PR #36181, opened 2023-09-03
Python 3.12.x

mkoeppe:
Continuation of:
- #34788

### :memo: Checklist

- [ ] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36176
- Depends on #36112
- Depends on #36123
- Depends on #36129
- Depends on #36477 
- Depends on #36444
- Depends on #36561
- Depends on #36869
- Depends on #36351
  - which depends on #36775
- Depends on #36945 (absorbed in #36351)
- Depends on #36868 (absorbed here)
- Depends on #37270 (split out from here)
- Depends on #36983
- Depends on #37834 (split out from here, to be merged)
- Depends on #37833 (split out from here)

Related / follow-ups:
- #37402
- https://github.com/yaml/pyyaml/pull/808
- #31295
- #35593

PR #36403, opened 2023-10-04
Fix AttributeError doctests when output includes a suggestion (part of python 3.12 support)

tornaria:
In python 3.12, attribute errors add a suggestion at the end of the
usual error message ("Did you mean ...?").
    
We add ... at the end of these doctest outputs to fix it.

This PR is split in two commits:    

 - The bulk of the changes, in the first commit, was obtained automatically with:
    ```
    git grep -l "AttributeError:" src | xargs sed -ie \
        's/^ *\(AttributeError: .*\)\?has no attribute.*$/&.../'
    ```
 - The second commit includes 3 changes not catched by the sed pattern above.

Since this changes a lot of files and is bound to get old, I'd appreciate a quick review + merge to avoid trouble (the PR is long but almost trivial and the review should be easy).

I'll soon make a separate PR (on top of this one) with the rest of the support python 3.12 changes, which are maybe not so trivial but much shorter than this one.

Related: #36181 

@mkoeppe :pray:


mkoeppe (approved):
LGTM.

vbraun:
merge conflict

PR #36407, opened 2023-10-05
Support python 3.12 on sagemath-standard

tornaria:
This PR adds support for running with python 3.12 from system. This has been tested on the python 3.12 branch of void linux (x86_64, x86_64-musl and i686).

The first two commits correspond to #36403. The rest is split is small pieces and I tried to add reasonable explanations in the commit messages. Reviewing by commit may be easier (ignoring the first two, already reviewed).

See also: #36181 

### :hourglass: Dependencies

#36403


kiwifb:
I was thinking about what was needed to enable 3.12 so it is timely. It is mostly doctest fix apart from that stuff with pylong. Can you expand on that stuff? It mentions some headers from python 3.12, naively I would have thought that was stuff removed from 3.12 that was still needed. If it is supposed to be internal, why do we need to use it?

tornaria:
> I was thinking about what was needed to enable 3.12 so it is timely. It is mostly doctest fix apart from that stuff with pylong. Can you expand on that stuff? It mentions some headers from python 3.12, naively I would have thought that was stuff removed from 3.12 that was still needed. If it is supposed to be internal, why do we need to use it?

As I mentioned in the commit message, the layout for python integers changed in python 3.12. This affects cypari2 and fpylll as well as sagemath, since all three projects implement functions to convert from/to python integers.

Since the change is forward and backward incompatible, I created a module `pycore_long` which implements some api to python integers. This is all declared in `pycore_long.pxd` and implemented in `pycore_long.h`. Note that these two files are shared verbatim with cypari2 (https://github.com/sagemath/cypari2/pull/138) and fpylll (https://github.com/fplll/fpylll/pull/254). Arguably this should be moved to cython itself (there is a cython module `cpython.longintrepr`) but for now this is the simpler solution.

I suggest looking at the raw files: in `pycore_long.pxd` there are a few functions declared, and this is all a user of this module needs to know. These declarations are meant to complement the ones in `cpython.longintrepr`.

In `pycore_long.h` is the implementation: one for python 3.12, one for < 3.12. The function names and prototypes, and the implementation for 3.12, is taken verbatim from cpython 3.12 source code. This way, if a later version of cpython decides to change again the representation of integers, we can just copy the corresponding code. The fall-back implementation for < 3.12 is mine, but you'll see everything is trivial (the function names are already very clear on intent, this is just "impedance matching" code).

The only function not taken from python is `ob_digit()` which is just an abstraction around getting the `ob_digit` struct member which they moved around.

As for the low level details: the main difference (besides moving the `ob_digit` member around) is that in the previous implementation they abused the size of the integer to place the sign. In the current implementation they added an additional member `lv_tag` to hold the sign. I expect they want to use `lv_tag` for other stuff in the future, presumably to have a "short" representation for integers that fit in one word.

tornaria:
rebased to beta6

tornaria:
FTR: before I was only testing this with cython 0.29 (this PR applied to sagemath 10.1).

Now I've tested this with cython 3:
 - cython 3.0.2 has a memory leak bug that causes three doctests to fail (this is fixed in https://github.com/cython/cython/pull/5725)
 - cython 3.0.3 includes this fix, but has another bug that breaks building scipy (this seems to be fixed in https://github.com/cython/cython/pull/5752 but I haven't tested that)

In summary, when actually updating python to 3.12 something has to be done with cython.

Everything seems to work ok when using cython 3.0.2 + https://github.com/cython/cython/pull/5725.


A further note: this PR is about *supporting* using python 3.12. Please review from that point of view: the important thing is that this makes sense and it doesn't cause any problem with older python.

I'd appreciate if this can be merged for 10.2 even if it might make more sense to wait a bit more to actually update the python in sage-the-distro. We are already shipping python 3.12 in void linux so this is well tested at least in the architectures we support.

mkoeppe (approved):
LGTM

src/sage/cpython/atexit.pyx
```diff
@@ -154,6 +154,10 @@ cdef extern from *:
     #undef _PyGC_FINALIZED
     #include "internal/pycore_interp.h"
     #include "internal/pycore_pystate.h"
+    #if PY_VERSION_HEX >= 0x030c0000
```
tobiasdiez, line 157: I don't have a deep understanding of these things, but reading the comments in `sage_setup/cython_options.py` it appears to me that the usage of PY_VERSION_HEX and PY_MAJOR_VERSION is discouraged. In fact, if I see this correctly than the code added in this PR is the only place now in sage that uses it. @tornaria what's your take on this

tornaria: Uh? This is C code, and `PY_VERSION_HEX` is part of the python stable ABI. See: https://docs.python.org/3/c-api/apiabiversion.html#c.PY_VERSION_HEX.

In fact, I don't think there's any code in sage source that uses `PY_VERSION_HEX` or `PY_MAJOR_VERSION` as defined in `sage_setup/cython_options.py`.

BTW, `git grep PY_VERSION_HEX` shows there are several uses of `PY_VERSION_HEX`, all in C code except for one use in `src/sage/cpython/dict_del_by_value.pyx` which I think it's easy to remove.

tobiasdiez: Thanks a lot. I got thrown off by the usage in `dict_del_by_value` and just did a naive "find all" search. You are right there is nothing wrong with using it in C code.

How would you "fix" `dict_del_by_value`?

dimpase:
Had "fun" with Python 3.12 :  https://github.com/python/cpython/issues/112282 
they hardcoded C stack limit, in the process they broke everything that relies on `sys.setrecursionlimit` and recursive calls of C modules, e.g. `@cache` decorator.

I really don't like the vibe of that change pushed through. https://github.com/python/cpython/issues/107263 and https://github.com/python/cpython/issues/112215

dimpase:
Sorry, why has this been merged without bumping the minimal acceptable version to 3.12.0 in `pkgs/python3/spkg-configure.m4` ? 
I am telling people on https://github.com/sagemath/sage/issues/36860 that it should just work! (while it can't).
Certainly one needs

```patch
--- a/build/pkgs/python3/spkg-configure.m4
+++ b/build/pkgs/python3/spkg-configure.m4
@@ -1,8 +1,8 @@
 SAGE_SPKG_CONFIGURE([python3], [
    m4_pushdef([MIN_VERSION],               [3.9.0])
    m4_pushdef([MIN_NONDEPRECATED_VERSION], [3.9.0])
-   m4_pushdef([LT_STABLE_VERSION],         [3.12.0])
-   m4_pushdef([LT_VERSION],                [3.12.0])
+   m4_pushdef([LT_STABLE_VERSION],         [3.13.0])
+   m4_pushdef([LT_VERSION],                [3.13.0])
    AC_ARG_WITH([python],
                [AS_HELP_STRING([--with-python=PYTHON3],
                                [Python 3 executable to use for the Sage venv; default: python3])])
```

mkoeppe:
Because the Sage distribution cannot support Python 3.12 yet.

dimpase:
> Because the Sage distribution cannot support Python 3.12 yet.

**This PR adds support for running with python 3.12 from system**

then wtf is this, pardon my French?

kiwifb:
sage 10.2 does run with python 3.12 in gentoo, with my overlay. I had to rush a few packages to support python 3.12 but the last one(s) where you need the overlay are gmpy/pplpy.

dimpase:
So it's all here, what's missing is the goodwill of @mkoeppe

mkoeppe:
> then wtf is this, pardon my French?

No, Dima, this is not pardoned. This is unwelcome aggression. https://github.com/sagemath/sage/blob/develop/CODE_OF_CONDUCT.md

mkoeppe:
@dimpase You seem to be confused about the distinction between the Sage library and the Sage distribution; just like 2 weeks ago in https://github.com/sagemath/sage/pull/36777#issuecomment-1833128275

Here in #36407, @tornaria did the necessary porting work for the Sage library.
Because he (and I assume @kiwifb) bring their own version of gmpy2, they can run the Sage library on Python 3.12.

dimpase:
@roed314 : @mkoeppe approved this PR, forgetting the version bump in spkg-configure.

Now my attempt to correct this omission is being shut down by @mkoeppe for no good reason.

A good case to look at by Sage-Abuse.

kiwifb:
Sorry for the noise, wrong ticket.

dimpase:
> @dimpase You seem to be confused about the distinction between the Sage library and the Sage distribution; just like 2 weeks ago in https://github.com/sagemath/sage/pull/36777#issuecomment-1833128275
> 
> Here in #36407, @tornaria did the necessary porting work for the Sage library.
> Because he (and I assume @kiwifb) bring their own version of gmpy2, they can run the Sage library on Python 3.12.
> 

it's yet another reason to split the two things - we are not obliged to read your mind at the same time as reading the title of the PR.

dimpase:
> > then wtf is this, pardon my French?
> 
> No, Dima, this is not pardoned. This is unwelcome aggression. https://github.com/sagemath/sage/blob/develop/CODE_OF_CONDUCT.md

I think you just don't understand that your hair-splitting 1-sentense reply was rude.

mkoeppe:
If this is your attempt to apologize -- that's far from sufficient.

tornaria:
Why is this PR "disputed" ? It's already merged. Please, let's not be childish.

Anyway, I came here to say that there is a typo in `pylong_core.h` which affects python 3.8. See: https://github.com/sagemath/cypari2/pull/146.

This file is in fpylll (https://github.com/fplll/fpylll/pull/254), cypari2 (https://github.com/sagemath/cypari2/pull/138), and sagemath.

dimpase:
This PR is a good example of sage the distro slowing down everything. @tornaria

tornaria:
> This PR is a good example of sage the distro slowing down everything. @tornaria

How so?

Python 3.12 was released oct 2, this PR was made oct 5, and it was merged oct 21.

In fact, this is a superb example of how things should work IMHO: we implement support for python 3.12 in sagelib, and sage the distro can move forward as it wishes.

The only thing I would like to be different is that requirements for sage-the-distro should not be enforced via requirements for sage-the-lib. Claiming that sage-the-lib supports python 3.12 and doesn't require sage_conf should not imply the same for sage-the-distro.

dimpase:
> > This PR is a good example of sage the distro slowing down everything. @tornaria
> 
> How so?
> 
> Python 3.12 was released oct 2, this PR was made oct 5, and it was merged oct 21.
> 
> In fact, this is a superb example of how things should work IMHO: we implement support for python 3.12 in sagelib, and sage the distro can move forward as it wishes.

as we have a monorepo, there is no well-defined border between sage the distro and sagelib.

it's perfectly possible to install sagelib on a system where all the external non-pypi deps are provided, by creating a venv, doing pip install <stuff>, where <stuff> is a relatively short list of packages, a dozen or two, and then
build sagelib by running
`./configure --enable-system-side-packages && make build` - but this only works if there are no upper bounds on PyPI package versions enforced by sage the distro.

In this case the PR title is misleading. Without a qualification it means that sage (the distro) accepts external python 3.12.

I.e., that's another reason for having sagelib in a separate repo.




> 
> The only thing I would like to be different is that requirements for sage-the-distro should not be enforced via requirements for sage-the-lib. Claiming that sage-the-lib supports python 3.12 and doesn't require sage_conf should not imply the same for sage-the-distro.

mkoeppe:
> this only works if there are no upper bounds on PyPI package versions enforced by sage the distro.

The Sage distribution (without the experimental `--enable-system-site-packages`) does not use any of the version bounds set in `build/pkgs/*/install-requires.txt`.
These files are there for uses of the Sage library outside of the Sage distribution, and now also for `--enable-system-site-packages`.

dimpase:
> > this only works if there are no upper bounds on PyPI package versions enforced by sage the distro.
> 
> The Sage distribution (without the experimental `--enable-system-site-packages`) does not use any of the version bounds set in `build/pkgs/*/install-requires.txt`. These files are there for uses of the Sage library outside of the Sage distribution, and now also for `--enable-system-site-packages`.

In this particular case, we talk about the version of accepted Python - not a PyPI package - which is controlled by
`LT_VERSION` in its `spkg-configure.m4`, not by an `install-requires.txt`.

One should have bumped `LT_VERSION` to 3.13, but it hasn't happen.

No experimental features,  `--enable-system-site-packages` etc not relevant here - except that it prevents the use of Python 3.12 in the way I outlined in the previous comment.

mkoeppe:
> One should have bumped `LT_VERSION` to 3.13, but it hasn't happen.

Nope, as I have explained several times already above.

dimpase:
> > One should have bumped `LT_VERSION` to 3.13, but it hasn't happen.
> 
> Nope, as I have explained several times already above.

As I explained at least twice, the PR title is misleading. You could have avoided this mess by adding "partially" there, but you preferred not to, for reasons unclear.

mkoeppe:
> Why is this PR "disputed"

@tornaria I think you can see the pattern.

tornaria:
> > In fact, this is a superb example of how things should work IMHO: we implement support for python 3.12 in sagelib, and sage the distro can move forward as it wishes.
> 
> as we have a monorepo, there is no well-defined border between sage the distro and sagelib.

This is not about the monorepo, but about having the same version number for everything (sage-the-distro, sage-setup, sagemath-standard, etc). It's not modular if everything must share the same version number.

The monorepo makes some things more difficult and some things easier.

> build sagelib by running `./configure --enable-system-side-packages && make build`

For pure sage-the-library see https://github.com/sagemath/sage/pull/37138#issuecomment-1904824866

> In this case the PR title is misleading. Without a qualification it means that sage (the distro) accepts external python 3.12.

Not my goal with this PR.

In any case, this PR is done and merged. With no regrets afaict.

tornaria:
> > Why is this PR "disputed"
> 
> @tornaria I think you can see the pattern.

This PR is done and merged. With no regrets afaict.

mkoeppe:
> This PR is done and merged. With no regrets afaict.

That's right, there's nothing wrong with this PR. 

"Disputed" is a warning label. It's marked with this label because the discussion in it is affected by abusive conduct by a participant in the discussion.

dimpase:
There is nothing abusive in my conduct here. Your conduct here is abusive, yes. You just cannot stand any criticism, that is the problem, you immediately consider it "abuse".

tornaria:
> > This PR is done and merged. With no regrets afaict.
> 
> That's right, there's nothing wrong with this PR.
> 
> "Disputed" is a warning label. It's marked with this label because the discussion in it is affected by abusive conduct by a participant in the discussion.

No. The label applies to the PR (my PR, btw), and it means the PR is disputed (as in: some reviewers think it's ok, some reviewers think it's not). I already changed the title of the PR, is it ok now?

I don't condone @dimpase outbursts as much as I disapprove the witch hunt.

tornaria:
Could I be given the powers required to close PR discussions, etc?

mkoeppe:
> The label applies to the PR (my PR, btw), and it means the PR is disputed (as in: some reviewers think it's ok, some reviewers think it's not).

I agree that it would be great if the "disputed PRs" were just disagreements of the kind that you describe. But that's unfortunately not true.

In any case, nobody's going to revert this PR because of the label. It's just a warning label -- and a cross-referencing tool for those who are interested in seeing the full picture of the pattern of abusive conduct. and the damage that it has already and continues to inflict.

dimpase:
@mkoeppe - please stop insinuating here and elsewhere about "abusive conduct". 

You can't take the any criticism of your work, it's  obvious, and you are driving the project into the ground by forcing everyone to work out your grand ideas of a conda competitor.

mkoeppe:
I'm not "insinuating". 
I'm "calling out" the abuse that is happening in plain sight.

mkoeppe:
For readers interested in this concept -- https://edib.harvard.edu/calling-and-calling-out-guide

roed314:
This is not an appropriate place for this discussion.  I've deleted several comments.

dimpase:
> > > Why is this PR "disputed"
> > 
> > 
> > @tornaria I think you can see the pattern.
> 
> This PR is done and merged. With no regrets afaict.

my problem with this PR was that it was only possible to understand what it does, and  what it does not do, by reading the diffs. I assumed that with it one can use `./configure` to use Python 3.12, as this was always, or almost always the default, when a PR said "Support Foobar version 42", that now a system-installed Foobar 42 is accepted by `./configure`.

In turn, I confused a number of other people by pointing at this PR and saying that now Python 3.12 is fully supported (see https://github.com/sagemath/sage/pull/36407#issuecomment-1852664647).

Thus, the insufficient detail in the PR description, and a break of the usual Sage PR conventions by it, led to a wtf, here, and not only here, I suppose.

roed314:
@dimpase Is there a followup issue or PR for adding the missing functionality to `./configure`?

mkoeppe:
@roed314 He opened #36868, which is now superseded by another PR as noted therein.

tornaria:
> Thus, the insufficient detail in the PR description, and a break of the usual Sage PR conventions by it, led to a wtf, here, and not only here, I suppose.

My apologies. I will try to be more clear in future PRs. Could you suggest a good title that communicates correctly? Some ideas:
 - Support python 3.12 on sagemath-standard
 - sagemath-standard: support for python 3.12
 - support for python 3.12 on sagelib
 - ...

I'm open to any suggestions. There's three levels of "support" to take into account:

1. "support" the version in `sagemath-standard` (i.e. the pypi distribution builds and checks)
2. "accept" the version in `./configure` if installed in system (what you understood)
3. "ship" the version as a spkg to use if not installed in system

Please suggest better replacements for the words I quoted above. I only care about (1) and I have been using "support" in that sense.

dimpase:
To my ear, "support" is stronger than "accept".
You can't support things without accepting them.
E.g. you cannot support constructive maths without accepting it as maths.

As it's mostly about python 3.12-compatible python longs, a title saying this explicitly might work better. E.g.

"python 3.12-compatible long ints and related sagelib issues"

And then, in the body, "with this, sagelib is ready for Python 3.12".

Or without getting technical in the title: "making sagelib ready for Python 3.12".

antonio-rojas:
We just upgraded to 3.12. I'm experiencing a segfault in `libs/gap/element.pyx` at
```
sage: for i in range(100):
    # compute the sum in GAP
    _ = libgap.Sum(rnd)
    try:
        libgap.Sum(*rnd)
    except ValueError:
        pass ## line 2490 ##
```
with backtrace
```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=11, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007aea958ab393 in __pthread_kill_internal (signo=11, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007aea9585a6c8 in __GI_raise (sig=sig@entry=11) at ../sysdeps/posix/raise.c:26
#3  0x00007aea955be2f9 in sigdie (sig=sig@entry=11, s=s@entry=0x7aea955c6f78 "Unhandled SIGSEGV: A segmentation fault occurred.")
    at build/src/cysignals/implementation.c:713
#4  0x00007aea955c267d in sigdie_for_sig (inside=0, sig=11) at build/src/cysignals/implementation.c:256
#5  cysigs_signal_handler (sig=11) at build/src/cysignals/implementation.c:361
#6  <signal handler called>
#7  0x00007ae9ba6a5d60 in _Py_IsImmortal (op=0x0) at /usr/include/python3.12/object.h:243
#8  Py_DECREF (op=0x0) at /usr/include/python3.12/object.h:701
#9  Py_XDECREF (op=0x0) at /usr/include/python3.12/object.h:799
#10 __pyx_pf_4sage_4libs_3gap_7element_19GapElement_Function_2__call__ (__pyx_v_self=__pyx_v_self@entry=0x7ae9b3e90a00, 
    __pyx_v_args=__pyx_v_args@entry=0x7ae9b57a0400) at build/cythonized/sage/libs/gap/element.c:26540
#11 0x00007ae9ba6a6bab in __pyx_pw_4sage_4libs_3gap_7element_19GapElement_Function_3__call__ (__pyx_v_self=0x7ae9b3e90a00, __pyx_args=0x7ae9b57a0400, 
    __pyx_kwds=<optimized out>) at build/cythonized/sage/libs/gap/element.c:26110
#12 0x00007aea95bc1949 in _PyObject_Call (tstate=0x7aea96031788 <_PyRuntime+459656>, callable=0x7ae9b3e90a00, args=0x7ae9b57a0400, 
    kwargs=<optimized out>) at Objects/call.c:367
#13 0x00007aea95a8d591 in PyCFunction_Call (kwargs=0x0, args=0x7ae9b57a0400, callable=0x7ae9b3e90a00) at Objects/call.c:387
```
and two hangs in `interfaces/octave.py` and `rings/integer.pyx`, related to handling of big integers:
```
Tests run before process (pid=1709947) timed out:
sage: t = '"{}"'.format(10^10000) ## line 178 ##
```
```
sage: a = Integer(19)**100000  # long time ## line 6937 ##
sage: c = a.inverse_mod(a*a)   # long time ## line 6938 ##
```

Any ideas?

dimpase:
I think this libgap error with python 3.12 has been reported already.

The trace looks scary - seems to be multithreading going on in Python? @tornaria

tornaria:
The libgap error is in #37026, but I don't think I've been able to reproduce it myself.

The other two I don't think I've seen before; the second one just works, and the third one raises `ZeroDivisionError(...)` as expected.

kiwifb:
I did assume that it was one of those errors caused by upgrading to 
gap-4.13.0 but it is indeed a python 3.12 error. The test pass with 
python 3.11 and segfault with python 3.12.

antonio-rojas:
I can reproduce the hang in plain Python
```
>>> import sys; sys.set_int_max_str_digits(0); 10**10000
***hangs***
^C
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/_pylong.py", line 85, in int_to_decimal_string
    return str(int_to_decimal(n))
               ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_pylong.py", line 77, in int_to_decimal
    result = inner(n, n.bit_length())
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_pylong.py", line 64, in inner
    return inner(lo, w2) + inner(hi, w - w2) * w2pow(w2)
           ^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_pylong.py", line 64, in inner
    return inner(lo, w2) + inner(hi, w - w2) * w2pow(w2)
           ^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_pylong.py", line 64, in inner
    return inner(lo, w2) + inner(hi, w - w2) * w2pow(w2)
           ^^^^^^^^^^^^^
  [Previous line repeated 6 more times]
  File "/usr/lib/python3.12/_pylong.py", line 45, in w2pow
    result = D2**w
             ~~^^~
  File "/usr/lib/python3.12/_pydecimal.py", line 2339, in __pow__
    ans = self._power_exact(other, context.prec + 1)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/_pydecimal.py", line 2187, in _power_exact
    if xc > 10**p:
            ~~^^~
KeyboardInterrupt
```
The `_pylong` module is indeed new in 3.12: https://github.com/python/cpython/commit/de6981680bcf6496e5996a853b2eaa700ed59b2c

tornaria:
> I can reproduce the hang in plain Python
> 
> ```
> >>> import sys; sys.set_int_max_str_digits(0); 10**10000
> ***hangs***
> ^C
>   File "<stdin>", line 1, in <module>
>   File "/usr/lib/python3.12/_pylong.py", line 85, in int_to_decimal_string
>     return str(int_to_decimal(n))
>                ^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/_pylong.py", line 77, in int_to_decimal
>     result = inner(n, n.bit_length())
>              ^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/_pylong.py", line 64, in inner
>     return inner(lo, w2) + inner(hi, w - w2) * w2pow(w2)
>            ^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/_pylong.py", line 64, in inner
>     return inner(lo, w2) + inner(hi, w - w2) * w2pow(w2)
>            ^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/_pylong.py", line 64, in inner
>     return inner(lo, w2) + inner(hi, w - w2) * w2pow(w2)
>            ^^^^^^^^^^^^^
>   [Previous line repeated 6 more times]
>   File "/usr/lib/python3.12/_pylong.py", line 45, in w2pow
>     result = D2**w
>              ~~^^~
>   File "/usr/lib/python3.12/_pydecimal.py", line 2339, in __pow__
>     ans = self._power_exact(other, context.prec + 1)
>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/_pydecimal.py", line 2187, in _power_exact
>     if xc > 10**p:
>             ~~^^~
> KeyboardInterrupt
> ```
> 
> The `_pylong` module is indeed new in 3.12: [python/cpython@de69816](https://github.com/python/cpython/commit/de6981680bcf6496e5996a853b2eaa700ed59b2c)

Can't reproduce this either. Tried with 3.12.3 (current) and 3.12.2. What's your version of python?

dimpase:
_Py_Isimmortal, which we see in the backtrace, is new in Python 3.12.
So this might be something to fix in cysignals, or
in Sage guts somewhere 

https://peps.python.org/pep-0683/

antonio-rojas:
> _Py_Isimmortal, which we see in the backtrace, is new in Python 3.12. So this might be something to fix in cysignals, or in Sage guts somewhere

But `op` is a nullptr long before reaching `_Py_Isimmortal`, so this may be a previously existing issue that only now is being detected by `_Py_Isimmortal` trying to dereference it.

antonio-rojas:
Calling any libgap function in three arguments triggers the crash.
```
sage: libgap.AbelianGroup(0,0,0)
-> crash
```

dimpase:
Can you attach your `sage/libs/gap/element.c` here?

antonio-rojas:
> Can you attach your `sage/libs/gap/element.c` here?

[element.tar.gz](https://github.com/sagemath/sage/files/15080499/element.tar.gz)

tornaria:
> libgap.AbelianGroup(0,0,0)

Can't reproduce this one either:
```
$ sage
│ SageMath version 10.3, Release Date: 2024-03-19                    │
│ Using Python 3.12.3. Type "help()" for help.                       │
sage: libgap.AbelianGroup(0,0,0)
Cell In[1], line 1
----> 1 libgap.AbelianGroup(Integer(0),Integer(0),Integer(0))

File /usr/lib/python3.12/site-packages/sage/libs/gap/element.pyx:2514, in sage.libs.gap.element.GapElement_Function.__call__ (build/cythonized/sage/libs/gap/element.c:26294)()
   2512 try:
   2513     sig_GAP_Enter()
-> 2514     sig_on()
   2515     if n == 0:
   2516         result = GAP_CallFunc0Args(self.value)

GAPError: Error, usage: AbelianGroup( [<filter>, ]<ints> )
sage: 
```

FWIW, we are on gap 4.12.2, cython 3.0.10, cysignals 1.11.4, ... (what else would be relevant?)

We don't have mpdecimal.

dimpase:
> Calling any libgap function in three arguments triggers the crash.
Do you mean, any libgap function which does not take 3 arguments?

> 
> ```
> sage: libgap.AbelianGroup(0,0,0)
> -> crash
> ```
same crash place?

Probably, to reproduce one needs to know compiler version, cython version, maybe even glibc and pthreads version, cysignals version...

tornaria:
For me gcc is 13.2.0, glibc is 2.39 (and gap 4.12.2, cython 3.0.10, cysignals 1.11.4, ... )

Unrelated, but how do you deal with #37616 if you are on gap 4.13?

antonio-rojas:
> same crash place?
> 
> Probably, to reproduce one needs to know compiler version, cython version, maybe even glibc and pthreads version, cysignals version...

Yes, same backtrace.
GCC 13.2.1, cython 3.0.10, glibc 2.39, cysignals 1.11.4.
Tried rebuilding python and sage without any Arch specific build flags, same result.

What baffles me is that the definition of `Py_XDECREF` is
```
static inline void Py_XDECREF(PyObject *op)
{
    if (op != _Py_NULL) {
        Py_DECREF(op);
    }
}
```
so this part of the backtrace doesn't make any sense to me:
```
#8  Py_DECREF (op=0x0) at /usr/include/python3.12/object.h:701
#9  Py_XDECREF (op=0x0) at /usr/include/python3.12/object.h:799
```

antonio-rojas:
> For me gcc is 13.2.0, glibc is 2.39 (and gap 4.12.2, cython 3.0.10, cysignals 1.11.4, ... )
> 
> Unrelated, but how do you deal with #37616 if you are on gap 4.13?

I ignore the test failures. None of them is a serious issue.

antonio-rojas:
Narrowed this down to GCC optimization. Compiling with `-O2` or higher triggers this, compiling with `-O1` works fine. This would explain why this is only reproducible in some distros.

antonio-rojas:
> Is this compiling libgap? or element.pyx ?
> 
> If this is compiling libgap with different -O values then it's GAP bug.

It's compiling the previously attached `element.c`

tornaria:
> Narrowed this down to GCC optimization. Compiling with `-O2` or higher triggers this, compiling with `-O1` works fine. This would explain why this is only reproducible in some distros.
> 
> Compiling with clang works OK with any optimization level.

This is our cc line for that file: 
`
cc -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O3 -Wall -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe -g -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe -g -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe -ffile-prefix-map=/builddir/sagemath-10.4.beta3/pkgs/sagemath-standard=. -fPIC -Isage/cpython -I/usr/lib/python3.12/site-packages/cysignals -I/builddir/sagemath-10.4.beta3/pkgs/sagemath-standard -I/usr/lib/python3.12/site-packages/numpy/core/include -I/usr/include/python3.12 -Ibuild/cythonized -I/usr/include/python3.12 -c build/cythonized/sage/libs/gap/element.c -o build/temp.linux-x86_64-cpython-312/build/cythonized/sage/libs/gap/element.o -fno-strict-aliasing -DCYTHON_CLINE_IN_TRACEBACK=1
`

dimpase:
> What baffles me is that the definition of `Py_XDECREF` is
> [...]

```C
// Static inline functions should use _Py_NULL rather than using directly NULL
// to prevent C++ compiler warnings. On C23 and newer and on C++11 and newer,
// _Py_NULL is defined as nullptr.
#if (defined (__STDC_VERSION__) && __STDC_VERSION__ > 201710L) \
        || (defined(__cplusplus) && __cplusplus >= 201103)
#  define _Py_NULL nullptr
#else
#  define _Py_NULL NULL
#endif
```
`Py_NULL` and `NULL` are the same thing, IMHO. So probably some kind of optimisation causes a corruption of call frames.

 Something to do with `_Py_IsImmortal` - `Py_XDECREF` and `Py_DECREF` are kicking the can down the road to `_Py_IsImmortal`  for some reason.

Can you insert prints, or otherwise  see the actual values of `op` at the calling moments, not post mortem?

Also, can you add print in the body of the `sig_GAP_Enter` macro (defined close to the top of element.c)
```
#define sig_GAP_Enter()  {int t = GAP_Enter(); if (!t) sig_error();}
```
to see what value of `t` we're getting from this `libgap.Sum(1,2,3)`, or whatever the crash's trigger is.

Maybe `t` must be declared `volatile` there? (to prevent it being optimized away)

Of course, sometimes print statements cause sudden changes in how the code operates, but let's try anyway...

dimpase:
something like 
```diff
--- a/src/sage/libs/gap/gap_includes.pxd
+++ b/src/sage/libs/gap/gap_includes.pxd
@@ -29,7 +29,7 @@ cdef extern from "gap/calls.h" nogil:
 
 cdef extern from "gap/libgap-api.h" nogil:
     """
-    #define sig_GAP_Enter()  {int t = GAP_Enter(); if (!t) sig_error();}
+    #define sig_GAP_Enter()  {volatile int t = GAP_Enter(); if (!t) sig_error();}
     """
     ctypedef void (*GAP_CallbackFunc)()
     void GAP_Initialize(int argc, char ** argv,
```

antonio-rojas:
This is a regression in the GCC stable branch. I downgraded to the previous Arch snapshot (commit 860b0f0ef787f756c0e293671b4c4622dff63a79) and it works fine. I will try to create a minimal test case and report upstream, so far my naive attempts are not working.

antonio-rojas:
Upstream is understandably having trouble diagnosing the issue, with such a complex setup needed to reproduce it, and I am not able to help much further. If someone more knowledgeable could chime in and help make this more digestable for upstream, that would be helpful.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114872

antonio-rojas:
Let's move the discussion to https://github.com/sagemath/sage/issues/37026 since this is not directly caused by Python 3.12 (it only makes the issue manifest itself as a crash, but likely still exists on 3.11)

dimpase:
> Calling any libgap function in three arguments triggers the crash.
> 
> ```
> sage: libgap.AbelianGroup(0,0,0)
> -> crash
> ```

I am able to reproduce this on Gentoo (with Python 3.12.3, and gcc version 13.2.1 20240210 (Gentoo 13.2.1_p20240210 p14)). 

In view of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114872#c12 it's of course a tough task to produce a "small" example (it could be that gcc code generation only breaks down on huge files like element.c)

dimpase:
> This is a regression in the GCC stable branch. I downgraded to the previous Arch snapshot (commit 860b0f0ef787f756c0e293671b4c4622dff63a79) and it works fine. I will try to create a minimal test case and report upstream, so far my naive attempts are not working.

Upstream has concluded it's not a gcc bug. It's actually their optimiser getting more efficient, holding more local variables in registers. So on can either go the route of adding `volatile` all over the place (#37951), or, better, do not blindly create ref-counted GAP objects, which need postprocessing at `sign_on/off` recovery, and use as much as possible temporary GAP objects (currently not supported by out interface, but it's certainly doable).

PR #36408, opened 2023-10-05
refine the category of free groups and their quotients

fchapoton:
As these groups are finitely generated.

### :memo: Checklist

### DIFF
--- a/src/sage/groups/finitely_presented.py
+++ b/src/sage/groups/finitely_presented.py
@@ -130,7 +130,7 @@ AUTHOR:
 
 from sage.arith.misc import GCD as gcd
 from sage.categories.morphism import SetMorphism
-from sage.functions.generalized import sign
+from sage.categories.groups import Groups
 from sage.groups.free_group import FreeGroup
 from sage.groups.free_group import FreeGroupElement
 from sage.groups.group import Group
@@ -144,6 +144,7 @@ from sage.rings.polynomial.laurent_polynomial_ring import LaurentPolynomialRing
 from sage.rings.rational_field import QQ
 from sage.sets.set import Set
 from sage.structure.unique_representation import UniqueRepresentation
+
 
 class GroupMorphismWithGensImages(SetMorphism):
     r"""
@@ -804,6 +805,10 @@ class FinitelyPresentedGroup(GroupMixinLibGAP, UniqueRepresentation, Group, Pare
         self._free_group = free_group
         self._relations = relations
         self._assign_names(free_group.variable_names())
+        if category is None:
+            category = Groups().FinitelyGenerated()
+        else:
+            category &= Groups().FinitelyGenerated()
         parent_gap = free_group.gap() / libgap([rel.gap() for rel in relations])
         ParentLibGAP.__init__(self, parent_gap)
         Group.__init__(self, category=category)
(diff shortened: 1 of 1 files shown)


fchapoton:
not clear if this passes the CI or not ; the CI seems badly borked

src/sage/groups/finitely_presented.py
```diff
@@ -804,6 +808,10 @@ def __init__(self, free_group, relations, category=None):
         self._free_group = free_group
         self._relations = relations
         self._assign_names(free_group.variable_names())
+        if category is None:
+            category = Groups().FinitelyGenerated()
+        else:
+            category &= Groups().FinitelyGenerated()
         parent_gap = free_group.gap() / libgap([rel.gap() for rel in relations])
```
kwankyu, line 812: I agree with `if` part, but if the category is given by the user, shouldn't we use it as given, instead of joining with `Groups().FinitelyGenerated()`? Do you have a reason for doing that?

src/sage/groups/free_group.py
```diff
@@ -775,7 +778,7 @@ def __init__(self, generator_names, libgap_free_group=None):
         if not generator_names:
             cat = Groups().Finite()
         else:
-            cat = Groups().Infinite()
+            cat = Groups().Infinite().FinitelyGenerated()
         Group.__init__(self, category=cat)
```
kwankyu: Is there no *infinitely-generated* free group in GAP?

src/sage/groups/finitely_presented.py
fchapoton: well, this is adding information, so there is no harm and this could be useful

kwankyu: As far as I know, the idiom is just to pass the user-supplied category on.

> this is adding information, ...

I imagine that someone might not want the additional information...

But really I don't have a substantial argument. So OK if you insist.

src/sage/groups/free_group.py
kwankyu: I found this from the GAP documentation:

> Called in the fourth form, FreeGroup returns a free group on infinitely many generators. The optional argument name must be a string; its default value is "f", and the optional argument init must be a finite list of nonempty strings; its default value is an empty list. The generators are initially labelled according to the list init, followed by namei for each i in the range from Length(init)+1 to infinity.

Source: https://docs.gap-system.org/doc/ref/chap37_mj.html#X8215999E835290F0

Not relevant here?

fchapoton: As far as I can tell, our `FreeGroup` class cannot create free groups with infinitely many generators. If fails by trying to make the list of all names. For example, try `FreeGroup(NN)`

kwankyu: Yes. It seems that `FreeGroup_class` wraps only finitely-generated free groups of GAP, though GAP itself supports infinitely-generated free groups.

kwankyu:
```
sage -t src/sage/groups/cubic_braid.py  # 1 doctest failed
sage -t src/sage/groups/braid.py  # Timed out
sage -t src/sage/graphs/generators/families.py  # 1 doctest failed
sage -t src/sage/groups/artin.py  # Timed out
sage -t src/sage/categories/homset.py  # 2 doctests failed
sage -t src/sage/groups/libgap_wrapper.pyx  # 1 doctest failed
sage -t src/sage/groups/free_group.py  # 1 doctest failed
sage -t src/sage/groups/raag.py  # Timed out  
```

fchapoton:
We cannot hope for an iterator, as (citing gap documentation) : There exist finitely presented groups for which no algorithm exists (it is known that no such algorithm can exist) that will tell for two arbitrary words in the generators whether the corresponding elements in the FpGroup are equal.

fchapoton:
I have decied to refactor the pull request into something simpler. There remains only a tiny change to the category. Other changes are cosmetic code details.

fchapoton:
Damn.. It seem that for instance the Artin groups of type D4 cannot be made into an enumerated set either. Maybe I should drop the pull request completely.

fchapoton:
I would say that there is a flaw in our categorical framework : it is wrong that a finitely generated semigroup is an enumerated set, as this require to be able to check equality of elements, which may be undecidable..

tscrim:
We might have to refine the categories at bit or put some other kind of hold.

Side comment: IIRC all finite Artin groups are known to have solvable word problems by the Garside structure (like for the braid group coming from lifting the long element). So the type $D_4$ Artin group should have a solvable word problem.

PR #36409, opened 2023-10-05
Interface to HoG (House of Graphs) database

darijgr:
This provides support for retrieving graphs from the House of Graphs ( https://houseofgraphs.org/ ), along with their computed invariants and embeddings. No search functionality is provided here (other than lookup by ID); the user is supposed to search on the HoG website and then use this interface to import the graph into Sage. (Further functionality could be done in future tickets.)

Alas I have not had the time to write fake objects for offline testing.


darijgr:
What is going on with the Build & Test? Why is it failing? Is there a way to see the log?

PR #36367, opened 2023-09-29
Simplify experimental all-conda installation instructions via `pkgs/sage-conf_conda`

mkoeppe:
This is a version of sage-conf for the experimental all-conda installation method.

It calls `configure` with the whole long set of options, which simplifies the instructions for the user.

The new instructions use editable mode (PEP 660) for `pkgs/sage-conf_conda`, but it also works to install it in non-editable mode.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36411, opened 2023-10-05
Fix conda install instructions after `setuptools` update

mkoeppe:
- Updates install instructions for use with modern `setuptools` versions.
- Restore `bootstrap` step in instructions, lost in #36367 
- Unpin `setuptools`.

### :memo: Checklist

### DIFF
--- a/.github/workflows/ci-conda.yml
+++ b/.github/workflows/ci-conda.yml
@@ -92,7 +92,7 @@ jobs:
         run: |
           # Use --no-deps and pip check below to verify that all necessary dependencies are installed via conda.
           pip install --no-build-isolation --no-deps -v -v -e ./pkgs/sage-conf ./pkgs/sage-setup
-          pip install --no-build-isolation --no-deps -v -v -e ./src
+          pip install --no-build-isolation --no-deps --config-settings editable_mode=compat -v -v -e ./src
         env:
           SAGE_NUM_THREADS: 2
         
--- a/build/pkgs/setuptools/distros/conda.txt
+++ b/build/pkgs/setuptools/distros/conda.txt
@@ -1,4 +1,1 @@
-# Set this bound until https://github.com/sagemath/sage/issues/34209 adds support for PEP660 editable builds
-# By setting this version bound, we avoid having to include the following in our installation instructions.
-#   export SETUPTOOLS_ENABLE_FEATURES=legacy-editable
-"setuptools<64"
+setuptools
--- a/build/pkgs/setuptools_wheel/distros
+++ b/build/pkgs/setuptools_wheel/distros
@@ -1,0 +1,1 @@
+../setuptools/distros
--- a/build/pkgs/setuptools_wheel/distros/repology.txt
+++ b/build/pkgs/setuptools_wheel/distros/repology.txt
@@ -1,1 +1,0 @@
-python:setuptools
--- a/pkgs/sage-conf_conda/setup.py
+++ b/pkgs/sage-conf_conda/setup.py
@@ -30,22 +30,22 @@ class build_py(setuptools_build_py):
 
         if os.path.exists(os.path.join(SAGE_ROOT, 'config.status')):
             print(f'Reusing configured SAGE_ROOT={SAGE_ROOT}')
-        else:
-            cmd = f"cd {SAGE_ROOT} && ./configure --enable-build-as-root --with-system-python3=force --disable-notebook --disable-sagelib --disable-sage_conf --disable-doc"
-            cmd += ' --with-python=$CONDA_PREFIX/bin/python --prefix="$CONDA_PREFIX"'
-            cmd += ' $(for pkg in $(PATH="build/bin:$PATH" build/bin/sage-package list :standard: --exclude rpy2 --has-file spkg-configure.m4 --has-file distros/conda.txt); do echo --with-system-$pkg=force; done)'
-            print(f"Running {cmd}")
-            sys.stdout.flush()
-            if os.system(cmd) != 0:
-                if os.path.exists(os.path.join(SAGE_ROOT, 'config.status')):
-                    print("Warning: A configuration has been written, but the configure script has exited with an error. "
-                          "Carefully check any messages above before continuing.")
-                else:
-                    print(f"Error: The configure script has failed; this may be caused by missing build prerequisites.")
-                    sys.stdout.flush()
-                    PREREQ_SPKG = "_prereq bzip2 xz libffi"  # includes python3 SPKG_DEPCHECK packages
-                    os.system(f'cd {SAGE_ROOT} && export PACKAGES="$(build/bin/sage-get-system-packages conda {PREREQ_SPKG})" && [ -n "$PACKAGES" ] && echo "You can install the required build prerequisites using the following shell command" && echo "" && build/bin/sage-print-system-package-command conda --verbose --sudo install $PACKAGES && echo ""')
-                    raise SetupError("configure failed")
+
+        cmd = f"cd {SAGE_ROOT} && ./configure --enable-build-as-root --with-system-python3=force --disable-notebook --disable-sagelib --disable-sage_conf --disable-doc"
+        cmd += ' --with-python=$CONDA_PREFIX/bin/python --prefix="$CONDA_PREFIX"'
+        cmd += ' $(for pkg in $(PATH="build/bin:$PATH" build/bin/sage-package list :standard: --exclude rpy2 --has-file spkg-configure.m4 --has-file distros/conda.txt); do echo --with-system-$pkg=force; done)'
+        print(f"Running {cmd}")
+        sys.stdout.flush()
+        if os.system(cmd) != 0:
+            if os.path.exists(os.path.join(SAGE_ROOT, 'config.status')):
+                print("Warning: A configuration has been written, but the configure script has exited with an error. "
+                      "Carefully check any messages above before continuing.")
+            else:
+                print(f"Error: The configure script has failed; this may be caused by missing build prerequisites.")
+                sys.stdout.flush()
+                PREREQ_SPKG = "_prereq bzip2 xz libffi"  # includes python3 SPKG_DEPCHECK packages
+                os.system(f'cd {SAGE_ROOT} && export PACKAGES="$(build/bin/sage-get-system-packages conda {PREREQ_SPKG})" && [ -n "$PACKAGES" ] && echo "You can install the required build prerequisites using the following shell command" && echo "" && build/bin/sage-print-system-package-command conda --verbose --sudo install $PACKAGES && echo ""')
+                raise SetupError("configure failed")
 
         # In this mode, we never run "make".
 
--- a/pkgs/sage-conf_conda/setup.py
+++ b/pkgs/sage-conf_conda/setup.py
@@ -28,9 +28,6 @@ class build_py(setuptools_build_py):
             raise SetupError('No conda environment is active. '
                              'See https://doc.sagemath.org/html/en/installation/conda.html on how to get started.')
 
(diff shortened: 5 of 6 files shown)


tobiasdiez:
There are now build errors.

mkoeppe:
possibly from Cython 3.0.3

mkoeppe:
Yes, that did the trick.

Remaining error in pip check: "sagemath-standard 10.2b5 has requirement ipython<8.9.0,>=7.13.0, but you have ipython 8.16.1."

build/pkgs/cython/distros/conda.txt
```diff
@@ -1 +1 @@
-cython
+cython==3.0.2
```
tobiasdiez: Wouldn't it be better to exclude the broken 3.0.3 instead of fixing the version (here and in install-requires)? From what I read, the bug should be fixed in 3.0.4.

mkoeppe: No, we have to expect further breakage in the 3.0.x series

tobiasdiez: Any basis for this expectation? Looking at https://github.com/sagemath/sage/pull/36110, there was no problem upgrading  from 3.0.0 to 0.1 and 0.2.

I think we can always exclude newer versions once they are known to break something...

mkoeppe: > we can always exclude newer versions once they are known to break something...

I think I have explained this to you several times in the past already. This type of breakage affects releases that we have already made retroactively. As our project does not make bug fix releases, we *cannot* fix such breakage after it happens.

So when our experience indicates that we have to expect breakage from updates, it is prudent to put in suitable version constraints.

The breakage with 3.0.3 affected a major project (scipy), so this is an indication that the Cython project is currently struggling with release management. This is understandable because after the rollout of Cython 3 in distributions, there is likely very high demand for developer attention as lots of projects are forced to add support for Cython 3.

tobiasdiez: > So when our experience indicates that we have to expect breakage from updates, it is prudent to put in suitable version constraints.

Yes, that's why you put in the major version constraint for cython. Bugs in other projects can always happen, but that's not a valid reason to fix the version. (A better solution in the form of a lock file is waiting for review: https://github.com/sagemath/sage/pull/35986)

For example, if we release sage with this version constraint, then it is no longer possible to use it in a conda environment with a newer version of cython. But someone might need 3.0.4 because earlier versions break of them.

So please loosen the constraint again. Thanks

mkoeppe: > that's not a valid reason to fix the version.

Sorry, Tobias, you don't get to declare that.

tobiasdiez: By that logic you would need to introduce a lot more fixed version constraints for a lot more packages. A lot of our dependencies have weaker ci than cython and thus may introduce new bugs.

For me these version constraints say "we expect that sage is compatible with these versions". Since that range usual contains future releases, it implicitly has the addition "of course, only given that the dependency doesn't introduce breaking changes/bugs". 

From what I can see the cython issue that affects us is likely fixed in 3.0.4 (it already has patches discussed etc). So until you have evidence that there might be more bugs that affect us, I have the expectation that 3.0.4 works.

mkoeppe: It's not helpful to repeat the same points every time that a version constraint has to be set. I'm fully aware of any and all of these points, and they are already factored in.

tobiasdiez: > I'm fully aware of any and all of these points, and they are already factored in.

Perfect! In that case you might want to point out where our evaluation differs, since apparently we come to a different end result. For example, you might want to start with discussing why we didn't needed a fixed version before but now we do. For all I can see the cython 3 gets more stable and stable, so there are less and less bugs expected. By this logic, we would need a fixed version less and less. 

Would the following compromise work for you: we simply exclude 3.0.3 for now, and if 3.0.4 breaks things again, we introduce an upper version constraint `<= 3.0.2`?

tobiasdiez:
Matthias, may I ask you not to mark discussions as resolved just because you are not agreeing with the points being brought forward.

Since you were so kind recently in pointing out where I can improve my way of discussing things, I may return the favor: it usually helps in finding a compromise if both parties are open to deviate from their final goal.

mkoeppe:
Tobias, you misunderstand. I've marked the thread as resolved because there were no useful points in it, nor do I expect to learn anything new from it if were to continue. It's all empty calories.

tobiasdiez:
An alternative is suggested in https://github.com/sagemath/sage/pull/35986 (ready for review), where the cython lock is present in the environment files that users/devs will us to generate the conda env, but not in the `conda.txt`. I think this is a nice compromise.

mkoeppe:
I don't see the conda lock as a solution for this problem. It can be nice for creating stability for deployments on the application level, such as our gitpod or devcontainers. 
But it's hardly a replacement for the thoughtful use of version ranges.

dimpase:
I'm trying to run "[Using conda to provide all dependencies for the Sage library (experimental)](https://github.com/mkoeppe/sage/blob/355b39cce59ad9e5cb3e5c8b16915410e83f8591/src/doc/en/installation/conda.rst#using-conda-to-provide-all-dependencies-for-the-sage-library-experimental)" on this branch, and get
```
...
  adding 'sage_conf-9.8.dist-info/top_level.txt'
  adding 'sage_conf-9.8.dist-info/RECORD'
  removing build/bdist.linux-x86_64/wheel
  Building wheel for sage-conf (pyproject.toml) ... done
  Created wheel for sage-conf: filename=sage_conf-9.8-py3-none-any.whl size=5929 sha256=2452cdf180ceebae3ac425dc4be670ed835ba9e14dabadcca092846ec7d3003b
  Stored in directory: /home/scratch/dimpase/.cache/pip/wheels/42/8d/24/ee2834b08081c68f0473dced1cfa0c801641dffdf6648cb811
Successfully built sage-conf
Failed to build sagemath-standard
ERROR: Could not build wheels for sagemath-standard, which is required to install pyproject.toml-based projects
Exception information:
  File "/home/scratch/scratch2/dimpase/mambaforge/envs/sage-dev/lib/python3.11/site-packages/pip/_internal/cli/base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
             ^^^^^^^^^^^^^^^
  File "/home/scratch/scratch2/dimpase/mambaforge/envs/sage-dev/lib/python3.11/site-packages/pip/_internal/cli/req_command.py", line 248, in wrapper
    return func(self, options, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scratch/scratch2/dimpase/mambaforge/envs/sage-dev/lib/python3.11/site-packages/pip/_internal/commands/install.py", line 429, in run
    raise InstallationError(
pip._internal.exceptions.InstallationError: Could not build wheels for sagemath-standard, which is required to install pyproject.toml-based projects
Remote version of pip: 23.2.1
Local version of pip:  23.2.1
Was pip installed by pip? False
```
Any idea what it could be?

I don't know whether it's a regression or not, though.

dimpase:
it appears that one needs to run ./bootstrap, not just ./bootstrap-conda, to set everything up properly.

mkoeppe:
> I don't see where `./bootstrap` is called in [sage](https://github.com/mkoeppe/sage/tree/355b39cce59ad9e5cb3e5c8b16915410e83f8591)/[pkgs](https://github.com/mkoeppe/sage/tree/355b39cce59ad9e5cb3e5c8b16915410e83f8591/pkgs)/[sage-conf_conda/setup.py](https://github.com/mkoeppe/sage/tree/355b39cce59ad9e5cb3e5c8b16915410e83f8591/pkgs/sage-conf_conda/setup.py)

You are right, that was a mistake in #36367. I'll fix it here.

dimpase:
There seems to be one more bug/feature in setup.py: configure is not run if config.status is present. 

And if you have a stale config.status lying around, then there is a good chance you get a misconfiguration. Not sure how to resolve this in general, but here removing it first of all is a workaround..

mkoeppe:
You are right, it should probably at least run `./config.status --recheck && ./config.status` in this case, and maybe check that prefix is configured to be the same as the conda prefix

dimpase:
as a quick semi-fix, you can add the call to ./configure right after ./bootstrap in the instructions

mkoeppe:
No, the very point of sage-conf_conda is to simplify the instructions.

mkoeppe:
Keeping it simple.

dimpase:
What's the meaning of
```
        if os.path.exists(os.path.join(SAGE_ROOT, 'config.status')):
             print(f'Reusing configured SAGE_ROOT={SAGE_ROOT}')
```
if you anyway just run `./confugure` after it ?

mkoeppe:
It's a warning of sorts

tobiasdiez (changes requested):
This PR is introducing an additional test failure in interpreter.py. At least its likely, I don't see it in https://github.com/sagemath/sage/pull/35986 but in every other conda PR. It's hard to verify this, since we cannot run the conda workflow without the auto-fix applied atm. Removing the blocker label because of this.

mkoeppe:
> This PR is introducing an additional test failure in interpreter.py.

Link to the failure please. There is no such failure in the Build & Test of this very PR.

tobiasdiez:
e.g. https://github.com/sagemath/sage/actions/runs/6451909843/job/17624459499?pr=36431. I've opened https://github.com/sagemath/sage/pull/36453 for the cython version fix, so we can see if it's really related to the other changes of this PR here.

mkoeppe:
You mean this?
```
sage -t --random-seed=219809728447624716937891191463986234770 src/sage/repl/interpreter.py
File "src/sage/repl/interpreter.py", line 77, in sage.repl.interpreter
Failed example:
    print("dummy line"); shell.run_cell('1/0') # see #25320 for the reason of the `...` and the dummy line in this test
Expected:
    dummy line
    ...
    ZeroDivisionError...Traceback (most recent call last)
    ...
    ----> 1 Integer(1)/Integer(0)
    .../sage/rings/integer.pyx... in sage.rings.integer.Integer...div...
    ...
    -> ...                  raise ZeroDivisionError("rational division by zero")
       ...            x = <Rational> Rational.__new__(Rational)
       ...            mpq_div_zz(x.value, ....value, (<Integer>right).value)
    <BLANKLINE>
    ZeroDivisionError: rational division by zero
Got:
    dummy line
    ZeroDivisionError                         Traceback (most recent call last)
    Cell In[1], line 1
    ----> 1 Integer(1)/Integer(0)
    <BLANKLINE>
    File sage/rings/integer.pyx:2016, in sage.rings.integer.Integer.__truediv__()
    <BLANKLINE>
    ZeroDivisionError: rational division by zero
*********************************************************************
```

mkoeppe:
The ipython version bound set here on this PR branch is the only thing that could be relevant for this doctest

tobiasdiez:
Yes, that's the one. https://github.com/sagemath/sage/pull/36453 is working.

dimpase:
"re-using" might mean different things, e.g. not ignoring rather than ignoring the existing settings

dimpase:
how about using the approach of #36453, i.e. not pinning a concrete version of Cython, but restricting the range and excluding 3.0.2, here?

mkoeppe:
I've narrowed the scope of the PR, removing the change to the `ipython` (https://github.com/sagemath/sage/pull/36411#issuecomment-1758773922) and `cython` version constraints, and have removed the message.

mkoeppe:
> how about using the approach of #36453

See previous discussion here on this PR on this topic. To be continued on #36453.

dimpase (approved):

mkoeppe:
Thanks

Issue #36322, opened 2023-09-24
Replace tmp_dir() and tmp_filename()


orlitzky:
### Steps To Reproduce

Look at the sage source code.

### Expected Behavior

We use Python's `tempfile` module for temporary file and directory handling.

### Actual Behavior

We use two custom functions, `tmp_filename()` and `tmp_dir()` that more or less reimplement Python's own tempfile handling except with additional downsides.

### Additional Information

Our own `tmp_filename()` and `tmp_dir()` are superfluous:

* They don't do anything that python's own `tempfile` module can't do.
* They do clean up after themselves, but only when sage exits. This wastes resources if you launch multiple sage processes (e.g. https://groups.google.com/g/sage-devel/c/ZctT9fdDXtU). The python functions allow finer-grained control.
* The python functions support context managers that clean up after themselves automatically.
* Python programmers are already familiar with its `tempfile` module, but not the quirks of `tmp_filename()` and `tmp_dir()`.

It's not an onerous list... they just don't need to exist.

### `tmp_dir()`

Most uses of `tmp_dir()` are easy to replace:

1. If the directory is only used in one place, and only temporarily, you can wrap the block of code in a `with tempfile.TemporaryDirectory() as d` block. At the end of the block, the directory is removed.
2. If the directory is used in a long-running doctest example, it's nicer for the reader if you instead create the directory with `d = tempfile.TemporaryDirectory()`, and then, at the end of the example, clean it up with `d.cleanup()`. The explicit cleanup is not strictly necessary, but doing so keeps our cumulative resource usage low.
3. The existing behavior of `tmp_dir()` leaves the directory around until sage exits. If you actually _need_ that behavior, you can replicate it by creating the directory with `d = tempfile.TemporaryDirectory()` and then omitting the call to `d.cleanup()`. Per the documentation, "On ... destruction of the temporary directory object, the newly created temporary directory and all its contents are removed from the filesystem."

### Environment

```markdown
Any
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

orlitzky:
Leaving a comment because the python-3.12 docs confused the hell out of me and I'm sure to forget what I discovered.

The python-3.11 (and earlier) docs say that both temporary files and temporary directories are cleaned up when the object goes out of scope. But now in 3.12, the `NamedTemporaryFile` docs say,

[...]

PR #36436, opened 2023-10-10
src/sage/misc/latex.py: replace tmp_dir()

orlitzky:
These bits of code only use one temporary file, but they call a function `_run_latex_()` that writes its output file to the same directory as its input file. Since we don't want the output path to be predictable (for security reasons), we have to "hide" the one temporary file inside of a temporary directory that will only be writable by the Sage user.

In other words: standard `tempfile.TemporaryDirecrory()` replacement.

Issue: https://github.com/sagemath/sage/issues/36322

### DIFF
--- a/src/sage/misc/latex.py
+++ b/src/sage/misc/latex.py
@@ -30,9 +30,9 @@ import random
 import re
 import shutil
 from subprocess import call, PIPE
+from tempfile import TemporaryDirectory
 
 from sage.misc.cachefunc import cached_function, cached_method
-from sage.misc.temporary_file import tmp_dir
 from sage.structure.sage_object import SageObject
 
 from sage.misc.lazy_import import lazy_import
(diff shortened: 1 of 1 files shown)


mezzarobba (commented):
Is there a reason not to use `with TemporaryDirectory()` (or a `try`-`finally` block) — especially in `view`, which has two possible exit points even in the absence of any exception?

orlitzky:
> Is there a reason not to use `with TemporaryDirectory()` (or a `try`-`finally` block) — especially in `view`, which has two possible exit points even in the absence of any exception?

The context manager is better but it makes the diff awful if you have to indent the entire method. I've made the change in a separate commit if you don't mind looking through it though.

mezzarobba (approved):
thanks!

EmmanuelCharpentier:
This bungles `view()`. Highly damageable to console (or emacs' `sage-shell-mode`) users.

I opened https://github.com/sagemath/sage/issues/36526.

PR #36437, opened 2023-10-10
src/sage/parallel/map_reduce.py: replace tmp_dir()

orlitzky:
Standard `tempfile.TemporaryDirectory()` replacement for `tmp_dir()` and fix a deprecation warning while we're at it.

Issue: https://github.com/sagemath/sage/issues/36322


mezzarobba (approved):
lgtm, thanks!

PR #36440, opened 2023-10-10
trying ruff on some posets files

fchapoton:
This is changing two files in the posets folder.

- use a smaller doctest in one place
- replace one annoying use of `deepcopy` by the method `copy` of `DiGraph`
- apply `ruff check --fix --select UP src/sage/combinat/posets` to make a few automatic modifications, about strings mostly, and about the syntax of `super`
- avoid one symbolic number in one doctest

### :memo: Checklist


dcoudert (approved):
LGTM.

PR #36443, opened 2023-10-10
CI / tox: Update and speed up pyright

mkoeppe:
We bring `tox -e pyright` and the pyright invocation that is part of the CI "Build & Test" back in sync:
- They are now both pinned to today's latest pyright version
- node.js is provisioned via https://pypi.org/project/pyright/ and never accepts a preinstalled node version 
- node.js is configured to provide sufficient heap size. This reduces the runtime dramatically as pyright no longer has to fight with the garbage collector and eliminates spurious out of memory conditions.
- Both run on the entire source tree.

This is done so that even if the output of pyright may be incomprehensible to developers, at least there are no gratuitous differences between what's shown on CI and what is shown locally.

Also done as part of this ticket: A limited effort to get rid of some warnings, just enough to establish that it would be  premature to create problem matchers that annotate these warnings in the source code. Most of the warnings are meaningless and distracting noise because pyright is struggling with too many aspects of the Sage source code: lazy imports, methods in `ParentMethods`, `ElementMethods`, etc. 

### :memo: Checklist

### :hourglass: Dependencies

- #36515 (split out from here)
- #36516 (split out from here)
- #36517 (split out from here)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,20 +103,9 @@ jobs:
           MAKE: make -j2 --output-sync=recurse
           SAGE_NUM_THREADS: 2
 
-      - name: Set up node to install pyright
-        if: always() && steps.worktree.outcome == 'success'
-        uses: actions/setup-node@v3
-        with:
-          node-version: '12'
-
-      - name: Install pyright
-        if: always() && steps.worktree.outcome == 'success'
-        # Fix to v232 due to bug https://github.com/microsoft/pyright/issues/3239
-        run: npm install -g pyright@1.1.232
-
       - name: Static code check with pyright
         if: always() && steps.worktree.outcome == 'success'
-        run: pyright
+        run: ./sage -tox -e pyright
         working-directory: ./worktree-image
 
       - name: Clean (fallback to non-incremental)
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -22,6 +22,7 @@
     "reportUndefinedVariable": "warning",
     "reportOptionalOperand": "warning",
     "reportMissingImports": "warning",
+    "reportPrivateImportUsage": "warning",
     // Suppress because it flags our standard pattern @staticmethod __classcall__(cls, ...)
     "reportSelfClsParameterName": "none",
 }
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -22,5 +22,6 @@
     "reportUndefinedVariable": "warning",
     "reportOptionalOperand": "warning",
     "reportMissingImports": "warning",
+    // Suppress because it flags our standard pattern @staticmethod __classcall__(cls, ...)
     "reportSelfClsParameterName": "none",
 }
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -22,4 +22,5 @@
     "reportUndefinedVariable": "warning",
     "reportOptionalOperand": "warning",
     "reportMissingImports": "warning",
+    "reportSelfClsParameterName": "none",
 }
--- a/src/sage/misc/lazy_attribute.pyi
+++ b/src/sage/misc/lazy_attribute.pyi
@@ -1,3 +1,9 @@
 # This type-stub file helps pyright understand the decorator @lazy_attribute.
 
-lazy_attribute = property
+# Adapted from https://github.com/python/typeshed/blob/b9640005eb586afdbe0a57bac2b88a7a12465069/stdlib/builtins.pyi#L1237-L1254
+class lazy_attribute:
+    def __init__(
+        self,
+        f: Callable[[Any], Any] | None = ...
+    ) -> None: ...
+    def __get__(self, a: Any, cls: type) -> Any: ...
--- a/src/sage/misc/lazy_attribute.pyi
+++ b/src/sage/misc/lazy_attribute.pyi
@@ -1,0 +1,1 @@
+lazy_attribute = property
--- a/src/sage/misc/lazy_attribute.pyi
+++ b/src/sage/misc/lazy_attribute.pyi
@@ -1,1 +1,3 @@
+# This type-stub file helps pyright understand the decorator @lazy_attribute.
+
 lazy_attribute = property
--- a/src/sage/schemes/toric/variety.py
+++ b/src/sage/schemes/toric/variety.py
@@ -2250,20 +2250,21 @@ class ToricVariety_field(AmbientSpace):
             1
         """
         Td = QQ.one()
-        if self.dimension() >= 1:
+        dim = self.dimension()
+        if dim >= 1:
             c1 = self.Chern_class(1)
             Td += QQ.one() / 2 * c1
-        if self.dimension() >= 2:
-            c2 = self.Chern_class(2)
-            Td += QQ.one() / 12 * (c1**2 + c2)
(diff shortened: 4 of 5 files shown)


kwankyu (approved):
Working fine.

mkoeppe:
Thank you!

tobiasdiez (dismissed):
Sadly it doesn't really work. It now runs the check only on a subset of the files (the instructions in the tox file are obsolete).
Moreover, I don't think its a good idea to install a node package via an additional tour over tox + pip. I also would like to keep some version constraint to prevent that this test fails due some new releases. If you want to keep it up to date, add a simple rennovate config https://docs.renovatebot.com/modules/manager/regex/

mkoeppe:
> It now runs the check only on a subset of the files (the instructions in the tox file are obsolete).

Well, that this had fallen out of sync was a bug. Do we really want to run it on all files?

mkoeppe:
> I also would like to keep some version constraint to prevent that this test fails due some new releases.

Please suggest a version constraint

mkoeppe:
> I don't think its a good idea to install a node package via an additional tour over tox + pip.

It's better because we're reusing what is already configured, so it can't fall out of sync as it had.

tobiasdiez:
> > I don't think its a good idea to install a node package via an additional tour over tox + pip.
> 
> It's better because we're reusing what is already configured, so it can't fall out of sync as it had.

Well, the tox version is still not working.

If you want to simplify the setup, you can use https://github.com/marketplace/actions/run-pyright. Also has the advantage that it registers annotation matchers.

mkoeppe:
> the tox version is still not working

Thanks, I see that.

pyrightconfig.json
```diff
@@ -22,4 +22,7 @@
     "reportUndefinedVariable": "warning",
     "reportOptionalOperand": "warning",
     "reportMissingImports": "warning",
+    "reportPrivateImportUsage": "warning",
+    // Suppress because it flags our standard pattern @staticmethod __classcall__(cls, ...)
```
tobiasdiez, line 26: comments in json files are sadly not possible. the common workaround I've seen is to add them as another field, i.e. `reportSelfClsParameterName.comment` (not sure if it works here or if pyright complains)

.github/workflows/build.yml
```diff
@@ -103,20 +103,9 @@ jobs:
           MAKE: make -j2 --output-sync=recurse
           SAGE_NUM_THREADS: 2
 
-      - name: Set up node to install pyright
```
tobiasdiez, line 106: I still think that this way of installing pyright is better. I also don't get the consistency argument (as it comes down to calling pyright without any other arguments). Tox just installs (and calls) pyright in a very convoluted way. And this is exemplified by the memory issues. What to you think about using https://github.com/jakebailey/pyright-action? Also has the advantage of adding problem matchers

pyrightconfig.json
```diff
@@ -22,4 +22,7 @@
     "reportUndefinedVariable": "warning",
     "reportOptionalOperand": "warning",
     "reportMissingImports": "warning",
+    "reportPrivateImportUsage": "warning",
```
tobiasdiez, line 25: Why not fixing the single instance of this error (by properly importing the offending method)?

src/sage/misc/lazy_attribute.pyi
```diff
@@ -0,0 +1,3 @@
+# This type-stub file helps pyright understand the decorator @lazy_attribute.
+
+lazy_attribute = property
```
tobiasdiez: But this is not the correct type, right? lazy_attribute just happens to have the same typing as property (at the moment). In this case, just replace it by https://github.com/python/typeshed/blob/b9640005eb586afdbe0a57bac2b88a7a12465069/stdlib/builtins.pyi#L1237-L1254

tobiasdiez (changes requested):

pyrightconfig.json
mkoeppe: It's a Microsoft extension ("JSONC"), see https://github.com/microsoft/node-jsonc-parser

pyrightconfig.json
mkoeppe: I didn't see the immediate fix, and it did not seem worth it.

.github/workflows/build.yml
mkoeppe: > I [...] don't get the consistency argument (as it comes down to calling pyright without any other arguments).

It comes down to not only that, but also to running the same pyright version, on the same Node version.

> Tox just installs (and calls) pyright in a very convoluted way. And this is exemplified by the memory issues.

There's nothing convoluted about it. This is how one installs and runs pyright using Python infrastructure without having to rely on a system installation of Node. 

And there is nothing mysterious about "memory issues", it is a default heap size of 2 GB, which is too small for our purpose. Here I configure it explicitly so that proper function does not depend on an externally set configuration.

> What to you think about using https://github.com/jakebailey/pyright-action? Also has the advantage of adding problem matchers

I looked at it, as you had mentioned it before. As I explained in the PR description, adding Check Warnings for pyright's complaints just makes no sense.

src/sage/misc/lazy_attribute.pyi
mkoeppe: Thanks, done in 05e34ef

.github/workflows/build.yml
tobiasdiez: > There's nothing convoluted about it. This is how one installs and runs pyright using Python infrastructure without having to rely on a system installation of Node.

Yes, but we have a node environment here, so no need to use the Python infrastructure. I'm aware that its more convenient for the average Python dev to run it that way locally. But that doesn't mean the ci needs to do that way, and in fact the node-setup action is the githuby way of installing node in a github workflow. So please use it.

pyrightconfig.json
tobiasdiez: I don't remember the names, but sage is importing x from y, and y is importing x from z (without reexporting it). So just import x from z.

.github/workflows/build.yml
mkoeppe: > in fact the node-setup action is the githuby way of installing node in a github workflow. 

No, there's not only "one way" of doing things right.

pyrightconfig.json
mkoeppe: We don't even have such a thing as a convention of what reexports should look like in Sage. So warning instead of error is quite appropriate.

tobiasdiez: Here `y` is an external library. So sage was accessing a non-public api. I think this is a good catch and should be fixed properly.

.github/workflows/build.yml
tobiasdiez: You can also easily disable comments for the pyright action if that's your concern. I would say we should run it twice: once without commenting for all functions, and a second time with comments on type-annotated functions.

pyrightconfig.json
mkoeppe: I don't plan to work on this import. If you want to fix stuff, just open a PR.

.github/workflows/build.yml
mkoeppe: Beyond the scope of this PR.

mkoeppe:
Closing as it has been replaced by #36515

PR #36442, opened 2023-10-10
CI: Refactor `build*.yml` through a new script `.ci/retrofit-worktree.sh`

mkoeppe:
Removes some duplicated code, no changes in functionality.
Also preparation for merging 3 workflows into one, as proposed in https://github.com/sagemath/sage/pull/36349#issuecomment-1755882567

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36435, opened 2023-10-09
Retrieve upstream tarballs from GitHub release assets

mkoeppe:
We propose to add GitHub release assets as a source from which upstream tarballs can be retrieved. This would take priority over trying the Sage mirrors, thus reducing the reliance of the project on self-maintained file servers. https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas

For the purpose of testing this PR, one tarball missing on the mirrors (https://github.com/sagemath/sage/pull/36381#issuecomment-1752268975) has been uploaded to https://github.com/sagemath/sage/releases/tag/10.1

Creating GitHub releases and depositing the upstream tarballs can later be automated in a GH Actions workflow, e.g., using https://github.com/softprops/action-gh-release. As we don't make GitHub releases for beta versions, the previous stable release can be used.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36446, opened 2023-10-11
CI: Merge workflows into build.yml

mkoeppe:
Simplifying our workflows running on every ticket, as suggested in https://groups.google.com/g/sage-devel/c/Cisdhf3nSbM

Unchanged as of this PR: build, doc-build, doc-build-pdf jobs all build the image. This can be improved by building a container first.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36443 
- Depends on #36442
- Depends on #36435 
- Depends on #36411
- Depends on #36473


mkoeppe:
@kwankyu Here's an attempt to reorganize the CI workflows. Checks tab: https://github.com/sagemath/sage/pull/36446/checks 

This is not the final implementation; the merge of the separate workflows also enables sharing the build phases of build and docbuild workflows via Docker.

mkoeppe:
Also ci-conda could be merged here, but I'm not touching it for the moment because other PRs are working on it

kwankyu:
We may have separate "Build & Test" and "Build & Test using Conda". They are well distinguished. As far as I understand, Building on conda is still not as stable as on the vanilla system. Hence merging them would make "Build & Test" unstable and too big. "Build & Test" is our core workflow, and better to be kept rock-reliable.

tobiasdiez:
How do you plan to handle the fact that some of the workflows currently have different triggers? In particular, the Linux work flow only being run when there is a label? Similarly, I think it will be hard to trigger the publish doc workflow directly after the docbuild has been finished.

mkoeppe:
Thanks, these are relevant points to consider. 

Current status (mainly for future reference):
- Pushing the image to a registry works great for pushes but, as we know, cannot work for PRs because there aren't (and likely should not be) permissions.
- Saving the Docker image as an artifact is too slow, much slower than pushing the image to a registry (in which case existing layers are reused). Pushing a multi-gigabyte image via upload-artifact takes very long, so not a suitable approach.
- One can of course easily run a local registry (e.g. via the `services` keyword), but it will only be accessible by a job, not by all jobs in a workflow, so nothing gained.
- Saving the changes as an incremental tarfile (https://www.gnu.org/software/tar/manual/html_node/Incremental-Dumps.html), which I was using successfully in the multi-stage Cygwin workflow, unfortunately also does not work well because of the type of file system used within Docker containers.
- Approaches such as https://github.com/aidanhs/dlgrab are possible, but I won't pursue this direction.

Next approach that I'll try: 
- Docker layer caching via https://docs.docker.com/build/cache/backends/gha/ -- after building the image incrementally, the jobs for docbuild, ptest, etc. will just again rebuild the image incrementally from the same source image using the same steps; when the cache can be retrieved, this will be very fast; otherwise, it just falls back to building the image incrementally (= current status).

mkoeppe:
I'm closing this and will do the docker layer caching in a separate PR; which will keep the workflows separate

Issue #30956, opened 2020-11-24
sdh_store_and_pip_install_wheel: Install from the persistent wheel directory


mkoeppe:
<div id="comment:0"></div>

Python packages no longer use SAGE_DESTDIR except for staging/removing the `.whl` file.

This last bit of staging leads to `sage -pip freeze` showing wheel URLs in temporary build locations:

```
$ ./sage -pip freeze
alabaster @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-algebraic-2018-spring/local/var/tmp/sage/build/alabaster-0.7.12/src
appdirs==1.4.4
appnope @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-algebraic-2018-spring/local/var/tmp/sage/build/appnope-0.1.0.p0/src
argon2-cffi==20.1.0
attrs @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-algebraic-2018-spring/local/var/tmp/sage/build/attrs-19.3.0/src
Babel @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-algebraic-2018-spring/local/var/tmp/sage/build/babel-2.6.0/src
```

(These locations are recorded in `direct_url.json` according to https://www.python.org/dev/peps/pep-0610/; see https://packaging.python.org/en/latest/specifications/direct-url/)

In this ticket, we remove the staging, writing the `.whl` file directly into `SAGE_VENV/var/lib/sage/wheels`. (Uninstallation of the `.whl` file will be done by adding to the `spkg-piprm` script.) 

Thus, `sage -pip freeze` will show the actual file URLs of the stored wheels.

CC:  @jhpalmieri @tobiasdiez

Component: **build**

_Issue created by migration from https://trac.sagemath.org/ticket/30956_

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -10,4 +10,5 @@
 Babel @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-algebraic-2018-spring/local/var/tmp/sage/build/babel-2.6.0/src
 ```
 
+(These locations are recorded in `direct_url.json` according to https://www.python.org/dev/peps/pep-0610/; see https://packaging.python.org/en/latest/specifications/direct-url/)
 
``````

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -1,4 +1,6 @@
[...]

PR #36452, opened 2023-10-11
`make sagemath_categories-check`, `make pypi-wheels-check`

mkoeppe:
We add `make` targets for testing wheel-building script packages that have already been installed. For example:
```
$ make SAGE_WHEELS=yes sagemath_categories  # builds and stores (but does not install) a wheel
$ make sagemath_categories-check            # tests the wheel using tox
```
The target `make pypi-wheels-check` calls all of them. We restructure the Build & Test CI so that all tests are run after build has been completed. Later, as a follow up of this and #36446, we can store the result of Build as a container image and then parallelize the various tests on top of it.

This is implemented on top of an improvement of how we record information on our wheels.
```
$ cat venv/var/lib/sage/scripts/sagemath_categories/spkg-requirements.txt  
sagemath_categories @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-algebraic-2018-spring/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/sagemath_categories-10.2b6-cp311-cp311-macosx_13_0_x86_64.whl
```
Previously, this information was only known during the execution of the `spkg-install` script.

In a follow-up, we can address #30956 by delaying the wheel installation to spkg-postinst.

### :memo: Checklist


.github/workflows/build.yml
```diff
@@ -165,6 +164,25 @@ jobs:
           MAKE: make -j2 --output-sync=recurse
           SAGE_NUM_THREADS: 2
 
+      # Testing
+
+      - name: Test changed files (sage -t --new)
+        run: |
+          # We run tests with "sage -t --new"; this only tests the uncommitted changes.
```
kwankyu, line 172: `if: always() && steps.worktree.outcome == 'success'`?

mkoeppe: thanks, i've made a similar change in 8224cb0

kwankyu:
Thus previously 
```
make SAGE_CHECK=yes sagemath_categories                    
```
worked, but will not work after this PR. Instead we will use 
```
make sagemath_categories-check                                           
```
Am I right?

If so, then do we need to update some relevant part of the developer guide?

mkoeppe:
> Thus previously
> 
> ```
> make SAGE_CHECK=yes sagemath_categories                    
> ```
> 
> worked, but will not work after this PR

No, it still works the same way. The makefile invokes the `spkg-check` script after running the `spkg-install`.

build/bin/sage-spkg
```diff
@@ -567,24 +567,6 @@ for script in $WRAPPED_SCRIPTS; do
...
-    elif [ -f setup.py ]; then
-        echo 'python setup.py install' >> spkg-install
-    else
-        echo >&2 "Error: There is no spkg-install script, no setup.py, and no configure"
-        echo >&2 "script, so I do not know how to install $PKG_SRC."
-        exit 1
-    fi
-    chmod +x spkg-install
-fi
-
```
kwankyu, line 587: Is this never used, and so removed?

kwankyu:
OK. It looks good as far as I can tell.

It works well as described.

kwankyu (approved):

build/bin/sage-spkg
mkoeppe: Yes, that's right. It was a leftover from the abandoned support for "old-style packages"

mkoeppe:
Thank you!

PR #36462, opened 2023-10-15
using ruff for UP004, UP008, UP028

fchapoton:
this was scripted using

`ruff check --fix --select UP004,UP008,UP028 src/sage`

here is the link to `ruff` documentation: https://docs.astral.sh/ruff/rules/

### :memo: Checklist


fchapoton:
not sure that the changes to super will work ; I think I kept them as they were breaking something ; let's see

kwankyu (approved):
LGTM.

kwankyu:
Do you plan to add ruff to our linter fleet?

fchapoton:
maybe, but the CI si such a mess for the moment...

mkoeppe:
@fchapoton At least add ruff to the developer documentation please. https://doc.sagemath.org/html/en/developer/tools.html

kwankyu:
> not sure that the changes to super will work ; I think I kept them as they were breaking something ; let's see

I worried that `super()` replacing `super(CLASS, cls)` would work. It seems that `super()` works as expected in a static method.

fchapoton:
bakc to positive after trivial merge

PR #36466, opened 2023-10-16
Run incremental linux ci only when packages changed

tobiasdiez:
... so that it only shows up in the "checks" list below in these cases.

### :memo: Checklist


kwankyu:
@mkoeppe ?

.github/workflows/ci-linux-incremental.yml
```diff
@@ -18,12 +18,11 @@ name: CI Linux incremental
 on:
   pull_request:
     types:
-      # Defaults
-      - opened
-      - synchronize
-      - reopened
```
mkoeppe, line 24: Why this change?

tobiasdiez: Isn't it sufficient to run the workflow when you add the label?

mkoeppe: No, that's not how it's intended.

tobiasdiez: okay, reverted.

kwankyu: I think we should rather remove "labeled" filter. This would trigger the workflow whenever any new label is added such as "needs review". Is this intended? Why?

tobiasdiez:
Putting this to positive review since @kwankyu gave his okay and the comments by @mkoeppe were addressed. (please correct if I'm not quick here)

mkoeppe:
Yes, that's fine

PR #36473, opened 2023-10-17
CI docbuild: Do not rebuild sagelib from scratch

mkoeppe:
Reworking #35652 as discussed in https://github.com/sagemath/sage/pull/36469#issuecomment-1765662418

Also applying the changes made in #35866 (cosmetic improvements to the Actions logs of doc-build.yml) to doc-build-pdf.yml

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36160, opened 2023-08-30
Fix documentation previews using mathjax cdns

kwankyu:
Fixes #36159 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36475, opened 2023-10-17
.github/workflows/doc-build.yml: Repair display of changes

mkoeppe:
From https://github.com/sagemath/sage/pull/36473#discussion_r1362509153

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36473 (merged here)
- Depends on #36442 (merged here)

### DIFF
--- a/.ci/retrofit-worktree.sh
+++ b/.ci/retrofit-worktree.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+if [ $# != 2 ]; then
+    echo >&2 "usage: $0 WORKTREE_NAME WORKTREE_DIRECTORY"
+    echo >&2 "Ensures that the current working directory is a git repository,"
+    echo >&2 "then makes WORKTREE_DIRECTORY a git worktree named WORKTREE_NAME."
+fi
+WORKTREE_NAME="$1"
+WORKTREE_DIRECTORY="$2"
+
+export GIT_AUTHOR_NAME="ci-sage workflow"
+export GIT_AUTHOR_EMAIL="ci-sage@example.com"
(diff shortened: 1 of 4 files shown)


kwankyu:
As you pointed out, this is due to the change of using mathjax cdns from 10.2.beta1 and, as I understand it, should have been transient hiccups if ci didn't fail to build images of 10.2.beta2~5 (is this also because of the change?). Now that the images of 10.2.beta6 and beta7 are available, the problem will disappear. Then do we need this fix?

mkoeppe:
The change (in #36160) was done by setting an environment variable in the doc-build workflow. 

The Docker images however are built with the default configuration, without this environment variable set. (I don't know if/why that should be changed.)

kwankyu:
I see. This is a tricky situation.

How about comparing the built doc with the previous doc which was built in the previous run of the build doc workflow on the push to develop branch and stored somewhere?

kwankyu:
The current approach is easy and quick but prone to be broken again someday (when we upgrade mathjax).

mkoeppe:
> How about comparing the built doc with the previous doc which was built in the previous run of the build doc workflow on the push to develop branch and stored somewhere?

Sure, another possible solution, but I don't think it would be more robust

mkoeppe:
OK, here's a better version that avoids duplicating that specific CDN URL.

I've also made the change discussed in https://github.com/sagemath/sage/pull/36473#issuecomment-1766872361, which gives a cleaner log

kwankyu (approved):
LGTM.

I am preparing a PR building on this.

mkoeppe:
Thank you!

Issue #34411, opened 2022-08-23
sage-download-file: Limit number of mirrors contacted


mkoeppe:
<div id="comment:0"></div>

In particular when `--enable-download-from-upstream-url` is in use.

CC:  @jhpalmieri @dimpase @vbraun @williamstein

Component: **build**

_Issue created by migration from https://trac.sagemath.org/ticket/34411_

PR #36477, opened 2023-10-18
fplll 5.4.5, fpylll 0.6

mkoeppe:
### :memo: Checklist

### :hourglass: Dependencies

- https://github.com/sagemath/sage/pull/36482 (merged here)


build/pkgs/fpylll/install-requires.txt
```diff
@@ -1 +1 @@
-fpylll >=0.5.9, <=0.5.9
+fpylll ==0.6.0
```
tobiasdiez: Conda needs to be changed accordingly. Again, the usual question: is such a hard version constraint really necessary? It seems like there are no changes necessary between 0.5.9 and 0.6.0, so it would also be fine to accept both, or not?

(if the coupling of fpylll and fplll is the reason, i would say its the users task to provide a working duo)

mkoeppe:
- The packages are unconstrained in our conda.txt.
- The packages currently have the strict pins in spkg-configure.m4 / install-requires.txt.
 
It's beyond the scope of the PR to change any of this. It is an upgrade ticket for Cython 3 and Python 3.12.

tobiasdiez:
That the current practice is not working is shown by the fact that the conda workflow is currently failing because of it. But, okay, I've opened https://github.com/sagemath/sage/pull/36482 if fixing it is out of the scope of this PR.

mkoeppe:
Tested ([with some other update PRs](https://github.com/mkoeppe/sage/commits/ci-10.2.beta8%2Bupdates-1)) in https://github.com/mkoeppe/sage/actions/runs/6622955701

tornaria:
FWIW in void linux we are on fplll-5.4.5 and fpylll-0.6.0 (both unpatched). No issue at all with sagemath so it seems an easy upgrade.

mkoeppe:
Let's get it in please.

dimpase:
seems that we really need a way to limit (or perhaps explicitly specify) the range of Sage mirrors - errors are due to a .cn mirror just hanging.

Once this is in, we can use it throughout CI to prevent this sort of trouble.

mkoeppe:
Yes, that's #34411

dimpase (approved):
lgtm

PR #36482, opened 2023-10-19
Loosen version requirement on fpylll and align its conda version

tobiasdiez:
Currently, the conda workflow fails (e.g https://github.com/sagemath/sage/actions/runs/6556683128/job/17807012736) because it installs fpylll 0.6 and we still require 0.5.9.
To fix this, the version constraint is loosened, and to prevent it from happening in the future the same version constraint is introduced for conda.

### :memo: Checklist


malb:
FPyLLL only works with Cython 3.0, though, would that be a problem?

mkoeppe:
> FPyLLL only works with Cython 3.0, though, would that be a problem?

No, Sage has fully switched to Cython 3.0 too.

mkoeppe:
From the PR description:
> to prevent it from happening in the future the same version constraint is introduced for conda.

This explanation makes no sense, but yes, it's fine to add a lower version bound to conda.txt.

mkoeppe (approved):

tobiasdiez:
Thanks!

PR #36487, opened 2023-10-19
Add a lift method to LieGenerator

ehaka:
The `FreeLieAlgebraElement.lift` method can try to call both `LieGenerator.lift` and `LieBracket.lift`, but the former is not defined. As a consequence coercion into the universal enveloping algebra fails for elements which are not brackets:

```
sage: L = LieAlgebra(QQ, ['X','Y']).Lyndon()
sage: X,Y = L.gens()
sage: U = L.universal_enveloping_algebra()
sage: U.one() + X.bracket(Y) # works fine
1 + X*Y - Y*X
sage: U.one() + X
...
AttributeError: 'sage.algebras.lie_algebras.lie_algebra_element.Lie' object has no attribute 'lift'
```

Copying the pre-existing code in `LieBracket.lift` to `LieGenerator.lift` with straight-forward modifications (since the generator case is even simpler) fixes this problem.

```
sage: U.one() + X
1 + X
```

### DIFF
--- a/src/sage/algebras/lie_algebras/lie_algebra_element.pxd
+++ b/src/sage/algebras/lie_algebras/lie_algebra_element.pxd
@@ -50,6 +50,7 @@ cdef class LieObject(SageObject):
 
 cdef class LieGenerator(LieObject):
     cdef public str _name
+    cpdef lift(self, dict UEA_gens_dict)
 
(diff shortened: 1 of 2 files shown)


tscrim (approved):
Good catch. Thank you for fixing this!

PR #36488, opened 2023-10-19
Fix func_persist: do not use the (now removed) inspect.formatargspec, but instead use inspect.signature.

### LINKED ISSUE #36302: `func_persist` fails with Python 3.11
```sage
@func_persist
def test(n):
    return n*n
```

fails with

```python
AttributeError                            Traceback (most recent call last)
Cell In[1], line 1
----> 1 @func_persist
      2 def test(n):
      3     return n*n

File /usr/lib/python3.11/site-packages/sage/misc/func_persist.py:60, in func_persist.__init__(self, f, dir)
     56 self.__dir = dir
     57 os.makedirs(dir, exist_ok=True)
     58 self.__doc__ = '%s%s%s' % (
     59     f.__name__,
---> 60     inspect.formatargspec(*inspect.getargs(f.__code__)),
     61     f.__doc__)

AttributeError: module 'inspect' has no attribute 'formatargspec'
```

This is because `inspect.formatargspec` has been [deprecated in Python 3.5 and removed in 3.11](https://docs.python.org/3/whatsnew/3.11.html#removed). Probably related to #33842.

jhpalmieri:
Fix func_persist: do not use inspect.formatargspec (gone as of Python 3.11)

The method `inspect.formatargspec` has been deprecated since Python 3.5, now removed as of 3.11. `func_persist` still uses it, so we replace it with `inspect.signature`. 

Another option would be to replace it with `sageinspect.sage_formatargspec`. I think that func_persist should only be used with Python functions, not Cython, and `inspect.signature` is nicer for Python functions. If the function is defined as `def test(n: int) -> int:`, For example, then `inspect.signature` will return `(n: int) -> int` while `sageinspect.sage_formatargspec` will just return `(n)`.

This should fix #36302.

### :memo: Checklist

### DIFF
--- a/src/sage/misc/func_persist.py
+++ b/src/sage/misc/func_persist.py
@@ -45,7 +45,6 @@ import os
 
 from . import persist
 
-
 class func_persist:
     r"""
     Put ``@func_persist`` right before your function
@@ -57,7 +56,7 @@ class func_persist:
         os.makedirs(dir, exist_ok=True)
         self.__doc__ = '%s%s%s' % (
             f.__name__,
-            inspect.formatargspec(*inspect.getargs(f.__code__)),
+            inspect.signature(f),
             f.__doc__)
 
     def __call__(self, *args, **kwds):


balejk:
It seems to work well when imported from Sage 10.1 running on Python 3.12. Thank you!

jhpalmieri:
If you are willing, you may approve the changes and/or give this a positive review using "Labels".

balejk (approved):

balejk:
I have approved the changes, but I don't seem to have enough permissions to change labels.

Issue #34630, opened 2022-10-02
Switch modularized sagelib's build systems from setuptools to meson-python / ninja


mkoeppe:
<div id="comment:0"></div>

(from [#32874 comment:27](https://github.com/sagemath/sage/issues/32874#comment:27))

SciPy has switched to meson-python / ninja (see #34081).
See for example https://github.com/scipy/scipy/blob/main/scipy/linalg/meson.build; this replaces older infrastructure like https://github.com/scipy/scipy/blob/main/tools/cythonize.py

Cython recently added the `--depfile` switch 
- https://github.com/cython/cython/pull/4949 (`cython --depfile`)

This finally enables build systems to do their own dependency tracking (caching dependencies in depfiles) instead of passing all source files to Cython - where it takes a good chunk of the time for a trivial `./sage -b` ([#32874 comment:27](https://github.com/sagemath/sage/issues/32874#comment:27)).

What is needed is to make use of `cython --depfile` in meson:
- https://github.com/mesonbuild/meson/issues/9049 (Cython dep handling)

Then we can switch from setuptools to meson-python. We would continue to use our own package/module discovery code (`sage_setup.find`).

In this ticket, we change the build systems of `pkgs/sagemath-*` (including `pkgs/sagemath-standard`) to meson-python. Editable builds will continue to use setuptools for a monolithic build via `src/setup.py`. (meson-python as of 0.11.x does not support PEP 660 editable wheels - https://github.com/mesonbuild/meson-python/issues/47)

Depends on #33577
Depends on #34081
Depends on #34855
Depends on #34897

CC:  @dimpase @kwankyu

Component: **build**

_Issue created by migration from https://trac.sagemath.org/ticket/34630_

mkoeppe:
Dependencies: **#33577, #34081**

dimpase:
<div id="comment:2" align="right">comment:2</div>

good idea.

mkoeppe:
Changed dependencies from **#33577, #34081** to **#33577, #34081, #33577**

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -13,3 +13,6 @@
 
 Then we can switch from setuptools to meson-python. We would continue to use our own package/module discovery code (`sage_setup.find`).
 
[...]

Issue #33577, opened 2022-03-27
Migrate from setup.cfg to pyproject.toml


tobiasdiez:
<div id="comment:0"></div>

Starting with v61, setuptools (experimentally) support pyproject.toml for configuration, see https://github.com/pypa/setuptools/issues/1688. 

setuptools 68.1.0 declared support for pyproject.toml as stable.

Moreover, setup.cfg will eventually be deprecated so it's a good idea to migrate from cfg to toml.

Migration can be done mostly automatically using https://ini2toml.readthedocs.io/en/latest/readme.html:

```
  (cd pkgs/sage-conf && ini2toml setup.cfg > pyproject.toml && git rm -f setup.cfg && git add pyproject.toml)

  (cd pkgs/sagemath-objects && (echo && echo '[project]' && ini2toml -p setup.cfg setup.cfg.m4 | sed "1,/[[]project/d;s/\"esys/esys/;s/install-requires /install-requires-toml /;s/ *\\\\\\\\\"/\')/;/conf-unix/d;" ) >> pyproject.toml.m4 && git rm -f setup.cfg.m4)
```
(the conversion of the m4 files loses some install-requires and needs to be fixed up manually.)

This is also preparation for #34630 (migration from setuptools to meson-python)

Depends on #33842
Depends on #30666

CC:  @mkoeppe

Component: **build**

Author: **Matthias Koeppe**

Branch/Commit: **[u/mkoeppe/migrate_from_setup_cfg_to_pyproject_toml](https://github.com/sagemath/sagetrac-mirror/tree/u/mkoeppe/migrate_from_setup_cfg_to_pyproject_toml) @ [`f81a0c8`](https://github.com/sagemath/sagetrac-mirror/commit/f81a0c8e12d01cbc3a3c15e99a614fde3b8021cf)**

_Issue created by migration from https://trac.sagemath.org/ticket/33577_

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -1,3 +1,6 @@
 Starting with v61, setuptools (experimentally) support pyproject.toml for configuration, see https://github.com/pypa/setuptools/issues/1688. Moreover, setup.cfg will eventually be deprecated so it's a good idea to migrate from cfg to toml.
 
 Migration can probably be done using https://ini2toml.readthedocs.io/en/latest/readme.html.
+
+This is also preparation for #34630 (migration from setuptools to meson-python)
+
``````

mkoeppe:
[...]

Issue #30025, opened 2020-06-29
Clean up SAGE_ROOT directory by moving stuff into SAGE_ROOT/build/


mkoeppe:
<div id="comment:0"></div>

- Move `configure.ac`, `m4`, `config`, `Makefile` into `build/`.  
- Leave behind symlinks `configure -> build/configure`, `Makefile -> build/Makefile` (or `-> build/make/Makefile-main`)
- Split out `build/bootstrap` from `SAGE_ROOT/bootstrap`, call it from there.

CC:  @dimpase @orlitzky @jhpalmieri

Component: **build**

_Issue created by migration from https://trac.sagemath.org/ticket/30025_

mkoeppe:
superseded by #31662

PR #36489, opened 2023-10-20
Use meson instead of configure for conda install

tobiasdiez:
Switches from `configure` to a simple meson-based setup (when using conda) by implementing a `sage-conf_meson` package that calls meson to looks for the dependencies. I've kept the `sage-conf_conda` for now, as comparison and fall-back.

This is preparation for https://github.com/sagemath/sage/issues/34630. 

### :memo: Checklist

### :hourglass: Dependencies

- #36468 to fix the conda workflow


src/doc/en/installation/conda.rst
```diff
@@ -136,11 +136,15 @@ Here we assume that you are using a git checkout.
...
 
-  - Install the build prerequisites and the Sage library::
+  - Bootstrap the source tree and install the build prerequisites and the Sage library::
 
-      $ pip install --no-build-isolation -v -v --editable ./pkgs/sage-conf_conda ./pkgs/sage-setup
+      $ ./bootstrap
+      $ pip install --no-build-isolation -v -v --editable ./pkgs/sage-conf_meson ./pkgs/sage-setup
       $ pip install --no-build-isolation --config-settings editable_mode=compat -v -v --editable ./src
 
+    In case of errors, try to use `sage-conf_conda` instead of `sage-conf_meson`,
```
mkoeppe:
```suggestion
    In case of errors, try to use ``sage-conf_conda`` instead of ``sage-conf_meson``,
```

meson.build
```diff
@@ -0,0 +1,40 @@
+project('sage', 'c')
```
mkoeppe, line 1: I'd be more comfortable with this file being local to sage-conf_meson, so as not to create expectations that Sage-the-distribution will switch to meson

bootstrap-conda
```diff
@@ -95,6 +95,7 @@ echo >&2 $0:$LINENO: generate conda environment files
         for pkg in $BOOTSTRAP_SYSTEM_PACKAGES; do
             echo "  - $pkg"
         done
+        echo "  - meson"
```
mkoeppe: shouldn't this only go into src/environment-template?

pkgs/sage-conf_meson/sage_conf.py
```diff
@@ -0,0 +1,6 @@
+from _sage_conf.__main__ import _main
+
+from builddir._conf import *
+from builddir.build_info import CONDA_PREFIX, SAGE_ROOT
```
mkoeppe: This "from builddir... import" makes me a bit uncomfortable. 
This seems to assume that `.` is in sys.path
Please test that also non-editable installs (via wheels) of sage-conf_meson work.

mkoeppe:
This goes in a great direction, thanks for working on it. 

For the conda CI, I think it would be crucial to keep a run of the configure-based sage-conf_conda around, just for the tests that it runs (instead of just trusting that the package versions installed in the environment are OK). (For example, by building but not installing the sage-conf_conda wheel.)

src/doc/en/installation/conda.rst
tobiasdiez: Thanks!

meson.build
tobiasdiez: The idea in the middle to long term is to support running `meson compile` from the root directory to compile sagelib with meson (using conda). For this the meson build file needs to be there (and its pretty much convention to have such a build file in the root). Sage-the-distro supporting meson would mean that there is a meson file in the build directory, or not?

bootstrap-conda
tobiasdiez: I guess you are right since in this mode we are still using make. 

Can we not delete the env files completely (along with the documentation)? Or at least move the conda env files that are currently in the root into the `build` dir, and then move the ones from `src` to the root?

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: The `builddir` here is generated under `pkgs/sage-conf_meson`

tobiasdiez:
Thanks for the review. Should have addressed all points now.

meson.build
mkoeppe: I'm not aware of such a plan, and I'd suggest that if you have a plan to use a new Issue in which you explain it.

From what you wrote above, it seems unlikely that it would be the same meson.build file that would be suitable for both purposes, so nothing is gained by putting it there.

bootstrap-conda
mkoeppe: > Can we not delete the env files completely (along with the documentation)? 

We could, but it would not be an improvement.

> Or at least move the conda env files that are currently in the root into the `build` dir, and then move the ones from `src` to the root?

`src` is the root directory of the monolithic Sage library. There's a pyproject.toml there, and right next to it is the environment*.yml. That's exactly the right place for it.

pkgs/sage-conf_meson/sage_conf.py
mkoeppe: Yes, I knew that when I wrote the above. So?

pkgs/sage-conf/pyproject.toml
```diff
@@ -1,3 +1,10 @@
 [build-system]
 requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "sage-conf"
+dynamic = ["version"]
+
+[tool.setuptools.dynamic]
+version = {file = ["VERSION.txt"]}
```
mkoeppe, line 10: It looks like you are trying to move things from setup.cfg to here, but now you'd need to remove it there to complete the job.

tobiasdiez: Nope, just wanted to have the new conf package without setup.cfg. Removing the existing ones can be done in a follow-up.

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: I don't see the difference between `configure` creating the `_conf.py` file and here we are using `meson` to create the `builddir` folder. Can you please be more specific what problem you encounter?

meson.build
tobiasdiez: It's part of https://github.com/sagemath/sage/issues/34630. In the long-term you want to use of course meson-python (configured via pyproject.toml). But in the short to midterm its easier to just call meson directly. That's at least how scipy/numpy proceeded: https://github.com/scipy/scipy/pull/14847/files#diff-e42998d51257e6f84aa51ffa5f4ed1544cb12f97a94978c44f3bc281b5324d79R390-R500

And yes, the meson file added in this PR can be used for this purpose. In fact, I believe we can get ride of the conf and setup packages, but of course there still needs to be some way to provide the runtime-config (probably by just putting the generated conf file in the correct place in the builddir).

pkgs/sage-conf_meson/sage_conf.py
mkoeppe: > I don't see the difference

In https://github.com/sagemath/sage/blob/develop/pkgs/sage-conf/sage_conf.py#L1 we have:
`from _sage_conf._conf import *`

`_sage_conf` is an installed top-level package.

Here: `from builddir._conf import *`

`builddir` is not an installed top-level package.

More questions?

meson.build
mkoeppe: Well, #34630 has nothing to with stuff that belongs in SAGE_ROOT.

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: I don't know what an "installed top-level package" is, but regardless, the editable install works (or do you have a problem with that?!) so that should be good for now. Can be improved in a follow-up.

meson.build
tobiasdiez: Yes, devs usually start in the root.

pkgs/sage-conf_meson/sage_conf.py
mkoeppe: https://docs.python.org/3/reference/import.html

pkgs/sage-conf/pyproject.toml
mkoeppe: It's a regression in sage-conf to duplicate metadata.

mkoeppe:
Tobias, there are clear work items here. They don't go away by marking them as "resolved".

meson.build
tobiasdiez: I have opened https://github.com/sagemath/sage/pull/36524. This should give you an idea on why its handy to have a top-level meson file. (It's very similar to having a top-level make file)

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: What documented behavior is not working for you? As this is a temporary package anyway, I will not invest more time in it than necessary.

pkgs/sage-conf/pyproject.toml
tobiasdiez: I don't understand enough about how setup cfg and pyproject toml interact to solve this. But it appears that the duplication of information results in no problem, so we (you?) can work on improving the setup.cfg file later on. It's definitely not in the scope of this PR to do this.

pkgs/sage-conf_meson/sage_conf.py
mkoeppe: Calling something "temporary" does not make it exempt from normal scrutiny in a review.
In particular, when no plan has been made known anywhere.

tobiasdiez: Can you then please at least say what is not working for you, and why you need this?

mkoeppe: Tobias, I said it clearly in the first message. 

The lines `from builddir._conf import *` are faulty because they depend on "." being in PYTHONPATH etc. This is obviously never correct in library code.

The command `sage-config` must work no matter the current working directory.

tobiasdiez: There is no sage-config script in the meson version. Please test it yourself and report back if you encounter any issues.

mkoeppe: Providing a sage-config script is part of the contract of sage-conf.
So that's another work item.

tobiasdiez: The meson variant is a minimal version of sage-conf that is good enough for an editable install. That's the documented behavior. Please don't extend the scope of this PR by your wishes. If you have a problem that it is only implements a subset of the sage-confirm package, I can rename it. Suggestions welcome

mkoeppe: Something of this quality is obviously not mergeable. So please refrain from setting it back to "needs review".

tobiasdiez: Always a pleasure to work with you. Not sure how that's possible but I'm still surprised by your flexibility to accept solutions that are slightly different from how you envisioned them.

So you really think without the sage-conf script the Sage world will end? There is no way that this script could be added later when other devs are missing it?

mkoeppe: https://github.com/sagemath/sage/blob/develop/CODE_OF_CONDUCT.md

dimpase: if this is working it can be merged.

more features may be handled by a follow-up

pkgs/sage-conf/pyproject.toml
tobiasdiez: I've now simply removed it from the setup.cfg file. Not sure if its working or something else is needed. Please test it and let me know.

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: Thanks @dimpase!

mkoeppe: Note that @dimpase said "if this is working".

I've already made you aware that it does not work. It is not ready to even be tested yet because the defect is obvious from the source code.

tobiasdiez: Matthias, what's your problem? Are you triggered because this PR replaces the configure scripts that you put so much time into? Do you want to pay back because I criticized some of your PRs? You have a lot of things going on in your 'real' life and that stresses you? Whatever it is, please take a deep breath before you further comment here. I'm working on sage in my free time, and if I have to deal with too much negativity I'll just stop. 

Let me tell you a story: A few years ago we had a new contributor at JabRef. He was opening a lot of PRs with huge refactorings. Generally, it was good work but had a few rough edges. At that time, I knew almost any line of the project by heart and thus felt justified to propose alternatives / improvements to these PRs. But this new contributor had also quite strong opinions about how to do things correctly, so we had quite a few long discussions about many things - things which in retrospect were more bike-shedding than actual important issues. At some point, in the middle of one of these PR discussions, he just disappeared. Only at that point, I realized how much good work was in these PRs and that merging 'slightly suboptimal' PRs is still better than having no PRs at all (seems to be a human thing to only realize what you have until you don't have it anymore). It was a good learning experience for me (and I still sometimes have to remind me about it), so maybe this story also has some takeaway points for you.

So that being said, `pip install --no-build-isolation -v -v --editable ./pkgs/sage-conf_meson` produces
```
...
Preparing editable metadata (pyproject.toml) ... done
....
Building wheels for collected packages: sage-conf
....
Building editable for sage-conf (pyproject.toml) ... done
  Created wheel for sage-conf: filename=sage_conf-10.2b7-0.editable-py3-none-any.whl size=2512 sha256=4958feb00d99c295c3e0aa500f37d6197d982b7bbefb7d81c2fe318e5e3f65da
  Stored in directory: /tmp/pip-ephem-wheel-cache-wkm2csaz/wheels/5c/ba/39/add681a35d9ceb3083cbf17c0ed691dcb3688722ca75fb2441
Successfully built sage-conf
```
So could you please test it and report what is not working for you, in the ideal case with suggestions on how to improve things. Thanks!

mkoeppe: Tobias, this is highly inappropriate. I've taken this to sage-abuse.

williamstein: Tobias, I'm writing on behalf of the sage-abuse committee.   We request that you remove your [comment](https://github.com/sagemath/sage/pull/36489#discussion_r1372452493) from this thread and replace it by one that is consistent with our code of conduct [1], or request that an admin delete it your comment.    Your comment violates points of the code of conduct.  If you feel that anybody else's behavior violates the code of conduct, we encourage you to email [2] and explain why.   We also encourage you to write whatever rant you want to [3].     -- Thank you.

[1] https://github.com/sagemath/sage/blob/develop/CODE_OF_CONDUCT.md
[2] https://groups.google.com/g/sage-abuse
[3] https://groups.google.com/g/sage-flame

tobiasdiez: @williamstein I honestly don't see where I've violated the code of conduct. I merely (wanted to) express that I don't enjoy Matthias negativity (and were genuinely asking where this negativity is coming from). Anyway, I've now hidden my comment.

@mkoeppe Would you please review the PR in line with the code of conduct? In particular, finally explain to me what errors you encounter. I get: 

`pip install --no-build-isolation -v -v --editable ./pkgs/sage-conf_meson` produces
```
...
Preparing editable metadata (pyproject.toml) ... done
....
Building wheels for collected packages: sage-conf
....
Building editable for sage-conf (pyproject.toml) ... done
  Created wheel for sage-conf: filename=sage_conf-10.2b7-0.editable-py3-none-any.whl size=2512 sha256=4958feb00d99c295c3e0aa500f37d6197d982b7bbefb7d81c2fe318e5e3f65da
  Stored in directory: /tmp/pip-ephem-wheel-cache-wkm2csaz/wheels/5c/ba/39/add681a35d9ceb3083cbf17c0ed691dcb3688722ca75fb2441
Successfully built sage-conf
```
So could you please test it and report what is not working for you, in the ideal case with suggestions on how to improve things. Thanks!

williamstein: I've gone ahead and just deleted the comment (as an admin).  The code says basically be "friendly, patient, welcoming, considerate, respectful and polite".  Let's all do that and move on. 

Look I'm incredibly appreciative and amazed that all of you are contributing to Sage, almost exactly 20 years after I started the project.  Please, please keep doing so, and in a friendly manner.  Thanks!!

dimpase: > Note that @dimpase said "if this is working".
> 
> I've already made you aware that it does not work. It is not ready to even be tested yet because the defect is obvious from the source code.
> 

Please provide a step by step reproducer for the error you see.

mkoeppe: There's only one step to reproduce: `import sage_conf`.

meson.build
mkoeppe: Yes, looking at #36524 confirms: This is work on configuring the Sage library, not the Sage distribution.

Yes, obviously a top-level `meson.build` file is like a top-level `Makefile`. 

Our top-level Makefile is an element of the Sage distribution, not the Sage library.

So as I said, this new file does not belong in SAGE_ROOT.

pkgs/sage-conf/pyproject.toml
mkoeppe: Thanks.

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: Right, this is indeed not working well. But with the build passing without issues, this means we don't need the sage conf package at all with the conda install?! So we just don't install it, and everything is fine?

mkoeppe: sage-conf is a declared install-requires of sagelib, so some version of it needs to be installed. 
This is all explained in a fairly detailed way in https://github.com/sagemath/sage/blob/develop/pkgs/sage-conf/README.rst, do let me know if anything is missing there and we can improve this documentation.

At this time, I don't have a comment on what variables provided by `sage_conf` are needed at the runtime of the Sage library. We have multiple fallbacks for configuration -- the defaults set in `sage.env`, runtime detection via `sage.features`, and possibly more in various parts of the Sage library.

pkgs/sage-conf/pyproject.toml
mkoeppe: So that's the first little step of #33577.

pkgs/sage-conf_meson/sage_conf.py
tobiasdiez: > At this time, I don't have a comment on what variables provided by sage_conf are needed at the runtime of the Sage library.

That's my observation: apparently none. The fallbacks are good enough for conda.

>  sage-conf is a declared install-requires of sagelib, so some version of it needs to be installed.

So can I change that to an optional dependency? (since it seems to be optional)

mkoeppe: > > At this time, I don't have a comment on what variables provided by sage_conf are needed at the runtime of the Sage library.
> 
> That's my observation: apparently none. The fallbacks are good enough for conda.

You may be right about this, but a lot currently still depends on the assumption "$SAGE_LOCAL = $CONDA_PREFIX" (e.g., see #30914 for pointers), which is guaranteed by the all-conda installation instructions, but I am not sure that we will want to keep it in the future.

A problem with this assumption is that we currently have an inflexible either-or situation: 
- We either take all packages from conda and do not use the Sage build system at all, for anything; 
- or we use the Sage build system for everything and then cannot use Python packages from conda. (For the purpose of this discussion, let's ignore the `--enable-system-site-packages` switch.)

mkoeppe: > > sage-conf is a declared install-requires of sagelib, so some version of it needs to be installed.
> 
> So can I change that to an optional dependency? (since it seems to be optional)

No, by design it's required, but as is explained in https://github.com/sagemath/sage/blob/develop/pkgs/sage-conf/README.rst, you can pick your own implementation.

tobiasdiez: > > > At this time, I don't have a comment on what variables provided by sage_conf are needed at the runtime of the Sage library.
> > 
> > 
> > That's my observation: apparently none. The fallbacks are good enough for conda.
> 
> You may be right about this, but a lot currently still depends on the assumption "$SAGE_LOCAL = $CONDA_PREFIX" (e.g., see #30914 for pointers), which is guaranteed by the all-conda installation instructions, but I am not sure that we will want to keep it in the future.
> 
> A problem with this assumption is that we currently have an inflexible either-or situation:
> 
>     * We either take all packages from conda and do not use the Sage build system at all, for anything;
> 
>     * or we use the Sage build system for everything and then cannot use Python packages from conda. (For the purpose of this discussion, let's ignore the `--enable-system-site-packages` switch.)

Right, but in the "we take everything from conda" the assumption will always be valid, right? So in this case we don't need a sage_conf package. If in the future some semi-mixed versions are supported, then this method needs its own sage_conf package.

> > So can I change that to an optional dependency? (since it seems to be optional)
> 
> No, by design it's required, but as is explained in https://github.com/sagemath/sage/blob/develop/pkgs/sage-conf/README.rst, you can pick your own implementation.

This document doesn't say anything on why we need the package nor why the design cannot be changed.

tobiasdiez: Either way, I've fixed the package now. Please give it a try.

meson.build
tobiasdiez: Sorry, but there are a lot of files in the root folder that are not related to the distribution. pyright config, tox ini, bootstrap conda, pipfile, contributing, gitpod, gitignore.

I think we also agree that sage's main purpose is sagelib, and not the distribution. So reserving the root folder for distribution seems to be conterproductive and giving unncessary prio to it.

It is natural to run `meson compile` from the root of the project. And since this workflow is easy to support, why shouldn't we?!

pkgs/sage-conf_meson/sage_conf.py
mkoeppe: > Right, but in the "we take everything from conda" the assumption will always be valid, right? 

My point is that this dichotomy of the two installation methods is not sustainable; so it would be unwise to be invested in a "pure" mode that seeks to eliminate sage_conf altogether.

meson.build
mkoeppe: > I think we also agree that sage's main purpose is sagelib, and not the distribution. So reserving the root folder for distribution seems to be counterproductive and giving unnecessary prio to it.

I don't disagree with this. And the work on `sage-conf_pypi` and `sage-conf_conda` certainly goes in the direction of "flipping the repository/distribution inside out", i.e., to make the configure script + the build system for SAGE_LOCAL just a detail of *one* implementation of sage-conf. See also #31662 and #30025.

But we cannot make such as structural change just casually on a ticket like this one:
- There's simply no technical need for the meson.build file of this PR here to be in the root.
- (Assuming that by "since this workflow is easy to support" you mean meson + conda for dependencies; correct me if you mean something else) There's no agreement in the community whatsoever that the project should do that.

tobiasdiez: >But we cannot make such as structural change just casually on a ticket like this one:

There is no structural change in this PR. The meson build file provides an alternative to using autoconf, and is placed in the same folder as `configure.ac`. If you move the configure file to `pkgs/sage_conf` at some point, then the meson build file can be moved as well to `pkgs/sage_conf_meson`. Until then, for consistency, both (or none) should be in the root.

mkoeppe: > There is no structural change in this PR. The meson build file provides an alternative to using autoconf

No, it doesn't.

tobiasdiez: I've now moved the config writing to a meson file in the pkgs folder. The meson file in the root is thus there only for convenience of developers (similar to how the tox ini in the root is directing a lot of stuff to the tox ini in src)

mkoeppe: How is it convenient and for what?

tobiasdiez: To test the compilation/config gen. Meson supports eg multiple build directories, so you can play around with changing things while keeping the other generated version(s) around. Also calling meson directly is quicker than going through the pip install.

mkoeppe: So it's for compiling the Sage library, for the single use case of the all-conda install?

dimpase: Meson is somewhere between make and autoconf IMHO.

I don't see a point in insisting that its configs must be somewhere one has to find out from docs.

An alternative would be to have `meson` target in the top-level makefile.

mkoeppe: > Meson is somewhere between make and autoconf IMHO.

Yes, I think everyone here on this PR knows exactly what meson is.

.vscode/settings.json
```diff
@@ -34,5 +34,6 @@
         "sagemath",
         "Cython"
     ],
-    "editor.formatOnType": true
+    "editor.formatOnType": true,
+    "esbonio.sphinx.confDir": ""
```
mkoeppe, line 39: What does this change do? Does it have a relation to the goal of this PR?

.github/workflows/ci-conda.yml
```diff
@@ -95,6 +94,12 @@ jobs:
         shell: bash -l {0}
         run: ./sage -t --all -p0
 
+      # We keep this step for now to make sure that the configure-based setup still works.
+      - name: Test configure
+        shell: bash -l {0}
+        run: |
+          ./configure --enable-build-as-root --with-python=$CONDA_PREFIX/bin/python --prefix=$CONDA_PREFIX --enable-system-site-packages $(for pkg in $(./sage -package list :standard: --has-file spkg-configure.m4 --has-file distros/conda.txt --exclude rpy2); do echo --with-system-$pkg=force; done) || true
```
mkoeppe, line 101: What is this, why re-introduce this long command line that is invoked by sage-conf_conda?

pkgs/sage-conf/pyproject.toml
mkoeppe: I've cherry-picked this part into #36561 and completed the job there, please take a look.

.vscode/settings.json
tobiasdiez: Vscode is adding this every time I'm opening the project. I was bound to commit it at some point ;-). I doesn't hurt...

.github/workflows/ci-conda.yml
tobiasdiez: The conda-ci never used the sage conf package (and its not the point of this PR to change that).

meson.build
tobiasdiez: > So it's for compiling the Sage library, for the single use case of the all-conda install?

There is nothing conda-specific about the meson files, it's only that conda provides a nice fixed environment and thus is a good starting point. You can try to run `meson setup` outside of a conda environment and it should work (if you have all the necessary packages installed in the system).

tobiasdiez:
Can we please get this in. It's blocking further work on meson. As we will gain more experience with meson, the setup will naturally evolve.

dimpase (changes requested):
why do you hardcode version, 10.2.beta8 ? This seems wrong to me.

tobiasdiez:
You mean in the `VERSION.txt` file? These file are automatically updated by the version-update script (there are similar files in the the other sage-conf packages).

dimpase:
yes. can it be left out of the git tree?

tobiasdiez:
I don't know, maybe it would be enough to have one top-level file with the version info (or use git to determine the version). But for now its the equivalent of this file here: https://github.com/sagemath/sage/blob/develop/pkgs/sage-conf_conda/VERSION.txt

meson.build
mkoeppe: I'll explain and summarize once more: 
- What you just described ("it should work (if you have all the necessary packages installed in the system).") means that you are configuring / building the Sage library. 
- This is different from what our `configure.ac` and `Makefile` do. They configure / build the Sage distribution.
- There is no plan to switch the Sage distribution to using meson. 
- When people see a file `meson.build` next to `configure.ac` and `Makefile`, they are bound to think that using meson is an alternative for building the same thing. So this has the potential to mislead.

mkoeppe:
@dimpase Tobias is correct regarding VERSION.txt. It is our standard practice, as `ls -l pkgs/*/VERSION.txt` shows.
Yes, these files could as well be made into symlinks, but if it were decided to make this change, it should be changed uniformly for all distributions packages, and so it's clearly outside of the scope of this PR.

mkoeppe:
> Can we please get this in. It's blocking further work on meson.

Nothing is blocked. 

A problem here on this PR is the inappropriate bundling, which is part of what we may want to call "manufactured friction".  

Our Developer Guide is clear about the importance of keeping PRs narrowly focused. https://doc.sagemath.org/html/en/developer/github.html#planning-an-enhancement
https://doc.sagemath.org/html/en/developer/review.html#reasons-to-invalidate-prs

This PR, as described in the PR description above, could be really simple and straightforward, as it would fit seamlessly into the existing design of sage-conf -- with a suitably focused implementation and participation in the normal review process. Just make a copy of `sage-conf_conda` and replace a little bit of it. 

But all of the following are unnecessary for the goal of the PR and create a disproportionate, gratuitous effort for everyone involved -- counting 80 comments so far:
- the SAGE_ROOT/meson.build file, unnecessary for the goal of the PR 
- leaving out the sage-config script, which is part of the documented API, and insisting that it does not matter
- demanding that reviewers justify the status quo.

meson.build
tobiasdiez: > * When people see a file `meson.build` next to `configure.ac` and `Makefile`, they are bound to think that using meson is an alternative for building the same thing. So this has the potential to mislead.

So if you see a hammer and a saw next to each other you are bound to see them as alternatives for building the same thing?

Conceptually, `meson setup` does the same thing as `configure`: go through the declared dependencies and write out a config file. Similarly, `meson compile` will do the same thing as `make sage-all` (or whatever the correct target is for building sagelib). 

It's also not technically possible to move the meson file in the root to `src` since meson's `subdir` only walks down the tree.

Anyway, I'm tired of this stupid discussion and will stop participating in it. If you think that the harm created by the meson file in the root outweighs the positives of this PR, then please go ahead and continue blocking it. Please leave the "ready for review" tag on, as there are no work items that referee and PR creator agree on (it just that you don't like a small design choice in the PR).

tobiasdiez:
@mkoeppe "Please leave the "ready for review" tag on, as there are no work items that referee and PR creator agree on (it just that you don't like a small design choice in the PR)." from above. Thanks!

meson.build
mkoeppe: > Similarly, `meson compile` will do the same thing as `make sage-all` (or whatever the correct target is for building sagelib).

The target `make sage-all` does not exist. `make all`, the default target, builds the Sage distribution.

mkoeppe:
There are clear work items.

meson.build
mkoeppe: > Sorry, but there are a lot of files in the root folder that are not related to the distribution. 
> [...] tox ini [...]

Actually `SAGE_ROOT/tox.ini` defines the portability tests of the Sage distribution.

tobiasdiez:
No longer needed.

PR #36492, opened 2023-10-21
Conditional Oriented Matroids on SageMath

he-rand:
This package is proposed for optional inclusion: SageMath functions manipulating conditional oriented matroids. Among other functions, the check if a set corresponds to a conditional oriented matroid, the generating of that latter from its tope set, its Varchenko determinant computing, and the solving of its Aguiar-Mahajan system are implemented.
[Conditional Oriented Matroids.pdf](https://github.com/sagemath/sage/files/13060177/Conditional.Oriented.Matroids.pdf)


jhpalmieri:
Can you please clarify what you have in mind with this? Sage doesn't currently distribute any worksheet files. Maybe it's not a bad idea to have a `contrib` directory and put such files there, but until that is done, what should be done with this?

he-rand:
> Can you please clarify what you have in mind with this? Sage doesn't currently distribute any worksheet files. Maybe it's not a bad idea to have a `contrib` directory and put such files there, but until that is done, what should be done with this?

Thank you very much for your message. I wish to propose this package for optional inclusion. For that, I wanted to add the label c: packages: optional , however did not find a way to insert in the pull request. Or is there another thing I should to get this package considered as optional inclusion?

jhpalmieri:
> > Can you please clarify what you have in mind with this? Sage doesn't currently distribute any worksheet files. Maybe it's not a bad idea to have a `contrib` directory and put such files there, but until that is done, what should be done with this?
> 
> Thank you very much for your message. I wish to propose this package for optional inclusion. For that, I wanted to add the label c: packages: optional , however did not find a way to insert in the pull request. Or is there another thing I should to get this package considered as optional inclusion?

As I tried to imply before, there is no obvious place to put this kind of contribution in Sage. Your proposal creates a new directory `conditional-oriented-matroids` on the top level, on equal footing with `src` (the entire Sage library) and `build` (the build infrastructure). I don't think that's an appropriate location. Here are a few suggestions: (1) look at the structure of the Sage distribution and figure out where it might make sense to include something like this. (2) For the code itself, please include documentation. Take a look at the Sage library code, for example https://github.com/sagemath/sage/blob/develop/src/sage/matroids/catalog.py#L56: every function has documentation, including examples. This is described in the [Sage Developer's Guide](https://doc.sagemath.org/html/en/developer/coding_basics.html). (3) For the code itself, how are users expected to use it? That's something else you should explain.

jhpalmieri:
With the current structure of the Sage library, one possible option would be to just add your code (once it follows the guidelines) into the Sage library in `src/sage/matroids`. Perhaps it could be supplemented with a [thematic tutorial](https://doc.sagemath.org/html/en/thematic_tutorials/) illustrating the use, or perhaps just the documentation itself would be sufficient.

he-rand:
> > > Can you please clarify what you have in mind with this? Sage doesn't currently distribute any worksheet files. Maybe it's not a bad idea to have a `contrib` directory and put such files there, but until that is done, what should be done with this?
> > 
> > 
> > Thank you very much for your message. I wish to propose this package for optional inclusion. For that, I wanted to add the label c: packages: optional , however did not find a way to insert in the pull request. Or is there another thing I should to get this package considered as optional inclusion?
> 
> As I tried to imply before, there is no obvious place to put this kind of contribution in Sage. Your proposal creates a new directory `conditional-oriented-matroids` on the top level, on equal footing with `src` (the entire Sage library) and `build` (the build infrastructure). I don't think that's an appropriate location. Here are a few suggestions: (1) look at the structure of the Sage distribution and figure out where it might make sense to include something like this. (2) For the code itself, please include documentation. Take a look at the Sage library code, for example https://github.com/sagemath/sage/blob/develop/src/sage/matroids/catalog.py#L56: every function has documentation, including examples. This is described in the [Sage Developer's Guide](https://doc.sagemath.org/html/en/developer/coding_basics.html). (3) For the code itself, how are users expected to use it? That's something else you should explain.

I have already a documentation at the beginning of the conversation as pdf. It is attached to this reply. Or maybe I should have uploaded it in another location or in mandatory in another form?
[Conditional Oriented Matroids.pdf](https://github.com/sagemath/sage/files/13266297/Conditional.Oriented.Matroids.pdf)

he-rand:
> With the current structure of the Sage library, one possible option would be to just add your code (once it follows the guidelines) into the Sage library in `src/sage/matroids`. Perhaps it could be supplemented with a [thematic tutorial](https://doc.sagemath.org/html/en/thematic_tutorials/) illustrating the use, or perhaps just the documentation itself would be sufficient. I don't know anything about matroids, and so I don't know how specialized this material is and whether it would make sense as part of the Sage library.

Thank you very much. I will look for the best option for it.

jhpalmieri:
> > > > Can you please clarify what you have in mind with this? Sage doesn't currently distribute any worksheet files. Maybe it's not a bad idea to have a `contrib` directory and put such files there, but until that is done, what should be done with this?
> > > 
> > > 
> > > Thank you very much for your message. I wish to propose this package for optional inclusion. For that, I wanted to add the label c: packages: optional , however did not find a way to insert in the pull request. Or is there another thing I should to get this package considered as optional inclusion?
> > 
> > 
> > As I tried to imply before, there is no obvious place to put this kind of contribution in Sage. Your proposal creates a new directory `conditional-oriented-matroids` on the top level, on equal footing with `src` (the entire Sage library) and `build` (the build infrastructure). I don't think that's an appropriate location. Here are a few suggestions: (1) look at the structure of the Sage distribution and figure out where it might make sense to include something like this. (2) For the code itself, please include documentation. Take a look at the Sage library code, for example https://github.com/sagemath/sage/blob/develop/src/sage/matroids/catalog.py#L56: every function has documentation, including examples. This is described in the [Sage Developer's Guide](https://doc.sagemath.org/html/en/developer/coding_basics.html). (3) For the code itself, how are users expected to use it? That's something else you should explain.
> 
> I have already a documentation at the beginning of the conversation as pdf. It is attached to this reply. Or maybe I should have uploaded it in another location or in mandatory in another form? [Conditional Oriented Matroids.pdf](https://github.com/sagemath/sage/files/13266297/Conditional.Oriented.Matroids.pdf)

As I said, every function should have its own documentation, and that documentation is part of the function's definition:
```
def my_function(x, y, z):
    """
    Return ...

    INPUTS:

    - ``x`` -- an integer
    - ``y`` -- a matrix
    - ``z`` -- a real number

    More information about what the function does

    EXAMPLES::

        sage: my_function(3, identity_matrix(2), 3.1416)
        37

        sage: more examples which test all parts of the actual code
    """
    Python code defining the function...
```
Users should be able to get help by typing `my_function?`, and there should be examples that can be tested using Sage's doctesting framework. As I said, please read the developer's guide and look at the source code for the Sage library.

he-rand:
@jhpalmieri The code is now written according to the Sage Developer's Guide. Besides, the index.rst file was also modified. I would suggest to follow your first proposition, that is why I created the contrib directory "com" for the code. But, I naturally remain open to any suggestion. I look forward to hearing from you.

tscrim:
You should follow good object-oriented-programming principles with having a good set of classes that group together common functionality and features.

You should also follow standard Python conventions (mainly, see PEP8) with functions/methods being `lower_case` and class names being `CamelCase`.

Avoid using `Set` and instead use `frozenset`.

You should have documentation that provides definitions, not just references, which will indicate the restrictions on the input types. This also goes to @jhpalmieri previous comment: the doctests and simple descriptions don't carry enough information for me to tell what they do.

Your doctests won't pass because you need to also have the appropriate import statements in your doctests.

tscrim:
This should not be added as a new top-level directory, nor contain a copy of the GPL (only a simple statement in the top of the files is sufficient), nor the README file. As @jhpalmieri said, this would be most naturally in the matroids folder.

In order for me to consider the code acceptable for Sage, you must implement this using object-oriented-programming paradigms. More specifically, things should be organized into `class`es with methods that act on them. For example, a `SignSystem`.

You should also avoid the Sage `Set`  (which is slow) in favor of Python's (`frozen`)`set`.

Within your doctests, you need a `sage: from path.to.file import Foo` as each method's doctests are run as if it was a fresh Sage session. Related, you should *not* import everything into the global namespace; you're not currently, but I want to make it clear that that is not how it should be fixed.

dimpase:
@thecaligarmo - can you have a look, how big is the overlap with your package?

thecaligarmo:
> @thecaligarmo - can you have a look, how big is the overlap with your package?

Looking over it quickly, there is some overlap. In particular, here is what I notice from a quick glance.

Similarities
1. They also allow for checking if something is an oriented matroid.
2. Checks for negatives, zeroes, support, composition, seperation, tope, face, partial order, loops, parallel elements, restriction, deletion

Differences:
1. They seem to be using procedural code whereas mine is using OOP
2. They have a few extra functions that I haven't implemented yet (contraction, generalized mandel, the Aguiar-Mahajan distance between topes, Varchenko matrix/determinant)
3. Their system only allows for one way to define oriented matroids, whereas ours allows for various ways to define oriented matroids (and a system to add more ways as there are numerous cryptomorphic definitions for oriented matroids). 
4. Our system allows for changing the "type" of oriented matroid the user is working with (as in, you can change which definition/structure you are using which makes some calculations/ideas faster/easier.)
5. Our system has some extra functions theirs doesn't. Such as: getting underlying matroid, getting the rank, returning the face poset, on whether the oriented matroid is simplicial or not (is acyclic or not, is simple or not)

I think some of this stuff we can add into my current package although I haven't worked with conditional oriented matroids in order to know whether that particular part can also be included. (From their paper linked above, seems like oriented matroids are conditional oriented matroids with some additional property, but I don't know if this is true for every cryptomorphic definition, or just this one particular definition). I also don't know much about Lopsided systems.

Issue #36486, opened 2023-10-19
Doctest failures with networkx 3.2


tornaria:
### Steps To Reproduce

```
sage -t --random-seed=324850534885798131910668214652534087473 src/sage/graphs/generic_graph.py
**********************************************************************
File "src/sage/graphs/generic_graph.py", line 5105, in sage.graphs.generic_graph.GenericGraph.cycle_basis
Failed example:
    g.cycle_basis()                                                       # needs networkx
Expected:
    [[1, 6, 8, 5, 0], [4, 9, 6, 8, 5, 0], [7, 9, 6, 8, 5],
     [4, 3, 8, 5, 0], [1, 2, 3, 8, 5, 0], [7, 2, 3, 8, 5]]
Got:
    [[6, 8, 5, 7, 9],
     [2, 3, 8, 5, 7],
     [4, 3, 8, 5, 7, 9],
     [4, 0, 5, 7, 9],
     [2, 1, 0, 5, 7],
     [6, 1, 0, 5, 7, 9]]
**********************************************************************
File "src/sage/graphs/generic_graph.py", line 5111, in sage.graphs.generic_graph.GenericGraph.cycle_basis
Failed example:
    g.cycle_basis(output='edge')                                          # needs networkx
Expected:
    [[(1, 6, None), (6, 8, None), (8, 5, None), (5, 0, None),
     (0, 1, None)], [(4, 9, None), (9, 6, None), (6, 8, None),
     (8, 5, None), (5, 0, None), (0, 4, None)], [(7, 9, None),
     (9, 6, None), (6, 8, None), (8, 5, None), (5, 7, None)],
     [(4, 3, None), (3, 8, None), (8, 5, None), (5, 0, None),
     (0, 4, None)], [(1, 2, None), (2, 3, None), (3, 8, None),
     (8, 5, None), (5, 0, None), (0, 1, None)], [(7, 2, None),
     (2, 3, None), (3, 8, None), (8, 5, None), (5, 7, None)]]
Got:
    [[(6, 8, None), (8, 5, None), (5, 7, None), (7, 9, None), (9, 6, None)],
     [(2, 3, None), (3, 8, None), (8, 5, None), (5, 7, None), (7, 2, None)],
     [(4, 3, None),
      (3, 8, None),
      (8, 5, None),
      (5, 7, None),
      (7, 9, None),
      (9, 4, None)],
[...]

PR #36493, opened 2023-10-21
improve method `cycle_basis` for graphs with multiple edges

dcoudert:
We use the notion of nearest common ancestor in a tree to find cycles more efficiently than using DFS.
This is used only when the graph allows multiple edges, since otherwise we use the method of networkx. Therefore, the failing doctests reported in #36486 when using networkx 3.2 are independent from the changes done here.

An additional optimization would be to determine the minimum spanning tree, the ranks and the parents at the same time. We don't do it here.

### :memo: Checklist


src/sage/graphs/generic_graph.py
```diff
@@ -5192,11 +5192,11 @@ def cycle_basis(self, output='vertex'):
...
             sage: G.cycle_basis()                                                       # needs networkx
-            [[2, 3], [4, 3, 2, 1], [4, 3, 2, 1]]
+            [[3, 2], [4, 1, 2, 3], [4, 1, 2, 3]]
             sage: G.cycle_basis(output='edge')                                          # needs networkx
             [[(2, 3, 'c'), (3, 2, 'b')],
-             [(4, 3, 'd'), (3, 2, 'b'), (2, 1, 'a'), (1, 4, 'f')],
-             [(4, 3, 'e'), (3, 2, 'b'), (2, 1, 'a'), (1, 4, 'f')]]
+             [(3, 4, 'd'), (4, 1, 'f'), (1, 2, 'a'), (2, 3, 'b')],
+             [(3, 4, 'e'), (4, 1, 'f'), (1, 2, 'a'), (2, 3, 'b')]]
         """
```
kwankyu, line 5200: Then for these cases, you may remove `# needs networkx`?

kwankyu:
Looks good to me though I don't know the algorithm.

Can you provide some benchmarks showing that the new algorithm is more efficient?

dcoudert:
Some benchmarks on the graphs used in the doctests:
```py
sage: G = Graph([(0, 2, 'a'), (0, 2, 'b'), (0, 1, 'c'), (1, 2, 'd')], multiedges=True)
sage: H = Graph([(1, 2), (2, 3), (2, 3), (3, 4), (1, 4), (1, 4), (4, 5), (5, 6), (4, 6), (6, 7)], multiedges=True)
```
Before:
```py
sage: %timeit G.cycle_basis()
57.3 µs ± 177 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit H.cycle_basis()
112 µs ± 190 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
After:
```py
sage: %timeit G.cycle_basis()
42.5 µs ± 104 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit H.cycle_basis()
56.8 µs ± 176 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

and a larger graph
```py
sage: G = graphs.Grid2dGraph(10, 10)
sage: G.allow_multiple_edges(True)
sage: G.add_edges(G.edges())
```
Before
```py
sage: %timeit G.cycle_basis()
48.1 ms ± 359 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
After
```py
sage: %timeit G.cycle_basis()
1.13 ms ± 4.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

kwankyu (approved):
Thanks. Brilliant.

kwankyu:
If you want to attract attention from graph theory experts for more time, you may discard my review.

dcoudert:
Thank you.

tobiasdiez:
Conda workflow fails because of this PR: https://github.com/sagemath/sage/actions/runs/6700323341/job/18206090258#step:11:10801 (with python 3.9)

tobiasdiez:
No problem 🐱 . Yes, a follow-up compatible with 3.9 would be good (since its still a bit until we can talk about dropping this version).

dcoudert:
This is fixed in #36609

PR #36469, opened 2023-10-16
Revert to non-incremental docbuild

tobiasdiez:
The doc build workflow currently fails because it runs out of space (e.g https://github.com/sagemath/sage/actions/runs/6527955359/job/17723528313?pr=36466). The incremental build before is also not really working, as the sagelib is still built again. We remove the incremental build for now, until https://github.com/sagemath/sage/pull/36459, and use an older image for now.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36349, opened 2023-09-28
ci: improve things around auto-apply of fixes

tobiasdiez:
Fixes some of the mess that was introduced by merging https://github.com/sagemath/sage/pull/36338. In particular,
- Having PRs open with blocker label no longer influences workflows triggered by push to the main branch (see eg https://github.com/sagemath/sage/actions/runs/6332061127/job/17197852786)
- Reduce the number of "checks" displayed below the PR 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36338, opened 2023-09-25
CI: Merge open blocker PRs in all CI workflows + other improvements

mkoeppe:
The new script `.ci/merge-fixes.sh` uses the GitHub CLI to retrieve open PRs with the "blocker" label from sagemath/sage and tries to merge them into the tested branch, ignoring any PRs that give merge failures.

We call this script in all CI workflows. Thus, 
- hotfixes for the linter (such as #36327) or 
- fixes for build problems (such as #36276)
 
will take effect in the CI workflows of all PRs.

This can be seen in the runs of the workflows of the present PR (section "Merge CI fixes from sagemath/sage").

Additionally,
- we fix `debian-trixie`, which failed because a system package is no longer available
- we fix `macos-conda-maximal` by removing some fictitious distros/conda.txt files
- we refactor ci-macos using a new reusable workflow; example runs: 
   - https://github.com/mkoeppe/sage/actions/runs/6308124398
   - https://github.com/sagemath/cysignals/actions/runs/6316714387?pr=185
- we update the ci-macos platforms, including a configuration for the new and problematic Xcode 15.0 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #35652, opened 2023-05-19
CI Build & Test: First build incrementally and test changed files only

mkoeppe:
### :books: Description

This gives a faster turnaround for errors in the committed changes. 
When the incremental build (or the test) fails, we clean thoroughly and build sagelib from scratch (= what happened before this PR.)

Changing the docbuild CI in the same way.

The full doctests are now run with `--long`.

Also:
- Fixes `sage -t --new` so it works in a git worktree and respects file-level annotations `# sage.doctest: optional - ...`
- Resolves #35732

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #35751 (merged here)

PR #36498, opened 2023-10-21
CI build, doc-build: Run containers explicitly, separate jobs for pyright, build, modularized tests, long tests

### LINKED ISSUE #35814: build.yml: Report failures (e.g. from incremental test) earlier
as suggested in https://github.com/sagemath/sage/pull/35652#issuecomment-1603365433

possibly using https://github.com/LouisBrunner/checks-action

and/or problem matchers for:
- sphinx (see https://github.com/python/cpython/pull/20325/files)
- linter errors
- doctest failures

mkoeppe:
Running the containers explicitly, instead of using the declarative `container:` feature of GH Actions gives us more control:
- we can create more space on the host if necessary; we just scraped by an out of space condition in #36473 / https://github.com/sagemath/sage/pull/36469#issuecomment-1765624306
- we can run some operations outside of the container but in the same job; this will make the separate "Get CI fixes" jobs unnecessary, addressing the cosmetic concerns from https://github.com/sagemath/sage/pull/36338#discussion_r1337974043, #36349
- it enables caching between the various workflows (as first discussed in #36446).

We split out static type checking with pyright into its separate workflow:
- **pyright.yml**: As static checking does not need  a build of the Sage library, for PRs that do not make any changes to packages, there's nothing to build, and the workflow gives a fast turnaround just after 10 minutes. For applying the CI fixes from blocker tickets, this workflow uses the technique favored in #36349.

The workflow **build.yml** first launches a job:
- **test-new:** It builds incrementally (using a tox-generated `Dockerfile` and https://github.com/marketplace/actions/build-and-push-docker-images) and does the quick incremental test. This completes within 10 to 20 minutes when there's no change.

After this is completed, more jobs are launched:
- **test-mod:** It again builds incrementally and tests a modularized distribution. Later (with more from #35095), more jobs will be added to this matrix job for other distributions.
- **test-long:** It again builds incrementally and runs the long test.

The workflows **doc-build.yml** and **doc-build-pdf.yml** again build incrementally and then build the documentation. The diffing code for the HTML documentation now runs in the host, not the container, which simplifies things. (To show that diffing still works, we include a small change to the Sage library.)

Splitting the workflow jobs implements @kwankyu's  suggestion from:
- https://github.com/sagemath/sage/pull/35652#issuecomment-1603365433 (Fixes #35814)

The steps "again builds incrementally" actually use the GH Actions cache, https://docs.docker.com/build/ci/github-actions/cache/#cache-backend-api. When nothing is cached and the 3 workflows are launched in parallel, they may each build the same thing. But when there's congestion that leads to the workflows to be run serially, the 2nd and 3rd workflow will be able to use the cache from the 1st workflow. This elasticity may be helpful in reducing congestion.

There is a rather small per-project limit of 10 GB for this cache, so we'll have to see how effectively caching works in practice. See https://github.com/sagemath/sage/actions/caches

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36938 (merged here)
- Depends on #36951 (merged here)
- Depends on #37351 (merged here)
- Depends on #37750 (merged here)
- Depends on #37277 (merged here)


kwankyu:
`src/sage/schemes/toric/variety.py` sneaked in?

mkoeppe:
via https://github.com/sagemath/sage/pull/36443

tobiasdiez:
I don't get the point of this PR. It seems to mostly complicate the setup and essentially reimplements the container feature of github. For the out-of-space issue it would be nice to first know what's the route cause to be able to evaluate if steps to fixing it go into the correct direction. For potential speedups by caching the docker image, I'm not sure if this is really needed. Yes, from a conceptual viewpoint you only want to build sagelib once, and then reuse it - but well, it mostly takes 2-10 mins of incremental compilation anyway. Personally, I think we have more pressing issues with the ci setup and better not introduce more moving parts (atm at least).

mkoeppe:
> For the out-of-space issue it would be nice to first know what's the root cause to be able to evaluate if steps to fixing it go into the correct direction.

It's really not very complicated. GitHub runs the container from its very full runner image. We know that it does not have enough space for our large containers.

mkoeppe:
> It seems to [...] essentially reimplements the container feature of github.

That's right. The declarative `container` feature is convenient, but we have run into the two restrictions described in the PR description. So we just use docker directly; problem solved.

mkoeppe:
> Yes, from a conceptual viewpoint you only want to build sagelib once, and then reuse it - but well, it mostly takes 2-10 mins of incremental compilation anyway.

Actually, the compilation can take much longer when low level parts of sagelib are touched, and when certain packages are updated.

tobiasdiez:
> > For the out-of-space issue it would be nice to first know what's the root cause to be able to evaluate if steps to fixing it go into the correct direction.
> 
> It's really not very complicated. GitHub runs the container from its very full runner image. We know that it does not have enough space for our large containers.

Github provides 14gb of space (after its runner image), that should be sufficient (and was sufficient) to run a 7gb image in it and install sage. Also our image got bigger by a few houndred mbs.

mkoeppe:
If you want to understand more details, I'd suggest that you experiment with the space parameters in https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml#L148

tobiasdiez:
It's already quite easy to run commands outside the container. Here is an example for exactly your first use case: https://github.com/osquery/osquery/blob/a916e80a895f8abb9b749161d774d528d0466b31/.github/workflows/hosted_runners.yml#L257-L263 I don't think we need more flexibility.

mkoeppe:
> It's already quite easy to run commands outside the container. Here is an example for exactly your first use case: https://github.com/osquery/osquery/blob/a916e80a895f8abb9b749161d774d528d0466b31/.github/workflows/hosted_runners.yml#L257-L263 

Thanks for finding this.

> I don't think we need more flexibility.

Well, I already know that we do need it, namely for sharing the build, which I'll do in a follow up.

kwankyu:
We are getting ["Out of space" errors](https://github.com/sagemath/sage/actions/runs/8607601590) in CI.

Is this PR ready and will it cure the out-of-space errors?

mkoeppe:
> We are getting ["Out of space" errors](https://github.com/sagemath/sage/actions/runs/8607601590) in CI.

Oh! That's new.

> Is this PR ready and will it cure the out-of-space errors?

Yes, it's ready. I'm not sure if I fully tested the live-doc deployment with it though

.github/workflows/build.yml
```diff
@@ -26,192 +26,308 @@ concurrency:
...
+  BUILD_IMAGE:            "localhost:5000/${{ github.repository }}/sage-${{ github.event.inputs.platform || 'ubuntu-jammy-standard' }}-with-targets:ci"
+  FROM_DOCKER_REPOSITORY: "ghcr.io/sagemath/sage/"
+  FROM_DOCKER_TARGET:     "with-targets"
+  FROM_DOCKER_TAG:        ${{ github.event.inputs.docker_tag || 'dev'}}
+  EXTRA_CONFIGURE_ARGS:   --enable-fat-binary
+
 jobs:
-  get_ci_fixes:
+  build:
     runs-on: ubuntu-latest
```
kwankyu, line 80: How about
```
test-new  test-mod  test-long
```
instead of
```
build     mod       test-long
```
which are obscure now?

In each job, a docker image is built. What are shared between these docker builds? Layers?

kwankyu: Ah, current working directory is shared between the containers. Right?

.github/workflows/doc-build-pdf.yml
```diff
@@ -20,96 +20,109 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  # Same as in build.yml
+  TOX_ENV:                "docker-${{ github.event.inputs.platform || 'ubuntu-jammy-standard' }}-incremental"
+  BUILD_IMAGE:            "localhost:5000/${{ github.repository }}/sage-${{ github.event.inputs.platform || 'ubuntu-jammy-standard' }}-with-targets:ci"
+  FROM_DOCKER_REPOSITORY: "ghcr.io/sagemath/sage/"
```
kwankyu, line 27: Does `localhost:5000` in `BUILD_IMAGE` mean that the built image is stored in the local registry?

What is the local registry for? The built image is shared among different jobs?

.github/workflows/doc-build.yml
```diff
@@ -9,140 +9,161 @@ on:
...
           git config --global --add safe.directory $(pwd)
           git config --global user.email "ci-sage@example.com"
-          git config --global user.name "Build & Test workflow"
-          # mathjax path in old doc
-          mathjax_path_from=$(SAGE_USE_CDNS=no /sage/sage -python -c "from sage_docbuild.conf import mathjax_path; print(mathjax_path)")
-          .ci/retrofit-worktree.sh worktree-image /sage
+          git config --global user.name "Build documentation workflow"
+          # mathjax path in old doc (regex)
+          mathjax_path_from="[-./A-Za-z_]*/tex-chtml[.]js"
           # mathjax path in new doc
```
kwankyu, line 124: The original way to get `mathjax_path_from` seems safer.

.github/workflows/pyright.yml
```diff
@@ -0,0 +1,84 @@
...
+concurrency:
+  # Cancel previous runs of this workflow for the same branch
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pyright:
+    runs-on: ubuntu-latest
+    container: ghcr.io/sagemath/sage/sage-ubuntu-focal-standard-with-targets:dev
+    steps:
```
kwankyu, line 22: No `jammy`?

build/bin/write-dockerfile.sh
```diff
@@ -55,7 +55,10 @@ case $SYSTEM in
                 ;;
             *)
                 cat <<EOF
-RUN (yes | unminimize) || echo "(ignored)"
+RUN if command -v unminimize > /dev/null; then  \
+        (yes | unminimize) || echo "(ignored)"; \
+        rm -f "$(command -v unminimize)";       \
+    fi
 EOF
```
kwankyu, line 72: What is `unminimize`? Why is it removed?

.github/workflows/build.yml
mkoeppe: Nothing is shared explicitly between the different jobs. Per PR description:
> The steps "again builds incrementally" actually use the GH Actions cache, https://docs.docker.com/build/ci/github-actions/cache/#cache-backend-api. When nothing is cached and the 3 workflows are launched in parallel, they may each build the same thing. But when there's congestion that leads to the workflows to be run serially, the 2nd and 3rd workflow will be able to use the cache from the 1st workflow. This elasticity may be helpful in reducing congestion.

.github/workflows/doc-build-pdf.yml
mkoeppe: It's just a very robust way to have the "docker build action" store its result, for loading it later.

It's not shared between the different jobs; there's no mechanism for doing that. Each job is separate.

build/bin/write-dockerfile.sh
mkoeppe: The Ubuntu Docker images are "minimized", meaning that some large bits such as documentation has been removed. We have to unminimize once (which reinstalls the full versions of some minimized packages), or e.g. the maxima documentation (which we depend on for correct operation) will be missing (IIRC). But we only have to do this once. To save time in incremental builds, I'm removing the `unminimize` binary here.

.github/workflows/build.yml
mkoeppe: > How about
> 
> ```
> test-new  test-mod  test-long
> ```

Thanks. I've renamed it as suggested

.github/workflows/pyright.yml
mkoeppe: Thanks for catching this; fixed

.github/workflows/build.yml
kwankyu: Then the workflow may take longer than before (where there is only one job), because each job builds the image and the container anew?

mkoeppe: Only when there is so much pressure on the cache that the layers are no longer there when the next of these jobs is started. Haven't observed that yet.

.github/workflows/doc-build.yml
mkoeppe: Perhaps but I would have to start the old container just for that...

.github/workflows/build.yml
kwankyu: So this
```
          cache-from: type=gha
          cache-to:   type=gha,mode=max
```
sets up the cache? If this cache is local to a job, then how could this allow layers to be shared (via the cache) between three docker image buildings...?

.github/workflows/doc-build-pdf.yml
kwankyu: OK.

.github/workflows/doc-build.yml
kwankyu: No. This
```
 # mathjax path in old doc
          mathjax_path_from=$(SAGE_USE_CDNS=no /sage/sage -python -c "from sage_docbuild.conf import mathjax_path; print(mathjax_path)")
```
just runs the same sage but with `SAGE_USE_CDNS=no`. You may run the same container for `mathjax_path_to`.

kwankyu: Yes, this would take a couple of seconds more though.

build/bin/write-dockerfile.sh
kwankyu: OK. 

Generally speaking, I hope that any cryptic part of a script is annotated with ample comments like the one you just gave, to ease the work of the future maintainer of our CI infrastructure.

.github/workflows/build.yml
mkoeppe: No, the cache is global.

kwankyu: I see. That's the reason.

build/bin/write-dockerfile.sh
mkoeppe: Great suggestion, done in 17583ab633a and c18ba755eba

kwankyu (approved):
This PR looks good to me, except the `mathjax_path` thing, which is a minor issue.

mkoeppe:
Thanks! I think I'd like to keep the `mathjax_path` as is, at least for now.

.github/workflows/doc-build.yml
kwankyu: OK for now.

kwankyu:
We need this PR to be merged ASAP.

mkoeppe:
Thanks for the review!

vbraun:
merge conflict

.github/workflows/doc-build.yml
```diff
@@ -9,140 +9,161 @@ on:
...
```
kwankyu, line 124: Note that we are suffering from spurious diffs like
```diff
     <script>window.MathJax = {"tex": {"macros": {"ZZ": "\\Bold{Z}", "NN": "\\Bold{N}", "RR": "\\Bold{R}", "CC": "\\Bold{C}", "QQ": "\\Bold{Q}", "QQbar": "\\overline{\\QQ}", "GF": ["\\Bold{F}_{#1}", 1], "Zp": ["\\Bold{Z}_{#1}", 1], "Qp": ["\\Bold{Q}_{#1}", 1], "Zmod": ["\\ZZ/#1\\ZZ", 1], "CDF": "\\Bold{C}", "CIF": "\\Bold{C}", "CLF": "\\Bold{C}", "RDF": "\\Bold{R}", "RIF": "\\Bold{I} \\Bold{R}", "RLF": "\\Bold{R}", "SL": "\\mathrm{SL}", "PSL": "\\mathrm{PSL}", "lcm": "\\mathop{\\operatorname{lcm}}", "dist": "\\mathrm{dist}", "Bold": ["\\mathbf{#1}", 1]}, "inlineMath": [["$", "$"], ["\\(", "\\)"]], "maxBuffer": 51200, "autoload": {"color": [], "colorv2": ["color"]}}}</script>
-    <script defer="defer" src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml.js?v=0d9741c5"></script>
+    <script defer="defer" src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml.js"></script>
     <script src="_static/jupyter-sphinx-furo.js?v=a8e52432"></script>
```
See the parameter string `?v=0d9741c5`. This is the checksum of the file `tex-chtml.js` automatically added by (recent) Sphinx. Would you add a regex to catch this string as well? 

I guess that adding `?v=[0-9a-f]*` to the end of `mathjax_path_from` in line 123 would do it. But I am not sure since I didn't experiment, and doing regex correct is quite tricky (even more within sed).

kwankyu: `mathjax_path_from="[-./A-Za-z_]*/tex-chtml[.]js?v=[0-9a-f]*"` works locally, that is, on mac.

mkoeppe: Thanks; done in a6224a3f04b

kwankyu: Thank you!

PR #36380, opened 2023-10-01
Demote brial (= polybori) from standard to optional, add distribution sagemath-brial, enlarge sagemath-objects, sagemath-categories

mkoeppe:
as proposed in https://groups.google.com/g/sage-devel/c/ewHP20bmskQ, because (among other things):
It has been dropped from Debian testing (`debian-trixie`) and `ubuntu-noble`, where it seems to block SageMath upgrades (Sage is stuck at 9.5 in Debian/Ubuntu). See https://repology.org/project/brial/versions

- Doctests are already conditional on the dynamically detected feature `sage.rings.polynomial.pbori`
- We create a distribution **sagemath-brial** that packages the corresponding extension modules (cherry-picked from #35095).
- Documentation is conditionalized by using Sphinx tags 

The new distribution builds on top of enlarged distributions **sagemath-objects**, **sagemath-categories**. This is a major update of these distributions, which makes them much more useful.

Also includes style changes to `all*.py` files cherry-picked from #37900.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36495 (split out from here)
- ~~Depends on #36566 (split out from here)~~
- Depends on #37857 (merged here)
- Depends on #38088 (merged here)
- Depends on #38143 (merged here)
- Depends on #38142 (merged here)
- Depends on #38180 (split out from here)
- Depends on #38200

PR #36503, opened 2023-10-22
Add config for ruff

tobiasdiez:
[Vscode removed the (built-in) support for pycodestyle and now recommends to use ruff.](https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions). For this reason, we add ruff config. The config corresponds to an "ideal version" (instead of a minimal version that is in the linter-ci), so that devs don't introduce unnecessary new issues that would need fixing later.
There are probably a few rules that we can disable as they don't apply / cause to many false negatives. I would say we start with this maximal config and then disable these offending rules later as we get more experience. A notable exception is the line length rule, which I've directly disabled as there is no point in enforcing it currently.

Next step: use it in ci, this is now https://github.com/sagemath/sage/pull/36512

### :memo: Checklist


pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.ruff]
```
mkoeppe, line 1: Please, no pyproject.toml in SAGE_ROOT. 
SAGE_ROOT is not a Python package.

mkoeppe:
Next step = create a tox environment.

pyproject.toml
tobiasdiez: It applies to all python code (e.g. also the one in pkgs), so should be in the root. It's also quite common that the pyproject config is in the root, but the actual code in a subfolder. eg https://github.com/scipy/scipy/blob/main/pyproject.toml or https://github.com/pandas-dev/pandas/blob/main/pyproject.toml

mkoeppe: > It applies to all python code (e.g. also the one in pkgs), so should be in the root.

No, that's not how the Sage repository is structured. Again, SAGE_ROOT is not the source directory of a Python package (distribution)... which is why there is no setup.py, no pyproject.toml, no setup.cfg, etc.

tobiasdiez: How is that different from scipy?

mkoeppe: The source directory of the Python distribution package scipy (per PEP 518, PEP 621, the location of the `pyproject.toml` file) is the root of its repository https://github.com/scipy/scipy/tree/main

The source directory of the Python distribution package sagemath-standard is not the root of the repository https://github.com/sagemath/sage (= SAGE_ROOT), but rather `src/`. As I know that you know, this is where you find `src/pyproject.toml`, `src/setup.py`, etc.

tobiasdiez: I didn't changed anything about a distribution package. See it similar to the tox.ini file.

Do you also have a suggestion on how to apply config otherwise for tools that look at pyproject and that should apply to all python code?

mkoeppe: > Do you also have a suggestion on how to apply config otherwise for tools that look at pyproject and that should apply to all python code?

By design, all source code of the Sage library is in the monolithic source directory `SAGE_ROOT/src`. So there is no need for tools to look at the source files of the modularized distribution packages (sagemath-environment etc.)
Some special distribution packages (currently only pkgs/sage-sws2rst, pkgs/sage-conf*) do have their own source files, so tool configuration could go into pkgs/sage-sws2rst/pyproject.toml etc.

However, Ruff also supports configuration using `ruff.toml` https://docs.astral.sh/ruff/configuration/#using-rufftoml, so this would be a good way to put the tool configuration that is "global to the repository" there

tobiasdiez: > Some special distribution packages (currently only pkgs/sage-sws2rst, pkgs/sage-conf*) do have their own source files, so tool configuration could go into pkgs/sage-sws2rst/pyproject.toml etc.

That doesn't scale. I'll not add ruff config to each pyproject toml file.

> However, Ruff also supports configuration using `ruff.toml` https://docs.astral.sh/ruff/configuration/#using-rufftoml, so this would be a good way to put the tool configuration that is "global to the repository" there

I think there is a reason why we put all the config's in tox ini so far instead of having 20 config files for all tools.

Not everything that has a pyproject file is a "distribution package". Especially not if it only contains tool config...

mkoeppe: > > Some special distribution packages (currently only pkgs/sage-sws2rst, pkgs/sage-conf*) do have their own source files, so tool configuration could go into pkgs/sage-sws2rst/pyproject.toml etc.
> 
> That doesn't scale.

It does not have to scale. As I just explained to you, there is a fixed short list of packages.

mkoeppe: > Not everything that has a pyproject file is a "distribution package".

That's the very idea of pyproject.toml. It marks the top of a Python distribution package (= project). That's what "py" and "project" refer to in the name of this file.

tobiasdiez: > > > Some special distribution packages (currently only pkgs/sage-sws2rst, pkgs/sage-conf*) do have their own source files, so tool configuration could go into pkgs/sage-sws2rst/pyproject.toml etc.
> > 
> > 
> > That doesn't scale.
> 
> It does not have to scale. As I just explained to you, there is a fixed short list of packages.

Copying the config 4 times is already too much for my taste. Thanks for the suggestion, but I'll not implement it.

mkoeppe: So do ruff.toml, as I explained.

mkoeppe:
I think I was quite clear about what the work item is.
Change it to ruff.toml because a pyproject.toml has no business to be in SAGE_ROOT.

tobiasdiez:
I've already thanked you for that suggestion, but I don't think its a good idea.

mkoeppe:
No, Tobias, you haven't. You were referring to the solution of putting tool configuration into distribution-package-specific pyproject.toml files in `src` and `pkgs/sage-conf*` and `pkgs/sage-sws2rst`.

tobiasdiez:
https://github.com/sagemath/sage/pull/36380#discussion_r1366018242

mkoeppe:
this link does not point to anywhere remotely related to this PR.

tobiasdiez:
No idea what happened there, but here is the correct link: https://github.com/sagemath/sage/pull/36503#discussion_r1367947146

mkoeppe:
It's not a suggestion, Tobias.

You are using pyproject.toml incorrectly, as I have explained, so it cannot be merged in this form. 
It would mark SAGE_ROOT as the source directory of a Python distribution package, which would be incorrect and would mislead developers into trying to do "pip install ." from SAGE_ROOT.

mkoeppe:
People have been using tox.ini as a convenient container because it was/is a convenient central place for tool configuration items when tox was already in use for testing the package.
Likewise, people now like to use pyproject.toml as a convenient container for tool configuration when pyproject.toml is already there ... in its function of defining a Python distribution package.

pyproject.toml
tobiasdiez: > So do ruff.toml, as I explained.

I've tried that but couldn't make it work reliable. For some files it just doesn't seem to work (although, I agree, that the docs claim that ruff.toml and pyproject.toml are equivalent). As a starter, you need to put the config in the root so that it applies to all python files in all directories. Moreover, it sometimes just didn't apply the config when reformatting a file - not sure if it's a problem with ruff or vscode, or something else. My guess is that the somewhat strange directory layout with a pyproject.toml in src is resulting in hick ups.

Since the config in the root pyproject.toml worked exellently for me, I stopped investigating. The added pyproject.toml in the root doesn't cause any other issues, is helpful in other contexts as well (eg #37446) and `pip install .` will work correctly with #36524.

Sorry, should have let you know before you invested the time to prepare https://github.com/sagemath/sage/pull/37452.

tobiasdiez:
@vbraun would be nice to get this in. Since it's only a bunch of config files updates, it should be safe to be marked as "blocker".

pyproject.toml
tobiasdiez: > > Not everything that has a pyproject file is a "distribution package".
> 
> That's the very idea of pyproject.toml. It marks the top of a Python distribution package (= project). That's what "py" and "project" refer to in the name of this file.

That's a very narrow viewpoint. Multiple tools support non-distribution/non-package scenarios via `pyproject.toml`, see eg https://python-poetry.org/docs/basic-usage#operating-modes. In https://discuss.python.org/t/projects-that-arent-meant-to-generate-a-wheel-and-pyproject-toml/29684/5 a Cpython core developer addressed the concern that pyproject.toml indicates that code which will be installed as a distribution as follows:
 > [...] storing configuration for black/ruff etc is not a reasonable indicator that code contained in that directory is intended to be installed as a distribution. The project section, yes, is currently only usable with distribution; but it’s certainly not an indicator without that. [...] The concern was (as I understand it) that pyproject.toml indicates that code which will be installed as a distribution. That isn’t the case and it’s not a packaging-exclusive marker file today; and it hasn’t been since its introduction.

[In the same thread](https://discuss.python.org/t/projects-that-arent-meant-to-generate-a-wheel-and-pyproject-toml/29684/193), another Cpython core developer answers the question "if a project (in the broadest possible sense of the term) is not intended to be built into a wheel, is pyproject.toml an appropriate place to store project-level configuration?" with
> I think so. With the level of adoption of [tool] I think this has already happened.

According to this viewpoint, since I don't add a `project` section, the `pyproject.toml` in the root does not function as a indicator for a distribution package and it's perfectly fine to add tool configs there.

jhpalmieri:
Creating a top-level file pyproject.toml just to support an editor is a really bad idea.

culler:
The pyproject.toml file definitely does not belong in SAGE_ROOT.  But it is true, and somewhat confusing, that the directory structure used in sage is non-standard.  I think it would be more standard with the following changes:

1. rename sage/src as sage/project_root
2. create sage/project_root/src
3. move sage/src/sage to sage/project_root/src/sage

Then the setup.py and setup.cfg would be in the project root directory, and the sage source code would be in the src/sage subdirectory of the project root.  I think that would be closer to what is suggested in [the packaging guide](https://packaging.python.org/en/latest)

Because pyproject.toml definitely does not belong in SAGE_ROOT I vote -1 on this PR.

saraedum:
:-1: nothing terribly wrong with this PR I think but https://github.com/sagemath/sage/pull/37452 seems like a cleaner solution to me.

mkoeppe:
This should be closed as a duplicate.

Issue #30813, opened 2020-10-22
try to upgrade tdlib


dimpase:
<div id="comment:0"></div>

tdlib appears in development, at https://github.com/freetdi/tdlib - now at version 0.9.

Not sure whether it it easy to upgrade, but one should try.

CC:  @dcoudert @sagetrac-llarisch

Component: **packages: optional**

_Issue created by migration from https://trac.sagemath.org/ticket/30813_

mkoeppe:
- Latest release = 0.9.1 on gitlab
- Latest commit = Sep 11, 2023
- Arch package https://archlinux.org/packages/extra/any/tdlib/ advertises itself as 0.9.2 @antonio-rojas 
- repo https://github.com/freetdi/tdlib/commits/develop/ is synced but has no releases, no tags

mkoeppe:
- Sage patch by @antonio-rojas https://gitlab.archlinux.org/archlinux/packaging/packages/sagemath/-/blob/main/sagemath-tdlib-0.9.patch?ref_type=heads

felix-salfelder:
There's some unfinished work on new python bindings for treedec. The goal is to replace the "tdlib" cython code by ordinary c++, with a more pythonic interface. And there is an experimental but stand-alone setup.py.

See the python_build branch in the git source repo and https://www.algok.uni-bamberg.de/freetdi.html for possible future work.

mkoeppe:
@felix-salfelder Thanks for the pointer. Note we are packaging https://pypi.org/project/sagemath-tdlib/ (you may have already seen that)

antonio-rojas:
> * Sage patch by @antonio-rojas https://gitlab.archlinux.org/archlinux/packaging/packages/sagemath/-/blob/main/sagemath-tdlib-0.9.patch?ref_type=heads

Patch credit goes to Fedora, TBF.

antonio-rojas:
@felix-salfelder hi, is there a reason why there is no 0.9.2 tag? Last commit being "Release 0.9.2" suggests that it's done.

felix-salfelder:
On Fri, Jun 07, 2024 at 12:13:01AM -0700, Antonio Rojas wrote:
> @felix-salfelder hi, is there a reason why there is no 0.9.2 tag? Last
> commit being "Release 0.9.2" suggests that it's done.

Not sure where it went. Perhaps I forgot to push it...  I have just
updated git tags and tarballs [1]. The migration from git**b remains in
limbo for now.

[...]

PR #36504, opened 2023-10-22
Functions for nice tree decomposition and its labelling

guojing0:
This PR intends to add two new functions:

1. `make_nice_tree_decomposition` takes a graph and its tree decomposition. It returns a *nice* tree decomposition.
2. `label_nice_tree_decomposition` takes a tree decomposition, assumed to be nice. It returns the same result, with vertices labelled accordingly.

### :memo: Checklist


dcoudert:
If you need, I have a code that takes a tree decomposition and returns a nice tree decomposition. It is not optimized but it can be useful. I had in my todo list to add it to Sagemath, as well as other methods...

guojing0:
> If you need, I have a code that takes a tree decomposition and returns a nice tree decomposition. It is not optimized but it can be useful. I had in my todo list to add it to Sagemath, as well as other methods...

Merci pour la suggestion. But at least for the nice tree decomposition one, I would like to implement it myself first.

What other methods do you have in mind? The ones mentioned in TODO in the code files?

Also, by "optimization", do you mean math/algo level or code level (like making the tree more balanced?)

guojing0:
> approximation of tree length based on BFS layering. I have a code for computing the "leveled tree" or "layered tree" as proposed by Chepoi et al. I need to add it and extend it to get a tree decomposition. The same can be done using lex-M and MCS-M.

Sounds interesting.

Currently I don't know too much about these, but can check the literatures.

guojing0:
> If you need, I have a code that takes a tree decomposition and returns a nice tree decomposition. It is not optimized but it can be useful. I had in my todo list to add it to Sagemath, as well as other methods...

May I see the code for reference, or improve upon it...?

dcoudert:
My code is below. I hope it is correct

```py
def nice_tree_decomposition(T):
    r"""
    Return a *nice* tree-decomposition of the tree-decomposition `T`.

    https://kam.mff.cuni.cz/~ashutosh/lectures/lec06.pdf

    A *nice* tree-decomposition is a rooted binary tree with four types of
    nodes:
    - Leaf nodes have no children and bag size 1;
    - Introduce nodes have one child. The child has the same vertices as the
      parent with one deleted;
    - Forget nodes have one child. The child has the same vertices as the parent
      with one added.
    - Join nodes have two children, both identical to the parent.

    .. WARNING::

        This method assumes that the vertices of the input tree `T` are hashable
        and have attribute ``issuperset``, e.g., ``frozenset`` or
        :class:`~sage.sets.set.Set_object_enumerated_with_category`.

    INPUT:

    - ``T`` -- a tree-decomposition
    """
    from sage.graphs.graph import Graph
    if not isinstance(T, Graph):
        raise ValueError("T must be a tree-decomposition")

    name = "Nice tree-decomposition of {}".format(T.name())
    if not T:
        return Graph(name=name)

    # P1: The tree is rooted.
    # We choose a root and orient the edges in root to leaves direction.
    leaves = [u for u in T if T.degree(u) == 1]
    root = leaves.pop()
    from sage.graphs.digraph import DiGraph
    DT = DiGraph(T.breadth_first_search(start=root, edges=True),
                 format='list_of_edges')

    # We relabel the graph in range 0..|T|-1
    bag_to_int = DT.relabel(inplace=True, return_map=True)
    # get the new name of the root
    root = bag_to_int[root]
    # and force bags to be of type Set to simplify the code
    bag = {ui: Set(u) for u, ui in bag_to_int.items()}

    # P2: The root and the leaves have empty bags.
    # To do so, we add to each leaf node l of DT a children with empty bag.
    # We also add a new root with empty bag.
    root, old_root = DT.add_vertex(), root
    DT.add_edge(root, old_root)
    bag[root] = Set()
    for vi, u in enumerate(leaves, start=root + 1):
        DT.add_edge(bag_to_int[u], vi)
        bag[vi] = Set()

    # P3: Ensure that each vertex of DT has at most 2 children.
    # If b has k > 2 children (c1, c2, ..., c_k). We disconnect (c1, ... ck-1)
    # from b, introduce k - 2 new vertices (b1, b2, .., bk-2), make ci the
    # children of bi for 1 <= i <= k-2, make ck-1 the second children of bk-2,
    # make bi the second children of b_i-1, and finally make b1 the second
    # children of b. Each vertex bi has the same bag has b.
    for ui in list(DT):  # the list(..) is needed since we modify DT
        if DT.out_degree(ui) > 2:
            children = DT.neighbors_out(ui)
            children.pop()  # one vertex remains a child of ui
            DT.delete_edges((ui, vi) for v in children)
            new_vertices = [DT.add_vertex() for _ in range(len(children) - 1)]
            DT.add_edge(ui, new_vertices[0])
            DT.add_path(new_vertices)
            DT.add_edges(zip(new_vertices, children))
            DT.add_edge(new_vertices[-1], children[-1])
            bag.update((vi, bag[ui]) for vi in new_vertices)

    # P4: If b has 2 children c1 and c2, then bag[b] == bag[c1] == bag[c2]
    for ui in list(DT):
        if DT.out_degree(ui) < 2:
            continue
        for vi in DT.neighbor_out_iterator(ui):
            if bag[ui] != bag[vi]:
                DT.delete_edge(ui, vi)
                wi = DT.add_vertex()
                DT.add_path([ui, wi, vi])
                bag[wi] = bag[ui]

    # P5: if b has a single child c, then one of the following conditions holds:
    #       (i)  bag[c] is a subset of bag[b] and |bag[b]| == |bag[c]| + 1
    #       (ii) bag[b] is a subset of bag[c] and |bag[c]| == |bag[b]| + 1

    def add_path_of_introduce(ui, vi):
        """
        Replace arc (ui, vi) by a path of introduce vertices.
        """
        if len(bag[ui]) + 1 == len(bag[vi]):
            return
        diff = list(bag[vi] - bag[ui])
        diff.pop()  # when all vertices are added, we are on vi
        xi = ui
        for w in diff:
            wi = DT.add_vertex()
            bag[wi] = bag[xi].union(Set((w,)))
            DT.add_edge(xi, wi)
            xi = wi
        DT.add_edge(xi, vi)
        DT.delete_edge(ui, vi)

    def add_path_of_forget(ui, vi):
        """
        Replace arc (ui, vi) by a path of forget vertices.
        """
        if len(bag[vi]) + 1 == len(bag[ui]):
            return
        diff = list(bag[ui] - bag[vi])
        diff.pop()  # when all vertices are removed, we are on vi
        xi = ui
        for w in diff:
            wi = DT.add_vertex()
            bag[wi] = bag[xi] - Set((w,))
            DT.add_edge(xi, wi)
            xi = wi
        DT.add_edge(xi, vi)
        DT.delete_edge(ui, vi)

    for ui in list(DT):
        if DT.out_degree(ui) != 1:
            continue
        
        vi = next(DT.neighbor_out_iterator(ui))
        bag_ui, bag_vi = bag[ui], bag[vi]

        if bag_ui == bag_vi:
            # We can merge the vertices
            if DT.in_degree(ui) == 1:
                parent = next(DT.neighbor_in_iterator(ui))
                DT.add_edge(parent, vi)
            else:
                root = vi
            DT.delete_vertex(ui)

        elif bag_ui.issubset(bag_vi):
            add_path_of_introduce(ui, vi)

        elif bag_vi.issubset(bag_ui):
            add_path_of_forget(ui, vi)

        else:
            # We must first forget some nodes and then introduce new nodes
            wi = DT.add_vertex()
            bag[wi] = bag[ui] & bag[vi]
            DT.add_path([ui, wi, vi])
            DT.delete_edge(ui, vi)
            add_path_of_forget(ui, wi)
            add_path_of_introduce(wi, vi)

    # We now return the result
    nice = Graph(DT, name=name)
    nice.relabel(inplace=True,
                 perm={u: (i, bag[u]) for i, u in enumerate(nice.breadth_first_search(start=root))})
    return nice
```

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +758,186 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+    TESTS::
+
+    TODO
+
+    EXAMPLES::
+
+    TODO
+    """
+    from sage.graphs.graph import Graph
+    if not isinstance(T, Graph):
```
dcoudert: This test is not enough to ensure that T is a valid tree decomposition. May be the method should take both a graph and a tree decomposition as input and first call method `is_valid_tree_decomposition` ?

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +758,186 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+        raise ValueError("Input must be a valid tree decomposition")
```
dcoudert: the error message should not start with a capital letter.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +758,186 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+    # To each leaf node of `directed_tree`, we add a child with empty bag.
+    # We also add a new root with empty bag.
+    root, old_root = directed_tree.add_vertex(), root
+    directed_tree.add_edge(root, old_root)
+    bag[root] = Set()
+    for vi, u in enumerate(leaves, start=root + 1):
+        directed_tree.add_edge(bags_to_int[u], vi)
+        bag[vi] = Set()
+
+    # P3: Ensure that each vertex of directed_tree has at most 2 children.
```
dcoudert: P3 -> Step 3

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +758,186 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
     G = reduced_tree_decomposition(G)
 
     G.name("Tree decomposition")
     return G
 
+def make_nice_tree_decomposition(T):
+    r"""
+    Return a *nice* tree decomposition (TD) of the TD `T`.
+
+    A *nice* TD `NT` is a rooted tree with four types of nodes:
```
dcoudert, line 806: don't forget to add a reference to a paper defining a nice tree decomposition.

The reference should be added to file `src/doc/en/reference/references/index.rst`.

dcoudert (changes requested):

guojing0:
@dcoudert Thank you for the reviews and suggestions. I didn't think you would review and comment until I add the label `needs_review`. I was working on CoCalc web-based VS code, so I pushed the codes for back-up.

dcoudert:
You should add examples for which we can check that the solution is what we expect.

guojing0:
> You should add examples for which we can check that the solution is what we expect.

Do intend to.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +759,226 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+    See page 161 of [CFKLMPPS15]_ for a description of the nice tree decomposition.
+
+    A *nice* TD `NT` is a rooted tree with four types of nodes:
+
+    - *Leaf* nodes have no children and bag size 1;
+    - *Introduce* nodes have one child: If `v \in NT` is an introduce node and
+      `w \in NT` its child, then `Bag(v) = Bag(w) \cup { x }`, where `x` is the
+      introduced node;
+    - *Forget* nodes have one child: If `v \in NT` is a forget node and
+      `w \in NT` its child, then `Bag(v) = Bag(w) \setminus { x }`, where `x` is the
```
dcoudert: same here

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +759,226 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
-    # We merge all edges between two sets S,S' where S is a subset of S'
+    # We merge all edges between two sets S, S' where S is a subset of S'
     G = reduced_tree_decomposition(G)
 
     G.name("Tree decomposition")
     return G
 
+def make_nice_tree_decomposition(graph, tree_decomp):
+    r"""
+    Return a *nice* tree decomposition (TD) of the TD `tree_decomp`.
```
dcoudert: You can not use latex for tree_decomp. Change  to ` ``tree_decomp`` `.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -768,12 +759,226 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+
+    See page 161 of [CFKLMPPS15]_ for a description of the nice tree decomposition.
+
+    A *nice* TD `NT` is a rooted tree with four types of nodes:
+
+    - *Leaf* nodes have no children and bag size 1;
+    - *Introduce* nodes have one child: If `v \in NT` is an introduce node and
+      `w \in NT` its child, then `Bag(v) = Bag(w) \cup { x }`, where `x` is the
```
dcoudert: use `\{ x \}`. Otherwise the `{` and `}` will not appear in the documentation.

dcoudert (changes requested):
May be you can add parameter `nice` to method `treewidth` in order to return a nice tree decomposition ?

guojing0:
> May be you can add parameter `nice` to method `treewidth` in order to return a nice tree decomposition ?

I agree with your suggestion. Just added. For some reason `treewidth` doesn't capture `nice` even if it's set to `True`. Will dig into the issue later.

dcoudert:
> For some reason `treewidth` doesn't capture `nice` even if it's set to `True`. Will dig into the issue later.

This is certainly due to the use of clique separators. Make sure to transform the result before returning the result.

guojing0:
> > For some reason `treewidth` doesn't capture `nice` even if it's set to `True`. Will dig into the issue later.
> 
> This is certainly due to the use of clique separators. Make sure to transform the result before returning the result.

I have uploaded the latest code. I don't think this is the case. The following was run with the latest code. As you can see on line 786, I have `print(nice)`, which was set to `True`, but it printed `False`.

```
sage: graphs.PetersenGraph().treewidth(certificate=True, nice=True)
False
Tree decomposition: Graph on 6 vertices
```

guojing0:
You are right; I thought it was caused by line 693.

guojing0:
```
sage: import sage.graphs.graph_decompositions.tdlib as tdlib
....: 
ModuleNotFoundError                       Traceback (most recent call last)
Cell In [2], line 1
----> 1 import sage.graphs.graph_decompositions.tdlib as tdlib

ModuleNotFoundError: No module named 'sage.graphs.graph_decompositions.tdlib'
```

I am thinking about testing the changes with `tdlib`. But after I read relevant documentations, it seems that `tdlib` is only bundled with a distribution called `sagemath-tdlib`.

So I was wondering if you have any suggestions? Also, one of the TO-DO items is `upgrade tdlib to 0.9.0 :trac: 30813` -- are there any significant changes, such as performance?

dcoudert:
> I am thinking about testing the changes with `tdlib`. But after I read relevant documentations, it seems that `tdlib` is only bundled with a distribution called `sagemath-tdlib`.

Depending on the installation of sagemath you have, you should be able to install `tdlib` using `./sage -I tdlib`. It works for me.

You can of course add some tests/examples in `treewidth` to show the behavior of parameter `nice` and that you get the same result when using algorithm `"tdlib"` or `"sage"`.

> So I was wondering if you have any suggestions? Also, one of the TO-DO items is `upgrade tdlib to 0.9.0 :trac: 30813` -- are there any significant changes, such as performance?
See #30813.
It is faster and it adds new methods that could be interesting. Authors won the 2017 PACE challenge with it. https://pacechallenge.org.

guojing0:
Will try.

I always thought that Java one was the only 2017 PACE winner.

guojing0:
> Depending on the installation of sagemath you have, you should be able to install tdlib using ./sage -I tdlib. It works for me.
>
> You can of course add some tests/examples in treewidth to show the behavior of parameter nice and that you get the same result when using algorithm "tdlib" or "sage".

Good suggestion. Just did.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -499,14 +499,26 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None, nice=Fals
...
     The PetersenGraph has treewidth 4::
 
-        sage: graphs.PetersenGraph().treewidth()
+        sage: petersen = graphs.PetersenGraph()
+        sage: petersen.treewidth(algorithm='sage')
+        4
+        sage: petersen.treewidth(algorithm='sage', certificate=True)
+        Tree decomposition: Graph on 6 vertices
+
+    The PetersenGraph has treewidth 4 (with `tdlib`)::
```
dcoudert: use double `

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -499,14 +499,26 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None, nice=Fals
...
+
+        sage: petersen = graphs.PetersenGraph()
+        sage: petersen.treewidth(algorithm='tdlib')
```
dcoudert: you will have to mark the test as optional in case it is tested by someone without tdlib installed.
It should be something like
```
sage: petersen.treewidth(algorithm='tdlib')  # optional tdlib
```
@mkoeppe can certainly tell us the right formulation.

dcoudert:
@mkoeppe  can you check the last commit. I'm not sure of the right formulation for doctests with optional dependency. Thanks.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -493,19 +500,36 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+        4                                                             # optional tdlib
+        sage: petersen.treewidth(algorithm='tdlib', certificate=True) # optional tdlib
+        Tree decomposition: Graph on 6 vertices                       # optional tdlib
+
+    Nice tree decomposition of the PetersenGraph has 28 nodes::
+
+        sage: petersen = graphs.PetersenGraph()
+        sage: petersen.treewidth(algorithm='sage', certificate=True, nice=True)
+        Nice tree decomposition of Tree decomposition: Graph on 28 vertices
+        sage: petersen.treewidth(algorithm='tdlib', certificate=True, nice=True) # optional tdlib
```
dcoudert: it seems that the right form is `# optional - tdlib`. And please make sure we have at least 2 spaces before `#`

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -761,18 +781,275 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+
+    INPUT:
+
+    - ``nice_TD`` -- a nice tree decomposition
+
+    OUTPUT:
+
+    A nice tree decomposition with nodes labelled.
+    """
+    directed_TD = DiGraph(nice_TD.breadth_first_search(start=root,
```
dcoudert: `root` is not defined.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -761,18 +781,275 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+            if len(current_bag) == len(child_bag) + 1:
+                directed_TD.set_vertex(node, 'intro')
+            else:
+                directed_TD.set_vertex(node, 'forget')
+        else:
+            directed_TD.set_vertex(node, 'leaf')
 
-    G.name("Tree decomposition")
-    return G
+    return Graph(directed_TD, name=name)
```
dcoudert: `name` is not defined.

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -761,18 +781,275 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+    # Return the nice tree decomposition after the processing
+    nice_tree_decomp = Graph(directed_tree, name=name)
+
+    bfs_ordering = nice_tree_decomp.breadth_first_search(start=root)
+    relabeling = {u: (i, bag[u]) for i, u in enumerate(bfs_ordering)}
+    nice_tree_decomp.relabel(inplace=True, perm=relabeling)
+
+    return nice_tree_decomp
+
+def label_nice_tree_decomposition(nice_TD):
```
dcoudert: If this is useful, why not doing it during the construction of the nice tree decomposition ? You could add an extra parameter to the method to do it, or do it my default. The extra cost is rather limited.

dcoudert (changes requested):

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
guojing0: I also thought about doing it inside `make_nice_tree_decomposition` function directly, given the relatively small overhead. But I'm thinking about taking a modular approach, and don't want to make the nice tree decomp function too complicated (since this function or `treewidth` already have many arguments). So I am slightly unsure which approach to take.

dcoudert:
The tag `# optional - tdlib` is needed only for the call of a doctest, not for the result.
```diff
-        sage: petersen = graphs.PetersenGraph()                        # optional - tdlib
+        sage: petersen = graphs.PetersenGraph()
        sage: petersen.treewidth(algorithm='tdlib')                    # optional - tdlib
-        4                                                              # optional - tdlib
+        4
        sage: petersen.treewidth(algorithm='tdlib', certificate=True)  # optional - tdlib
        Tree decomposition: Graph on 6 vertices                        # optional - tdlib
    Nice tree decomposition of the PetersenGraph has 28 nodes::
        sage: petersen = graphs.PetersenGraph()
        sage: petersen.treewidth(algorithm='sage', certificate=True, nice=True)
        Nice tree decomposition of Tree decomposition: Graph on 28 vertices
        sage: petersen.treewidth(algorithm='tdlib', certificate=True, nice=True)  # optional - tdlib
-        Nice tree decomposition of Tree decomposition: Graph on 28 vertices       # optional - tdlib
+        Nice tree decomposition of Tree decomposition: Graph on 28 vertices
```

Concerning `label_nice_tree_decomposition`, I see 2 options:
- let it as a separate method that a user can call independently if needed. However, don't make it directly accessible from Graph. It is too specialized
- set the labels during construction of the nice tree decomposition without asking if the user wants it or not. This is certainly the easiest way and you don't need extra parameters. The extra cost is small.

guojing0:
> Concerning `label_nice_tree_decomposition`, I see 2 options:
> 
>     * let it as a separate method that a user can call independently if needed. However, don't make it directly accessible from Graph. It is too specialized
> 
>     * set the labels during construction of the nice tree decomposition without asking if the user wants it or not. This is certainly the easiest way and you don't need extra parameters. The extra cost is small.

I think I will go with the first approach. It's in `tree_decomposition.pyx` so not in `graph.py`, or I misunderstand what you mean?

src/sage/graphs/graph_decompositions/tree_decomposition.pyx
```diff
@@ -761,18 +781,305 @@ def treewidth(g, k=None, kmin=None, certificate=False, algorithm=None):
...
+
+    EXAMPLES::
+
+        sage: from sage.graphs.graph_decompositions.tree_decomposition import make_nice_tree_decomposition, label_nice_tree_decomposition
+        sage: bip_one_four = graphs.CompleteBipartiteGraph(1, 4)
+        sage: bip_one_four_TD = bip_one_four.treewidth(certificate=True)
+        sage: nice_TD = make_nice_tree_decomposition(bip_one_four, bip_one_four_TD)
+        sage: root = sorted(nice_TD.vertices())[0]
+        sage: label_TD = label_nice_tree_decomposition(nice_TD, root)
+        sage: for node in label_TD:
```
dcoudert: Here I suggest to use `for node in sorted(label_TD):`.

It will not change the order of the nodes in this example, but it's safer. Indeed, we can not rely on the order of the vertices without sorting. The data structure of the graph may change and so the order of the vertices.

dcoudert (changes requested):

dcoudert (approved):
It now looks good to me.
Thank you.

guojing0:
@dcoudert I am not sure if this is the right place to ask, if not, I will either email or resort to other channel(s): As continuation of this PR, I am working on some functionality related to homomorphism counting. I am aware that Sage already has `has_homomorphism_to` function, but it only returns one homomorphism even if multiple exist.

It seems that this function was/is implemented with some LP solver. Would it be possible to extend it to support returning all homomorphisms?

dcoudert:
> @dcoudert I am not sure if this is the right place to ask, if not, I will either email or resort to other channel(s): As continuation of this PR, I am working on some functionality related to homomorphism counting. I am aware that Sage already has `has_homomorphism_to` function, but it only returns one homomorphism even if multiple exist.
> 
> It seems that this function was/is implemented with some LP solver. Would it be possible to extend it to support returning all homomorphisms?

It's better to open a specific issue to discuss that. It is certainly possible.

dcoudert:
Can you try to clean this PR. it has conflicts.

antonio-rojas:
This is causing a test failure on Arch:
```
File "/usr/lib/python3.11/site-packages/sage/graphs/graph_decompositions/tree_decomposition.pyx", line 1037, in sage.graphs.graph_decompositions.tree_decomposition.label_nice_tree_decomposition
Failed example:
    for node in sorted(label_TD):
        print(node, label_TD.get_vertex(node))
Expected:
    (0, {}) root
    (1, {0}) forget
    (2, {0, 1}) intro
    (3, {0}) forget
    (4, {0, 4}) join
    (5, {0, 4}) intro
    (6, {0, 4}) intro
    (7, {0}) forget
    (8, {0}) forget
    (9, {0, 3}) intro
    (10, {0, 2}) intro
    (11, {3}) intro
    (12, {2}) intro
    (13, {}) leaf
    (14, {}) leaf
Got:
    (0, {}) root
    (1, {0}) forget
    (2, {0, 4}) intro
    (3, {0}) forget
    (4, {0, 1}) join
    (5, {0, 1}) intro
    (6, {0, 1}) intro
    (7, {0}) forget
    (8, {0}) forget
    (9, {0, 2}) intro
    (10, {0, 3}) intro
    (11, {2}) intro
    (12, {3}) intro
    (13, {}) leaf
    (14, {}) leaf
```

guojing0:
@antonio-rojas There's indeed a bug in this code, I have fixed it in https://github.com/sagemath/sage/pull/36846. Does the new PR fix your issue?

antonio-rojas:
> @antonio-rojas There's indeed a bug in this code, I have fixed it in #36846. Does the new PR fix your issue?

No, I'm getting the exact same failure with #36846

dcoudert:
@guojing0, can you modify the test in #36846 to make it more robust (may be using a smaller graph). We can always decide to add it tag `# random` or `# not tested`..
Take also the opportunity to double check that labels are correctly set.

guojing0:
@dcoudert Done in https://github.com/sagemath/sage/pull/36846, also use the claw instead of $K_{1, 4}$.

PR #36510, opened 2023-10-23
Add pyright ci annotations

tobiasdiez:
We use the [pyright-action](https://github.com/jakebailey/pyright-action) to install and run pyright in the ci. This action is also used by https://github.com/python/typeshed (among others) and has the advantage that we can easily add annotations to surface typing errors. Due to the limited typing info currently in the sage repo, we only add these annotations for errors and not warnings.

image

Also fix one import that was complained about by the new pyright version.

### :memo: Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,21 +103,26 @@ jobs:
           MAKE: make -j2 --output-sync=recurse
           SAGE_NUM_THREADS: 2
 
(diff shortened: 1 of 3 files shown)


mkoeppe (approved):
This seems to work well.

tobiasdiez:
Thanks for the review!

PR #36515, opened 2023-10-23
`sage -tox -e pyright`: Update, speed up, isolate

mkoeppe:
Part split out from #36443 that does not conflict with #36510:
- Update to today's latest pyright version
- node.js, as provisioned via https://pypi.org/project/pyright/, no longer accepts a preinstalled node version.
- node.js is configured to provide sufficient heap size. This reduces the runtime dramatically as pyright no longer has to fight with the garbage collector and eliminates spurious out of memory conditions. (Same change for the CI "Build & Test" is made also in #36510.)
- Like the pyright invocation by the CI "Build & Test", it now runs on the entire source tree.
- It no longer assumes that the Sage library has been built (and no longer uses the Sage venv). It installs the runtime dependencies from the SAGE_SPKG_WHEELS in an isolated tox environment.

### :memo: Checklist

### :hourglass: Dependencies

- #36510 (merged here because it fixes one type error)


pyrightconfig.json
```diff
@@ -22,4 +22,5 @@
     "reportUndefinedVariable": "warning",
     "reportOptionalOperand": "warning",
     "reportMissingImports": "warning",
+    "reportSelfClsParameterName": "none",
```
tobiasdiez, line 20: What's the problem with this check? (I know you had a comment here before explaining it, but even then I didn't understand it)

mkoeppe: Sorry, I lost the commit with the comment, I'll find it.

mkoeppe: The comment is back and does not need further clarification

kwankyu:
Sorry, but the latest version is 1.1.336...

PR #36400, opened 2023-10-04
`build/pkgs/setuptools_scm`: Update to 8.0.4, add fixes for version 8

mkoeppe:
Version 8 needs some new workarounds when packages that do not use `setuptools_scm` are installed with `--no-build-isolation` in an environment with `setuptools_scm`. This is relevant in particular for conda, see #35593.

`setuptools_scm` inserts itself into the build process of unsuspecting packages via entry points declared by setuptools. One gets a warning:
-   `[sage_conf-10.2.beta5]     WARNING setuptools_scm._integration.setuptools pyproject.toml does not contain a tool.setuptools_scm section` (https://github.com/sagemath/sage/actions/runs/6409373032/job/17400573034?pr=36400#step:8:3978)

These warnings show up at build time, but also in runtime uses of Cython, where it causes doctest failures.

Here we do the update (for no specific reason other than to be up to date) and implement the workarounds.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #35095, opened 2023-02-12
Modularization of sagelib: Distributions sagemath-{brial,combinat,eclib,flint,gap,giac,glpk,graphs,groups,homfly,lcalc,libbraiding,libecm,linbox,modules,mpmath,ntl,pari,plot,polyhedra,schemes,singular,standard-no-symbolics,symbolics}

mkoeppe:
### 📚 Description
Screenshot 2023-09-09 at 11 57 02 PM

- Fixes #32432
- Fixes #32601
- Fixes #33674
- Fixes #33673
- Fixes #32718
- Fixes #33660 
- Fixes #31444 
- Fixes #32702
- Fixes #31695
- Fixes #30666
- Fixes #29912

### 📝 Checklist

- [x] I have made sure that the title is self-explanatory and the description concisely explains the PR.
- [x] I have linked an issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### ⌛ Dependencies

- Depends on https://github.com/sagemath/sage/issues/32779
- Depends on #35663
- Depends on #35661 
- Depends on https://github.com/sagemath/sage/pull/35774
- Depends on #35716
- Depends on #35717
- Depends on #35812
- Depends on #37012 
- Depends on #37009
- Depends on #37008
- Depends on #36964
- Depends on #36951 
- Depends on #36566 
- Depends on #37486 
- Depends on #36561
- Depends on #36649 
- Depends on #36676 
- Depends on #36498
- Depends on #36380 
- Depends on #35906 
- Depends on #37180

This is part of:
- #29705

PR #36520, opened 2023-10-23
`pkgs/sagemath-{bliss,sirocco,tdlib}`: Fix sdists

mkoeppe:
Since 10.2.beta8, likely as a side effect of the setuptools_scm update (#36400), some of our modularized sdists contain way too many files. https://pypi.org/project/sagemath-bliss/10.2b8/#files

We fix this by updating the MANIFEST.in files.
- Cherry-picked from #35095.

### :memo: Checklist

### DIFF
--- a/pkgs/sagemath-bliss/MANIFEST.in
+++ b/pkgs/sagemath-bliss/MANIFEST.in
@@ -1,6 +1,9 @@
+prune sage
+
 include VERSION.txt
 
 graft sage/graphs/bliss_cpp
+include sage/graphs/bliss.p*
 
 global-exclude *.c
 global-exclude *.cpp
--- a/pkgs/sagemath-sirocco/MANIFEST.in
+++ b/pkgs/sagemath-sirocco/MANIFEST.in
@@ -1,4 +1,8 @@
+prune sage
+
 include VERSION.txt
+
+include sage/libs/sirocco.p*
 
 global-exclude *.c
 global-exclude *.cpp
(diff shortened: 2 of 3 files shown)


kiwifb:
How bizarre. The sdist that I am generating now for sage-on-gentoo do not have all that extra baggage. I cannot see the difference before and after the patch. I had a long think after some of our previous conversation and came up with something, I am now generating sdist for the develop branch as part of the ebuild, I just had to put my brain cells at work for a bit and now at some point `python -m build -n -x -s` is run.

mkoeppe:
Our GH Actions workflow that builds them just uses `make pypi-sdists` in a freshly configured source tree, which ultimately invokes the same command that you quoted.  setuptools_scm is installed before any of the sdists are built because it is part of `PYTHON_TOOLCHAIN`.

kiwifb:
And I have setuptools-scm too, but I have stayed at 7.1.0 on that system. That may make a difference, I will test that.

kiwifb:
Yes, setuptools-scm-9.0.4 is the guilty party and causes a lot of stuff to be included. The patch does have an impact in that case. Much more reasonable.

kiwifb (approved):
Does what it says on the box. 
Positive review from me.

mkoeppe:
Thanks!

kiwifb:
I think we need a follow up. The sdist problem did not stop at optional packages. The sagemath-standard sdist contains `sage/libs/sirocco.pyx` which does clash with `sagemath-sirocco`. There is nothing from sagemath-meataxe or sagemath-bliss in the sdist (otherwise emerge would refuse to merge them as it did with sirocco), but I do not know about other sagemath-* optional packages.

kiwifb:
I should add that this is with the current merge branch of Volker which includes this PR.

PR #36538, opened 2023-10-25
Implement Drinfeld modular forms

DavidAyotte:
This PR aims to implement Drinfeld modular forms into SageMath. These special types of modular forms live in the function field theory and are linked with Drinfeld modules. 

#### Definitions

Without going too much into the details, we give some definitions and insight of what is going on here. Let $K = \mathbb{F}_q(T)$ together with the $1/T$-adic norm and $A = \mathbb{F}_q[T]$. We define $K\_{\infty} = \mathbb{F}_q((1/T))$ be the completion of $K$ and let $\mathbb{C}\_{\infty}$ be the completion of an algebraic closure of $K\_{\infty}$. In this specific setting, the *Drinfeld period domain* of rank $r$ is

$$
\Omega^r(\mathbb{C}\_{\infty}) := \lbrace (w_1, \ldots, w_{r-1}, 1)\in \mathbb{C}\_{\infty}^r : \text{the }w_i \text{'s are }K\_{\infty}\text{-linearly independant} \rbrace.
$$

(this period domain is to be seen as the analogue of the complex upper half plane)

So, for each $w = (w_1, \ldots, w_{r-1}, 1)\in \Omega^r(\mathbb{C}\_{\infty})$, we have an associated $A$-lattice (i.e. a discrete projective $A$-module of rank r): $\Lambda_w = A w_1 + \ldots A w_{r-1} + A$.

The analytic theory of Drinfeld modules tells us that, for each such $w$, there exist a Drinfeld module 

$$
\phi^w_T : T \mapsto T + g\_1(w) \tau + \ldots + g\_{r-1}(w) \tau^{r-1} + g\_r(w) \tau^r
$$

associated with this lattice. This association is what we call the universal Drinfeld module over $\Omega^r(\mathbb{C}\_{\infty})$. We see the coefficients as functions $g_i : \Omega^r(\mathbb{C}\_{\infty}) \rightarrow \mathbb{C}\_{\infty}$. These functions satisfy modular transformation property under the action of the group $\mathrm{GL}_r(A)$:

$$
g_i(\gamma(w)) = j(\gamma, w)^i g_i(w)
$$

where $j$ is some automorphic factor.

The ring of Drinfeld modular forms for $\mathrm{GL}_r(A)$ is

$$
\mathcal{M}\_{\bullet} = \mathbb{C}\_{\infty}[g_1, \ldots g_{r-1}, g_r].
$$

#### Implementations details

In this implementation, we create a new parent `DrinfeldModularForms` which represents the polynomial ring $K[g_1,\ldots, g_{r-1}, g_r]$ and a new element class `DrinfeldModularFormsElement`. Then, for arbitrary rank, we only see a Drinfeld modular form symbolically as a multivariate polynomial in $g_1, \ldots, g_r$, the coefficients forms. In the special case when the rank is two, we aim to implement algorithms for computing the expansion at infinity of any Drinfeld modular forms. These expansions are analogues of the usual $q$-expansion of classcial modular forms. This will be achieved via the Lopez-Petrov nonstandard expansion theory.

We note that this implementation was initially done as a pip-installable package:

https://davidayotte.github.io/drinfeld_modular_forms

This work has been an ongoing project for almost two years now and since the addition of Drinfeld modules in SageMath, the author has seen an interest in the implementation of Drinfeld modular forms to SageMath. In the rank two case, the ability to compute their expansion at infinity shows some interesting computational properties of these objects.

We are aware that this is a quite specialized topic, so we aim to make this implementation as convivial as possible for the user and well documented. Any comments and suggestions are welcome!

CC: @xcaruso @kryzar @ymusleh 

### :memo: Checklist


xcaruso:
Great addition, thanks David!

I noticed that you added special classes for Drinfeld modular forms of rank 2 but those are basically empty. Is it normal?

DavidAyotte:
> I noticed that you added special classes for Drinfeld modular forms of rank 2 but those are basically empty. Is it normal?

Yes these are currently just placeholder, but I plan to implement them in this PR. If you want to get familiar with the expansion theory I'm going to implement, I recommend reading chapter 7 of my thesis and/or the associated reference by Petrov and Lopez:

- Petrov: https://arxiv.org/abs/1207.6479
- Lopez: https://link.springer.com/article/10.1007/s00013-010-0148-7

I was recently thinking about some implementation details and wondering whether or not it would be desirable to implement a specific class for the $A$-expansion, but in the end, I don't think I will go this way because I'm not sure if this will really add something.

xcaruso:
I suggest actually to split your contribution in two separate PR. It would be easier for review.

DavidAyotte:
> I suggest actually to split your contribution in two separate PR. It would be easier for review.

OK, that's a good advice, I will do that. In this current PR I will only implement arbitrary ranks Drinfeld modular forms and in a seperate PR I will implement rank 2 specific stuff.

DavidAyotte:
I pretty much finished most of the major changes to this PR. I still have some enhancements in mind, but I think it's best to keep them for further PR. As discussed previously, I'm planning to implement the expansion in rank two casse. Next, I also would like to add a keyword argument `generator` to the `DrinfeldModularForms` class which would create the ring with respect to a different set of generators.

Essentially, there would be two options: `generator="coefficients"` or `generator="eisenstein"`. The `"coefficients"` flag would create the ring as it is implemented right now, and the `"eisenstein"` flag would return the ring, but with its generators assumed to be the Eisentein series. More precisely, I want to implement this behavior:
```
sage: A = GF(3)['T']; K.<T> = Frac(A)
sage: M = DrinfeldModularForms(K, 3, generators='eisenstein')
sage: M.gens()
# Return E1, E2, E3 (Ei is the Eisentein series of weight q^i - 1).
```
However, this would require a little bit of work and refactoring, especially for the coefficient forms methods, since those methods assume the generators to be the coefficient forms. In the Eisenstein case, the coefficient forms method should return an algebraic combination of the Eisenstein series, which is computable by a recursive procedure.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+        sage: g2 in M
+        True
+    """
+    def __init__(self, parent, polynomial):
+        if not isinstance(polynomial, MPolynomial):
+            raise TypeError("input must be a multivariate polynomial")
+        if not parent.base_ring().has_coerce_map_from(polynomial.base_ring()):
+            raise ValueError("unable to coerce base ring of the given "
+                             "polynomial into Drinfeld modular form ring")
+        poly = parent._poly_ring(polynomial)
```
kryzar, line 100: In this method, there is no explicit verification that the number of variables of the multivariate polynomial `polynomial` is correct. I assume this is done here. Can you confirm?

Also, wouldn't `mpolynomial` be a better name?

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+    sage: A = GF(q)['T']
+    sage: K.<T> = Frac(A)
+    sage: M = DrinfeldModularForms(K, 2)  # rank 2
+    sage: M.inject_variables()
+    Defining g1, g2
+    sage: g1.parent()
+    Ring of Drinfeld modular forms of rank 2 over Fraction Field of Univariate Polynomial Ring in T over Finite Field of size 3
+
+A *graded Drinfeld modular form* is a sum of modular forms having
+potentially different weights::
```
kryzar: See my main comment for this review.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+    def __init__(self, parent, polynomial):
+        if not isinstance(polynomial, MPolynomial):
+            raise TypeError("input must be a multivariate polynomial")
+        if not parent.base_ring().has_coerce_map_from(polynomial.base_ring()):
+            raise ValueError("unable to coerce base ring of the given "
+                             "polynomial into Drinfeld modular form ring")
+        poly = parent._poly_ring(polynomial)
+        self._polynomial = poly
+
+        ModuleElement.__init__(self, parent)
```
kryzar: Why not use `super()`?

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+        sage: M = DrinfeldModularForms(K, 2)
+        sage: M.inject_variables()
+        Defining g1, g2
+        sage: (T^2 + 1)*(g1 + g1*g2)
+        (T^2 + 1)*g1*g2 + (T^2 + 1)*g1
+        sage: (g1).parent()
+        Ring of Drinfeld modular forms of rank 2 over Fraction Field of Univariate Polynomial Ring in T over Finite Field of size 3
+        sage: g2 in M
+        True
+    """
```
kryzar: Maybe a few comments to assist the reader on the init process would be valuable. For example, you could say that this function is mainly formal and that most of the work is handled by the parent. Up to you :)

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: A = GF(3)['T']; K = Frac(A)
+            sage: M = DrinfeldModularForms(K, 2)
+            sage: M.0 + M.1  # indirect doctest
+            g2 + g1
+        """
+        return self.__class__(self.parent(), self._polynomial + other._polynomial)
+
+    def _mul_(self, other):
+        r"""
+        Return the multiplication of self with other.
```
kryzar:
```suggestion
        Return the multiplication of `self` with `other`.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: M.0*(M.0 + M.1)
+            g1*g2 + g1^2
+            sage: (M.0 + M.1)*M.0
+            g1*g2 + g1^2
+        """
+        return self.__class__(self.parent(), self._polynomial*other._polynomial)
+
+    def _lmul_(self, c):
+        r"""
+        Return the scalar multiplication of self by `c`.
```
kryzar:
```suggestion
        Return the scalar multiplication of `self` by `c`.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: 0 * M.1
+            0
+            sage: M.0 * 0
+            0
+        """
+        return self.__class__(self.parent(), c*self._polynomial)
+
+    def __neg__(self):
+        r"""
+        Return the negation of self.
```
kryzar:
```suggestion
        Return the negation of `self`.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: A = GF(3)['T']; K = Frac(A)
+            sage: M = DrinfeldModularForms(K, 2)
+            sage: -M.0  # indirect doctest
+            -g1
+        """
+        return self.__class__(self.parent(), -self._polynomial)
+
+    def __bool__(self):
+        r"""
+        Return True whether self is nonzero.
```
kryzar:
```suggestion
        Return `True` whether `self` is nonzero.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: A = GF(3)['T']; K = Frac(A)
+            sage: M = DrinfeldModularForms(K, 2)
+            sage: bool(M.0)
+            True
+        """
+        return bool(self._polynomial)
+
+    def _latex_(self):
+        r"""
+        Return the latex expression of self.
```
kryzar:
```suggestion
        Return the LaTeX expression of `self`.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: latex(g2)
+            g_{2}
+            sage: latex(1/T*g1^5 + g2*g1)
+            \frac{1}{T} g_{1}^{5} + g_{1} g_{2}
+        """
+        return self._polynomial._latex_()
+
+    def _richcmp_(self, other, op):
+        r"""
+        Return the comparison of self with other.
```
kryzar:
```suggestion
        Return the comparison of `self` with `other`.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: latex(g1)
+            g_{1}
+            sage: latex(g2)
+            g_{2}
+            sage: latex(1/T*g1^5 + g2*g1)
+            \frac{1}{T} g_{1}^{5} + g_{1} g_{2}
+        """
+        return self._polynomial._latex_()
+
+    def _richcmp_(self, other, op):
```
kryzar, line 214: Question: what is the benefit of using `_richcmp_` other something like `__eq__`?

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: M.0 == M.0
+            True
+        """
+        if op != op_EQ and op != op_NE:
+            raise TypeError('invalid comparison between modular forms ring elements')
+        return richcmp(self._polynomial, other._polynomial, op)
+
+    def rank(self):
+        r"""
+        Return the rank of the graded Drinfeld form.
```
kryzar: And you could say that the rank depends only on the parent.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            2
+            sage: M5 = DrinfeldModularForms(K, 5)
+            sage: (M5.0 + M5.3).rank()
+            5
+        """
+        return self.parent()._rank
+
+    def is_one(self):
+        r"""
+        Return ``True`` whether the given graded Drinfeld form is the
```
kryzar: Here and subsequently you use "given". Sometimes you don't. You might want to uniformize the notation. Up to you !)

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            True
+        """
+        return not bool(self)
+
+    def is_drinfeld_modular_form(self):
+        r"""
+        Return whether ``self`` is a Drinfeld modular form.
+
+        We recall that elements of Drinfeld modular forms ring are not
+        necessarily modular forms as they may have mixed weight components.
```
kryzar: See my main comment for this review.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            g2^4
+        """
+        M = self.parent()
+        components = self._polynomial.homogeneous_components().items()
+        return {k: self.__class__(M, p) for k, p in components}
+
+    def polynomial(self):
+        r"""
+        Return self as a multivariate polynomial over the generators of
+        the ring.
```
kryzar:
```suggestion
        Return `self` as a multivariate polynomial over the generators
        of the ring.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+            sage: F = g1*h2^9
+            sage: F.type_m()
+            9
+            sage: (h2^11).type_m()
+            1
+            sage: g1.type_m()
+            0
+        """
+        if not self.is_drinfeld_modular_form():
+            raise ValueError("self should be a Drinfeld modular form")
```
kryzar: You might wanna mention this restriction in the docstring.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+        if not self.parent()._has_type:
+            return ZZ(0)
+        q = self.base_ring().base_ring().cardinality()
+        return self.polynomial().degrees()[-1] % (q-1)
+
+    def weight(self):
+        r"""
+        Return the weight of self.
```
kryzar:
```suggestion
        Return the weight of `self`.
```

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,397 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#                  https://www.gnu.org/licenses/
+# ****************************************************************************
+
+from sage.structure.element import ModuleElement
+from sage.structure.richcmp import richcmp, op_NE, op_EQ
+
+from sage.rings.polynomial.multi_polynomial import MPolynomial
+
```
kryzar, line 27:
```suggestion


```

kryzar (commented):
Here is my review for the file `element.py`. I will review `ring.py` as well, but figured we could already start the discussion.

As a general remark, and even though you address that a bit in `ring.py`, I fear that there is a lack of clarity between what is considered a Drinfeld modular form and what is simply a graded Drinfeld modular form, both in the naming and in your documentation. The fact that those are all instances `DrinfeldModularFormsElement` certainly doesn't help, imo. I also find `is_drinfeld_modular_form` confusing.

kryzar:
Oh, and another comment, I think you should stress that those forms are formal objects, not actually computed objects (like a series expansion).

src/sage/modular/drinfeld_modform/element.py
DavidAyotte: The benefit is that this handles other operators like "<", "<=", ... (not only "==") and returns a more user-friendly error message.

src/sage/modular/drinfeld_modform/element.py
DavidAyotte: Thanks for spotting this!! I'll make the changes. Note that raw strings are denoted using pair of two backticks like so: ` ``self`` `. Unlike Github comments, pair of single backtick is to denote LaTeX notations.

DavidAyotte:
Thank you Antoine for taking the time to review! I addressed the backtick comments locally and I will address the other comments a bit later.

xcaruso:
By the way, in general, we do not write `"self"` but `this Drinfeld modular form`. But I think that both are acceptable.

xcaruso:
Another comment: I'm not sure it's a good idea to have the tutorial in a separate file because it becomes more difficult to access it in the command line using `?`.

src/sage/modular/drinfeld_modform/element.py
kryzar: TIL, thanks!

kryzar:
> Oh, and another comment, I think you should stress that those forms are formal objects, not actually computed objects (like a series expansion).

Understood. I shall also review the other file.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
+r"""
+Introduction to Drinfeld modular forms
+======================================
+
+This tutorial outlines the definitions, the notations, and the
+implementation of Drinfeld modular forms in SageMath. We assume that the
+reader has basic knowledge of classical modular forms, as we
+will often make analogies to this setting. We also assume little
+knowledge of Drinfeld modules. The interested reader can consult the
+Sage reference manual about this topic :ref:`Drinfeld modules`.
```
kryzar: Modify last sentence + reflow

```suggestion
This tutorial outlines the definitions, the notations, and the
implementation of Drinfeld modular forms in SageMath. We assume that the
reader has basic knowledge of classical modular forms, as we will often
make analogies to this setting. We also assume little knowledge of
Drinfeld modules; for this topic, the interested reader can consult the
SageMath reference manual: :ref:`Drinfeld modules`.
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+.. RUBRIC:: Preliminary notations
+
+Let `q` be a prime power and let `A` be the ring of functions of
+`\mathbb{P}^1/\mathbb{F}_q` which are regular outside a closed point
+`\infty`. This ring is the polynomial ring `\mathbb{F}_q[T]`. We denote
```
kryzar, line 17: Is it important here to define $A$ as this ring of functions rather than $A = \mathbb F_q [T]$?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+by `K := \mathbb{F}_q(T)` its fraction field. We endow `K` with the
+`1/T`-adic valuation and let `K_{\infty} := \mathbb{F}_q((1/T))` be the
+completion of `K`. Next, we define `\mathbb{C}_{\infty}` to
```
kryzar: For the whole construction of $\mathbb C_\infty$, I would suggest adding a note after this paragraph to present the parallel construction in the classical case. If you feel that would be overkill, by all means keep it like this.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+be the completion of an algebraic closure of `K_{\infty}`. Lastly, We
+denote by `\tau : x\mapsto x^q` the `q`-Frobenius.
```
kryzar: Reflow

```suggestion
completion of an algebraic closure of `K_{\infty}`. Lastly, We denote by
`\tau : x\mapsto x^q` the `q`-Frobenius.
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+In SageMath, we create the rational function field by first creating a
+univariate polynomial ring over `\mathbb{F}_q` and then constructing its
+field of fractions::
```
kryzar: Reformulation
```suggestion
univariate polynomial ring over `\mathbb{F}_q`, and following this, by
constructing its field of fractions::
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+======================================
+
+This tutorial outlines the definitions, the notations, and the
+implementation of Drinfeld modular forms in SageMath. We assume that the
+reader has basic knowledge of classical modular forms, as we
+will often make analogies to this setting. We also assume little
+knowledge of Drinfeld modules. The interested reader can consult the
+Sage reference manual about this topic :ref:`Drinfeld modules`.
+
+.. RUBRIC:: Preliminary notations
```
kryzar, line 13: I would suggest be more precise and titling something like "Base setting", as you are not just giving notations, but also instantiating objects.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+    \Omega^r(\mathbb{C}_{\infty}) :=
+    \mathbb{P}^{r-1}(\mathbb{C}_{\infty})
+    \setminus \{K_{\infty}\text{-rational hyperplanes}\}.
+
+This space is a rigid analytic space and we identify its elements
+with the set of column vectors
+`(w_1,\ldots, w_{r-1}, w_{r})^{\mathrm{T}}` in `\mathbb{C}_{\infty}^r`
+such that the `w_i` are `K_{\infty}`-linearly independant and
+`w_r = \xi`, a nonzero constant in `\mathbb{C}_{\infty}`.
```
kryzar: Precise that $\xi$ is fixed?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+field of fractions::
+
+    sage: A = GF(3)['T']
+    sage: K.<T> = Frac(A)
+    sage: K
+    Fraction Field of Univariate Polynomial Ring in T over Finite Field of size 3
+    sage: K.base()  # returns A
+    Univariate Polynomial Ring in T over Finite Field of size 3
+
+.. RUBRIC:: Drinfeld period domain
```
kryzar: What about something like "Drinfeld period domain and action of `\mathrm{GL}_r(K_{\infty})` on `\Omega^r(\mathbb{C}_{\infty})`"?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+`(w_1,\ldots, w_{r-1}, w_{r})^{\mathrm{T}}` in `\mathbb{C}_{\infty}^r`
+such that the `w_i` are `K_{\infty}`-linearly independant and
+`w_r = \xi`, a nonzero constant in `\mathbb{C}_{\infty}`.
+
+We define a left action of `\mathrm{GL}_r(K_{\infty})` on
+`\Omega^r(\mathbb{C}_{\infty})` by setting
+
+.. MATH::
+
+    \gamma(w) := j(\gamma, w)^{-1}\gamma w
```
kryzar, line 68: I assume the notation $\gamma(w)$ (as opposed to $\gamma \cdot w$) is standard.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+where `j(\gamma, w) := \xi^{-1} \cdot (\text{last entry of }\gamma w)`.
+
+.. RUBRIC:: Universal Drinfeld module over `\Omega^r(\mathbb{C}_{\infty})`
+
+For any `w = (w_1, \ldots, w_{r-1}, \xi)` in
+`\Omega^r(\mathbb{C}_{\infty})` we have a corresponding discrete
+`A`-module which is free of rank `r`:
```
kryzar:
```suggestion
`A`-module `\Lambda^w` which is free of rank `r`:
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+.. MATH::
+
+    \Lambda^w := Aw_1 \oplus \cdots \oplus Aw_{r-1} \oplus A\xi.
+
+By analytic uniformization, there exists a corresponding Drinfeld module
```
kryzar: Maybe add a reference to what analytic uniformization is, or explain briefly.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+.. MATH::
+
+    \phi^w : T \mapsto T + g_1(w)\tau + \cdots
+    + g_{r - 1}(w)\tau^{r-1} + g_{r}(w)\tau^{r}.
+
+This Drinfeld module is called the *universal
+Drinfeld `\mathbb{F}_q[T]`-module over `\Omega^r(\mathbb{C}_{\infty})`*
+and its coefficients
+`g_i : \Omega^r(\mathbb{C}_{\infty}) \rightarrow \mathbb{C}_{\infty}`
```
kryzar:
```suggestion
`g_i : \Omega^r(\mathbb{C}_{\infty}) \to \mathbb{C}_{\infty}`
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+are rigid analytic functions satisfying the invariance property:
```
kryzar: Italics
```suggestion
are rigid analytic functions satisfying the *invariance property*:
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+In the rank two case, the expansion at infinity is of the form
+
+.. MATH::
+
+    g_i(w) = \sum_{i = 0}^{\infty} a_n(g_i)u(w)^i
+
+where `u(w) := e(w)^{-1}` and `e` is the exponential of the Carlitz
+module `\rho:T\mapsto T + \tau`. The analytic parameter `u` is called
+the *parameter at infinity*.
```
kryzar: Added "function"
```suggestion
where `u(w) := e(w)^{-1}` and `e` is the exponential function of the
Carlitz module `\rho:T\mapsto T + \tau`. The analytic parameter `u` is
called the *parameter at infinity*.
```

Also maybe add a reference.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+A *Drinfeld modular form* of rank `r`, weight `k`, type `m` for
+`\mathrm{GL}_r(A)` is a rigid analytic function
+`f:\Omega^r(\mathbb{C}_{\infty}) \rightarrow \mathbb{C}_{\infty}`
```
kryzar: I would put this line in a `.. MATH::` block. Up to you though.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+* `f(\gamma(w)) = \mathrm{det}(\gamma)^m j(\gamma, w)^k f(w)` for all
+  `\gamma` in `\mathrm{GL}_r(A)` and
+  `w\in \Omega^r(\mathbb{C}_{\infty})`;
+
+* `f` is holomorphic at infinity.
+
+The second condition is similar to the classical case. In the rank two
+situation, this expansion is simply given by
+`f = \sum_{n\geq 0} a_n(f) u^n` where `a_n(f)\in \mathbb{C}_{\infty}`.
```
kryzar:
```suggestion
The second condition is similar to the classical case. Furthemore, in
the rank two situation, the expansion of `f` is simply given by
`f = \sum_{n\geq 0} a_n(f) u^n` where `a_n(f)\in \mathbb{C}_{\infty}`.
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+The reader is refered to
+part I of [BRP2018]_ for more information about the analytic theory of
+Drinfeld modular form of arbitrary rank.
```
kryzar: You don't really mention the analytic theory of Drinfeld modules before, so perhaps it is not obvious why the reader should be interested in that.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+.. RUBRIC:: Ring of Drinfeld modular forms
+
+Letting `M_k^{r, m}(\mathrm{GL}_r(A))` denote the space of rank `r`,
+weight `k\in (q - 1)\mathbb{Z}` and type `m~(\mathrm{mod}~q-1)`
+Drinfeld modular forms, we define
```
kryzar:
```suggestion
Drinfeld modular forms. We define
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+.. MATH::
+
+    M^{r, 0}(\mathrm{GL}_r(A)) :=
+    \bigoplus_{k\in ZZ} M_k^{r}(\mathrm{GL}_r(A))
+
+to be the graded ring of all Drinfeld modular forms of type 0. The
+graduation is given by the weight of a modular form. Similarly, we let
+`M^{r}(\mathrm{GL}_r(A)) \supset M^{r, 0}(\mathrm{GL}_r(A))` be the ring
+of all Drinfeld modular forms of arbitrary type. By
```
kryzar:
```suggestion
of all Drinfeld modular forms of rank `r` and arbitrary type. By
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+    M^{r, 0}(\mathrm{GL}_r(A))
+    = \mathbb{C}_{\infty}[g_1,\ldots, g_{r-1}, g_{r}].
+
+and
+
+.. MATH::
+
+    M^{r, 0}(\mathrm{GL}_r(A))
+    = \mathbb{C}_{\infty}[g_1,\ldots, g_{r-1}, h_{r}].
```
kryzar: Join the equalities?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+where `h_r` is a weight `(q^r - 1)/(q - 1)` modular forms of type `1`
+which is a `(q-1)`-root of `g_r`.
+
+To create the ring of type zero forms in SageMath we use the class
+`DrinfeldModularForms`::
```
kryzar:
```suggestion
To create the ring of type zero and rank `r` Drinfeld modular forms in
SageMath, one uses the class `DrinfeldModularForms`::
```

Also, shouldn't you write class:`.../DrinfeldModularForms`?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+
+    sage: A = GF(3)['T']
+    sage: K.<T> = Frac(A)
```
kryzar, line 206: Can't you reuse the objects of the first rubric?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+    sage: M = DrinfeldModularForms(K, 3)  # rank 3
+    sage: M
+    Ring of Drinfeld modular forms of rank 3 over Fraction Field of Univariate Polynomial Ring in T over Finite Field of size 3
+
+This ring is generated by the coefficient forms that we see purely as
+symbolic objects, each of which we attach its respective weight::
```
kryzar: > each of which we attach its respective weight

I'm not sure to understand your point. Do you just mean to say that each of those forms has its own weight? Maybe you should reformulate this whole sentence.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+    sage: h4.weight()
+    40
+
+.. RUBRIC:: Coefficients forms
+
+There are bound methods for accessing the coefficients forms at any
+`a`::
+
+    sage: A = GF(2)['T']
+    sage: K.<T> = Frac(A)
```
kryzar: Why change the base setting?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,228 @@
...
+    sage: M.coefficient_forms(T^2)
+    [(T^2 + T)*g1,
+     g1^3 + (T^4 + T)*g2,
+     g1^4*g2 + g1*g2^2 + (T^8 + T)*g3,
+     g1^8*g3 + g1*g3^2 + g2^5,
+     g2^8*g3 + g2*g3^4,
+     g3^9]
+
+To access a single coefficient form, use the method
+``coefficient_form``::
```
kryzar: More consistent + pointer to the method
```suggestion
To access a single coefficient form, one uses the method
meth:`coefficient_form`::
```

kryzar (changes requested):
I reviewed the tutorial and now only have `ring.py` remaining. I think it's very good, thank you!

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
+r"""
+This module defines a class named :class:`DrinfeldModularForms`.
+
+Currently, the implementation only supports the full group modular group
```
kryzar: You mean this?
```suggestion
Currently, the implementation only supports the full modular group
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+`\mathrm{GL}_r(A)` where `A = \mathbb{F}_q[T]`.
+
+The implementation is based on the following identification:
+
+.. MATH::
+
+    M^r(\mathrm{GL}_r(A))
+    = \mathbb{C}_{\infty}[g_1, \ldots, g_{r-1}, g_{r}].
+
+where `g_i` the `i`-th coefficient form of weight `q^{i} - 1`.
```
kryzar: Maybe add more details, or a reference. As you wish.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+
+EXAMPLES::
```
kryzar: Shouldn't the examples be in the class, to have them with `?`? Anyway, as Xavier suggested, you may have some refactoring to do on this aspect.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+     g2^91]
+
+One can compute basis for any subspace of given weight::
+
+    sage: M = DrinfeldModularForms(K, 4)
+    sage: M.basis_of_weight(q^3 - 1)
+    [g3, g1*g2^3, g1^5*g2^2, g1^9*g2, g1^13]
+
+We note that the elements of this ring may not be *modular forms* as
+as they may have mixed weight components::
```
kryzar: As mentioned in the review of `element.py`, I think the terminology is quite ambiguous and should be modified. However, I am not an expert by any mean, so please tell me if this is standard.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+
+    sage: M = DrinfeldModularForms(K, 4)
+    sage: M.inject_variables()
+    Defining g1, g2, g3, g4
+    sage: F = g1 + g2 + g3 + g4
+    sage: F.is_drinfeld_modular_form()
+    False
+
+This is why we call these elements *graded Drinfeld modular forms*.
```
kryzar: Not super clear imo. But again, you can just refer to another part of the documentation.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+    sage: M.inject_variables()
+    Defining g1, g2, g3, h4
+    sage: h4.weight()
+    400
+    sage: h4.type_m()
+    1
+    sage: (g1*h4^4).type_m()
+    4
+
+The last generator is known as Gekeler's `h` function.
```
kryzar: Add a reference?

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+
+AUTHORS:
+
+- David Ayotte (2022): initial version
+"""
+
+# ****************************************************************************
+#       Copyright (C) 2022 DAVID AYOTTE <davidayotte94@outlook.com>
```
kryzar: Quick remark, you might want to `git config` to set your email to this one.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+
+class DrinfeldModularForms(Parent, UniqueRepresentation):
+    r"""
+    Base class for the graded Drinfeld modular forms ring.
+
+    INPUT:
+
+    - ``base_ring`` -- The fraction field of a univariate polynomial
+      ring over `\mathbb{F}_q`.
+    - ``rank`` (integer, default: 2) -- the rank of the ring
```
kryzar:
```suggestion
    - ``rank`` (integer, default: ``2``) -- the rank of the ring
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+      comma seperated string of character representing the names of the
+      generators.
+
+    TESTS::
+
+        sage: K = Frac(GF(3)['T'])
+        sage: TestSuite(DrinfeldModularForms(K)).run()
+        sage: TestSuite(DrinfeldModularForms(K, 3)).run()
+        sage: TestSuite(DrinfeldModularForms(K, 4)).run()
+        sage: K = Frac(GF(7)['T'])
```
kryzar: Also try when $q$ is not prime?

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+
+    @staticmethod
+    def __classcall_private__(cls, base_ring, rank=2, group=None,
+                              has_type=False, names=None):
+        rank = ZZ(rank)  # check the type of rank
+        if not isinstance(base_ring, FractionField_generic):
+            raise TypeError("base ring must be a fraction field of a "
+                            "polynomial ring")
+        if not isinstance(base_ring.base(), PolynomialRing_general):  # not sure if this test is relevant
+            raise TypeError("the base of the base ring must be a "
```
kryzar: Codecov says this line is not tested.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+            raise NotImplementedError("Drinfeld modular forms are currently "
+                                      "only implemented for the full group")
+        if names is None: # default names
+            names = f"g1, "
+            for i in range(2, rank, 1):
+                names += f"g{i}, "
+            if has_type:
+                names += f"h{rank}"
+            else:
+                names += f"g{rank}"
```
kryzar:
```suggestion
            last = f"h{rank}" if has_type else f"g{rank}"
            names += last
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+        elif isinstance(names, str):
+            nb_names = len(names.split())
+            if nb_names != rank:
```
kryzar: Wouldn't it be possible not to mention the rank if the user gives names?

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+        return cls.__classcall__(cls, base_ring, rank, group, has_type, names)
+
+    def __init__(self, base_ring, rank, group, has_type, names):
+        self._has_type = has_type
+        self._rank = rank
+        self._base_ring = base_ring
+        q = base_ring.base_ring().cardinality()
+        if has_type:
+            degs = [q**i - 1 for i in range(1, rank, 1)]
+            degs.append((q**rank - 1)/(q - 1))
```
kryzar:
```suggestion
            degs.append((q**rank - 1) // (q - 1))
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+            degs = [q**i - 1 for i in range(1, rank + 1, 1)]
+        self._poly_ring = PolynomialRing(base_ring, rank, names=names,
+                                         order=TermOrder('wdeglex', degs))
+        self._assign_names(names)
+
+        Parent.__init__(self, base=base_ring, category=GradedAlgebras(base_ring))
+
+    def _an_element_(self):
+        r"""
+        Return an element of self.
```
kryzar, line 344: As Xavier pointed out, `self` should not be employed here.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+
+        TESTS::
```
kryzar, line 346: Why put these in a `TESTS` and not `EXAMPLES`?
Btw, I noticed that `M.an_element?` returns the docstring of `Parent.an_element`:

```
sage: M.an_element?
Docstring:     
   Returns a (preferably typical) element of this parent.

   This is used both for illustration and testing purposes. If the set
   "self" is empty, "an_element()" raises the exception
   "EmptySetError".

   This calls "_an_element_()" (which see), and caches the result.
   Parent are thus encouraged to override "_an_element_()".

   EXAMPLES:

      sage: CDF.an_element()
      1.0*I
      sage: ZZ[['t']].an_element()
      t

   In case the set is empty, an "EmptySetError" is raised:

      sage: Set([]).an_element()
      ...
      EmptySetError
Init docstring: Initialize self.  See help(type(self)) for accurate signature.
File:           ~/pro/code/sage/src/sage/structure/parent.pyx
Type:           builtin_function_or_method
sage: 
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+             g1^36*g2 + g1^28*g2^3 + g1^4*g2^9 + (T^81 + T^9 + T)*g2^10,
+             g1^81*g2^10 + g1^9*g2^28 + g1*g2^30,
+             g2^91]
+        """
+        if a is None:
+            return [self._generator_coefficient_form(i) for i in range(1, self.rank() + 1)]
+        try:
+            a = self._base_ring(a)
+        except TypeError:
+            raise TypeError("the input should be an element of the base ring")
```
kryzar: Codecov indicates that this `raise` is never testes. See also line 486.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,653 @@
...
+        if not a.denominator().is_one():
+            raise ValueError("a should be in the ring of regular functions")
+        a = a.numerator()
+        if i < 1 or i > a.degree()*self.rank():
+            raise ValueError(f"index (={i}) must be >= 1 and <= deg(a)*rank "
+                             f"(={a.degree()*self.rank()})")
+        coeff_forms = self._coefficient_forms(a)
+        return coeff_forms[i - 1]
+
+    def coefficient_forms(self, a=None):
```
kryzar, line 536: Why not removing `_coefficient_forms`, putting it in `coefficient_forms` and updating `coefficient_form` accordingly?

kryzar (changes requested):
Done with `ring.py`. I noticed that `FunctionField`s are not accepted at input. I don't know if that is something you want to change or keep like this.

```
sage: q = 3
sage: Fq = GF(q)
sage: K.<T> = FunctionField(Fq)
sage: M = DrinfeldModularForms(K, 2)
Cell In [13], line 1
----> 1 M = DrinfeldModularForms(K, Integer(2))

File ~/pro/code/sage/src/sage/misc/classcall_metaclass.pyx:320, in sage.misc.classcall_metaclass.ClasscallMetaclass.__call__()
    318 """
    319 if cls.classcall is not None:
--> 320     return cls.classcall(cls, *args, **kwds)
    321 else:
    322     # Fast version of type.__call__(cls, *args, **kwds)

File ~/pro/code/sage/src/sage/modular/drinfeld_modform/ring.py:204, in DrinfeldModularForms.__classcall_private__(cls, base_ring, rank, group, has_type, names)
    202 rank = ZZ(rank)  # check the type of rank
    203 if not isinstance(base_ring, FractionField_generic):
--> 204     raise TypeError("base ring must be a fraction field of a "
    205                     "polynomial ring")
    206 if not isinstance(base_ring.base(), PolynomialRing_general):  # not sure if this test is relevant
    207     raise TypeError("the base of the base ring must be a "
    208                     "polynomial ring")

TypeError: base ring must be a fraction field of a polynomial ring
sage: 
```

(I am available for any question, remark on critic on my review. And keep in mind that I am not very familiar with modular forms!)

xcaruso:
Hi guys,
where are we with this ticket?

kryzar:
If I reckon correctly, David's main occupation lately was starting his new job. He probably put the ticket on hold for a while!

DavidAyotte:
> Hi guys,
> where are we with this ticket?

Hello, yes I'm still alive and occupied, but please bear with me I plan to make a come back in a few weeks.

xcaruso:
Take your time; there's no rush.

DavidAyotte:
Here's an update. I just pushed some local commits I had on my machine from a couple of months ago. They pretty much addressed Antoine's comments on the `src/sage/modular/drinfeld_modform/element.py` file. I hope that the lack of clarity with what is considered a Drinfeld modular form vs graded Drinfeld modular form is fixed. I decided to stick with the term "homogeneous" which I think is clearer. 

I'm now looking at the other comments!

src/sage/modular/drinfeld_modform/tutorial.py
DavidAyotte: Yes exactly!

src/sage/modular/drinfeld_modform/tutorial.py
DavidAyotte: I've always been `\rightarrow` guy, but okay I'll use `\to`!

src/sage/modular/drinfeld_modform/tutorial.py
DavidAyotte: Yes I could, but my preference at the time was to make it a little bit easier to follow, assuming the reader could forget what are the variables `A` and `K` (even if they specified in the `_repr_`). If you really think that it is unnecessary I can make the change.

src/sage/modular/drinfeld_modform/tutorial.py
DavidAyotte: Yep I agree this sentence is indeed strange, I'll change it.

src/sage/modular/drinfeld_modform/tutorial.py
DavidAyotte: The reason I did that is because the number of coefficient forms grows quickly in function of $q$, $r$ and the degree of the polynomial $a$. So to keep the output small, I chose a small $q$. I could I gone with this setting from the start, but I just don't like even prime number (no logical reason 😅)

src/sage/modular/drinfeld_modform/tutorial.py
DavidAyotte: Thanks, in fact there was a typo here, the second equality was supposed to be
`M^{r}(\mathrm{GL}_r(A)) = \mathbb{C}_{\infty}[g_1,\ldots, g_{r-1}, h_{r}]`
(no 0 in the superscript of `M`)

DavidAyotte:
Thank you for the comments on the tutorial, they were all very relevant. I'm still not finished with some of them.

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: So classy.

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: Yes it would probably be possible and a good idea! I'll work on that.

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: Here we really want a single backslash as the weight of the form $h$ is $(q^r - 1)/(q-1)$

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: The block here doesn't really matter because this method is hidden to the user and therefore it doesn't show in the reference manual.

Also, about the docstring, that's normal because `M.an_element` is implemented in `Parent`. To see the docstring of the Drinfeld modular form implementation, you need to type `M._an_element_?`

DavidAyotte:
Small update, addressing some comments of Antoine. I'm still not finished!

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: I implemented your idea. It is now possible to only specify the names and not the rank. One can also type
`M.<g1, g2, ..., gr> = DrinfeldModularForms(K)` to name and assign the generators right from the start (similar to polynomial rings).

kryzar:
> Small update, addressing some comments of Antoine. I'm still not finished!

Merci and good luck!

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: The technicality here which led me to this implementation is because the following check in the method `coefficient_form`:
```python
if i < 1 or i > a.degree()*self.rank():
    raise ValueError(f"index (={i}) must be >= 1 and <= deg(a)*rank "
                     f"(={a.degree()*self.rank()})")
``` 
is necessary as I need to get the degree of `a` in order to check if the index `i` is in the correct range. So calling the method `coefficient_forms` (plural) would do a redundant check.

Anyway, maybe I'm missing a point here so I'm open if you have a cleaner solution.

DavidAyotte:
I've put back the need review label for this PR. I'm always open for further comments and discussion.

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,726 @@
...
+`\mathrm{GL}_r(A)` where `A = \mathbb{F}_q[T]`.
+
+The implementation is based on the following identification:
+
+.. MATH::
+
+    M^r(\mathrm{GL}_r(A))
+    = \mathbb{C}_{\infty}[g_1, \ldots, g_{r-1}, g_{r}].
+
+where `g_i` the `i`-th coefficient form of weight `q^{i} - 1`.
```
kryzar:
```suggestion
where `g_i` is the `i`-th coefficient form of weight `q^{i} - 1`.
```

src/sage/modular/drinfeld_modform/ring.py
kryzar: Okay, fair enough.

kryzar (changes requested):
Found a tiny typo. It's the only change I request. Once you have implemented it, I will give a final review to approve the changes.

kryzar (approved):

DavidAyotte:
Yay ! Thanks !

vbraun:
Documentation doesn't build, see CI

Please don't set tickets to positive review where there are obvious CI fails

kryzar:
> Documentation doesn't build, see CI
> 
> Please don't set tickets to positive review where there are obvious CI fails

Thank you for catching that; sorry for the overlook.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,445 @@
...
+            sage: M.0 == M.0
+            True
+            sage: M.0 < M.1
+            Traceback (most recent call last):
+            ...
+            TypeError: invalid comparison between Drinfeld modular forms ring elements
+        """
+        if op != op_EQ and op != op_NE:
+            raise TypeError('invalid comparison between Drinfeld'
+                            ' modular forms ring elements')
```
xcaruso: I'm not sure it's the correct way to handle this.
In general, we just return `NotImplemented`.

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,445 @@
...
+
+    def type(self):
+        r"""
+        Return the type of this graded Drinfeld form.
+
+        Recall that the *type* is the integer `m` such that
+
+        .. MATH::
+
+            f(\gamma(w)) = \mathrm{det}(\gamma)^m j(\gamma, w)^k f(w).
```
xcaruso: Is `m` really defined as an integer, or as an integer modulo q-1?

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,230 @@
...
+
+.. RUBRIC:: Preliminary notations
+
+Let `q` be a prime power and let `A` be the ring of functions of
+`\mathbb{P}^1/\mathbb{F}_q` which are regular outside a closed point
+`\infty`. This ring is the polynomial ring `\mathbb{F}_q[T]`. We denote
+by `K := \mathbb{F}_q(T)` rational function field. We endow `K` with the
+`1/T`-adic valuation and let `K_{\infty} := \mathbb{F}_q((1/T))` be the
+completion of `K`. Next, we define `\mathbb{C}_{\infty}` to be the
+completion of an algebraic closure of `K_{\infty}`. Lastly, We denote
```
xcaruso:
```suggestion
completion of an algebraic closure of `K_{\infty}`. Lastly, we denote
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,230 @@
...
+    sage: A = GF(3)['T']
+    sage: K.<T> = Frac(A)
+    sage: K
+    Fraction Field of Univariate Polynomial Ring in T over Finite Field of size 3
+    sage: K.base()  # returns A
+    Univariate Polynomial Ring in T over Finite Field of size 3
+
+.. RUBRIC:: Drinfeld period domain and action of `\mathrm{GL}_r(K_{\infty})`
+
+In the classical setting, the domain of any modular forms is the complex
```
xcaruso: I would say
```suggestion
In the classical setting, the domain of any modular form is the complex
```
but I'm not sure.

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,230 @@
...
+`\Omega^r(\mathbb{C}_{\infty})` we have a corresponding discrete
+`A`-module `\Lambda^w` which is free of rank `r`:
+
+.. MATH::
+
+    \Lambda^w := Aw_1 \oplus \cdots \oplus Aw_{r-1} \oplus A\xi.
+
+An important result is that we have analytic uniformization which is the
+analogue of complex uniformization for elliptic curves. In our setting,
+elliptic curves re replaced by Drinfeld modules. In short, there exists
```
xcaruso:
```suggestion
elliptic curves are replaced by Drinfeld modules. In short, there exists
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,230 @@
...
+.. RUBRIC:: Ring of Drinfeld modular forms
+
+Letting `M_k^{r, m}(\mathrm{GL}_r(A))` denote the space of rank `r`,
+weight `k\in (q - 1)\mathbb{Z}` and type `m~(\mathrm{mod}~q-1)`
+Drinfeld modular forms, we define
+
+.. MATH::
+
+    M^{r, 0}(\mathrm{GL}_r(A)) :=
+    \bigoplus_{k\in ZZ} M_k^{r}(\mathrm{GL}_r(A))
```
xcaruso:
```suggestion
    M^{r, 0}(\mathrm{GL}_r(A)) :=
    \bigoplus_{k\in ZZ} M_k^{r, 0}(\mathrm{GL}_r(A))
```

src/sage/modular/drinfeld_modform/tutorial.py
```diff
@@ -0,0 +1,230 @@
...
+
+In SageMath, we model the ring of type 0 Drinfeld modular forms over `K`
+as a finitely generated ring in the coefficients forms `g_i`:
+
+.. MATH::
+
+    K[g_1,\ldots, g_{r-1}, g_r].
+
+Hence, any ring element is seen as a formal algebraic combination of the
+coefficient forms forms `g_i` over `K`. Likewise, the ring of arbitrary
```
xcaruso:
```suggestion
coefficient forms `g_i` over `K`. Likewise, the ring of arbitrary
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,726 @@
...
+        Traceback (most recent call last):
+        ...
+        TypeError: base ring must be a fraction field of a polynomial ring
+
+        sage: k.<x, y> = GF(5)[]
+        sage: K = k.quotient(x+y).fraction_field()
+        sage: DrinfeldModularForms(K, 2)
+        Traceback (most recent call last):
+        ...
+        TypeError: the base of the base ring must be a polynomial ring
```
xcaruso: `TypeError` or `NotImplementedError`?

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,726 @@
...
+        Traceback (most recent call last):
+        ...
+        TypeError: rank or names must be specified
+    """
+
+    Element = DrinfeldModularFormsElement
+
+    @staticmethod
+    def __classcall_private__(cls, base_ring, rank=None, group=None,
+                              has_type=False, names=None):
```
xcaruso, line 191: Shouldn't you add some doctest here (in order to please to ``sage --coverage``)? Same remark with `__init__`?

(I think that it is okay if you split the doctests of the class into three parts.)

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,726 @@
...
+            if rank is None:
+                rank = nb_names
+            else:
+                rank = ZZ(rank)
+                if nb_names != rank:
+                    raise ValueError("the number of generators "
+                                     f"must be equal to the rank (={rank})")
+        else:
+            raise TypeError("names must be None, a comma seperated string "
+                            "or a list of string")
```
xcaruso: Would not you like also to allow to pass a prefix?
For example, for usual polynomials, one can do this
```
sage: S = PolynomialRing(ZZ, 3, names='g')
sage: S
Multivariate Polynomial Ring in g0, g1, g2 over Integer Ring
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,726 @@
...
+        T = K.gen()
+        if a is None:
+            return [self._generator_coefficient_form(i)
+                    for i in range(1, self.rank() + 1)]
+        try:
+            a = K(a)
+        except TypeError:
+            raise TypeError("a should be an element of the base ring")
+        if a.valuation(T) < 0:
+            raise ValueError("a should be in the ring of regular functions")
```
xcaruso: I'm not sure this is correct.
I think you want to avoid also e.g. $\frac 1{T+1}$, don't you?

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,726 @@
...
+            g1
+            sage: M._generator_coefficient_form(2)
+            g2
+            sage: M._generator_coefficient_form(3)
+            h3^2
+        """
+        if self._has_type and i == self.rank():
+            q = self._base_ring.base_ring().cardinality()
+            return self.gen(i-1)**(q - 1)
+        return self.gen(i - 1)
```
xcaruso, line 405: This shift is a little bit weird. Even more that it is not the same for `gen`.

xcaruso (commented):
Sorry, I have some further comments.

src/sage/modular/drinfeld_modform/element.py
DavidAyotte: In most literature I read it is defined as an integer even though it depends only on its class modulo $q-1$. That's usually a fact that you prove *after* defining modular forms, similarly to the fact that the weight of nonzero classical modular forms of level 1 must be even.

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: That's a good question. This exception comes from this check:

```python
        if not isinstance(base_ring.base(), PolynomialRing_general):
            raise TypeError("the base of the base ring must be a "
                            "polynomial ring")
```

where I check that the base of the fraction field is indeed a polynomial ring. However, I can't seem to find a way in SageMath to create the fraction field of a ring that is not an instance of a polynomial ring. Do you know if it's possible? In other word, I want something like:
```
sage: R = ring
sage: K = R.fraction_field()
sage: K
Fraction field of the ring R...
```
but R is not a polynomial ring.

src/sage/modular/drinfeld_modform/element.py
xcaruso: But then, it is not uniquely determined, isn't it?

DavidAyotte: Oh, my bad, yes that's definitely true. I will fix that. Thanks for pointing this out.

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: Indeed, I will fix it!

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: It's simply a matter of indexing and definitions. Usually, the coefficients forms at $T$ are indexed from 1 to $r$ (the $i$-th coefficient form is the coefficient in front of $\tau^i$). 

When I did that a long time ago, I wanted to follow the indexing of multivariate polynomial ring's `gen` method (which starts at zero). Also, I wanted to separate the concepts of "coefficient forms" and "generators" as coefficients form generates the ring, but not all generators are coefficients forms. Maybe it would be nice to add a functionality later were user can specify their own set of generators when instantiating the class similarly to how it is done in `ModularFormsRing`.

However, I do hear you and understand the possible confusion this might arise. What do you think about leaving this like so, but adding a note to the documentation explaining the indexing choice?

DavidAyotte:
Hello, here's an update. Please do not hesitate if you have more comments and suggestions!

src/sage/modular/drinfeld_modform/element.py
xcaruso: I saw that you raised a `NotImplementedError`.
It was not what I meant; I think that we have to `return NotImplemented`

src/sage/modular/drinfeld_modform/ring.py
xcaruso: I don't know.
Anyway, I think that we can keep as it is, that's not a big deal.

src/sage/modular/drinfeld_modform/ring.py
xcaruso: Thanks for implementing this.
Do you want to update the error message and the documentation accordingly?

By the way, there is a typo : `seperated` -> `separated`

src/sage/modular/drinfeld_modform/ring.py
xcaruso: Okay, I'm fine with this though I think I prefer changing the indexing and starting from 0 (as we do almost all the time in python and sage).

xcaruso:
Thanks. I just added a few more comments to my previous review.
After this, I think that everything will be fine with me!

xcaruso:
By the way, it seems that there are still errors in building the documentation; be careful!

DavidAyotte:
The CI tests are all green now!

src/sage/modular/drinfeld_modform/ring.py
xcaruso: The typo is still here.

kryzar:
If Xavier is happy with the changes, then so am I (I had already given green flag previously).

src/sage/modular/drinfeld_modform/element.py
```diff
@@ -0,0 +1,447 @@
...
+            sage: M.0 == M.0
+            True
+            sage: M.0 < M.1
+            Traceback (most recent call last):
+            ...
+            NotImplementedError: invalid comparison between Drinfeld modular forms ring elements
+        """
+        if op != op_EQ and op != op_NE:
+            raise NotImplementedError('invalid comparison between Drinfeld'
+                                      ' modular forms ring elements')
```
xcaruso: Again, I think that the correct way to handle this is as follows.

```suggestion
            return NotImplemented
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,767 @@
...
+            else:
+                rank = ZZ(rank)
+                if nb_names == 1 and rank > 1:
+                    g = names[0]
+                    names = [f'{g}{i}' for i in range(1, rank + 1, 1)]
+                elif nb_names != rank:
+                    raise ValueError(f"the number of generators (={nb_names}) "
+                                     f"must be equal to the rank (={rank})")
+        else:
+            raise TypeError("names must be None, a comma seperated string "
```
xcaruso:
```suggestion
            raise TypeError("names must be None, a comma separated string "
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,767 @@
...
+        if not base_ring.base().base().is_finite():
+            raise ValueError("the ring of constants must be finite")
+        if group is not None:  # placeholder
+            raise NotImplementedError("Drinfeld modular forms are currently "
+                                      "only implemented for the full group")
+        if names is None: # default names
+            if rank is None:
+                raise TypeError("rank or names must be specified")
+            rank = ZZ(rank)  # check the type of rank
+            names = [f'g{i}' for i in range(1, rank, 1)]
```
xcaruso, line 283:
```suggestion
            names = [f'g{i}' for i in range(1, rank)]
```

src/sage/modular/drinfeld_modform/ring.py
```diff
@@ -0,0 +1,767 @@
...
+            raise ValueError("base ring characteristic must be finite")
+        if not base_ring.base().base().is_field():
+            raise ValueError("the ring of constants must be a field")
+        if not base_ring.base().base().is_finite():
+            raise ValueError("the ring of constants must be finite")
+        if group is not None:  # placeholder
+            raise NotImplementedError("Drinfeld modular forms are currently "
+                                      "only implemented for the full group")
+        if names is None: # default names
+            if rank is None:
```
xcaruso, line 280: There is no doctest corresponding to this branch, I think.

xcaruso (commented):
Just a few additional (or not) small comments.

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: Indeed, sorry for overlooking this.

src/sage/modular/drinfeld_modform/element.py
DavidAyotte: Done!

src/sage/modular/drinfeld_modform/ring.py
DavidAyotte: Yes there is one, see the [line 259](https://github.com/DavidAyotte/sage/blob/drinfeld_modform/src/sage/modular/drinfeld_modform/ring.py#L259).

DavidAyotte:
I'm not sure I understand why all the following doctests run fail:
```
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/groups/misc_gps/argument_groups.py  # 6 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/interfaces/gp.py  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/libs/pari/tests.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/lazy_series.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/lazy_series_ring.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/padics/padic_generic_element.pyx  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/qqbar.py  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/stats/basic_stats.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/stats/distributions/discrete_gaussian_lattice.py  # Timed out
```

src/sage/modular/drinfeld_modform/ring.py
xcaruso: Oh sorry, I missed it.

xcaruso:
I'm not sure but I think they are all unrelated to this PR.
So, I give a positive review.

DavidAyotte:
Yay! Thank you @xcaruso and @kryzar for your reviews!

kryzar:
> Yay! Thank you @xcaruso and @kryzar for your reviews!

You did the hard work. Merci !

kryzar:
Yeaaaah, congrats @DavidAyotte!

PR #36524, opened 2023-10-24
Compile everything with meson

tobiasdiez:
Add meson configuration to compile sagelib with meson. Basic developer docs are added as well (meant as background info, not real installation instructions and thus under "developer").

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36543, opened 2023-10-25
Cleanup conditional assert

tornaria:
The code we remove here is a legacy assert which has no equivalent in python 3.12. The way it's implemented uses compile time `IF` which is deprecated in cython 3 so it's desirable to remove.

See also: https://github.com/sagemath/sage/pull/36524#discussion_r1369530294

### :memo: Checklist


tobiasdiez (approved):
Thanks! Tests are passing, so it seems to be fine to remove it. But I'll let a cython export have the final say.

mkoeppe (approved):

PR #36545, opened 2023-10-26
CI: Doctest problemmatcher

mkoeppe:
### :memo: Checklist


mkoeppe:
@tobiasdiez Any help here would be very welcome. I can't seem to get the matchers to match

tobiasdiez:
Since we have full control over the doctest output, instead of adding a problem matcher, I would add in the doctest writer a switch (checking for the `CI` env variable) and then print annotated logs. https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

Example: https://github.com/adrienverge/yamllint/blob/f63e56fd982e8dcee9f7849f80926e032cdabebd/yamllint/cli.py#L81-L88

mkoeppe:
OK, that's certainly something easy that we can do, thanks.
To human readers of our log files, it would add unfamiliar noise though, so there's some tradeoff to be considered.

mkoeppe:
Closing in favor of 
- #36558

Issue #36536, opened 2023-10-25
Failures with `--enable-system-site-packages` on various systems


mkoeppe:
`debian-bookworm` https://github.com/mkoeppe/sage/actions/runs/6622955701/job/18026345582#step:10:3976
```
  [sagemath_doc_html-none]   [reference] AttributeError: type object 'Axis' has no attribute '_set_ticklabels'
```
likewise `debian-trixie`, `debian-sid`

`fedora-33` https://github.com/mkoeppe/sage/actions/runs/6622955701/job/18026347408#step:10:5780
```
 [argon2_cffi_bindings-21.2.0]     ModuleNotFoundError: No module named 'setuptools_scm.file_finder_git'
```
(likewise for many other packages, and for `fedora-34`)

`fedora-37` https://github.com/mkoeppe/sage/actions/runs/6622955701/job/18026348122#step:10:6023
```
[matplotlib-3.8.0]     ModuleNotFoundError: No module named 'pybind11'
```

`opensuse-15.5-gcc_11-python3.11` https://github.com/mkoeppe/sage/actions/runs/6622955701/job/18026350285
```
 [pillow-10.0.1]   Usage:   
  [pillow-10.0.1]     /sage/local/var/lib/sage/venv-python3.11/bin/python3 -m pip wheel [options] <requirement specifier> ...
  [pillow-10.0.1]     /sage/local/var/lib/sage/venv-python3.11/bin/python3 -m pip wheel [options] -r <requirements file> ...
  [pillow-10.0.1]     /sage/local/var/lib/sage/venv-python3.11/bin/python3 -m pip wheel [options] [-e] <vcs project url> ...
  [pillow-10.0.1]     /sage/local/var/lib/sage/venv-python3.11/bin/python3 -m pip wheel [options] [-e] <local project path> ...
  [pillow-10.0.1]     /sage/local/var/lib/sage/venv-python3.11/bin/python3 -m pip wheel [options] <archive url/path> ...
  [pillow-10.0.1]   
  [pillow-10.0.1]   no such option: -C
  [pillow-10.0.1]   Warning: building with "python3 -m pip wheel --wheel-dir=dist --verbose --no-deps --no-index --isolated --ignore-requires-python --find-links=/sage/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels" failed.
[...]

PR #35251, opened 2023-03-09
Upgrade `ipython` to 8.17.x

mkoeppe:
### 📝 Checklist

This is the upgrade of `ipython` to 8.17 + upgrades of dependencies.

As the following issue is still open:
- https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1576
- ipython now requires prompt-toolkit >= 3.0.30, so we can no longer use the workaround of  https://github.com/sagemath/sage/issues/33428 (pinning prompt-toolkit)
we provide a proper workaround here.

- [x] I have made sure that the title is self-explanatory and the description concisely explains the PR.
- [ ] I have linked an issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36548, opened 2023-10-26
Support a few more Cython system packages

orlitzky:
The usual stuff, for:

- cypari2
- cysignals
- fpylll
- memory_allocator
- pplpy
- primecountpy

### Dependencies

- https://github.com/sagemath/sage/pull/36482


dimpase (approved):
looks good.

please put the correct tags up.

build/pkgs/primecountpy/spkg-configure.m4
```diff
@@ -0,0 +1,5 @@
+SAGE_SPKG_CONFIGURE([primecountpy], [
+  SAGE_SPKG_DEPCHECK([cysignals primecount], [
```
mkoeppe, line 2: Not necessarily for this PR, but a possible next refinement:
`primecount` is only used via `primecountpy`; so instead of listing it in `SAGE_SPKG_DEPCHECK`, we could mark `primecount` as "not required" when `primecountpy` is available -- like we do for `virtualenv` as the dependency of only `tox`.

orlitzky:
> please put the correct tags up.

?? what did I miss?

mkoeppe:
Best to rebase it on top of #36482 for the fpylll version

mkoeppe:
The changes look straightforward, but it's not quite clear what the standard of review should be for this experimental feature because so many tests are failing. In addition to what I reported in #36536, I see:

- `debian-bullseye` https://github.com/sagemath/sage/actions/runs/6658324345/job/18094988046?pr=36548
Numerous failures with the message `    sage.repl.rich_output.display_manager.RichReprWarning: Exception in _rich_repr_ while displaying object: No module named 'matplotlib.docstring'`
- same on `ubuntu-jammy`
- `fedora-33` - build errors in various packages - https://github.com/sagemath/sage/actions/runs/6658324345/job/18094988999?pr=36548
- `gentoo-python3.11` - many doctest failures from `DeprecationWarning: pkg_resources is deprecated as an API. ` https://github.com/sagemath/sage/actions/runs/6658324345/job/18094990142?pr=36548#step:10:10288

orlitzky:
> Best to rebase it on top of #36482 for the fpylll version

Done

orlitzky:
> * `gentoo-python3.11` - many doctest failures from `DeprecationWarning: pkg_resources is deprecated as an API. `

Does the container need an update? That's my configuration and I don't see any of those failures. I'm sure Francois would have heard about it too.

orlitzky:
> The changes look straightforward, but it's not quite clear what the standard of review should be for this experimental feature because so many tests are failing.

These are real bugs being exposed in our build system / metadata, but if we can't guess the cause from the logs and if no one has these old distros to poke around on, then it shouldn't be a blocker IMO.

This may still improve somewhat when we announce the feature. I've been putting it off until I was done writing `spkg-configure.m4`s, and this is one of the last sets. (IPython and friends remain blocked on prompt_toolkit, https://github.com/sagemath/sage/pull/35251#issuecomment-1716928104). But still, if we announce it and if no one cares about e.g. Fedora 33, then no one cares about Fedora 33...

mkoeppe:
> if no one has these old distros to poke around on

One does not need to "have these old distros". This all runs on Docker.

build/pkgs/primecountpy/spkg-configure.m4
dimpase: how do we do this? where is this "not required" thing should be?

orlitzky: You can see an example in e.g. `filelock/spkg-configure.m4`. In theory we could do this for any package but it requires hard-coding the entire dependency tree in m4 and even I am not in a rush to do that. This is a relatively easy case though.

mkoeppe (approved):
I guess this is fine, as the feature is marked experimental.

PR #36031, opened 2023-08-03
Cythonize `LatticePolytope.normal_form(algorithm='palp_native')`, change to default, add as a `Polyhedron` method

mkoeppe:
The current default algorithm, using the external, unmaintained `palp` package, is unreliable.

Here, as a follow-up on #35997, we make the reimplementation of the PALP algorithm in Python (`palp_native`) suitable as the new default by speeding it up.  

We also fix a bug in introduced in #35997.

We also make it available as a method of `Polyhedron` (for base ring `ZZ`).

@xuluze 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36551, opened 2023-10-26
Fix bug in QF.find_primitive_p_divisible_vector__next

tornaria:
When the vector [0,...,0,1] is a zero mod p, it was missed from the iteration. Fix by check and return it if this is the case.
    
A doctest triggering the bug is included.

The second commit adds a return_matrix option for QF.find_p_neighbor_from_vec which is related and useful, similar to QF.is_globally_equivalent_to, etc.

### :memo: Checklist

### DIFF
--- a/src/sage/quadratic_forms/quadratic_form__neighbors.py
+++ b/src/sage/quadratic_forms/quadratic_form__neighbors.py
@@ -74,6 +74,10 @@ def find_primitive_p_divisible_vector__next(self, p, v=None):
         sage: v = Q.find_primitive_p_divisible_vector__next(5, v); v
         (1, 0)
         sage: v = Q.find_primitive_p_divisible_vector__next(5, v); v
+        sage: v = Q.find_primitive_p_divisible_vector__next(2) ; v
+        (0, 1)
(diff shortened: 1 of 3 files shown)


tornaria:
Ping @yyyyx4 : if you find the time for a quick review, this is a very small PR that fixes a few annoying bugs in ternary quadratic forms. Most of the changed lines are new (simple) doctests for the bugs. Thanks.

yyyyx4:
Your changes look good to me.

As a side remark (which is orthogonal to this patch), the API with `__next` here is really strange by Python standards: I suppose it should just be a generator, using `yield`.

yyyyx4 (approved):

tornaria:
@vbraun I'd appreciate if this could be merged for 10.2, as it fixes two real annoying bugs and it's quite unobtrusive. I'm finishing a paper+code and it'd be nice if it works on unpatched 10.2.

tornaria:
Quoting https://github.com/sagemath/sage/pull/36031#issuecomment-1823669718:
> Marking this as a blocker in the hope it can sneak into 10.2 with the other open blocker

mkoeppe:
Yes, that's the documented mechanism to request that something be included in the release.

PR #36552, opened 2023-10-26
`src/sage/misc/cython.py`: Fix the workaround for setuptools_scm

mkoeppe:
The part of #36400 that was intended to fix the runtime use of cython did not actually fix it because `finalize_distribution` is called already by `Distribution.__init__`.

This shows up when invoked from a directory with `pyproject.toml` or a similar config file. As the Build & Test workflow invokes the doctester from `SAGE_ROOT/src`, these errors show up in all tests. Marking it as a blocker so it is [applied automatically in CI workflows](https://github.com/sagemath/sage/wiki/Sage-10.2-Release-Tour#open-blocker-prs-are-applied-automatically-in-ci-workflows)

Fixes https://github.com/sagemath/sage/pull/36400#issuecomment-1781525600

### :memo: Checklist

### DIFF
--- a/src/sage/misc/cython.py
+++ b/src/sage/misc/cython.py
@@ -421,7 +421,7 @@ def cython(filename, verbose=0, compile_message=False,
                         os.curdir)
 
     # This emulates running "setup.py build" with the correct options
-    dist = Distribution()
+    #
     # setuptools plugins considered harmful:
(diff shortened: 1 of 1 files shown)


tobiasdiez (approved):
Seems to work well, thanks!

PR #36562, opened 2023-10-28
`pkgs/sage-{docbuild,setup,sws2rst}`: Migrate from `setup.cfg` to `pyproject.toml`

mkoeppe:
- Split out from and part of #33577

### :memo: Checklist


pkgs/sage-docbuild/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+    "Operating System :: MacOS :: MacOS X",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Scientific/Engineering :: Mathematics",
+]
+urls = {Homepage = "https://www.sagemath.org"}
+dependencies = ["sphinx"]
```
kiwifb, line 25: Doesn't sage_docbuild also depend on jupyter-sphinx?

pkgs/sage-setup/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Scientific/Engineering :: Mathematics",
+]
+urls = {Homepage = "https://www.sagemath.org"}
+requires-python = ">=3.9, <3.12"
+dependencies = [
+    "pkgconfig",
+    "jinja2",
```
kiwifb: What about cython? I believe both jinja2 and cython are runtime dependencies.

pkgs/sage-docbuild/pyproject.toml
mkoeppe: I think we are treating the dependencies on Sphinx extensions (declared in sage_docbuild/conf.py) as belonging to sagemath_doc_html

pkgs/sage-setup/pyproject.toml
mkoeppe: The distributions that use Cython declare that in their build-system requires. I don't think we need it here.

pkgs/sage-docbuild/pyproject.toml
kiwifb: I understand the logic but I am not very comfortable with it. Part of me feels that if you are going that route, sphinx should be treated the same way. It is not needed at install either, I think.

mkoeppe: A difference is that sphinx is imported by `sage_docbuild.sphinxbuild`.

pkgs/sage-setup/pyproject.toml
kiwifb: That's an argument for distribution to declare their use of jinja2 as well. And there is actually no call to pkgconfig in sage_setup.

pkgs/sage-docbuild/pyproject.toml
kiwifb: I do not think I will argue with that one. Where to declare runtime dependency seems to be a bit unsettled.

pkgs/sage-setup/pyproject.toml
mkoeppe: Probably jinja2 should actually be an optional dependency (formerly known as extras-require) because only the packages that you sage_setup.autogen use it.

mkoeppe: Good point about pkgconfig.

kiwifb: > Probably jinja2 should actually be an optional dependency (formerly known as extras-require) because only the packages that you sage_setup.autogen use it.

Yes, and only called from one key distribution right now, I guess it is fair.

kiwifb (commented):
Hit the wrong button, this is not an extra review task.

pkgs/sage-setup/pyproject.toml
mkoeppe: After modularization, there will be more uses, but I've made the change now.

kiwifb (approved):
LGTM

kiwifb:
Last comment that came to me after approving the other request. Should we add 3.12 in the list of supported python?

kiwifb:
> Last comment that came to me after approving the other request. Should we add 3.12 in the list of supported python?

And I meant to post that in #36563

mkoeppe:
We should probably do that soon, but not in these PRs.

kiwifb:
OK

kiwifb (approved):
Let's get going.

mkoeppe:
Thank you!

mkoeppe:
@kiwifb OK to set to positive review after the last changes?

kiwifb (approved):
Let's ship it.

mkoeppe:
Thank you!

PR #36578, opened 2023-10-29
add an example of a divisor on a curve

dimpase:
complete the example of a divisor on a curve, which was missing, as noticed on sage-develop and on
https://github.com/sagemath/sage/commit/977ace651af9b99689f7b6507f91f8b4e2588ae9#r131117132


kwankyu (approved):
Thank you! LGTM.

PR #36581, opened 2023-10-30
`sage/stats/distributions`: Implement non-spherical Gaussian sampling over lattices

grhkm21:
I implemented non-spherical Gaussian sampling over lattices. It's useful for cryptography.

~~There are some TODOs that I hope to fix as soon as possible before merging, hence it's in draft status.~~ Done!

There are some missing features e.g. when $B$ is not full-rank, or computing the normalisation factor for non-integral lattices, etc. But those should be easy to fix.

Suggestions to variable naming would be appreciated too :)

TODO:
- [x] Finish writing doctest
- [x] Support taking a basis $B$ and compute $\Sigma = BB^T$ instead.
- [x] Allow setting $c$ dynamically.
- [x] Fix merge conflicts :(
- [x] Fix linting issues, coding style issues, ...


grhkm21:
Hi, it seems that the CI `build-docs` is failing because it cannot import `DGI` from `discrete_gaussian_integer`, but it is literally defined [here](https://github.com/sagemath/sage/pull/36581/files#diff-ceb24946bc277caa2df64aabf0b6de0013a67ef2d86b34fab2d9000d0517da0dR499). Does anyone know how to fix the issue? Thanks!

src/sage/stats/distributions/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .discrete_gaussian_integer import DGI
+from .discrete_gaussian_lattice import DGL
+from .discrete_gaussian_polynomial import DGP
```
tscrim: Generally speaking, we keep the `__init__.py` files empty. The `all.py` files handle the imports.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -279,34 +449,103 @@ def __init__(self, B, sigma=1, c=None, precision=None):
...
             sage: v = vector(ZZ, n, (0, 0))
             sage: v.set_immutable()
             sage: while v not in counter:
             ....:     add_samples(1000)
-            sage: while abs(m*f(v)*1.0/c/counter[v] - 1.0) >= 0.1:                      # needs sage.symbolic
+            sage: while abs(m*f(v)*1.0/nf/counter[v] - 1.0) >= 0.1:                     # needs sage.symbolic
             ....:     add_samples(1000)
 
-            sage: from sage.stats.distributions.discrete_gaussian_lattice import DiscreteGaussianDistributionLatticeSampler
+        Spherical covariance are automatically handled.
```
tscrim:
```suggestion
        Spherical covariance are automatically handled::
```
Also all of the ones below should have this ending as it signifies the next block is code (which is meant to be tested).

tscrim (changes requested):
Thank you for the contribution. I have a few general comments.

I am generally against having acronyms like `DGI` imported into the global namespace. It clutters the tab completion and the acronym doesn't really describe what it does. There can be exceptions to this, but generally speaking, you should at the very least import the fully spelled out object. I could also see this running into conflicts with something else down the road too. In particular, this is evidenced by the original doctests.

Another is all methods and functions need doctests (with functions defined in functions being an exception, but you should have a fairly decent reason for not splitting it off as a separate method). Similarly, the one-line description should be in the form "Return foo" (not "Returns foo"); this is something we are moving towards universally, so even if other methods do this, let's not make the issue worse.

Any (new) imports to the global namespace should be done with `lazy_import` to make sure the startup time is not increased.

We generally do not have setters (as we generally don't use `@properties`). Even more so, we consider most objects immutable based on their input data. Of course, there are very notable exceptions (matrices, vectors, graphs), but by having things (effectively) immutable, it means that you cannot change things and get unintended side effects.

Lastly,

grhkm21:
Hi, sorry for the delay. I will finish this tomorrow. Thanks for the really useful feedback :) I was wondering why the `@property` stuff weren't used.

grhkm21:
@tscrim Do you mind explaining this part:

> We generally do not have setters (as we generally don't use @properties). Even more so, we consider most objects immutable based on their input data. Of course, there are very notable exceptions (matrices, vectors, graphs), but by having things (effectively) immutable, it means that you cannot change things and get unintended side effects.

In particular, the lattice sampler implemented here should support changing `c`, since it's more efficient than recomputing all necessary data from scratch (see `_precompute_data`, which is run once at initialisation but not when changing `c`), so it shouldn't be immutable. How should I implement it instead?

grhkm21:
Also just to clarify, should all properties / class variables be accessed with method calls? For example, if `L` is my sampler and I want to obtain the basis, should I write `L.B` or `L.B()`? (or should I name it `L._B`?)

tscrim:
> @tscrim Do you mind explaining this part:
> 
> > We generally do not have setters (as we generally don't use @properties). Even more so, we consider most objects immutable based on their input data. Of course, there are very notable exceptions (matrices, vectors, graphs), but by having things (effectively) immutable, it means that you cannot change things and get unintended side effects.
> 
> In particular, the lattice sampler implemented here should support changing `c`, since it's more efficient than recomputing all necessary data from scratch (see `_precompute_data`, which is run once at initialisation but not when changing `c`), so it shouldn't be immutable. How should I implement it instead?

You can share immutable data across implementations by using settings (which can be the same object in memory since it is (effectively) immutable). The issue would be if you create a list `L` with a sampler `S`, modify `S`, and then use `L[0]`, you are using the modified one rather than the original. So by having (effectively) immutable objects, you can reduce potential logic errors. Of course, this isn't a requirement, but the reasoning for the general principle we follow. I am not objecting to the current implementation for the record.

tscrim:
> Also just to clarify, should all properties / class variables be accessed with method calls? For example, if `L` is my sampler and I want to obtain the basis, should I write `L.B` or `L.B()`? (or should I name it `L._B`?)

Yes, that is correct as it protects the defining data from being accidentally mutated. Of course, the `@property` does that too, but making them methods would fit with (most of) the rest of Sage. Since this exists as a more standalone, it is okay to be different. The choice is up to you.

grhkm21:
I have made changes according to your suggestions, it should be fine now. Do you mind reviewing the part dealing with `lazy_import`? I am unsure if I am doing it correctly. Thanks! :)

tscrim:
Yes, they are done correctly, thank you. I will try to finish my review this week.

You don't have to do anything about this comment, but it might be worth future-proofing this a bit. In particular, I am wondering if there should be a top-level `distributions` or `distribution_samplers` "catalog" (basically, just (lazily) importing a module that has the corresponding imports) as a common entry point. Would you find this easier to use and/or discover functionality or more complicated?

tscrim:
See `algebras.<tab>` for instance. This reduces the clutter, simplifies the import structure, and makes common things easier to discover for users.

grhkm21:
> See `algebras.<tab>` for instance. This reduces the clutter, simplifies the import structure, and makes common things easier to discover for users.

Done! That's very nice, now people can just use `distributions.DiscreteGaussianDistributionLatticeSampler`.

src/sage/stats/all.py
```diff
@@ -1,7 +1,9 @@
+import sage.stats.distributions.catalog as distributions
 
 from .r import ttest
 from .basic_stats import (mean, mode, std, variance, median, moving_average)
 from .hmm import all as hmm
+from .distributions.all import *
```
tscrim:
```suggestion
```
Now that we have the catalog, we can remove the global imports.

src/sage/stats/distributions/all.py
```diff
@@ -0,0 +1,6 @@
+# We lazy_import the following modules since they import numpy which
+# slows down sage startup
+from sage.misc.lazy_import import lazy_import
+lazy_import("sage.stats.distributions.discrete_gaussian_integer", ["DiscreteGaussianDistributionIntegerSampler"])
+lazy_import("sage.stats.distributions.discrete_gaussian_lattice", ["DiscreteGaussianDistributionLatticeSampler"])
+lazy_import("sage.stats.distributions.discrete_gaussian_polynomial", ["DiscreteGaussianDistributionPolynomialSampler"])
```
tscrim, line 6: Likewise, we can make this file empty. Although its fine to keep it as-is.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -14,17 +14,19 @@
...
 - Martin Albrecht (2014-06-28): initial version
 
+- Gareth Ma (2023-09-22): implement non-spherical sampling
+
 EXAMPLES::
 
-    sage: from sage.stats.distributions.discrete_gaussian_lattice import DiscreteGaussianDistributionLatticeSampler
-    sage: D = DiscreteGaussianDistributionLatticeSampler(ZZ^10, 3.0)
+    sage: from sage.stats.distributions.discrete_gaussian_lattice import DiscreteGaussianDistributionLatticeSampler as DGL
+    sage: D = DGL(ZZ^10, 3.0)
```
tscrim: It would be good to use the catalog here (and in many other doctests).

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -185,28 +190,30 @@ def compute_precision(precision, sigma):
...
         INPUT:
 
-        - ``tau`` -- all vectors `v` with `|v|_∞ ≤ τ·σ` are enumerated
-          (default: ``3``).
+        - ``tau`` -- (default: ``None``) all vectors `v` with `|v|_2^2 ≤ τ·σ`
+          are enumerated; if none is provided, enumerate vectors with
+          increasing norm until the sum converges to given precision. For high
+          dimension lattice, this is recommended.
+
+        - ``prec`` -- (default: ``None``) Passed to :meth:`compute_precision`
```
tscrim:
```suggestion
        - ``prec`` -- (default: ``None``) passed to :meth:`compute_precision`
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +229,214 @@ def _normalisation_factor_zz(self, tau=3):
...
+            return sum(self.f((vector(u) + base) * self.B) for u in coords)
+
+        if self.B.nrows() != self.B.ncols():
+            raise NotImplementedError("Basis must be a square matrix for now.")
+
+        if self.is_spherical and not self._c_in_lattice:
+            raise NotImplementedError("Lattice must contain 0 for now.")
+
+        if self.B.base_ring() != ZZ:
+            raise NotImplementedError("Lattice must be integral for now.")
```
tscrim:
```suggestion
            raise NotImplementedError("only implemented for the basis being a square matrix")

        if self.is_spherical and not self._c_in_lattice:
            raise NotImplementedError("only implemented when the lattice contains 0")

        if self.B.base_ring() != ZZ:
            raise NotImplementedError("only implemented for integral lattices")
```
We also follow Python's convention for error messages (lower case without a period/full-stop).

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +229,214 @@ def _normalisation_factor_zz(self, tau=3):
...
+            sage: Sigma = Matrix(ZZ, [[5, -2, 4], [-2, 10, -5], [4, -5, 5]])
+            sage: c = vector(ZZ, [7, 2, 5])
+            sage: D = DGL(ZZ^n, Sigma, c)
+            sage: r = D._maximal_r(); r
+            0.58402...
+            sage: e_vals = (D.sigma() - r^2 * D.Q).eigenvalues()
+            sage: assert all(e_val >= -1e-12 for e_val in e_vals)
+        """
+        if self.is_spherical:
+            raise RuntimeError("You have encountered a bug. File it! :)")
```
tscrim:
```suggestion
        assert self.is_sperical
```
I am not `00% sure versus the `RuntimeError. I think the assert is slightly better here, but I am not sure I could explain why I feel that way.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +229,214 @@ def _normalisation_factor_zz(self, tau=3):
...
+        while cnt < 10000:
+            nv = (Q * v).normalized()
+            if (nv - v).norm() < 1e-12:
+                break
+            v = nv
+            cnt += 1
+        res = (v[0] / (Q * v)[0]).sqrt()
+        return res
+
+    def _randomise(self, v):
```
tscrim, line 386: This method needs a doctest.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +229,214 @@ def _normalisation_factor_zz(self, tau=3):
...
+
+        - ``r`` -- (default: None) rounding parameter `r` as defined in
+          [Pei2010]_. Ignored for spherical Gaussian parameter. If not provided,
+          set to be the maximal possible such that Σ - r²BBᵀ is positive
+          definite.
+
         - ``precision`` -- bit precision `≥ 53`.
 
+        - ``sigma_basis`` -- (default: False) When set, ``sigma`` is treated as
+            a (row) basis, i.e. the covariance matrix is computed by ``Σ = SSᵀ``.
```
tscrim:
```suggestion
        - ``c`` -- (default: 0) center `c`, any vector in `\ZZ^n` is
          supported, but `c ∈ \Lambda(B)` is faster
        - ``r`` -- (default: ``None``) rounding parameter `r` as defined in
          [Pei2010]_; ignored for spherical Gaussian parameter; if not provided,
          set to be the maximal possible such that `\Sigma - SS^T` is positive
          definite
        - ``precision`` -- bit precision `≥ 53`
        - ``sigma_basis`` -- (default: ``False``) when set, ``sigma`` is treated as
            a (row) basis, i.e. the covariance matrix is computed by ``Σ = SSᵀ``
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -317,132 +556,236 @@ def __init__(self, B, sigma=1, c=None, precision=None):
...
-            D = []
-            for i in range(self.B.nrows()):
-                sigma_ = self._sigma / self._G[i].norm()
-                D.append(DiscreteGaussianDistributionIntegerSampler(sigma=sigma_))
-            self.D = tuple(D)
-            self.VS = FreeModule(ZZ, B.nrows())
+        .. note::
+
+           Do not call this method directly, it is called automatically from :func:`DiscreteGaussianDistributionLatticeSampler.__init__`.
+        """
```
tscrim:
```suggestion
        .. NOTE::

           Do not call this method directly, it is called automatically from
           :func:`DiscreteGaussianDistributionLatticeSampler.__init__`.
        """
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -317,132 +556,236 @@ def __init__(self, B, sigma=1, c=None, precision=None):
...
+            v = self._call_non_spherical()
+        elif self._c_in_lattice:
             v = self._call_in_lattice()
         else:
             v = self._call()
         v.set_immutable()
         return v
 
-    @property
+    def f(self, x):
```
tscrim, line 664: This needs a doctest.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -479,7 +822,51 @@ def _call(self):
...
             assert sigma_ > 0
-            z = DiscreteGaussianDistributionIntegerSampler(sigma=sigma_, c=c_, algorithm="uniform+online")()
+            z = DGI(sigma=sigma_, c=c_, algorithm="uniform+online")()
             c = c - z * B[i]
             v = v + z * B[i]
         return v
+
+    def add_offline_samples(self, cnt=1):
+        """
+        Precompute samples from B⁻¹D₁ to be used in :meth:`_call_non_spherical`
```
tscrim:
```suggestion
        Precompute samples from `B^{-1} D_1` to be used in :meth:`_call_non_spherical`.
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -479,7 +822,51 @@ def _call(self):
...
+            sage: Sigma = Matrix([[5, -2, 4], [-2, 10, -5], [4, -5, 5]])
+            sage: D = DGL(ZZ^3, Sigma, c=(1/2,0,0))
+            sage: L = [D._call_non_spherical() for _ in range(2^12)]
+            sage: mean_L = sum(L) / len(L)
+            sage: norm(mean_L.n() - D.c()) < 0.25
+            True
+
+        .. note::
+
+           Do not call this method directly, call :func:`DiscreteGaussianDistributionLatticeSampler.__call__` instead.
```
tscrim:
```suggestion
        .. NOTE::

           Do not call this method directly, call
           :func:`DiscreteGaussianDistributionLatticeSampler.__call__` instead.
```

tscrim (changes requested):
Sorry for taking so long to continue the review. Thank you for the changes. I am glad to like the catalog approach.

I have a few more specific comments below.

src/sage/stats/distributions/discrete_gaussian_lattice.py
grhkm21: Ooo I didn't know it's a Python convention, but I learned that in the past few months

grhkm21:
I turned all the unicode docstrings into math code (`` `\sigma` ``) but I should check that it works when docs are done

src/sage/stats/distributions/catalog.py
```diff
@@ -0,0 +1,33 @@
...
+- :class:`distributions.discrete_gaussian_integer.DiscreteGaussianDistributionIntegerSampler
+  <sage.stats.distributions.discrete_gaussian_integer.DiscreteGaussianDistributionIntegerSampler>`
+- :class:`distributions.discrete_gaussian_lattice.DiscreteGaussianDistributionLatticeSampler
+  <sage.stats.distributions.discrete_gaussian_lattice.DiscreteGaussianDistributionLatticeSampler>`
+- :class:`distributions.discrete_gaussian_polynomial.DiscreteGaussianDistributionPolynomialSampler
+  <sage.stats.distributions.discrete_gaussian_polynomial.DiscreteGaussianDistributionPolynomialSampler>`
+
+To import these names into the global namespace, use::
+
+    sage: from sage.stats.channels_catalog import *
```
tscrim: This doctest is failing. Looks like the import is not quite correct (older version?).

src/sage/stats/distributions/catalog.py
```diff
@@ -0,0 +1,33 @@
...
+
+"""
```
tscrim, line 20:
```suggestion
"""
```
We typically don't have the blankline here (at least, it is what we are largely moving towards).

src/sage/stats/distributions/catalog.py
```diff
@@ -0,0 +1,33 @@
...
+#  Distributed under the terms of the GNU General Public License (GPL),
+#  version 2 or later (at your preference).
+#
+#                  http://www.gnu.org/licenses/
+#*****************************************************************************
+
+from sage.misc.lazy_import import lazy_import as _lazy_import
+_lazy_import("sage.stats.distributions.discrete_gaussian_integer", ["DiscreteGaussianDistributionIntegerSampler"])
+_lazy_import("sage.stats.distributions.discrete_gaussian_lattice", ["DiscreteGaussianDistributionLatticeSampler"])
+_lazy_import("sage.stats.distributions.discrete_gaussian_polynomial", ["DiscreteGaussianDistributionPolynomialSampler"])
```
tscrim:
```suggestion
from sage.misc.lazy_import import lazy_import
lazy_import("sage.stats.distributions.discrete_gaussian_integer", ["DiscreteGaussianDistributionIntegerSampler"])
lazy_import("sage.stats.distributions.discrete_gaussian_lattice", ["DiscreteGaussianDistributionLatticeSampler"])
lazy_import("sage.stats.distributions.discrete_gaussian_polynomial", ["DiscreteGaussianDistributionPolynomialSampler"])

del lazy_import
```
No need for the special handling.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -1,30 +1,31 @@
 # -*- coding: utf-8 -*-
-r"""
-Discrete Gaussian Samplers over Lattices
+r""" Discrete Gaussian Samplers over Lattices
```
tscrim:
```suggestion
r"""
Discrete Gaussian Samplers over Lattices
```
This is how we do our documentation. All one-lines begin on the next line. (Also, IMO, it is also easier to parse the intent and read it.)

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -1,30 +1,31 @@
...
-`\exp(-|x-c|_2^2/(2σ^2))`. More precisely lattice vectors in `x ∈ Λ` are
-returned with probability:
+discrete Gaussian distribution. That is, if `\sigma` is big enough relative to
+the provided basis, then vectors are returned with a probability proportional
+to `\exp(-|x - c|_2^2 / (2\sigma^2))`. More precisely lattice vectors in `x \in
+\Lambda` are returned with probability:
 
-    `\exp(-|x-c|_2^2/(2σ²))/(∑_{x ∈ Λ} \exp(-|x|_2^2/(2σ²)))`
+    `\exp(-|x - c|_2^2 / (2\sigma^2)) / (\Sigma_{x \in \Lambda} \exp(-|x|_2^2 /
+    (2\sigma^2)))`
```
tscrim:
```suggestion
.. MATH::

    \frac{\exp(-|x - c|_2^2 / (2\sigma^2))}{\sum_{x \in \Lambda} \exp(-|x|_2^2 /
    (2\sigma^2))}.
```
Also $\sum_{a \in A}$ is different than $\Sigma_{a \in A}$.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -185,28 +187,29 @@ def compute_precision(precision, sigma):
...
         return precision
 
-    def _normalisation_factor_zz(self, tau=3):
+    def _normalisation_factor_zz(self, tau=None, prec=None):
         r"""
-        This function returns an approximation of `∑_{x ∈ \ZZ^n}
-        \exp(-|x|_2^2/(2σ²))`, i.e. the normalisation factor such that the sum
-        over all probabilities is 1 for `\ZZⁿ`.
+        This function returns an approximation of `\Sigma_{x \in B}
+        \exp(-|x|_2^2 / (2\sigma²))`, i.e. the normalisation factor such that the sum
```
tscrim:
```suggestion
        This function returns an approximation of `\sum_{x \in B}
        \exp(-|x|_2^2 / (2\sigma^2))`, i.e. the normalization factor such that the sum
```
We also typically use American English spellings (but you don't have to go and fix this everywhere; I just noticed it here).

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +225,218 @@ def _normalisation_factor_zz(self, tau=3):
...
+            cnt = ZZ(Q.__pari__().qfrep(bound, 0)[bound - 1])
+            inc = 2 * cnt * f_or_hat(bound / det**self.n)
+            if cnt > 0 and res == res + inc:
+                return R(norm_factor * res)
+            res += inc
+
+    @cached_method
+    def _maximal_r(self):
+        r"""
+        This function computes the largest value `r > 0` such that `Σ - r²BBᵀ`
```
tscrim:
```suggestion
        This function computes the largest value `r > 0` such that `\Sigma - r^2 BB^T`.
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +225,218 @@ def _normalisation_factor_zz(self, tau=3):
...
+            sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^3, Sigma)
+            sage: all(D._randomise([0, 0, 0]).norm() <= 16 for _ in range(100))
+            True
+        """
+        return vector(ZZ, [DiscreteGaussianDistributionIntegerSampler(self.r, c=vi)() for vi in v])
+
+    def __init__(self, B, sigma=1, c=0, r=None, precision=None, sigma_basis=False):
         r"""
-        Construct a discrete Gaussian sampler over the lattice `Λ(B)`
+        Construct a discrete Gaussian sampler over the lattice `\LAmbda(B)`
```
tscrim:
```suggestion
        Construct a discrete Gaussian sampler over the lattice `\Lambda(B)`
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -317,137 +555,250 @@ def __init__(self, B, sigma=1, c=None, precision=None):
...
+            sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^3, 3.0, c=(1,0,0))
+            sage: D.set_c((2, 0, 0))
+            sage: D
+            Discrete Gaussian sampler with Gaussian parameter σ = 3.00000000000000, c=(2, 0, 0) over lattice with basis
+            <BLANKLINE>
+            [1 0 0]
+            [0 1 0]
+            [0 0 1]
+
+        .. note:
```
tscrim:
```suggestion
        .. NOTE::
```

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -458,16 +809,16 @@ def _call(self):
...
-            sage: L = [D._call() for _ in range(2^12)]  # long time
-            sage: mean_L = sum(L) / len(L)              # long time
-            sage: norm(mean_L.n() - D.c) < 0.25         # long time
+            sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^3, 3.0, c=(1/2,0,0))
+            sage: L = [D._call() for _ in range(2^12)]
+            sage: mean_L = sum(L) / len(L)
+            sage: norm(mean_L.n() - D.c()) < 0.25
             True
 
         .. note::
```
tscrim:
```suggestion
        .. NOTE::
```
A while-we-are-at-it thing.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -483,3 +834,46 @@ def _call(self):
             c = c - z * B[i]
             v = v + z * B[i]
         return v
+
+    def add_offline_samples(self, cnt=1):
+        """
+        Precompute samples from `B^{-1}D_1` to be used in :meth:`_call_non_spherical`
```
tscrim:
```suggestion
        Precompute samples from `B^{-1}D_1` to be used in :meth:`_call_non_spherical`.
```

tscrim (changes requested):
Thank you. I am just slightly worried about the unicode (within latex markings too) not translating properly into mathjax and/or the pdf docbuild. In principle., it should be fine, but proper latex should always work (provided it is not too fancy). So I am probably just being paranoid, but I definitely know this way will work.

Some other little details that are in the specific comments:

- All one-line descriptions in the docstring start on their own line and end with a period/full-stop except for the module "title".
- We use `.. NOTE::` for all sphinx note formatted blocks.

I think this is just about ready otherwise. Thank you for all the changes and polishing!

src/sage/stats/distributions/discrete_gaussian_lattice.py
grhkm21: Oh no 😂 What should the method be named then? If the American spelling is preferred *and* this is an internal method, might as well change it right?

tscrim: Internal methods can be changed without any deprecation (which is only needed for anything public). Feel free to change it or decide to keep it as-is. (For reference, I am American, so I am biased in this. :P )

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -3,28 +3,32 @@
...
+discrete Gaussian distribution. That is, if `\sigma` is big enough relative to
+the provided basis, then vectors are returned with a probability proportional
+to `\exp(-|x - c|_2^2 / (2\sigma^2))`. More precisely lattice vectors in `x \in
+\Lambda` are returned with probability:
 
-    `\exp(-|x-c|_2^2/(2σ²))/(∑_{x ∈ Λ} \exp(-|x|_2^2/(2σ²)))`
+.. MATH::
+
+    `\frac{\exp(-|x - c|_2^2 / (2\sigma^2))}{\sum_{v \in \Lambda} \exp(-|v|_2^2 /
+    (2\sigma^2))}`
```
tscrim:
```suggestion
    \frac{\exp(-|x - c|_2^2 / (2\sigma^2))}{\sum_{v \in \Lambda} \exp(-|v|_2^2 /
    (2\sigma^2))}.
```
The extra backticks are not needed here.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -222,51 +228,217 @@ def _normalisation_factor_zz(self, tau=3):
...
+            cnt = ZZ(Q.__pari__().qfrep(bound, 0)[bound - 1])
+            inc = 2 * cnt * f_or_hat(bound / det**self.n)
+            if cnt > 0 and res == res + inc:
+                return R(norm_factor * res)
+            res += inc
+
+    @cached_method
+    def _maximal_r(self):
+        r"""
+        This function computes the largest value `r > 0` such that `\Sigma - r^2BB^@`
```
tscrim: Is the `@` here intentional or should it be a `T`?

tscrim (commented):
Sorry for loosing track of this. Thank you for the changes. Just a few small more little details, then I will set a positive review.

src/sage/stats/distributions/discrete_gaussian_lattice.py
grhkm21: It's a `T`, thanks.

grhkm21:
Thank you!!!!! The issue should be okay now.

tscrim (approved):
Thank you.

tornaria:
There's something weird going on here:
```
$ sage -t --random-seed=170991570709228473284945102686749566194 src/sage/stats/distributions/discrete_gaussian_lattice.py 
 
Running doctests with ID 2024-04-22-23-05-42-4d9a14f7.
Running with SAGE_LOCAL='/usr' and SAGE_VENV='/usr'
Using --optional=pip,sage
Features to be detected: 4ti2,benzene,bliss,buckygen,conway_polynomials,csdp,cvxopt,cvxopt,database_cremona_ellcurve,database_cremona_mini_ellcurve,database_cubic_hecke,database_ellcurves,database_graphs,database_jones_numfield,database_knotinfo,dvipng,ecm,fpylll,fricas,gap_package_atlasrep,gap_package_design,gap_package_grape,gap_package_guava,gap_package_hap,gap_package_polycyclic,gap_package_qpa,gap_package_quagroup,gfan,graphviz,imagemagick,ipython,jmol,jupymake,kenzo,latte_int,lrcalc_python,lrslib,matroid_database,mcqd,meataxe,mpmath,msolve,nauty,networkx,numpy,palp,pandoc,pdf2svg,pdftocairo,pexpect,phitigra,pillow,plantri,polytopes_db,polytopes_db_4d,pplpy,primecountpy,ptyprocess,pynormaliz,pyparsing,python_igraph,requests,rubiks,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.sat,sage.schemes,sage.symbolic,sage_numerical_backends_coin,sagemath_doc_html,scipy,singular,sphinx,symengine_py,sympy,tdlib,threejs
Doctesting 1 file.
sage -t --warn-long 54.1 --random-seed=170991570709228473284945102686749566194 src/sage/stats/distributions/discrete_gaussian_lattice.py
    Timed out
Tests run before process (pid=3842) timed out:
sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^10, 3.0) ## line 24 ##
[...]
sage: n = 3 ## line 478 ##
sage: Sigma = Matrix(ZZ, [[5, -2, 4], [-2, 10, -5], [4, -5, 5]]) ## line 479 ##
sage: c = vector(ZZ, [7, 2, 5]) ## line 480 ##
sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^n, Sigma, c) ## line 481 ##
sage: nf = D._normalisation_factor_zz(); nf # This has not been properly implemented ## line 482 ##
63.7692733335502
sage: while v not in counter: add_samples(1000) ## line 484 ##
sage: while abs(m*f(v)*1.0/nf/counter[v] - 1.0) >= 0.1: add_samples(1000) ## line 485 ##
------------------------------------------------------------------------
[... irrelevant stack trace ...]
sage -t --warn-long 54.1 --random-seed=170991570709228473284945102686749566194 src/sage/stats/distributions/discrete_gaussian_lattice.py  # Timed out
Total time for all tests: 600.7 seconds
    cpu time: 0.0 seconds
    cumulative wall time: 0.0 seconds
Features detected for doctesting: sage.plot,sage.symbolic
```

The last test seems wrong to me, it only passes by chance.

Can you have a look @grhkm21 @tscrim ?

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -279,34 +451,102 @@ def __init__(self, B, sigma=1, c=None, precision=None):
...
+        matrix::
+
+            sage: n = 3
+            sage: Sigma = Matrix(ZZ, [[5, -2, 4], [-2, 10, -5], [4, -5, 5]])
+            sage: c = vector(ZZ, [7, 2, 5])
+            sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^n, Sigma, c)
+            sage: nf = D._normalisation_factor_zz(); nf # This has not been properly implemented
+            63.76927...
+            sage: while v not in counter: add_samples(1000)
+            sage: while abs(m*f(v)*1.0/nf/counter[v] - 1.0) >= 0.1: add_samples(1000)
```
tornaria, line 485: This test seems wrong... 

Unless we are lucky, this will never finish. Note that `counter` was not reset, we are running the test with the vector from the previous sample, and sometimes the test passes even with a different normalization factor. But if the test doesn't pass on the first few attempts, it won't pass because `m` tends to infinity while `counter[v]` remains constant.

Make sure you reset `counter` and you will see this test looping forever always.

grhkm21: Sorry about this, I didn’t test with —long much (but why did CI not catch it?)

Can you fix it for me? I am in exam seasons right now, sorry

tornaria: This is not with `--long`. It's a particular seed that causes the failure.

The test passes by chance, from leftover samples of the previous test.

In fact, the normalization factor seems about 20% off, so the test (if properly implemented) would not have much chance of passing:
```
$ sage
│ SageMath version 10.4.beta3, Release Date: 2024-04-12              │
│ Using Python 3.12.3. Type "help()" for help.                       │
sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(ZZ^3, Matrix(ZZ, [[5, -2, 4], [-2, 10, -5], [4, -5, 5]]), vector(ZZ, [7, 2, 5]))
sage: nf = D._normalisation_factor_zz()
/usr/lib/python3.12/site-packages/sage/stats/distributions/discrete_gaussian_lattice.py:298: UserWarning: Note: `_normalisation_factor_zz` has not been properly implemented for non-spherical distributions.
  warn("Note: `_normalisation_factor_zz` has not been properly "
sage: %time m = 100_000; sum(D() == vector(ZZ,[7,2,5]) for _ in range(m)) * nf / m
CPU times: user 11.5 s, sys: 67.4 ms, total: 11.5 s
Wall time: 11.6 s
0.813695927736100
sage: %time m = 100_000; sum(D() == vector(ZZ,[7,2,5]) for _ in range(m)) * nf / m
CPU times: user 11.3 s, sys: 62.7 ms, total: 11.4 s
Wall time: 11.4 s
0.790738989336022
sage: %time m = 100_000; sum(D() == vector(ZZ,[7,2,5]) for _ in range(m)) * nf / m
CPU times: user 11.5 s, sys: 46.8 ms, total: 11.5 s
Wall time: 11.6 s
0.783086676535996
```

tornaria: Ping @grhkm21 @tscrim

Can we either get this fixed before 10.4 or revert it until it can be fixed properly?

tscrim: Indeed, the test is bad because it does not reset `v` nor `counter`, so it is using the same one from the previous test (which has a similar enough normalization factor). This is why it wasn't getting picked up generally by the tester.

I don't think I would be able to fix this myself. I propose we effectively disable this test with `# known bug`. Reverting the whole PR feels a bit extreme since the normalization doesn't seem to be a key part of the PR (it is hidden, not called from anywhere as far as I can tell, and emits a warning).

grhkm21: Sorry for the late reply. I fixed it, it's just copying some lines above. Please kindly review #38209 and sorry for the bad code :(

PR #36584, opened 2023-10-30
implemented power of graph function under basic methods

### LINKED ISSUE #36582: Implement a function to find the power of graph
### Problem Description

The problem we are seeking to solve with this feature request is the absence of a "power" function in the SageMath library for graph theory. In graph theory, a branch of mathematics, the `k`th power `G^k` of an undirected graph `G` is another graph that has the same set of vertices, but in which two vertices are adjacent when their distance in `G` is at most `k`. 

### Proposed Solution

One implementation for the power of the graph is a brute-force one that computes the shortest distances between all pairs of nodes using Breadth-First Search (BFS). 

### Alternatives Considered

The above approach is inefficient for large graphs and results in a time complexity of O(n^3) in the worst case, where n is the number of vertices in the graph. This can be a performance bottleneck when working with graphs of significant size.

Some optimization methods: 
- For unweighted graphs, we can utilize matrix exponentiation techniques to compute the kth power of the graph efficiently.
- For weighted graphs, we can adapt the Floyd-Warshall algorithm to compute all-pairs shortest paths efficiently and then construct the kth power graph based on these shortest paths.

### Additional Information

_No response_

saatvikraoIITGN:
### Description

This PR introduces a new function, `power_of_graph`, to compute the kth power of an undirected, unweighted graph efficiently using the shortest distances method. The proposed function leverages Breadth-First Search (BFS) to calculate the power graph, providing a practical and scalable solution.

### Why is this change required?

The change is required to address the need for efficiently calculating the kth power of a graph, a fundamental operation in graph theory. The PR aims to incorporate this feature into the SageMath library. 

Fixes #36582 

### Checklist

- [x] I have linked a relevant issue or discussion (if applicable).

### :hourglass: Dependencies

#36582 : To implement power of graph


dcoudert:
It would be better to implement method `power` in `generic_graph.py` to address both directed and undirected graphs, and to use builtin fonctions .

For instance, 
```py
other = self.copy()
for u in self:
    for v in self.breadth_first_search(u, distance=k):
        other.add_edge(u, v)
```
One difficulty is to deal with (di)graphs with loops and/or multiple edges.

dcoudert:
For your information, we also have method `distance_graph` that returns the graph with only edges between vertices at distance exactly `k` in the original graph.

saatvikraoIITGN:
> It would be better to implement method `power` in `generic_graph.py` to address both directed and undirected graphs, and to use builtin fonctions .
> 
> For instance,
> 
> ```python
> other = self.copy()
> for u in self:
>     for v in self.breadth_first_search(u, distance=k):
>         other.add_edge(u, v)
> ```
> 
> One difficulty is to deal with (di)graphs with loops and/or multiple edges.

Sir the code suggested by your works fine with digraphs with loops and multiple edges

I will raise the PR with the changes

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,41 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
                                             algorithm=algorithm,
                                             weight_function=weight_function,
                                             check_weight=check_weight)[0]
+                                
+    def power(self, k):
+        r"""
+        Compute the kth power graph of an undirected, unweighted graph based on
```
dcoudert: it's also working for directed graphs.

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,41 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+        shortest distances between nodes using BFS.
+
+        INPUT:
+        - graph: An undirected, unweighted graph.
```
dcoudert: please use the same coding style as other methods.

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,41 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+        - k: The maximum path length for considering edges in the power graph.
+
+        OUTPUT:
+        - The kth power graph based on shortest distances between nodes.
+
+        EXAMPLE:
+
+        sage: G = Graph([(0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
```
dcoudert: 4 more spaces needed before `sage:`

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,41 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+        sage: k = 2
+        sage: PG = G.power(k)
+        sage: PG.edges()
```
dcoudert: the reported order of edges may change. You can use `PG.edges(sort=True, labels=False)` here as vertex labels are comparable.

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,41 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+        [(0, 1), (0, 2), (0, 3), (0, 4), (1, 0), (1, 2), (1, 3), (1, 4), (2, 0), (2, 1), (2, 3), (2, 4), (3, 0), (3, 1), (3, 2), (3, 4), (4, 0), (4, 1), (4, 2), (4, 3), (5, 4), (4, 5)]
+
+        sage: G = DiGraph([(0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
+        sage: k = 3
+        sage: PG = g.power(k)
+        sage: PG.edges()
+        [(0, 1, None), (0, 2, None), (0, 3, None), (0, 4, None), (1, 0, None), (1, 2, None), (1, 3, None), (1, 4, None), (1, 5, None), (2, 0, None), (2, 1, None), (2, 3, None), (2, 4, None), (2, 5, None), (3, 0, None), (3, 1, None), (3, 2, None), (4, 5, None)]
+
+        """
+        power_of_graph = self.copy()
```
dcoudert: if the graph allows multiple edges, the original edges of the graph might be duplicated. You must take care of such case.

dcoudert (changes requested):
The coding style must be improved (see how it is done in other methods) and some care is needed for graphs with multiple edges.

src/sage/graphs/generic_graph.py
saatvikraoIITGN: I can't find a function to test for multigraphs

saatvikraoIITGN:
```
power_of_graph = self.copy()
    added_edges = set()

    for u in self:
        for v in self.breadth_first_search(u, distance=k):
            if u != v and (u, v) not in added_edges and (v, u) not in added_edges:
                power_of_graph.add_edge(u, v)
                added_edges.add((u, v))

    return power_of_graph
```

Is this what you meant?

dcoudert: The first question to ask is why do you need to copy the graph ? The only reason I see is to copy the position of the vertices (if set). Otherwise, I don't know. Furthermore, if `self` is immutable, the copy should be immutable as well in which case your code will fail.

So instead of using a copy of the graph, it might be better to initialize a (Di)Graph. You can add it the set of vertices, and you can set the position of the vertices. Then the code should be ok.

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,50 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+            sage: PG = g.power(k)
+            sage: PG.edges(sort=True, labels=False)
+            [(0, 1), (0, 2), (0, 3), (0, 4), (1, 0), (1, 2), (1, 3), (1, 4), (1, 5), (2, 0), (2, 1), (2, 3), (2, 4), (2, 5), (3, 0), (3, 1), (3, 2), (4, 5)]
+
+        """
+        from sage.graphs.graph import DiGraph, Graph
+
+        power_of_graph = DiGraph() if self.is_directed() else Graph()
+
+        added_edges = set()
```
dcoudert: not needed anymore

src/sage/graphs/generic_graph.py
```diff
@@ -15745,6 +15745,50 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+        Testing on directed graphs::
+
+            sage: G = DiGraph([(0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
+            sage: k = 3
+            sage: PG = g.power(k)
+            sage: PG.edges(sort=True, labels=False)
+            [(0, 1), (0, 2), (0, 3), (0, 4), (1, 0), (1, 2), (1, 3), (1, 4), (1, 5), (2, 0), (2, 1), (2, 3), (2, 4), (2, 5), (3, 0), (3, 1), (3, 2), (4, 5)]
+
+        """
+        from sage.graphs.graph import DiGraph, Graph
```
dcoudert: no ! `DiGraph` is in `sage.graphs.digraph`.

Are you testing your code before pushing it ?

dcoudert (changes requested):

saatvikraoIITGN:
@dcoudert does the code seem fine?

dcoudert (changes requested):
1. Check and correct the doctest errors reported by the CI
2. Although it is currently possible to import `DiGraph` from `sage.graphs.graph`,  please don't do that.
3. Check and document the expected result when `k < 0`, `k = 0`, `k=1`, `k=+oo`
4. Respect the 80 columns mode in comments and doctests, as done in other methods.
5. Add empty line after `INPUT:` and `OUTPUT:` as done in other methods.
6. Set the name of the graph before returning it.
7. Add a `TESTS::` block and check the behavior of the method in corner cases, when the graph has multiple edges

saatvikraoIITGN:
@dcoudert What do you mean by `Set the name of the graph before returning it.`

src/sage/graphs/generic_graph.py
```diff
@@ -15746,6 +15746,93 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
                                             check_weight=check_weight)[0]
 
+    def power(self, k):
+        r"""
+        Compute the kth power graph of an unweighted graph based on shortest
+        distances between nodes using BFS.
+
+        INPUT:
+
+        - ``k`` -- integer; the maximum path length for considering edges in the power graph.
```
dcoudert: When the description of a parameter is on multiple lines, the second line has 2 spaces more.
```py
        - ``k`` -- integer; the maximum path length for considering edges in the
          power graph.
```
See other methods for examples.

saatvikraoIITGN:
> > @dcoudert What do you mean by `Set the name of the graph before returning it.`
> 
> ```python
> sage: G = graphs.PetersenGraph()
> sage: G
> Petersen graph: Graph on 10 vertices
> sage: G.name("a cool name")   # Here we change the name
> sage: G
> a cool name: Graph on 10 vertices
> sage: G = Graph(name="My empty graph")    # Here we set the name of a new graph
> sage: G
> My empty graph: Graph on 0 vertices
> ```

For eg.: in the following test
```
sage: G = Graph([(0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
sage: k = 2
sage: PG = G.power(k)
sage: PG.edges(sort=True, labels=False)```

We have to name the graph, `G`, before `PG = G.power(k)`?

**Comment by dcoudert** [comment]
> We have to name the graph, `G`, before `PG = G.power(k)`?

No, there is no need to name the graph here.

I'm saying that you can set the name inside the method, as is done for instance in method `complement`
```py
sage: G = graphs.PetersenGraph()
sage: H = G.complement()
sage: H
complement(Petersen graph): Graph on 10 vertices
sage: H.name()
'complement(Petersen graph)'
```

src/sage/graphs/generic_graph.py
```diff
@@ -15746,6 +15746,97 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
                                             weight_function=weight_function,
                                             check_weight=check_weight)[0]
 
+    def power(self, k):
+        r"""
+        Compute the kth power graph of an unweighted graph based on shortest
```
dcoudert:
```diff
-        Compute the kth power graph of an unweighted graph based on shortest
-        distances between nodes using BFS.
+        Return the `k`-th power graph of ``self``.
+
+        In the `k`-th power graph of a graph `G`, there is an edge between
+        any pair of vertices at distance at most `k` in `G`, where the
+        distance is considered in the unweighted graph. In a directed graph,
+        there is an arc from a vertex `u` to a vertex `v` if there is a path
+        of length at most `k` in `G` from `u` to `v`.
```

src/sage/graphs/generic_graph.py
```diff
@@ -15746,6 +15746,97 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+            sage: PG.edges(sort=True, labels=False)
+            [(0, 1), (0, 2), (0, 3), (0, 4), (1, 0), (1, 2), (1, 3), (1, 4), (1, 5), (2, 0), (2, 1), (2, 3), (2, 4), (2, 5), (3, 0), (3, 1), (3, 2), (4, 5)]
+
+        TESTS:
+
+        Testing when k < 0::
+
+            sage: G = Graph([(0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
+            sage: k = -2
+            sage: PG = G.power(k)
```
dcoudert: you can simplify by directly calling `G.power(-2)`.  No need to define `k`

dcoudert (changes requested):

src/sage/graphs/generic_graph.py
```diff
@@ -15746,6 +15746,95 @@ def distance_all_pairs(self, by_weight=False, algorithm=None,
...
+
+            sage: G = Graph([(0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
+            sage: PG = G.power(Infinity)
+            Traceback (most recent call last):
+            ...
+            ValueError: distance must be a non-negative integer, not +Infinity
+
+        Testing on graph with multiple edges::
+
+            sage: G = DiGraph([(0, 1), (0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)])
```
dcoudert: By default, (di)graphs are simple even if the input list of edges has multiple copies of an edge. So you must specify `muliedges=True`. Sorry for not seing that earlier.
```
sage: G = DiGraph([(0, 1), (0, 1), (1, 2), (2, 3), (3, 0), (2, 4), (4, 5)], multiedges=True)
```

dcoudert (changes requested):

dcoudert (approved):
LGTM.

PR #36585, opened 2023-10-30
implemented the transpose of graph function in graphs

### LINKED ISSUE #36583: Implement a function to find the transpose of a graph
### Problem Description

The problem is that the SageMath graph library currently lacks a built-in function to compute the transpose of a directed graph efficiently. The transpose of a directed graph is a fundamental operation in graph theory and graph algorithms, and having an efficient implementation is essential for various applications.

In the mathematical and algorithmic study of graph theory, the converse, *transpose*, or reverse of a directed graph `G` is another directed graph on the same set of vertices with all of the edges reversed compared to the orientation of the corresponding edges in `G`. That is, if `G` contains an edge `(u, v)`, then the converse/transpose/reverse of `G` contains an edge `(v, u)` and vice versa.

### Proposed Solution

We propose the addition of a new function called `transpose_graph(G`) to the SageMath graph library. This function will provide a convenient and efficient way to calculate the transpose of a directed graph `G`. The transpose of a directed graph has the same vertices as the original graph, but the direction of each edge is reversed.

Here's how the function works:

- Input: The function takes one argument, `G`, which is a directed graph (DiGraph) that you want to find the transpose of.
- Output: The function returns the transpose of the input graph `G`.
- Operation: The function first checks whether the input graph G is simple, meaning it has no parallel edges or self-loops, ensuring the validity of the operation.
- It creates a copy of G with a dense data structure, which is suitable for efficiently modifying the graph.
- Next, it iterates through the edges of `G`, reverses the direction of each edge, and adds them to the new graph GT. This process efficiently computes the transpose of the graph.
- If the original graph has a name, the function updates the name of the resulting transpose graph `GT` to indicate that it's the transpose of the original graph.
- If the original graph `G` is immutable, the function returns an immutable copy of the transpose graph to maintain consistency with the SageMath immutable data structure.

This proposed solution offers an efficient and user-friendly way to compute the transpose of a directed graph in SageMath, enhancing the functionality and usability of the graph library for graph theory and related applications.

### Alternatives Considered

Here are some alternative approaches to computing the transpose of a directed graph:

Manual Transpose Implementation:

One alternative is for users to implement the transpose operation manually using the existing functionality of SageMath's DiGraph class. Users can iterate through the edges of the original graph, reverse the direction of each edge, and construct a new graph to represent the transpose. While this approach provides flexibility, it may not be as efficient or user-friendly as having a dedicated function.

Graph Transformation:

Another alternative is to use graph transformation libraries or tools external to SageMath. Various graph manipulation libraries in other programming languages, such as NetworkX in Python, provide functionality to compute the transpose of directed graphs. Users could export their graph from SageMath, perform the transpose operation in an external tool, and then import the result back into SageMath. This approach can be useful when the transpose operation is part of a more extensive graph manipulation pipeline.

### Additional Information

_No response_

saatvikraoIITGN:
### Description

This PR introduces a new function, `transpose_graph`, which enables the computation of the transpose of a directed graph within the SageMath graph library. The transpose of a directed graph retains the same vertices but reverses the direction of each edge.

### Why is this change required?

The change is necessary to address the absence of a built-in method for calculating the transpose of a graph in SageMath. The transpose operation is fundamental in graph theory and graph algorithms, and having an efficient implementation simplifies various applications that require this operation.

This PR aims to solve this problem by providing a dedicated and efficient function for calculating the transpose of a directed graph.

Fixes #36583 

### Checklist

- [x] I have linked a relevant issue or discussion (if applicable).

### Dependencies

#36583 : to implement the function for the transpose of a graph

### DIFF
--- a/src/sage/graphs/graph.py
+++ b/src/sage/graphs/graph.py
@@ -4890,6 +4890,55 @@ class Graph(GenericGraph):
 
         return repr
 
+    @doc_index("Basic methods")
+    def transpose_of_graph(self):
+        r"""
+        Return the transpose of a directed graph G.
+        
+        The transpose of a directed graph has the same vertices, but the direction
+        of each edge is reversed.
+
+        INPUT:
+        - G: A directed graph (DiGraph).
+
+        OUTPUT:
+        - The transpose of the graph G.
+
+        EXAMPLES::
+
+            sage: G = DiGraph([(1, 2), (2, 3), (3, 1)])
+            sage: GT = transpose_graph(G)
+            sage: GT.edges()
+            [(2, 1), (3, 2), (1, 3)]
+
+            sage: G = DiGraph([(1, 2), (2, 2), (3, 3), (3, 1)])
+            sage: GT = transpose_graph(G)
+            sage: GT.edges()
+            [(2, 1), (2, 2), (3, 3), (1, 3)]
+
+            sage: G = DiGraph()
+            sage: GT = transpose_graph(G)
+            sage: GT.edges()
+            []
+
+        """
+        self._scream_if_not_simple()  # Ensure G is a simple graph.
+
+        GT = self.copy(data_structure='dense')  # Create a copy of G with dense data structure.
+        
+        for u, v, label in self.edges():
+            GT.delete_edge(u, v, label)  # Remove the original edge.
+            GT.add_edge(v, u, label)     # Add the reversed edge.
+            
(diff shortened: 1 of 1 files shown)


dcoudert:
We already have method `reverse` which returns a copy of a digraph with edges reversed in direction.

dcoudert:
So I propose to close this PR and the corresponding issue.

PR #36228, opened 2023-09-09
Use Cython 3 implicit namespace package support

mkoeppe:
Making `sage.misc.package_dir.cython_namespace_package_support` a no-op for Cython 3.

Keeping the code for Cython 0.x around for use in downstream packages.

- Cherry-picked from #35095 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36207 (merged here) 
- Depends on #36110 (merged here)
- Depends on #36573 (need to rebase on top)
- Depends on the changes in dfceb98478d, 7d042415e05, ef2f651caa1 in #35095
- Depends on https://github.com/sagemath/sage/pull/36594 (cherry-picked from above)
- Depends on https://github.com/sagemath/sage/pull/36595 (cherry-picked from above)
- Depends on https://github.com/sagemath/sage/pull/36596 (cherry-picked from above)
- Depends on #36605
- Depends on https://github.com/sagemath/sage/pull/36612 (cherry-picked from above)
- Depends on #36666 (cherry-picked from above, merged here)
- Depends on #36572 (dup)
- Depends on #36588 (dup)
- Depends on #36589 (dup)

PR #36594, opened 2023-10-30
Replace relative imports by absolute ones in `sage.{matrix,matroids,modules,stats}`

mkoeppe:
- Cherry-picked from #35095
- This is for #36228

### :memo: Checklist

### DIFF
--- a/src/sage/matrix/matrix2.pyx
+++ b/src/sage/matrix/matrix2.pyx
@@ -99,7 +99,7 @@ from sage.arith.numerical_approx cimport digits_to_bits
 from copy import copy
 
 import sage.modules.free_module
-from . import berlekamp_massey
+from sage.matrix import berlekamp_massey
 from sage.modules.free_module_element import is_FreeModuleElement
 from sage.matrix.matrix_misc import permanental_minor_polynomial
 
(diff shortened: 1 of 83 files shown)


mkoeppe:
The pyright error is unrelated. Ready for review

src/sage/matrix/matrix2.pyx
```diff
@@ -99,7 +99,7 @@ from sage.arith.numerical_approx cimport digits_to_bits
 from copy import copy
 
 import sage.modules.free_module
-from . import berlekamp_massey
```
dcoudert, line 102: here we used `from . import berlekamp_massey` to import a module and it is replaced by `from sage.matrix import berlekamp_massey`. But in other places, for instance in `src/sage/matroids/unpickling.pyx`, we use `import sage.matroids.matroid` (not modified). I'm not sure of the difference between both forms. Which one is the most suited ?

dcoudert (commented):
otherwise LGTM and passes tests on Fedora 35.

src/sage/matrix/matrix2.pyx
mkoeppe: After `from sage.matrix import berlekamp_massey`, the module can be referred to as `berlekamp_massey`. This style tends to appear in older portions of the Sage source code. Personally I don't like it very much, but I have not made an effort to change it.

dcoudert: I see.

dcoudert (approved):
LGTM.

mkoeppe:
Thank you!

PR #36337, opened 2023-09-25
In doctesting, ignore two ld warnings from recent OS X:

jhpalmieri:
in doctesting, ignore two warnings coming from ld in new OS X/command-line tools/Xcode 15

As of Xcode 15, I see warnings of this sort arising in doctesting:
```
ld: warning: duplicate -rpath '/path/to/SAGE_ROOT/local/lib' ignored
ld: warning: ignoring duplicate libraries: '-lflint', '-lgmp', '-lmpfr'
```
These all seem to arise from doctests involving `cython(...)`.

These warnings cause doctests to fail, so we ignore them, as we ignore similar warning messages in src/doctest/parsing.py.

### :memo: Checklist

- [X] The title is concise, informative, and self-explanatory.
- [X] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36523, opened 2023-10-23
`scipy`: Fix build on macOS with Xcode CLT 15.0

mkoeppe:
- Fixes #36342 
- using the fix discussed at https://github.com/mesonbuild/meson/issues/12282

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36599, opened 2023-10-30
sage-env: identify the version of command-line tools (OS X) and set LDFLAGS accordingly

### LINKED ISSUE #36342: Build failures with Xcode CLT 15.0: scipy (via meson), ecm
SciPy build failure when homebrew's gfortran is used (via meson), reported in https://groups.google.com/g/sage-release/c/1wOBmhvNJqc/m/dT-6HGAjCAAJ
- see also https://github.com/mesonbuild/meson/issues/12282

ECM build failure reported in https://groups.google.com/g/sage-release/c/dvPti2UkyjQ/m/Nj4e5zApCQAJ:

```
 checking what assembly label suffix to use... :
checking if globals are prefixed by underscore... configure: error: Test program links neither with nor without underscore.
********************************************************************************
Error configuring ecm-7.0.5 See the file
    /Users/iohara/Desktop/sage-10.2.beta4/local/var/tmp/sage/build/ecm-7.0.5/src/config.log
for details.
```
This looks similar to https://gitlab.inria.fr/zimmerma/ecm/-/issues/21856 @dimpase @zimmermann6 

Our latest ECM upgrade was made in https://github.com/sagemath/sage/issues/31325.

jhpalmieri:
Identify the version of command-line tools in use (OS X), and use the version to set LDFLAGS.

In sage-env, save the version of command-line tools being used in XCLT_VERSION. If it's too recent, we set LDFLAGS to solve some problems:
- Fixes #36337 (doctest failures)
- Fixes https://github.com/sagemath/sage/issues/36342 (scipy building, replacing #36523).

### :memo: Checklist


jhpalmieri:
In a perfect world, we would document variables like this somewhere in the developer's guide...

jhpalmieri:
We could also set LDFLAGS depending on XCLT_VERSION, either here or on another PR. (See https://github.com/sagemath/sage/pull/36337#issuecomment-1741485061.)

jhpalmieri:
With this change, `scipy` builds and I don't get the "ignoring duplicate libraries" messages in the sagelib log or in doctests.

jhpalmieri:
I only have access to OS X machines with the latest command-line tools, so I can't test as thoroughly as I would like.

src/bin/sage-env
```diff
@@ -277,6 +277,15 @@ export UNAME=`uname | sed 's/CYGWIN.*/CYGWIN/' `
 # Mac OS X-specific setup
 if [ "$UNAME" = "Darwin" ]; then
     export MACOSX_VERSION=`uname -r | awk -F. '{print $1}'`
+    # Try to identify command-line tools version.
+    # See https://apple.stackexchange.com/q/180957/351985.
+    XCLT_VERSION=`pkgutil --pkg-info=com.apple.pkg.CLTools_Executables | awk '/version: / { print $NF }' | cut -d. -f-1`
```
mkoeppe: I'm not sure if querying with `pkgutil` is the right approach. 
I think on a system several versions of CLT (and separately several versions of Xcode) can be installed.

mkoeppe:
I think it may be better to just check whether `$(xcode-select -p)/usr/bin/ld-classic` exists.

jhpalmieri:
> I think it may be better to just check whether `$(xcode-select -p)/usr/bin/ld-classic` exists.

Like this?
```
    if [ -x "$(xcode-select -p)/usr/bin/ld-classic" ] ; then
        CLT_LD_CLASSIC=yes
    else
        CLT_LD_CLASSIC=no
    fi
    export CLT_LD_CLASSIC
```
(Might as well test whether it's executable, right?)

jhpalmieri:
> > I think it may be better to just check whether `$(xcode-select -p)/usr/bin/ld-classic` exists.
> 
> Like this?
> 
> ```
>     if [ -x "$(xcode-select -p)/usr/bin/ld-classic" ] ; then
>         CLT_LD_CLASSIC=yes
>     else
>         CLT_LD_CLASSIC=no
>     fi
>     export CLT_LD_CLASSIC
> ```
> 
> (Might as well test whether it's executable, right?)

(Or move this check later when setting LD_FLAGS, and don't define a variable to track it — that's a little cleaner, if we're just using this to set LD_FLAGS.)

mkoeppe:
Yes, looks good to me

jhpalmieri:
Okay, here's a new version.

mkoeppe (approved):

jhpalmieri:
Thanks!

PR #36600, opened 2023-10-31
Support giac 1.9.0-67

tornaria:
The function `gb_giac()` is less noisy, so we adjust doctests in a way they accept the noisy version (old giac) but don't fail with the silent version (new giac).

### :memo: Checklist


orlitzky:
Running testlong now, back in about 24h.

antonio-rojas:
LGTM

kiwifb:
Does LGTM too. It is nice that giac is less chatty when used as a library. Still need to get those blank lines out though.

orlitzky (approved):
I added giac-1.9.0.67 to Gentoo and ran the full test suite. This does fix the only new issues, thanks.

PR #36604, opened 2023-10-31
Calculation of Maximum Leaf Number graph parameter

### LINKED ISSUE #36603: Maximum leaf number parameter
### Problem Description

The Max Leaf Number (MLN) represents the maximum number of leaf nodes in any spanning tree of a graph G. It is an incredibly useful graph parameter, particularly when constructing wireless sensor networks. It is also a minimal upper bound for both bandwidth and distance to linear forest.

Currently, SageMath does not provide a function to calculate the MLN. However, now that Minimum Connected Dominating Set (MCDS) is implemented, such a function can be easily added.

### Proposed Solution

The MLN of a graph can be calculated by generating a MCDS. This is because the MLN is the size of the difference set between the graph and the MCDS (or equivalently, the order of the graph minus the size of the dominating set).

### Alternatives Considered

The MLN could also be calculated through generating a Maximum Leaf Spanning Tree; however, this would require implementing an algorithm from scratch and has no time complexity advantage.

### Additional Information

_No response_

MrBanananator:
Added function to calculate the Max Leaf Number graph parameter. 

Fixes #36603 


src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,45 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
         return list(dom)
+
+
+# ==============================================================================
+# Maximum leaf number
+# ==============================================================================
+
+def max_leaf_number(G):
+    r"""
+    Return the maximum leaf number of the graph, i.e. the maximum possible number
```
dcoudert: Could you change to:
```
    Return the maximum leaf number of the graph.

    The maximum leaf number is the maximum possible number of leaves of a
    spanning tree of `G`. This is also the complement of the caridnality of a
    minimum connected dominating set.
    See the :wikipedia:`Connected_dominating_set`.
```

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,45 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
         return dom
     else:
         return list(dom)
+
+
+# ==============================================================================
+# Maximum leaf number
+# ==============================================================================
+
+def max_leaf_number(G):
```
dcoudert: add input parameters `solver`, `verbose` and `integrality_tolerance` as documented in method `dominating_set`.

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,45 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+
+    Petersen graph::
+
+        sage: G = graphs.PetersenGraph()
+        sage: G.max_leaf_number()
+        6
+
+    Unconnected graph::
+
+        sage: G = 2 * graphs.PetersenGraph()
```
dcoudert: You can simply try with `Graph(2)`.

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,45 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+        sage: G.max_leaf_number()
+        0
+
+    Petersen graph::
+
+        sage: G = graphs.PetersenGraph()
+        sage: G.max_leaf_number()
+        6
+
+    Unconnected graph::
```
dcoudert, line 1349: This could be the `TESTS::` block.

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,45 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+
+        sage: G = 2 * graphs.PetersenGraph()
+        sage: G.max_leaf_number()
+        Traceback (most recent call last):
+        ...
+        ValueError: the graph must be connected
+    """
+    if G.order() < 2:
```
dcoudert: may be you should document more explicitly the expected value for a graph with 0, 1 or 2 vertices.

src/sage/graphs/domination.py
dcoudert: it might be better to use `maximum_leaf_number`.

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,45 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
         return dom
     else:
         return list(dom)
+
+
+# ==============================================================================
```
dcoudert: we don't need these lines. We already have the definition of the function.

dcoudert (changes requested):
Thank you for this addition. I have a few comments for improving the method.

src/sage/graphs/domination.py
DaveWitteMorris: Instead of "the complement of the caridnality", this needs to be "the cardinality of the complement". (In addition to changing the order, notice that there is a typo in "caridnality".)

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,78 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+    Unconnected graph::
+
+        sage: G = Graph(2)
+        sage: G.maximum_leaf_number()
+        Traceback (most recent call last):
+        ...
+        ValueError: the graph must be connected
+    """
+    # The MLN of a graph with less than 2 vertices is 0, while the
+    # MLN of a connected graph with 2 or 3 vertices is 1 or 2 respectively.
```
dcoudert: This is not consistant with the examples above. Furthermore, these assumptions should be placed before the `INPUT:` statement so that users don't have to look at the code to know it.

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,78 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+    if G.order() <= 1:
+        return 0
+    elif not G.is_connected():
```
dcoudert: you can use `if` instead of `elif` here. It seems better... but I'm not sure why ;)

dcoudert (changes requested):

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,79 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+        sage: G.maximum_leaf_number()
+        0
+
+    Petersen graph::
+
+        sage: G = graphs.PetersenGraph()
+        sage: G.maximum_leaf_number()
+        6
+
+    Tests:
```
dcoudert: this keyword should be in capital letters, so `TESTS:`.

src/sage/graphs/domination.py
```diff
@@ -1284,3 +1285,79 @@ def greedy_dominating_set(G, k=1, vertices=None, ordering=None, return_sets=Fals
...
+        ...
+        ValueError: the graph must be connected
+    """
+    if G.order() <= 1:
+        return 0
+    if not G.is_connected():
+        raise ValueError('the graph must be connected')
+    if G.order() <= 3:
+        return G.order() - 1
+    return G.order() - dominating_set(G, connected=True, value_only=True, solver=solver, verbose=verbose, integrality_tolerance=integrality_tolerance)
```
dcoudert: can you use a form like
```
    return G.order() - dominating_set(G, connected=True, value_only=True,
    	                              solver=solver, verbose=verbose,
				      integrality_tolerance=integrality_tolerance)
```
This is for the coding style.

dcoudert (changes requested):
minor changes needed. Otherwise, it passes all tests.

dcoudert (approved):
LGTM. Thank you for adding this method.

PR #36609, opened 2023-10-31
avoid using `itertools.pairwise`

dcoudert:
Method `pairwise` has been introduced in `itertools` version 3.10. As reported in #36493,  its use breaks the condo workflow with Python 3.9. So we modify the code to avoid using `pairwise`.

### :memo: Checklist

### :hourglass: Dependencies
See discussion https://github.com/sagemath/sage/pull/36493#issuecomment-1786350920,


tobiasdiez (approved):
LGTM, thanks a lot for the quick follow-up!

I'll wait with setting it to positive review until the ci workflows are green, but already set it to "blocker" so that the fix applies to other PRs as well.

PR #36572, opened 2023-10-29
Replace relative imports by absolute ones in categories

tobiasdiez:
As preparation for compiling with meson, we replace relative imports by absolute ones.
Relative imports are not used consistently in the codebase and result in issues for doctesting with pytest (which admittedly is a limitation of pytest). We normalize the relative imports in `sage.categories` to be absolute imports. Small code-style improvements along the way (mainly to nicely order the imports)

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36430, opened 2023-10-09
CI Linux: Run `ubuntu-focal-standard` without waiting for other platforms

mkoeppe:
This is so that the image for "build.yml" etc. is ready faster.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36610, opened 2023-10-31
Limit max number of parallel ci runs for docker-based runs

tobiasdiez:
Currently, the ci-linux workflow runs a lot of parallel runs which prevents other workflows (from PR) to be executed. We add a few more `max_parallel` statements to spread the stress on the ci over a longer time.

### :memo: Checklist


.github/workflows/ci-linux.yml
```diff
@@ -101,6 +102,7 @@ jobs:
       tox_packages_factors: >-
           ["standard"]
       docker_push_repository: ghcr.io/${{ github.repository }}/
+      max_parallel: 12
```
kwankyu, line 87: I guess you picked up these numbers heuristically.  But what is the reason that 12 for "standard" is less than 15 for "standard-pre"?

tobiasdiez: The heuristic was that "standard-pre" is rather important as you want to be notified early that there are profound compilation issues on a certain system. "standard" is a bit less important (and results seem to be more uniform across the systems tested), also there are a few systems dropping out as their pre step failed.

mkoeppe: Note that the workflows do not fail when there are doctest failures. 
It's crucial for "standard" to be completed as quickly as possible so that developers can inspect the results.

mkoeppe: Also note that there is an overall walltime limit of 3 days (?) for the whole workflow to pass. So whatever job restrictions we put in, we need to make sure that this does not cause the later workflows to never run.

mkoeppe:
From the moment that a release has been pushed and the merged PRs are closed:
- PR workflows still run with the previous releases's Docker image.
- For PRs that have already been rebased on the new release, this will be slow as it has to rebuild all the changes in the new release.
- For PRs that have not been rebased, the blocker PRs just closed are no longer applied, so failures already fixed there will show up again.
This is the situation until the "default-pre" and "default" jobs have completed and pushed the new Docker image.

So what we should do is ensure that this new Docker image is available as soon as possible. #36430, by making "default-pre" and "default" jobs separate from "standard-pre" and "standard", already did part of this, but we can do better: Currently, because "default" is queued after the "standard-pre" jobs, it has to wait until a runner becomes available.

So I would suggest the following:
- standard-pre: *increase* the `max-parallel` from 30 (the default set in docker.yml) to 50 so that all standard-pre jobs can start at the same time

.github/workflows/ci-linux.yml
tobiasdiez: There is nothing crucial about these workflow runs. If they fail, then its just an ordinary bug that needs to be fixed. There is no need to have them finished as soon as possible. You can always remove some older/similar systems from the ci if things are then taking too long.

In my opinion it's not acceptable that other workflows are queued for 14+ hours because of it.

mkoeppe: > There is nothing crucial about these workflow runs. If they fail, then its just an ordinary bug that needs to be fixed. There is no need to have them finished as soon as possible.

Not sure for what audience you're writing this, Tobias...

kwankyu: > Also note that there is an overall walltime limit of 3 days (?) for the whole workflow to pass. So whatever job restrictions we put in, we need to make sure that this does not cause the later workflows to never run.

That is 35 days according to https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration. So no problem here.

mkoeppe: Thanks for checking; then the time to the release (~ 7 days) is the closer deadline

kwankyu: Yes.

Another question. What is the limit on the number of concurrent jobs? The linked page says that it depends on our plan. If we are on "Free" plan, it is 20 concurrent jobs. Is this the case?

kwankyu: If it is 20, then even though we increase "max-parallel" to 50, only 20 of them will be run together in parallel. Am I right?

mkoeppe: The GitHub people gave us the Team plan for free, so we have 60 concurrent Linux runners (and 5 macOS runners).

kwankyu: We have 45 systems in ci-linux. So even though we increase "max-parallel" to 50 for "standard-pre", "default" may not run before two passes of bundle of 20 parallel runs, unless "default" is lucky.

kwankyu: Ah. 60 runners. Then increasing "max-parallel" to 50 is a good solution!

Also we may decrease "max-parallel" for "minimal", "maximal", "experimental" to allow rooms for other PR workflow runs.

mkoeppe: Sure. When I added the "standard-sitepackages" run (with a limit of 10 and running parallel with "standard"), I didn't decrease the max-parallel of "standard". We can reduce it a bit, from 30 perhaps to 25. I wouldn't go much lower than that.

kwankyu: How about basing your PR on Tobias's?

mkoeppe: I've instead pushed a small change to https://github.com/sagemath/sage/pull/36616#issuecomment-1789410410

mkoeppe:
Let's close this one here in favor of #36616.

tobiasdiez:
> #36616 adjusted "standard" reasonably.

I don't agree that this is enough reduction.

> Decreasing max-parallel for "minimal" and others will make more rooms for other PRs. But as we have 60 runners, max-parallel 30 (current default) for them already gives enough room for other PRs. On the other hand, decreasing max-parallel for "minimal" and others will result in taking more time to finish "minimal" and others.

Sadly not true. Already during the first hours after a new release you have a long queue. Normally the conda workflow takes 2 hours, but as you see from https://github.com/sagemath/sage/actions/workflows/ci-conda.yml?page=4 and page 3 there are a lot of workflows that take 5 hours, so they are queued for 3h. Some are even in the queue for 12 hours, and that's usually the case when one of the "blocks" in the ci-linux workflow is finished so that the next block gets almost all runners.

.github/workflows/ci-linux.yml
tobiasdiez: > Thanks for checking; then the time to the release (~ 7 days) is the closer deadline

According to https://github.com/sagemath/sage/actions/runs/6598197147/usage, the overall run time is 28d, so an average number of 5 runners should be more than enough to make that deadline. With the proposed 10-15 runners, you get it in 3 to 4 days.

kwankyu:
OK. Then we would need decreasing max-parallel for "minimal" and others in addition to #36616.

If you rebase your PR on #36616 and make adjustments only to "minimal" and others, I am willing to get this PR in for the next beta.

mkoeppe:
I don't think this PR is a serious proposal. If Tobias wants to demonstrate that after this radical cut of number of allotted runners by a factor 3 is suitable for running the portability tests, he can do so by running it in his own repo so we can have a look at the run.

tobiasdiez:
> I don't think this PR is a serious proposal.

Please stop these comments. This PR is clearly not an April fools joke.

> If Tobias wants to demonstrate that after this radical cut of number of allotted runners by a factor 3 is suitable for running the portability tests, he can do so by running it in his own repo so we can have a look at the run.

I don't want to demonstrate that this is possible, I want to reduce the completion time of PR-related workflows after a release.
Anyway, here is the run: https://github.com/tobiasdiez/sage/actions/runs/6831315056 (standard will finish about 16h after the run is started; not sure what other information you expect to get out of the test run).

mkoeppe:
> (standard will finish about 16h after the run is started; not sure what other information you expect to get out of the test run).

The other tests that are part of the CI Linux.

mkoeppe:
> Anyway, here is the run: https://github.com/tobiasdiez/sage/actions/runs/6831315056 (standard will finish about 16h after the run is started

Well, it has finished. Total duration 2d 15h. And that's without really running the optional + experimental workflows (most failed instantly; this is repaired in #36667).

As I expected -- not suitable for merging.

tobiasdiez:
> > Anyway, here is the run: https://github.com/tobiasdiez/sage/actions/runs/6831315056 (standard will finish about 16h after the run is started
> 
> Well, it has finished. Total duration 2d 15h. And that's without really running the optional + experimental workflows (most failed instantly; this is repaired in #36667).
> 
> As I expected -- not suitable for merging.

Can you expand on why this is too slow? It is clearly in the 7d timelimit you set in https://github.com/sagemath/sage/pull/36610#discussion_r1378408422. What would be an acceptable time for you?

mkoeppe:
> It is clearly in the 7d timelimit you set in [#36610 (comment)](https://github.com/sagemath/sage/pull/36610#discussion_r1378408422).

No, Tobias, I didn't "set" this timelimit.

tobiasdiez:
> > It is clearly in the 7d timelimit you set in [#36610 (comment)](https://github.com/sagemath/sage/pull/36610#discussion_r1378408422).
> 
> No, Tobias, I didn't "set" this timelimit.

Okay, then you 'said' it. Whatever, what about my other questions?

 > Can you expand on why this is too slow? [...] What would be an acceptable time for you?

mkoeppe:
Proposing to close this PR. As I said in https://github.com/sagemath/sage/pull/36610#issuecomment-1806621316, this PR is not a serious proposal. 
From the beginning, it has lacked a justification for the radical changes that it makes. Instead, the author demands that reviewers justify the status quo.

tobiasdiez:
There is nothing radical about this PR. It just gives priority to the PR workflows instead of the ci-linux, by reducing the priority of the later by roughly a factor of 2 (relative to the number of runners assigned at the point of creation of this PR).

tobiasdiez:
Maybe start with answering the following question instead of insulting me:
> Can you expand on why this is too slow? [...] What would be an acceptable time for you?

tobiasdiez:
> For reference: [#36572 (comment)](https://github.com/sagemath/sage/pull/36572#issuecomment-1803044465)

I fail to see how this answers my questions.

mkoeppe:
As discussed in https://github.com/sagemath/sage/pull/36726#issuecomment-1847870000

tobiasdiez:
That doesn't address my question at all.

mkoeppe:
@kcrisman responding to your suggestion in https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/ciVrZ7x0AQAJ:

Confirming my vote of -1 on this PR.

kcrisman:
> That doesn't address my question at all.

Perhaps surprisingly, here I agree more that it's reasonable to ask for a concrete number and have discussion about that before closing the ticket.  See the end of my comment here: https://github.com/sagemath/sage/pull/36725#issuecomment-2004330310  Am I correct in inferring that some of the CI limit tickets are, at their core, about a disagreement in philosophy on whether PRs or "standard" builds should get priority?  I'm a little in the dark here (hence my initial reluctance to get drawn in ...). 

But anyway, I feel like that is all a sideshow.  It seems like this should be a number that people can come to **some** consensus on.  Can we at least agree on a hard upper and lower bound imposed by hardware/price/something?  (And then maybe make a random choice based on some interesting distribution ... I'm totally not joking, randomized elections of this type are definitely proposed from time to time, and maybe it will help end the bad feelings here a *tiny* bit.)

jhpalmieri:
-1 from me

PR #36630, opened 2023-11-02
Allow to specify output directory for generated interpreters

tobiasdiez:
Add the possibility to specify the output directory where the generated interpreters files will be created. This is preparation for the compilation with meson.

### :memo: Checklist


src/sage_setup/autogen/interpreters/__main__.py
```diff
@@ -1,9 +1,18 @@
-from __future__ import print_function, absolute_import
+from __future__ import absolute_import, print_function
```
mkoeppe: this line can be deleted, it's no longer needed with python 3.

src/sage_setup/autogen/interpreters/__main__.py
```diff
@@ -1,9 +1,18 @@
...
 
+import argparse
 import os
-from sage.env import SAGE_SRC
 
 from . import rebuild
 
+parser = argparse.ArgumentParser()
+parser.add_argument("output_dir")
```
mkoeppe: add a help string?

tobiasdiez: I don't know what to write that would be more explanatory than "output directory"

src/sage_setup/autogen/interpreters/__main__.py
tobiasdiez: Done

src/sage_setup/autogen/interpreters/__main__.py
```diff
@@ -1,9 +1,16 @@
...
+args = parser.parse_args()
+
+output_dir = args.output_dir
+if not output_dir:
+    from sage.env import SAGE_SRC
+    output_dir = os.path.join(SAGE_SRC, "sage", "ext", "interpreters")
```
orlitzky: Does anything else invoke this via `__main__`? If not I'd recommend dropping the default and making the output directory required. Otherwise we risk someone complaining about breaking the existing behavior if we later try to remove the dead code.

It'd also be nice to have a short docstring at the top explaining how you'd run this via `python -m ...` and why the variable output directory is useful to the meson build.

tobiasdiez: Good point, done. (The fallback was also not working before, since positional args are always required)

Since meson's philosophy is to have a "out-of-source" build, we need to create the interpreters in a location specified by meson.

orlitzky:
I think this is OK now, thanks.

tobiasdiez:
Thank you for the review and the good suggestions!

PR #36635, opened 2023-11-03
clean the file hexad.py

fchapoton:
various little pep8 details, some fixes by ruff, unicode art, etc

### :memo: Checklist

### DIFF
--- a/src/sage/games/hexad.py
+++ b/src/sage/games/hexad.py
@@ -107,7 +107,7 @@ def view_list(L):
     return matrix(GF(2), 3, 3, lambda x, y: 1 if (x, y) in L else 0)
 
 
-def picture_set(A, L):
+def picture_set(A, L) -> set:
     """
     This is needed in the :meth:`Minimog.find_hexad` function below.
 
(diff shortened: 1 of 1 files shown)


src/sage/games/hexad.py
```diff
@@ -545,77 +543,77 @@ def find_hexad(self, pts):
...
             return H, WHAT
         if len(L2) == 2:  # type 0 or 3
             if (MINIMOG[0][2] in LL and MINIMOG[2][1] in LL):
-                H, WHAT = self.find_hexad3(LL - set([MINIMOG[0][2],MINIMOG[2][1]]),
+                H, WHAT = self.find_hexad3(LL - {MINIMOG[0][2], MINIMOG[2][1]},
                                            MINIMOG[0][2], MINIMOG[2][1])
-                if H != []:   # must be type 3
+                if H:   # must be type 3
                     return list(H), WHAT
                 if H == []:   # could be type 0
```
dcoudert: This could be `if not H:`, or we can remove this test as we checked `if H:` 2 lines before.

The same situation appears multiple times below.

dcoudert (commented):
otherwise LGTM.

src/sage/games/hexad.py
fchapoton: voila, merci pour la relecture

dcoudert (approved):
LGTM. Passes tests on Fedora 35.

PR #36601, opened 2023-10-31
Doc preview for all languages

kwankyu:
Currently our doc preview for PR is uploading only English docs. We fix it for all languages.

Along the way, we remove more spurious diffs in CHANGES.

Note that the doc preview for this PR is somewhat broken. But this would be ironed out in the next beta.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36617, opened 2023-10-31
CI: Make jobs more responsive to canceling

mkoeppe:
... by replacing `always()` by `success() || failure()`, except for steps such as uploading / printing out logs or similar artifacts that have already been built.

Even upon canceling a workflow (manually or automatically when a new commit has been pushed to the same branch), a new step that uses `if: always() ....` will still be started, which can clog the GH Actions pipeline.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36614 (merged here to remove merge conflict)

PR #36614, opened 2023-10-31
.github/workflows/doc-build-pdf.yml: Do not build HTML documentation, fix upload of PDFs

mkoeppe:
Our `make` target `doc-pdf` depends on `doc-html` only to avoid concurrency bugs when updating the inventory files.

To speed up the Build Documentation (PDF) workflow, we skip the HTML build.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36601 (to resolve merge conflict, merged here)

PR #36636, opened 2023-11-03
Deploy live doc on push to develop 

kwankyu:
Live doc is built separately from the usual doc preview on push to `develop`, and deployed to Netlify.

Sample deployment: https://deploy-livedoc--sagemath.netlify.app/html/en/a_tour_of_sage/

### :memo: Checklist

### :hourglass: Dependencies

- #36601

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -200,7 +200,7 @@ jobs:
 
       - name: Build live doc
         id: buildlivedoc
-        if: always() && steps.copy.outcome == 'success' && github.ref == 'refs/heads/develop'
+        if: (success() || failure()) && steps.copy.outcome == 'success' && github.repository == 'sagemath/sage' && github.ref == 'refs/heads/develop'
         run: |
           set -ex
           export SAGE_USE_CDNS=yes
@@ -215,7 +215,7 @@ jobs:
 
       - name: Copy live doc
         id: copylivedoc
-        if: always() && steps.buildlivedoc.outcome == 'success'
+        if: (success() || failure()) && steps.buildlivedoc.outcome == 'success'
         run: |
           set -ex
           mkdir -p ./livedoc
@@ -224,7 +224,7 @@ jobs:
           zip -r livedoc.zip livedoc
 
       - name: Upload live doc
-        if: always() && steps.copylivedoc.outcome == 'success'
+        if: (success() || failure()) && steps.copylivedoc.outcome == 'success'
         uses: actions/upload-artifact@v3
         with:
           name: livedoc
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -63,12 +63,12 @@ jobs:
           mathjax_path_to=$(SAGE_USE_CDNS=yes /sage/sage -python -c "from sage_docbuild.conf import mathjax_path; print(mathjax_path)")
           new_version=$(cat src/VERSION.txt)
           # Wipe out chronic diffs between old doc and new doc
-          (cd /sage/local/share/doc/sage/html/en && \
+          (cd /sage/local/share/doc/sage/html && \
            find . -name "*.html" | xargs sed -i -e '/class="sidebar-brand-text"/ s/Sage [0-9a-z.]* /Sage '"$new_version"' /' \
                                                 -e 's;'"$mathjax_path_from"';'"$mathjax_path_to"';' \
                                                 -e '\;<script type="application/vnd\.jupyter\.widget-state+json">;,\;</script>; d')
           # Create git repo from old doc
-          (cd /sage/local/share/doc/sage/html/en && \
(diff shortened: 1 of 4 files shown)


.github/workflows/doc-build.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           path: upstream
           name: upstream
 
-  build-docs:
```
mkoeppe: are these renames really necessary?

kwankyu: No.. Just cosmetic changes. Does it incur merge conflict?

kwankyu: I will revert if that causes problems.

mkoeppe: Yes, likely with #36508, #36614, #36498...

kwankyu: Let me know.

kwankyu: Sorry. Reverted.

mkoeppe: Thank you! Also makes it easier to look through the changes~

kwankyu: Checking that I didn't break anything while reverting, in

https://github.com/kwankyu/sage/actions/runs/6750354430

and doc-publish workflow (pending to run). Not likely.

.github/workflows/doc-build.yml
```diff
@@ -196,3 +197,36 @@ jobs:
         with:
           name: docs
           path: docs.zip
+
+      - name: Build live doc
+        id: buildlivedoc
+        if: always() && steps.copy.outcome == 'success' && github.ref == 'refs/heads/develop'
```
mkoeppe: Perhaps should also conditionalize on `github.repository == 'sagemath/sage'` so that it does not run in users' forks if they push a change to their develop (instead of creating a specific branch)

kwankyu: OK. Thanks.

kwankyu: Do I need `always()` for "Build live doc", "Copy live doc" and "Upload live doc" steps? Perhaps yes.

mkoeppe: It's better to use `success() || failure()`, see #36617

kwankyu: Pending to commit the change `github.repository == 'sagemath/sage` until https://github.com/sagemath/sage/pull/36636#discussion_r1382224796 finishes on my own repo.

kwankyu: > It's better to use `success() || failure()`, see #36617

OK. 

> Even upon canceling a workflow (manually or automatically when a new commit has been pushed to the same branch), a new step that uses if: always() .... will still be started

But this seems to be a bug on github side.

.github/workflows/doc-build.yml
kwankyu: Checked.

.github/workflows/doc-build.yml
kwankyu: Done.

mkoeppe: > But this seems to be a bug on github side.

No, this behavior is by design, but one needs to use it sparingly. Essentially only for releasing resources or for saving logs.

mkoeppe:
Which version of Sage does the binder environment run? Is it always running the current develop version? Is there a mechanism that triggers the rebuild?

kwankyu:
The default branch (master) runs the latest release. The dev branch runs the latest develop version.

Both branches are updated everyday.

Thus the deployed live doc runs on the latest release of Sage.

.github/workflows/doc-build.yml
kwankyu: I see. Thanks.

mkoeppe:
Thanks. Would this be a good moment to revisit:
-  #33862

kwankyu:
You mean including sage version in `NETLIFY_ALIAS` so that the url of the deployed doc include the version number?

That has one problem. Currently the url of the deployed doc for beta releases does not change and we are relying on the fact in the doc links in the github wiki. A solution would be to deploy the doc twice: once with say `deploy-release-10.2.beta9--...` and once with say `deploy-release-latest--...`

I am not sure if this is a good time to revisit #33862 here. As doc build workflows are now touched by a few PRs, I think it is wise to postpone further touch to them.

mkoeppe (approved):
LGTM.

.github/workflows/doc-build.yml
mkoeppe: > Pending to commit the change `github.repository == 'sagemath/sage`

When done, please go ahead and set the label "positive review".

kwankyu:
Thank you!

.github/workflows/doc-build.yml
kwankyu: Done. Thanks.

PR #35546, opened 2023-04-21
compute traces of elliptic-curve endomorphisms

yyyyx4:
In this patch, we add a generic method to compute the trace of an elliptic-curve endomorphism.

The computation uses a variant of Schoof's algorithm, which relies (only) on evaluating the endomorphism on small-order points, hence the method is totally independent of the internal representation of the `EllipticCurveHom`.

Adding this method is yet another step towards implementing endomorphism rings, especially for supporting the quaternionic (supersingular) case. I'm sure some speedups are possible, but they can come later.

PR #36637, opened 2023-11-03
sums of elliptic-curve morphisms

yyyyx4:
Here we add a new `EllipticCurveHom` child class representing formal sums of elliptic-curve morphisms with the same domain and codomain. One of the main features is a conversion function from formal sums to a more explicit representation of the same morphism as an isogeny chain, which will come in handy for endomorphism-ring-related computations.

In the process, we also generalize `point_of_order()` from primes to prime powers.

There is certainly a lot of potential for optimizations; some ideas are mentioned in the code. Since I think it's fair to say that this version is already much better than what we currently have for sums of isogenies (i.e., nothing), I'd suggest merging this as is and dealing with bottlenecks later as they arise.

Cc: @JohnCremona @defeo @GiacomoPope @remyoudompheng


JohnCremona:
I can review this, but not until Tuesday, sorry.  Feel free to proceed without me!

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -1359,9 +1359,10 @@ def set_order(self, value=None, *, multiple=None, check=True):
                 raise ValueError('Value %s illegal for point order' % value)
             E = self.curve()
             q = E.base_ring().cardinality()
-            low, hi = Hasse_bounds(q)
-            if value > hi:
-                raise ValueError('Value %s illegal: outside max Hasse bound' % value)
+            if q < oo:
+                _, hi = Hasse_bounds(q)
+                if value > hi:
```
JohnCremona, line 1364: The upper bound for the group order (and hence the simplest upper bound for the order of a point) is q+1+hi, not hi.  [This is an old bug, not introduced here.]

src/sage/schemes/elliptic_curves/hom_sum.py
```diff
@@ -0,0 +1,644 @@
...
+            sage: pi = E.frobenius_endomorphism()
+            sage: m1 = E.scalar_multiplication(1)
+            sage: (pi - m1)._degree_bound()
+            486
+            sage: (pi - m1).degree()
+            433
+        """
+        B = ZZ(0)
+        for phi in self._phis:
+            B += phi.degree() + 2*sqrt(B * phi.degree())
```
JohnCremona: Is this Cauchy-Schwartz? A comment would be nice. 
Would it not be better to take the integer part each time (within the loop)?  It might give a slightly better bound; more importantly, doing it as written is likely to create some nasty nested square root symbolic expressions.  Probably using isqrt() instead of sqrt) would be a good idea.

src/sage/schemes/elliptic_curves/hom_sum.py
```diff
@@ -0,0 +1,644 @@
...
+                    P = point_of_order(self._domain, l)
+                except ValueError:
+                    continue   # supersingular and l == p
+
+                Q = self.dual()._eval(self._eval(P))
+                d = discrete_log(Q, P, ord=l, operation='+')
+                deg = deg.crt(Mod(d, l))
+
+            self._degree = ZZ(deg)
+            self.dual()._degree = self._degree
```
JohnCremona, line 434: Does this line really do anything?  Doesn't it create the dual, set its degree, and then throw it away?

src/sage/schemes/elliptic_curves/hom_sum.py
```diff
@@ -0,0 +1,644 @@
...
+        deg = self.degree()
+        if deg.is_zero():
+            raise ValueError('zero morphism cannot be written as a composition of isogenies')
+
+        p = self.base_ring().characteristic()
+        insep = 0
+        scalar = 1  #TODO Can we detect scalar factors earlier to save some extensions below?
+
+        ker = []
+        for l,m in deg.factor():
```
JohnCremona, line 264: In this loop are you computing the inseparable degree of the sum?  It might be preferable to implement a method called inseparable_degree() using this sort of code and call it here?

src/sage/schemes/elliptic_curves/hom_sum.py
```diff
@@ -0,0 +1,644 @@
...
+            486
+            sage: (pi - m1).degree()
+            433
+        """
+        B = ZZ(0)
+        for phi in self._phis:
+            B += phi.degree() + 2*sqrt(B * phi.degree())
+        return integer_floor(B)
+
+    def _compute_degree(self):
```
JohnCremona, line 373: I feel sure that there must be a way to compute the degree of the sum of two morphisms; if so you caould call it repeateldy here.  But I'll need to think about that a bit more.

JohnCremona (changes requested):
I have made a few comments, at least one of which requires a change, and will think further about the question of computing deg(phi+psi) more directly.

src/sage/schemes/elliptic_curves/hom_sum.py
yyyyx4: I suppose what definitely works is the identity $\deg(\varphi+\psi) = \deg(\varphi)+\deg(\psi)+\mathrm{tr}(\varphi\widehat\psi)$, now that we have #35546. But there should be cases where computing multiple such trace pairings will certainly be slower than the current one-shot method, so this might require some thinking...

src/sage/schemes/elliptic_curves/ell_point.py
yyyyx4: Really? It looks correct to me; for example, `Hasse_bounds(101)` returns $(82,122)$.

src/sage/schemes/elliptic_curves/hom_sum.py
yyyyx4: In this class `.dual()` is marked `@cached_method`, so the same object will be returned on subsequent calls.

JohnCremona:
My stupid mistake, apologies.

JohnCremona:
Yes, that's what I was thinking of. Without experimenting I cannot tell
which would be fastest.

src/sage/schemes/elliptic_curves/hom_sum.py
yyyyx4: Excellent suggestion, this was indeed mentioned as a TODO elsewhere! Done.

src/sage/schemes/elliptic_curves/hom_sum.py
yyyyx4: Thanks, done. I also changed it to compute a lower bound in addition to the upper bound: in many cases the lower bound ends up being trivial anyway, but it could save a little bit of work sometimes.

src/sage/schemes/elliptic_curves/hom_sum.py
JohnCremona: Thanks -- that was more work than I thought it would be (as always happens), so the extra work is appreciated.

JohnCremona: Will I be very unpopular to suggest that we now also need method(s) separable_degree()?  It's only a matter of time before some user complains it is not there, so now would be a good time.  You could then change the injective() method to checking that the separable_degree is 1.

JohnCremona: More importantly, I don't think it is correct that a sum is separable iff the summands are.  For example (Frob-1)+(1).  The reverse implication also fails, e.g. (Frob)+(1).  So I think that is_separable() for sums will have to do some work.

src/sage/schemes/elliptic_curves/hom_sum.py
JohnCremona: Excellent!

src/sage/schemes/elliptic_curves/hom_sum.py
yyyyx4: Thanks, added `.separable_degree()`.

> More importantly, I don't think it is correct that a sum is separable iff the summands are. For example (Frob-1)+(1). The reverse implication also fails, e.g. (Frob)+(1). So I think that is_separable() for sums will have to do some work.

Certainly, but I don't think the current branch does it wrong, does it? The `.is_separable()` method inherited from `EllipticCurveHom` simply checks that the inseparable degree equals one, so having a correct implementation of the `.inseparable_degree()` method suffices, unless I'm missing something.

JohnCremona: Sorry, you are moving faster than I can keep up! I'm sure I did once see an implementation which just checked if the summands were seperable.  It's fine now.

JohnCremona:
All the suggestions I made in my review and follow-ups have been implemented (or corrected), so this can have a positive review from me assuming that tests pass.

yyyyx4:
Thanks a lot, @JohnCremona! The failures appear to be unrelated to these changes.

PR #36638, opened 2023-11-03
return Weierstrass scaling factor in base field for EllipticCurveIsogeny

yyyyx4:
In the absence of pre- and post-isomorphisms, this value was returned as an `Integer` instead of in the base field, which is unexpected and sometimes causes things to break spontaneously.


JohnCremona:
I can review this but not until Tuesday -- it's a good idea and looks a trivial change in the code.

JohnCremona:
Sorry for the delay, I forgot this one.  It's good.

PR #36271, opened 2023-09-16
`sage.{dynamics,schemes}`: Modularization fixes, docstring cosmetics, update `# needs`

mkoeppe:
- Part of: #29705 
- Cherry-picked from: #35095 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36639, opened 2023-11-03
fix broken pyright in CI

fchapoton:
This is trying to fix our broken CI, where pyright complains about the type of `PariError`.

### :memo: Checklist


tobiasdiez (approved):
Thanks for taking the initiative. Your fix works and looks good to me (Thus marking it as blocker).

However, it reverts the changes of https://github.com/sagemath/sage/pull/36271 so I'm not sure if @mkoeppe is happy with it.

mkoeppe:
It's fine for now, I'll find a solution that is compatible with pyright in #35095, where this change was cherry-picked from.

Issue #36174, opened 2023-09-01
Absolute length returning wrong value in affine type


thecaligarmo:
### Steps To Reproduce

When trying to get the absolute length of an element in an Affine Coxeter group, we get the wrong value.

```
sage: W = CoxeterGroup(["A",2,1])
sage: (r, s, t) = W.simple_reflections()
sage: (r * s * r * t).absolute_length()
1
```

**Mathematical Background:** Given an affine Coxeter group $W$ generated by a set $S$, the set $R$ (of reflections) is the conjugates of $S$. In other words, $R = \cup_{w \in W} wSw^{-1}$. The absolute length of an element $w$ is the minimal length of words which represent $w$ over the generating set $R$.

In the above case, $S = \left\\{ r, s, t\right\\}$ and $rsr$ and $t$ are both reflections in $R$ implying that $rsrt$ has absolute length $2$. (Reflections always have an odd number of generators and so $rsrt$ must have an even absolute length).

### Expected Behavior

```
sage: W = CoxeterGroup(["A",2,1])
sage: (r, s, t) = W.simple_reflections()
sage: (r * s * r * t).absolute_length()
2
```

### Actual Behavior

```
sage: W = CoxeterGroup(["A",2,1])
sage: (r, s, t) = W.simple_reflections()
sage: (r * s * r * t).absolute_length()
1
```

### Environment

```markdown
- **OS: Ubuntu 22.04**:
- **Sage Version: 9.8 and 10.0**:
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

fchapoton:
This is using the definition as `(M-1).rank()`, maybe not valid outside of finite type ?

thecaligarmo:
Yeah, I think `(M-1).rank()` only works for finite type. In affine case you can get around it using the paper by Lewis, McCammond, Peterson and Schwer (https://arxiv.org/abs/1710.06920) where they basically state that in an affine Coxeter group, the absolute length is equal to $2 \cdot dim(w) - dim(p(w))$ where $p(w)$ is a projection of $w$ to its finite part.

fchapoton:
feel free to propose a pull request

thecaligarmo:
[...]

PR #36650, opened 2023-11-05
Implement quick fix for absolute length for non-finite Coxeter groups

thecaligarmo:
Implement a quick fix for absolute length in non-finite Coxeter groups

Puts in a temporary fix for #36174 . The issue of fixing absolute length in sagemath is turning out to be a little more complicated than previously assumed. So until a proper fix can be put in place, this PR removes the ability to calculate the absolute length for Coxeter groups which are not finite (as the number being produced is wrong). 

### :memo: Checklist

### :hourglass: Dependencies

#36174


tscrim:
What about general types? I don't think we should disable this in general, but rather raise a warning to the user to be careful.

thecaligarmo:
I mean, as far as I'm aware, it's only true in finite types (the dimension definition). So (currently) there is no accurate/correct way to give absolute length for elements coming from non-finite groups. By "warning" do you mean like a warning error rather than a not implemented error? (I'm looking at the other comment you had made on the other issue now. This could be null and moot cause it looks like the other option is fast enough that we can just implement that instead without needing a quick fix. I hadn't seen that article)

tscrim:
I was thinking of something like “warning(“don’t trust this”),” albeit with a more descriptive message of course.

As you said, the proposed general fix on #36174 would make this moot. I would keep the current implementation for finite types. Perhaps for “long” words, wrt some reasonable cutoff based on some testing, switching between the two algorithms, but that might be more complicated than needed though (and implementation dependent, which would make the cutoff somewhat arbitrary and require a length computation, making it more expensive than just doing the current implementation).

thecaligarmo:
Closing this pull request as it is no longer needed.

PR #36651, opened 2023-11-05
enhancements in infinite polynomial rings

fchapoton:
This make a few little changes, nothing serious.

This catches more precise exceptions, not `Exception`.

### DIFF
--- a/src/sage/rings/polynomial/infinite_polynomial_ring.py
+++ b/src/sage/rings/polynomial/infinite_polynomial_ring.py
@@ -491,7 +491,7 @@ class InfiniteGenDict:
         try:
             if len(L) == 2:
                 return self._D[L[0]][int(L[1])]
-        except Exception:
+        except (TypeError, ValueError):
             pass
         raise KeyError("%s is not a variable name" % k)
 
@@ -614,7 +614,7 @@ class GenDictWithBasering:
                 return D[k]
             except KeyError:
                 pass
-        raise KeyError("%s is not a variable name of %s or its iterated base rings" % (k, self._P))
+        raise KeyError("{} is not a variable name of {} or its iterated base rings".format(k, self._P))
 
 
 ##############################################################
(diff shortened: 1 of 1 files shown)


mantepse:
Thank you!  It appears to me that the `all_gens` method is completely new (in Sage), do you have any plans with it?

fchapoton:
well, not really. The problem is that the usual semantic of `gen` and `gens` is not respected in this file. They do not return some elements of the ring. I am not sure what to do. I can remove `all_gens` or call it otherwise or make it hidden..

mantepse:
Do you know whether there is any good reason for not respecting the usual semantics?

fchapoton:
it's because there are infinitely many generators, in a finite number of families

mantepse:
I don't see how that's a good reason. Is this the only infinite dimensional ring in sage which has gen and gens?

fchapoton:
It is also a matter of offering a simple and practical access to the variables.

mantepse:
Can we have both, in case it makes sense?  I find
```
sage: P.<a, b> = InfinitePolynomialRing(QQ)
```
practical, and I do not see why we couldn't have, at the same time
```
sage: [P.gen(i) for i in range(5)]
[a[0], b[0], a[1], b[1], a[2]]
```
I looked at the code: on the plus side, `ParentWithGens` is deprecated, on the minus side, there is no method returning a family of generators.

I think that there are quite a few things that would work better with `gen` returning a proper generator of the ring, for examples substitution and defining homomorphisms.

mantepse:
Old discussion: https://groups.google.com/g/sage-devel/c/-CVdz6H7dTc/m/_x1px4bHAgAJ

fchapoton:
See also #34120

fchapoton:
I do not want to do big things and changes here. So maybe we can keep only the changes that enlarge the possible input of the `gen` method of infinite polynomial rings ? Would you approve that ?

mantepse:
I would rather not modify the semantics before agreeing on the desirable semantics, especially since we seem to agree that the current semantics breaks stuff quite badly. Do you have any concrete plans with your proposed improvement?

fchapoton:
I dot not  have any specific plan about `gen` and  `gens`. I may try to get rid of `ParentWithGens` when I have time.

So we can close the present poule-requete, vielleicht ?

fchapoton:
Ok, I have replaced the branch by a new one only making very simple changes, and not touching anything about gens

mantepse:
Can I assume that the failure in singular is unrelated?
```
File "src/sage/interfaces/singular.py", line 530, in sage.interfaces.singular.Singular._send_interrupt
Failed example:
    2*a
Expected:
    2
Got:
    <BLANKLINE>
```

If so, positive review from my side.

mantepse (approved):
Thank you!

PR #36664, opened 2023-11-06
ruff auto-fix for C4 in modular

fchapoton:
this was scripted using
```
ruff check --select=C4 --fix src/sage/modular/
```
and is about simplifying list,set,dict comprehensions

### :memo: Checklist

### DIFF
--- a/src/sage/modular/abvar/homspace.py
+++ b/src/sage/modular/abvar/homspace.py
@@ -570,7 +570,7 @@ class Homspace(HomsetWithBase):
             return
 
         if (self.domain() == self.codomain()) and (self.domain().dimension() == 1):
-            self._gens = tuple([ identity_matrix(ZZ,2) ])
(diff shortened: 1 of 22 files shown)


src/sage/modular/abvar/homspace.py
```diff
@@ -570,7 +570,7 @@ def calculate_generators(self):
             return
 
         if (self.domain() == self.codomain()) and (self.domain().dimension() == 1):
-            self._gens = tuple([ identity_matrix(ZZ,2) ])
+            self._gens = ( identity_matrix(ZZ,2), )
```
DavidAyotte: Small detail: there are extra spacing characters here near the parentheses.

DavidAyotte (changes requested):

DavidAyotte:
Thank you!

PR #36665, opened 2023-11-06
implement Bröker's algorithm for constructing supersingular curves

yyyyx4:
In this patch, we add a function `special_supersingular_curve()` to construct a supersingular elliptic curve defined over a given $\mathbb F_p$ having a small-degree endomorphism which anticommutes with Frobenius. The function optionally constructs and returns the endomorphism as an explicit isogeny; this should prove useful later when we have better tools for endomorphism rings.

Bröker's algorithm is both the standard approach for finding **a** supersingular elliptic curve in given characteristic, as well as part of the setup for several algorithms dealing with endomorphism rings: The special endomorphism together with Frobenius forms a generating set of the endomorphism *algebra*.

Cc: @JohnCremona @defeo @GiacomoPope @remyoudompheng


src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2414,3 +2416,172 @@ def is_j_supersingular(j, proof=True):
     # expensive since it involves counting the number of points on E):
 
     return E.trace_of_frobenius() % p == 0
+
+def special_supersingular_curve(F, *, endomorphism=False):
+    r"""
+    Given a finite field ``F`` of degree `1` or `2`, construct a
+    "special" supersingular elliptic curve `E` defined over ``F``.
+
```
JohnCremona, line 2424: It's not obvious to me while the degree has to be 1 or 2.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2414,3 +2416,172 @@ def is_j_supersingular(j, proof=True):
...
+    """
+    if not isinstance(F, FiniteField) or (deg := F.degree()) > 2:
+        raise TypeError('input must be a finite field of degree 1 or 2')
+    p = F.characteristic()
+
+    if endomorphism and deg % 2:
+        raise ValueError('endomorphism was requested but is not defined over given field')
+
+    E = None
+
```
JohnCremona, line 2550: Can you add a comment to say what q is?  i.e. (I think) it is the degree of the special endo being constructed.

JohnCremona:
I amde a couple of comments, not in a formal review (yet).

src/sage/schemes/elliptic_curves/ell_finite_field.py
yyyyx4: To slightly simplify the implementation, mostly. I changed it to allow arbitrary finite fields.

src/sage/schemes/elliptic_curves/ell_finite_field.py
yyyyx4: It was already mentioned in the docstring, but I also added a comment in the code.

JohnCremona (approved):

yyyyx4:
Thank you!

vbraun:
pdf docs fail
```
[sagemath_doc_pdf-none] Overfull \hbox
[sagemath_doc_pdf-none] ! Undefined control sequence.
[sagemath_doc_pdf-none] l.12331 has coefficients in \(\FF
[sagemath_doc_pdf-none]                                  _p\);
```

PR #36670, opened 2023-11-06
.github/workflows/docker.yml: Interrupt the build before the 6 hour cancellation

mkoeppe:
When a CI Linux workflow is canceled, because we are using `docker build`, no image is written out.

Here we add an input parameter `timeout`, defaulting to 5.5 hours, after which we kill the  `make` process inside of the build container.
Then our script (in tox) writes out and pushes an image (marked as "-failed") to ghcr.io, enabling better diagnostics.

See https://github.com/mkoeppe/sage/actions/runs/6831246509/job/18580446041 for a test run where for demonstration purposes I set the timeout to 6 minutes.

### :memo: Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -78,6 +78,10 @@ on:
       free_disk_space:
         default: false
         type: boolean
+      timeout:
+        description: 'Elapsed time (seconds) at which to kill the build'
+        default: 20000
+        type: number
       #
       # Publishing to GitHub Packages
       #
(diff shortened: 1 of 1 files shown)


kwankyu:
Is this for workflow that is canceled by timeout? Does this allow to investigate the image (with `-failed`) for the cause of timeout? Is there other (real) use cases?

mkoeppe:
> Is this for workflow that is canceled by timeout? Does this allow to investigate the image (with `-failed`) for the cause of timeout? Is there other (real) use cases?

Yes, the main use case is for debugging these timeouts. But I could also imagine using it to get earlier access to the image of a failing build.

kwankyu (approved):
OK.

mkoeppe:
Thanks... but it wasn't ready for review; it doesn't work yet

kwankyu:
Ah, sorry. 

> Thanks... but it wasn't ready for review; 

Ah, sorry. 

> it doesn't work yet

The code doesn't work? Or do you plan to do more?

mkoeppe:
This approach works locally as described, but in GH Actions the runner dies too quickly. I'll have to make changes and experiments

kwankyu:
Isn't `always()` for this case?

mkoeppe:
It's possible that I may be making a stupid mistake in setting up the trap in this step. I cannot see any sign of it being triggered.

Help with this is very welcome!

mkoeppe:
I've switched to a simpler approach, suitable for the main use case.

kwankyu (approved):
OK, if it works for you.

mkoeppe:
Thanks!

mkoeppe:
Rebased onto the proper rc2.

mkoeppe:
Setting to blocker because there are more build hangs, possibly from contourpy, in https://github.com/sagemath/sage/actions/runs/6842945144/job/18611623959#step:11:1932, https://github.com/sagemath/sage/actions/runs/6842945144/job/18611622942#step:11:2109, https://github.com/sagemath/sage/actions/runs/6842945144/job/18611624721

PR #36332, opened 2023-09-25
Yet more more spkgs confs

dimpase:
a continuation of #36276  (more system Python packages support via `--enable-system-site-packages`)

Also, added a Cython 3.0.2 patch to fix the usage of `--enable-system-site-packages` option. Without it, and sufficiently many Python system packages, sagelib won't build. (Therefore, blocker label added)

PR #36129, opened 2023-08-24
Notebook 7, ipykernel 6.27, ipython 8.17

mkoeppe:
This is now based on JupyterLab. There are many changes in dependencies.

We are trying to use platform-independent wheels when available, to avoid any Node.JS activity during our build.
Hence we remove our `nodeenv`, `nodejs`, `jupyter_packaging`, `hatch_nodejs_version` packages.

There is trouble on the horizon regarding our model of building everything from source: The latest versions of `jsonschema` have switched from `pyrsistent` to a Rust-based package.
We use the newest versions that don't pull in the build dependency on the Rust compiler.

- Resolves #24904
- Resolves https://github.com/sagemath/sage/issues/30399
- Resolves #33772

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #35251 (merged here)

PR #36674, opened 2023-11-07
spkg-configure.m4 for python: tzdata, pytest, vcversioner

dimpase:
also, remove leftovers of `setuptools_scm_git_archive`, left in the tree by an oversight.

This is continuation of #36332


mkoeppe:
Note that #36129 is going to remove vcversioner, so perhaps this part can be omitted

orlitzky:
tzdata can be skipped in most cases because it provides timezone data that's already installed on every system; the system python will be configured to use the system timezone data. In Gentoo the tzdata package literally does nothing.

To check, we would need to run some zoneinfo command to see if it fails. I once started to write an autoconf macro for that but it's probably overkill here. Something like `python -s -c 'import zoneinfo; print(zoneinfo.ZoneInfo("America/New_York"));'`could work? (The `-s` is to simulate our runtime poisoning of `PYTHONUSERBASE`.)

orlitzky:
"python" should probably be `$PYTHON_FOR_VENV` in that command, but I don't remember what the difference is any more

orlitzky:
Maybe we don't have to think about `tzdata` at all. It's used only by `pytz_deprecation_shim`, which is used only by `tzlocal`, which is used only by `rpy2`. Both `pytz_deprecation_shim` and `tzlocal` are already disabled with `--disable-r`. I think we should add `tzdata` to the list of R packages in `configure.ac` and forget about it until there's enough consensus on sage-devel to delete R.

mkoeppe:
> I think we should add `tzdata` to the list of R packages in `configure.ac`

I agree

orlitzky:
Rebase on top of https://github.com/sagemath/sage/pull/36129 with no vcversioner and with tzdata disabled by `--disable-r`?

I think it's about time we switch the default to `--disable-r` too. The last meaningful change to the R interface was,

```
commit 2c069c9c1224b2f9f26b9ae282e6ac22991dfa91
Author: Karl-Dieter Crisman <kcrisman@gmail.com>
Date:   Mon Apr 19 21:26:35 2010 -0400

    Trac 7665 - updates R interface to fully enable R graphics in nb and command line

```

mkoeppe:
> Rebase on top of #36129 with no vcversioner and with tzdata disabled by `--disable-r`?

Yes please

> I think it's about time we switch the default to `--disable-r` too. The last meaningful change to the R interface was,

Hm? No, we have working rpy2, which is installed whenever there's a compatible system R.

orlitzky:
> > I think it's about time we switch the default to `--disable-r` too. The last meaningful change to the R interface was,
> 
> Hm? No, we have working rpy2, which is installed whenever there's a compatible system R.

And when there's not, we build one for no reason.

What I should have said though is "switch R to optional" rather than disabled. Then if people want to use R, they install R.

orlitzky:
> No, we don't. R is a dummy package.

Of course it is. I'm going to take a nap.

mkoeppe:
Should rebase and remove vcversioner now

PR #36666, opened 2023-11-06
Replace relative imports by absolute ones in `sage.{algebras,arith,categories,cpython,data_structures,misc,modular,rings,sat,symbolic}`

mkoeppe:
Final chunk of the cherry-pick from #35095 for #36228, see also https://github.com/sagemath/sage/pull/36572#issuecomment-1784216595

This PR overlaps with #36572, #36588, #36589, but it does not touch `.all*` files (no changes there are needed for #36228), thus avoiding conflicts with #35095 (see also https://github.com/sagemath/sage/pull/36524#issuecomment-1792680978)

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36372, opened 2023-10-01
Fix/ignore all remaining test failures when run under conda

tobiasdiez:
Add a `skip_conda` flag that skips the test when executed in an active conda environment. This is used to disable the remaining handful of tests that are still failing under linux in combination with https://github.com/sagemath/sage/pull/35593 (refs https://github.com/sagemath/sage/issues/35597).

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- #36411 (as it fixes some of the tests)
- https://github.com/sagemath/sage/pull/36468 (as it fixes the conda setup)

PR #36647, opened 2023-11-04
fix random doctest error in `src/sage/rings/polynomial/skew_polynomial_finite_field.pyx`

dcoudert:
Fixes #36646.

Add `# random` tag to a doctest to avoid random failure reported in #36646.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36373, opened 2023-10-01
Enable conda ci for all PRs and remove experimental label

tobiasdiez:
Using conda to install all sage dependencies is pretty stable and has been long enough in "experimental mode". We remove here the experimental flag and in order to make sure that the workflow is not breaking enable the conda-ci on all PRs. This also provides a) faster feedback as the whole build+test takes only about 1.5h (which is less time then what only the "incremental build" currently takes) and b) tests that the code is working on all supported python versions. The last remaining failing tests under linux are temporarily disabled in https://github.com/sagemath/sage/pull/36372.

### :memo: Checklist

- [ ] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36678, opened 2023-11-08
CI conda: Ignore baseline test failures 

mkoeppe:
This makes the CI conda pass by marking affected files as "known baseline failures", thus reducing the noise that this workflow makes on PRs (including https://github.com/sagemath/sage/pull/36666#issuecomment-1805078169)

Less intrusive version of #36372.

In addition to the failure marked there, here we collect a number of more failures observed over the course of a week in https://github.com/sagemath/sage/actions/workflows/ci-conda.yml?query=is%3Acompleted

### :memo: Checklist

### DIFF
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -5,6 +5,9 @@
     "sage.combinat.cluster_algebra_quiver.quiver": {
         "failed": true
     },
+    "sage.geometry.cone": {
+        "failed": true
+    },
     "sage.groups.matrix_gps.finitely_generated_gap": {
         "failed": true
     },
@@ -15,6 +18,9 @@
         "failed": true
     },
     "sage.matrix.matrix2": {
+        "failed": true
+    },
+    "sage.matrix.matrix_integer_sparse": {
         "failed": true
     },
     "sage.misc.lazy_import": {
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -11,6 +11,9 @@
     "sage.misc.lazy_import": {
         "failed": true
     },
+    "sage.misc.weak_dict": {
+        "failed": true
+    },
     "sage.rings.polynomial.multi_polynomial_ideal": {
         "failed": true
     },
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -1,0 +1,23 @@
+{
+    "sage_setup.clean": {
+        "failed": true
+    },
+    "sage.groups.matrix_gps.finitely_generated_gap": {
+        "failed": true
+    },
+    "sage.libs.singular.singular": {
+        "failed": true
+    },
+    "sage.misc.lazy_import": {
+        "failed": true
+    },
+    "sage.rings.polynomial.multi_polynomial_ideal": {
+        "failed": true
+    },
+    "sage.rings.polynomial.multi_polynomial_libsingular": {
+        "failed": true
+    },
+    "sage.tests.gap_packages": {
+        "failed": true
+    }
+}
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -20,6 +20,9 @@
     "sage.modular.modform.l_series_gross_zagier": {
         "failed": true
     },
+    "sage.rings.function_field.drinfeld_modules.morphism": {
+        "failed": true
+    },
     "sage.rings.polynomial.multi_polynomial_ideal": {
         "failed": true
     },
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -11,6 +11,9 @@
     "sage.groups.matrix_gps.finitely_generated_gap": {
         "failed": true
     },
+    "sage.interfaces.expect": {
+        "failed": true
+    },
     "sage.libs.gap.element": {
         "failed": true
     },
@@ -39,6 +42,9 @@
         "failed": true
     },
     "sage.rings.polynomial.multi_polynomial_libsingular": {
+        "failed": true
+    },
+    "sage.rings.polynomial.skew_polynomial_finite_field": {
         "failed": true
     },
     "sage.tests.gap_packages": {
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -5,6 +5,9 @@
     "sage.groups.matrix_gps.finitely_generated_gap": {
         "failed": true
     },
+    "sage.libs.gap.element": {
+        "failed": true
+    },
     "sage.libs.singular.singular": {
         "failed": true
     },
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -14,6 +14,9 @@
     "sage.misc.weak_dict": {
         "failed": true
     },
+    "sage.modular.modform.l_series_gross_zagier": {
+        "failed": true
+    },
     "sage.rings.polynomial.multi_polynomial_ideal": {
         "failed": true
     },
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -2,6 +2,9 @@
     "sage_setup.clean": {
         "failed": true
     },
+    "sage.combinat.cluster_algebra_quiver.quiver": {
+        "failed": true
+    },
     "sage.groups.matrix_gps.finitely_generated_gap": {
         "failed": true
     },
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -14,6 +14,9 @@
     "sage.libs.singular.singular": {
         "failed": true
     },
+    "sage.matrix.matrix2": {
+        "failed": true
+    },
     "sage.misc.lazy_import": {
         "failed": true
     },
--- a/.github/workflows/ci-conda.yml
+++ b/.github/workflows/ci-conda.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Test
         if: success() || failure()
         shell: bash -l {0}
-        run: ./sage -t --all -p0
+        run: ./sage -t --all --baseline-stats-path=.github/workflows/ci-conda-known-test-failures.json -p0
 
       - name: Print logs
         if: always()


tobiasdiez (dismissed):
There are only a couple of test failures in these files. These ignore patterns are way to coarse.

(removing blocker label for that reason to not hide possible issues, and because it's not working anyway in this context)

tobiasdiez:
So if another new test in the same file fails, the conda ci is red?

mkoeppe:
Perhaps you should clarify what you think the purpose of the ci-conda workflows is. 
To my understanding they are being run so that the major breakage that we have seen in the past (from unmonitored package upgrades in conda-forge) does not go unnoticed.
I don't think it's really the job of this workflow to duplicate the test coverage done by build.yml

Marking these files as baseline failures is done so that on unrelated workflows, the red checkmarks on the conda workflows do not annoy developers.

mkoeppe:
And obviously work is needed to actually diagnose, report, and fix these failures. The approach of #36372 ("skip_conda") hides these failures too well.

tobiasdiez:
> Perhaps you should clarify what you think the purpose of the ci-conda workflows is. To my understanding they are being run so that the major breakage that we have seen in the past (from unmonitored package upgrades in conda-forge) does not go unnoticed. I don't think it's really the job of this workflow to duplicate the test coverage done by build.yml

They have the same purpose as the Build & Test workflow for people using the conda environment. So yes, they should duplicate the test coverage.

(In the long term, once the conda runs are a bit more stable and there are no more conda-specific issues, we can remove the Build & Test workflow).

kwankyu:
> They have the same purpose as the Build & Test workflow for people using the conda environment. So yes, they should duplicate the test coverage.

B&T is mainly for checking sage library changes to help developers and reviewers. It is unlikely that some code change in the sage library works on a platform and fails on another. That is why B&T runs only on a specific decent platform and tests on all other platforms are done only for weekly release.

I do not use conda (linux neither). I regard conda as another platform. So it looks strange to me that currently 6 B&T workflows for conda platforms run for each PR. I think there is already too much duplication. 

Perhaps conda is special, and separate B&T for conda is necessary for PRs (perhaps upgrading packages). Then why not just two B&T for conda (one for ubuntu and one for mac) for PRs and more tests for weekly release?  

About #36372, I object to introducing conda-specific doctest tags into sage library. That is another cluttering of sage library, in addition to all doctest tags for modularization. We decided to live with modularization. But please never platform-specific doctest tags.

mkoeppe:
> it looks strange to me that currently 6 B&T workflows for conda platforms run for each PR. I think there is already too much duplication.

FWIW, I approved the change to run these jobs on all PRs in #36373. Tobias has already implemented a mitigation (see https://github.com/sagemath/sage/pull/36373#issuecomment-1742386153) there on my request by making it "fail-fast" and setting a max-parallel 2.

But I think at least the macOS jobs need reconsideration / additional adjustment. Given that we only have 5 macOS runners, launching 3 jobs for every PR is clearly too much. 

For example, at the time that 10.2.rc1 was released (over an 1 hour ago), at least 7 "ci-conda" workflows were already in the queue (https://github.com/sagemath/sage/actions/workflows/ci-conda.yml?query=is%3Aqueued), with at least 14 macOS jobs waiting for a runner. As a result, the CI-macos run of 10.2.rc1 (https://github.com/sagemath/sage/actions/runs/6827163283) has not been able to start yet.

mkoeppe:
> I think at least the macOS jobs need reconsideration / additional adjustment. Given that we only have 5 macOS runners, launching 3 jobs (each 1 hour) for every PR is clearly too much.

Implemented in  #36694

mkoeppe:
> Then why not just two B&T for conda (one for ubuntu and one for mac) for PRs and more tests for weekly release?

I agree and have updated #36694 to do this.

kwankyu:
Currently cond ci is the main reason breaking and delaying checks.

Let's get this in.

tobiasdiez:
https://github.com/sagemath/sage/pull/36678#pullrequestreview-1721526377 is still not addressed. If a single test in the "Build & Test" workflow would fail, you also wouldn't want to to exclude the whole file, would you?

mkoeppe:
> [#36678 (review)](https://github.com/sagemath/sage/pull/36678#pullrequestreview-1721526377) is still not addressed. If a single test in the "Build & Test" workflow would fail, you also wouldn't want to to exclude the whole file, would you?

We have addressed it in the discussion (@kwankyu in https://github.com/sagemath/sage/pull/36678#issuecomment-1804901930). The conda tests do *not* serve the same function as the Build & Test workflow:
- I understand that you hope to replace Build&Test by the conda workflow at some point, but no such decision has been taken; and I have seen no effort to make it ready for that.
- Amplifying the failure probability from random failures by running 7 workflows instead of 1 does not help with diagnosing or fixing them. (It also does not help that these random failures are scattered across these 7 workflows.)
- Currently, developers simply do not know what to do when the conda workflows fail. It has been a mixture of conda-specific failures and random errors.

kwankyu:
> [#36678 (review)](https://github.com/sagemath/sage/pull/36678#pullrequestreview-1721526377) is still not addressed. If a single test in the "Build & Test" workflow would fail, you also wouldn't want to to exclude the whole file, would you?

I would. We should. "B&T for conda" is specifically for conda. If there is a failure with it, I would exclude the file from  being tested for other PRs until the failure is fixed. Keeping it fail for other PRs betrays the purpose of checks. We expect all checks pass unless the PR itself has a problem.

If you want the file to be tested again for conda, you should spend time on fixing the failing test instead of worrying that other test might fail in the same file (which seems quite imaginary to me).

tobiasdiez:
> > [#36678 (review)](https://github.com/sagemath/sage/pull/36678#pullrequestreview-1721526377) is still not addressed. If a single test in the "Build & Test" workflow would fail, you also wouldn't want to to exclude the whole file, would you?
> 
> We have addressed it in the discussion (@kwankyu in [#36678 (comment)](https://github.com/sagemath/sage/pull/36678#issuecomment-1804901930)). The conda tests do _not_ serve the same function as the Build & Test workflow:
> 
>     * I understand that you hope to replace Build&Test by the conda workflow at some point, but no such decision has been taken; and I have seen no effort to make it ready for that.
> 
>     * Amplifying the failure probability from random failures by running 7 workflows instead of 1 does not help with diagnosing or fixing them. (It also does not help that these random failures are scattered across these 7 workflows.)
> 
>     * Currently, developers simply do not know what to do when the conda workflows fail. It has been a mixture of conda-specific failures and random errors.

I fail to see how any of that is relevant to the question whether to hide conda failures on the file or test level. As can be seen in https://github.com/sagemath/sage/pull/36372, the test failures happen once per file (with the exception of gap).
Concerning your first point: https://github.com/search?q=repo%3Asagemath%2Fsage+conda&type=pullrequests

> > [#36678 (review)](https://github.com/sagemath/sage/pull/36678#pullrequestreview-1721526377) is still not addressed. If a single test in the "Build & Test" workflow would fail, you also wouldn't want to to exclude the whole file, would you?
> 
> I would. We should. "B&T for conda" is specifically for conda. If there is a failure with it, I would exclude the file from being tested for other PRs until the failure is fixed. 

Really? You would exclude the whole file and not just mark the single failing test with "# random"?

kwankyu:
> > I would. We should. "B&T for conda" is specifically for conda. If there is a failure with it, I would exclude the file from being tested for other PRs until the failure is fixed.
> 
> Really? You would exclude the whole file and not just mark the single failing test with "# random"?

We are talking about some doctest failing only in conda ci, and there is no problem with the doctest on the rest of the platforms. Why tag it with `# random`?

kwankyu:
By the way, `# random` is used for inherently random outputs. It is not for random failures.

tobiasdiez:
> > > I would. We should. "B&T for conda" is specifically for conda. If there is a failure with it, I would exclude the file from being tested for other PRs until the failure is fixed.
> > 
> > 
> > Really? You would exclude the whole file and not just mark the single failing test with "# random"?
> 
> We are talking about some doctest failing only in conda ci, and there is no problem with the doctest on the rest of the platforms. Why tag it with `# random`?

It seems to be the established practice: https://github.com/sagemath/sage/pull/36647
But I agree with you that `random` is not a good solution for this, but since there are currently no means to tag a doctest "this breaks on system xyz, but not others" it's the next best thing. To have a more conceptual solution, I introduced the coda-specific exception.

Anyway, I'm okay with ignoring the failing tests using an external file. But this mechanism should only exclude the tests that are really failing to a) prevent that additional errors are introduced and b) to make it easier to work on these issues (it's essential to know which line actually breaks randomly in source file of +1000 lines).

kwankyu:
> > > > I would. We should. "B&T for conda" is specifically for conda. If there is a failure with it, I would exclude the file from being tested for other PRs until the failure is fixed.
> > > 
> > > 
> > > Really? You would exclude the whole file and not just mark the single failing test with "# random"?
> > 
> > 
> > We are talking about some doctest failing only in conda ci, and there is no problem with the doctest on the rest of the platforms. Why tag it with `# random`?
> 
> It seems to be the established practice: #36647 (among others) But I agree with you that `random` is not a good solution for this, but since there are currently no means to tag a doctest "this breaks on system xyz, but not others" it's the next best thing. To have a more conceptual solution, I introduced the coda-specific exception.
> 
> Anyway, I'm okay with ignoring the failing tests using an external file. But this mechanism should only exclude the tests that are really failing to a) prevent that additional errors are introduced and b) to make it easier to work on these issues (it's essential to know which line actually breaks randomly in source file of +1000 lines).

Just focus on the failing test that made the file excluded. We may add a comment telling the failing test (or tests) when we add the file to the list of the ignored files. It would make things complicated if we try to exclude only the failing tests. Why make things complicated just for rare cases (almost looking imaginary to me)?

mkoeppe:
> b) to make it easier to work on these issues (it's essential to know which line actually breaks randomly in source file of +1000 lines).

One sees this in the logs -- because with the approach here, the failures are not hidden but only marked as "[failed in baseline]".

tobiasdiez:
> > b) to make it easier to work on these issues (it's essential to know which line actually breaks randomly in source file of +1000 lines).
> 
> One sees this in the logs -- because with the approach here, the failures are not hidden but only marked as "[failed in baseline]".

That clearly only works for the issues that occur often/reproducible. I was talking about the random failures that happen once every 100 runs or so.

kwankyu:
The random failure you are talking about seems not related with PRs. Why do you want to observe it in PR checks? To investigate such a random failure, I would just devise massive repetitive test runs with just one version of sage. No?

mkoeppe:
I also don't understand the comment in https://github.com/sagemath/sage/pull/36678#issuecomment-1826258022

tobiasdiez:
> Why do you want to observe it in PR checks? 

I don't want to observe them at all. It just reality that these failing tests are occurring very rarely, which makes it hard to reproduce them locally. Heck, even now there are random test failures of the Build & Test workflow (e.g. https://github.com/sagemath/sage/actions/runs/6970122859/job/18967437726). Do you want to exclude the whole file from the tests because of this?

> To investigate such a random failure, I would just devise massive repetitive test runs with just one version of sage. No?

I actually did this. But that only revealed a couple of errors. But if you have computation time over that you don't need, I would appreciate it very much if you could run the conda tests in a loop and let me know about the failures.

> I also don't understand the comment in [#36678 (comment)](https://github.com/sagemath/sage/pull/36678#issuecomment-1826258022)

Take a failure that occurs once every 100 runs. If you only know that the failure happens in a given file, you need to run all doctests in that file for 100 times (I know, I know statistics...), which could easily take a few hours depending on the file, only to reproduce it locally. And that would imply that the error actually happens on the given system (some of the errors are occurring only on eg macos). If you know the exact test that fails, you copy that one to a test file, put a loop around it and have it reproduced in a couple of mins.

mkoeppe:
> even now there are random test failures of the Build & Test workflow [...]. Do you want to exclude the whole file from the tests because of this?

Asked and answered already.

kwankyu:
> 1. Test failure is noticed
> 2. You record it in a GitHub Issue. There you include the necessary information so that it does not take 2 hours to reconstruct it.
> 3. One adds stuff to the exclusion list to eliminate noise in the tests.

and items in the exclusion list are commented with the the GitHub Issue numbers...

kwankyu (approved):
LGTM.

mkoeppe:
> There are clear short comings with the approach of this PR as described above.

Nope, all have been refuted.

Issue #36685, opened 2023-11-09
`.ci/merge-fixes.sh`: GraphQL: API rate limit exceeded 


mkoeppe:
As observed in https://github.com/sagemath/sage/pull/36372#discussion_r1388281048

mkoeppe:
I think this has been fixed by https://github.com/sagemath/sage/pull/36686

PR #36686, opened 2023-11-09
`.ci/merge-fixes.sh`: Obtain patches via URL, make customizable by repository variable

mkoeppe:
Instead of using `gh pr checkout`, we obtain the CI fixes via their patch URLs.

- This is faster because typically we do not have to unshallow the repo to apply the patches (seconds instead of ~2 minutes)
- Fewer surprises when applied to a PR based on an older release
- Conjecturally uses fewer API queries, helping avoid #36685

When a repository variable `SAGE_CI_FIXES_FROM_REPOSITORIES` is set in a fork, it is used instead of the hardcoded sagemath/sage as the source(s) of the CI fixes; this gives better control in decentralized development. When set to "none", this also makes it possible to see the "ground truth", addressing a concern raised in https://github.com/sagemath/sage/pull/36349#issuecomment-1738400399. See comments at the top of `.ci/merge-fixes.sh` for details.

Also improving the display of details of the merged PRs, as requested by @tornaria in https://github.com/sagemath/sage/pull/36696#issuecomment-1806916614

Example runs:
- https://github.com/sagemath/sage/actions/runs/6854931832/job/18638901859?pr=36686 (here it recovers gracefully from failed merges due to the 10.2.rc1/10.2.rc2 tagging confusions)
- https://github.com/mkoeppe/sage/actions/runs/6855410957/job/18640380398#step:3:8 (I have set `SAGE_CI_FIXES_FROM_REPOSITORIES=none` in my fork -- for testing this PR)

### :memo: Checklist

### DIFF
--- a/.ci/merge-fixes.sh
+++ b/.ci/merge-fixes.sh
@@ -16,7 +16,6 @@ else
     git tag -f test_base
     git commit -q -m "Uncommitted changes" --no-allow-empty -a
     for a in $PRs; do
-        echo "::group::Applying PR https://github.com/$REPO/pull/$a as a patch"
         git tag -f test_head
         # We used to pull the branch and merge it (after unshallowing), but when run on PRs that are
         # based on older releases, it has the side effect of updating to this release,
@@ -29,26 +28,30 @@ else
         # which squashes everything into one diff without commit metadata.
         PATH=build/bin:$PATH build/bin/sage-download-file "https://github.com/$REPO/pull/$a.patch" upstream/$a.patch
         date -u +"%Y-%m-%dT%H:%M:%SZ" > upstream/$a.date  # Record the date, for future reference
-        if git am --empty=keep < upstream/$a.patch; then
-            echo "Applied patch"
+        LAST_SHA=$(sed -n -E '/^From [0-9a-f]{40}/s/^From ([0-9a-f]{40}).*/\1/p' upstream/$a.patch | tail -n 1)
+        echo "::group::Applying PR https://github.com/$REPO/pull/$a @ https://github.com/$REPO/commit/$LAST_SHA as a patch"
+        export GIT_COMMITTER_NAME="$GIT_AUTHOR_NAME applying https://github.com/$REPO/pull/$a @ https://github.com/$REPO/commit/$LAST_SHA"
+        if git am --signoff --empty=keep < upstream/$a.patch; then
+            echo "---- Applied patch ------------------------------------------------------------"
             cat upstream/$a.patch
+            echo "--------------------------------------------------------------------8<---------"
             echo "::endgroup::"
-            echo "Applied #$a as a patch"
         elif git am --abort \
                 && if git fetch --unshallow --all > /dev/null 2>&1; then echo "Unshallowed"; fi \
                 && echo "Retrying with 3-way merge" \
                 && git am --empty=keep --3way < upstream/$a.patch; then
-            echo "Applied patch"
+            echo "---- Applied patch ------------------------------------------------------------"
             cat upstream/$a.patch
+            echo "--------------------------------------------------------------------8<---------"
             echo "::endgroup::"
-            echo "Applied #$a as a patch"
         else
-            echo "Failure applying patch"
-            git am --show-current-patch=diff
+            echo "---- Failure applying patch ---------------------------------------------------"
+            git am --signoff --show-current-patch=diff
+            echo "--------------------------------------------------------------------8<---------"
             echo "::endgroup::"
             echo "Failure applying #$a as a patch, resetting"
-            git am --abort
+            git am --signoff --abort
         fi
     done
-    git log test_base..HEAD
+    #git log test_base..HEAD
 fi
--- a/.ci/merge-fixes.sh
+++ b/.ci/merge-fixes.sh
@@ -1,57 +1,82 @@
 #!/bin/sh
-# Merge open PRs from sagemath/sage labeled "blocker".
-REPO="sagemath/sage"
-GH="gh -R $REPO"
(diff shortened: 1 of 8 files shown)


.ci/merge-fixes.sh
```diff
@@ -2,7 +2,9 @@
 # Merge open PRs from sagemath/sage labeled "blocker".
 REPO="sagemath/sage"
 GH="gh -R $REPO"
-PRs="$($GH pr list --label "p: blocker / 1" --json number --jq '.[].number')"
+mkdir -p upstream
+PRs="$($GH pr list --label "p: blocker / 1" --json number --jq '.[].number' | tee upstream/ci-fixes.txt)"
```
tornaria: Why not add `--label "s: positive review"`? Multiple `--label` arguments work just fine.

.ci/merge-fixes.sh
```diff
@@ -14,23 +16,42 @@ else
...
+        # Here, we obtain the "git am"-able patch of the PR branch.
+        # This also makes it unnecessary to unshallow the repository.
+        #
+        # Considered alternative: Use https://github.com/$REPO/pull/$a.diff,
+        # which squashes everything into one diff without commit metadata.
+        PATH=build/bin:$PATH build/bin/sage-download-file "https://github.com/$REPO/pull/$a.patch" upstream/$a.patch
+        date -u +"%Y-%m-%dT%H:%M:%SZ" > upstream/$a.date  # Record the date, for future reference
+        LAST_SHA=$(sed -n -E '/^From [0-9a-f]{40}/s/^From ([0-9a-f]{40}).*/\1/p' upstream/$a.patch | tail -n 1)
+        echo "::group::Applying PR https://github.com/$REPO/pull/$a @ https://github.com/$REPO/commit/$LAST_SHA as a patch"
+        export GIT_COMMITTER_NAME="$GIT_AUTHOR_NAME applying https://github.com/$REPO/pull/$a @ https://github.com/$REPO/commit/$LAST_SHA"
```
tornaria: I realized `https://github.com/$REPO/commits/$LAST_SHA` is even better.

Compare:
 - https://github.com/sagemath/sage/commit/4a70c1e535bc621bdbe95d8f531375e5700d7c3e
 - https://github.com/sagemath/sage/commits/4a70c1e535bc621bdbe95d8f531375e5700d7c3e

tornaria (commented):
LGTM in general.

.ci/merge-fixes.sh
mkoeppe: Does that work in logs?

.ci/merge-fixes.sh
mkoeppe: We can do that later when the label-warring phase has been outgrown

.ci/merge-fixes.sh
mkoeppe: I want to be careful there so that when the workflow runs in a fork, these links still correctly point to sagemath/sage

tornaria: In the logs only full URLs are clickable, and there is no translation like here. The point is not that it looks shorter, but that the url `https://github.com/sagemath/sage/commit/4a70c1e535bc621bdbe95d8f531375e5700d7c3e` that you are using now is a little bit less nice than the url `https://github.com/sagemath/sage/commits/4a70c1e535bc621bdbe95d8f531375e5700d7c3e`. From the latter one can click on the individual commit to see the patch but it's nicer as a summary of the PR changes when there's more than one commit.

mkoeppe: Ah thanks, I missed that from how github formatted your comment https://github.com/sagemath/sage/pull/36686#discussion_r1390519957

mkoeppe: Done in 9dac7295709.

tornaria: The links still work using `sagemath/sage` as the repo even if the commit doesn't belong to that repo but a fork.

tornaria:
LGTM, but how is it tested?

mkoeppe:
So far, only on its own CI run...

mkoeppe:
I've reworked and generalized this a bit. Ready for review.

kwankyu:
> LGTM, but how is it tested?

See it in action here:

https://github.com/kwankyu/sage/actions/runs/6887591924/job/18735031877

It seems working well. You can even watch it recovering from applying-patch failure.

mkoeppe:
Thanks for testing!

> Comparing with the original `merge-fixes.sh` [...]
> 
> (1) It is just slightly faster.

Note it is much faster in runs where patches apply directly

> (2) It fails in applying the blocker #36617 but the original `merge-fixes.sh` succeeds.

I think this failure would go away if we rebased #36617.

kwankyu:
> > Comparing with the original `merge-fixes.sh` [...]
> > (1) It is just slightly faster.
> 
> Note it is much faster in runs where patches apply directly

Right. It was slow this time because of #36617 that failed to be applied directly.

> > (2) It fails in applying the blocker #36617 but the original `merge-fixes.sh` succeeds.
> 
> I think this failure would go away if we rebased #36617.

Failing in applying non-rebased blockers may be a good thing, by the way.

kwankyu (approved):
LGTM too.

mkoeppe:
> Failing in applying non-rebased blockers may be a good thing, by the way.

I think so, too. It sort of enforces some restraint on what to do in these CI fixes.

mkoeppe:
Thanks both for the review!

tobiasdiez:
It fails to apply https://github.com/sagemath/sage/pull/36372 now, although this PR is up-to-date with the newest dev branch. It's seems like an unnecessary constraint that blocker PR's need to be rebased (instead of merging the latest develop).

PR #36687, opened 2023-11-09
Codestyle improvements (mostly sort imports)

mkoeppe:
Split out to preserve the incidental changes made in @tobiasdiez's original branch of #36372.

Author: Tobias Diez

### :memo: Checklist


tobiasdiez (approved):
Thanks for helping upstreaming these changes.

Next time, please use the magic co-author comment to give proper attribution.

mkoeppe:
Tobias, these are changes authored by you. You can't "positively review" them.

mkoeppe:
I've rebased with --author

tobiasdiez:
Okay, then I misunderstood the purpose of this PR. Could you then please review these changes? Thanks!

src/sage/rings/polynomial/pbori/fglm.py
```diff
@@ -1,5 +1,5 @@
-from .PyPolyBoRi import Polynomial, BoolePolynomialVector
-from .pbori import FGLMStrategy, BooleSet
+from .pbori import BooleSet, FGLMStrategy
```
dcoudert, line 1: We should avoid relative imports, no ?

dcoudert (changes requested):
Why is it important to modify the order of the imports ? I'm missing something here.

src/sage/rings/polynomial/pbori/fglm.py
tobiasdiez: Yes, you are right. But I prefer to do this migration in a separate PR.

tobiasdiez:
> Why is it important to modify the order of the imports ? I'm missing something here.

There is of course nothing important in the import order. As with any code style fix, it merely improves the readability of the code (slightly). For example, it's easier to spot related imports. Perhaps I should also add that these changes were done automatically by ruff.

dcoudert:
Here also we have failing doctests due to nauty. I don't understand what's going on.

dcoudert (approved):
So then, LGTM.

tobiasdiez:
Thanks for the review!

Issue #36690, opened 2023-11-10
Twisted homology of simplicial sets


miguelmarco:
Twisted homology of simplicial sets wrt abelian representations of the fundamental group can be computed using matrices over polynomial rings, modulo some relations.

PR #36691, opened 2023-11-10
Implement twisted homology of simplicial sets.

miguelmarco:
Right now we can compute universal covers of simplicial sets if the fundamental group is finite. This PR implements
twisted homology with respect to the abelianization character. 

In particular, if the fundamental group is abelian, this gives modules whose underlying additive group are isomorphic to the homology groups of the universal cover.

Addresses #36690 

### :memo: Checklist


src/sage/categories/simplicial_sets.py
```diff
@@ -8,13 +8,22 @@
...
+from sage.functions.generalized import sign
+from sage.matrix.special import identity_matrix
 from sage.misc.cachefunc import cached_method
+from sage.misc.lazy_import import lazy_import
 from sage.categories.category_singleton import Category_singleton
 from sage.categories.category_with_axiom import CategoryWithAxiom
 from sage.categories.sets_cat import Sets
 from sage.categories.homsets import HomsetsCategory
+from sage.matrix.constructor import matrix
+from sage.misc.flatten import flatten
```
jhpalmieri: I think this line (20) can be deleted: I don't see `flatten` anywhere in the rest of this file.

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+
+                    sage: X = simplicial_sets.Torus()
+                    sage: d = X._canonical_twisting_operator()
+                    sage: d
+                    {(s_0 v_0, sigma_1): f2*f3^-1, (sigma_1, s_0 v_0): f3, (sigma_1, sigma_1): f2}
+                    sage: list(d.values())[0].parent()
+                    Multivariate Laurent Polynomial Ring in f2, f3 over Integer Ring
+                    sage: Y = simplicial_sets.RealProjectiveSpace(2)
+                    sage: Y._canonical_twisting_operator()
+                    {f: F1bar}
```
jhpalmieri: Perhaps delete lines 600 and 601, since they're covered in 603-604.

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                r"""
+                Return the normalized chain complex twisted by some operator.
+
+                A twisting operator is a map from the set of simplices to some algebra.
+                The differentials are then twisted by this operator.
+
+                INPUT:
+
+                - ``twisting_operator`` -- a dictionary, associating the twist of each
+                simplex. If it is not given, the canonical one (associated to the
```
jhpalmieri:
```suggestion
                  simplex. If it is not given, the canonical one (associated to the
```

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                laurent polynomial ring abelianization of the fundamental group, ignoring
```
jhpalmieri:
```suggestion
                  laurent polynomial ring abelianization of the fundamental group, ignoring
```

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                torsion) is used.
```
jhpalmieri:
```suggestion
                  torsion) is used.
```

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+
```
jhpalmieri, line 631: I think that all of these lines need to be indented. At least I'm getting an error building the documentation, and I think it's coming from this. Same in `twisted_homology` docstring.

Edit: the suggested changes are a bit garbled. Every INPUT item should look like
```
- ``item`` -- text
  more text
```
with "more" aligned with the first backquote before "item".

jhpalmieri:
My first impression is that it adds important functionality to Sage, and it looks good. Should `K.twisted_homology(n)` (etc.) exit gracefully if `K` is simply connected? Right now I just get an error; should it just return the ordinary homology? Should `_canonical_twisting_operator` return a trivial operator, assigning 1 to each edge?

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                    dimensions = list(range(self.dimension() + 1))
+                else:
+                    if not isinstance(dimensions, (list, tuple, range)):
+                        dimensions = list(range(dimensions - 1, dimensions + 2))
+                    else:
+                        dimensions = [n for n in dimensions if n >= 0]
+                    if not dimensions:
+                        # Return the empty chain complex.
+                        if cochain:
+                            return ChainComplex(base_ring=base_ring, degree=1)
```
jhpalmieri, line 735: As written, `K.twisted_chain_complex(dimensions=None)` will return something different from `K.twisted_chain_complex(dimensions=[])`. I'm not sure, but maybe it would make more sense if they both behaved the way `None` is handled.

jhpalmieri:
I have a design question: when I wrote the initial simplicial set code, I set it up so that `fundamental_group` was only available to pointed simplicial sets. Mathematically this is certainly the right thing to do, but should Sage be sloppier and (for example) if the simplicial set is path connected, choose an arbitrary base point?

This would allow moving some methods around between files, and it might make it possible to unify `twisted_chain_complex` with `chain_complex` — `chain_complex` could just take an optional argument for the twisting operator.

None of this would necessarily be done on this PR, but is it worth thinking about, or should we keep it as is?

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                if cochain:
+                    new_diffs = {}
+                    for d in differentials:
+                        new_diffs[d-1] = differentials[d].transpose()
+                    return ChainComplex(new_diffs, degree_of_differential=1,
+                                        check=check)
+                return ChainComplex(differentials, degree_of_differential=-1,
+                                    check=check)
+
+            def twisted_homology(self, n, reduced=False):
```
jhpalmieri, line 815: Should this accept a `twisting_operator` argument, to pass to the chain complex method?

src/sage/categories/simplicial_sets.py
enriqueartal, line 631: Ignoring torsion, can it lead to wrong results?

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                r"""
+                The `n`-th twisted homology module of the simplicial set with respect to
+                the abelianization of the fundamental_group.
+
+                It is a module over a polynomial ring, including relations to make some
+                variables the multiplicative inverses of others.
```
enriqueartal, line 821: I think the option to express it in the Laurent polynomial rings should be done at some point.

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +582,391 @@ def universal_cover(self):
...
+                    sage: Y._canonical_twisting_operator()
+                    {f: F1bar}
+                    sage: d2 = Y._canonical_twisting_operator()
+                    sage: d2
+                    {f: F1bar}
+                    sage: list(d2.values())[0].parent()
+                    Quotient of Univariate Laurent Polynomial Ring in F1 over Integer Ring by the ideal (-1 + F1^2)
+
+                """
+                from sage.rings.polynomial.laurent_polynomial_ring import LaurentPolynomialRing
```
enriqueartal: Apparently LaurentPolynomialRing is not been used

src/sage/rings/polynomial/laurent_polynomial.pyx
```diff
@@ -1988,3 +1988,26 @@ cdef class LaurentPolynomial_univariate(LaurentPolynomial):
...
+            sage: f = t-t^-2
+            sage: f._as_extended_polynomial()
+            -tinv^2 + t
+            sage: _.parent()
+            Multivariate Polynomial Ring in t, tinv over Rational Field
+
+        """
+        dres = {}
+        for (e, c) in self.dict().items():
+            if e > 0 :
```
enriqueartal, line 2009: An extra space before :

src/sage/rings/polynomial/laurent_polynomial.pyx
```diff
@@ -1988,3 +1988,26 @@ cdef class LaurentPolynomial_univariate(LaurentPolynomial):
...
+                dres[(e,0)] = c
+            else:
+                dres[(0,-e)] = c
```
enriqueartal:
```suggestion
            if e > 0:
                dres[(e, 0)] = c
            else:
                dres[(0, -e)] = c
```

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -555,6 +555,34 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+            -t2inv^2 + t1
+            sage: _.parent()
+            Multivariate Polynomial Ring in t1, t1inv, t2, t2inv over Rational Field
+
+        """
+        dres = {}
+        for (e, c) in self.dict().items():
+            exps = []
+            for t in e:
+                if t > 0 :
```
enriqueartal:
```suggestion
                if t > 0:
```

src/sage/categories/simplicial_sets.py
miguelmarco: In the case of twisted homology, it is taken into consideration and the necessary relations are added.

src/sage/categories/simplicial_sets.py
miguelmarco: I agree, but in order for it to work properly, we would need to implement several functionalities in Laurent polynomials (involving matrices, modules, ideals....).

I think that should be done in a separate issue.

src/sage/categories/simplicial_sets.py
miguelmarco: My idea with the interface was that `None`  means "unspecified", and hence, the default is used; whereas `[]` means "i explicitely want the empty list of dimensions to be computed".

src/sage/categories/simplicial_sets.py
miguelmarco: I considered that, but couldn't decide how the interface should  handle the torsion case (right now we gather that info from `abelianization_to_algebra`).

I propose to discuss it in a later issue.

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +581,391 @@ def universal_cover(self):
...
+
+                It is a module over a polynomial ring, including relations to make some
+                variables the multiplicative inverses of others.
+
+                INPUT:
+
+                - ``n`` - a positive integer.
+
+                - ``reduced`` - (default: False) if set to True, the presentation matrix
+                will be reduced.
```
jhpalmieri:
```suggestion
                  will be reduced.
```

jhpalmieri: (One more indentation fix.)

src/sage/categories/simplicial_sets.py
miguelmarco: Solved.

src/sage/categories/simplicial_sets.py
miguelmarco: Fixed.

src/sage/categories/simplicial_sets.py
miguelmarco: Before the changes in `universal_cover_dict`, these two examples behaved differently (the hashes of the simplices in the n-skeleton of a wedge didn't necessarily match the ones in the total space).

I think it is good to have a test that makes sure that that problem is solved.

src/sage/categories/simplicial_sets.py
jhpalmieri: I understand. It doesn't feel ideal to me, but we can keep it as you have it.

src/sage/categories/simplicial_sets.py
jhpalmieri: Okay.

src/sage/categories/simplicial_sets.py
jhpalmieri: I'm talking about (a) a line that evaluates `Y._canonical_twisting_operator()` as compared to (b) a line that does `d2 = Y._canonical_twisting_operator()` and then evaluates `d2`. Are you saying that those behaved differently? If so, it would be good to add a comment explaining that, or else someone may come along later, not understand their purpose, and delete those tests.

miguelmarco: Oh sorry, you are right. What worked differently was the example with `X`. The lines you mention are indeed unnecessarily duplicated.

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +581,414 @@ def universal_cover(self):
...
+                singred = singular_function("reduce")
+                singlift = singular_function("lift")
+                G, d = self._universal_cover_dict()
+                phi = G.abelianization_map()
+                abelG, R, I, images = G.abelianization_to_algebra(ZZ)
+                CC = self.twisted_chain_complex()
+                M1 = CC.differential(n).T
+                M2 = CC.differential(n + 1).T
+
+                def convert_to_polynomial(p):
```
enriqueartal, line 906: Would it be better to define a function that convert to a polynomial and also take out positive powers of the variables?

src/sage/categories/simplicial_sets.py
```diff
@@ -574,6 +581,414 @@ def universal_cover(self):
...
+                    if hasattr(p, "lift"):
+                        return p.lift()._as_extended_polynomial()
+                    else:
+                        return p._as_extended_polynomial()
+                M1 = M1.apply_map(convert_to_polynomial)
```
enriqueartal, line 911: I guess this only gives a matrix of the differential if it is applied column-or-row-wise

enriqueartal (commented):
There are some small comments.

enriqueartal (approved):

vbraun:
merge conflict

miguelmarco:
> merge conflict

Fixed.

PR #36694, opened 2023-11-10
CI conda: On `pull_request`, only run 1 macOS job and 1 Linux job

mkoeppe:
as discussed in https://github.com/sagemath/sage/pull/36678#issuecomment-1806192342, https://github.com/sagemath/sage/pull/36616#issuecomment-1811508709

On pushes to a tag, all 3 Linux and 3 macOS jobs are still run, as demonstrated in https://github.com/mkoeppe/sage/actions/runs/6829619271

After merging, the branch protection rule https://github.com/sagemath/sage/settings/branch_protection_rules/33965567 will need to be updated. It controls what workflows show as "Required" in the PR Checks.

### :memo: Checklist

### DIFF
--- a/.github/workflows/ci-conda.yml
+++ b/.github/workflows/ci-conda.yml
@@ -22,11 +22,18 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu, macos]
+        os: >-
+          ${{ github.event_name == 'pull_request'
+                && fromJson('["ubuntu"]')
+                || fromJson('["ubuntu, "macos]') }}
         python: ['3.9', '3.10', '3.11']
         # Optional environment is disabled for now as its not yet working
         # environment: [environment, environment-optional]
         conda-env: [environment]
+        include: >-
+          ${{ github.event_name == 'pull_request'
+                && fromJson('[{"os": "macos, "python": "3.11", "conda-env": "environment"}]')
+                || fromJson('[]') }}
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci-conda.yml
+++ b/.github/workflows/ci-conda.yml
@@ -22,17 +22,25 @@ jobs:
 
     strategy:
       matrix:
+        # On pushes to tags or branches, test the whole matrix.
         os: >-
           ${{ github.event_name == 'pull_request'
                 && fromJson('["ubuntu"]')
-                || fromJson('["ubuntu, "macos]') }}
-        python: ['3.9', '3.10', '3.11']
+                || fromJson('["ubuntu", "macos"]') }}
+        python: >-
+          ${{ github.event_name == 'pull_request'
+                && fromJson('["3.9"]')
+                || fromJson('["3.9", "3.10", "3.11"]') }}
         # Optional environment is disabled for now as its not yet working
         # environment: [environment, environment-optional]
         conda-env: [environment]
+        # On pull requests, only test two jobs:
+        # Ubuntu with Python 3.9, macOS with Python 3.10.
+        # Build & Test currently uses Python 3.11 (on ubuntu-focal).
+        # Together, they cover the supported minor Python versions.
         include: >-
           ${{ github.event_name == 'pull_request'
-                && fromJson('[{"os": "macos, "python": "3.11", "conda-env": "environment"}]')
+                && fromJson('[{"os": "macos", "python": "3.10", "conda-env": "environment"}]')
                 || fromJson('[]') }}
 
     steps:


tobiasdiez (dismissed):
Except for a the time after a release, there is no shortage of macos runners. So I don't think we need to change anything there atm. (Also in my experience there was more of a difference between the behavior on different python versions, e.g. some of the failing tests only occurred under python 3.11 but not on other python versions)

mkoeppe:
What exactly is the change you're requesting here?

mkoeppe:
> Except for a the time after a release, there is no shortage of macos runners. 

This is clearly false. For example, right now (the macOS jobs from the release have finished already), https://github.com/sagemath/sage/actions/workflows/ci-conda.yml?query=is%3Aqueued shows 16 workflows waiting for 36 macOS runners. 

> in my experience there was more of a difference between the behavior on different python versions, e.g. some of the failing tests only occurred under python 3.11 but not on other python versions)

That doesn't imply that all python versions have to run on all PRs.

mkoeppe:
I think we could also reduce Linux in PRs to one Python version.
Since Build & Test currently uses Python 3.11, we could do Linux with Python 3.9 and macOS with Python 3.10, so that we still have the nice coverage of the supported minor Python versions.

mkoeppe:
> I think we could also reduce Linux in PRs to one Python version.
> Since Build & Test currently uses Python 3.11, we could do conda Linux with Python 3.9 and conda macOS with Python 3.10, so that we still have the nice coverage of the supported minor Python versions.

Done in 9ff04ef3a49

kwankyu (approved):
Thanks. LGTM.

tobiasdiez:
As described above, I consider this a huge step backwards.

mkoeppe:
> As described above, I consider this a huge step backwards.

No, Tobias, you haven't said anything like that.

kwankyu:
Currently, conda ci is the main source that breaks and delays the checks.

Let's get this in.

tobiasdiez:
> Currently, conda ci is the main source that breaks and delays the checks.

This PR is not helping with fixing any of the random breaks. In contrast, it makes it very hard to spot if a failure is random or actual a result of the changes in the PR.

Let me know if you see consistent failures at https://github.com/sagemath/sage/pull/36372 and I'll add exclude them (or better yet, provide a fix).

If the conda workflow is too slow, I propose we remove the "max-parallel" constraint. Would also help with discovering the last random failures.

mkoeppe:
> very hard to spot if a failure is random or actual a result of the changes in the PR.

This is not a problem in actual practice that only uses the Build&Test workflow.

mkoeppe:
> If the conda workflow is too slow, I propose we remove the "max-parallel" constraint.

The problem is not that it is too slow in the sense that for a given isolated PR it takes long to completion, so changing "max-parallel" does not help. The problem is that launching 3 Linux jobs and 3 macOS jobs, each of an hour runtime, for each PR creates long backlogs in particular for the macOS runners, as I said in https://github.com/sagemath/sage/pull/36694#issuecomment-1806928427

mkoeppe:
> This PR is not helping with fixing any of the random breaks.

Running fewer instances of the CI conda on PRs does reduce the amplification of the error probability from random failures.

tobiasdiez:
> > very hard to spot if a failure is random or actual a result of the changes in the PR.
> 
> This is not a problem in actual practice that only uses the Build&Test workflow.

Do you remember the time after I've introduced the "Build & Test" workflow on trac? It was constantly red because of newly discovered random failures. Just give the conda workflow a bit more time, and it will be mostly green (and fail only due to problems with the PR at hand).

> > If the conda workflow is too slow, I propose we remove the "max-parallel" constraint.
> 
> The problem is not that it is too slow in the sense that for a given isolated PR it takes long to completion, so changing "max-parallel" does not help. The problem is that launching 3 Linux jobs and 3 macOS jobs, each of an hour runtime, for each PR creates long backlogs in particular for the macOS runners, as I said in [#36694 (comment)](https://github.com/sagemath/sage/pull/36694#issuecomment-1806928427)

Based on https://github.com/sagemath/sage/actions/workflows/ci-conda.yml?query=is%3Asuccess, the usual executation time of the conda tests are between 4 to 5 hours. In the last 50 runs, there where 5 runs with execuation times over 6h (and 3 of them where close to a release). After a release, there can indeed be longer waiting times - but that can hardly be blamed on the conda runs. Removing "max-parallel" would already help to reduce it to an average runtime in the range of 2h.

kwankyu:
> Based on https://github.com/sagemath/sage/actions/workflows/ci-conda.yml?query=is%3Asuccess, the usual executation time of the conda tests are between 4 to 5 hours. In the last 50 runs, there where 5 runs with execuation times over 6h (and 3 of them where close to a release). After a release, there can indeed be longer waiting times - but that can hardly be blamed on the conda runs. Removing "max-parallel" would already help to reduce it to an average runtime in the range of 2h.

Adding the waiting time in the queue, conda test would delay checks result a few hours. When the conda test fails, it does not help the PR author and the reviewer since they have nothing to do but to just ignore it.  Isn't this the current situation?

mkoeppe:
> > > very hard to spot if a failure is random or actual a result of the changes in the PR.
> > 
> > This is not a problem in actual practice that only uses the Build&Test workflow.
> 
> Do you remember the time after I've introduced the "Build & Test" workflow on trac? It was constantly red because of newly discovered random failures.

Yes, I remember everything. Also the tests on the patchbot were suffering from random test failures. And no, it is not hard for our developers to spot whether a failure is random or not. One does not need to run 6 additional workflows for a PR for that.

tobiasdiez:
> Adding the waiting time in the queue, conda test would delay checks result a few hours. When the conda test fails, it does not help the PR author and the reviewer since they have nothing to do but to just ignore it. Isn't this the current situation?

Nope, most failures right now are actual problems, see e.g: https://github.com/sagemath/sage/pull/36765

> And no, it is not hard for our developers to spot whether a failure is random or not. 

As you did in https://github.com/sagemath/sage/pull/36765#pullrequestreview-1748526453?

tobiasdiez:
> If the conda workflow is too slow, I propose we remove the "max-parallel" constraint. Would also help with discovering the last random failures.

That's now https://github.com/sagemath/sage/pull/36767.

mkoeppe:
> > And no, it is not hard for our developers to spot whether a failure is random or not.
> 
> As you did in [#36765 (review)](https://github.com/sagemath/sage/pull/36765#pullrequestreview-1748526453)?

Would you explain what you mean by this comment?

vbraun:
Editor decision

CI is going to be more useful if it gives results faster. Its important to search for randomly failing tests, but that should be separate from CI.

tobiasdiez:
@vbraun Please note that the execution times posted above are outdated. With https://github.com/sagemath/sage/pull/36767 the situation improved dramatically. Apart from the time after a release (where runners are heavily used by the portability tests), all three linux runs always finish around the 1h mark. In fact, since we have 60 runners, we would need to have pushes to 12 PRs per hour (one PR triggers 5 linux workflows) in order for the runs to pile up. The situation for macos is similar, although they do sometimes pile up a little bit. But, generally, they do finish around the 3h/4h mark, which is only a bit slower than the Build & Test run. You can see the latest run times at https://github.com/sagemath/sage/actions/workflows/ci-conda.yml.

> Its important to search for randomly failing tests, but that should be separate from CI.

In theory, I very much agree. In practice, however, the free CI github infrastructure is the only place for reproducible and convenient mass-scale tests we have at the moment. In fact, me watching the conda runs has lead to identifying and fixing a lot of random test failures (all of which don't seem to be conda specific), see https://github.com/sagemath/sage/issues?q=is%3Aissue+author%3Atobiasdiez+. Randomization as a means to increase test coverage naturally has the tradeoff to trigger sporadic failures in situations where you don't want them.

mkoeppe:
> all of which don't seem to be conda specific

If they are not conda specific, why would you need to watch conda workflows to catch them? 
(I don't need an answer.)

tobiasdiez:
> > all of which don't seem to be conda specific
> 
> If they are not conda specific

As you confirmed yourself (https://github.com/sagemath/sage/issues/37071, https://github.com/sagemath/sage/issues/37068) and eg in https://github.com/sagemath/sage/pull/37360.

> why would you need to watch conda workflows to catch them?

The conda workflow is very convenient for this purpose since we don't spend the majority of the time building dependencies (as in your portability runs). So they run very fast (1h for build and testing). But, of course, if you have a nice grant to pay for computational resources, I would very much appreciate if you could setup a stress test with conda or without, and report all issues you find that way.

mkoeppe:
OK, sounds we are all in agreement that it makes no sense whatsoever to run this activity as part of the CI conda on PR.

Thanks, Volker.

PR #36696, opened 2023-11-10
Fix SSL timeout in doctest / internet feature

tornaria:
Recently I've been getting timeouts in a couple of doctests in `src/sage/doctest/control.py`, including earlier today when testing 10.2.rc1

This has two causes:
1. The test for the internet feature doesn't catch a `TimeoutError`.
2. These two doctests cause internet feature to be tested.

The first one is clearly a bug: sagemath only expects `urllib.error.URLError`, but in some cases urllib raises `TimeoutError`. Arguably a bug in urlib, but easier to fix (or workaround) on our side. Done in the first commit of this PR.

The second one is IMO a bug, since I didn't use `--optional=internet` no test should hit internet (in particular, should not test internet feature).

These two doctests are meant to test options `--hide=all` and `--hide=optional`. For some reason the semantics of these include testing for the internet feature. I believe the semantics of these options should be similar to `--optional=all`, that is, exclude "external" software  (internet feature is considered "external").

That's what I implemented in the second commit in this PR.

### DIFF
--- a/src/sage/doctest/control.py
+++ b/src/sage/doctest/control.py
@@ -457,11 +457,15 @@ class DocTestController(SageObject):
                 options.hide.discard('all')
                 from sage.features.all import all_features
                 feature_names = {f.name for f in all_features() if not f.is_standard()}
+                from sage.doctest.external import external_software
+                feature_names.difference_update(external_software)
                 options.hide = options.hide.union(feature_names)
             if 'optional' in options.hide:
                 options.hide.discard('optional')
                 from sage.features.all import all_features
                 feature_names = {f.name for f in all_features() if f.is_optional()}
+                from sage.doctest.external import external_software
+                feature_names.difference_update(external_software)
                 options.hide = options.hide.union(feature_names)
 
         options.disabled_optional = set()
--- a/src/sage/features/internet.py
+++ b/src/sage/features/internet.py
@@ -56,7 +56,7 @@ class Internet(Feature):
         try:
             urlopen(req, timeout=1, context=default_context())
             return FeatureTestResult(self, True)
-        except urllib.error.URLError:
+        except (urllib.error.URLError, TimeoutError):
             return FeatureTestResult(self, False)
 
 


tornaria:
Note that this is not trivial to test, since the timeout bug is hard to reproduce (it triggers *only* if the timeout happens after the https request is made and before the response is read back, or something like that).

A reasonable way to test the second commit is to temporarily patch the method `Internet._is_present()` in `src/sage/features/internet.py` to unconditionally `raise TimeoutError`.

tornaria:
Are the following failures my fault?
```
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/combinat/degree_sequences.pyx  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/combinat/tutorial.py  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/comparability.pyx  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/connectivity.pyx  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/digraph_generators.py  # 8 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/generic_graph.py  # 5 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/genus.pyx  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/graph.py  # 3 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/graph_coloring.pyx  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/graph_generators.py  # 17 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/hypergraph_generators.py  # 9 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/repl/rich_output/pretty_print.py  # 3 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/sets/set_from_iterator.py  # 15 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/tests/books/computational-mathematics-with-sagemath/combinat_doctest.py  # 1 doctest failed
```

tornaria:
I'll assume is some of this shit:
```
Applying: Merge https://github.com/sagemath/sage/pull/36671
Applying: Merge https://github.com/sagemath/sage/pull/36529
Applying: Merge https://github.com/sagemath/sage/pull/36372
```
Heck, these are moving targets and the log doesn't even print the commit hash that was merged!

Bear in mind this PR was pushed within a day of 10.2.rc1 being released.

#36372 goes to show how broken this act is: there is a disagreement on the status of that PR, and CI is being hold hostage.

mkoeppe:
> Are the following failures my fault?
> 
> ```
> ----------------------------------------------------------------------
> sage -t --long --random-seed=286735480429121101562228604801325644303 sage/combinat/degree_sequences.pyx  # 1 doctest failed
> sage -t --long --random-seed=286735480429121101562228604801325644303 sage/combinat/tutorial.py  # 1 doctest failed
> sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/comparability.pyx  # 1 doctest failed
> sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/connectivity.pyx  # 2 doctests failed
> ```

No, they also show up in other PRs. A problem with nauty.

mkoeppe:
> I'll assume is some of this shit:
> 
> ```
> Applying: Merge https://github.com/sagemath/sage/pull/36671
> Applying: Merge https://github.com/sagemath/sage/pull/36529
> Applying: Merge https://github.com/sagemath/sage/pull/36372
> ```

I don't think so.

> Heck, these are moving targets and the log doesn't even print the commit hash that was merged!

Actually the log does print all that information, just not where you looked. https://github.com/sagemath/sage/actions/runs/6830550892/job/18578658840?pr=36696#step:3:20

The applied patches are also in the "upstream" artifact - https://github.com/sagemath/sage/actions/runs/6830550892?pr=36696

tornaria:
> > I'll assume is some of this shit:
> > ```
> > Applying: Merge https://github.com/sagemath/sage/pull/36671
> > Applying: Merge https://github.com/sagemath/sage/pull/36529
> > Applying: Merge https://github.com/sagemath/sage/pull/36372
> > ```
> 
> I don't think so.
> 
> > Heck, these are moving targets and the log doesn't even print the commit hash that was merged!
> 
> Actually the log does print all that information, just not where you looked. https://github.com/sagemath/sage/actions/runs/6830550892/job/18578658840?pr=36696#step:3:20
> 

If you mean this:
```
commit 3d28c298df87999cb38a8b94aa05270aaff0550f
Author: ci-sage workflow <ci-sage@example.com>
Date:   Sat Nov 11 00:56:50 2023 +0000

    Merge https://github.com/sagemath/sage/pull/36372

commit 768cf3ff76a63ffb38b0b4f9b9b0b6a089e89695
Author: ci-sage workflow <ci-sage@example.com>
Date:   Sat Nov 11 00:56:49 2023 +0000

    Merge https://github.com/sagemath/sage/pull/36529

commit 1520c1f6571f9aeb6a16f270a7d53d68c35ffd76
Author: ci-sage workflow <ci-sage@example.com>
Date:   Sat Nov 11 00:56:46 2023 +0000

    Merge https://github.com/sagemath/sage/pull/36671
```

There's not enough information here. These are the merge commits, but I don't expect they will be anywhere I can find them (i.e., in the PR).

Usually when I do `git log --first-parent` I get something like
```
$ git log --first-parent develop..
commit af8dbaa6977dadf0ab9b1e59e68930f084cbfa7c (HEAD -> develop-merge)
Merge: 8aa61ce6392 3f92c49026b
Author: Gonzalo Tornaría <tornaria@cmat.edu.uy>
Date:   Fri Nov 10 20:28:33 2023 -0300

    Merge PR #36696: Fix SSL timeout

[...]
```
At least this way I can see 3f92c49026b is the commit that was merged, and I can find it in the PR (in fact it's this PR). Let's compare with 3d28c298df87999cb38a8b94aa05270aaff0550f <--- see, not even GH knows about this commit.

> The applied patches are also in the "upstream" artifact - https://github.com/sagemath/sage/actions/runs/6830550892?pr=36696

That's so much friction that I'm already sweating (no good since summer is coming here in the south). Also, I worry that different jobs may be started at different times; so it could be that different jobs in the same check merge different commits (e.g. in case some blocker pr changes in between). For the same reason it'd be nice to print the same (complete) information in every job, and don't print incomplete information anywhere.


If showing the `Merge:` line is not feasible, or just for bonus points, the commit message for the merge could include a link to the parent commit, something like

`Merge https://github.com/sagemath/sage/pull/36696 @ https://github.com/sagemath/sage/commit/3f92c49026b16f94da78240066e41ef6f27d9aa0`

which would be hopefully clickable (needs the full URL since only complete URLs are clickable in logs afaik). Or else, just print a link to the PR + link to the commit instead of only a link to the PR.

My constructive suggestion above should not be confused with me endorsing this scheme, which I still think it's too brittle (both from technical *and* social sides).

tornaria:
> > Are the following failures my fault?
> > ```
> > ----------------------------------------------------------------------
> > sage -t --long --random-seed=286735480429121101562228604801325644303 sage/combinat/degree_sequences.pyx  # 1 doctest failed
> > sage -t --long --random-seed=286735480429121101562228604801325644303 sage/combinat/tutorial.py  # 1 doctest failed
> > sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/comparability.pyx  # 1 doctest failed
> > sage -t --long --random-seed=286735480429121101562228604801325644303 sage/graphs/connectivity.pyx  # 2 doctests failed
> > ```
> 
> No, they also show up in other PRs. A problem with nauty.

CI seems to be fixed now (see https://github.com/sagemath/sage/pull/36702).

Shall I then restart this check? (only way I know is to amend and force push).

mkoeppe:
> CI seems to be fixed now (see #36702).

No, I think it's a stochastically appearing problem. My current guess is that the recently updated nauty does not respect SAGE_FAT_BINARY and thus it depends on CPU details of the runner

mkoeppe:
> > Actually the log does print all that information, just not where you looked. https://github.com/sagemath/sage/actions/runs/6830550892/job/18578658840?pr=36696#step:3:20
> 
> If you mean this:
> 
> ```
> commit 3d28c298df87999cb38a8b94aa05270aaff0550f
> Author: ci-sage workflow <ci-sage@example.com>
> Date:   Sat Nov 11 00:56:50 2023 +0000
> 
>     Merge https://github.com/sagemath/sage/pull/36372
> [...]
> ```
> 
> There's not enough information here. 

You are right.

> [...] Usually when I do `git log --first-parent` I get something like
> 
> ```
> $ git log --first-parent develop..
> commit af8dbaa6977dadf0ab9b1e59e68930f084cbfa7c (HEAD -> develop-merge)
> Merge: 8aa61ce6392 3f92c49026b
> Author: Gonzalo Tornaría <tornaria@cmat.edu.uy>
> Date:   Fri Nov 10 20:28:33 2023 -0300
> 
>     Merge PR #36696: Fix SSL timeout
> 
> [...]
> ```
> 
> At least this way I can see [3f92c49](https://github.com/sagemath/sage/commit/3f92c49026b16f94da78240066e41ef6f27d9aa0) is the commit that was merged, and I can find it in the PR (in fact it's this PR). Let's compare with 3d28c298df87999cb38a8b94aa05270aaff0550f <--- see, not even GH knows about this commit.

OK, I'll see what I can do.

> > The applied patches are also in the "upstream" artifact - https://github.com/sagemath/sage/actions/runs/6830550892?pr=36696
> 
> That's so much friction that I'm already sweating [...]

I agree, it's far from convenient.

> If showing the `Merge:` line is not feasible, or just for bonus points, the commit message for the merge could include a link to the parent commit, something like
> 
> `Merge https://github.com/sagemath/sage/pull/36696 @ https://github.com/sagemath/sage/commit/3f92c49026b16f94da78240066e41ef6f27d9aa0`
> 
> which would be hopefully clickable (needs the full URL since only complete URLs are clickable in logs afaik). Or else, just print a link to the PR + link to the commit instead of only a link to the PR.

Good idea and easy to do.

tornaria:
> > CI seems to be fixed now (see #36702).
> 
> No, I think it's a stochastically appearing problem. My current guess is that the recently updated nauty does not respect SAGE_FAT_BINARY and thus it depends on CPU details of the runner

Ah, that's easy to fix: add `--enable-generic --disable-popcnt` to configure args. See: https://github.com/void-linux/void-packages/blob/master/srcpkgs/nauty/template

tornaria:
> > > CI seems to be fixed now (see #36702).
> > 
> > 
> > No, I think it's a stochastically appearing problem. My current guess is that the recently updated nauty does not respect SAGE_FAT_BINARY and thus it depends on CPU details of the runner
> 
> Ah, that's easy to fix: add `--enable-generic --disable-popcnt` to configure args. See: https://github.com/void-linux/void-packages/blob/master/srcpkgs/nauty/template

I mean possibly. I didn't test.

Also note `--disable-popcnt` might be a bit too much. In void we are supposed to support x86_64 baseline which has only SSE2, but I guess all the runners should be at least sse4 + popcnt (iirc nehalem already has popcnt). But the `--enable-generic` might be important, e.g. if the builder supports avx-512 but the runner does not.

mkoeppe:
Thanks a lot! Happening in #36707

mkoeppe:
Not sure if your change to `--hide` is the right fix. If I understand it correctly, the actual problem is that the code below unconditionally runs `is_present` even for external features. Here's a fix that changes this by going through `AvailableSoftware.__contains__`
```
diff --git a/src/sage/doctest/control.py b/src/sage/doctest/control.py
index 8583d7a447d..297da6133ce 100644
--- a/src/sage/doctest/control.py
+++ b/src/sage/doctest/control.py
@@ -1502,18 +1502,14 @@ class DocTestController(SageObject):
             available_software._allow_external = self.options.optional is True or 'external' in self.options.optional
 
             for h in self.options.hide:
-                try:
+                if h in available_software:
                     i = available_software._indices[h]
-                except KeyError:
-                    pass
-                else:
                     f = available_software._features[i]
-                    if f.is_present():
-                        f.hide()
-                        self.options.hidden_features.add(f)
-                        for g in f.joined_features():
-                            if g.name in self.options.optional:
-                                self.options.optional.discard(g.name)
+                    f.hide()
+                    self.options.hidden_features.add(f)
+                    for g in f.joined_features():
+                        if g.name in self.options.optional:
+                            self.options.optional.discard(g.name)
 
             for o in self.options.disabled_optional:
                 try:
```

tornaria:
> Not sure if your change to `--hide` is the right fix. If I understand it correctly, the actual problem is that the code below unconditionally runs `is_present` even for external features. Here's a fix that changes this by going through `AvailableSoftware.__contains__`

Indeed, that is the problema, and this solution feels much better. I've replaced my commit by yours.

mkoeppe:
Doctest failure at https://github.com/sagemath/sage/actions/runs/6837912542/job/18594471128?pr=36696#step:14:29

tornaria:
> Doctest failure at https://github.com/sagemath/sage/actions/runs/6837912542/job/18594471128?pr=36696#step:14:29

The logic is not working properly. It seems meataxe is not being hide (in my local test this didn't hit me because I don't have meataxe installed). IOW, I think the reason this gives "5 tests" instead of "4 tests" is because it's actually running the last test (labeled "optional meataxe") in spite of the hidden meataxe.

It's possible that it's another bug so that calling `f.hide()` before `f.is_present()` does not work (I can't imagine another cause).

I'm inclined to revert to my previous workaround, since this looks like a slippery slope (cf. a series of unfortunate events vol 10).

mkoeppe:
> I'm inclined to revert to my previous workaround

That would be fine with me for getting 10.2 out

mkoeppe:
Downgrading until fixed

tornaria:
> Downgrading until fixed

That feels a bit rude. What is the point?

I've reverted the branch to 4a70c1e which is what was tested before.

mkoeppe:
> > Downgrading until fixed
> 
> That feels a bit rude. What is the point?

Sorry, just so that the doctest failure doesn't show up in the CI runs of all PRs.

mkoeppe:
Let's get this in.

tornaria:
> > > Downgrading until fixed
> > 
> > 
> > That feels a bit rude. What is the point?
> 
> Sorry, just so that the doctest failure doesn't show up in the CI runs of all PRs.

I assumed "blocker" without "positive review" is not automatically merged, is it? If it is, you did the right thing here of course, but it kind of violates the principle of least surprise.

Thanks for the review.

mkoeppe:
> I assumed "blocker" without "positive review" is not automatically merged, is it? 

It is, actually. 
The "positive review" label is currently not very stable. That's part of the weakness of this mechanism social side that you mentioned in https://github.com/sagemath/sage/pull/36696#issuecomment-1806916614

tornaria:
> > I assumed "blocker" without "positive review" is not automatically merged, is it?
> 
> It is, actually. The "positive review" label is currently not very stable. That's part of the weakness of this mechanism social side that you mentioned in [#36696 (comment)](https://github.com/sagemath/sage/pull/36696#issuecomment-1806916614)

That's a problem, because when you do something completely correct from a technical point of view, it could still feel a little bit rude from a social point of view, specially if the person involved is not 

Wouldn't it make sense to have a separate label for "hotfixes" ?

Although maybe having a "hotfix" branch (with some person in charge) would be better. Maybe call it `main` and make it the default branch (so workflows issues can actually be fixed). Or even do it in `develop`. This might need some coordination with the release manager.

mkoeppe:
> > > I assumed "blocker" without "positive review" is not automatically merged, is it?
> > 
> > It is, actually. The "positive review" label is currently not very stable. That's part of the weakness of this mechanism social side that you mentioned in [#36696 (comment)](https://github.com/sagemath/sage/pull/36696#issuecomment-1806916614)
> 
> That's a problem, because when you do something completely correct from a technical point of view, it could still feel a little bit rude from a social point of view, specially if the person involved is not aware of the reason.

I agree, so the best practice should be to always include a clear comment when relabeling. I've updated https://github.com/sagemath/sage/wiki/Sage-10.2-Release-Tour#open-blocker-prs-are-applied-automatically-in-ci-workflows, take a look

soehms:
Sorry I wasn't able to attend at the weekend.

> It's possible that it's another bug so that calling `f.hide()` before `f.is_present()` does not work (I can't imagine another cause).

Yes! The sole purpose of calling `is_present`  in `control.py` is informational, i.e. restricting the content of the `Hidden features:...` message. Therefore, I will propose a smarter implementation for this in a new PR, which (incidentally) provides more accurate information, i.e. `Features that have been hidden` instead of `Hidden features`. More precisely, I will make the feature class count calls to the `is_present` method in a state where the instance is hidden and present at the same time. This can then be used to filter the message content.

soehms:
> Sorry I wasn't able to attend at the weekend.
> 
> > It's possible that it's another bug so that calling `f.hide()` before `f.is_present()` does not work (I can't imagine another cause).
> 
> Yes! The sole purpose of calling `is_present` in `control.py` is informational, i.e. restricting the content of the `Hidden features:...` message. Therefore, I will propose a smarter implementation for this in a new PR, which (incidentally) provides more accurate information, i.e. `Features that have been hidden` instead of `Hidden features`. More precisely, I will make the feature class count calls to the `is_present` method in a state where the instance is hidden and present at the same time. This can then be used to filter the message content.

This is https://github.com/sagemath/sage/pull/36741!

PR #36700, opened 2023-11-11
use less _element_constructor

fchapoton:
trying to avoid ` _element_constructor` with no final underscore

the only serious change is the one in finite fields morphisms.

### :memo: Checklist


src/sage/combinat/posets/hasse_cython.pyx
```diff
@@ -71,7 +71,7 @@ class IncreasingChains(RecursivelyEnumeratedSet_forest):
             self._greater_than = positions
             self._vertices = list(range(n))
 
-        self._element_constructor = element_constructor
+        self._constructor_ = element_constructor
```
mkoeppe: This name `_constructor_` so far does not appear anywhere in our codebase. Why is this an improvement?

fchapoton: because it is now a specific name used only in this context, not related to any general mechanism

mkoeppe: But why use `_constructor_`, not `_constructor`?

fchapoton: better now ?

src/sage/combinat/posets/hasse_cython.pyx
```diff
@@ -30,8 +30,8 @@ class IncreasingChains(RecursivelyEnumeratedSet_forest):
     - ``positions`` -- a list of sets of integers describing the poset,
       as given by the lazy attribute ``_leq_storage`` of Hasse diagrams.
 
-    - ``element_constructor`` -- used to determine the type of chains,
-      for example ``list`` or ``tuple``
+    - ``constructor`` -- used to determine the type of chains,
+      for example :class:`list` or :class:`tuple`
```
mkoeppe, line 34: Can this be renamed without deprecation for the old name?

fchapoton: this is not a named argument, just an argument

mkoeppe: All arguments are named arguments unless you use `, /,`. Users can do `D = IncreasingChains([{0,1},{1}], element_constructor=list, exclude=[]); D`

fchapoton: We have `rename_keyword` but only for keywords, not for args. If you insist, I can make an ad-hoc warning, but I would say it's ok not to deprecate. In particular, it's not in the global namespace.

mkoeppe: maybe the name of the parameter can just be left as is?

fchapoton: ok, done

mkoeppe (approved):
LGTM, thanks.

PR #36711, opened 2023-11-12
CI macOS: Build following stages even when some job failed; adjust platforms

mkoeppe:
To fix what can be seen in https://github.com/sagemath/sage/actions/runs/6827163283 - stage-2 was not attempted because some of the stage-1 jobs failed.

We also replace one of the failing platforms and remove another.

### :memo: Checklist

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -40,30 +40,35 @@ jobs:
     with:
       stage: "2"
     needs: [stage-1]
+    if: ${{ success() || failure() }}
 
   stage-2-optional-0-o:
     uses: ./.github/workflows/macos.yml
     with:
       stage: "2-optional-0-o"
     needs: [stage-2]
+    if: ${{ success() || failure() }}
 
   stage-2-optional-p-z:
     uses: ./.github/workflows/macos.yml
     with:
       stage: "2-optional-p-z"
     needs: [stage-2-optional-0-o]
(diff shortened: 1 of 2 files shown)


jhpalmieri:
What are stages 1 and 2? I have a guess about `stage-2-optional-p-z` or `stage-2-experimental-0-o`, for example, but what are `stage-1` and `stage-2`?

mkoeppe:
Stage 1 builds non-Python packages, stage 2 builds Python packages

.github/workflows/ci-macos.yml
```diff
@@ -40,30 +40,35 @@ jobs:
     with:
       stage: "2"
     needs: [stage-1]
+    if: ${{ success() || failure() }}
```
jhpalmieri, line 43: Is this equivalent to `if: ${{ always() }}`? (I found that at https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#example-not-requiring-successful-dependent-jobs.) Any reason to prefer one approach to the other?

mkoeppe: We just got rid of a lot of `always()` in:
- #36617

jhpalmieri:
Okay, looks good.

jhpalmieri (approved):

mkoeppe:
Thanks! 

Setting it to blocker so it gets merged in the next rc

PR #36718, opened 2023-11-14
Adding cellular bases for the SymmetricGroupAlgebra over general fields

tscrim:
Currently the SymmetricGroupAlgebra uses the seminormal basis for computing a cell basis, but this only is guaranteed to work over a field where Maschke's theorem holds. However, it is known that the Kazhdan-Lusztig C' basis at q = 1 for the type A Hecke algebra is a cell basis, We implement that as the fallback when we cannot use the seminormal basis.

We also fix a few typos in the doc that were noticed.

### :memo: Checklist


tscrim:
@fchapoton I hope you can do the review for this. Mathematically it is just implementing the KL $C'$ basis at $q=1$. Everything else is handled by the general framework for cellular algebras already implemented.

fchapoton:
there are a few failing doctests, some trivial and other less so

tscrim:
This should fix all of the failing doctests. They were all trivial coming from the fact that the cell poset was in the reverse order.

fchapoton:
testing wihtout coxeter3 does not work:
```
KLG = SymmetricGroup(self.n)
    UnboundLocalError: local variable 'SymmetricGroup' referenced before assignment
```

tscrim:
Good catch. This should fix it.

fchapoton (approved):
works with and without coxeter3, ok

tscrim:
Merci

PR #36727, opened 2023-11-16
more consistent multiple-variable sanity check in solve()

melnikovsky:
Currently a non-symbol variable is allowed for 1D solve() but not for nD:

```
sage: var('x y');
sage: solve(x+sqrt(2),sqrt(2))
[sqrt(2) == -x]
sage: solve(y+sqrt(3),sqrt(3))
[sqrt(3) == -y]
sage: solve([x+sqrt(2),y+sqrt(3)],[sqrt(2),sqrt(3)])
...TypeError: sqrt(2) is not a valid variable....
```
Proposed patch lifts this restriction and enables things like:

```
sage: var('x y'); u=function('u')(x,y);
sage: solve(u.gradient()[0], u.gradient()[0])
[diff(u(x, y), x) == 0]
sage: solve(u.gradient()[1], u.gradient()[1])
[diff(u(x, y), y) == 0]
sage: solve(list(u.gradient()), list(u.gradient()))
[[diff(u(x, y), x) == 0, diff(u(x, y), y) == 0]]
```


fchapoton:
you need to add doctests for the new behaviour

melnikovsky:
> you need to add doctests for the new behaviour
Thanks for the feedback, done.

PR #36731, opened 2023-11-18
`build/pkgs/pillow`: Allow discovery of libjpeg, upgrade to 10.1.0

mkoeppe:
As discussed in https://groups.google.com/g/sage-devel/c/A4nY3bh1oPI @egourgoulhon 

Here we make`libjpeg` a new "recommended" dummy package with spkg-configure script -- same category as pandoc, ffmpeg, imagemagick, texlive, so it appears in the installation instructions in https://deploy-preview-36731--sagemath-tobias.netlify.app/html/en/installation/source#debian-ubuntu-package-installation (in the 3rd command line for each of the shown package systems).

Use of libjpeg can be disabled using `./configure --without-system-libjpeg`.

Everything done here would also be necessary if we decided to make libjpeg a normal standard/optional package; that is beyond the scope of this PR.

We also refine the "CI Linux incremental" workflow, which determines what packages to uninstall (so that spkg-configure is tested) or to build (important for optional packages), based on changed files. Failures seen in https://github.com/sagemath/sage/actions/runs/6914890950/job/18813211070?pr=36731#step:4:8 came from the workflow trying to build dummy packages. We suppress these now, as seen in https://github.com/sagemath/sage/actions/runs/6915792754/job/18815135590?pr=36731#step:4:8

Tests at https://github.com/mkoeppe/sage/actions/runs/6911535503 (Linux), https://github.com/mkoeppe/sage/actions/runs/6911535500 (macOS)

### :memo: Checklist


orlitzky:
Will pillow still build on a system with no libjpeg? [Their docs](https://pillow.readthedocs.io/en/latest/installation.html) imply that it won't:

> Starting with Pillow 3.0.0, libjpeg is required by default, but may be disabled with the `--disable-jpeg` flag.

Either required or disabled is better than having it be auto-detected, but if it's going to be required we should probably put it in `_prereq` and not `_recommended`.

mkoeppe:
> Will pillow still build on a system with no libjpeg? [Their docs](https://pillow.readthedocs.io/en/latest/installation.html) imply that it won't:
> 
> > Starting with Pillow 3.0.0, libjpeg is required by default, but may be disabled with the `--disable-jpeg` flag.

Thanks, I'll check

mkoeppe:
I wouldn't like to make it a required dummy package - because as Marc pointed out, it's not available on macOS (without homebrew).

mkoeppe:
> [Their docs](https://pillow.readthedocs.io/en/latest/installation.html) imply that it won't:

Code agrees - https://github.com/python-pillow/Pillow/blob/main/setup.py#L810

orlitzky:
The risk with it being auto-detected is that if you happen to have libjpeg installed temporarily for some unrelated reason, building sage (even with `--disable-libjpeg`!) will unconditionally build a copy that requires libjpeg to run. Then when you later remove libjpeg, sage will crash.

mkoeppe:
Well, that's true for pretty much any use of shared libs from system packages

orlitzky:
> Well, that's true for pretty much any use of shared libs from system packages

Yeah but with a required standard package you can at least avoid the eventual crash with  `--without-system-foo`.

mkoeppe:
That's a fair point. Not sure if a major point, but a fair point.

mkoeppe:
We do have a variable `PILLOW_CONFIG_SETTINGS` that an ephemeral-system-packages-wary user could use for locking down what libraries pillow uses.

mkoeppe:
And note that Pillow autodetects a bunch of other libraries and we are not forbidding that.

egourgoulhon (approved):
Thank you for implementing this!
LGTM. I've checked it on Ubuntu 22.04, with libjpeg-dev installed as a system package. More precisely, I've built Sage from scratch (starting from a git clone (-> Sage 10.2.rc4)  + pull of this branch) and the output of `./configure` is 
```
Checking whether SageMath should install SPKG libjpeg...
checking for libjpeg... yes
configure: will use system package and not install SPKG libjpeg
...
libjpeg:                        using system package
```
Then I've checked that running 
```
from PIL import Image
plot(sin(x)).save("fig.png")
img = Image.open("fig.png")
img
```
in a Jupyter notebook does not trigger any error (contrary to Sage 10.2.rc4), and that the subsequent code
```
img.convert("RGB").save("fig.jpg")
```
properly creates a jpeg image file. 
I've also checked that the public notebook
https://nbviewer.org/github/sagemanifolds/SageManifolds/blob/master/Notebooks/SM_black_hole_rendering.ipynb
runs without any error.

mkoeppe:
Thanks for testing!

dimpase:
> The risk with it being auto-detected is that if you happen to have libjpeg installed temporarily for some unrelated reason, building sage (even with `--disable-libjpeg`!) will unconditionally build a copy that requires libjpeg to run. Then when you later remove libjpeg, sage will crash.

In that sense, `./sage --pip install pillow` is safer - it will most probably fetch a binary wheel with everything included.

culler:
I agree with Dima.  The binary wheel is the simplest way to do this.

I wonder whether Sage should perhaps reconsider the policy about never
installing binary wheels into the Sage tree.

- Marc

mkoeppe:
> In that sense, `./sage --pip install pillow` is safer - it will most probably fetch a binary wheel with everything included.

Note that the included libraries in such binary wheels (built using auditwheel or delocate) may cause library conflicts of the same kind that we otherwise avoid using `SAGE_SPKG_DEPCHECK` files in `spkg-configure.m4` 
Not saying that I'd be overly concerned about this in the particular case of pillow, but it is something that we do not have control over in general.

culler:
In the case of libjpeg, it is not a dynamic library, but just an ar library
archive.  So I am pretty sure that the Pillow python extension is
statically linked with libjpeg.

- Marc

mkoeppe:
That's true about libjpeg (after all, we know that Pillow is the only package using it!), but the concern extends to the other libraries bundled in the wheel:
```
$ unzip Pillow-10.1.0-cp311-cp311-macosx_10_10_x86_64.whl
[...]
  inflating: PIL/.dylibs/liblcms2.2.dylib  
  inflating: PIL/.dylibs/libpng16.16.dylib  
  inflating: PIL/.dylibs/libxcb.1.1.0.dylib  
  inflating: PIL/.dylibs/libbrotlidec.1.1.0.dylib  
  inflating: PIL/.dylibs/libfreetype.6.dylib  
  inflating: PIL/.dylibs/libXau.6.0.0.dylib  
  inflating: PIL/.dylibs/libbrotlicommon.1.1.0.dylib  
  inflating: PIL/.dylibs/libharfbuzz.0.dylib  
  inflating: PIL/.dylibs/libsharpyuv.0.dylib  
  inflating: PIL/.dylibs/libz.1.3.dylib  
  inflating: PIL/.dylibs/libwebp.7.dylib  
  inflating: PIL/.dylibs/libwebpmux.3.dylib  
  inflating: PIL/.dylibs/libopenjp2.2.5.0.dylib  
  inflating: PIL/.dylibs/libwebpdemux.2.dylib  
  inflating: PIL/.dylibs/liblzma.5.dylib  
  inflating: PIL/.dylibs/libtiff.6.dylib  
[...]
```

orlitzky:
> Note that the included libraries in such binary wheels (built using auditwheel or delocate) may cause library conflicts of the same kind that we otherwise avoid using `SAGE_SPKG_DEPCHECK` files in `spkg-configure.m4` Not saying that I'd be overly concerned about this in the particular case of pillow, but it is something that we do not have control over in general.

I think libjpeg is actually pretty high-risk. It's a common system library and lots of math software needs to work with JPEG graphics for e.g. handwriting recognition. You'd have to be able to promise (a) that no other sage dependency would ever want to link to libjpeg, and (b) that the user himself wouldn't want to link a sage program to libjpeg.

mkoeppe:
@culler Of course, in preparing your macOS app, you can simply test that the resulting binary works. But for the unattended from-source builds of Sage on user systems, relying on the published wheels from PyPI introduces a lot of potential for mysterious failures.

culler:
On Mon, Nov 20, 2023 at 12:25 PM Michael Orlitzky ***@***.***>
wrote:

> I think libjpeg is actually pretty high-risk. It's a common system library
> and lots of math software needs to work with JPEG graphics for e.g.
> handwriting recognition. You'd have to be able to promise (a) that no other
> sage dependency would ever want to link to libjpeg, and (b) that the user
> himself wouldn't want to link a sage program to libjpeg.
>

I think that concern applies to dynamic libraries, not to statically linked
libraries.  Nonetheless, Matthias seems to be saying that
there are several dynamic libraries packages with the Pillow python
extension.  I know that delocate embeds .dylib files into Python packages,
but they put the libraries inside the package (in a directory named
.dylibs) and use relative rpaths.  So, if Pillow is using delocate, I don't
actually think that any other parts of sage would be able to find the
dynamic libraries packaged with Pillow.

- Marc

mkoeppe:
> I don't actually think that any other parts of sage would be able to find the dynamic libraries packaged with Pillow.

I agree with this; but I think the failure mechanism is through possible global symbol clashes of different versions of the same library.

orlitzky:
I'd have to try it to be sure. The issue isn't with finding the bundled libraries but rather with the conflicting symbol names. All of this stuff eventually gets linked into the same executable, so if two libraries (even transitively) provide the same symbols, you can get a conflict if they're not binary-compatible.

For example, the bundled pillow .so and the one on my system:
```
$ nm -gD /usr/lib/python3.11/site-packages/PIL/_imaging.cpython-311-x86_64-linux-gnu.so  | grep jpeg
                 U jpeg_CreateCompress@LIBJPEG_6.2
                 U jpeg_CreateDecompress@LIBJPEG_6.2
                 U jpeg_add_quant_table@LIBJPEG_6.2
00000000000375f0 T jpeg_buffer_dest
0000000000037090 T jpeg_buffer_src
                 U jpeg_destroy_compress@LIBJPEG_6.2
                 U jpeg_destroy_decompress@LIBJPEG_6.2
                 U jpeg_finish_compress@LIBJPEG_6.2
                 U jpeg_finish_decompress@LIBJPEG_6.2
                 U jpeg_read_header@LIBJPEG_6.2
                 U jpeg_read_scanlines@LIBJPEG_6.2
                 U jpeg_resync_to_restart@LIBJPEG_6.2
                 U jpeg_set_defaults@LIBJPEG_6.2
                 U jpeg_set_quality@LIBJPEG_6.2
                 U jpeg_simple_progression@LIBJPEG_6.2
                 U jpeg_start_compress@LIBJPEG_6.2
                 U jpeg_start_decompress@LIBJPEG_6.2
                 U jpeg_std_error@LIBJPEG_6.2
                 U jpeg_suppress_tables@LIBJPEG_6.2
                 U jpeg_write_marker@LIBJPEG_6.2
                 U jpeg_write_scanlines@LIBJPEG_6.2
0000000000064ee0 D libjpeg_turbo_version

$ nm -gD PIL/_imaging.pypy310-pp73-x86_64-linux-gnu.so | grep jpeg
                 U jpeg_CreateCompress@LIBJPEG_6.2
                 U jpeg_CreateDecompress@LIBJPEG_6.2
                 U jpeg_add_quant_table@LIBJPEG_6.2
000000000003b1f0 T jpeg_buffer_dest
000000000003ac60 T jpeg_buffer_src
                 U jpeg_destroy_compress@LIBJPEG_6.2
                 U jpeg_destroy_decompress@LIBJPEG_6.2
                 U jpeg_finish_compress@LIBJPEG_6.2
                 U jpeg_finish_decompress@LIBJPEG_6.2
                 U jpeg_read_header@LIBJPEG_6.2
                 U jpeg_read_scanlines@LIBJPEG_6.2
                 U jpeg_resync_to_restart@LIBJPEG_6.2
                 U jpeg_set_defaults@LIBJPEG_6.2
                 U jpeg_set_quality@LIBJPEG_6.2
                 U jpeg_simple_progression@LIBJPEG_6.2
                 U jpeg_start_compress@LIBJPEG_6.2
                 U jpeg_start_decompress@LIBJPEG_6.2
                 U jpeg_std_error@LIBJPEG_6.2
                 U jpeg_suppress_tables@LIBJPEG_6.2
                 U jpeg_write_marker@LIBJPEG_6.2
                 U jpeg_write_scanlines@LIBJPEG_6.2
000000000006b5e0 D libjpeg_turbo_version
```

In this case they agree, but... I'd have to try it. (This is why there are different wheels for different libc)

vbraun:
Imho this needs more testing than cramming it in at the last minute

mkoeppe:
People who use Jupyter seem to consider it a severe regression from the previous release

mkoeppe:
And this was tested on the full set of platforms, see ticket description.

egourgoulhon:
> People who use Jupyter seem to consider it a severe regression from the previous release

I confirm...

egourgoulhon:
> The issue wasn't just that you can't play with jpegs, but that some other functionality of pillow happens to be broken at the moment if you compile it without jpeg support, right?

Yes indeed. For instance, as reported in https://groups.google.com/g/sage-devel/c/A4nY3bh1oPI/m/0lxrC67HAgAJ, the following code does not manipulate any jpeg but is broken in Sage 10.2.rc4:
```
from PIL import Image
plot(sin(x)).save("fig.png")
img = Image.open("fig.png")
img
```
For some reason, it yields
```
ValueError: Could not save to JPEG for display
```
while on the user side, we are not asking for any jpeg. 
The same piece of code works nicely in Sage 10.1

orlitzky:
> while on the user side, we are not asking for any jpeg.
> The same piece of code works nicely in Sage 10.1

This is probably fixed in pillow-10.1.0: https://github.com/python-pillow/Pillow/issues/7259

If so, maybe an upgrade to pillow is a less intrusive fix?

PR #36732, opened 2023-11-18
build/pkgs/pari: require pari-2.5.14, drop tests for patched v2.5.13

### LINKED ISSUE #36729: OS X: system Pari is rejected because of unknown shell command
### Steps To Reproduce

`timeout` is not a valid shell command in OS X, so the test
```
        AC_MSG_CHECKING([whether factor() bug 2469 of pari 2.15.3 is fixed])
        result=`echo "f=factor(2^2203-1); print(\"ok\")" | timeout 1 $GP -qf`
```
introduced in #35302 fails even with a recent version of `pari` installed via `homebrew`.

### Expected Behavior

Sage should be able to use `homebrew`'s version of `pari`.

### Actual Behavior

`config.log` says
```
configure:24946: checking whether factor() bug 2469 of pari 2.15.3 is fixed
configure:24953: result: no; cannot use system pari/GP with known bug
...
configure:25276: no suitable system package found for SPKG pari
```
If I remove `timeout` from that line, then the system `pari is used, Sage builds correctly, and all tests pass. Also the shell command
```
echo "f=factor(2^2203-1); print(\"ok\")" | gp -qf 
```
executes without error and finishes very quickly.

### Additional Information

_No response_

### Environment

```markdown
- **OS**: OS X 14.1
- **Sage Version**: 10.2.rc3
```

### Checklist

orlitzky:
We currently check for either pari-2.5.14, or an unofficial (patched) version of pari-2.5.13 that is missing two bugs present in the upstream release. However, one of those bug-checks fails on macOS, making it impossible to use the system pari there.

pari-2.5.14 has been out for a few months. Any distro fast enough to have backported patches to pari-2.5.13 will likely have pari-2.5.14 by now. On that assumption, we simply require pari-2.5.14 to fix the macOS issue, and hope it does not affect anyone still using a patched pari-2.5.13.

- Also fixes #36729

### DIFF
--- a/build/pkgs/pari/spkg-configure.m4
+++ b/build/pkgs/pari/spkg-configure.m4
@@ -1,6 +1,6 @@
 SAGE_SPKG_CONFIGURE([pari], [
   dnl See gp_version below on how the version is computed from MAJV.MINV.PATCHV
-  m4_pushdef([SAGE_PARI_MINVER],["134912"])dnl this version and higher allowed
+  m4_pushdef([SAGE_PARI_MINVER],["134916"])dnl this version and higher allowed
   m4_pushdef([SAGE_PARI_MAXVER],["999999"])dnl this version and higher not allowed
   SAGE_SPKG_DEPCHECK([gmp readline], [
     AC_PATH_PROG([GP], [gp])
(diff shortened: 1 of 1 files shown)


jhpalmieri:
It looks great to me, but I really only use OS X + homebrew. I think we need feedback from people on other platforms about which versions of `pari` are standard these days.

orlitzky:
You can get some idea here: https://repology.org/project/pari/versions

The only people affected are those with a *patched* pari-2.15.3. One would have to go digging to find out which patches are applied on the few distros that still have 2.15.3.

I'm a strong proponent of supporting multiple versions of things, but patching them in sage and then requiring everyone else to apply the same patches and checking for them in `./configure` isn't the way to go about it. This version has sailed.

jhpalmieri:
> You can get some idea here: https://repology.org/project/pari/versions

What do the colors mean in that table? In particular, why are some of the 2.15.3 yellow and some red?

orlitzky:
> > You can get some idea here: https://repology.org/project/pari/versions
> 
> What do the colors mean in that table? In particular, why are some of the 2.15.3 yellow and some red?

The red ones are where no newer version is available on that distro. The yellow ones are where an old version exists, but a newer one is available too. (They key is on the repology homepage).

For example, the Fedora 38 release has a yellow v2.15.13 because 2.15.14 is also available via the updates channel.

jhpalmieri:
Okay, let's move ahead with this.

jhpalmieri (approved):

jhpalmieri:
Note: should fix #36729.

dimpase:
On Homebrew, one just has to install `coreutils` (as we do on all the Linux systems).
Then `timeout` is there.

PR #36738, opened 2023-11-19
`build/bin/sage-spkg`: Add options for checking, cleaning separately from build/install

mkoeppe:
The `sage-spkg` script installs normal and wheel packages.
We create new options that make it possible to invoke the phases of the installation separately.
```
   -d: only download the package
   -b: build and install (stage) only, do not run post-install
       or check
   -p: post-install only
   -x: exclusively run the test suite; this may assume that:
        * the package has been installed already and/or that
        * the temporary build directory has not been deleted
   -e: erase (delete) the temporary build directory only
```
This is preparation for improvements in `Makefile.in` in combination with https://github.com/sagemath/sage/pull/36740

### :memo: Checklist

### DIFF
--- a/build/bin/sage-spkg
+++ b/build/bin/sage-spkg
@@ -90,20 +90,25 @@ Search Sage's list of packages (see 'sage --package list') for a
 matching package, and if a match is found, install it.
 
 Options:
-   -s: do not delete the temporary build directory
-   -c: after installing, run the test suite for the package;
-       exit with an error on test suite failures
-   -w: after installing, run the test suite for the package;
-       print a warning on test suite failures
-   -d: only download the package
    -y: automatically reply "y" for all prompts regarding
        experimental and old-style packages; warning: there
        is no guarantee that these packages will build correctly;
        use at your own risk
    -n: automatically reply "n" for all prompts regarding
        experimental and old-style packages
+   -d: only download the package
    -o: allow fetching the package from its upstream URL
        when it is not available from the Sage mirrors (yet)
+   -b: build and stage only, do not install or check
+   -x: only run the test suite; this may assume that:
+        * the package has been installed already and/or that
+        * the temporary build directory has not been deleted
+   -c: after installing, run the test suite for the package;
+       exit with an error on test suite failures
+   -w: after installing, run the test suite for the package;
+       print a warning on test suite failures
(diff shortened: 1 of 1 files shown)


orlitzky:
No major issues but I noticed a few small things that could be improved:

1. `sage-spkg -p` doesn't work twice in a row because `unload_destdir()` removes the staged files itself when it's done. Maybe it could say "Moving package files..." instead of "Copying..."? Regardless, this is mainly confusing because the error message accuses you of deleting the _build_ directory which still exists if you used `-s`.
2. The "stage" vs "install" terminology is confusing throughout. There's no hope of fixing this, just mentioning it.
3. Conflicting options should throw an error. Some combinations do nothing (`-d -p`, `-d -x`), while others choose one "randomly" based on whichever case comes first in the script (`-p -x`, `-p -x -e`).

mkoeppe:
> 3\. Conflicting options should throw an error. Some combinations do nothing (`-d -p`, `-d -x`), while others choose one "randomly" based on whichever case comes first in the script (`-p -x`, `-p -x -e`).

Improved in d41c0320f06

mkoeppe:
> * `sage-spkg -p` doesn't work twice in a row because `unload_destdir()` removes the staged files itself when it's done. Maybe it could say "Moving package files..." instead of "Copying..."? Regardless, this is mainly confusing because the error message accuses you of deleting the _build_ directory which still exists if you used `-s`.

Thanks. I've made improvements in 756ae53a73b.

> * The "stage" vs "install" terminology is confusing throughout. There's no hope of fixing this, just mentioning it.

Indeed. But we can perhaps revisit this after #36747, which unifies handling of "script" and "normal" packages. I also want to clean up the "clean" vs "uninstall" terminology.

build/bin/sage-spkg
```diff
@@ -257,11 +279,15 @@ while true; do
...
     shift
 done
-
-
-##################################################################
-# Figure out the package filename, download it if needed.
-##################################################################
+MODE_SWITCHES=${MODE_SWITCHES#, }
+case "$MODE_SWITCHES" in
+    *,*)
```
orlitzky, line 284: This makes e.g. `-d -d` an error, but it's good enough, thanks.

orlitzky (approved):

PR #36740, opened 2023-11-20
Refactor `bootstrap-conda`, `m4/sage_spkg_collect.m4` through sage-package; handle `dependencies_build`

mkoeppe:
Using new commands `sage --package dependencies` and `sage --package properties`, we eliminate direct references to `build/pkgs/` from various scripts and reduce code duplication in reading SPKG metadata.

In particular, we simplify the `configure`-time compilation of dependency information for `build/make/Makefile` in `m4/sage_spkg_collect.m4`; this also gives a bit of a speedup when running `./configure`. 

We complement the `build/make/Makefile` variables `deps_SPKG` by new variables `build_deps_SPKG` and `check_deps_SPKG`. This is refactoring only; there should be no change in behavior (also note that no `dependencies_build` file is added here.) Together with #36738, this is also preparation for improving the parallelization of build phases of various packages. As it gives us a clearer mapping of "build-system requires" and "install requires" of Python packages, it is also preparation for automating more steps of the creation of Python SPKGs (`sage -package create --pypi`).

Review instructions:
- There is a change to the `bootstrap`-generated file `m4/sage_spkg_configures.m4`: The macro `SAGE_SPKG_COLLECT_INIT` is now called explicitly, with an argument (the full list of packages)
- To view the effective Makefile rules after macroexpansion, use `make DEBUG_RULES=1 `. (Note there is no change to the macro for script packages here; these rules will be merged into the macro for normal packages as a follow-up to #36747._

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37018 (split out from here)
- Depends on #37350 (split out from here)
- Depends on #37430 (split out from here)

### DIFF
--- a/bootstrap-conda
+++ b/bootstrap-conda
@@ -11,7 +11,7 @@ STRIP_COMMENTS="sed s/#.*//;"
 shopt -s extglob
 
 eval $(sage-package dependencies --format=shell :all:)
-DEVELOP_SPKG_PATTERN="@(_develop${runtime_deps__develop// /|})"
+DEVELOP_SPKG_PATTERN="@(_develop|${runtime_deps__develop// /|})"
 BOOTSTRAP_PACKAGES=_bootstrap
 PACKAGES=
 OPTIONAL_PACKAGES=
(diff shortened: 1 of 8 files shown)


mkoeppe:
@dimpase @orlitzky Please review

mkoeppe:
@dimpase please apply sense of duty here

PR #36745, opened 2023-11-21
pep8 for axiom and sympy interfaces

fchapoton:
pep8 cleanup for the 2 modified files

### :memo: Checklist

### DIFF
--- a/src/sage/interfaces/axiom.py
+++ b/src/sage/interfaces/axiom.py
@@ -457,7 +457,7 @@ class PanAxiom(ExtraTabCompletion, Expect):
             return out
         # out = out.lstrip()
         i = out.find('\n')
-        out = out[i + 1:]
+        out = out[i+1:]
         outs = out.split("\n")
         i = 0
         for line in outs:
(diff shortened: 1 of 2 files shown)


src/sage/interfaces/axiom.py
```diff
@@ -458,14 +457,14 @@ def _eval_line(self, line, reformat=True, allow_use_file=False,
             return out
         # out = out.lstrip()
         i = out.find('\n')
-        out = out[i+1:]
+        out = out[i + 1:]
```
mkoeppe: I think this is incorrect per PEP 8 - https://peps.python.org/pep-0008/#pet-peeves
"in a slice the colon acts like a binary operator, and should have equal amounts on either side (treating it as the operator with the lowest priority)."

mkoeppe (commented):
Otherwise LGTM

src/sage/interfaces/axiom.py
kwankyu: There is equal amounts of space on either side of `:`, namely, zero. I see no problem here.

kwankyu: Ah, you meant
```
# Correct:
ham[lower+offset : upper+offset]
# Wrong:
ham[lower + offset:upper + offset]
```
You are right. 

But personally I cannot agree with the recommendation. I like the "wrong" format :-)

mkoeppe (approved):

PR #36724, opened 2023-11-15
Compute dimensions of simple symmetric group modules over positive characteristic

tscrim:
We can compute Specht module dimensions, but an interesting open problem is to compute the dimensions of simple modules for `S_n` over fields of positive characteristic. We provide a way to compute dimensions by using the pullback of the polytabloid inner product and computing the rank of the Gram matrix.

This is based off of code provided by Jackson Walters.

We also implement a more efficient column stabilizer method for standard tableaux.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

PR #36748, opened 2023-11-22
Implement Specht modules in the tabloid basis

tscrim:
This provides an implementation of the Specht modules over arbitrary fields in the standard tableaux basis. To do so, we also implement the tabloid module, where the tabloids are implemented as `OrderedSetsPartitions`. Using this, we can construct the maximal submodule and simple submodule from a Specht module.

We also implement a method to compute the Brauer character of the representations.

### :memo: Checklist

### :hourglass: Dependencies

- #36724: Trivial conflict with working on the same files.
- #36985: So the 0 dimensional tests will pass.


src/sage/combinat/specht_module.py
```diff
@@ -18,15 +25,227 @@
...
+        The Frobenius map is defined as the map to symmetric functions
+
+        .. MATH::
+
+            F(\chi) = \frac{1}{n!} \sum_{w \in S_n} \chi(w) p_{\rho(w)},
+
+        where `\chi` is the character of the `S_n`-module ``self``,
+        `p_{\lambda}` is the powersum symmetric function basis element
+        indexed by `\lambda`, and `\rho(w)` is partition of the cycle type
+        of `w`. Specifically, this map takes irreducible representations
```
darijgr: By "and `\rho(w)` is partition of the cycle type of `w`", do you mean ``... is the cycle type of `w` as a partition?"

src/sage/combinat/specht_module.py
```diff
@@ -18,15 +25,227 @@
...
+
+        An example using the tabloid module::
+
+            sage: SGA = SymmetricGroupAlgebra(QQ, 5)
+            sage: TM = SGA.tabloid_module([2, 2, 1])
+            sage: s(TM.frobenius_image())
+            s[2, 2, 1] + s[3, 1, 1] + 2*s[3, 2] + 2*s[4, 1] + s[5]
+        """
+        from sage.combinat.sf.sf import SymmetricFunctions
+        BR = self.base_ring()
```
darijgr: Are you sure you want ``self.base_ring()`` instead of ``QQ``? I can easily find myself applying this to Spechts over `\ZZ`, and the way you compute this using the p-basis will probably not work there.

src/sage/combinat/specht_module.py
```diff
@@ -317,25 +615,385 @@ def _acted_upon_(self, x, self_on_left=False):
...
+        EXAMPLES::
+
+            sage: SGA = SymmetricGroupAlgebra(QQ, 5)
+            sage: SM = SGA.specht_module([2,2,1])
+            sage: TestSuite(SM).run()
+        """
+        self._diagram = ambient._shape
+        SymmetricGroupRepresentation.__init__(self, ambient._SGA)
+
+        from sage.combinat.symmetric_group_algebra import e
```
darijgr: Where do you need this `e`?

src/sage/combinat/specht_module.py
```diff
@@ -317,25 +615,385 @@ def _acted_upon_(self, x, self_on_left=False):
...
+        ambient_basis = ambient.basis()
```
darijgr, line 835: The ambient space is the whole tabloid module M^\lambda, right?

src/sage/combinat/specht_module.py
tscrim: Yes; I will change that on my next push.

src/sage/combinat/specht_module.py
tscrim: I guess this is a point I am slightly confused about with the Frobenius map. What is the image supposed to be? I am pretty sure you are correct with having this always be `QQ`, but is that what should happen over $F_p$ (assuming we just map irreducibles)? I don't know the precise details enough to answer this...

src/sage/combinat/specht_module.py
tscrim: No, that is a leftover from a previous implementation attempt. Will remove.

src/sage/combinat/specht_module.py
tscrim: Yes, that's correct.

src/sage/combinat/specht_module.py
darijgr: I would be quite surprised if there was a Frobenius characteristic map over `GF(2)`. After all, there is a Specht module over `GF(2)` that decomposes as a direct sum of two Specht modules (see Exercise 1 (d) in https://www.ma.rhul.ac.uk/~uvah099/Maths/Sym/SymProblems2.pdf ), but no such identity holds for Schur functions even over `GF(2)`. Maybe there are some targets better than `QQ`, but `QQ` is at the very least a good start.

tscrim: Indeed, the modular branching rule is complicated (and one of the reasons for implementing this). Perhaps this should simply be disabled unless in char 0. There are definitely reasons for allowing more generic rings (mainly polynomial rings to compute the graded multiplicities). However, for the current implementations, it is either irreducible (hence, concentrated in a single degree) nor do I know of a generic grading for the general diagrams. So everything would be ungraded. I am now convinced to do this only over `QQ`.

Something else I have been pondering is to just return as Schur functions (of course, going through the powersums for the computation). This is the most useful output, but it is not the "natural" one from the computation. What do you think?

tscrim: Changed to `QQ` with also converting the coefficients for the powersums.

tscrim:
I have addressed all of the comments and added a few more doctests for the codecov (those I felt were reasonable).

tscrim:
@darijgr ping?

mantepse:
@tscrim, is there any reason you know of that this code is not yet in sage?

darijgr:
@mantepse Lazy reviewers...

mantepse:
@darijgr, do you want to look more thoroughly?  For me it certainly looks "good enough" - I would be extremely surprised to see any fundamental flaws that cannot be fixed in follow up tickets.  What do you think?

darijgr:
I don't have time any more in this quarter :(

Martin, can you review this?

src/sage/combinat/tabloid.py
```diff
@@ -0,0 +1,320 @@
...
+from sage.typeset.ascii_art import ascii_art
+from sage.rings.integer_ring import ZZ
+
+class Tabloid(ClonableArray):
+    r"""
+    A tabloid.
+
+    A tabloid is an ordered set partition of `\{1, 2, \ldots, n\}` such that
+    the sizes of the sets are weakly decreasing (that is, they form an
+    integer partition).
```
mantepse: It might make sense to allow compositions - it shouldn't make any difference in the code and might be used in educational settings.

src/sage/combinat/specht_module.py
```diff
@@ -213,99 +455,176 @@ def _unicode_art_(self):
...
+            # Check for a scalar first
+            ret = super()._acted_upon_(x, self_on_left)
+            if ret is not None:
+                return ret
+            # Check if it is in the symmetric group algebra
+            P = self.parent()
+            if x in P._SGA or x in P._SGA.group():
+                if self_on_left:  # it is only a left module
+                    return None
+                else:
```
mantepse, line 539: can be simplified to `return`.

src/sage/combinat/specht_module.py
```diff
@@ -213,99 +455,176 @@ def _unicode_art_(self):
...
+            sage: v = sum(TM.basis())
+            sage: TM.bilinear_form(u, v)
+            7
+            sage: TM.bilinear_form(u, TM.zero())
+            0
+        """
+        if len(v) < len(u):
+            u, v = v, u
+        R = self.base_ring()
+        return R.sum(c * v[T] for T,c in u)
```
mantepse: missing blank after `,`.

mantepse (approved):
Thank you!

(I'll leave it up to you whether you want to allow compositions now or not)

src/sage/combinat/specht_module.py
tscrim: Added.

src/sage/combinat/specht_module.py
tscrim: I would prefer to leave it explicitly returning `None` to signify that it is meant to be returned as the output (rather than as a non-output).

src/sage/combinat/tabloid.py
tscrim: Basically, yes, but I would prefer to leave it only for partitions. I am not sure if this should be completely replaced by an appropriate subclass of `OrderedSetPartitions`. So I don't want to put too much more into it at this point. It also has the construction from tableaux, which currently only works for partition shape.

mantepse: I agree.  In fact, I now think that `Tabloids` should really be `OrderedSetPartitions_scomp(n, alpha)`, without any extra methods.  Maybe the name of the method `symmetric_group_action` is not ideal, do we need it at all?

For example, `__iter__` of `OrderedSetPartitions_scomp` *should* be the same as `__iter__`, so we don't need to optimize two iterators.

I'll give it positive review anyway and leave this to you.

mantepse (approved):
Thank you!

tscrim:
> Copying this from above, because apparently, once resolved, text is not shown anymore.
> 
> I agree. In fact, I now think that `Tabloids` should really be `OrderedSetPartitions_scomp(n, alpha)`, without any extra methods. Maybe the name of the method `symmetric_group_action` is not ideal, do we need it at all?

I have two reasons for this:

1. I use the `symmetric_group_action` method to build the action. Although this could be moved into the tabloid module itself.
2. I want to actually have latex/ascii_art output as tabloids. This can also be moved into the tabloid module code, but it is a less natural place.

> For example, `__iter__` of `OrderedSetPartitions_scomp` _should_ be the same as `__iter__`, so we don't need to optimize two iterators.

Indeed.

> I'll give it positive review anyway and leave this to you.

Thank you, but I seriously considering dumping the `Tabloids` class for OSPs. Its duplicating too much for no clear benefits or extensions that OSPs wouldn't be able to do themselves (with perhaps a special subclass for those whose base set is $[n]$). It seems like you are also in favor of this as well, right?

mantepse:
Yes, I am always in favour of less duplication.

mantepse:
PS:  I think eventually we should have a clear framework for combinatorial objects 'naturally defined on the set [n]'  and on arbitrary sets, that is used consistently for `SetPartitions`, `Combinations`, `Tableaux`, etc.

(and also walk through possibly misleading classes, eg., `Tuples`, and eventually deprecate them)

tscrim:
I have removed the `Tabloid(s)` classes in my latest push since they are (currently) redundant and moved the corresponding functionality into the `TabloidModule` class. I also fixed a bug with partitions not being recognized as compositions (necessary for the code now). I reshuffled the class hierarchy a bit in order to bring in more functionality from `Representation_abstract`. To get everything to work, I needed to add a `gens()` method to `StandardPermutations_n`.

src/sage/combinat/composition.py
```diff
@@ -1754,7 +1756,7 @@ def _element_constructor_(self, lst) -> Composition:
             sage: P([3,3,1]) # indirect doctest
             [3, 3, 1]
         """
-        if isinstance(lst, Composition):
+        if isinstance(lst, (Composition, Partition)):
```
fchapoton, line 1761: a doctest for that, maybe ?

src/sage/combinat/partition.py
```diff
@@ -5495,9 +5495,9 @@ def specht_module(self, base_ring=None):
         EXAMPLES::
 
             sage: SM = Partition([2,2,1]).specht_module(QQ); SM
-            Specht module of [(0, 0), (0, 1), (1, 0), (1, 1), (2, 0)] over Rational Field
+            Specht module of [2, 2, 1] over Rational Field
             sage: s = SymmetricFunctions(QQ).s()
```
fchapoton: can remove this line

src/sage/combinat/permutation.py
```diff
@@ -7342,6 +7342,19 @@ def cardinality(self):
         """
         return factorial(self.n)
 
+    @cached_method
+    def gens(self):
```
fchapoton: please add typing annotation `-> tuple` so that we known its is ok by grepping

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -1586,6 +1586,27 @@ def specht_module(self, D):
...
+
+            :class:`~sage.combinat.specht_module.TabloidModule`
+
+        EXAMPLES::
+
+            sage: SGA = SymmetricGroupAlgebra(QQ, 5)
+            sage: TM = SGA.tabloid_module(Partition([3,1,1]))
+            sage: TM
+            Tabloid module of [3, 1, 1] over Rational Field
+            sage: s = SymmetricFunctions(QQ).s()
```
fchapoton, line 1602: really need the conversion to Schur here ?

tscrim: Relic of the previous commits. Removes.

src/sage/combinat/composition.py
tscrim: I thought I added one here, but I forgot. (It was tested indirectly, but direct tests are good.)

tscrim:
Come gather 'round people, wherever you roam, and finish the PR review with doctest passin'.

So it was a deeper issue (and more strangely/worrisome, something I couldn't reproduce locally!). The `is_submodule` check should not have raised an error if the ambient spaces were different but instead simply returned `False` because of the generic coercion check for modules is relying on that. This is now fixed in the last commit and the doctests in `specht_module.py` now pass for me.

fchapoton (approved):
ok, let's go

tscrim:
Merci!

PR #36749, opened 2023-11-22
Choice of the name of primitive elements for subfields of QQbar

enriqueartal:
Added optional keywords `name` for the method `union` and the function `number_field_elements_from_algebraics` from `qqbar.py`. 

Without this option the name, the name is always `a` and it may conflict if  `a` is another object, see https://groups.google.com/g/sage-devel/c/mr9AzpUcbUM
### :memo: Checklist


src/sage/rings/qqbar.py
```diff
@@ -2508,6 +2509,8 @@ def number_field_elements_from_algebraics(numbers, minimal=False, same_field=Fal
     - ``embedded`` -- Boolean (default: ``False``). Whether to make the
       NumberField embedded.
 
+    - ``name`` -- string (default: ``a``). Name of the primitive element.
+
```
tscrim:
```suggestion
    - ``name`` -- string (default: ``'a'``); name of the primitive element
```
Let's not make things worse; instead, let's bring everything into the correct format.

src/sage/rings/qqbar.py
```diff
@@ -3249,11 +3252,18 @@ def _interval_fast(self, prec):
...
+        r"""
+        Given generators ``self``, `\alpha`, and ``other``, `\beta`,
+        ``self.union(other)`` gives a generator for the number field
         `\QQ[\alpha][\beta]`.
 
+        INPUT:
+
+        - ``other`` - an algebraic number.
+
+        - ``name`` - string (default: 'a'). A name for the primitive element.
```
tscrim:
```suggestion
        - ``other`` -- an algebraic number
        - ``name`` -- string (default: ``'a'``); a name for the primitive element
```

tscrim (changes requested):
Just some simple doc fixes. LGTM otherwise.

fchapoton:
Travis also suggested to remove the final dot in the description of parameters

enriqueartal:
> Travis also suggested to remove the final dot in the description of parameters

Sorry, I read it too quickly. Done.

src/sage/rings/qqbar.py
```diff
@@ -3249,11 +3254,17 @@ def _interval_fast(self, prec):
...
+    def union(self, other, name='a'):
+        r"""
+        Given generators ``self``, `\alpha`, and ``other``, `\beta`,
+        ``self.union(other)`` gives a generator for the number field
         `\QQ[\alpha][\beta]`.
 
+        INPUT:
+
+        - ``other`` - an algebraic number
+        - ``name`` - string (default: ``'a'``); a name for the primitive element
```
tscrim:
```suggestion
        - ``other`` -- an algebraic number
        - ``name`` -- string (default: ``'a'``); a name for the primitive element
```
there is also the second `-` needed to make the correct length hyphen.

tscrim (approved):
Thanks. LGTM.

 Whomever thought that a default commit message for some GH way of incorporating reviewer suggested changes should actually work on a project of any meaningful size and then if they cannot see the clear gravity of their error should be immediately fired.

PR #36750, opened 2023-11-22
`networkx` 3.2.1, `igraph` 0.10.8, `python_igraph` 0.11.3

mkoeppe:
- https://networkx.org/documentation/stable/release/release_3.2.html - in particular adds Python 3.12 support
- https://networkx.org/documentation/stable/release/release_3.2.1.html
- https://igraph.org/news.html

### :memo: Checklist


mkoeppe:
On gentoo (https://github.com/sagemath/sage/actions/runs/6961820497/job/18944262177?pr=36750#step:11:6562):
```
hecking whether SageMath should install SPKG igraph...
checking whether any of glpk openblas gmp is installed as or will be installed as SPKG... no
checking for IGRAPH... yes
configure: will use system package and not install SPKG igraph

.....

File "src/sage/graphs/digraph.py", line 447, in sage.graphs.digraph.DiGraph
Failed example:
    import igraph                                  # optional - python_igraph
Exception raised:
      File "/sage/src/sage/doctest/forker.py", line 709, in _run
      File "/sage/src/sage/doctest/forker.py", line 1144, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.graphs.digraph.DiGraph[23]>", line 1, in <module>
        import igraph                                  # optional - python_igraph
        ^^^^^^^^^^^^^
      File "/sage/local/var/lib/sage/venv-python3.11/lib/python3.11/site-packages/igraph/__init__.py", line 25, in <module>
        from igraph._igraph import (
    ImportError: /sage/local/var/lib/sage/venv-python3.11/lib/python3.11/site-packages/igraph/_igraph.abi3.so: undefined symbol: igraph_biadjacency
```

szhorvat:
This is because python-igraph 0.11.3 requires C/igraph 0.10.8. Gentoo has an outdated version of C/igraph, 0.10.4.

mkoeppe:
> This is because python-igraph 0.11.3 requires C/igraph 0.10.8

Thanks for confirming this, that's the lower bound I set in c7c7eb9. Is there an upper bound that I should be setting?

szhorvat:
`< 0.11` is definitely an upper bound, as 0.11 won't be compatible.

However, the accurate answer is that each python-igraph release is only guaranteed to be compatible with the same C/igraph that it bundles. If you want bullet-proof, then compile python-igraph with the bundled version of the C core.

The reason for the incompatibilities are: (1) newly added functions, as it happened in this case (2) experimental functions that are changed.

igraph has "experimental functions" which are excluded from API compatibility guarantees. We reserve the right to change them even across bugfix releases, although we do try to get the design right the first time and avoid such changes. python-igraph does use some of these experimental functions, so in principle, the current version of python-igraph may become incompatible with a future 0.10.x release of C/igraph.

dcoudert (approved):
Tested on Fedora 39 and macOS 12.7.1 (M1).

LGTM.

PR #35366, opened 2023-03-27
Many more namespace packages – follow up

mkoeppe:
### :books: Description

Follow-up from #35322.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #35322, opened 2023-03-21
Many more namespace packages

mkoeppe:
### 📚 Description
As discussed in https://github.com/sagemath/sage/pull/35100#issuecomment-1477174851, we remove many of the empty `__init__.py` files, turning these packages into PEP-420 implicit namespace packages.

The following packages will keep their `__init__.py` files because they are intended to remain ordinary packages indefinitely:
- `sage.cpython`, `sage.structure` – because they are shipped as a whole by **sagemath-objects**
- `sage.doctest`, `sage.repl` – because they are shipped as a whole by **sagemath-repl**
- `sage.features` – because it is shipped as a whole by **sagemath-environment**

This is part of:
- #29705 

### 📝 Checklist

- [x] I have made sure that the title is self-explanatory and the description concisely explains the PR.
- [x] I have linked an issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36753, opened 2023-11-23
Readd init files

tobiasdiez:
There are various issues with using namespace packages because essential dependencies of sage are not fully supporting them. In particular,

- As seen in https://github.com/sagemath/sage/pull/36228 and #36524, cython currently doesn't support namespace packages (it creates the wrong module name, perhaps related to https://github.com/cython/cython/issues/5237). Sagelib (more specifically sage-setup) currently implements an ugly workaround by monkey-patching Cython internals
- Sage's on-the-fly cython compilation is currently completely broken as none of the namespace packages can be used in it (see the changed doctest in this PR)
- There are various issues with pytest, some of the are discussed in https://github.com/sagemath/sage/issues/33826
- importlib.resources doesn't support namespace packages on Python 3.9. https://github.com/sagemath/sage/pull/37344#issuecomment-1951560051
- It is not possible to compile and use a cython file using the standard cython cli. This, however, can be very helpful in certain situations where you want to experiment with different cython options but don't want to wait for a complete recompilation of all cython files. I used this quite often while preparing #36524.

Until these issues are fixed upstream, we revert https://github.com/sagemath/sage/pull/35322 and https://github.com/sagemath/sage/pull/35366 by readding the `__init__.py` files.

The proposed alternative https://github.com/sagemath/sage/pull/37009 would only provide an (ugly) workaround for the last issue.

### :memo: Checklist


mkoeppe:
This change is not needed in Sage, so why not just keep this workaround in your experimental branch

tobiasdiez:
> This change is not needed in Sage

It fixes the issues you see in https://github.com/sagemath/sage/pull/36228. Not sure if you count your PR as being part of Sage or not.

> so why not just keep this workaround in your experimental branch

Because https://github.com/sagemath/sage/pull/36524 is already in a good shape (there is only one compilation issue left), so to ease review of it I started to extract changes in that branch to new PRs.

mkoeppe:
> > This change is not needed in Sage
> 
> It fixes the issues you see in #36228. Not sure if you count your PR as being part of Sage or not. This PR also shows that the only reason why its currently working in the develop branch is that sage is monkey patching cython's package discovery.

Yes, that is precisely the purpose of our customization of Cython.

Obviously, #36228 is not part of Sage. It is a draft PR that investigates what the status of upstream Cython namespace package support is. We cannot yet switch to it.

tobiasdiez:
> We cannot yet switch to it.

Should work with the changes of this PR. Give it a try: https://github.com/mkoeppe/sage/pull/20

mkoeppe:
Obviously, when you replace namespace packages by ordinary packages, then namespace package support is not needed. What's there to try out.

But we need the namespace packages for the modularization -- which is why we removed the init files.

mkoeppe:
So why not instead of calling `cython` from meson, call a custom script that uses out customization.

tobiasdiez:
> Obviously, when you replace namespace packages by ordinary packages, then namespace package support is not needed. What's there to try out.

Matthias, it's not my fault that Cython's namespace package support has bugs. So what's the problem with not using it (for now, until it's ready)?

> But we need the namespace packages for the modularization -- which is why we removed the init files.

It runs perfectly fine right now. If additional modularized distributions need namespace support, just don't install the init files (they are needed for compilation only). (Or design around the cython limitation for now and don't use namespace packages.)

> So why not instead of calling `cython` from meson, call a custom script that uses our customization.

I'm not calling cython myself. It's all handled internally by meson, see https://mesonbuild.com/Python-module.html#extension_module.

mkoeppe:
> > why not instead of calling `cython` from meson, call a custom script that uses our customization.
> 
> I'm not calling cython myself. It's all handled internally by meson, see https://mesonbuild.com/Python-module.html#extension_module.

Without doubt, trivial to customize.

mkoeppe:
> Cython's namespace package support has bugs. So what's the problem with not using it (for now, until it's ready)?

Once again, Sage is not affected by this bug because we have a working customization of Cython.

tobiasdiez:
> > Cython's namespace package support has bugs. So what's the problem with not using it (for now, until it's ready)?
> 
> Once again, Sage is not affected by this bug because we have a working customization of Cython.

What's the point of this posturing? Do you just want to make my work on https://github.com/sagemath/sage/pull/36524 harder?

Clearly there are bugs in Cython. The options (brought forward so far) are either a) the current approach of monkey patching or b) just not using namespace modules during compilation (i.e. this PR here). Both are workarounds for the same Cython bug. Option a) does not work (easily) for https://github.com/sagemath/sage/pull/36524, while b) does work well. So why not go with b)?

dimpase:
Unfortunately @mkoeppe prefers quick "clever" workarounds to proper, standard ways to do things;  instead of using external packages as everybody does, Sage vendors most everything, and is adding more and more vendored packages at an alarming rate.

mkoeppe:
> What's the point of this posturing? Do you just want to make my work on #36524 harder?

@tobiasdiez I have made clear to you in https://github.com/sagemath/sage/pull/36572#issuecomment-1803044465 that insinuations of this type are unwelcome. I'm referring this to sage-abuse again.

dimpase (approved):
nice - so we can revive our pytest efforts, right?

mkoeppe:
This is incompatible with the modularization work and cannot be merged.

tornaria:
> > Cython's namespace package support has bugs. So what's the problem with not using it (for now, until it's ready)?
> 
> Once again, Sage is not affected by this bug because we have a working customization of Cython.

Could we please have as major goal to stop using "customization" of standard tools? This only adds a lot of friction to making sagemath more modular as in: work in a normal environment with reasonably up to date dependencies (and don't require them to be patched just to suit sagemath).

If sage-the-distro is meant as a "reference distribution" of sagemath, it should strive to ship packages from unpatched, upstream released versions.

dimpase:
> This is incompatible with the modularization work and cannot be merged.
> 

any CI failures indicating this?

mkoeppe:
> I systematically need to `touch src/sage/__init__.py` because sagemath is broken without this file.

Because of a system install of Sage components as discussed previously?

mkoeppe:
> Could we please have as major goal to stop using "customization" of standard tools? 

Of course, that's a major goal.
- #31543

tornaria:
> > I systematically need to `touch src/sage/__init__.py` because sagemath is broken without this file.
> 
> Because of a system install of Sage components as discussed previously?

Yes, and before you tell me this is "unsupported", it works very well, thank you very much, except for the missing `__init__` files (which I think also cause some trouble when compiling cython code external, perhaps due to relative imports that also provoke other issues but the root cause seems to always be the lack of __init__ files).

Maybe your goal is a "modular walled garden".

mkoeppe:
And concretely for the problem with namespace packages in current Cython, this is:
- #36228

This will need some work with upstream, which would be very valuable. I don't have time for this at the moment, so any help is very welcome.

As I have said in comments above, for Sage there is nothing to fix because we have the customization in place.

mkoeppe:
> "modular walled garden".

Gardening on balconies is a valid use case

tornaria:
> > Could we please have as major goal to stop using "customization" of standard tools?
> 
> Of course, that's a major goal.
> 
>     * [Meta-ticket: Eliminate use of patches for Python packages #31543](https://github.com/sagemath/sage/issues/31543)

You know, what strikes me the most, is that you, @dimpase and @tobiasdiez (and I?) seem to be all really in a very strong agreement on most of the substantial matters, and you (we?) keep arguing about implementation details instead of focusing on what's important.

Having or not having __init__ files, using or not using namespace packages, are just implementation details. Having feedback on the form "this breaks my workflow, please don't be a sucker" is an asset, not a liability. I keep doing it because I know you care, you listen, and you are able to figure out more robust ways to achieve the goals. I also try to adjust my workflow, often with your help. I think that's more constructive than turning the review labels into a flip-flop.

mkoeppe:
@tornaria Of course your feedback is always appreciated.

But some "implementation details" matter very much -- because of external constraints, namely the Python "ecosystem". Implicit namespace packages are the standard defined by PEP 420. I am using them because it allows us to separate distribution management from namespace management: I have been able to modularize without renaming hundreds of modules.

mkoeppe:
Yes, we are on the cutting edge with this extensive use of namespace packages, but the remaining minor troubles that come from this are easy to solve.

tornaria:
> @tornaria Of course your feedback is always appreciated.
> 
> But some "implementation details" matter very much -- because of external constraints, namely the Python "ecosystem". Implicit namespace packages are the standard defined by PEP 420. I am using them because it allows us to separate distribution management from namespace management: I have been able to modularize without renaming hundreds of modules.

But do we really want to ship the sage library split into pieces? You convinced me before that the real gain is that one can not install e.g. the sagemath-pari distribution, so that pari doesn't have to be installed. But maybe there is another way: ship the sage library as a whole, but activate the "sagemath-pari" part only if pari is available, etc. IOW, keep sage the library as a whole, but make it more "dynamic" on what software it works with. There could still be a "sagemath-pari" pip package, but it won't need to install anything into the sage library but only make sure pari is working.

mkoeppe:
> but activate the "sagemath-pari" part only if pari is available, etc. IOW, keep sage the library as a whole, but make it more "dynamic" on what software it works with. 

No, this is incompatible with Python packaging.
Functionality is keyed to distribution names.

tornaria:
> > but activate the "sagemath-pari" part only if pari is available, etc. IOW, keep sage the library as a whole, but make it more "dynamic" on what software it works with.
> 
> No, this is incompatible with Python packaging. Functionality is keyed to distribution names.

Why? Python itself is dynamic, and there are tons of python packages with functionality that is disabled or enabled on the presence of another python package.

dimpase:
Python isn't really in a great shape - you can argue that PEP 420 is standard, but neither Cython nor Pytest implemented it.

CPython is taken by "speeduppers", who try to make sure that the COBOL-like tasks important for their employers are fast, submit PEPs which get rejected only to sneak them in via backdoor, breaking things in the process. 

Packaging... let's not get started on this. "pip PTSD" is a thing

mkoeppe:
Of course, and that's exactly what we do at the **runtime** in the modularized Sage library; either with `try... import ... except` or `lazy_import` or Feature tests. The runtime does not care about the distributions, it only cares about the available functionality.

But at **build time**, whenever Cython modules are involved, things are more complicated!

tornaria:
> > > but activate the "sagemath-pari" part only if pari is available, etc. IOW, keep sage the library as a whole, but make it more "dynamic" on what software it works with.
> > 
> > 
> > No, this is incompatible with Python packaging. Functionality is keyed to distribution names.
> 
> Why? Python itself is dynamic, and there are tons of python packages with functionality that is disabled or enabled on the presence of another python package.

As a quick example: mpmath optional dependency on gmpy2. In fact, mpmath backend defaults to python numbers. If a good enough gmpy2 is available (and `MPMATH_NOMGPY` is not set in the environment) then it will use gmpy2 as the backend (which is faster).

Have a look into `.../site-packages/mpmath/libmp/backend.py`.

tornaria:
> Of course, and that's exactly what we do at the **runtime** in the modularized Sage library; either with `try... import ... except` or `lazy_import` or Feature tests. The runtime does not care about the distributions, it only cares about the available functionality.
> 
> But at **build time**, whenever Cython modules are involved, things are more complicated!

Isn't that why `cypari2`, etc. exist?

mkoeppe:
I've written about the different types of dependencies at https://doc.sagemath.org/html/en/developer/packaging_sage_library.html#dependencies-and-distribution-packages

mkoeppe:
> > But at **build time**, whenever Cython modules are involved, things are more complicated!
> 
> Isn't that why `cypari2`, etc. exist?

They solve the problem when you only need to `import` from them. They don't solve the problem when you need to `cimport` from them.

mkoeppe:
`cypari2` is a good distribution, but it only packages modules that have no dependency on Sage. 

In the Sage library, we have modules that also have a compile-time dependency on PARI (and cypari2); and the point of `sagemath-pari` is exactly to package these in a separately buildable distribution.

dimpase:
anyhow, if PEP 420 turns out to be a dud (just as e.g. relative  imports were proven harmful) then we'd have really wasted a lot of time for nothing.

mkoeppe:
Yes, and if Python is discontinued, then we also have a problem.

dimpase:
> `cypari2` is a good distribution, but it only packages modules that have no dependency on Sage.
> 
> In the Sage library, we have modules that also have a compile-time dependency on PARI (and cypari2); and the point of `sagemath-pari` is exactly to package these in a separately buildable distribution.

nothing prevents cypari2 to package PARI's headers - the library to link is already there (although perhaps with a mangled name, by default ?). Then `cypari2` may be used to provide PARI lib to other parts of Sage on C level.

mkoeppe:
> Then `cypari2` may be used to provide PARI lib to other parts of Sage on C level.

Yes, that would be a useful change, similar to what purportedly gmpy2 will be doing.

But it does not solve what we're discussing here.

tobiasdiez:
@mkoeppe, I would like to kindly request a bit more detailed feedback than

> This is incompatible with the modularization work and cannot be merged.

Understanding the specifics of the compatibility issues and the reasons behind considering the proposed changes incompatible with the modularization work would greatly assist in addressing the concerns effectively. Could you please provide more details on what aspects are affected, why the suggested workarounds might not be suitable, and if there are alternative solutions you would recommend?

I may also point out that the claim "This change is not needed in Sage" is a bit short-sighted. This has been already mentioned above, but let me summarize:
- Sagelib (more specifically sage-setup) needs to implement a workaround by monkey-patching Cython internals
- Sage's on-the-fly cython compilation is currently completely broken as none of the namespace packages can be used in it (see the changed doctest in this PR)
- There are various issues with pytest, some of the are discussed in https://github.com/sagemath/sage/issues/33826
- It is not possible to compile and use a cython file by using the standard cython cli. This, however, can be very helpful in certain situations where you want to experiment with different cython options but don't want to wait for a complete recompilation of all cython files. I used this quite often while preparing #36524.

In view of these issues, it's only reasonable to wait a bit longer for the ecosystem to mature until we jump on the "namespace package" train.

tobiasdiez:
@mkoeppe should I set this to disputed or are you intending to answer to the questions raised towards you in https://github.com/sagemath/sage/pull/36753#issuecomment-1852397103 and https://github.com/sagemath/sage/pull/36753#issuecomment-1860808781?

mkoeppe:
Obsolete with
- #37009

tobiasdiez:
> Obsolete with
> 
>     * [`src/bin/sage-cython`: Repurpose as PEP 420 fixer #37009](https://github.com/sagemath/sage/pull/37009)

Apart from the fact that this doesn't seem to work (the tests fail in https://github.com/sagemath/sage/pull/37012), this would only address one of the issues mentioned in https://github.com/sagemath/sage/pull/36753#issuecomment-1860808781. They other ones are still open.

mkoeppe:
> Apart from the fact that this doesn't seem to work

I don't believe that you have even tried it out.

tornaria:
It seems that namespace packages are still half-baked, and cause several surprising and unexpected trouble. Several issues have been already mentioned above, and the last straw is https://github.com/sagemath/sage/pull/37344#issuecomment-1951560051

The PR looks good to me, so I'm giving it a _second_ positive review.

dimpase (approved):

tornaria:
Please see: https://github.com/sagemath/sage/pull/36726#issuecomment-1820148873

tornaria:
It seems there is also something about namespace packages when importing from a zip file/wheel

See: https://github.com/sagemath/sage/pull/37399#discussion_r1496250813

dimpase:
@mkoeppe - you are not helping here.

roed314:
This PR was given positive review by @dimpase on December 12 before being given a disputed label.  I'm restoring the positive review that @mkoeppe removed.

Voting on disputed tickets will start soon; I will make a post on sage-devel shortly.

kwankyu:
-1 from me. This PR is undoing things done for modularization of the sage library, deliverables of which are distributions packages. Namespace package is the major tool that makes distributions packages possible. The PR descriptions says that namespace packages do not work well with some other parts of sage. These problems seem more or less concerns of developers. I think continuing the efforts to ship working distribution packages is more important objective than curing some inconveniences of developers.

kwankyu:
I reset to "needs review" status, as this PR needs voting to be resolved.

dimpase:
Namespace packages aren't widely adopted, and are more of a hassle than anything.
Due to them, standard Python tools, e.g. Cython, pytest, are hard to use with Sage.

+1 here

tobiasdiez:
> -1 from me. This PR is undoing things done for modularization of the sage library, deliverables of which are distributions packages. Namespace package is the major tool that makes distributions packages possible. The PR descriptions says that namespace packages do not work well with some other parts of sage. These problems seem more or less concerns of developers. I think continuing the efforts to ship working distribution packages is more important objective than curing some inconveniences of developers.

Thanks for your vote! Could you please give more details what issues you experienced when using this PR with the modularizated packages? Currently, these packages explicitly exclude the init files from being distributed, e.g.
https://github.com/sagemath/sage/blob/e417e2205be84d6d567b8897527fa6945ad09bdb/pkgs/sagemath-categories/MANIFEST.in.m4#L31-L32
So the modularized packages should not be affected by the changes in this PR. But perhaps I missed something and we need to declare more such exclusions.

kwankyu:
> ... Could you please give more details what issues you experienced when using this PR with the modularizated packages? 

I didn't test this PR, and there is no more details. I speculated that the distribution packages would not work because the PR description says you added `__init__` files back.

> Currently, these packages explicitly exclude the init files from being distributed, e.g.
> 
> https://github.com/sagemath/sage/blob/e417e2205be84d6d567b8897527fa6945ad09bdb/pkgs/sagemath-categories/MANIFEST.in.m4#L31-L32
> 
> So the modularized packages should not be affected by the changes in this PR, and still install namespace packages. 

Ah, okay. Then my speculation is wrong. 

@mkoeppe Is it right that this PR is compatible with the modularization infrastructure in sage?

mkoeppe:
> Due to this favourite pet project of one developer, the project benefits of which we don't see, and won't see any time soon, Sage is turning into a bug swamp, as it prevents sane tooling and forces developers off more important things.

Once again Dima repeats his denunciations of my work.

We don't do that here, Dima. I had thought the sage-abuse committee made this very clear to you. @roed314 @jhpalmieri @williamstein

mkoeppe:
> > ... Could you please give more details what issues you experienced when using this PR with the modularizated packages?
> 
> I didn't test this PR, and there is no more details. I speculated that the distribution packages would not work because the PR description says you added `__init__` files back.
> 
> > Currently, these packages explicitly exclude the init files from being distributed, e.g.
> > https://github.com/sagemath/sage/blob/e417e2205be84d6d567b8897527fa6945ad09bdb/pkgs/sagemath-categories/MANIFEST.in.m4#L31-L32
> > 
> > So the modularized packages should not be affected by the changes in this PR, and still install namespace packages.
> 
> @mkoeppe Is it right that this PR is compatible with the modularization infrastructure in sage?

No, it is incompatible, as I have explained clearly here and elsewhere, and is obvious from reading the documentation in our developer guide (@kwankyu, I know that you have read it).

Tobias is correct that the exclusions "should" take care of a common use case, in which one builds the modularized distributions as wheels and then installs them, without access to the source directory. None of the extra `__init__` files shipped, so no change of behavior, fine. (That's why some of the automated tests here may be passing.)

But as soon as the source directory is in the PYTHONPATH, everything breaks. The `__init__` files in the source directory make all of the installed namespace packages invisible to Python. For example, I build and install a wheel of sagemath-bliss but then try to use it with the editable installation in the source tree. It will be as if it's not installed, a fresh cause for confusion and frustration by users and developers.

mkoeppe:
@kcrisman Responding to your suggestion in https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/ciVrZ7x0AQAJ

Here's my confirmation of my vote of -1 for this PR.

The bullet list of issues in the PR description is a laundry list of trivialities, compiled for the specific purpose of obstructing my modularization work. This PR does not fix a problem -- it is engineered to create problems. It is artificial friction.

orlitzky:
+1

I have no strong desire to obstruct the modularization work, and apropos of nothing else I think init files are stupid. But I also think that, in the short term, this is the least-bad option.

There are several annoyances caused by the namespaces packages, and adding custom tooling to work around it is not only itself an annoyance, but a barrier to making sage a "normal" project. Making sage more normal and separating sagelib from sage-the-distribution are the higher priorities for me. In any case, whatever we decide to do here is only temporary because the upstream issues will eventually be resolved.

dimpase:
>> Due to this favourite pet project of one developer, the project benefits of which we don't see, and won't see any time soon, Sage is turning into a bug swamp, as it prevents sane tooling and forces developers off more important things.

> Once again Dima repeats his denunciations of my work.

It is not denunciation, it's merely putting it into perspective. We swallowed your evaluation of `--enable-system-site-packages` as an experimental feature you feel like breaking now and then with impunity. Please allow us to proceed over your questionable (see below) objection here.

> But as soon as the source directory is in the PYTHONPATH, everything breaks.

Well, then don't put the source directory there, problem solved.

mkoeppe:
@dimpase As I explained to you previously in https://github.com/sagemath/sage/pull/36726#issuecomment-1870544967, a discussion is not possible with a participant who blatantly violates our Code of Conduct.

> We 

@dimpase Who is "we"?

> swallowed your evaluation of `--enable-system-site-packages` as an experimental feature 

@dimpase This is called "playing the victim". A standard technique of manipulation and abuse. Unacceptable.

The feature It is experimental per the authors of this feature and clearly marked as such in `./configure --help`.
```
  --enable-system-site-packages
                          allow the use of python packages from the system
                          (experimental; default: no)
```

> you feel like breaking now and then with impunity.

Another denunciation.

mkoeppe:
So let's investigate the bullet list of issues one by one:

> There are various issues with using namespace packages because essential dependencies of sage are not fully supporting them. In particular,
> 
> * As seen in [Use Cython 3 implicit namespace package support #36228](https://github.com/sagemath/sage/pull/36228) and [Compile everything with meson #36524](https://github.com/sagemath/sage/pull/36524), cython currently doesn't support namespace packages (it creates the wrong module name, perhaps related to [[BUG] Building namespace packages does not work with glob patterns cython/cython#5237](https://github.com/cython/cython/issues/5237)). Sagelib (more specifically sage-setup) currently implements an ugly workaround by monkey-patching Cython internals

A non-problem. (As I said from the beginning - https://github.com/sagemath/sage/pull/36753#issuecomment-1826228388.)
This is merely critiquing aesthetics ("ugly workaround") of our customization.

> * Sage's on-the-fly cython compilation is currently completely broken as none of the namespace packages can be used in it (see the changed doctest in this PR)

Unsubstantiated. We use the same customization of Cython in the on-the-fly cython compilation (`sage.misc.cython`). There's not even a bug report.

> * There are various issues with pytest, some of the are discussed in [pytest - fix scope in which doctests are run #33826](https://github.com/sagemath/sage/issues/33826)

The effort to use more pytest in Sage stalled a long time ago, and it's not because of namespace packages.

No effort has been made to work with upstream to add namespace package support.

My estimation: 3h of work for creating something upstreamable for pytest. 
That's many orders of magnitude less than the cost from the artificial friction here.

> * importlib.resources doesn't support namespace packages on Python 3.9. [GAP: make GAP_ROOT_PATHS configuration unnecessary #37344 (comment)](https://github.com/sagemath/sage/pull/37344#issuecomment-1951560051)

- This PR does not help with this at all -- because in the modularized setting, which this PR purportedly supports, the __init__ files are not present
- It's a non-problem because the official backport package https://importlib-resources.readthedocs.io/en/latest/ is provided for the very purpose of providing the latest features to older Python versions.

> * It is not possible to compile and use a cython file using the standard cython cli. This, however, can be very helpful in certain situations where you want to experiment with different cython options but don't want to wait for a complete recompilation of all cython files. I used this quite often while preparing [Compile everything with meson #36524](https://github.com/sagemath/sage/pull/36524).

- It's a non-problem for the build system of Sage.
- This PR is only needed for the meson-python-based build system of the monolithic Sage library (PR #36524).
- This PR does not help at all with meson-python-based build systems of the modularized distributions -- because in the modularized distribution, which this PR purportedly supports, the `__init__` files are not present
- My PR #37009 covers both use-cases (I am using it already in #35095)

kwankyu:
OK. Then I keep my vote -1. This PR is a regression in the modularization infrastructure set in place by years efforts.

culler:
I think that having to work around a reported cython bug is unfortunate, but it does not justify derailing the modularization effort.  So I vote -1.

dimpase:
Modularized packages are work in progress.

Yes, and we (yes, we: Tobias, Michael, me)  want to compile sagelib with meson, now - and it is ready. We do not want to wait for another N years until your modularisation is done, maybe.

dimpase:
> The effort to use more pytest in Sage stalled a long time ago, and it's not because of namespace packages.

This is untrue. This effort has been stalled precisely for this reason, and you know it.

You can  prove me wrong by spending an estimated 3h doing a PR for pytest which fixes this (and everything else what's wrong with pytest), a PR which will be merged in another 3h upon arrival. But you prefer arguing here.

mkoeppe:
For those who were wondering, @dimpase is blocked on my own PRs because of persistent abuse and noise along the same lines.

mkoeppe:
> Modularized packages are work in progress.
> 
> [...] we [...] want to compile sagelib with meson, now - and it is ready.

Obviously you have not put the necessary work in to review either of the two projects.


Once again: There is no technical conflict between using meson and the modularization. 

I'm using meson-python myself in:
- #35095 

Using meson-python does not depend on adding back the `__init__.py` files. The conflict is gratuitous, and obviously a vehicle of hostility. 

This is not acceptable in our community.

tobiasdiez:
> Tobias is correct that the exclusions "should" take care of a common use case, in which one builds the modularized distributions as wheels and then installs them, without access to the source directory. None of the extra `__init__` files shipped, so no change of behavior, fine. (That's why some of the automated tests here may be passing.)

That's the major use case for modularized distributions. In fact, it's the only documented way to test them.

> For example, I build and install a wheel of sagemath-bliss but then try to use it with the editable installation in the source tree. It will be as if it's not installed, a fresh cause for confusion and frustration by users and developers.

That setup makes a lot of sense. You install `src/sage/graphs/bliss.pyx` twice and then find it surprising that python hides one of them.

dimpase:
> For those who were wondering, @dimpase is blocked on my own PRs because of persistent abuse and noise along the same lines.

For those who are wondering, as far as I know, @mkoeppe is the only developer here who blocks Sage developers on GitHub. For disagreeing with him,  due to good, technical, reasons he's unable to find counterarguments for. He considers this abuse. Well, I can understand this - it's indeed painful and frustrating, being unable to find a counterargument.

As Sage has no technical committee or a BDFL to make rulings on technical matters, nowadays we "solve" these by popular vote - as we lost the ability to reach consensus quite a while ago. Would you want to, say, fly on an airplane designed this way, by popular vote? Really?

dimpase:
> > Modularized packages are work in progress.
> > 
> > [...] we [...] want to compile sagelib with meson, now - and it is ready.
> 
> Obviously you have not put the necessary work in to review either of the two projects.

Your only complaint is that this breaks editable installs, right?
I am old enough to remember the times when Sage had no editable installs, and, surprise, we managed...

Besides, how about a possibility that I am just not interested in modularisation atm?
Currently it's a kludge with no real benefits, and with rather frustrating, and performance-degrading, things like sage.features.


> 
> 
> Once again: There is no technical conflict between using meson and the modularization. 
> 
> I'm using meson-python myself in:
> - #35095 
> 
> Using meson-python does not depend on adding back the `__init__.py` files. The conflict is gratuitous, and obviously a vehicle of hostility. 

Maybe you have to realise that you are not a BDFL, Sage is not "your" project, on which you pull all the shots, that other developers might have views conflicting with your views, and that your ongoing work is not flawless? 

One has to seek consensus all the time, in the absence of common view. Seek consensus, and there won't be hostility. If if you just try to push through your point of view as the only acceptable to you, you inevitably get hostility.

src/sage/categories/__init__.py
saraedum, line 1: Is there a reason why the comment is missing in this one?

src/sage/libs/__init__.py
saraedum, line 1: Any reason why the comment is missing in this one?

src/sage/libs/gap/__init__.py
saraedum, line 1: Any reason why the comment is missing in this one?

src/sage/misc/__init__.py
saraedum, line 1: Any reason why the comment is missing in this one?

saraedum:
@mkoeppe I don't understand why this change is so consequential for modularization efforts at the moment. Sure, eventually, the changes here need to be reverted but that's rather trivial it would seem. In the meantime this seems to solve a couple of problems for several people.

I don't fully understand your examples about `PYTHONPATH` and the bliss package. I understand that things break in these cases but I can't really see that these use cases are that important but maybe I am missing something. Obviously there are other ways to fix these problems as you pointed out but why would it be such a consequential mistake to merge this PR in the meantime to make things work again? In particular, I am not very happy with bringing something like `sage-cython` back as a workaround. Why do you think the workaround here is so much worse than the `sage-cython` workaround?

mkoeppe:
> I don't understand why this change is so consequential for modularization efforts at the moment.

Not sure what you mean with "at the moment". The modularization work is ready and overdue to be upstreamed. As soon as #36676 and #36964 (both held back for months by senseless obstruction) are merged, I will prepare the PR that replaces sagemath-standard by the empty distribution with dependencies on the modularized distributions prepared in #35095.

> Why do you think the workaround here is so much worse than the `sage-cython` workaround?

`sage-cython` is a simple build tool that implements for meson-python what our current build system already does -- namely support the package layout of the Sage library. It can be used for both the monolithic build out of src/ and for the modularized builds out of pkgs/* and the corresponding sdists. It is only at use at build time and it has no drawbacks at runtime.

The workaround here, as I explained above in https://github.com/sagemath/sage/pull/36753#issuecomment-2000517759, is limited to Tobias's experimental monolithic build system, cannot be used for the modularized builds, and at runtime creates novel failure modes for combining editable and non-editable installations as I explained on the example with bliss.

kwankyu:
> ... @kwankyu I don't understand why this change is so consequential for modularization efforts at the moment. Sure, eventually, the changes here need to be reverted but that's rather trivial it would seem. In the meantime this seems to solve a couple of problems for several people.

Though Matthias is the single person working for modularized sage, it is an objective of the community, established by several discussions on the sage-devel. It is an objective that impacts the users of sage. The problems mentioned in the PR description seem narrow-scoped technical problems (of a few developers) that has little impact to the users of sage. If the solution of the problems is not compatible with the modularization efforts, then we should choose the more important objective. I don't understand the technical details, on which I rely on Matthias' expertise.

dimpase:
I don't think that modularisation, in the shape it's taking, makes much sense. It splinters the code into too many parts, at expense of making the design very complex and hard to use, and deviating too much from the usual Python design.

E.g. no user would care about sage-numpy, simply  because numpy is so easily available and used in a lot of places. And we don't need design which is so complex that only one developer understands it.

mantepse:
> > I don't understand why this change is so consequential for modularization efforts at the moment.
> 
> Not sure what you mean with "at the moment". The modularization work is ready and overdue to be upstreamed. As soon as #36676 and #36964 (both held back for months by senseless obstruction) are merged, I will prepare the PR that replaces sagemath-standard by the empty distribution with dependencies on the modularized distributions prepared in #35095.

Wouldn't it then make sense to get this one in, under the condition that it is reverted as soon as these other PRs are ready (i.e., have positive review)?  Given the new voting mechanism, it shouldn't be possible to hold back a ticket for "senseless obstruction".

Besides, I strongly dislike the language in "senseless obstruction", it insinuates that others are acting in bad faith.  Even if you honestly feel they do, I think it doesn't really help the situation to use that language.  At the very least it puts me off, and, as already mentioned, I would quit developing sage if it wasn't that my job depends on it.

(I have absolutely no idea of the technical issues and merits, I exclusively use sage on ubuntu where I am root.  In particular, I don't see the advantage of the modularization effort yet, and I dislike the `# needs sage.combinat` tags all over the place, **but** I respect that this may be important to others.)

dimpase:
deciding technical matters by voting is just silly. Would you want to fly in an airplane designed this way? "the majority didn't feel the need for emergency exits, so we just boarded them up"

kwankyu:
> ... In particular, I don't see the advantage of the modularization effort yet, and I dislike the `# needs sage.combinat` tags all over the place, **but** I respect that this may be important to others.

@mkoeppe I think this is an important point. Modularization has been progressing under the hood, but now is affecting other developers, badly in several disputed PRs or mildly in other ways like `# needs` tag . On the other hand, none of them, including myself, has seen promised benefits of the modularization. 

We need to strengthen Public Relationship to defend the effort. What distribution package is useful to users in what situation? If none is yet, when will we have one? There is no user guide or tutorial for distribution packages.

dimpase:
I think modularisation as it stands should be re-thought and possibly done differently at several places.

Syntactic analysis of code, done automatically, can often tell you what features are used in a particular fragment, and this information can be kept separately from the source code.
And one can think of an automatic process which  determines at runtime which parts of sagelib are involved in execution of a particular code fragment, e.g. a doctest (that's basically profiling, a rather well-know technique).

At the moment the source code gets increasingly overloaded with various tags which can very well instead be 

1. computed.

2. stored separately from the source code.

And overloaded with bloatware kludges such as sage.features - which are imposing unneeded runtime burden, consist mostly of boilerplate code, and illustrate deficiencies in the design more than anything else.

As well, the user perspective has to be taken into account. A user doing number theory computations doesn't care whether he needs sage-numpy, or sage-bliss, or whatever that looks tangentially relevant. The user wants sage-NT.
Again, a user doing combinatorics wants an appropriately defined subsets of sagelib, as well.
And as computers get bigger, using monolithic sagelib is less and less of an obstacle.

From development of sagelib  itself point of view, the current design of modularisation is mere a hassle.

From development of sage-dependent Python modules, the biggest obstacle is not absence of modularisation, it's absence of a clean and easy way to install and possibly modify sagelib in an existing venv.
And this is something that held back by the conflict around this and related PRs. Developers of such modules are giving up and deserting Sage.

kwankyu:
> I think modularisation as it stands should be re-thought and possibly done differently at several places.

I respect you complaints. I agree with some of them that I can understand. But as you know, no change is made in sage just by an idea. Someone that believes the idea should work and put forth PRs. Until then, the current state of the modularization that was hardly achieved by energy and time of some people should be respected.

orlitzky:
> (I have absolutely no idea of the technical issues and merits, I exclusively use sage on ubuntu where I am root. In particular, I don't see the advantage of the modularization effort yet, and I dislike the `# needs sage.combinat` tags all over the place, **but** I respect that this may be important to others.)

In theory, the `# needs foo` tags should not be necessary. They allow Matthias to split the packages without reorganizing the tests, which, ultimately, should be done so that each package tests its own code. Likewise, sage-conf, sage-bootstrap, and related workarounds allow the modularization to proceed without first solving some serious underlying problems.

We're getting into conflict because a lot of people would now like to solve those underlying problems, and it requires re-doing parts of the modularization in a better way. Both will have some benefits to end-users, but the path that ultimately requires the least work is to fix our build system first and modularize properly afterwards. Of course, opinions differ.

mantepse:
I think there is something I misunderstand. For me personally, the only real reason to use and develop for sage (and not fricas) is that I can use combinatorics, algebra, graph theory, geometry, number theory, symbolics, plotting and some interfaces to online databases.

It is extremely important to me that the doctest show the interactions between all of these.

It would be a nightmare for me if a module only doctests the things within the same directory.   Lazy series wouldn't exist, for example.

dimpase:
> > I think modularisation as it stands should be re-thought and possibly done differently at several places.
> 
> I respect you complaints. I agree with some of them that I can understand. But as you know, no change is made in sage just by an idea. Someone that believes the idea should work and put forth PRs. Until then, the current state of the modularization that was hardly achieved by energy and time of some people should be respected.
> 

Well, Tobias has put forward a bunch of PRs.
They were rather unwelcome.
Gonzalo put forward a bunch of PRs.
They were rather unwelcome.
I've put forward a bunch of trac tickets dealing with the related issue of slimming down sage-distro. They were rather unwelcome - by the same entity.
I have offered an alternative to recode pkill with a shell script, as if we don't have enough of them already. It was rather unwelcome.

I started to see a pattern :-)

We need a technical committee to decide on such technical things, not a popular vote. A popular vote on such things totally ignores opinions of the minority - and indeed, e.g. people involved in packinging Sage for distros are a minority. Do you want them to give up on distributing Sage by ignoring them?

mkoeppe:
> For me personally, the only real reason to use and develop for sage (and not fricas) is that I can use combinatorics, algebra, graph theory, geometry, number theory, symbolics, plotting and some interfaces to online databases.
> 
> It is extremely important to me that the doctest show the interactions between all of these.
> 
> It would be a nightmare for me if a module only doctests the things within the same directory. Lazy series wouldn't exist, for example.

@mantepse I fully agree with you, and my plan for the modularization states clearly: The documentation will not be modularized.

mkoeppe:
> > The modularization work is ready and overdue to be upstreamed. As soon as #36676 and #36964 [...] are merged, I will prepare the PR that replaces sagemath-standard by the empty distribution with dependencies on the modularized distributions prepared in #35095.
> 
> Wouldn't it then make sense to get this one in, under the condition that it is reverted as soon as these other PRs are ready (i.e., have positive review)?

No, it wouldn't make sense -- and I've already explained why in detail above. 

What would make sense is to review and merge the following PRs:
- https://github.com/sagemath/sage/pull/37009 (which solves the problem of building with meson-python in a way that works for both monolithic and modularized builds)
- https://github.com/sagemath/sage/pull/37008 (which supports the meson-python build system in a way that is suitable for both monolithic and modularized builds)
- https://github.com/sagemath/sage/pull/36740 (which refactors our distribution build scripts)

-

orlitzky:
> I think there is something I misunderstand. For me personally, the only real reason to use and develop for sage (and not fricas) is that I can use combinatorics, algebra, graph theory, geometry, number theory, symbolics, plotting and some interfaces to online databases.
> 
> It is extremely important to me that the doctest show the interactions between all of these.
> 
> It would be a nightmare for me if a module only doctests the things within the same directory. Lazy series wouldn't exist, for example.

Not the same directory, but the same package, or in the packages it depends on.

Suppose **B** depends on **A**. Obviously they interact in some way. If you put the tests for those interactions in **A**, then then you have to add a bunch of `# needs B` everywhere because **A** does not depend on **B**, so **B** might not be installed when you run the tests for **A**.

If, on the other hand, you put the tests in **B**, then no `# needs A` tags are required, because **B** depends on **A**. 

This is a very simple example. In real life, this rests precariously on our ability to split the sage library into modules that form a nice dependency tree -- and a very hard job.

We have a lot of "unnecessary" `# needs` tags right now because, until recently, no one has cared about what-depends-on-what. So when we split **A** and **B** out of the sage library, the tests may be in **A** for no good reason. Adding the `# needs` tag lets us put off moving them into **B** until a later time.

saraedum:
> > > The modularization work is ready and overdue to be upstreamed. As soon as #36676 and #36964 [...] are merged, I will prepare the PR that replaces sagemath-standard by the empty distribution with dependencies on the modularized distributions prepared in #35095.
> > 
> > 
> > Wouldn't it then make sense to get this one in, under the condition that it is reverted as soon as these other PRs are ready (i.e., have positive review)?
> 
> No, it wouldn't make sense -- and I've already explained why in detail above.

Sorry, I don't understand what you are referring to with "above." I read through the entire PR history here but I am not sure where the problem with @mantepse's approach is exactly.

saraedum:
> deciding technical matters by voting is just silly. Would you want to fly in an airplane designed this way? "the majority didn't feel the need for emergency exits, so we just boarded them up"

@dimpase I believe this is the wrong place to question the merits of voting on disputed PRs. If you are unhappy with the process, I'd recommend to discuss the process on sage-devel. (Btw., at least in the EU, it's my understanding that the EN norms (which cover emergency exits I think) are voted on by the national committees and then accepted if at least 71% voted in favor. Then again, the EU is known for not being the most efficient entity when it comes to decision making…but I don't want to distract the discussion here and am happy to continue this on sage-devel or even sage-flame.)

mantepse:
> This is a very simple example. In real life, this rests precariously on our ability to split the sage library into modules that form a nice dependency tree -- and a very hard job.

I possibly did not communicate my concern clear enough, so I try again:

It would be a nightmare for me, if forming such a "nice dependency tree" would be encouraged.

It is extremely important to me, that I can be sure that interactions between all the packages are tested **and** demonstrated.  In other words, I want that the dependency graphs of the mathematical parts is a clique.

There are many classes in sage that I only use occasionally.  For example, I regularly look through recent additions in algebra and number theory, whether some new invariants appeared that I might want to transform into statistics for findstat.  This has been quite successful in the past, even though I know absolutely nothing about the mathematics involved.  But since there are many examples that interact with stuff that I know, I was able to figure out whether I could use something or not.

Another reason to do that is to make my own stuff more robust.  For example, the lazy series code (originally only used in `sage.combinat.species`!) is now relatively reliable because we test their interaction with very many rings and algebras - and almost every time somebody tells me about another algebra in sage - at this point I would like to send warmest greetings to Frederic :-)  - we discover bugs and shortcomings, either in the algebra, or in the lazy series code.

So, at the very least, the root of the depency tree should contain all of groups, rings and algebras, symmetric functions, linear algebra, graph theory, mixed integer linear solvers, categories and number theory, and a big chunk of combinatorics.

I think it might make sense to split out plotting and visualization, but I'm not quite sure.  I do think it makes sense to split out jupiter and the emacs mode, and the doctesting and documentation facilities.

mkoeppe:
> I read through the entire PR history here but I am not sure where the problem with @mantepse's approach is exactly.

@saraedum https://github.com/sagemath/sage/pull/36753#issuecomment-2000517759, https://github.com/sagemath/sage/pull/36753#issuecomment-2011193627

dimpase:
> > deciding technical matters by voting is just silly. Would you want to fly in an airplane designed this way? "the majority didn't feel the need for emergency exits, so we just boarded them up"
> 
> @dimpase I believe this is the wrong place to question the merits of voting on disputed PRs. If you are unhappy with the process, I'd recommend to discuss the process on sage-devel. (Btw., at least in the EU, it's my understanding that the EN norms (which cover emergency exits I think) are voted on by the national committees and then accepted if at least 71% voted in favor. Then again, the EU is known for not being the most efficient entity when it comes to decision making…but I don't want to distract the discussion here and am happy to continue this on sage-devel or even sage-flame.)

I meant "voted by plane designers", certainly not by an outside committee. Sorry for getting off-topic.
Seriously, the main problem with deciding PRs by popular voting is that interests of a sizeable minority are ignored.
A minority (e.g. people packaging Sage on Linuxes) often must not be ignored.

saraedum:
> > I read through the entire PR history here but I am not sure where the problem with @mantepse's approach is exactly.
> 
> @saraedum [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2000517759), [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2011193627)

Ok, let me try to summarize what these comments are saying in the way I understand them.

In one, you explain why you think that the problems that people are trying to solve here are not actual problems/unsubstantiated/… . I am giving these people the benefit of the doubt that this PR would solve these problems and would make their lives easier until the problems are fixed by other means (e.g. upstream.)

The other one is about you wanting other PRs to be reviewed and merged. @mantepse's proposal was exactly to revert this PR once these other PRs are ready to be merged and positively reviewed. So formally, I don't see why this is an argument against this PR.

Sorry, I still fail to understand what's so problematic about this PR if it's so easy to revert.

dimpase:
By reading e.g.

> We can define some meaningful small distributions that just consist of a single or a few Cython modules. For example, sagemath-tdlib ([github issue #29864](https://github.com/sagemath/sage/issues/29864)) would just package the single Cython module that must be linked with tdlib, sage.graphs.graph_decompositions.tdlib. Starting with the Sage 9.6 development cycle, as soon as namespace packages are activated, we can start to create these distributions. This is quite a mechanical task.

it becomes clear that these modules have nothing to do with what users need, and it merely creates a large number of such packages in a mechanical way.  It's an exercise in using Python in a way it's not designed for. The benefits are totally unclear, the overhead (ugly code, slowdowns, quite a bit of extra boilerplate) is considerable.

mantepse:
> @mantepse Key to the design is to understand that there are several different _types_ of dependencies, with different technical underpinnings; and that modularization goes hand-in-hand with integration. 

I must admit that I don't really care much about a theory of dependencies.

If at some point it will be **encouraged** to avoid depending on other parts of sage in doctests, for example in the form that @orlitzky outlined, I will heavily protest.

Too long, didn't read: for me, the sentence "The documentation will not be modularized." is of critical importance.

mkoeppe:
> > @mantepse Key to the design is to understand that there are several different _types_ of dependencies, with different technical underpinnings; and that modularization goes hand-in-hand with integration.
> 
> I must admit that I don't really care much about a theory of dependencies.

That's fine.

> If at some point it will be **encouraged** to avoid depending on other parts of sage in doctests, for example in the form that @orlitzky outlined, I will heavily protest.

So would I.

> Too long, didn't read: for me, the sentence "The documentation will not be modularized." is of critical importance.

Yes.

dimpase:
> I think it might make sense to split out plotting and visualization, but I'm not quite sure. I do think it makes sense to split out jupiter and the emacs mode, and the doctesting and documentation facilities.

there is no jupyter _in_ sagelib, and we are talking about modularisation of sagelib specifically. And emacs mode is already split off as far as I know, it has never been in sagelib, anyway.

Modularisation of sage the distro is a quite different story.

mkoeppe:
> > > I read through the entire PR history here but I am not sure where the problem with @mantepse's approach is exactly.
> > 
> > @saraedum [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2000517759), [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2011193627)
>
> In one, you explain why you think that the problems that people are trying to solve here are not actual problems/unsubstantiated/… .

If you read to the end, you'll also see that I explain why the PR here is a non-solution. 

> The other one is about you wanting other PRs to be reviewed and merged. @mantepse's proposal was exactly to revert this PR once these other PRs are ready to be merged and positively reviewed. 

The ones that I just linked in https://github.com/sagemath/sage/pull/36753#issuecomment-2013235147 **are** ready to be positively reviewed and merged now.

saraedum:
> If you read to the end, you'll also see that I explain why the PR here is a non-solution.

I did indeed read the entire comment several times. I also trust the people in favor of this PR that this does solve problems they are facing.

> > The other one is about you wanting other PRs to be reviewed and merged. @mantepse's proposal was exactly to revert https://github.com/sagemath/sage/pull/36777#issuecomment-1837593395 this PR once these other PRs are ready to be merged and positively reviewed.
> 
> The ones that I just linked in [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2013235147) **are** ready to be positively reviewed and merged now.

The proposal was to revert when they **are** positively reviewed. One of these PRs has been waiting for review since at least three months. It seems to be very hard to review and might still take a while to get to "positive review".

NathanDunfield:
> > The ones that I just linked in [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2013235147) **are** ready to be positively reviewed and merged now.
> 
> The proposal was to revert when they **are** positively reviewed. One of these PRs has been waiting for review since at least three months. It seems to be very hard to review and might still take a while to get to "positive review".

@saraedum The PR's that @mkoeppe referred to are labeled "disputed", just like this one, and until this week there was no method to resolve whether to merge them.  Now that the system of per-ticket voting on disputed items has been created, I think they could be merged within a week or two if folks voted in favor, depending on what (if anything) needs to be done to get them up-to-date compared to the main branch.

saraedum:
> > > The ones that I just linked in [#36753 (comment)](https://github.com/sagemath/sage/pull/36753#issuecomment-2013235147) **are** ready to be positively reviewed and merged now.
> > 
> > 
> > The proposal was to revert when they **are** positively reviewed. One of these PRs has been waiting for review since at least three months. It seems to be very hard to review and might still take a while to get to "positive review".
> 
> @saraedum The PR's that @mkoeppe referred to are labeled "disputed", just like this one, and until this week there was no method to resolve whether to merge them. Now that the system of per-ticket voting on disputed items has been created, I think they could be merged within a week or two if folks voted in favor, depending on what (if anything) needs to be done to get them up-to-date compared to the main branch.

Actually, neither https://github.com/sagemath/sage/pull/36740 nor its dependency https://github.com/sagemath/sage/pull/37430 have been disputed. But they've been at needs review since December 16, and Feb 23.

But sure, if we think that these get merged quickly now then that proposal has less of a benefit.

saraedum:
@mkoeppe could you explain why these linked PRs break if the init files are present? I had a quick look but it's too much m4 to understand what's going on right away. I didn't see the immediate problem with the init files present there.

orlitzky:
> I must admit that I don't really care much about a theory of dependencies.
> 
> If at some point it will be **encouraged** to avoid depending on other parts of sage in doctests, for example in the form that @orlitzky outlined, I will heavily protest.

I'm not encouraging the removal of doctests, only shuffling them around to avoid `# needs sage.foo`. It sounds like you are not going to be a fan of the modularization in any case. I find it more trouble than it's worth at the moment, but I'm trying to be honest about the pros and cons. The `# needs sage.foo` are not a huge con if they can be avoided by copy & pasting the tests into a different file.

mantepse:
> > If at some point it will be **encouraged** to avoid depending on other parts of sage in doctests, for example in the form that @orlitzky outlined, I will heavily protest.
> 
> I'm not encouraging the removal of doctests, only shuffling them around to avoid `# needs sage.foo`.

What precisely do you mean with shuffling around?  Inside the same doctest, no problem.  Inside the same class, maybe.  Into a different file, very problematic in my opinion.

It seems to me, that the healthy situation is, when the applications of classes are heavily interweaved - probably in stark contrast to other computer programs.  It seems to me that these interactions increase the probability that our definitions and choices are consistent, and play nicely with each other.

orlitzky:
> What precisely do you mean with shuffling around? Inside the same doctest, no problem. Inside the same class, maybe. Into a different file, very problematic in my opinion.

Into another file in another package. Why is that problematic? GAP tests work that way, for example. There are still a few circular dependencies, but if package **A** depends on **B**, **C**, and **D**, then typically the tests that combine features from all four will live in **A**.

> It seems to me, that the healthy situation is, when the applications of classes are heavily interweaved - probably in stark contrast to other computer programs. It seems to me that these interactions increase the probability that our definitions and choices are consistent, and play nicely with each other.

In one sense, spaghetti is the strongest pasta :)

Leaving the tests in the "wrong" file with `# needs sage.foo` will actually reduce the amount of real world testing that it gets. If users are installing the packages and running the test suite at install time, the tests will be run one package at a time, in order. If **B** depends on **A**, then **A** will be installed before **B**. If you have a test in **A** that `# needs B`, it will be skipped, because **B** isn't installed yet. Then when you go to install **B**... the test isn't there, because it's in **A**. Moving the test out of **A** and into **B** would actually improve the situation by allowing the end users to run the test at install-time.

I'm happy to keep discussing this but it really seems like you're just not interested in trying to logically split sage into modules, in which case these hypotheticals are never going to satisfy you. It's certainly a lot less confusing and a lot less work if we _don't_ split them. That's a real con.

mkoeppe:
> @mkoeppe could you explain why these linked PRs break if the init files are present? 

@saraedum No, that's not at all what I'm saying. I think some miscommunication has happened here in part by referring to several different sets of PRs that were mentioned for different reasons.

> I mean, #37009 (sage-cython) shouldn't have a problem with the init files for example?

I brought up #37009 (sage-cython) as relevant to the discussion in this PR because it is the proper short-term solution to the problem of building Cython extension modules with meson-python -- short of an upstreamable fix of Cython itself.

saraedum:
@mkoeppe apologies, I might have misunderstood why you were pointing me at these PRs then.

I think I didn't understand your answer to this question then:

> I don't understand why this change is so consequential for modularization efforts at the moment.

Let me rephrase: Which other PRs break badly if this gets merged?

mantepse:
> I'm happy to keep discussing this but it really seems like you're just not interested in trying to logically split sage into modules

I am interested in the question.  I assure you, I can be convinced, but it needs convincing.

> > What precisely do you mean with shuffling around? Inside the same doctest, no problem. Inside the same class, maybe. Into a different file, very problematic in my opinion.
> 
> Into another file in another package. Why is that problematic? GAP tests work that way, for example. There are still a few circular dependencies, but if package **A** depends on **B**, **C**, and **D**, then typically the tests that combine features from all four will live in **A**.

I think this is problematic for the following reason, which I would like to present in the form of my favorite example, lazy series.

We currently have the following `# needs`:
```
lrcalc_python sage.combinat sage.libs.flint sage.libs.pari sage.libs.singular sage.modules sage.rings.finite_rings sage.rings.number_field sage.symbolic
```
I hope that `sage.groups` will soon become another one, once I have an implementation of the Burnside ring.

Quite a few of these use lazy series (even though the series code is a very recent addition), and there are several more places where it could be used.

It seems to me that the proper place to test whether it works with different base rings, as completion of different algebras, etc., is in the series code.  Only if all these tests in are in one place, I can see what is supposed to work and what is not yet working.

One might think that the dependencies for testing are roughly inverse to the dependencies in the code, I'm not sure about that.

> In one sense, spaghetti is the strongest pasta :)

I did start with Commodore BASIC V2.0.  I know my stuff.
However, I actually thought that spaghetti code refers to something quite different: https://en.wikipedia.org/wiki/Spaghetti_code

> Leaving the tests in the "wrong" file with `# needs sage.foo` will actually reduce the amount of real world testing that it gets. If users are installing the packages and running the test suite at install time, the tests will be run one package at a time, in order. If **B** depends on **A**, then **A** will be installed before **B**. If you have a test in **A** that `# needs B`, it will be skipped, because **B** isn't installed yet. Then when you go to install **B**... the test isn't there, because it's in **A**. Moving the test out of **A** and into **B** would actually improve the situation by allowing the end users to run the test at install-time.

I admit that I don't see why one would want to install only a part of the mathematics.  Even in the currently envisioned scheme, I think that there are only a few parts one could possibly remove: graph theory, symbolics and polyhedra, right?  I must say that's quite an odd triple of subjects, and I would guess that without, say, graphs, a big chunk of combinatorics would go, too.

mkoeppe:
> I admit that I don't see why one would want to install only a part of the mathematics.

@mantepse The tl;dr of that is: The math is harmless, it's really all because of the compiled non-Python library dependencies.

orlitzky:
> I think this is problematic for the following reason, which I would like to present in the form of my favorite example, lazy series.
> 
> We currently have the following `# needs`:
> 
> ```
> lrcalc_python sage.combinat sage.libs.flint sage.libs.pari sage.libs.singular sage.modules sage.rings.finite_rings sage.rings.number_field sage.symbolic
> ```
> 
> I hope that `sage.groups` will soon become another one, once I have an implementation of the Burnside ring.
> 
> Quite a few of these use lazy series (even though the series code is a very recent addition), and there are several more places where it could be used.

Based on [the documentation](https://doc.sagemath.org/html/en/developer/packaging_sage_library.html#hierarchy-of-distribution-packages), it's likely that most of those will wind up in the same packages sage-objects and/or sage-categories. The Singular and symbolics (maxima/ecl) parts are going to go in sage-singular and sage-symbolics which both depend on sage-categories and sage-objects. So to eliminate the `# needs`, you would move the tests that need singular and symbolics into files within sage-singular or sage-symbolics. That cleans up the examples a little bit and makes sure the tests are all run unconditionally. If you're worried about users not finding the examples split across multiple files, it's easy to cross-reference the methods and classes with sphinx.


> It seems to me that the proper place to test whether it works with different base rings, as completion of different algebras, etc., is in the series code. Only if all these tests in are in one place, I can see what is supposed to work and what is not yet working.
> 
> One might think that the dependencies for testing are roughly inverse to the dependencies in the code, I'm not sure about that.
> 
> > In one sense, spaghetti is the strongest pasta :)
> 
> I did start with Commodore BASIC V2.0. I know my stuff. However, I actually thought that spaghetti code refers to something quite different: https://en.wikipedia.org/wiki/Spaghetti_code

I was mostly making a joke, but if you have several files each of which depends on all of the others, you get a dependency graph that literally looks like spaghetti. To disentangle it, you have to cut edges, eilminating dependencies by e.g. moving tests that require symbolic from one node (sage-objects?) to another (sage-symbolics).

> I admit that I don't see why one would want to install only a part of the mathematics. Even in the currently envisioned scheme, I think that there are only a few parts one could possibly remove: graph theory, symbolics and polyhedra, right? I must say that's quite an odd triple of subjects, and I would guess that without, say, graphs, a big chunk of combinatorics would go, too.

I don't disagree. I think there's _some_ benefit to it, and so long as it doesn't interfere with anything else, who am I to tell someone how to spend their free time. But now it's starting to interfere with everything else.

mkoeppe:
> @mkoeppe apologies, I might have misunderstood why you were pointing me at these PRs then.

@saraedum No worries.

> I think I didn't understand your answer to this question then:
> 
> > I don't understand why this change is so consequential for modularization efforts at the moment.
> 
> Let me rephrase: Which other PRs break badly if this gets merged?

Wrong question. Just merging this PR here creates the bad breakage, as explained in https://github.com/sagemath/sage/pull/36753#issuecomment-2011193627

mkoeppe:
By the way, in this follow-up PR by the same author, some non-empty init files are added, which can of course no longer just be omitted from the modularized distributions.
- https://github.com/sagemath/sage/pull/36580

tornaria:
> > I don't understand why this change is so consequential for modularization efforts at the moment.
> 
> Not sure what you mean with "at the moment". The modularization work is ready and overdue to be upstreamed.

I'm confused. You claim otherwise in https://github.com/sagemath/sage/pull/35095#issuecomment-1998399020

> As soon as #36676 and #36964 (both held back for months by senseless obstruction) are merged,

Your PRs are held back as much as this PR, #36964, #37138, etc. namely because they are incompatible between each other. The only way out of this is to find a common solution, but this needs good communication that we seem to be lacking very badly.

I fear that if each one of us starts trying to escape the deadlock by creating more PRs with variations, we are only going to end up with a worse deadlock.

> I will prepare the PR that replaces sagemath-standard by the empty distribution with dependencies on the modularized distributions prepared in #35095.

Please don't. This will likely break distro packages of sagemath.

Let's first figure out a way out of this mess that allows distros (and conda?) to package sagemath. Even better, let's try to improve and simplify this process, and ideally move to a standard and supported simple process that we can recommend to all distros.

mkoeppe:
> > > I don't understand why this change is so consequential for modularization efforts at the moment.
> > 
> > 
> > Not sure what you mean with "at the moment". The modularization work is ready and overdue to be upstreamed.
> 
> I'm confused. You claim otherwise in [#35095 (comment)](https://github.com/sagemath/sage/pull/35095#issuecomment-1998399020)

No, I don't, Gonzalo. As I have explained there to you, #35095 is not intended for merging. I upstream by cherry-picking from it.

mkoeppe:
> This will likely break distro packages of sagemath.

No, Gonzalo, it won't. I know what I'm doing.

> Let's first figure out a way out of this mess that allows distros (and conda?) to package sagemath. Even better, let's try to improve and simplify this process, and ideally move to a standard and supported simple process that we can recommend to all distros.

Nonsense. All of Sage library packaging is following best packaging practices -- plain PEP-517 and PEP-518. There's absolutely nothing Sage-specific in it, and it's not a "mess".

tornaria:
> > Let's first figure out a way out of this mess that allows distros (and conda?) to package sagemath. Even better, let's try to improve and simplify this process, and ideally move to a standard and supported simple process that we can recommend to all distros.
> 
> Nonsense. All of Sage library packaging is following best packaging practices -- plain PEP-517 and PEP-518. There's absolutely nothing Sage-specific in it, and it's not a "mess".

You keep dismissing all attempts to create bridges and find a consensual solution.

tornaria:
I think I'm not the first person to tell you that :eyes: looks offensive to me. Please don't tag my comments with this reaction.

mkoeppe:
Gonzalo, the definition of this reaction is "seen". I am using this reaction to mark unwelcome and inappropriate comments.

tornaria:
> Gonzalo, the definition of this reaction is "seen". I am using this reaction to mark unwelcome and inappropriate comments.

The meaning you give it makes it even more offensive. Please don't do it.

mkoeppe:
> Who is the intended audience of this marking? Are you trying to indicate to the person making the comment that it's inappropriate? Bookkeeping for yourself for finding it later? Or marking it for the committee to look at?

@roed314 The intended audience is (1) the author of the inappropriate comment and (2) the other readers. It makes clear that I will not respond to it because it is inappropriate. (3) Yes, it is also bookkeeping to report as abuse later.

mkoeppe:
> @mkoeppe, please avoid referring to other people's arguments as "Nonsense."

@roed314 Are you writing on behalf of the sage-conduct committee? I have pointed this out to you and the old committee before, but more restraint and thought is necessary before you make public reprimands. @nbruin @jhpalmieri @VivianePons @jplab @saraedum

mkoeppe:
On this occasion, let me once more share a helpful resource, the "Calling In and Calling Out Guide" 
- https://edib.harvard.edu/files/dib/files/calling_in_and_calling_out_guide_v4.pdf

dimpase:
I have complained about "rolling eyes" 👀 emoji applied to my comments too, and was assured by @mkoeppe it merely means "seen", no negative connotations.

It's  interesting to see the development of this story. Looks like a gaslighting masterclass ...

tornaria:
> On this occasion, let me once more share a helpful resource, the "Calling In and Calling Out Guide"
> 
>     * https://edib.harvard.edu/files/dib/files/calling_in_and_calling_out_guide_v4.pdf

Thanks for sharing. Here's a snippet:

**What Do I Do When Someone Calls Me In or Out?**
 - **Pause** - take a breath. Ground yourself to receive what they have to say.
 - **Listen** - with the intent of learning and seeing things from their point of view
 - **Acknowledge** - take responsibility for the impact of your words or actions
 - **Reflect** - process your thoughts and emotions. It may help to do so with a trusted partner who will not only have grace for your mistake, neither defending your actions nor condemning you.
 - **Repair the harm done** - change your behavior going forward, inviting trusted
partners to hold you accountable to learning and doing better

tornaria:
> There's absolutely nothing Sage-specific in it, and it's not a "mess".

Just to try to mend spirits. When I used the word "mess" I was not referring to code (and perhaps @mkoeppe read it as that). I was using "mess" to refer to the situation where we are unable to progress and find a solution to the "disputed" PRs.

I hope that, at the very least, we agree that we have a problem; and that is worth trying to find a solution.

mkoeppe:
Thanks, Gonzalo, this clarification is much appreciated.

Obviously, I agree that the situation of the "disputed" PRs is a mess.

roed314:
> > @mkoeppe, please avoid referring to other people's arguments as "Nonsense."
> 
> @roed314 Are you writing on behalf of the sage-conduct committee? I have pointed this out to you and the old committee before, but more restraint and thought is necessary before you make public reprimands. @nbruin @jhpalmieri @VivianePons @jplab @saraedum

I'm sorry, I should have been more clear that I was replying as an individual rather than a representative of the sage-conduct committee.  I am glad that Gonzalo was able to clarify his use of the word "mess" in a way that helped defuse the argument.

We are discussing some of these issues internally on the committee, but it takes sometime to get everyone up to speed and we have a lot of issues to address.

mkoeppe:
As a matter of fact, I do encourage individuals in our community to speak up when inappropriate conduct is happening in plain sight. As I have shared with some of you in private, I believe that our community has already been severely damaged by toxicity and its normalization; not just by the most recent excesses but already over a much longer term. The silence (bystanderism) when harmful conduct is happening in plain sight is one of the symptoms of this damage. I'll say that the eagerness to frame clearly abusive behavior as mere technical disputes between equals is but another symptom of the same damage.

kwankyu:
> As a matter of fact, I do encourage individuals in our community to speak up when inappropriate conduct is happening in plain sight. ... The silence (bystanderism) when harmful conduct is happening in plain sight is one of the symptoms of this damage.

People become easily bystanders, and I think it is not realistic to expect them to behave differently just by encouraging. I am a bystander to all kinds of misconducts done in my neighborhood. 

Now that we have new CoC committee, it should be the entity to receive reports of misconduct, decide what is a misconduct (based on our CoC, a list of "bad words" may be enough for most cases), and to inform the offender and all of the decision, and to take action for repetitive misconducts. Do we have an official way to report in github? Does @sage-conduct or something work?

nbruin:
> Do we have an official way to report in github? Does @sage-conduct or something work?

Send a message to sage-conduct@googlegroups.com with a link to the github pull request/issue; perferrably a link to the particular comment you would like to report. And of course, include the reason for reporting.

Needless to say, this particular thread has already come to our attention.

mkoeppe:
> > As a matter of fact, I do encourage individuals in our community to speak up when inappropriate conduct is happening in plain sight. ... The silence (bystanderism) when harmful conduct is happening in plain sight is one of the symptoms of this damage.
> 
> People become easily bystanders, and I think it is not realistic to expect them to behave differently just by encouraging.

Indeed the culture of our community is not something that can be expected to be fixed overnight. 

This will be a long-term process for our community, which will need hard work, positive examples, and affirmative steps.

> Now that we have new CoC committee, it should be the entity to receive reports of misconduct, decide what is a misconduct [...], and to inform the offender and all of the decision, and to take action for repetitive misconducts.

Yes, and we have to be grateful to the members of the committee for their willingness to take on this difficult duty. 

But our community cannot expect to delegate all of this to the committee.

For example, there is a mismatch in timing: Per our new code of conduct, the CoC committee strives to acknowledge reports within 72 hours. Although this is already orders of magnitude faster than the past practice, many discussions are much faster paced than that. 

In the meantime, targets of abusive conduct will be needing the support of the community. Silence or continuing discussions as if nothing happened -- although natural on an individual level as self-protection -- is problematic because it implicitly legitimizes the misconduct.

kwankyu:
> For example, there is a mismatch in timing: Per our new code of conduct, the CoC committee strives to acknowledge reports within 72 hours. Although this is already orders of magnitude faster than the past practice, many discussions are much faster paced than that.

I look forward to this: As time goes on, cases of CoC decisions will be piled. Then this will have a preventive effect for the offender who fails in recognizing his/her own misconducts (in discussion, comments labeling, ...). Then bystanders can simply point to the piled CoC decisions for the offender, and cease to be bystanders anymore.

saraedum:
> As a matter of fact, I do encourage individuals in our community to speak up when inappropriate conduct is happening in plain sight.

We welcome all attempts at defusing a debate that has gotten heated. However, if you witness abusive behavior, please be very careful about publicly labeling it as such yourself. You might be misreading the intentions of the author and more importantly this can lead to a further escalation of the situation. Instead, please notify sage-conduct. We won’t tolerate any abusive behavior and will hide or delete inappropriate comments.

> The silence (bystanderism) when harmful conduct is happening in plain sight is one of the symptoms of this damage.

The committee is notified frequently, especially over the past month or so, perhaps due to increased visibility of the committee and its role due to recent votes and discussions in sage-devel. We therefore believe that the allegation of bystanderism is a misrepresentation of our community.

> I'll say that the eagerness to frame clearly abusive behavior as mere technical disputes between equals is but another symptom of the same damage.

The Code of Conduct Committee disagrees that there are no underlying technical disagreements underlying the many recent disputes. See our comment [here](https://github.com/sagemath/sage/pull/37501#issuecomment-2010579175)
However, the existence of a technical foundation for the dispute does not excuse abuse, and we do agree that abuse has occurred around these issues. Such abuse should be reported and the CoC will assess the situation with the parties involved.

Finally, let us please limit the discussion on this PR to the technical questions related to this PR.

The SageMath Code of Conduct Committee.

tornaria:
This PR was given positive review on dec 12: https://github.com/sagemath/sage/pull/36753#pullrequestreview-1777224517

I'm +1 on this, since having `sage` as a namespace package brings subtle and confusing interactions between a user install of sagemath (e.g., the git repo one is developing on) and a system install of sagemath.

Shall we set this back to positive review?

mkoeppe:
> Shall we set this back to positive review?

No, we shall not.

tornaria:
> > Shall we set this back to positive review?
> 
> No, we shall not.

Why not?

NathanDunfield:
> Why not?

As a disputed PR, it should only be set to positive review if the number of positive votes is [at least twice the number of negative votes](https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/kvmOlVb1AQAJ).  I see three negative votes and four positive ones (counting the PR author as positive), though I certainly could have miscounted.

mkoeppe:
I've already explained it in detail in:
- https://github.com/sagemath/sage/pull/36753#issuecomment-2000065224
- https://github.com/sagemath/sage/pull/36753#issuecomment-2000074516
- https://github.com/sagemath/sage/pull/36753#issuecomment-2000517759
- https://github.com/sagemath/sage/pull/36753#issuecomment-2011193627

tornaria:
> > Why not?
> 
> As a disputed PR, it should only be set to positive review if the number of positive votes is [at least twice the number of negative votes](https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/kvmOlVb1AQAJ). I see three negative votes and four positive ones (counting the PR author as positive), though I certainly could have miscounted.

Thanks, this was what I meant.

It's very hard to look with all the offtopic; I counted four +1 and two -1, that's why I asked.

src/sage/misc/__init__.py
dimpase: fixed

dimpase:
I don't think it humanely possible to count votes this way.
IMHO it should be done by the usual GitHub inteface: Approve/Request chnages. Then it's visible. Right now  I see one review from me, comments (I just addressed them) by @saraedum - and that's all.

saraedum:
@dimpase I agree that the system is not ideal. When you want to set it to positive review, you unfortunately have to browse through the entire PR history to count the votes. I guess we could reviews, somebody could build a vote-counting-bot, or we could use reactions on the PR itself. Anyway, I don't think this is a discussion that should be had on this PR but feel free to propose something on sage-devel.

So, for the record, currently these are the votes I see. (And hopefully, next time I remember and can start counting from this comment…)

## Current Votes (5-3)

:+1: @tobiasdiez (author)
:-1: @kwankyu https://github.com/sagemath/sage/pull/36753#issuecomment-1999078381
:+1: @dimpase https://github.com/sagemath/sage/pull/36753#issuecomment-1999164090
:-1: @mkoeppe https://github.com/sagemath/sage/pull/36753#issuecomment-2000074516
:+1: @orlitzky https://github.com/sagemath/sage/pull/36753#issuecomment-2000144146
:-1: @culler https://github.com/sagemath/sage/pull/36753#issuecomment-2000693842
:+1: @tornaria https://github.com/sagemath/sage/pull/36753#issuecomment-2041620469
:+1: @saraedum (here)

mkoeppe:
> > As a matter of fact, I do encourage individuals in our community to speak up when inappropriate conduct is happening in plain sight.
> 
> We welcome all attempts at defusing a debate that has gotten heated. However, if you witness abusive behavior, please be very careful about publicly labeling it as such yourself.

People can speak up without having to label anything. They can simply express that something that they have witnessed makes them uncomfortable. That's already helpful support for targets of abuse.

jhpalmieri:
I vote -1. I think that makes it 5-4.

tobiasdiez:
> I vote -1. I think that makes it 5-4.

Could you please shortly outline how/why you came to this conclusion? Thanks!

kwankyu:
> > I vote -1. I think that makes it 5-4.
> 
> Could you please shortly outline how/why you came to this conclusion? Thanks!

It is inappropriate to ask a voter to defend his voting. It is not part of the voting procedure.

roed314:
> It is inappropriate to ask a voter to defend his voting. It is not part of the voting procedure.

I'm just speaking for myself here, not the sage-conduct committee.  I don't think it's unreasonable to ask someone for their reasoning, @jhpalmieri is just not obligated to respond.

dimpase:
I think that people engage in purely political voting here. 

There is a status quo party, oddly united by most if not all of them running macOS as their primary OS, and the rest. Probably macOSers think of Sage as a missing package manager, and cannot understand why others hate this PoV.

dimpase:
@tobiasdiez - what's going on here nowadays?

tobiasdiez:
Not much. The tools we use have still the same limitations around namespace packages (eg https://github.com/cython/cython/issues/6287 and https://github.com/cython/cython/issues/5335 and https://github.com/cython/cython/issues/5237), so I still think it's a good idea to readd the init files until these issues are fixed.

But given the current vote is 5-4, I don't think this can be solved by the "disputed" process except if some of the people that voted with -1 in the meantime support the change (or at least are no longer against it). The other resolution would be a vote on the mailing list.

tobiasdiez:
As part of autumn-cleanup, I'll close this PR now. It would still be highly desirably to readd these init files for now, but I guess we have to find a workaround for the tools limitations / namespace bugs at the moment....

tobiasdiez:
Yes, there is no an option [consider_namespace_packages](https://docs.pytest.org/en/stable/reference/reference.html#confval-consider_namespace_packages), but my understanding is that its either or: either you use namespace packages everywhere or never. And there are still a few issues with this option: https://github.com/search?q=repo%3Apytest-dev%2Fpytest+consider_namespace_packages&type=issues

PR #36755, opened 2023-11-23
Remove sage.gaprc

tobiasdiez:
The ̀ sage.gaprc` allowed to customize the initialization of gap. But it's not used (at least in the sage distro) for a couple of years. So we remove the file, while leaving the mechanism in place (in case a certain distro needs it).

### :memo: Checklist

### DIFF
--- a/src/sage/libs/gap/sage.gaprc
+++ b/src/sage/libs/gap/sage.gaprc
@@ -1,3 +1,0 @@
-# This file is run by Sage when initializing libgap via GAP_Initialize, and may
-# contain bug fixes/workarounds and/or any Sage-specific patches necessary for
-# Sage's libgap interface.
--- a/src/sage/libs/gap/util.pyx
+++ b/src/sage/libs/gap/util.pyx
@@ -192,6 +192,12 @@ cdef initialize():
     `GAP Manual <https://www.gap-system.org/Manuals/doc/ref/chap3.html>`_
     Specifically, the value is for `-s` and `-o` options.
 
+    .. NOTE::
+
+        Additional initialization commands (e.g., bug fixes/workarounds 
+        and/or any Sage-specific patches) can be put in 
+        ``sage.gaprc`` (in the same folder as this file here).
+
     TESTS::
 
         sage: libgap(123)   # indirect doctest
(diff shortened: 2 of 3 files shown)


mkoeppe:
What problem is solved by this?

dimpase:
in view of [GAP docs](https://docs.gap-system.org/doc/ref/chap3_mj.html#X87DF11C885E73583) on this, it appears to be a semi-obsolete feature; for compatibility these files, and these read by GAP, ought to be named appropriately and placed in GAP's standard dirs. Probably reading these things ought to be delegated to libgap, not 100% sure about it.

Note that `sage.gaprc` is also mentioned in `src/setup.cfg.m4`

dimpase (changes requested):
src/setup.cfg.m4 mentions the file you're removing

Also, we need to rectify how to influence the initialisation of libgap by using such a file

tobiasdiez:
> What problem is solved by this?

Just a bit of cleanup/removal of semi-dead code.

> src/setup.cfg.m4 mentions the file you're removing

Thanks, I've removed it now.

> Also, we need to rectify how to influence the initialisation of libgap by using such a file

Not sure what you mean by "rectify". The process of adding a `sage.gaprc` file still works, in case it is needed at some point.

dimpase:
> > Also, we need to rectify how to influence the initialisation of libgap by using such a file
> 
> Not sure what you mean by "rectify". The process of adding a `sage.gaprc` file still works, in case it is needed at some point.

It's got to be documented, that's what I meant.

tobiasdiez:
Okay, I've now added a bit of documentation.

dimpase:
one has to figure out what GAP docs to cite, I suppose these for `gap.ini` - as it's ahead of any other initialisations

tobiasdiez:
Like this?

dimpase:
The GAP URL looks volatile. Also, please mention `gap.ini` explicitly

tobiasdiez:
> The GAP URL looks volatile. 

I agree, but it's the only deep link that's available (and it seem to be the same for at least the last year).

> Also, please mention `gap.ini` explicitly

Done.

PR #36759, opened 2023-11-23
allow passing a generating set (not necessarily a basis) for quaternion ideals

yyyyx4:
Most `.ideal()`-type methods in Sage take a list of generators. For quaternion orders, however, it is required that the given generators actually form a basis of the ideal. This is unexpected and inefficient, as it forces users to write more code in order to first compute a basis on their own. In this patch we add support for passing a list of generators instead of a basis.


src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1735,47 +1735,91 @@ def discriminant(self):
...
+        Return the left ideal of this order generated by the given generators.
 
         INPUT:
 
         - ``gens`` -- a list of elements of this quaternion order
 
         - ``check`` -- bool (default: ``True``)
 
+        - ``is_basis`` -- bool (default: ``False``); if ``True`` then ``gens``
+          must be a `\ZZ`-basis of the ideal.
```
tscrim:
```suggestion
          must be a `\ZZ`-basis of the ideal
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1735,47 +1735,91 @@ def discriminant(self):
...
+        Return the right ideal of this order generated by the given generators.
 
         INPUT:
 
         - ``gens`` -- a list of elements of this quaternion order
 
         - ``check`` -- bool (default: ``True``)
 
+        - ``is_basis`` -- bool (default: ``False``); if ``True`` then ``gens``
+          must be a `\ZZ`-basis of the ideal.
```
tscrim:
```suggestion
          must be a `\ZZ`-basis of the ideal
```

tscrim (approved):
LGTM. Just remove the two periods/full-stops at the end of the `INPUT:` lines to match Sage's doc conventions and you can set a positive review.

tscrim:
Thanks!

PR #36760, opened 2023-11-23
print more information about number-field orders

yyyyx4:
Currently, the `repr()` method for orders in number fields is not very useful: It prints the same information for all orders in a given field. In this patch we add the generators, and, for quadratic fields, the conductor of the order to the output of `repr()`.


src/sage/rings/number_field/order.py
```diff
@@ -2082,7 +2082,16 @@ def _repr_(self):
                     s = "Gaussian Integers"
         else:
             s = "Order"
-        return s + " in " + repr(self._K)
+            if self.number_field().absolute_degree() == 2:
+                f = (self.discriminant() / self.number_field().discriminant()).sqrt()
+                s += f' of conductor {f}'
+        try:
```
kwankyu, line 2111: For separation of concerns, you may put the conductor formula into the method `self.conductor()` and just use the method here?

src/sage/rings/number_field/order.py
```diff
@@ -2082,7 +2082,16 @@ def _repr_(self):
...
+            gens = self.ring_generators()
+        except TypeError:
+            gens = list(self.gens())
+        s += f' generated by {gens[0] if len(gens) == 1 else gens}'
```
kwankyu: Do we need to convert to a list here? Perhaps it's for consistency with `self.ring_generators()` which returns a list. 

Instead of converting and creating a new list (or tuple), you may just print the generators and add '[' and ']' on both sides in L2092 if `len(gens) > 1`.

src/sage/rings/number_field/order.py
```diff
@@ -2226,9 +2235,17 @@ def _repr_(self):
...
+            'Relative Order generated by [(-b^2 + 6*b + 11)*a + 3*b^2 + b + 14, (-2*b^2 - 2)*a - 2*b^2 - b, (b + 6)*a + 3*b^2 + 6] in Number Field in a with defining polynomial x^2 + x + 1 over its base field'
         """
-        return "%sRelative Order in %r" % ("Maximal " if self._is_maximal() else "", self._K)
+        s = 'Maximal ' if self._is_maximal() else ''
+        s += 'Relative Order'
+        try:
+            gens = self.ring_generators()
+        except TypeError:
+            gens = list(self.gens())
+        s += f' generated by {gens}'
```
kwankyu: The same comment here.

src/sage/rings/number_field/order.py
kwankyu: This is a minor point. You may ignore this comment if you will.

kwankyu: By the way, you may only print the (comma-separated) generators without "[...]".

kwankyu: Thus `' generated by {}'.format(', '.join(gens))` suffices.

yyyyx4: Personally I would keep the brackets since those lists can get quite long for general number fields and the brackets are making it easier to parse visually. If you have a strong opinion about this let me know; I don't.

kwankyu (approved):
Thanks. LGTM.

src/sage/rings/number_field/order.py
kwankyu: OK, if you like it.

PR #36764, opened 2023-11-24
Improve performance of __getitem__ of CFiniteSequence

### LINKED ISSUE #36763: Performance of __getitem__ of CFiniteSequence
### Problem Description

The function `__getitem__` for CFiniteSequence employs the matrix multiplications. Let d be the length of the linear recurrence (= the degree of the denominator of the generating function). Then, the computation of the n-th element requires time proportinal to d^3 log n.
For example,
`sage: C(1/(1-x^1000))[10^18]`
requries a couple of seconds. This is because the current program computes 10^18-th power of a 1000 x 1000 matrix.

### Proposed Solution

There are mainly two algorithms requiring time proportinal to M(d) log n where M(d) is a time for computing a product of two polynomials of degree d. The naive algorithm for the polynomial multiplications gives M(d) = d^2. Other algorithms, e.g., Karatsuba, FFT, give asymptotically faster polynomial multiplications.
The newer algorithm for computing the n-th element of CFiniteSequence was proposed in https://arxiv.org/abs/2008.08822
This algorithm is easier to implement.

### Alternatives Considered

Fiduccia's algorithm (https://doi.org/10.1137/0214007) has asymptotically similar time complexity. But, it is 1.5 times slower than the above algorithm.

### Additional Information

_No response_

ryuhei-mori:
Efficiency of `__getitem__` for CFiniteSequence is improved.

Fixes #36763

### :memo: Checklist

### DIFF
--- a/src/sage/rings/cfinite_sequence.py
+++ b/src/sage/rings/cfinite_sequence.py
@@ -661,12 +661,15 @@ class CFiniteSequence(FieldElement,
             (quo, num) = num.quo_rem(den)
             P = self.parent().polynomial_ring()
             x = self.parent().gen()
-            m = n
-            while m:
-                num = P((num * den(-x)).list()[m % 2::2])
+            if quo.degree() < n:
+                wp = 0
+            else:
+                wp = quo[n]
+            while n:
+                num = P((num * den(-x)).list()[n % 2::2])
                 den = P((den * den(-x)).list()[::2])
-                m //= 2
-            return quo[n] + num[0] / den[0]
+                n //= 2
+            return wp + num[0] / den[0]
         else:
             raise TypeError("invalid argument type")
 
--- a/src/sage/rings/cfinite_sequence.py
+++ b/src/sage/rings/cfinite_sequence.py
@@ -644,6 +644,14 @@ class CFiniteSequence(FieldElement,
             [0, 0, 1, 2, 3, 4, 5, 6, 7, 8]
             sage: s = C(x^3 * (1 - x)^-2); s[0:10]
             [0, 0, 0, 1, 2, 3, 4, 5, 6, 7]
+            sage: s = C(1/(1-x^1000)); s[10^18]
+            1
+            sage: s = C(1/(1-x^1000)); s[10^20]
+            1
+
+        REFERENCES:
+
+        - [BM2021]_
         """
         if isinstance(key, slice):
             m = max(key.start, key.stop)
--- a/src/sage/rings/cfinite_sequence.py
+++ b/src/sage/rings/cfinite_sequence.py
@@ -649,25 +649,26 @@ class CFiniteSequence(FieldElement,
             m = max(key.start, key.stop)
             return [self[ii] for ii in range(*key.indices(m + 1))]
         elif isinstance(key, Integral):
+            n = key - self._off
+            if n < 0:
+                return 0
             den = self.denominator()
             num = self.numerator()
             if self._off >= 0:
                 num = num.shift(-self._off)
             else:
                 den = den.shift(self._off)
-            n = key - self._off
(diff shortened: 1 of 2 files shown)


src/sage/rings/cfinite_sequence.py
```diff
@@ -650,28 +649,27 @@ def __getitem__(self, key):
...
-                return wp
-            A = Matrix(QQ, 1, d, self._c)
-            B = Matrix.identity(QQ, d - 1)
-            C = Matrix(QQ, d - 1, 1, 0)
-            if quo == 0:
-                off = self._off
-                V = Matrix(QQ, d, 1, self._a[:d][::-1])
+            (quo, num) = num.quo_rem(den)
+            P = self.parent().polynomial_ring()
+            x = self.parent().gen()
```
mezzarobba: I find it suspect that `CFiniteSequences(ZZ).gen()` returns a polynomial rather than a C-finite sequence, so I would write `x = P.gen()` here to be safe.

src/sage/rings/cfinite_sequence.py
```diff
@@ -650,28 +649,27 @@ def __getitem__(self, key):
...
+                wp = 0
             else:
-                off = 0
-                V = Matrix(QQ, d, 1, self._aa[:d][::-1])
-            M = Matrix.block([[A], [B, C]], subdivide=False)
-
-            return wp + list(M ** (key - off) * V)[d - 1][0]
+                wp = quo[n]
+            while n:
+                num = P((num * den(-x)).list()[n % 2::2])
```
mezzarobba: `den(-x)` actually takes quite a bit of time to compute in practice with the current implementation, so you may want to do it only once.

mezzarobba (commented):
Thank you for your contribution! The code looks good to me. In addition to the two comments I already left, I would suggest that you add a reference to your paper with Alin in the docstring, and maybe an example that shows that the code can handle recurrences of higher order.

It could be nice to also improve the case where one wants a whole slice of the sequence, but maybe not in this PR.

src/sage/rings/cfinite_sequence.py
ryuhei-mori: Thank you. I think both are equivalent in the current implementation as in the definition of `gen()` in this file. But, 'P.gen()' is slightly more efficient.

src/sage/rings/cfinite_sequence.py
ryuhei-mori: Thank you. I improved it. In my enviorment, the improvment was very small. I also tried

```python
nden = P([v if i % 2 == 0 else -v for i, v in enumerate(den)])
```

But, it was slower.

ryuhei-mori:
Thank you! I found you teached this algorithm in your course :)
I will add the docstring. Yes, I plan to implement the algorithm for slices.

By the way, I hope `CFiniteSequences` can be defined for general ring or field.
I think only `closed_form()` requires that the ring is `QQ` (or QQbar essentially).
Probably, `guess()` requires that it is field since the Berlekamp--Massey is explained by the Euclidean algorithm.
Except for them, I guess that we can allow general ring for `CFiniteSequences`.

src/sage/rings/cfinite_sequence.py
mezzarobba: Yes, I agree they are equivalent, but I view the current definition of `gen()` as a bug.

mezzarobba:
> Except for them, I guess that we can allow general commutative rings for `CFiniteSequences`.

I haven't looked in detail at the rest of the code, but this seems likely, and yes, this would be nice to have.

Let's wait for someone with the right permissions to approve the test workflows. (@fchapoton maybe?)

mezzarobba (approved):
Thank you!

PR #35250, opened 2023-03-08
Fix doctests for nauty 2.8.6 output changes

tornaria:
### 📚 Description

See #35157 and #34133.

This changes doctests to accomodate to changes in nauty output (which are only used for debug purposes).

OTOH, there is a bug in the released nauty 2.8.6 which causes another doctest failure reported in #35157. On a system with patched nauty 2.8.6 the current PR is good enough.

Leave #35157 open. To fix it a good option might be running `gentreeg 2` at configure time and check the bug is not present so it's possible to use 2.8.6 from system if it's been patched. In case it's useful, the patch I'm using is https://gitweb.gentoo.org/repo/gentoo.git/plain/sci-mathematics/nauty/files/nauty-2.8.6-gentreeg-gentourng.patch

### 📝 Checklist

- [x] I have made sure that the title is self-explanatory and the description concisely explains the PR.
- [x] I have linked an issue or discussion.

PR #36771, opened 2023-11-25
fix doctest for nauty 2.8.8

tornaria:
Updating to nauty 2.8.8 has one test failing. Fix it in a backwards compatible way.

Note that nauty output is only debug, and the fact that sagemath tests this output is dubious (see #35250).


mkoeppe (approved):
LGTM

PR #36773, opened 2023-11-26
Raise a power series to a power series

### LINKED ISSUE #36772: Raise a power series to a power series
### Problem Description

Currently, it is not possible to raise a power series to a power series. Doing so is desirable for generating function computations. (See for example Lecture 5 Exercise 10 in [Stanley's hyperplane notes](https://www.cis.upenn.edu/~cis6100/sp06stanley.pdf).) 
If `f` and `g` are power series, `f^g` can be defined as `(g * f.log()).exp()` as long as `g * f.log()` has no constant term. (See Chapter 2 of [generatingfunctionology](https://www2.math.upenn.edu/~wilf/gfology2.pdf).)

### Proposed Solution

The proposed solution is to add `(g * f.log()).exp()` into power series ring element's `__pow__` method so that `f^g` does the desired computation.

### Alternatives Considered

An alternative is to maintain the status quo whereby one has to manually compute `(g * f.log()).exp()`. 

### Additional Information

_No response_

trevorkarn:
This adds the functionality to compute the power series `f^g` when it is well defined.

This adds functionality that is useful for generating function computations.

Fixes #36772.
 

### :memo: Checklist

### :hourglass: Dependencies

N/A


trevorkarn:
As I understand it, the lint issues are coming from other portions of the code than what I changed.

src/sage/rings/power_series_ring_element.pyx
```diff
@@ -1135,11 +1136,35 @@ cdef class PowerSeries(AlgebraElement):
...
+            Traceback (most recent call last):
+            ...
+            ValueError: exponent must be a rational number or power series
+
         """
-        try:
-            right = QQ.coerce(r)
-        except TypeError:
-            raise ValueError("exponent must be a rational number")
+        if parent(r) is self.parent() or QQ.has_coerce_map_from(parent(r)):
```
tscrim: I think it would be good to split things up into cases with a bit more care. I would first test if `r` can be coerced into `QQ` and then run that case. Then if that is not the case, check if `r` can be coerced into `self.parent()` and if so then do the $\exp(r \log)$. The reason for the care is that things that coerce into `QQ` might also coerce into `self.parent()` (which I expect the general case to be slower). In both cases, I would probably do it by `try`-`except` blocks.

tscrim (changes requested):

Pietro-D:
s: needs-work

> As I understand it, the lint issues are coming from other portions of the code than what I changed.

The lint issues are related to some trailing spaces in your code

src/sage/rings/power_series_ring_element.pyx
Pietro-D: I think is a good solution to manage all the cases

fchapoton: done

fchapoton:
@tscrim, let us try to resurrect this one, please

trevorkarn:
The only potential issue is lint, correct?

fchapoton:
I have fixed the lint issue and rewritten the code as suggested by tscrim. This should be good to go, but please double check.

trevorkarn:
It looks good to me. Thanks for getting this across the finish line!

tscrim (approved):

PR #36776, opened 2023-11-26
Skip backport packages for newer python

orlitzky:
Three of our SPKGs are backports that provide python-3.11 features:

* importlib_metadata
* importlib_resources
* typing_extensions

As a result, they are not needed when sage's python is v3.11 or newer. We un-require them in `spkg-configure.m4`

### DIFF
--- a/m4/sage_python_package_check.m4
+++ b/m4/sage_python_package_check.m4
@@ -58,8 +58,17 @@ AC_DEFUN([SAGE_PYTHON_PACKAGE_CHECK], [
       AC_MSG_RESULT(yes)
       dnl strip all comments from install-requires.txt; this should leave
       dnl only a single line containing the version specification for this
-      dnl package.
-      SAGE_PKG_VERSPEC=$(sed '/^#/d' "./build/pkgs/$1/install-requires.txt")
+      dnl package. Afterwards, convert all double-quotes to single quotes.
+      dnl Both work, but only single quotes are documented. However, at the
+      dnl time of writing, double quotes are more compatible with our toml
+      dnl generation in ./bootstrap. Converting them from double- to single-
+      dnl quotes on-the-fly here lets us support both (in this macro, at
+      dnl least).
+      SAGE_PKG_VERSPEC=$(sed                   \
+        -e '/^#/d'                             \
+        -e "s/\"/'/g"                          \
+        "./build/pkgs/$1/install-requires.txt"
+      )
       AC_MSG_CHECKING([for python package $1 ("${SAGE_PKG_VERSPEC}")])
 
       dnl To prevent user-site (pip install --user) packages from being
@@ -77,10 +86,10 @@ AC_DEFUN([SAGE_PYTHON_PACKAGE_CHECK], [
 
       dnl double-quote SAGE_PKG_VERSPEC because platform-specific
       dnl dependencies like python_version<'3.11' will have single
-      dnl quotes in them.
+      dnl quotes in them. (We normalized the quotes earlier with sed.)
       AS_IF(
-        [PYTHONUSERBASE="${PYTHONUSERBASE}" config.venv/bin/python3 -c dnl
-           "import pkg_resources;                                      dnl
+        [PYTHONUSERBASE="${PYTHONUSERBASE}" config.venv/bin/python3 -c   dnl
+           "import pkg_resources;                                        dnl
             pkg_resources.require(\"${SAGE_PKG_VERSPEC}\".splitlines())" dnl
 	 2>&AS_MESSAGE_LOG_FD],
         [AC_MSG_RESULT(yes)],
--- a/m4/sage_python_package_check.m4
+++ b/m4/sage_python_package_check.m4
@@ -75,10 +75,13 @@ AC_DEFUN([SAGE_PYTHON_PACKAGE_CHECK], [
         PYTHONUSERBASE="${HOME}/.sage/local"
       ])
 
+      dnl double-quote SAGE_PKG_VERSPEC because platform-specific
+      dnl dependencies like python_version<'3.11' will have single
+      dnl quotes in them.
       AS_IF(
         [PYTHONUSERBASE="${PYTHONUSERBASE}" config.venv/bin/python3 -c dnl
            "import pkg_resources;                                      dnl
-            pkg_resources.require('${SAGE_PKG_VERSPEC}'.splitlines())" dnl
+            pkg_resources.require(\"${SAGE_PKG_VERSPEC}\".splitlines())" dnl
 	 2>&AS_MESSAGE_LOG_FD],
         [AC_MSG_RESULT(yes)],
         [AC_MSG_RESULT(no); sage_spkg_install_$1=yes]
--- a/build/pkgs/importlib_metadata/install-requires.txt
+++ b/build/pkgs/importlib_metadata/install-requires.txt
@@ -1,3 +1,3 @@
 # According to https://pypi.org/project/importlib-metadata/,
 # 4.13 provides the features of Python 3.11 importlib.metadata
-importlib_metadata >=4.13
+importlib_metadata >=4.13; python_version<'3.11'
--- a/build/pkgs/importlib_metadata/install-requires.txt
+++ b/build/pkgs/importlib_metadata/install-requires.txt
@@ -1,3 +1,3 @@
 # According to https://pypi.org/project/importlib-metadata/,
 # 4.13 provides the features of Python 3.11 importlib.metadata
-importlib_metadata >=4.13; python_version<'3.11'
+importlib_metadata >=4.13; python_version<"3.11"
--- a/build/pkgs/importlib_resources/install-requires.txt
+++ b/build/pkgs/importlib_resources/install-requires.txt
@@ -1,3 +1,3 @@
 # According to https://pypi.org/project/importlib-resources/,
 # version 5.7 provides the features of Python 3.11 importlib.resources
-importlib_resources >= 5.7
+importlib_resources >= 5.7; python_version<'3.11'
--- a/build/pkgs/importlib_resources/install-requires.txt
+++ b/build/pkgs/importlib_resources/install-requires.txt
@@ -1,3 +1,3 @@
 # According to https://pypi.org/project/importlib-resources/,
 # version 5.7 provides the features of Python 3.11 importlib.resources
-importlib_resources >= 5.7; python_version<'3.11'
+importlib_resources >= 5.7; python_version<"3.11"
--- a/build/pkgs/typing_extensions/install-requires.txt
+++ b/build/pkgs/typing_extensions/install-requires.txt
@@ -1,3 +1,3 @@
 # According to https://github.com/python/typing_extensions/blob/main/CHANGELOG.md,
 # version 4.4.0 adds another Python 3.11 typing backport
-typing_extensions >= 4.4.0; python_version<'3.11'
+typing_extensions >= 4.4.0; python_version<"3.11"
--- a/build/pkgs/typing_extensions/install-requires.txt
+++ b/build/pkgs/typing_extensions/install-requires.txt
@@ -1,3 +1,3 @@
 # According to https://github.com/python/typing_extensions/blob/main/CHANGELOG.md,
 # version 4.4.0 adds another Python 3.11 typing backport
-typing_extensions >= 4.4.0
+typing_extensions >= 4.4.0; python_version<'3.11'


orlitzky:
I ran `./configure` with python-3.10 and python-3.11 to ensure that the checks work but I'm entrusting the build itself to the CI.

mkoeppe:
Need to change install-requires.txt for these packages then to conditionalize on Python version.

orlitzky:
Hopefully that's the right syntax, I based it on https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

orlitzky:
The new version checks `$PYTHON_MINOR`, which should be set regardless of where python comes from, instead of executing python code. I also added an `spkg-configure.m4` for the new exceptiongroup package we'll be getting.

tornaria:
- https://docs.python.org/3/library/importlib.resources.html "New in version 3.7."
- https://docs.python.org/3/library/importlib.metadata.html "New in version 3.8."

tornaria:
Also: nothing in sagelib uses either package, and afaiu python modules are pip-installed. Seems highly redundant to document dependencies of dependencies.

It seems to me only dependencies of packages with custom install should be kept. E.g. typing_extensions is only a dependency of setuptools_scm, so it's not really a dependency of sagemath at all, and it shouldn't be listed in requirements of sagemath-standard, etc. BTW, there are tons of packages where `spkg-install.in` is nothing more than `sdh_pip_install .`, can those be removed?  

BTW `six` seems to be completely unnecessary (as in: not even dependencies need it anymore).

mkoeppe:
Clearer dependencies will be possible with #36740, please review

mkoeppe:
> * https://docs.python.org/3/library/importlib.resources.html "New in version 3.7."
> * https://docs.python.org/3/library/importlib.metadata.html "New in version 3.8."

Yes that's when the libraries were added, but newer Python versions provide new functions and, important for us, better support for PEP 420 namespace packages.

mkoeppe:
> afaiu python modules are pip-installed.

Normal packages (with checksums.ini) do not have access to PyPI. So we are installing all dependencies; and when installed from source, also build dependencies.

tornaria:
> Clearer dependencies will be possible with #36740, please review

I'm not interested in `build/*` at all. In fact, I want to be able to build sagelib without it. There's nothing in there that is really necessary at all but it's all tangled up.

mkoeppe:
OK why talk about spkg-* anything then?

mkoeppe:
src/pyproject.toml.m4 and src/setup.cfg.m4 is all that will matter to you then.

mkoeppe:
It's true that **currently** these backport packages are not used in sagelib.
We have declared them as dependencies so that sagelib developers can rely on the new library features even though we still support Python 3.9. See https://doc.sagemath.org/html/en/developer/coding_in_python.html#python-language-standard

tornaria:
> OK why talk about spkg-* anything then?

I just came by this PR since `src/setup.cfg` lists as required some packages that I don't think belong there. The most obvious ones being six, importlib_metadata, importlib_resources, my search led here, etc.

I made a comment in good faith. My mistake, sorry. I'll retract back to my corner of the world.

mkoeppe:
You are right that `six` should be removed from `src/setup.cfg.m4`. PR please?

mkoeppe:
> BTW, there are tons of packages where `spkg-install.in` is nothing more than `sdh_pip_install .`, can those be removed?

Yes, I think that could be a nice simplification.

tornaria:
> You are right that `six` should be removed from `src/setup.cfg.m4`. PR please?

Sure. What about importlib_resources, importlib_metadata, typing_extensions ? They don't seem to be required by sagemath-standard (eventually as a transitive dependency).

Also: sphinx, not really necessary for running sage, afaict.

Another thing is sage_conf: I do fine with a two-line `sage_conf.py` (only one line after #36792) although I'd rather have this as a configuration file inside `sage` package.

mkoeppe:
> Sure. What about importlib_resources, importlib_metadata, typing_extensions ? They don't seem to be required by sagemath-standard

As I said in https://github.com/sagemath/sage/pull/36776#issuecomment-1848793274, there there deliberately

mkoeppe:
> Also: sphinx, not really necessary for running sage, afaict.

This one comes in through sage.misc.sagedoc / sage.misc.sphinxify

mkoeppe:
> Another thing is sage_conf:

That's by design a required dependency. If your version of sage_conf is trivial, that's fine. What's intended is that on your system, "pip install sage_conf" will be a no-op because your system package already provides this distribution.

mkoeppe:
I'd also like to get rid of the runtime dep on sphinx. See #33682 for a possible route.

tornaria:
> > Also: sphinx, not really necessary for running sage, afaict.
> 
> This one comes in through sage.misc.sagedoc / sage.misc.sphinxify

Nice to have: maybe. Required: nope.
```
$ python -m sphinxify
/usr/bin/python: No module named sphinxify
$ sage -t /usr/lib/python3.12/site-packages/sage/misc/{sagedoc.py,sphinxify.py}
...
All tests passed!
...
```

I don't think I'm missing any functionality here... am I?

tornaria:
> > Another thing is sage_conf:
> 
> That's by design a required dependency. If your version of sage_conf is trivial, that's fine. What's intended is that on your system, "pip install sage_conf" will be a no-op because your system package already provides this distribution.

1. sage_conf is optional, not available is equivalent to trivial (i.e. empty) afaict.
2. no, I don't provide a "distribution", just a 2 line python module is enough. I just package sagemath-standard and I write by hand the configuration that suits our distro. I'd rather prefer if there was a way to (optionally) do this configuration inside the sage package (I could patch sage/env.py but I'd rather not).
3. `pip install sage_conf` is broken ("Python 3.12.0 is too recent" ? it also lectures me to install a bunch of distro packages, but every single one of them is already installed).

dimpase:
>     3. `pip install sage_conf` is broken ("Python 3.12.0 is too recent" ? it also lectures me to install a bunch of distro packages, but every single one of them is already installed).

Python 3.12.0 is not supported yet, and we probably should skip it. It's a bit wonky: https://github.com/python/cpython/issues/112215 - one can't control recursion depth any more.

tornaria:
> > ```
> > 3. `pip install sage_conf` is broken ("Python 3.12.0 is too recent" ? it also lectures me to install a bunch of distro packages, but every single one of them is already installed).
> > ```
> 
> Python 3.12.0 is not supported yet, and we probably should skip it. It's a bit wonky: [python/cpython#112215](https://github.com/python/cpython/issues/112215) - one can't control recursion depth any more.

:man_shrugging: that's what I have. I'm not using a 1.5G tarball for 100M of code. If sagemath couldn't work with my distro, I'd move on. Luckily sage works just fine with python 3.12 and sage_conf is not required.

orlitzky:
> Sure. What about importlib_resources, importlib_metadata, typing_extensions ? They don't seem to be required by sagemath-standard (eventually as a transitive dependency).

I'm trying to skip them rather than remove them mainly because I don't have python-3.9 to test. If nothing in sagelib needs them, I do think it's kind of silly to keep them as dependencies _just in case_ someone wants to use python-3.11 features.

orlitzky:
> I'm not interested in `build/*` at all. In fact, I want to be able to build sagelib without it. There's nothing in there that is really necessary at all but it's all tangled up.

There is now a clear path to this. First, we need to figure out how to make the system GAP usable. I just did that. Next we need to package all of sage's "data" packages. I did conway-polynomials, and I think there are four more. Nothing exciting here. The data packages are needed so that a sagelib build system can find the data on its own. Then, once every sagelib dependency can be detected by a build system, we should copy/paste all of the `./configure` checks that we currently do into Tobias's meson branch. Meson should detect and configure everything itself with no sage-conf and no `build/*`. All we'd have to do is `meson setup; meson compile`.

tornaria:
> > Sure. What about importlib_resources, importlib_metadata, typing_extensions ? They don't seem to be required by sagemath-standard (eventually as a transitive dependency).
> 
> I'm trying to skip them rather than remove them mainly because I don't have python-3.9 to test. If nothing in sagelib needs them, I do think it's kind of silly to keep them as dependencies _just in case_ someone wants to use python-3.11 features.

Sure, I'm not advocating for removing them from "sage the distribution", just from requirements.txt of sagemath-standard distribution. But I see you already do that (conditional to python >= 3.11) so thumbs up.

dimpase:
> > > ```
> > > 3. `pip install sage_conf` is broken ("Python 3.12.0 is too recent" ? it also lectures me to install a bunch of distro packages, but every single one of them is already installed).
> > > ```
> > 
> > 
> > Python 3.12.0 is not supported yet, and we probably should skip it. It's a bit wonky: [python/cpython#112215](https://github.com/python/cpython/issues/112215) - one can't control recursion depth any more.
> 
> 🤷‍♂️ that's what I have. I'm not using a 1.5G tarball for 100M of code. If sagemath couldn't work with my distro, I'd move on. Luckily sage works just fine with python 3.12 and sage_conf is not required.

we'd put https://github.com/python/cpython/issues/112215 into a doctest. 
python x.y.0 are always weird a bit. So Void, unlike any other distro I know, only carries 1 version of Python?
How'd people port their Python code if needed?

tornaria:
> > I'm not interested in `build/*` at all. In fact, I want to be able to build sagelib without it. There's nothing in there that is really necessary at all but it's all tangled up.
> 
> There is now a clear path to this. First, we need to figure out how to make the system GAP usable. I just did that. Next we need to package all of sage's "data" packages. I did conway-polynomials, and I think there are four more. Nothing exciting here. The data packages are needed so that a sagelib build system can find the data on its own. Then, once every sagelib dependency can be detected by a build system, we should copy/paste all of the `./configure` checks that we currently do into Tobias's meson branch. Meson should detect and configure everything itself with no sage-conf and no `build/*`. All we'd have to do is `meson setup; meson compile`.

System gap is usable. The system data packages I place in `/usr/share/sagemath` and they are usable (although `SAGE_SHARE` needs to be set to that, which in turns means I have to set `GAP_SHARE_DIR` and `JMOL_DIR`). sagelib (i.e. sagemath-standard) can be built using setuptools but one has to do some acrobacies. It would be nice if one could just `cd sagemath-standard && python -m build --no-isolation --wheel` and be done with it.

tornaria:
> we'd put [python/cpython#112215](https://github.com/python/cpython/issues/112215) into a doctest.

If you were to put that in a doctest, you force me to either patch it out, or stop shipping sagemath (which, for me, it would mean stop using it). I think we should move in the direction of removing these kind of tests, instead of adding them. Sage is completely usable even if you think that is a bug.

> python x.y.0 are always weird a bit. So Void, unlike any other distro I know, only carries 1 version of Python? How'd people port their Python code if needed?

It's never that hard. The low level stuff is the more complicated (e.g. they seem to enjoy playing with integers lately) but it's mostly concentrated in a few packages that abstract the changes away. Other stuff usually has deprecation warnings and packages that are well maintained don't take too long to adjust.

It's ok if you prefer to move a little bit slower; I never proposed that sagemath moves to python 3.12, only that requirements.txt stops claiming that it won't work with python 3.12, because it does work. Ultimately, I don't really care about that either; except I tried to pip install sage_conf to see how it works and I was forbidden to do it (and I don't think sage_conf is going to be affected by the recursion issue, is it?).

dimpase:
my point is that major Python3 upgrades introduce incompatibilities.

A distro that's serious about its users ought to keep an older version of Python around, just in case.
(e.g. if you wrote serious deeply recursive code than 3.11 -> 3.12.0 will spell big trouble for you)

mkoeppe:
> I never proposed that sagemath moves to python 3.12, only that requirements.txt stops claiming that it won't work with python 3.12, because it does work.

@tornaria Given your report of this, I would support making this change. PR please?

mkoeppe:
> > > Also: sphinx, not really necessary for running sage, afaict.
> > 
> > This one comes in through sage.misc.sagedoc / sage.misc.sphinxify
> 
> Nice to have: maybe. Required: nope.

@tornaria I know. I made that work in #32759.

But as you observed elsewhere, it provides the formatting of the docstrings in the interactive help.

That's a standard functionality, so it's a dependency of **sagemath-standard**.

orlitzky:
> System gap is usable. The system data packages I place in `/usr/share/sagemath` and they are usable (although `SAGE_SHARE` needs to be set to that, which in turns means I have to set `GAP_SHARE_DIR` and `JMOL_DIR`). sagelib (i.e. sagemath-standard) can be built using setuptools but one has to do some acrobacies. It would be nice if one could just `cd sagemath-standard && python -m build --no-isolation --wheel` and be done with it.

Yeah that's exactly what I mean. You shouldn't have to run your own distro to be able to build sagelib. Sage should be able to detect GAP and find those databases on its own rather than requiring you to place them exactly where it wants them and/or hand-edit conf.py.

mkoeppe:
> > I never proposed that sagemath moves to python 3.12, only that requirements.txt stops claiming that it won't work with python 3.12, because it does work.
> 
> @tornaria Given your report of this, I would support making this change. PR please?

That's now #36869

build/pkgs/importlib_metadata/spkg-configure.m4
```diff
@@ -1,3 +1,26 @@
...
+  # python-3.11 features (see coding_in_python.rst):
+  #
+  #   * importlib_metadata
+  #   * importlib_resources
+  #   * typing_extensions
+  #
+  # These packages are therefore not needed with >=python-3.11. Here
+  # we test for a python minor version component greater than or equal
+  # to 11, and mark this package as "not required" if we succeed.
+  AC_MSG_CHECKING([for >=python-3.11])
```
mkoeppe: Why does the Python version need a separate check in `spkg-configure.m4`? 
Does the conditionalization in `install-requires.txt` not do the job already? Checking whether `importlib_metadata >=4.13; python_version<'3.11'` is available will be trivially true for Python >= 3.11

orlitzky: Oh. I wrote that check before you pointed out that we needed to change install-requires.txt. It's probably redundant now but I have to get flint-3.x packaged before I can run `./configure` now.

orlitzky: I had to change the quoting in the macro but it basically works, if you don't care that it says "will use the system foo" rather than "foo is not required" when it skips the package.

mkoeppe: I think that's fine

m4/sage_python_package_check.m4
```diff
@@ -75,10 +75,13 @@ AC_DEFUN([SAGE_PYTHON_PACKAGE_CHECK], [
         PYTHONUSERBASE="${HOME}/.sage/local"
       ])
 
+      dnl double-quote SAGE_PKG_VERSPEC because platform-specific
+      dnl dependencies like python_version<'3.11' will have single
+      dnl quotes in them.
```
mkoeppe: Actually, you can use double quotes in install-requires.txt.

And in fact double quotes are better because then `m4/sage_spkg_versions_toml.m4` will be generated correctly!

mkoeppe: (relevant for #36951)

orlitzky: Ok, double quotes it is. I changed the package-check macro to normalize them (back to single quotes) so that both are supported, at least in that one macro.

mkoeppe (approved):
LGTM, thanks.

PR #36783, opened 2023-11-26
Drop Cygwin handling in spkg-install and spkg-install.in 

orlitzky:
This mainly consists of

```
$ git grep -i cygwin build/pkgs/*/spkg-install*
```

and then deleting the resulting `[ "${UNAME}" = "CYGWIN" ]` cases. Two instances (under sagelib and pycygwin) have been omitted because they are being handled elsewhere.


orlitzky:
Argh, I didn't see that you made these same changes on the other PR.

PR #36794, opened 2023-12-01
`sage --package create --pypi`: Create a wheel package by default if possible

mkoeppe:
We change the default of `sage --package create --pypi` to create a wheel package (if the package has a platform-independent wheel on PyPI) instead of a normal package (built from source).
This is simpler (no spkg-install.in script, no build dependencies) and a bit faster.

As an illustration, we change one SPKG, `six`, from normal to wheel.

### :memo: Checklist

### DIFF
--- a/build/pkgs/six/dependencies
+++ b/build/pkgs/six/dependencies
@@ -1,4 +1,4 @@
- | $(PYTHON_TOOLCHAIN) $(PYTHON)
+ | pip $(PYTHON)
 
 ----------
 All lines of this file are ignored except the first.
--- a/build/sage_bootstrap/app.py
+++ b/build/sage_bootstrap/app.py
@@ -290,8 +290,9 @@ class Application(object):
         if pypi:
             if source is None:
                 try:
-                    source = 'wheel'
-                    if not PyPiVersion(package_name, source='wheel').tarball.endswith('-none-any.whl'):
+                    if PyPiVersion(package_name, source='wheel').tarball.endswith('-none-any.whl'):
+                        source = 'wheel'
+                    else:
                         source = 'normal'
                 except PyPiError:
                     source = 'normal'
--- a/build/sage_bootstrap/app.py
+++ b/build/sage_bootstrap/app.py
@@ -273,7 +273,7 @@ class Application(object):
             update.fix_checksum()
 
     def create(self, package_name, version=None, tarball=None, pkg_type=None, upstream_url=None,
-               description=None, license=None, upstream_contact=None, pypi=False, source='normal'):
+               description=None, license=None, upstream_contact=None, pypi=False, source=None):
         """
         Create a package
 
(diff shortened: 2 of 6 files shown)


build/sage_bootstrap/app.py
```diff
@@ -288,6 +288,13 @@ def create(self, package_name, version=None, tarball=None, pkg_type=None, upstre
...
             raise ValueError('package names must not contain dashes, use underscore instead')
         if pypi:
+            if source is None:
+                try:
+                    source = 'wheel'
+                    if not PyPiVersion(package_name, source='wheel').tarball.endswith('-none-any.whl'):
+                        source = 'normal'
+                except PyPiError:
+                    source = 'normal'
             pypi_version = PyPiVersion(package_name, source=source)
```
kwankyu, line 299:
```diff
             if source is None:
                 try:
-                    source = 'wheel'
-                    if not PyPiVersion(package_name, source='wheel').tarball.endswith('-none-any.whl'):
+                    if PyPiVersion(package_name, source='wheel').tarball.endswith('-none-any.whl'):
+                        source = 'wheel'
+                    else
                         source = 'normal'
                 except PyPiError:
```
a bit simpler code?

build/sage_bootstrap/cmdline.py
```diff
@@ -319,7 +319,7 @@ def make_parser():
         'package_name', default=None, type=str,
         help='Package name.')
     parser_create.add_argument(
-        '--source', type=str, default='normal', help='Package source (one of normal, wheel, script, pip)')
+        '--source', type=str, default=None, help='Package source (one of normal, wheel, script, pip; default depends on provided arguments)')
     parser_create.add_argument(
```
kwankyu, line 323:
```diff
     parser_create.add_argument(
-        '--source', type=str, default=None, help='Package source (one of normal, wheel, script, pip; default depends on provided arguments)')
+        '--source', type=str, default=None, help='Package source (one of normal, wheel, script, pip); default depends on provided arguments')
     parser_create.add_argument(
         '--version', type=str, default=None, help='Package version')
```
following the style of other arguments?

build/pkgs/six/dependencies
```diff
@@ -1,4 +1,4 @@
- | $(PYTHON_TOOLCHAIN) $(PYTHON)
+ | pip $(PYTHON)
```
kwankyu, line 2: `PYTHON_TOOLCHAIN` represents "a prerequisite to any pip-installed packages". It does not differentiate source and wheel python packages. Here changing it with just "pip" signals that `PYTHON_TOOLCHAIN` is only for source python packages.  Is it?

On the other hand, packages in `PYTHON_TOOLCHAIN` are installed anyway because of other packages. So it seems that the change has no real benefit...

mkoeppe: > changing it with just "pip" signals that `PYTHON_TOOLCHAIN` is only for source python packages. Is it?

Indeed. I have updated the documentation.

build/sage_bootstrap/app.py
mkoeppe: Thanks, done in 818468af835

build/sage_bootstrap/cmdline.py
mkoeppe: Done in 818468af835

build/pkgs/six/dependencies
kwankyu: Thanks. OK.

kwankyu (approved):
LGTM.

mkoeppe:
Thank you!

PR #36730, opened 2023-11-17
`pkgs/sagemath-doc-{html,pdf}`

mkoeppe:
We create pip-installable packages that provide the Sage documentation. This is a step toward #29868, which has the full design.

We want to install the HTML documentation in `share/doc/sage/html`. So we need a build system that supports such "non-package data".
- setuptools considers this deprecated: https://setuptools.pypa.io/en/latest/userguide/datafiles.html#non-package-data-files
- hatch supports it via `wheel.shared-data` https://hatch.pypa.io/latest/plugins/builder/wheel/#configuration; example: https://github.com/jupyter/nbconvert/blob/main/pyproject.toml#L93C27-L93C44; dynamic data: https://github.com/pypa/hatch/discussions/528
- meson-python is willing to ship whatever we throw in `{datadir}` https://meson-python.readthedocs.io/en/latest/reference/limitations.html#non-package-data-files; status of editable wheels to be checked.

For synergy with #34630, we use meson-python - with a trivial `pkgs/sagemath-doc-html/meson.build` file that just runs `make` with the existing Makefile.
The actual build happens using ninja.

In the current implementation:
- if `SAGE_EDITABLE` is set to `yes` (default), this uses `pip install -e`, which does the incremental docbuild (in `pkgs/sagemath-doc-*/build`) and creates symlinks in `$SAGE_LOCAL/share/doc/sage/` that point there.
- if `SAGE_WHEELS` is set to `yes`, this builds a wheel from scratch, which contains the HTML documentation as non-package data files. If `SAGE_EDITABLE` is not set, then it installs the wheel using `--prefix=$SAGE_LOCAL` so that the HTML documentation appears in the familiar location.

We do the same with PDF in sagemath-doc-pdf.

GH Actions workflows:
- On release tags, the workflow `dist.yml` 
 The workflow  (run on release tags) builds the sdists and uploads them to PyPI, along with the other pip-installable packages.
- On release tags, `doc-build.yml` builds the live doc (HTML + PDF) by going through the wheels and uploads to PyPI if the credentials are present.

To test:
- `make sagemath_doc_html` and `make sagemath_doc_pdf` should continue to work
- `make SAGE_WHEELS=yes sagemath_doc_html` and `make SAGE_WHEELS=yes sagemath_doc_pdf` should both work and create (architecture-independent) wheels in `prefix/var/lib/sage/wheels/`
[...]

PR #36799, opened 2023-12-02
Create index-pdf.html in html build

kwankyu:
When we `make doc-pdf`, `reference/index-pdf.html` is also built. With this PR, it  is built by default when we `make doc-html`.

Here it is: https://deploy-preview-36799--sagemath-tobias.netlify.app/html/en/reference/index-pdf

This is to fix the issue https://github.com/sagemath/sage/pull/36730#issuecomment-1837013688

### :memo: Checklist


mkoeppe:
Makes sense, thanks for this change!

kwankyu:
Thanks!

PR #36804, opened 2023-12-03
make the empty sum() a Sage Integer

yyyyx4:
As of Sage 10.2, the result of `sum([])` is a Python `int`, which can sometimes cause unexpected breakage. In this patch we make sure that `sum()` always returns a Sage object.


tscrim:
In principle I agree with you, but this can have major implications. If you want to guarantee something in `ZZ`, you should use `ZZ.sum()` (or whatever ring `R.sum()`). At the very least, we need to ask on sage-devel about this if you want to proceed.

yyyyx4:
Thanks for your comment! Closing this for now since I'm no longer convinced of this change.

PR #36805, opened 2023-12-03
fast path for Vélu isogenies with a single kernel generator

yyyyx4:
The current implementation of Vélu's formulas always assumes that the kernel subgroup is given by an arbitrary set of generators: The important special case of a single generator is handled using the same code as the general case.
In this patch we add a fast path to handle a single generator, which can yield big speedups. Example:
```sage
sage: E = EllipticCurve(GF(2^127-1), [1,1,1,1,1])
sage: P = E.lift_x(73833238617088645877502541346296796686)
sage: P.set_order(2543)
sage: %timeit E.isogeny(P)
```

Sage 10.2:
```
63.6 ms ± 588 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

With this patch:
```
33.3 ms ± 98.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


GiacomoPope (approved):
Tried to find something to suggest to change, but the PR is good and clean. 

This is an obvious thing to implement as most calls only include a single point rather than a list so this will be a "free speed boost" for most people.

Good addition to the isogeny code.

PR #36806, opened 2023-12-03
add generic has_order() function

yyyyx4:
Currently there is `order_from_multiple()` for generic groups, but sometimes we really only want to know if the order equals a particular value (e.g., when sampling random elements to find a generating set of a group with known structure). This means the computation can be aborted earlier than when finding the exact order is required.

In this patch we add such an optimized `has_order()` function as well as a convenience wrapper for elliptic-curve points.  An example of the speedup can be found in one of the new doctests.

### DIFF
--- a/src/sage/groups/generic.py
+++ b/src/sage/groups/generic.py
@@ -1181,6 +1181,8 @@ def linear_relation(P, Q, operation='+', identity=None, inverse=None, op=None):
 # 2. order_from_bounds: finds the order given an interval containing a
 # multiple of the order
 #
+# 3. has_order: check if the order is exactly equal to a given integer
+#
 ################################################################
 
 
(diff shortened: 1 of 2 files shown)


src/sage/groups/generic.py
```diff
@@ -1407,6 +1409,99 @@ def order_from_bounds(P, bounds, d=None, operation='+',
...
+def has_order(P, n, operation='+'):
+    r"""
+    Generic function to test if a group element `P` has order
+    exactly equal to a given positive integer `n`.
+
+    INPUT:
+
+    - `P` -- group element with respect to the specified ``operation``
+    - `n` -- positive integer, or its :class:`~sage.structure.factorization.Factorization`
+    - ``operation`` -- string, either ``'+'`` (default) or ``'*'``
```
tscrim:
```suggestion
    - ``P`` -- group element with respect to the specified ``operation``
    - ``n`` -- positive integer, or its :class:`~sage.structure.factorization.Factorization`
    - ``operation`` -- string, either ``'+'`` (default) or ``'*'``
```
We always make these code formatted.

src/sage/groups/generic.py
```diff
@@ -1407,6 +1409,99 @@ def order_from_bounds(P, bounds, d=None, operation='+',
...
+    .. NOTE::
+
+        In some cases, order *testing* can be much faster than
+        *computing* the order using :func:`order_from_multiple`.
+    """
+    if isinstance(n, sage.rings.integer.Integer):
+        n = n.factor()
+
+    G = P.parent()
+    if operation == '+':
```
tscrim: I think it would be better to use `operator.add` and `operator.mul` than string comparisons. Although strings seem to be used elsewhere...

src/sage/groups/generic.py
```diff
@@ -1407,6 +1409,99 @@ def order_from_bounds(P, bounds, d=None, operation='+',
...
+        mult = multiple
+    else:
+        raise ValueError('unknown group operation')
+
+    def _rec(Q, fn):
+        if not fn:
+            return isid(Q)
+
+        if len(fn) == 1:
+            p,k = fn[0]
```
tscrim:
```suggestion
            p, k = fn[0]
```
PEP8

src/sage/groups/generic.py
```diff
@@ -1407,6 +1409,99 @@ def order_from_bounds(P, bounds, d=None, operation='+',
...
+            for i in range(k):
+                if isid(Q):
+                    return False
+                Q = mult(Q, p)
+            return isid(Q)
+
+        fl, fr = fn[::2], fn[1::2]
```
tscrim:
```suggestion
        fl = fn[::2]
        fr = fn[1::2]
```
PEP8 (I also find this easier to read).

src/sage/groups/generic.py
```diff
@@ -1407,6 +1409,99 @@ def order_from_bounds(P, bounds, d=None, operation='+',
...
+    else:
+        raise ValueError('unknown group operation')
+
+    def _rec(Q, fn):
+        if not fn:
+            return isid(Q)
+
+        if len(fn) == 1:
+            p,k = fn[0]
+            for i in range(k):
```
tscrim:
```suggestion
            for _ in range(k):
```
lint (or PEP8?) since you don't use the iteration variable

tscrim (changes requested):

src/sage/groups/generic.py
yyyyx4: On the long run, all of these ad-hoc interfaces should be replaced by a proper wrapper class (which is work in progress). For now, I'd stick to doing it the same way as the existing functions in this file...

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -539,6 +539,47 @@ def __bool__(self):
...
+        Compare a cached order if available, otherwise use :func:`sage.groups.generic.has_order`.
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve('26b1')
+            sage: P = E(1, 0)
+            sage: P.has_order(7)
+            True
+            sage: P._order
+            7
```
tscrim, line 561: Can you add a test showing the cache is working (mainly calling it a second time)?

Can we also add the analogous tests for a factorization?

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -539,6 +539,47 @@ def __bool__(self):
         """
         return bool(self[2])
 
+    def has_order(self, n):
+        r"""
+        Test if this point has order exactly `n`.
+
+        INPUT:
+
+        - `n` -- integer, or its :class:`~sage.structure.factorization.Factorization`
```
tscrim:
```suggestion
        - ``n`` -- integer, or its :class:`~sage.structure.factorization.Factorization`
```

tscrim (changes requested):
Okay, fair enough. Then just a few more little things.

src/sage/schemes/elliptic_curves/ell_point.py
yyyyx4: Thanks, done.

tscrim: No, that actually doesn't test it, that "shows" the user that it is working. It needs to actually be called twice. Also, all of the factorization tests are `# not tested`.

yyyyx4: Added another (`tested`) test for `Factorization` objects. But I don't understand your other comment: The test checks that the `._order` attribute is set correctly after the call to `.has_order()`. How is this not sufficient?

tscrim: Ah, sorry, I missed that. Thank you.

tscrim: Actually, sorry, that isn’t correct. It isn’t that the attribute is set correctly but the code in the method that utilizes it is being run correctly. So the method needs to be called a second time after `_order` is set.

yyyyx4: Okay, I see. It should be done now.

tscrim (approved):
Thank you. LGTM now.

yyyyx4:
Thank you!

PR #36813, opened 2023-12-04
`m4/sage_spkg_configure.m4`: Improve doc of `SAGE_DEPCHECK`, add macro for reverse depcheck

orlitzky:
ipython is the only package that depends on backcall, so backcall can be skipped if ipython comes from the system. This micro-optimization is not so micro now that ipython has dropped backcall upstream:

  https://github.com/ipython/ipython/commit/1c2687ea5

As a result, systems with newer ipythons may not have backcall installed. We don't want to install a pointless SPKG on those systems.

### DIFF
--- a/m4/sage_spkg_configure.m4
+++ b/m4/sage_spkg_configure.m4
@@ -191,7 +191,7 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
 # DESCRIPTION
 #     *** to be called from SAGE_SPKG_CONFIGURE* ***
 #     check for space-separated list of package dependencies $1 of package SPKG_NAME
-#     do $2 if successful
+#     do $2 if none of $1 is (to be) installed by Sage, normally checks on suitability of the system SPKG_NAME
 #
 #     $3 is for off-label use - checking a reverse dependency, by REVDEPCHECK. Then the default action of
 #     setting sage_spkg_install_SPKG_NAME=yes whenever anything from $1 comes from Sage may be overwritten,
--- a/m4/sage_spkg_configure.m4
+++ b/m4/sage_spkg_configure.m4
@@ -208,3 +208,17 @@ AC_DEFUN([SAGE_SPKG_DEPCHECK], [
         $2
         ])
 ])
+
+
+# SYNOPSIS
+#
+#   SAGE_SPKG_REVDEPCHECK(PACKAGE-REVERSE-DEPENDENCIES)
+#                                 $1
+# DESCRIPTION
+#     *** to be called from SAGE_SPKG_CONFIGURE* ***
+#     check for space-separated list of package reverse dependencies $1 of package SPKG_NAME
+#     If they all come from the system, do not require SPKG_NAME
+#
+AC_DEFUN([SAGE_SPKG_REVDEPCHECK], [
+    SAGE_SPKG_DEPCHECK($1, [[sage_require_]SPKG_NAME=no], [])
+])
--- a/m4/sage_spkg_configure.m4
+++ b/m4/sage_spkg_configure.m4
@@ -185,7 +185,7 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
 
 # SYNOPSIS
 #
-#   SAGE_SPKG_DEPCHECK(PACKAGE-DEPENDENCIES, FURTHER-CHECK, [FAIL= [sage_spkg_install_]SPKG_NAME=yes])
+#   SAGE_SPKG_DEPCHECK(PACKAGE-DEPENDENCIES, FURTHER-CHECK, [DEPS_ARE_INSTALLED= [sage_spkg_install_]SPKG_NAME=yes])
 #                                 $1              $2                $3
 #
 # DESCRIPTION
@@ -193,8 +193,9 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
 #     check for space-separated list of package dependencies $1 of package SPKG_NAME
 #     do $2 if successful
 #
-#     $3 is for off-label use - checking a reverse dependency. Then the default action of
-#     setting sage_spkg_install_SPKG_NAME=yes may be overwritten, typically to [], i.e. nothing.
+#     $3 is for off-label use - checking a reverse dependency, by REVDEPCHECK. Then the default action of
+#     setting sage_spkg_install_SPKG_NAME=yes whenever anything from $1 comes from Sage may be overwritten,
+#     typically to [ ], i.e. a no-op (it can't be [], as m4_default checks non-emptiness)
 #
 AC_DEFUN([SAGE_SPKG_DEPCHECK], [
     m4_foreach_w([DEP], $1, [
@@ -202,7 +203,7 @@ AC_DEFUN([SAGE_SPKG_DEPCHECK], [
     AC_MSG_CHECKING([whether any of $1 is installed as or will be installed as SPKG])
     AS_IF([test x = y m4_foreach_w([DEP], $1, [ -o [x$sage_spkg_install_]DEP = xyes])], [
         AC_MSG_RESULT([yes; install SPKG_NAME as well])
-        m4_default($3, [sage_spkg_install_]SPKG_NAME=yes)
+        m4_default([$3], [sage_spkg_install_]SPKG_NAME=yes)
 	], [
         AC_MSG_RESULT([no])
         $2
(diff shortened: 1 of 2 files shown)


dimpase (changes requested):
why don't you use `SAGE_SPKG_DEPCHECK()` ?

orlitzky:
> why don't you use `SAGE_SPKG_DEPCHECK()` ?

IDK, this is how other packages such as filelock do it, and I copied that.

orlitzky:
> actually, any package `foo` with the unique reverse dependency `bar` coming from the system should automatically get `sage_spkg_install_foo=no`. 

Well, yes, obviously. Too bad our "package manager" is written in shell scripts and GNU make templates and can't create a dependency graph...

dimpase (approved):

orlitzky:
Ok, I'm not picky. Thanks.

dimpase:
Well, I tested it, it works. I don't know what is incorrect there.

mkoeppe:
No, Dima, you're mistaken. 
I'm pretty sure I had to explain it to you several times already.

dimpase:
here is my last change
```diff
-  AC_REQUIRE([SAGE_SPKG_CONFIGURE_IPYTHON])
-  AS_IF([test "${sage_spkg_install_ipython}" = "no"],[
-    sage_require_backcall=no
-  ])
+  SAGE_SPKG_DEPCHECK([ipython], [sage_require_backcall=no])
```
here is the macro body
```m4
AC_DEFUN([SAGE_SPKG_DEPCHECK], [
    m4_foreach_w([DEP], $1, [
       AC_REQUIRE([SAGE_SPKG_CONFIGURE_]m4_toupper(DEP))])
    AC_MSG_CHECKING([whether any of $1 is installed as or will be installed as SPKG])
    AS_IF([test x = y m4_foreach_w([DEP], $1, [ -o [x$sage_spkg_install_]DEP = xyes])], [
        AC_MSG_RESULT([yes; install SPKG_NAME as well])
        [sage_spkg_install_]SPKG_NAME=yes], [
        AC_MSG_RESULT([no])
        $2
        ])
])
```
evaluating for our case, removing AC_MSG and loops (or length 1), will become
```m4
       AC_REQUIRE([SAGE_SPKG_CONFIGURE_IPYTHON])
    AS_IF([test x$sage_spkg_install_ipython = xyes])], [
        [sage_spkg_install_backcall=yes], [
        sage_spkg_install_backcall=no
        ])
```

As `sage_spkg_install_foo` takes values `yes` or `no` only, the macro call is equivalent to what it is replacing.

mkoeppe:
> ```m4
>     AS_IF([test x$sage_spkg_install_ipython = xyes])], [
>         [sage_spkg_install_backcall=yes], [
> ```

Yes, this is what it does, and setting `sage_spkg_install_backcall=yes` is not correct.
As I explained to you in https://github.com/sagemath/sage/pull/36813#issuecomment-1839315794

dimpase:
> > ```m4
> >     AS_IF([test x$sage_spkg_install_ipython = xyes])], [
> >         [sage_spkg_install_backcall=yes], [
> > ```
> 
> Yes, this is what it does, and setting `sage_spkg_install_backcall=yes` is not correct. As I explained to you in [#36813 (comment)](https://github.com/sagemath/sage/pull/36813#issuecomment-1839315794)

What? If we are building `ipython`, we must build `backcall` too, as it is a dependency of `ipython`.

mkoeppe:
> If we are building `ipython`, we must build `backcall` too, as it is a dependency of `ipython`.

No.

dimpase:
> > If we are building `ipython`, we must build `backcall` too, as it is a dependency of `ipython`.
> 
> No.

Yes. `backcall` is a dependency of `ipython` 8.6.0, and as such it must be built from source if `ipython` is.

dimpase:
I have provided code which does the same as the original code. I have no idea why @mkoeppe is in denial about it.

mkoeppe:
> Yes. `backcall` is a dependency of `ipython` 8.6.0, and as such it must be built from source if `ipython` is.

Try to explain in more detail why that would be.

mkoeppe:
> I have provided code which does the same as the original code.

Also that is not true.

dimpase:
sorry, I am particularly dumb today. Here is a fix.

mkoeppe:
Strong -1 on overloading SPKG_DEPCHECK with a new and unrelated purpose.

dimpase:
> Strong -1 on overloading SPKG_DEPCHECK with a new and unrelated purpose.

it is a dependency check - just the reverse one. One can easily build a new macro to do exactly what's needed.

dimpase:
now we have SAGE_SPKG_REVDEPCHECK just for such occasions

m4/sage_spkg_configure.m4
```diff
@@ -185,22 +185,40 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
 
 # SYNOPSIS
 #
-#   SAGE_SPKG_DEPCHECK(PACKAGE-DEPENDENCIES, FURTHER-CHECK)
-#                                 $1              $2
+#   SAGE_SPKG_DEPCHECK(PACKAGE-DEPENDENCIES, FURTHER-CHECK, [FAIL= [sage_spkg_install_]SPKG_NAME=yes])
```
mkoeppe: The documentation needs more work, so that we don't have to repeat this type of conversation the next time.

`FAIL` is unclear -- what is failing. This needs clearer terminology.

m4/sage_spkg_configure.m4
```diff
@@ -185,22 +185,40 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
...
+#                                 $1              $2                $3
 #
 # DESCRIPTION
 #     *** to be called from SAGE_SPKG_CONFIGURE* ***
 #     check for space-separated list of package dependencies $1 of package SPKG_NAME
 #     do $2 if successful
```
mkoeppe: likewise here, the "successful" is unclear.

m4/sage_spkg_configure.m4
```diff
@@ -185,22 +185,40 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
...
+	], [
         AC_MSG_RESULT([no])
         $2
         ])
 ])
+
+
+# SYNOPSIS
+#
+#   SAGE_SPKG_REVDEPCHECK(PACKAGE-REVERSE-DEPENDENCIES)
```
mkoeppe, line 216: And the problem is that the abstraction (using this macro) obscures why it has to be done in the REQUIRED-CHECK ($2) argument of SAGE_SPKG_CONFIGURE.

m4/sage_spkg_configure.m4
```diff
@@ -185,22 +185,40 @@ AC_DEFUN([SAGE_SPKG_CONFIGURE], [
...
+#                                 $1
+# DESCRIPTION
+#     *** to be called from SAGE_SPKG_CONFIGURE* ***
+#     check for space-separated list of package reverse dependencies $1 of package SPKG_NAME
+#     If they all come from the system, do not require SPKG_NAME
+#
+AC_DEFUN([SAGE_SPKG_REVDEPCHECK], [
+    SAGE_SPKG_DEPCHECK($1, [[sage_require_]SPKG_NAME=no], [])
```
mkoeppe: Is this m4-quoted correctly?

dimpase: here yes, but need to pass `[ ]`, not `[]`, and quote `$3` in the call to `m4_define`.

m4/sage_spkg_configure.m4
dimpase: the abstraction always obscures the boilerplate it replaces. I'll document where it should go

m4/sage_spkg_configure.m4
dimpase: will be fixed

m4/sage_spkg_configure.m4
dimpase: please check the new version

m4/sage_spkg_configure.m4
dimpase: see https://github.com/sagemath/sage/pull/36813/commits/b5e2e6562b76913aa646e11de158b0880af93f40

m4/sage_spkg_configure.m4
mkoeppe: One can avoid this effect for example by finding the right name for the abstraction. `SAGE_SPKG_REVDEPCHECK` does not include the word REQUIRE. Perhaps it should

dimpase: > One can avoid this effect for example by finding the right name for the abstraction. `SAGE_SPKG_REVDEPCHECK` does not include the word REQUIRE. Perhaps it should

Feel free to propose a name.

mkoeppe: We could rename `SAGE_DEPCHECK` to `SAGE_SPKG_INSTALL_IF_ANY_DEP_INSTALLED_ELSE`

and `SAGE_SPKG_REVDEPCHECK` to `SAGE_SPKG_UNREQUIRE_IF_NO_REVDEP_INSTALLED`

which I think are high autoconfspeak style

mkoeppe:
> Why not just do the ipython upgrade and drop it altogether?

Done now in #36823

dimpase:
We need to extract the autoconf macros stuff from here - in a new PR?

mkoeppe:
If @orlitzky does not object, we can just reuse this PR for this.

orlitzky:
No, go ahead, you both have edit/force-push privileges.

orlitzky:
Dependency issues from a year ago + we can use the meson build for this now.

Issue #36527, opened 2023-10-24
IntegerVectorsModPermutationGroup confuses groups whose domains are different


jukkakohonen:
### Steps To Reproduce

In SageMath version 9.0, running this:

```
A = PermutationGroup([(1,2)], domain=[1,2])
print("with domain ", A.domain())
print(IntegerVectorsModPermutationGroup(A, 3).list())

B = PermutationGroup([(1,2)], domain=[1,2,3])
print("with domain ", B.domain())
print(IntegerVectorsModPermutationGroup(B, 3).list())
```

### Expected Behavior

In `A` , the domain has two elements, so I am expecting lists of **two elements** that add up to 3:
```
with domain  {1, 2}
[[3, 0], [2, 1]]
```

In `B`, the domain has three elements, so I am expecting lists of **three elements** that add up to 3:
```
with domain  {1, 2, 3}
[[3, 0, 0], [2, 1, 0], [2, 0, 1], [1, 1, 1], [1, 0, 2], [0, 0, 3]]
```

### Actual Behavior

From both groups we get the same listing:
```
with domain  {1, 2}
[[3, 0], [2, 1]]
with domain  {1, 2, 3}
[[3, 0], [2, 1]]
```

### Additional Information

If the order of execution is reversed, so that we do `B` first and `A` second, then `B` gives the correct three-element lists, and `A` gives incorrectly the three-element lists.

So, the first one gets computed correctly, and the second one incorrectly. It looks like IntegerVectorsModPermutationGroup is remembering that it has already seen this group, not noticing that is in fact a different permutation group with different domain.

### Environment

```markdown
- **OS**: Ubuntu 20.04
- **Sage Version**: 9.0
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

TheShiveshNetwork:
I'd like to work on this issue. Can it be assigned to me?

jukkakohonen:
According to my rudimentary tests, for the second group, the `__init__` method of the auxiliary class `IntegerVectorsModPermutationGroup_with_constraints` is never even called. This seems to be because the auxiliary class inherits from `UniqueRepresentation`. When called with a group argument that it has seen before, it just takes the existing, cached instance.

And indeed, those two permutation groups *are* "Sage-equal" as in
```
sage: A = PermutationGroup([(1,2)], domain=[1,2])                               
[...]

PR #36814, opened 2023-12-04
Fast cardinality method for IntegerVectorsModPermutationGroup

### LINKED ISSUE #36681: IntegerVectorsModPermutationGroup crashes on empty domain
### Steps To Reproduce

In SageMath version 9.0, Release Date: 2020-01-01, run the following:

```
sage: P=PermutationGroup([],domain=[])                                          
sage: V=IntegerVectorsModPermutationGroup(P, sum=1)                             
sage: V.cardinality()   
```

### Expected Behavior

The cardinality should be zero.

The domain is empty, so we are counting empty integer vectors whose sum is 1. There are no such vectors.

### Actual Behavior

SageMath crashes with the following dump.

```
------------------------------------------------------------------------
/usr/lib/python3/dist-packages/cysignals/signals.cpython-38-x86_64-linux-gnu.so(+0x8154)[0x7ff63a15a154]
/usr/lib/python3/dist-packages/cysignals/signals.cpython-38-x86_64-linux-gnu.so(+0x8329)[0x7ff63a15a329]
/usr/lib/python3/dist-packages/cysignals/signals.cpython-38-x86_64-linux-gnu.so(+0xad5f)[0x7ff63a15cd5f]
/lib/x86_64-linux-gnu/libc.so.6(+0x43090)[0x7ff63d26a090]
/usr/lib/python3/dist-packages/sage/combinat/enumeration_mod_permgroup.cpython-38-x86_64-linux-gnu.so(+0x6ffb)[0x7ff51eb7effb]
/usr/lib/python3/dist-packages/sage/combinat/enumeration_mod_permgroup.cpython-38-x86_64-linux-gnu.so(+0x884b)[0x7ff51eb8084b]
/usr/lib/python3/dist-packages/sage/combinat/enumeration_mod_permgroup.cpython-38-x86_64-linux-gnu.so(+0xcb2d)[0x7ff51eb84b2d]
python3(PyCFunction_Call+0x59)[0x5f6939]
python3(_PyObject_MakeTpCall+0x296)[0x5f7506]
python3(_PyEval_EvalFrameDefault+0x59c7)[0x570787]
python3[0x50b07e]
python3(_PyEval_EvalFrameDefault+0x5796)[0x570556]
python3[0x500e63]
python3(_PyEval_EvalFrameDefault+0xafd)[0x56b8bd]
python3(_PyEval_EvalCodeWithName+0x26a)[0x5697da]
python3[0x50b1f0]
python3(_PyEval_EvalFrameDefault+0x5796)[0x570556]
python3(_PyEval_EvalCodeWithName+0x26a)[0x5697da]
python3(PyEval_EvalCode+0x27)[0x68e547]
python3[0x601624]
python3[0x5c4ef0]
python3(_PyEval_EvalFrameDefault+0x72d)[0x56b4ed]
python3[0x5009c8]
python3(_PyEval_EvalFrameDefault+0x213d)[0x56cefd]
python3[0x5009c8]
python3(_PyEval_EvalFrameDefault+0x213d)[0x56cefd]
python3[0x5009c8]
python3[0x504716]
python3(_PyEval_EvalFrameDefault+0x859)[0x56b619]
python3(_PyFunction_Vectorcall+0x1b6)[0x5f6ce6]
python3(_PyEval_EvalFrameDefault+0x72d)[0x56b4ed]
python3(_PyFunction_Vectorcall+0x1b6)[0x5f6ce6]
python3(_PyEval_EvalFrameDefault+0x859)[0x56b619]
python3(_PyEval_EvalCodeWithName+0x26a)[0x5697da]
python3[0x50b1f0]
python3(_PyEval_EvalFrameDefault+0x1910)[0x56c6d0]
python3(_PyEval_EvalCodeWithName+0x26a)[0x5697da]
python3(_PyFunction_Vectorcall+0x393)[0x5f6ec3]
python3(_PyEval_EvalFrameDefault+0x859)[0x56b619]
python3(_PyEval_EvalCodeWithName+0x26a)[0x5697da]
python3(_PyFunction_Vectorcall+0x393)[0x5f6ec3]
python3(_PyEval_EvalFrameDefault+0x859)[0x56b619]
python3(_PyFunction_Vectorcall+0x1b6)[0x5f6ce6]
python3(_PyEval_EvalFrameDefault+0x859)[0x56b619]
python3(_PyEval_EvalCodeWithName+0x26a)[0x5697da]
python3(PyEval_EvalCode+0x27)[0x68e547]
python3[0x67dbf1]
python3[0x67dc6f]
python3[0x67dd11]
python3(PyRun_SimpleFileExFlags+0x197)[0x67fe37]
python3(Py_RunMain+0x212)[0x6b7c82]
python3(Py_BytesMain+0x2d)[0x6b800d]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7ff63d24b083]
python3(_start+0x2e)[0x5fb85e]
------------------------------------------------------------------------

```

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Ubuntu 20.05
- **Sage Version**: 9.0, Release Date: 2020-01-01

Also tested on Sage 10.0, Release Date: 2023-05-20, with similar results.
```

### Checklist

### LINKED ISSUE #36787: Fast cardinality method for IntegerVectorsModPermutationGroup
### Problem Description

`IntegerVectorsModPermutationGroup` has no implementation of `cardinality`, so it defaults to iterating over the whole set and tallying. This is obviously slow for large sets.

For example, with vectors of length 10, summing to 20, and with the permutation group where the first two elements can be swapped, it takes more than two minutes to compute the cardinality.

### Proposed Solution

The cardinality should be calculated using the cycle index theorem (see e.g. [Peter Cameron's blog post](https://cameroncounts.wordpress.com/2018/03/16/a-counting-problem/) for an explanation).

An example code is here. This should be more or less a ready plug-in for the cardinality method.

```
def fast_cardinality(self):
    G = self._permgroup
    k = G.degree()              # Number of boxes
    d = self._sum               # Number of balls
    m = self._max_part          # Maximum balls in one box, -1 for no limit
    if m == -1:
        m = d
        
    if k==0:
        # Special case, no boxes.
        if d==0:
            return 1
        else:
            return 0

    # Power series with enough precision
    R.<x> = PowerSeriesRing(QQ, default_prec = d+1)

    # The function-counting series, for maximum part size m, is
    # (1-t^(m+1)) / (1-t) = 1+t+...+t^m.
    #
    # For the figure-counting series, we substitute x^cyclen for t.
    
    Z = G.cycle_index()
    funcount = sum(coeff * prod((1-x^((m+1)*cyclen)) / (1-x^cyclen)
                                for cyclen in cyctype)
                   for (cyctype, coeff) in Z)

    # Extract the x^d coefficient.
    dic = funcount.dict()
    if d in dic:
        return Integer(dic[d])  # Convert from Rational to Integer
    else:
        return 0
```

### Alternatives Considered

We could leave it as is, and rely on users to perform the cycle index calculations.

### Additional Information

Here is a speed comparison.

Permutation group acting on 10 elements, and only containing two permutations, the trivial one and the swap `(1,2)`. Count the vectors that sum to 20.

The times are about 0.1 seconds versus 161 seconds.

```
sage: G=PermutationGroup([(1,2)],domain=range(10)); V=IntegerVectorsModPermutationGroup(G,20)                                             
sage: t0=time(); fast_cardinality(V), time()-t0                                                                                           
(5911763, 0.09431600570678711)
sage: t0=time(); V.cardinality(), time()-t0                                                                                               
(5911763, 160.82350969314575)
sage: version()                                                                                                                           
'SageMath version 9.0, Release Date: 2020-01-01'
```

jukkakohonen:
# Fast cardinality method for IntegerVectorsModPermutationGroup

This patch fixes #36787 by implementing a `cardinality` method for
`IntegerVectorsModPermutationGroup_with_constraints`.  The method
calculates the cardinality using the the [Polya enumeration
theorem](https://en.wikipedia.org/wiki/P%C3%B3lya_enumeration_theorem) (also known as the cycle index theorem).

It is faster than the the default implementation, which iterates through the full set to
find the cardinality.

Incidentally this PR fixes #36681 so that cardinality and iter no longer crash in empty-domain situations.

### :memo: Checklist

### :hourglass: Dependencies

- #36873: we can use `is_trivial` from that PR


src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -775,6 +785,79 @@ def __iter__(self):
...
+            sage: G = PermutationGroup([(1,2,3)])
+            sage: I = IntegerVectorsModPermutationGroup(G, sum=10, max_part=5)
+            sage: I.cardinality()
+            7
+
+            sage: G = SymmetricGroup(10)
+            sage: I = IntegerVectorsModPermutationGroup(G, max_part=1)
+            sage: I.cardinality()
+            11
+        """
```
mantepse, line 949: There are a few more tests (in a `TESTS::` section) that would be good:
```
Check that :issue:`36681` is fixed::
sage: P=PermutationGroup([],domain=[])                                          
sage: V=IntegerVectorsModPermutationGroup(P, sum=1)                             
sage: V.cardinality()   
0
```
Add a group which is not the full symmetric group and not a cyclic group, or maybe better, all permutation groups of small order (can we generate them?  I think you mentioned that?)

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -775,6 +785,79 @@ def __iter__(self):
...
+        G = self._permgroup
+        k = G.degree()          # Vector length
+        d = self._sum           # Required total
+        m = self._max_part      # Max of one entry, -1 for no limit
+        if m == -1:
+            m = d               # Any entry cannot exceed total
+        if k == 0:
```
mantepse, line 958: alternatively: `if not k:`

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -775,6 +785,79 @@ def __iter__(self):
...
+            # Special case: Empty vectors cannot have nonzero sum.
+            if d == 0 or d is None:
```
mantepse, line 962: alternatively: `return Integer(bool(d))`  (maybe better not :-)

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -775,6 +785,79 @@ def __iter__(self):
...
+                return Integer(1)
+            else:
+                return Integer(0)
+
+        if d is None:
```
mantepse, line 986: I think it would be better to do the case `d is None` separately.  In this case we simply have
`sum(c*prod((m+1)**cycle_len for cycle_len in cycle_type) for cycle_type, c in f)`, no?

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -775,6 +785,79 @@ def __iter__(self):
...
+        # For function-counting series, we substitute x**cyclen for t.
+        #
+        # If there is no fixed sum requirement, we sum over the
+        # possible sums.
+        Z = G.cycle_index()
+        funcount = sum(
+            series_prod((1 - x**((m+1)*cyclen)) / (1 - x**cyclen)
+                        for cyclen in cyctype)
+            * coeff
+            for (cyctype, coeff) in Z)
```
mantepse: I suggest to do `for cycle_type, c in G.cycle_index())` (and do `cycle_len` and `cycle_type` instead of the other abbreviations.

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: Yes, some fancier groups would be nice in TESTS.  For example, it is nice to demonstrate and test that `PermutationGroup([(1,2),(3,4)])` (independent swaps, so a group of order 4) is indeed different from `PermutationGroup([[(1,2),(3,4)]])` (simultaneous swap, so a group of order 2).

For a "long" test, we could take all subgroups of `SymmetricGroup(k)` for k=3 (6 groups) and k=4 (30 groups). Perhaps a few direct products of small symmetric and cyclic groups.

For extensive tests we have to use the `sgs` parameter to `IntegerVectorsModPermutationGroup`, because otherwise the tests *will* fail for groups that only differ by the domain, due to #36527 which is still unfixed (this PR does not fix it and it is unclear to me how and when it would be fixed).

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: No, you have to multiply the power series. To wit,
```
sage: G=SymmetricGroup(2)
sage: Z=G.cycle_index()
sage: m=2
sage: V=IntegerVectorsModPermutationGroup(G, max_part=m)
sage: V.cardinality()
6
sage: sum(c*prod((m+1)**cycle_len for cycle_len in cycle_type) for cycle_type, c in Z)
9
sage: list(V)
[[0, 0], [1, 0], [2, 0], [1, 1], [2, 1], [2, 2]]
```
The correct count is 6 as we can see from the list.

mantepse: Sorry, I made a silly mistake.  I should have written `sum(c*(m+1)^len(cycle_type) for cycle_type, c in G.cycle_index())`.  Is that right?

jukkakohonen: Yes, that is correct. This is because the function-counting series is then just a `k*m`-degree polynomial, and the sum of its coefficients is the value of the polynomial at `x==1`. (There is probably a more direct combinatorial way to see this also.)  This adds a bit of complication/duplication to the code, but I guess it is worth it because it reduces to simple integer arithmetic, avoiding power series computations altogether in the `d is None` case.

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: Umm... I'd rather not. We are really talking about integers here, I think it is much more natural to compare to zero than to treat the integer as a pseudo Boolean. Is there an established rule for using ```not k``` pro ```k == 0``` when ```k``` is really and conceptually an integer value?

tscrim: It is faster and more common (at least within Sage's library) to do this. Most integers end up not being Python `int`'s as well, meaning coercion gets involved.

jukkakohonen: Faster my left foot. The comparison is done once (as in, "1 time") per invocation of `cardinality`. The difference is what, 40 nanoseconds? Did you measure?
```
sage: V=IntegerVectorsModPermutationGroup(SymmetricGroup(5),100)
sage: timeit("V.fast_cardinality()", number=1000, precision=5)
1000 loops, best of 3: 2.4201 ms per loop
sage: timeit("V.faster_cardinality()", number=1000, precision=5)
1000 loops, best of 3: 2.4255 ms per loop
sage: timeit("V.cardinality()", number=1, precision=5)
1 loop, best of 3: 10.837 s per loop
```
Here `cardinality` is the old iterative version. `fast_cardinality` is the version with `k==0` and `faster_cardinality` is exactly same code with `not k`. Perhaps the `not k` version is faster by a few nanoseconds, but I am not seeing it here. We are replacing an old 10-second algorithm with a new 2.4-millisecond algorithm, and are we now seriously talking about shaving nanoseconds?

Since when did SageMath programmers become interested in shaving nanoseconds anyway? Quoting 
https://doc.sagemath.org/html/en/faq/faq-contribute.html
> Fast code is valuable, but clean, readable code is important. In the mathematics community, inaccurate results are unacceptable.  ... Don Knuth observes that: “We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil.”

Replacing the clean, semantically transparent integer comparison `k==0` with a pseudo-Boolean comparison `not k` is, in my humble opinion, a prime example of a futile attempt at premature optimization that obscures the meaning and works as an open invitation to bugs. And it is against the FAQ that I quoted above.

I ask again: Is this `not k` style for integer comparison **endorsed in some official documentation**, or is it just a habit that SageMath programmers have grown fond of, for better or worse?

tscrim: For the record, yes, I have repeatedly measured. Not every coding convention (in fact, many of them are not) is in some official documentation. Not everything is called "1 time" such as the method being called in a tight loop. This is also not premature optimization but a standard micro-optimization; premature optimization is writing very complicated code for something that is not a bottleneck (as one has not profiled code). For example, another oddity of Python is `tuple([i for i in range(5)])` is faster than `tuple(i for i in range(5))`, so we often use the former even though the latter is more "clean" (in some sense of the word).

I hope you will continue to want to be an active and collaborative part of the SageMath developers community. Yet, I feel I must also point out our [code of conduct](https://github.com/sagemath/sage/blob/develop/CODE_OF_CONDUCT.md) and that your tone is close to not adhering to those standards.

jukkakohonen: Well, if that thing is a "standard micro-optimization" routinely applied in SageMath code, I must say I am sorry to hear that.

With all due respect I disagree that it is a good idea to apply it all over the place, or here. I have expressed the reasons for my stance and I will leave it at that.

I will *not* be applying that kind of thing here, but if someone else wants to, then that's their right.

mantepse: Dear @jukkakohonen!

I am very sorry that this got so heated, and I would like to ask you to stay.  I cannot speak for Travis, but I think what happened is a mis-emphasis.  Initially, that is, when I switched to sage and therefore python, I was also extremely annoyed by some (and in fact, the sheer number of) idioms.  (My background is common lisp and later, Aldor, which are - in my opinion - very clean languages.  Lisp doesn't even have much syntax.)

In the case at hand, I think it is mostly a matter of coding convention.  Some of these have been written down, some of these have just formed over time.  I don't know whether `not x` instead of `x == 0` is written down somewhere, but it is certainly standard, because in a tight loop you want to choose the faster version, and if I don't read `not x` over and over again, I might not realize that it is a way to check whether `x` is zero.  I remark that it is not applied always: grep finds about 6000 matches of ` == 0` in sage library code.  Of course, many of those are leftovers of past times.

In any case, I like your contribution a lot, and find it very helpful.  So please, let's get it done and let's focus on what we can agree on.

jukkakohonen: @mantepse Thank you for your considerations. Informed opinions, agreeable or otherwise, are always welcome.

Agreeing is nice, just I'm not sure about how much agreement is achievable. I *hope* we agree that replacing a default algorithm with a $4000 \times$ faster one, cleanly coded, is a good idea. I *doubt* you can make me agree that proliferating local optimizations of the kind discussed here is a good idea. That goes not only against well-respected authorities in algorithmics, but also against my own experience. Perhaps I may mention that I *have* done my good share of assembly coding, bit-twiddling and processor pipeline arrangement, enough that I know when it is a good idea and when it is not.

*If* the `not k` kind of optimization is deemed crucial and mandatory in SageMath, well, I am happy to report that I won't be bothered. I have plenty of other things to do in research. How mandatory it is, well, that is up to the community.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -775,6 +784,79 @@ def __iter__(self):
             else:
                 return SF.elements_of_depth_iterator(self._sum)
 
+    def cardinality(self):
+        r"""Return the number of integer vectors in the set.
```
mantepse: This should be
```
r"""
Return ...
```
(note the newline) per sage conventions

src/sage/combinat/integer_vectors_mod_permgroup.py
mantepse: So, could you simply reshape the pull request otherwise as discussed (separating the `d is None` case, perhaps rename the variables to make their meaning even more obvious, and perhaps adding a systematic test, and if not, at least one other non-cyclic, non-symmetric group and the test showing that #36681 is fixed)?

jukkakohonen: > So, could you simply reshape the pull request otherwise as discussed

Of course, because those other things make sense.

Re: variable names, I think my variable names and in-code comments are already more descriptive and less terse than what I have seen in other SageMath code (again I am very sorry to see that). But I am not one to be against code clarity, as you may have noticed.

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: I do not see much difference in clarity between `cyctype` and `cycle_type` in this context, and I deliberately chose the shorter one for the benefit of the reader, to better differentiate a local variable from the `cycle_type` method of `PermutationGroupElement`.

But it may be argued that a well-equipped reader is well aware of what is a local variable, so yes, we can use a longer nonabbreviated name. What I don't understand is why in your other code snippet you are then asking for the opposite: for the descriptive `coeff` to be replaced with the terse `c`?

mantepse: I don't mind much.  I read `cyclen` as a weird spelling of the plural of `cycle` (I'm Austrian, i.e. German speaking), and I was confused (even though I know my Pólya-Redfield quite well).  Since separating words with underscores is very common, I suggested it.  It is really as minor as it gets.

jukkakohonen: Oh, I did not see *that* way of reading :-)
I have absolutely no issues against using full words and underscore here, I will change it the way you like. But is it ok to keep `coeff`? You do not want it to be tersified into `c`?

mantepse: I don't mind.  Use what fits you.

jukkakohonen: Oh good. I was afraid that the single letter was another house rule directed against code clarity. (You know, the compiler may be 2 nanoseconds faster that way, and you never know when someone is going to call the compiler a billion times in a tight loop.)

mantepse: Please try to keep this a happy, fun and productive space.  Please resist making jokes about other opinions.

jukkakohonen: I'm not joking (though I must admit that recently I have sometimes had difficulty keeping a straight face). I have seen so much terse SageMath code with single-letter variable names and nary a comment, that it did cross my mind that it is a house rule.

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: Exactly, better not.

src/sage/combinat/integer_vectors_mod_permgroup.py
tscrim: I believe we can all agree that there are many (subjective) descriptions of what constitutes sufficient code clarity. Because of that, we can work together as a community without judgement to improve the overall utility of the software.

src/sage/combinat/integer_vectors_mod_permgroup.py
dimpase: I am +1000 on this contribution. I'll provide few tests for "fancy groups", if you don't mind.

dimpase:
Few examples counting 0-1 vectors, e.g.
```
sage: G=PGL(2,17)
sage: I = IntegerVectorsModPermutationGroup(G, sum=4, max_part=1)
sage: I.cardinality()
3
```
Indeed, PGL(2,q), is 3-transitive:

```
sage: I = IntegerVectorsModPermutationGroup(G, sum=3, max_part=1)
sage: I.cardinality()
1
```
but not 4-transitive. The stabiliser of a 3-set is $S_3$, which in this case has 4 orbits on the 18-set (so that's why 3 above is correct).

jukkakohonen:
I separated the `d is None` case as suggested, and added test cases (that the crash on empty domain is fixed & all permutation groups of order 4 compared to the length of the explicit listing). Fixed some formatting issues in code and comments.

For the benefit of those who care about actual speed, this is what the line profiler says: With `V = IntegerVectorsModPermutationGroup(SymmetricGroup(5), 100)`, running `V.cardinality()` 10000 times in a loop, an average loop took $3.849$ milliseconds, of which 288.9 nanoseconds (0.008 %) were spent on the `if k==0` line. With that line replaced with `if not k`, on a new run that line took 325.5 nanoseconds. Both tests were on the same, otherwise idle laptop after a minute of sleep to allow the processor to cool down. With effects this small, random fluctuations will have placed a large role, so more extensive tests would be required to reveal the full savings of `not k` in their entirety.

If the performance of `cardinality` is now a big problem, one might look towards the two lines `Z = G.cycle_index()` and `funcount = sum ...`, each taking roughly 50% of the time. There are ways of making *those* faster in this particular usage, with some cost in code complexity, but that has not been suggested here.

jukkakohonen:
Nice extra examples @dimpase ! Indeed 0-1 vectors are an important use case. (Graphs in general are 0-1 vectors, when viewed that way.) Do you mean these as "examples" or "tests"?

jukkakohonen:
In some special cases there are faster ways of finding the cardinality than the full-fledged Pólya counting. I'm wondering how many special cases would be sensible here. Empty domain is already handled specially, but:

1. Trivial group, with arbitrarily large domain, would be much faster simply by stars and bars (just like IntegerVectors does after PR #36830).
2. If the sum argument is very small (implying very small cardinality), it may be faster just to iterate and tally.

As an example for 1., consider the trivial group over 10 elements, and let the sum argument be 20. The general method takes roughly 600 microseconds, but calculating the cardinality by the binomial coefficient $\binom{29}{9} = 10015005$ is just 20 microseconds.

As an example for 2., consider the following setting where only 2 balls are distributed over 10 boxes, with some nontrivial symmetry.
```
sage: P = PermutationGroup([[(1,2)],[(3,4),(5,6,7)]], domain=range(1,11))
sage: V = IntegerVectorsModPermutationGroup(P, sum=2)
sage: timeit("sum(1 for v in V)")
625 loops, best of 3: 466 μs per loop
sage: timeit("V.cardinality()")
125 loops, best of 3: 2.88 ms per loop
sage: V.cardinality()
24
```
Even here Pólya counting is not *much* worse than iterating. With the same group and sum=4, cardinality becomes 207 and iterating is already slower than Pólya. So it would really be *very* small cardinalities that are useful to calculate by iteration.

I guess that 1. is a very common use case (if you construct permutation groups e.g. from the automorphisms of some big graphs, it is quite common that the group is in fact trivial), and it is easy to recognize and handle it as a special case (so that the user does not have to).




Furthermore one *could* try to recognize some special cases of groups (like cyclic or symmetric) and invoke special code, but I somehow think it would be an overkill here. If some particular use case involves lots of groups of some specific structure, allowing perhaps faster counting, then probably the *user* is a better position to know about the structure and to exploit it if needed.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,149 @@ def __iter__(self):
...
+        k = G.degree()          # Vector length
+        d = self._sum           # Required sum
+        m = self._max_part      # Max of one entry, -1 for no limit
+        if m == -1:
+            m = d               # Any entry cannot exceed total
+        if k == 0:
+            # Special case: Empty vectors cannot have nonzero sum.
+            if d == 0 or d is None:
+                return Integer(1)
+            else:
```
mantepse, line 964: `else` is unnecessary here.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,149 @@ def __iter__(self):
...
+            []
+
+
         """
+        # Special cases when domain is empty.
+        if self.n == 0:
+            if self._sum is not None and self._sum > 0:
+                # No empty vector can have positive sum.
+                return iter(())
+            else:
```
mantepse, line 842: `else` is unnecessary here.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,149 @@ def __iter__(self):
...
-                              algorithm='breadth')
+            SF = RecursivelyEnumeratedSet_forest(
+                (self([0]*(self.n), check=False),),
+                lambda x: [self(y, check=False)
+                           for y in canonical_children(
+                                   self._sgs, x, self._max_part)],
+                algorithm='breadth')
             if self._sum is None:
                 return iter(SF)
             else:
```
mantepse, line 859: `else` is unnecessary here.

mantepse:
It seems to me that all of this is premature optimization.  I would, however, agree that making a special case for the trivial group makes sense.

mantepse:
maybe `is_trivial` should really be a method of `PermutationGroup_generic`.  This method exists already for `AbelianGroup_gap`.

jukkakohonen:
I have pushed a version that adds three special cases, and tests for them.

* Case 1. All-zero vectors (cardinality is 1).
* Case 2. Vectors with sum 1 (cardinality is number of orbits).
* Case 3. Trivial groups (compute by binomial coefficient).

A brief assessment of whether they are worth it:

* Each case takes two lines of code. One fast and easy condition, and one simple calculation.
* Each case occurs often in real use (from what I've seen).  
* Each case can speed up the cardinality call by more than tenfold (see below).

Here are some simple test results.  ``S`` is the symmetric group over 10 elements, and ``T`` is the trivial group over 10 elements. All times are averages over 1000 calls in similar conditions.

* Case 1: ``S`` with sum zero: from 15.4 ms to 17.0 µs (900 times faster).
* Case 2: ``S`` with sum one: from 15.5 ms to 18.3 µs (850 times faster).
* Case 3: ``T`` with sum 100: from 745.3 µs to 31.9 µs (23 times faster).

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,187 @@ def __iter__(self):
...
                 return SF.elements_of_depth_iterator(self._sum)
 
+    def has_trivial_group(self):
```
mantepse: As I mentioned before, this should be a method of `PermutationGroup`, and it's name should be `is_trivial`.

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: Not sure what you aim to achieve by these "else-unnecessary" comments. You probably know I am not going to change the code to something I consider inferior, in terms of style, clarity and reliability.

src/sage/combinat/integer_vectors_mod_permgroup.py
jukkakohonen: Sure it should. If it suddenly appears there, I will use it here.

src/sage/combinat/integer_vectors_mod_permgroup.py
mantepse: At the time I made the comment, I did not know that you consider it as inferior. As I said, I don't mind much.

src/sage/combinat/integer_vectors_mod_permgroup.py
mantepse: see #36873

jukkakohonen:
> Few examples counting 0-1 vectors, e.g.

Thanks, I added the PGL example.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -931,8 +936,9 @@ def cardinality(self):
             # The 1 can be placed in any orbit, and by symmetry
             # it will be on the first element of the orbit.
             return Integer(len(G.orbits()))
-        if d is not None and self.has_trivial_group() and m >= d:
-            # Simple calculation with stars and bars.
+        if d is not None and m >= d and all(g.is_one() for g in G.gens()):
+            # Trivial group (should use G.is_trivial() as soon as
```
mantepse: You can and should make the pr for ·is_trivial· a dependency, then we don't need to change things again.

jukkakohonen: Right, but for that I need to find out how to make a dependency between PR's (never did that before). The current code works fine as it is, regardless of the `is_trivial` pr, so I'm not convinced it pays off to complicate things by creating dependencies.

mantepse: There is a heading "Dependencies" at the very top of this PR, and there you simply write `#36873`.  Then, locally, you checkout the other PR (eg., `gh pr checkout 36873`, switch back to your PR, do `git merge perm_gps/is_trivial`, and push the result.

jukkakohonen: Thanks, I hope I did it right.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,192 @@ def __iter__(self):
             [3, 1, 1, 2]
             [3, 0, 2, 2]
             [2, 2, 2, 1]
+
+            Check that :issue:`36681` is fixed::
+            sage: I = IntegerVectorsModPermutationGroup(
```
mantepse: Indentation is wrong here: `Check ...` needs to go 4 spaces to the left.  Also, add a new blank line after `::`

Finally, for some editors (eg., mine), the following style is preferrable::
```
sage: G = PermutationGroup([], domain=[])
sage: I = IntegerVectorsModPermutationGroup(G, sum=0)
```
(i.e., avoid `....:`)

(same thing below)

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,192 @@ def __iter__(self):
...
+            ....:     PermutationGroup([], domain=[]), sum=0)
+            sage: list(iter(I))     # Single empty vector
```
mantepse: Personally, I'd remove the two comments, here and two lines below, because they say exactly what the output is saying anyway.

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,192 @@ def __iter__(self):
...
+    def cardinality(self):
+        r"""
+        Return the number of integer vectors in the set.
+
+        The number is computed using the cycle index theorem, which is
+        faster than listing the vectors.
+
+        EXAMPLES::
+
+            With a trivial group all vectors are canonical::
```
mantepse: As above, the indentation needs to be fixed.  Also, in this case, `EXAMPLES::` should be `EXAMPLES:` (because there will be no indent after).  [this is sphinx, not me :-)]

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +788,192 @@ def __iter__(self):
...
+            10
+            sage: IntegerVectorsModPermutationGroup(T10, 100).cardinality()
+            4263421511271
+
+        """
+        from sage.rings.power_series_ring import PowerSeriesRing
+        from sage.rings.rational_field import QQ
+        from sage.rings.integer import Integer
+        from sage.misc.misc_c import prod
+        from sage.arith.misc import binomial
```
mantepse: I don't care much, but usually the imports are at the top of the file: https://peps.python.org/pep-0008/#imports

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +794,195 @@ def __iter__(self):
...
+        With a trivial group all vectors are canonical::
+
+            sage: G = PermutationGroup([], domain=[1,2,3])
+            sage: IntegerVectorsModPermutationGroup(G, 5).cardinality()
+            21
+            sage: IntegerVectors(5, 3).cardinality()
+            21
+
+        With two interchangeable elements, the smaller one
+        ranges from zero to n//2::
```
mantepse: `n//2` needs double backticks (for the html doc to come out correctly)

src/sage/combinat/integer_vectors_mod_permgroup.py
```diff
@@ -763,18 +794,195 @@ def __iter__(self):
...
                 return iter(SF)
             else:
                 return SF.elements_of_depth_iterator(self._sum)
 
+    def cardinality(self):
+        r"""
+        Return the number of integer vectors in the set.
+
+        The number is computed using the cycle index theorem, which is
+        faster than listing the vectors.
```
mantepse: I'd actually remove `, which is faster than listing the vectors`, because it carries no useful information, at least not for the user.

dimpase: I think it can be left in - perhaps moved into ` ..NOTE` . It does carry useful info (if you only care about the number, don't try to list vectors)

mantepse: I don't really mind, but I don't see the point either - it is obvious that you wouldn't implement a formula if enumeration is faster, isn't it?

jukkakohonen: I don't understand why a user would *not* benefit from being told what is fast and what is not. Certainly that is more relevant from a user's viewpoint than which particular algorithm is being used. So perhaps we remove the cycle index theorem then.

mantepse: I disagree, but let's leave it at that, then.  Let's not argue about seven words.

jukkakohonen: I will remove the information as requested.

jukkakohonen: For some reason the default `rank` and `unrank` still talk about speed.

```
   This is the default (brute force) implementation from the category
   "EnumeratedSets()" which can be used when the method "__iter__" is
   provided. Its complexity is O(r), where r is the rank of "obj".
```

mantepse: Let me just add my personal reason why I find "uses the Redfield–Pólya theorem" valuable information and "faster than plain enumeration" not.

Before that, please note that I don't really mind, because I know the code and the math behind it quite well, so it doesn't affect me in this case, and I see that one can have other opinions.  In particular, you can leave it as it is, or modify it, as you wish.

The first gives me some rough information on how the result is obtained.  This can be helpful if I have to understand the code (eg., whether I can adapt it to other, similar classes) and also what I can expect from it performance-wise.  For example, if the doc string says: "We essentially use linear search to find the element of given rank.", this tells me immediately that, if I know a custom unranking algorithm, I will be able to improve the performance.  Also, the code might use some ad-hoc tricks to make it slightly faster, which will almost certainly make it harder to decipher what the underlying algorithm really is.

I don't see any information I could gain from "faster than plain enumeration".  If it would be slower than plain enumeration, I wouldn't use a (nontrivial) formula, would I?

dimpase: > I will remove the information as requested.

No, please don't do this. It is absolutely normal, and preferred, for Sage documentation to explain what mathematical tools are behind the implementation. 
If a user wants to understand how a piece of code works, it's not a good idea to leave it totally unexplained, just like a blackbox. Sage has users from all walks of mathematical and CS life, I don't know what percentage of them heard about the cycle index theorem, perhaps 5% or less. It's not a topic that comes up in every maths-related undergraduate curriculum. 

I'd even like to ask for a reference to the cycle index theorem to be added to the docstring.
It could be just 
```
:wikipedia:`Cycle_index`
```

dimpase: > I don't see any information I could gain from "faster than plain enumeration". If it would be slower than plain enumeration, I wouldn't use a (nontrivial) formula, would I?

Just imagine you didn't know what the cycle index theorem is all about. The choice of the algorithm might have been like "we chose an algorithm we knew, and implemented, just because we could". Only after an explanation that it's a huge speedup it'd made sense.

mantepse:
There is a linter failure, other than that looks good to me.

mantepse:
Thank you for the extra polish!

PR #36829, opened 2023-12-07
CI Linux incremental: `make reconfigure` after uninstalling packages

mkoeppe:
As observed in https://github.com/sagemath/sage/pull/36792#issuecomment-1841713301, the "CI Linux incremental" workflow, which is run when package scripts are updated, is not helpful when testing new or modified `spkg-configure.m4` scripts when the SPKG has been installed:

Although we uninstall packages with changed spkg-configure.m4 scripts already, a re-run of configure is not (always?) triggered.

So here we invoke it explicitly, using a new makefile target `make reconfigure`. 
(This new target also simplifies instructions issued by the system package advice facility.)

### :memo: Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,9 @@ SPKG_COLLECT_FILES = build/pkgs/*/type build/pkgs/*/package-version.txt build/pk
 # If configure was run before, rerun it with the old arguments.
 # Otherwise, run configure with argument $PREREQ_OPTIONS.
 build/make/Makefile: configure $(SPKG_COLLECT_FILES) $(CONFIG_FILES:%=%.in)
+	$(MAKE) reconfigure
+
+reconfigure:
 	rm -f config.log
 	mkdir -p logs/pkgs
 	ln -s logs/pkgs/config.log config.log
@@ -62,6 +65,8 @@ build/make/Makefile: configure $(SPKG_COLLECT_FILES) $(CONFIG_FILES:%=%.in)
 		echo >&2 '****************************************************************************'; \
 	        exit 1; \
 	fi
+
+
 
 # This is used to monitor progress towards Python 3 and prevent
 # regressions. Should be removed after the full switch to python3.
@@ -339,7 +344,7 @@ list:
 	@$(MAKE) --silent build/make/Makefile >&2
 	@$(MAKE) --silent -f build/make/Makefile SAGE_PKGCONFIG=dummy $@
 
-.PHONY: default build dist install micro_release \
+.PHONY: default build dist install micro_release reconfigure \
 	misc-clean bdist-clean distclean bootstrap-clean maintainer-clean \
 	test check testoptional testall testlong testoptionallong testallong \
 	ptest ptestoptional ptestall ptestlong ptestoptionallong ptestallong \
--- a/Makefile
+++ b/Makefile
@@ -66,8 +66,6 @@ reconfigure:
 	        exit 1; \
 	fi
 
-
-
 # This is used to monitor progress towards Python 3 and prevent
 # regressions. Should be removed after the full switch to python3.
 #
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -63,6 +66,8 @@ build/make/Makefile: configure $(SPKG_COLLECT_FILES) $(CONFIG_FILES:%=%.in)
 	        exit 1; \
 	fi
 
+
+
 # This is used to monitor progress towards Python 3 and prevent
```
kwankyu: Really three blank lines?

kwankyu:
Otherwise lgtm.

Makefile
mkoeppe: fixed in ce89bf6

kwankyu (approved):
Thanks.

mkoeppe:
Thank you!

Makefile
```diff
@@ -50,6 +50,9 @@ SPKG_COLLECT_FILES = build/pkgs/*/type build/pkgs/*/package-version.txt build/pk
 # If configure was run before, rerun it with the old arguments.
 # Otherwise, run configure with argument $PREREQ_OPTIONS.
 build/make/Makefile: configure $(SPKG_COLLECT_FILES) $(CONFIG_FILES:%=%.in)
+	$(MAKE) reconfigure
+
+reconfigure:
 	rm -f config.log
 	mkdir -p logs/pkgs
 	ln -s logs/pkgs/config.log config.log
```
kwankyu, line 58: You didn't touch lines 56--58, but now `make reconfigure` runs these lines repeatedly. I wonder if this is really necessary. What is the purpose of these lines (in particular, the symbolic link)? Why do we keep the log file in `logs/pkgs/config.log` instead of at the sage root directory?

kwankyu: I experimented with it. It seems that removing the three lines does not do any harm, and we can keep the log in the root directory...

mkoeppe: I don't understand what you mean. There was no `make reconfigure` target before this PR. 
The rule `build/make/Makefile` used to run lines 56–58 directly.
Now it runs them by delegating to `make reconfigure`

kwankyu: The lines are not run by the first run of `./configure`. Hence we have the file `config.log` at the root directory. The old message from `./configure` says 
```
    hint: After installation, re-run configure using:

      $ ./config.status --recheck && ./config.status
```
Hence running the above code does not create the symbolic link to the log file in `logs/pkgs/config.log`.

I meant that with your PR, the lines are run by `make reconfigure`, as recommended by the new message
```
    hint: After installation, re-run configure using:

      $ make reconfigure
``` 
and we have the symbolic link.

mkoeppe: Oh, now I understand.

mkoeppe: At least I think so. You mean that if the user just runs `./config.status --recheck && ./config.status` (as previously instructed by `configure`), the linking is not done, and now with the replacement `make reconfigure` it is done?

kwankyu: Yes. And if I run `make reconfigure` again, it recreates the symbolic link.

kwankyu: Above all, I see no reason to keep `config.log` in `logs/pkgs`.

mkoeppe: Neither do I, but it has been like that since at least 2015 (0e168b13f36e0b6af6de5d521fa18f9054cb1add), and I don't want to change it here.

mkoeppe: In #33262, I had an improvement for the handling of config.log though, please take a look

kwankyu: Ah, I now see that the symlink is created anyway by the target `build/make/Makefile`, and we end up with the symlink at the end of installing sage, though not immediately after `./configure`. Hence there is no change in this regard with this PR. OK then.

I still think the directory `logs/pkgs` is not the right place for `config.log` though. Anyway this is not a concern of this PR.

OK. Thanks.

mkoeppe: > In #33262, I had an improvement for the handling of config.log though, please take a look

That's now revived as #36998

PR #36830, opened 2023-12-07
improved integer vectors efficiency -Enhancement

### LINKED ISSUE #36816: IntegerVectors efficiency and corner cases
### Problem Description

A few things, so closely interrelated that it makes sense to collect them together:

1. `IntegerVectors_nk` has no cardinality method, so it defaults to the slow method of iterating: `IntegerVectors(200,5).cardinality()` takes something like a minute to compute `70058751`.
2. `IntegerVectors_nk` has no unrank method: similarly `IntegerVectors(200,5).unrank(70000000)` takes a minute.
3. `IntegerVectors_n` and `IntegerVectors_k` also have no unrank, making it slow.
5. `IntegerVectors_n` and `IntegerVectors_k` always claim to be `InfiniteEnumeratedSets` with cardinality `+Infinity`. More appropriately `IntegerVectors_n(0)` and `IntegerVectors_k(0)` should be finite sets of cardinality `1`, because they contain only the empty vector.
6. `IntegerVectors(k=0)[1]` loops forever, while it should just raise `IndexError` because the set has only one element.

### Proposed Solution

1. Compute cardinality by [stars and bars](https://en.wikipedia.org/wiki/Stars_and_bars_(combinatorics)). Then `IntegerVectors(200,5).cardinality()` will take something like a millisecond.
2. Use a fast method based on 1.
3. Similar to the previous, but slightly more complicated. Still a fast `unrank` should be doable.
4. Both cases should give a `FiniteEnumeratedSets` with cardinality 1, if the parameter is zero.
5. This bug would be fixed by the solution to 3.

### Alternatives Considered

For 5., a simpler fix (without doing 3. fully) would be just to check for the corner case.

### Additional Information

1., 2. and 3. are efficiency issues, but 4. and 5. are in fact bugs.

`IntegerVectors` has also other possible parameters, like `max_slope` and `min_slope`. This issue is *not* meant to change functionality in those cases (it would be more complicated), but only the base case with parameters `n` and/or `k`.

Related to item 1., a fast `rank` was already implemented in #15609.

amanmoon:
Fixes #36816

1)  Added the `cardinality function` using stars and bars to the `IntegerVector_nk` class 
2) Fixed the bug in IntegerVectors_n and IntegerVectors_k now **IntegerVectors_n(0)cardinality() and IntegerVectors_k(0).cardinality() returns 1 and both of them are Finite EnumeratedSets**
3) improved efficiency of `rank` and `unrank` methods in `IntegerVectors_n` and `IntegerVectors_k`
4) `IntegerVectors_n(0)` and `IntegerVectors_k(0)` should now show cardinality as 1


src/sage/combinat/integer_vector.py
```diff
@@ -839,7 +838,10 @@ def __init__(self, n):
             sage: TestSuite(IV).run()
         """
         self.n = n
-        IntegerVectors.__init__(self, category=InfiniteEnumeratedSets())
+        if self.n==0:
```
mantepse: There need to be spaces around `==`.  Alternatively (and likely more according to the unwritten sage conventions, but I don't care):
```
if self.n:
    IntegerVectors.__init__(self, category=InfiniteEnumeratedSets())
else:
    IntegerVectors.__init__(self, category=EnumeratedSets())
```
This comment also applies below.

src/sage/combinat/integer_vector.py
```diff
@@ -898,7 +900,22 @@ def __contains__(self, x):
...
+        Return the cardinality of ``self``.
+        
+        TESTS::
+        
+            sage: IntegerVectors(n=0).cardinality()
+            1
+            sage: IntegerVectors(n=10).cardinality()
+            +Infinity
+        """
+        if self.n==0:
```
mantepse, line 914: same as for comment above

src/sage/combinat/integer_vector.py
```diff
@@ -1163,7 +1198,18 @@ def rank(self, x):
...
         return r
-
+            
+    def cardinality(self):
+        """
+        Return the cardinality of ``self``.
+        
+        TESTS::
+        
+            sage: IntegerVectors(200,5).cardinality()
```
mantepse: This is not really a test, but an example (demonstrating that cardinality works for large numbers).  It would also be good if there is an example that can be checked by hand, or even better, in your head.

A test would be `len(list(IntegerVectors(10, 5))) == IntegerVectors(10, 5).cardinality()`.

Warning: `len` automatically calls `cardinality`, if available.

src/sage/combinat/integer_vector.py
```diff
@@ -1315,7 +1361,7 @@ def __init__(self, n=None, k=None, **constraints):
             del constraints['inner']
         self.constraints = constraints
 
-        if n is not None:
+        if n is not None :
```
mantepse: No space before `:`.

src/sage/combinat/integer_vector.py
```diff
@@ -1438,8 +1484,7 @@ def cardinality(self):
         if self.k is None:
             if self.n is None:
                 return PlusInfinity()
-            if ('max_length' not in self.constraints
-                    and self.constraints.get('min_part', 0) <= 0):
+            elif 'max_length' not in self.constraints and self.constraints.get('min_part', 0) <= 0:
```
mantepse: The original version is much more readable, please revert.

src/sage/combinat/integer_vector.py
jukkakohonen: For an example&test that is checkable by hand (for us mortals) or in your head (for Gauss), you could have
```
sage: IntegerVectors(99, 3).cardinality()
5050
```
The "check in your head" would be that if the sum of the last two elements is $x$, then there are $x+1$ ways of dividing $x$ between the two elements, namely, $(0, x), (1, x-1), (2, x-2), \ldots, (x,0)$. Since $x$ can range from $0$ to $n=99$, the total cardinality is $1+2+\ldots+100 = 50 \times (1+100) = 5050$. Readers familiar with the [Gauss story](https://www.americanscientist.org/article/gausss-day-of-reckoning) will be amused.

You can then make a bigger demonstration with
```
sage: IntegerVectors(10^9 - 1, 3).cardinality()
500000000500000000
```

src/sage/combinat/integer_vector.py
```diff
@@ -900,6 +900,63 @@ def __contains__(self, x):
...
+
+            sage: IntegerVectors(k=5).rank([0,0,1,2,2])
+            243
+        """
+        if sum(x)!=self.n:
+            raise ValueError("argument is not a member of IntegerVectors({},{})".format(self.n, None))
+            
+        n, k, r = self.n, len(x), 0
+        for i in range(k):
+            r += binomial(i+n-1, n)
```
mantepse:
```
r = sum(binomial(n+i-1, i-1) for i in range(1, k))
```
should be easier to understand.

src/sage/combinat/integer_vector.py
```diff
@@ -900,6 +900,63 @@ def __contains__(self, x):
...
+            
+        for i in range(k - 1):
+            k -= 1
+            n -= x[i]
+            r += binomial(k + n - 1, k)
+
```
mantepse, line 958: Maybe
```
for i in range(k-1):
    n -= x[i]
    r += binomial(n + k - i - 2, n - 1)
```
is easier to understand.  Possibly even better if you reverse the summation.  Please check.

adrinospy:
I also implemented the unrank() method using an algorithm similar to binary search
I also added a faster rank() method for `IntegerVector_n` and `IntegerVector_k` classes

src/sage/combinat/integer_vector.py
jukkakohonen: While we are on the topic of efficiency, it can be useful to notice that
```sum(binomial(m, k) for m in range(n)) == binomial(n, k+1)```
see Knuth *The art of computer programming* §1.2.6E or [Wikipedia](https://en.wikipedia.org/wiki/Binomial_coefficient#Sums_of_the_binomial_coefficients).

The right hand side can be slightly faster (try $n=10^7$ and $k=17$).

File: src/sage/combinat/integer_vector.py [hunk 8]
```diff
@@ -898,6 +900,85 @@ def __contains__(self, x):
...
+            sage: IntegerVectors(n=5).rank([5,0])
+            1
+            sage: IntegerVectors(n=5).rank([3,2])
+            3
+            
+        """
+        if sum(x) != self.n:
+            raise ValueError("argument is not a member of IntegerVectors({},{})".format(self.n, None))
+            
+        n, k, r = self.n, len(x), 0
```
mantepse: The assignment to `r` is now unnecessary.

src/sage/combinat/integer_vector.py
```diff
@@ -898,6 +900,85 @@ def __contains__(self, x):
...
+
+        r = sum(binomial(n + i - 1, i - 1) for i in range(1, k))
```
mantepse: Sorry for not realizing earlier, but this can be simplified further to `r = binomial(k+n-1, n+1)`, if I'm not mistaken.

src/sage/combinat/integer_vector.py
```diff
@@ -1438,7 +1643,7 @@ def cardinality(self):
         if self.k is None:
             if self.n is None:
                 return PlusInfinity()
-            if ('max_length' not in self.constraints
+            elif ('max_length' not in self.constraints
```
mantepse: no `elif` needed here, because there is a `return` just before.

mantepse:
(see https://doc.sagemath.org/html/en/developer/tools.html)

adrinospy:
Thanks, and sorry for not knowing that,
I have removed all the linting errors from my code, there were some errors when i ran `./sage -tox -e pycodestyle -- src/sage/combinat/integer_vector.py`: 
386:23: E741 ambiguous variable name 'l'
412:15: E741 ambiguous variable name 'l'
1813:1: E402 module level import not at top of file

I was not sure whether to change these, as they were present from the start, so I left them as they were.

src/sage/combinat/integer_vector.py
```diff
@@ -898,6 +901,83 @@ def __contains__(self, x):
...
+            sage: IntegerVectors(n=10).unrank(10)
+            [1, 9]
+        """
+        ptr = 0
+        rtn = [self.n]
+        while self.rank(rtn) < x:
+            rtn.append(0)
+        rtn.pop()
+
+        while True:
```
mantepse: maybe this while loop can be factored out into a helper function in `IntegerVectors`?

In any case, we should store `current_rank = self.rank(rtn)`, so we don't compute it twice.

mantepse:
Excellent, thank you!

There are now only a few comments left, I'll do the review once they are done.

src/sage/combinat/integer_vector.py
amanmoon: should i do the same for `rank()` method?

mantepse: I think no. If I see it correctly, it would only be 3 lines there.

src/sage/combinat/integer_vector.py
```diff
@@ -1160,10 +1323,58 @@ def rank(self, x):
...
+        - ``x`` -- an integer such that ``x < len(self)``
+
+        EXAMPLES::
+
+            sage: IntegerVectors(4,5).unrank(30)
+            [1, 0, 1, 0, 2]
+            sage: IntegerVectors(2,3).unrank(5)
+            [0, 0, 2]
+        """
+        if x >= len(self):
```
jukkakohonen: Using `len(self)` seems problematic. For me
```
sage: V=IntegerVectors(100000,6); c=V.cardinality(); c; V[c-1]
```
produces
```
83345834041685416895001
...
OverflowError: cannot fit 'int' into an index-sized integer
```

How about using `cardinality` instead?

jukkakohonen:
This is already a great improvement over the previous (default) cardinality and unrank methods, and worth including in the next version.

If the vector sum is large, I believe further improvements in unranking are possible. In the version currently pushed, `unrank(self, x, rtn)` is doing a linear search over the values for each position in the vector. The following takes half a minute to find the vector at position 3 million in the list of 10 million (plus one). I believe it is obvious that there are faster methods.

```
sage: V = IntegerVectors(10^7, 2)
sage: V.cardinality()
10000001
sage: %time V.unrank(3*10^6)
CPU times: user 38 s, sys: 0 ns, total: 38 s
Wall time: 38 s
[7000000, 3000000]
```

However, I reiterate that the current version is already a big improvement.

adrinospy:
@jukkakohonen  Thanks a lot, I'll keep working on finding a new and improved algorithm.

src/sage/combinat/integer_vector.py
```diff
@@ -968,6 +1055,78 @@ def __contains__(self, x):
...
+            0
+            sage: IntegerVectors(k=5).rank([1,1,0,0,0])
+            7
+        """
+        if len(x) != self.k:
+            raise ValueError("argument is not a member of IntegerVectors({},{})".format(None, self.k))
+
+        n, k = sum(x), self.k
+
+        r = sum(binomial(k + i - 1, k - 1) for i in range(n))
```
jukkakohonen: It may be useful to notice that
` sum(binomial(k + i - 1, k - 1) for i in range(n)) == binomial(n+k-1, k)`
avoiding summation (please check yourself before using).

Currently it seems that `IntegerVectors(k=1)[n]` takes time roughly quadratic to `n`, which is *worse* than the old linear version.
```
sage: V=IntegerVectors(k=1)
....: %time V[1000]
CPU times: user 791 ms, sys: 0 ns, total: 791 ms
Wall time: 791 ms
[1000]
sage: V=IntegerVectors(k=1)
....: %time V[2000]
CPU times: user 3.13 s, sys: 0 ns, total: 3.13 s
Wall time: 3.13 s
[2000]
sage: V=IntegerVectors(k=1)
....: %time V[4000]
CPU times: user 10.9 s, sys: 0 ns, total: 10.9 s
Wall time: 10.9 s
[4000]
```

How does it fare if you replace the summation with a single binomial?

jukkakohonen:
Seeing that the main point of this PR is efficiency (besides the corner case fixes of course), it would be worthwhile to perform some offline performance tests to ensure that at least the methods are not getting _slower_ than they used to be before the PR.

Seeing that the old version of `IntegerVectors(k=1)[5000]` simply iterated through the vectors `[0], [1], ..., [5000]` in 6.2 milliseconds (just measured that in Sage 9.0), and the proposed new method took 17.3 *seconds* to find the same result (before the latest fix that eliminated a summation of binomial coefficients), I believe such offline tests would be in order.

src/sage/combinat/integer_vector.py
```diff
@@ -780,6 +780,20 @@ def __contains__(self, x):
                 return False
         return True
 
+    def unrank(self, x, rtn):
```
mantepse: This method needs a docstring and an example.  Also, currently it is not an unrank method, because it takes a second argument, so it should be renamed.  I didn't think about it, but could it be made into a proper unrank method by providing a default argument for `rtn`?

src/sage/combinat/integer_vector.py
```diff
@@ -968,6 +1055,78 @@ def __contains__(self, x):
...
+        EXAMPLES::
+
+            sage: IntegerVectors(k=5).unrank(10)
+            [1, 0, 0, 0, 1]
+            sage: IntegerVectors(k=5).unrank(15)
+            [0, 0, 2, 0, 0]
+        """
+        if self.k == 0 and x != 0:
+            raise IndexError(f"Index {x} is out of range for the IntegerVector.")
+        else:
```
mantepse: `else:` is unnecessary here, because `if` can only raise an error.  Similarly below.

src/sage/combinat/integer_vector.py
mantepse: Here is the version with the summation reversed.  Don't know whether it is better.
```
+        r = binomial(k + n - 1, n + 1)
+        s = 0
+        for i in range(k - 1):
+            s += x[k - 1 - i]
+            r += binomial(s + i, i + 1)
```

src/sage/combinat/integer_vector.py
amanmoon: The value of `rtn` depends on `n` and `k`, and there are various ways the `rtn` list is created based on any one of the values of `n` and `k` being None. I believe providing a default argument is not possible. 
What I can do is add an if-else block inside the unrank function that checks whether n or k is None and constructs the rtn based on that. 
Alternatively, I can rename the function to something like `construct_element_from_rtn` and I can write a general doctest for the `unrank` method.

mantepse: I'm afraid I misled you, I am sorry.  Possibly this shouldn't be a method of `IntegerVectors` at all, but rather an auxiliary function of the module.  It is not completely clear to me what it does - do I understand correctly that it first finds a vector `rtn` that has rank at least equal to `x` by moving the content of `rtn[0]` to larger indices, and then iterating one by one through the vectors having smaller rank, until we find the vector with the correct rank?

If so (or something similar), this won't work, for example, for `IntegerVectorsConstraints`, because `rtn` is not guaranteed to satisfy the given conditions, right?

adrinospy: You got the algorithm right.

amanmoon: You, got the algorithm right and the IntegerVectorsConstraints won't work as intended.
changing the name of the function will work or should i revert the changes back when there was no helper function?

mantepse: I d make it a normal Funktion and call it ·_unrank_helper·, take self as an extra argument.

src/sage/combinat/integer_vector.py
```diff
@@ -855,8 +855,7 @@ def __init__(self, n):
         self.n = n
         if self.n == 0:
             IntegerVectors.__init__(self, category=EnumeratedSets())
-        else:
```
mantepse: This `else:` *is* necessary, there is no `return` or `raise` before.

src/sage/combinat/integer_vector.py
```diff
@@ -998,8 +996,7 @@ def __init__(self, k):
         self.k = k
         if self.k == 0:
             IntegerVectors.__init__(self, category=EnumeratedSets())
-        else:
```
mantepse: This `else:` is necessary!

mantepse:
> Since the main point of this PR is efficiency (besides the corner case fixes of course), it would be worthwhile to perform some offline performance tests to ensure that at least the methods are not getting _slower_ than they used to be before the PR.
> 
> Seeing that the old version of `IntegerVectors(k=1)[5000]` simply iterated through the vectors `[0], [1], ..., [5000]` in 6.2 milliseconds (just measured that in Sage 9.0), and the proposed new method took 17.3 _seconds_ to find the same result (before the latest fix that eliminated a summation of binomial coefficients), I believe such offline tests would be in order.

One way this can be done (but I'm not sure that all sagemath developers agree with me) is to have a doctest which takes a tolerable amount of time (e.g., half a second on a standard computer) with the good version, but will take very long if the performance degrades.

jukkakohonen:
> One way this can be done (but I'm not sure that all sagemath developers agree with me) is to have a doctest which takes a tolerable amount of time (e.g., half a second on a standard computer) with the good version, but will take very long if the performance degrades.

It would indeed perform the test, but since it has the possibility of making a "short" test take a long time, it might raise some eyebrows. I would ask for opinions [on Zulip](https://sagemath.zulipchat.com/) before incorporating such speed tests.

But I also believe in preliminary and offline tests, performed by the developer before pushing, that are more extensive than the "TESTS" cases.

src/sage/combinat/integer_vector.py
jukkakohonen: These things are _bound to happen_ if one goes else-hunting indiscriminately. It would be good to remember the saying "don't fix it if it isn't broken".

I have heard that some misguided coding standards try to fix "no else after return", indiscriminately, but I have not heard SageMath would be among those.

mantepse: I don't care much about any particular rule, however, I favour uniform coding style.  I don't really see why always having the `else:` would be any better or any worse, both have their pro and cons.  In any case, at least at some point, the linter checked for `else` after `return`, so a fair bit of the codebase adheres to this rule.

I remember that, when I was using Turbo Pascal in the 90s, I spent a night over a forgotten semicolon (or was it a comma?).  Who cares?

jukkakohonen: I have given my advice, based on years of coding, and will leave it at that. Word to the wise.

src/sage/combinat/integer_vector.py
```diff
@@ -780,6 +780,20 @@ def __contains__(self, x):
                 return False
         return True
 
+    def _unrank_helper(self, x, rtn):
```
mantepse, line 783: this needs a docstring and one test

src/sage/combinat/integer_vector.py
```diff
@@ -968,6 +1054,76 @@ def __contains__(self, x):
...
+            [0, 0, 2, 0, 0]
+        """
+        if self.k == 0 and x != 0:
+            raise IndexError(f"Index {x} is out of range for the IntegerVector.")
+        n = 0
+        rtn = [0]*self.k
+        while self.rank(rtn) <= x:
+            n += 1
+            rtn[0] = n
+        rtn[0] -= 1
```
mantepse, line 1126: The above can be simplified to
```
        rtn = [0]*self.k
        while self.rank(rtn) <= x:
            rtn[0] += 1
        rtn[0] -= 1
```
(no need for `n`)

Also, the corner case 
```
sage: IntegerVectors(k=0).unrank(0)
[]
```
needs a test (I am guessing that it needs a separate treatment).

src/sage/combinat/integer_vector.py
```diff
@@ -780,6 +780,37 @@ def __contains__(self, x):
                 return False
         return True
 
+    def _unrank_helper(self, x, rtn):
+        """
+        return an element at rank ``x``.
```
mantepse: `Return the` instead of `return an`.

I also forgot:
```
INPUT:

    - ``x`` - a nonnegative integer
    - ``rtn`` - a list of nonnegative integers

Return an element of rank ``x`` by iterating through all integer vectors beginning with ``rtn``.
```

src/sage/combinat/integer_vector.py
```diff
@@ -780,6 +780,37 @@ def __contains__(self, x):
...
+
+        EXAMPLES::
+
+        sage: i=IntegerVectors(k=5)
```
mantepse: After `::` you have to indent by 4 spaces.  Also, we try to follow the same coding recommendations in the doctests, that is, `sage: i = IntegerVectors(k=5)` would be better and `sage: IV = IntegerVectors(k=5) ` would be even better.

mantepse:
@adrinospy, please excuse me being so fussy!  I very much like your contribution!

There are only 3 comments left for positive review!

src/sage/combinat/integer_vector.py
```diff
@@ -782,21 +782,26 @@ def __contains__(self, x):
 
     def _unrank_helper(self, x, rtn):
         """
-        return an element at rank ``x``.
+        INPUT:
+
+        - ``x`` - a nonnegative integer
+        - ``rtn`` - a list of nonnegative integers
+
+        Return the element at rank ``x`` by iterating through all integer vectors beginning with ``rtn``.
```
mantepse: Sorry, slight miscommunication on my part: the sentence `Return ...` should be just below the `"""`.

amanmoon: @mantepse Thank you for your kind words! These are my mistakes, and I'll try to minimize them next time.

mantepse:
So, last comment :-) I think I like this version
```
+        r = binomial(k + n - 1, n + 1)
+        s = 0
+        for i in range(k - 1):
+            s += x[k - 1 - i]
+            r += binomial(s + i, i + 1)
```
best, it is the easiest to translate into something you would do by hand.  If you agree, please replace the snippet in the two `rank` methods.  Otherwise, just let me know.

mantepse:
Perfekt, thank you!

Issue #36838, opened 2023-12-07
Symbolic power of a nilpotent matrix is zero


phul-ste:
### Steps To Reproduce

If we run 
```
Z=matrix([[0,1],[0,0]])
n=var('n')
Zn=Z^n;Zn
```
we get the null matrix.

### Expected Behavior

I would have expected a result like
`matrix([0^n,n*0^(n-1)],[0,0^n]])
`
to see that the matrix is not automatically zero for small n.

### Actual Behavior

The actual result is 
`matrix([[0,0],[0,0]])`

### Additional Information

I highly suspect that it is caused by the `.simplify_full()` method which turns `0^n` into `0`.

### Environment

```markdown
- **OS**: Ubuntu 20.04
- **Sage Version**: 9.0
- Jupyter Notebook
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

RuchitJagodara:
I didn't get it why do you want the result in that format because after all it is zero only, right ? @phul-ste

phul-ste:
@RuchitJagodara 
It is not zero for small values of n.
If the 0^n=0 only fails for $n=0$, the failure is more significant for nilpotent matrices.

In fact, the problem arose with students in a practical class given by a colleague of mine. There was an exact formula for some matrices (e.g.  `matrix([1,1],[0,1]])` ) while for some nilpotent matrix, the brutal limit case result was given.
(and this is part of some trust issues on sage's results from the aforementioned colleague that I had recently convinced to try an opensource software)

RuchitJagodara:
Hmm... Got it. Thank you for the clarification, @phul-ste. I will try to solve this as soon as possible. (This type of bug/error is very rare because as soon as someone reports it, it gets resolved. It's good to see that people are trying out open-source softwares.)

RuchitJagodara:
I have corrected the code and now it is working fine. Please, review it, @phul-ste.

phul-ste:
@RuchitJagodara I wrote a modification suggestion in order to use kronecker_delta instead of powers of 0.
(Hoping I am doing it right, I have never developed neither reviewed anything before)

Apart from that, everything seemed ok to me,
Thank you

RuchitJagodara:
@phul-ste, can you please review it again? I have made some changes as per your suggestions.

Issue #36410, opened 2023-10-05
Simplification of unit_step() can be inconsistent with its definition/evaluation.


UncombedCoconut:
### Steps To Reproduce

Sage gives this paradoxical simplification:
```
sage: t = var('t')
sage: assume(t == 0)   
sage: simplify(unit_step(t) - unit_step(0))
-1
```

### Expected Behavior

The expected answer above is `0`. Sage documents `unit_step` as evaluating to `1` for all nonnegative inputs, so `unit_step` expressions should only simplify to `0` when the input is known to be *strictly* negative.
Granted -- when one is truly using `unit_step` as a distribution (that is, integrating against it), Sage should be allowed to ignore the edge case.

### Actual Behavior

Sage is simplifying `unit_step(t)` to `0` anytime it can deduce `t <= 0`.

### Additional Information

I suspect the problem could be unintended semantics [here](https://github.com/sagemath/sage/blob/develop/src/sage/symbolic/ginac/inifcns.cpp#L389) -- but I don't understand Sage internals/dependencies well enough to be sure.
Thanks!

### Environment

```markdown
Tested with `SageMath version 10.1, Release Date: 2023-08-20`, `Using Python 3.11.5`, on Arch Linux.
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

DaveWitteMorris:
I think the problem is that (by default) `simplify` is performed by maxima, and [maxima's convention is that the function is 0 at 0](https://maxima.sourceforge.io/docs/manual/maxima_338.html#unit_005fstep). The easiest fix would be to change our definition of `unit_step` to be consistent with maxima, but I'm not sure that's a good idea, especially if (has been discussed) we are thinking of moving away from using maxima as the default in the long run.

FWIW:

1. [Mathematica agrees with our convention that the value is 1.](https://reference.wolfram.com/language/ref/UnitStep.html)
2. Maple does not seem to have this function. [The related Heaviside function is undefined at 0](https://www.maplesoft.com/support/help/maple/view.aspx?path=Heaviside) (like [sagemath's heaviside function](https://doc.sagemath.org/html/en/reference/functions/sage/functions/generalized.html#sage.functions.generalized.FunctionHeaviside)).
[...]

PR #36841, opened 2023-12-08
Change default domain of SR.var()

### LINKED ISSUE #36833: factor ignores complex conjugate
### Steps To Reproduce

1. I run SageMath version 10.2, Release Date: 2023-12-03.
3. When I enter the following commands:
```
y = SR.var("y") # or y = var("y")
f = y * conjugate(y)
f.factor()
```

4. This returns y^2 which is clearly not true for y being complex.

### Expected Behavior

The expected result for this would be to just return y * conjugate(y) as by documentation when declaring y as a variable of SR it is assumed that y is in the complex plain.

### Actual Behavior

The factor function treats y as a real variable and thus returns the wrong result of y^2.

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Arch Linux
- **Kernel**: Linux 6.6.4-arch1-1
- **Sage Version**: 10.2
- **Python Version**: 3.11.6
```

### Checklist

RuchitJagodara:
This patch fixes #36833 by changing the default domain of SR.var()
function to complex. By this way, every variable whose domain is not defined will be considered
in complex plane, by default.

The issue was with the default value only because whenever we were defining a variable without specifying any domain it was calling symbol() function with domain=None; Subsequently, it calls new_Expression_symbol() function with domain=None. In this code, I found that it was not giving any specific domain to the variable and so when we define a new variable using SR.var() function, it is passing the is_real() (Because it does not have any specific domain) test but giving the wrong answer on factor() function, if we assume that the variable is in complex plane by default which is correct according to the documentation.

### :memo: Checklist


src/sage/symbolic/ring.pyx
```diff
@@ -862,6 +862,16 @@ cdef class SymbolicRing(sage.rings.abc.SymbolicRing):
             Traceback (most recent call last):
             ...
             ValueError: cannot specify n for multiple symbol names
+        
+        Check that :trac:`36833` is fixed: Variables whose domain is not
+        defined should be considered in complex plane as described in documentation::
```
dimpase: Could you point out where in documentation you see this " considered in complex plane" ?

dimpase:
we might see lots of doctest errors now, not sure.

src/sage/symbolic/ring.pyx
RuchitJagodara: Screenshot from 2023-12-10 19-07-00

RuchitJagodara:
Then, I think it would be good if we change the factor function instead of doing this. I'll change it and will update.

RuchitJagodara:
@dimpase Do you know any function that can return the domain of a symbolic variable?

dimpase:
> @dimpase Do you know any function that can return the domain of a symbolic variable?

no, I don't.

By the way, I just saw the place where it's claimed that `var` by default returns
complex variable - it's in the output `sage: var?` - so you can find it in the code by
e.g. looking at the bottom of `sage: var??`

RuchitJagodara:
Can't we implement a new class for variables of symbolic ring? By this method, we will be able to track domain of the variable and once we are able to access the domain of variable, I think it will be easier to handle these type of issues.

src/sage/symbolic/ring.pyx
dimpase: thanks!

dimpase:
I don't think it's a good idea to make `domain="complex"` the default. It seems that most everything in Sage assumes it to be real. E.g. if you make this change then differentiating with default variables will become complex differentiation, which is a rather different story, mathematically - and Sage does not do it.
E.g.
```python
sage: y = SR.var("y",domain="complex")
sage: f(y) = y * conjugate(y)
sage: f.diff(y)
y |--> y*diff(conjugate(y), y) + conjugate(y)
```
doesn't make much sense, as in fact $\frac{d\overline{y}}{dy}$ does not exist (e.g. as Cauchy equations don't hold).

RuchitJagodara:
Hmm, got it. But when we declare a variable without declaring any domain, it is clear that it is not real from the  following,

```python
sage: n = SR.var("n"); n
n
sage: n.is_real()
False
``` 

So it simply implies that it is complex. So I am not getting the point that what is wrong here, if the variable is not real then it should not be simplified to y^2, anyhow. But factor function is doing so. I think the error is occurring with the maxima. Actually maxima is simplifying many functions which should not be simplified like 0^n is being converted to 0 (this is referenced from the issue #36838 ) , here y * conjugate(y) is being converted to y^2.

dimpase:
On 12 December 2023 15:48:25 GMT, Ruchit Jagodara ***@***.***> wrote:
>Hmm, got it. But when we declare a variable without declaring any domain, it is clear that it is not real from the  following,
>
>```python
>sage: n = SR.var("n"); n
>n
>sage: n.is_real()
>False
>``` 
>
>So it simply implies that it is complex. 

it'd imply this if our docs were correct:-)

By default  `domain=None`, and it could be that `is_real()` just checked that.



> So I am not getting the point that what is wrong here, if the variable is not real then it should not be simplified to y^2, anyhow. But factor function is doing so. I think the error is occurring with the maxima. Actually maxima is simplifying many functions which should not be simplified like 0^n is being converted to 0 (this is referenced from the issue #36838 ) , here y * conjugate(y) is being converted to y^2. 

maxima has no concept of a domain for a variable, it seems to me.

RuchitJagodara:
> maxima has no concept of a domain for a variable, it seems to me.

But it has to be, because when we do operation like below 

```python
sage: n = var('n',domain="complex"); n
n
sage: f1 = n * conjugate(n); f1
n*conjugate(n)
sage: f1.factor()
n*conjugate(n)
```

Here, we can clearly see that it is giving correct answer if we declare domain="complex", but it is not giving the same answer when we do it as below, which means it has something to do with the domain.

```python
sage: b = var("b")
sage: f2 = b * conjugate(b); f2
b*conjugate(b)
sage: f2.factor()
b^2
```

dimpase:
> > maxima has no concept of a domain for a variable, it seems to me.
> 
> But it has to be, because when we do operation like below
> 
> ```python
> sage: n = var('n',domain="complex"); n
> n
> sage: f1 = n * conjugate(n); f1
> n*conjugate(n)
> sage: f1.factor()
> n*conjugate(n)
> ```
> 
> Here, we can clearly see that it is giving correct answer if we declare domain="complex", but it is not giving the same answer when we do it as below, which means it has something to do with the domain.
> 
> ```python
> sage: b = var("b")
> sage: f2 = b * conjugate(b); f2
> b*conjugate(b)
> sage: f2.factor()
> b^2
> ```

Sorry, indeed, you are right. Maxima has [domain](https://maxima.sourceforge.io/docs/manual/maxima_46.html#index-domain), which is a domain for computation to be done (although it's not to declare that a variable is real or complex) , and  a similar thing called `assume` (which allows things like `assume(x>0)`, or indeed `x` to be real).

Better, it has [declare](https://maxima.sourceforge.io/docs/manual/maxima_49.html#index-conjugate) - which seems to be exactly what is needed here. So one can do
```
$ maxima 
Maxima 5.47.0 https://maxima.sourceforge.io
using Lisp SBCL 2.3.9
Distributed under the GNU Public License. See the file COPYING.
Dedicated to the memory of William Schelter.
The function bug_report() provides bug reporting information.
(%i4) declare(x, complex);
(%o4)                                done
(%i5) x*conjugate(x);
(%o5)                           x conjugate(x)
(%i6) declare(y, real);
(%o6)                                done
(%i7) y*conjugate(y);
                                       2
(%o7)                                 y
(%i8) 

```

RuchitJagodara:
Hmm... But in sage when we are doing this 
```python
sage: y = SR.var("y","real"); y
y
sage: f = y * conjugate(y); f
y*conjugate(y)
sage: f
y*conjugate(y)
sage: f.simplify_full()
y^2
``` 

Here, you can see that for variable whose domain is real, it is not simplifying conjuage(y) and for that I have to use simplify_full() function but I think this should be done when we were multiplying y with conjuage(y) so the value of f should be y^2 not y*conjugate(y).

RuchitJagodara:
I found an issue where one had similar problem with maxima, [discussion link](https://github.com/sagemath/sage/issues/36410#issuecomment-1763476272)

RuchitJagodara:
Should I change the documentation instead, which may specify that the default domain of `var` is None? This means that the variable will not have any specific domain. Or should I just leave it as it is?

dimpase:
Yes, please, we are not going to make the default domain `complex`. Please fix the documentation instead, to specify that the default domain is `None`.

RuchitJagodara:
Sorry for the delay! Actually, I was checking if there is any other fix that can be applied, but I think we cannot do anything as the problem is with maxima only. So, I have changed the documentation.

Can you please review my changes now ?

dimpase (approved):
OK

PR #36845, opened 2023-12-09
Give more precise answer on symbolic power of a matrix

### LINKED ISSUE #36838: Symbolic power of a nilpotent matrix is zero

RuchitJagodara:
- Changed _matrix_power_symbolic function to consider mk=0 case
- Changed the _matrix_power_symbolic function to handle all the cases
- Created tests covering the changes

This PR improves the answer given by _matrix_power_symbolic() function, which involves finding the symbolic power of a matrix, and also gives more precise answer on evalution. eg, before, when we were evaluating the zeroth power of a matrix sometimes it was giving wrong answer like below,
```python
        sage: n = var("n")
        sage: A = matrix(ZZ, [[1,-1], [-1,1]])
        sage: An = A^(n) 
        [ 1/2*2^n -1/2*2^n]
        [-1/2*2^n  1/2*2^n]
        sage: An({n:0})
        [ 1/2 -1/2]
        [-1/2  1/2]
``` 
Here, for n=0, it is not giving an identity matrix which is wrong, but after the change it will give answer as below,
```python
        sage: n = var("n")
        sage: A = matrix(ZZ, [[1,-1], [-1,1]])
        sage: An = A^(n) 
        [ 1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1) -1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)]
        [-1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)  1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)]
        sage: An({n:0})
        [1 0]
        [0 1]
```

Also, This patch fixes #36838 . The function, _matrix_power_symbolic() was giving symbolic power
of a nilpotent matrix as a null matrix but the correct answer should be represented
in the form of kronecker_delta() function. In the case of mk=0, simplifying only binomial
term should work because afterall the form 0^(n-i) should be simplifed to kronecker_delta() function and no further simplification is needed.
And in every other case it is simplifying whole term so it will handle all other cases.

### :memo: Checklist

### :hourglass: Dependencies

None


src/sage/matrix/matrix2.pyx
```diff
@@ -18656,8 +18669,14 @@ def _matrix_power_symbolic(A, n):
...
             mk = mk.radical_expression()
-        vk = [(binomial(n, i) * mk**(n-i)).simplify_full()
-              for i in range(nk)]
+
+        # If mk=0 return mk^(n-i) instead of mk**(n-i)
+        if mk:
+            vk = [(binomial(n, i) * mk._pow_(n-i)).simplify_full()
+                  for i in range(nk)]
+        else:
+            vk = [(binomial(n, i).simplify_full() * mk._pow_(n-i))
```
phul-ste: In fact, a more suitable result would be given by             
`vk = [(binomial(n, i).simplify_full() * kronecker_delta(n,i))
                  for i in range(nk)]`
This even allows to evaluate with small values of n

src/sage/matrix/matrix2.pyx
```diff
@@ -18618,12 +18618,25 @@ def _matrix_power_symbolic(A, n):
...
+    nilpotent matrix::
+
+        sage: A = matrix([[0,1],[0,0]]); A
+        [0 1]
+        [0 0]
+        sage: n = var('n'); n
+        n
+        sage: An = A^n; An
+        [        0^n 0^(n - 1)*n]
+        [          0         0^n]
```
phul-ste: According to the comment on line 18678,
We would then have something like
`[        kronecker_delta(0, n) n*kronecker_delta(1, n)]`
`[        0                     kronecker_delta(0, n)]`
that can then be evaluated
`sage: An({n:0})`
`[1  0]` 
`[0  1]`

phul-ste (approved):
I checked and it works perfectly to me.

sage: An = _matrix_power_symbolic(A,n); An
#        [  kronecker_delta(0, n) n*kronecker_delta(1, n)]
#        [                      0   kronecker_delta(0, n)]
sage: print(all(A^k == An.subs({n: k}) for k in range(8)))
#True

src/sage/matrix/matrix2.pyx
```diff
@@ -18656,8 +18660,14 @@ def _matrix_power_symbolic(A, n):
...
         if hasattr(mk, 'radical_expression'):
             mk = mk.radical_expression()
-        vk = [(binomial(n, i) * mk**(n-i)).simplify_full()
-              for i in range(nk)]
+
+        # Return mk^(n-i) instead of mk**(n-i) if mk=0
+        if mk:
+            vk = [(binomial(n, i) * mk._pow_(n-i)).simplify_full()
+                  for i in range(nk)]
+        else:
```
phul-ste, line 18671: In fact, I realize that after the computation, we can get a value even for negative n (or even also for non-integer n).

I think that, somehow, the system should raise an error if mk=0 but the evaluation is made at bad values of n. I have no idea on how to do that properly.

RuchitJagodara: This can be done by declaring variable with domain="integer", here when you do not give any domain to a variable it will consider it in complex plane by default. 

But we can check whether the variable have an integer value of not at the time of evaluation.

> I think that, somehow, the system should raise an error if mk=0 but the evaluation is made at bad values of n.

I didn't get that, why system should give an error if mk=0 ?

RuchitJagodara: I'll implement this in a new PR as this topic is not concentrated on nilpotent matrices.

phul-ste: > This can be done by declaring variable with domain="integer", here when you do not give any domain to a variable it will consider it in complex plane by default.
> 

great!

> But we can check whether the variable have an integer value of not at the time of evaluation.
> 
> > I think that, somehow, the system should raise an error if mk=0 but the evaluation is made at bad values of n.
> 
> I didn't get that, why system should give an error if mk=0 ?

It should raise an error if mk=0 and n negative since the matrix is then non-invertible. For n positive non-integer, I don't know if it makes sense either.

phul-ste: > I'll implement this in a new PR as this topic is not concentrated on nilpotent matrices.

By the way, I already opened an other issue #36863 for general simplification of `0^n`. 

But the problem for matrices is somehow different since: 

- the exponent is generally assumed to be an integer
- the general formula for the power of a Jordan block with eigenvalue 0 involves some expression of the form `P(n)*0^{n-i}` (P polynomial)  where the meaning of `0^{n-i}`for `0<i <=n`should be understood as `kronecker_delta(i,n)` in this special case (assuming that n is a *non-negative integer)

RuchitJagodara:
```python
    Testing exponentiation in the integer ring::

        sage: A = matrix(ZZ, [[1,-1], [-1,1]])
        sage: A^(2*n+1)                                                                 # needs sage.symbolic
        [ 1/2*2^(2*n + 1) -1/2*2^(2*n + 1)]
        [-1/2*2^(2*n + 1)  1/2*2^(2*n + 1)]
``` 

@dimpase, Can I change the above doctest which was there before my changes, because here, according to my change, I think the answer should be 
    [ 1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1) -1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)]
    [-1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)  1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)]

But the answer given in doctest is different. And to support my answer, I'll say that this is working for n=0 where we should get an indentity matrix but the answer given in doctest is false as it will not give identity matrix for n=0. And becuase of this my PR is failing that doctest check.

phul-ste:
> @dimpase, Can I change the above doctest which was there before my changes, because here, according to my change, I think the answer should be
> 
> ```python
>     [ 1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1) -1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)]
>     [-1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)  1/2*2^(2*n + 1) + 1/2*kronecker_delta(0, 2*n + 1)]
> ```
> 
> But the answer given in doctest is different. And to support my answer, I'll say that this is working for n=0 where we should get an indentity matrix but the answer given in doctest is false as it will not give identity matrix for n=0. And becuase of this my PR is failing that doctest check.

*A precision (in case your argument is misunderstood by others),  you mean 0 as the power of the matrix (and not for n=0 when we consider the power 2*n+1)

RuchitJagodara:
Ohh, ya that's right, thank you for pointing out.

src/sage/matrix/matrix2.pyx
RuchitJagodara: > In fact, I realize that after the computation, we can get a value even for negative n (or even also for non-integer n).
> I think that, somehow, the system should raise an error if mk=0 but the evaluation is made at bad values of n. I have no idea on how to do that properly.

I checked what I can do to determine whether we are evaluating an integer power or not. I found that when evaluating the matrix, we only replace the variables with their values, and we cannot impose a condition that the variable should be an integer for obvious reasons because we are not sure that the variable is the symbolic power only. The only place where we can make a change is when calculating the power of the matrix (i.e., A^n). However, if we do so, it will not be ideal because we can define a real power of a matrix, and imposing a condition would not be appropriate. So, I think implementing a system for this is not practical. So it is better to use domain=integer while defining a variable.

dimpase:
It makes sense to talk about changing the doctest result to match the output of the changed function.

By the way, I don't quite understand the comment
```
# Return mk^(n-i) instead of mk**(n-i) if mk=0
```
Do you mean `^` as in Python? That's confusing. Better use a maths way to explain what you mean.

RuchitJagodara:
Okay, I'll change it.

RuchitJagodara:
In `Build & Test / build (pull_request)` it is showing that doctests are failing on some files. But when I checked it in my machine was working fine and passing all doctests for all files. Like for below it is showing that 130 doctests are failed but when I ran the same command in my machine it is showing that all doctests are passes.
```
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/libs/coxeter3/coxeter_group.py  # 130 doctests failed
```

RuchitJagodara:
@dimpase can you please review this or assign someone for this.

dimpase:
Please fix the linting errors
```
/sage/matrix/matrix2.pyx:3185:30: W291 trailing whitespace
./sage/matrix/matrix2.pyx:3191:29: W291 trailing whitespace
./sage/matrix/matrix2.pyx:3199:44: W291 trailing whitespace
./sage/matrix/matrix2.pyx:18665:106: W291 trailing whitespace
./sage/matrix/matrix2.pyx:18666:106: W291 trailing whitespace
```

dimpase (approved):
looks good, thanks

PR #36849, opened 2023-12-10
Problem with orientations of simplices in simplicial complex maps.

jhpalmieri:
Problem with orientation of simplices in simplicial complex maps

In rare cases, Sage incorrectly computes the chain complex map associated to a map of simplicial complexes. The problem is that the orientation of each simplex in the image is not correctly compared to its actual orientation in the codomain: an edge `(f(v), f(w))` is not compared correctly to see whether the actual simplex in the codomain is stored as `(f(v), f(w))` or `(f(w), f(v))`.

I think that the default sorting of vertices (and hence simplices) means that this usually doesn't come up, but it can arise when the vertices do not have an obvious ordering. See https://ask.sagemath.org/question/74754/chain-morphism-between-subdivisions/ for one such case.

### :memo: Checklist


tscrim (approved):
LGTM.

Although if you want to make one change, please add a reference to this PR in the `TESTS:` block. Once you do that or decide not to, you can change this to a positive review.

jhpalmieri:
Thanks very much. I'll make the change you suggested soon and then change the status.

tscrim:
Thanks.

PR #36857, opened 2023-12-10
use unicode characters to display tables

fchapoton:
This is using unicode box-drawing characters to make better-looking tables.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.

### :hourglass: Dependencies

- #36861

PR #36861, opened 2023-12-11
Use lualatex for better unicode support 

### LINKED ISSUE #18370: Unicode support in reference manual and other docs, e.g. developer manual
<div id="comment:0"></div>

LaTeX isn't particularly good at typesetting Unicode text. While the utf8 input encoding can help recognizing the most common symbols even if the input is Unicode, the output is essentially restricted to 8-bit encodings so every special symbol has to be set up somewhere. `src/doc/common/conf.py` has a list of such setup rules, although some are pretty hackish, replacing unicode symbols with stuff that only looks remotely similar.

Unicode in doctest output (and perhaps other parts of the reference manual as well) has caused problems in the past, e.g. ticket:15802 and [#18270 comment:18](https://github.com/sagemath/sage/issues/18270#comment:18). It is likely to cause more problems in the future, when we make more extensive use of Unicode in Sage output once #18357 gets implemented. So Unicode in the doctests is going to be increasingly common, and it would be nice if we didn't have to jump through hoops to get that represented in the PDF output.

As far as I understand it, both XeTeX / `xelatex` and luaTeX / `lualatex` provide a more modern approach to font handling, and considerably better Unicode support. I haven't yet gotten things to work that way, but my hope would be that we could gain access to all the symbols of a sufficiently complete Unicode font. Apparently, current versions of Sphinx don't officially support either of these engines yet. And things don't work out of the box. But perhaps things can be made to work with a bit of patching. If so, that might greatly improve the situation.

See also #14153

CC:  @nexttime @embray @jdemeyer @hivert

Component: **documentation**

Keywords: **unicode**

_Issue created by migration from https://trac.sagemath.org/ticket/18370_

kwankyu:
to support https://github.com/sagemath/sage/pull/36857.

There have been numerous problems due to limited support of unicode in pdflatex. I see no reason why we should not switch to modern tex engine lualatex, as suggested in

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_engine

We set lualatex as the default engine for
- building the sage documentation, in `sage_docbuild/`
- rendering pdf images, in `src/sage/misc/latex.py`

In  `src/sage/misc/latex.py`, the default engine is determined by availability of the engines  with the order of preference: lualatex, xelatex, pdflatex.

Along the way, we
- remove latex-related code deprecated in #32650.
- make lots of cosmetic edits in `src/sage/misc/latex.py`
- add support "lualatex" as an alternative to "[pdf|xe]latex" in `src/sage/misc/latex.py`

We add optional (dummy) packages `texlive_luatex`, `free_fonts`, `xindy` as new dependencies of `sagemath_doc_pdf` in addition to `texlive` optional package. It seems that depending on the version and the platform, `texlive` may already include the new dependencies. If not, you need to install the new dependencies (as `_recommended` by the dummy package)

On Ubuntu, the requirement is fulfilled by
```
sudo apt install texlive texlive-luatex fonts-freefont-otf xindy
```

The built pdf doc is available here: https://deploy-livedoc--sagemath.netlify.app

To test `view` via lualatex, for example,
```
latex.extra_preamble(r"\usepackage{fontspec}\setmainfont{Arial}\setmonofont{Arial}")
view('Εύρηκα')
```
you should build sage on your local platform with this PR. You may need to install `texlive-full` before testing. It seems that `texlive` (2019) is not enough even for the `view` via pdftex.

sage-devel thread seeking Unicode testers: https://groups.google.com/g/sage-devel/c/tG2LK6Jvw0I

Resolves #18370

### :memo: Checklist


dimpase:
AFAIK, the official successor of pfd(la)tex is lua(la)tex, not xe(la)tex. See e.g. https://tex.stackexchange.com/questions/126206/why-choose-lualatex-over-xelatex and https://en.wikipedia.org/wiki/LuaTeX

It appears that xetex is not developed further  since 2019  or so.

kwankyu:
> AFAIK, the official successor of pfd(la)tex is lua(la)tex, not xe(la)tex. See e.g. https://tex.stackexchange.com/questions/126206/why-choose-lualatex-over-xelatex and https://en.wikipedia.org/wiki/LuaTeX

There's no "official" successor. 

> It appears that xetex is not developed further since 2019 or so.

Like TeX itself, it is not a flaw that xetex is not further developed. It is a great flaw if it was not maintained. I have (and had) no problem at all with xelatex in typesetting my scripts in English and Korean with little bits of Chinese and Greek. I'd say xetex is just stable.

I have zero experience with LuaTeX. I know it is famous because of its scripting capability with Lua language. Do we need the capability?

I just hope that with minimal efforts (on my side), xelatex solves the problems we now face. 

If you want to solve more problems with more capable luatex, I welcome you opening a PR as a sequel. Then this PR will serve as a stepping stone.

dimpase:
luatex is at least the official successor of pdftex in TexLive: https://en.wikipedia.org/wiki/LuaTeX

I use lualatex, it's quite nice, and its unicode support is better in some ways. Also,
xetex is known not to work very well with TikZ and in bidirectional typesetting.

Why do we need two PRs, and why choose an older project which is being phased out in favour of a newer one?

kwankyu:
> I use lualatex, it's quite nice, and its unicode support is better in some ways. Also, xetex is known not to work very well with TikZ and in bidirectional typesetting.

OK. Good.

> Why do we need two PRs, and why choose an older project which is being phased out in favour of a newer one?

Then would you open a PR and work on it? I will happily close this PR then.

dimpase:
OK, let me try lualatex - it it's a bit more work, as Sage does know about xelatex, but does not know about lualatex.
If it does, we can re-purpose this PR.

dimpase:
by the way, I tested this PR, and saw problems with Japanese and Cyrillic. So it's one way or another, more work needed, it seems.

kwankyu:
> by the way, I tested this PR, and saw problems with Japanese and Cyrillic. So it's one way or another, more work needed, it seems.

Thanks. This PR is in testing phase (not ready for review yet). I will work on the problems. 

You may want to prepare your luatex PR as a sequel to this PR, to not duplicate efforts.

kwankyu:
> by the way, I tested this PR, and saw problems with Japanese and Cyrillic. 

I don't recognize any problem from the built pdf docs. What problems did you see? Are they new?

kwankyu:
Ah sorry, my pdf docs were built by pdflatex, somehow...

kwankyu:
> OK, let me try lualatex - it it's a bit more work, as Sage does know about xelatex, but does not know about lualatex. If it does, we can re-purpose this PR.

I added support of lualatex in

https://github.com/sagemath/sage/pull/36861/commits/8a966dd5c6449f435360be79399d741702432592

while setting xelatex as the default engine for rendering pdfs, in `src/sage/misc/latex.py`.

If this setup works out well, then later, in your PR, you may replace xelatex with lualatex in the places setting the default engine.

dimpase:
On Tue, Dec 12, 2023 at 01:33:51AM -0800, Kwankyu Lee wrote:
> 
> https://github.com/sagemath/sage/pull/36861/commits/8a966dd5c6449f435360be79399d741702432592
> 
> while setting xelatex as the default engine for rendering pdfs, in `src/sage/misc/latex.py`.
> 

one needs to change build/pkgs/texlive/spkg-configure.m4 too. The list
of LaTeX packages, 
e.g. [unicode-math](https://ctan.org/pkg/unicode-math?lang=en) is most probably a good idea,
and the programs (No need for dvi* things any more) required there should be adjusted.

kwankyu:
> one needs to change build/pkgs/texlive/spkg-configure.m4 too. The list of LaTeX packages, e.g. [unicode-math](https://ctan.org/pkg/unicode-math?lang=en) is most probably a good idea, 

That may be a useful addition, though it is not required.

> and the programs (No need for dvi* things any more) required there should be adjusted.

We support [pdf|xe|lua]latex in addition to latex. dvi* things are still necessary for latex. Anyway, that is a separate concern.

kwankyu:
From Sphinx doc,
```
 * LaTeX builder with :confval:`latex_engine` set to ``'xelatex'`` or to 
   ``'lualatex'`` requires (by default) the ``FreeFont`` fonts, 
   which in Ubuntu xenial are provided by package ``fonts-freefont-otf``, and 
   e.g. in Fedora 29 via package ``texlive-gnu-freefont``. 
```
Hence we need to work on an spkg to proceeed.

dimpase:
> We support [pdf|xe|lua]latex in addition to latex. dvi* things are still necessary for latex

Why do we need to support `pdflatex` once we switch ? 
There is no reason for this.
`pdflatex` precludes using unicode in maths mode. With [xe|lua]latex one can write unicode maths without the endless `\declareUnicodeCharacter` macros. Surely we can dump a large number (1000s of lines) of these in the source, setup various `\ifxetex`s etc, but why? 

I discussed pdftex/xetex/luatex compatibility with David Carlise here: https://tex.stackexchange.com/questions/601579/unicode-maths-in-pdflatex

kwankyu:
> > We support [pdf|xe|lua]latex in addition to latex. dvi* things are still necessary for latex
> 
> Why do we need to support `pdflatex` once we switch ? There is no reason for this. 

`src/sage/misc/latex.py` has supported latex, pdflatex, xelatex up to now. And we are adding luatex. It is the user's choice which engine is used. We may change the default engine. But why do we suddenly stop support for a tex engine for no benefit?

dimpase:
> `src/sage/misc/latex.py` has supported latex, pdflatex, xelatex up to now. And we are adding luatex. It is the user's choice which engine is used. We may change the default engine. But why do we suddenly stop support for a tex engine for no benefit?

The benefit of not using latex/pdflatex is that you can use unicode freely, without worrying about missing `\declareUnicodeCharacter`.

kwankyu:
By the way, "unicode in maths mode" is not in the scope of this PR.

dimpase:
> By the way, "unicode in maths mode" is not in the scope of this PR.

it is very much in the scope, as using xelatex will easily break compatibility with (pdf)latex. To ensure the compatibility is not broken, you'd need to test the non-default (pdf)latex settings.

kwankyu:
> > `src/sage/misc/latex.py` has supported latex, pdflatex, xelatex up to now. And we are adding luatex. It is the user's choice which engine is used. We may change the default engine. But why do we suddenly stop support for a tex engine for no benefit?
> 
> The benefit of not using latex/pdflatex is that you can use unicode freely, without worrying about missing `\declareUnicodeCharacter`.

If a user chooses latex/pdflatex, he simply has chosen to not use unicode freely. We don't have to support him as much as xelatex/lualatex users regarding unicode.

kwankyu:
> > By the way, "unicode in maths mode" is not in the scope of this PR.
> 
> it is very much in the scope, as using xelatex will easily break compatibility with (pdf)latex. To ensure the compatibility is not broken, you'd need to test the non-default (pdf)latex settings.

Sorry I don't understand what compatibility you are thinking. Please demonstrate with some sample code example or with some images if relevant. 

My current concern is the font problem. I don't want to be distracted by a problem that I don't see (yet).

kwankyu:
> > If a user chooses latex/pdflatex, he simply has chosen to not use unicode freely. We don't have to support him as much as xelatex/lualatex users regarding unicode.
> 
> So I submit a PR with unicode math, and it passes the default (say, xelatex) pdf docs builder. Not testing the PR with non-default builder means that our docs might be broken in sense they can't be built with (pdf)latex any more. The latter is not what I'd call "supporting (pdf)latex" - it's akin to saying we support Python 2 (just not test with it ;-))

Ah. I see what you are confused with. We use only xelatex to compile our documentation. We don't "support" pdflatex here.

I meant supporting pdflatex in `src/sage/misc/latex.py`, which is used for example `view(object)`. This has nothing to do with compiling our documentation.

dimpase:
OK, that's fine with me. Certainly then all these `\DeclareUnicodeCharacter` in  `\ifPDFTeX` can be removed.

kwankyu:
OK. I will consider that.

kwankyu:
> From Sphinx doc,
> 
> ```
>  * LaTeX builder with :confval:`latex_engine` set to ``'xelatex'`` or to 
>    ``'lualatex'`` requires (by default) the ``FreeFont`` fonts, 
>    which in Ubuntu xenial are provided by package ``fonts-freefont-otf``, and 
>    e.g. in Fedora 29 via package ``texlive-gnu-freefont``. 
> ```
> 
> How should we proceed? Perhaps `fonts-freefont-otf` should be included to [_recommended](https://doc.sagemath.org/html/en/reference/spkg/_recommended.html#spkg-recommended)?
> 
> We need a dummy package created from
> 
> https://repology.org/project/fonts:gnu-freefont/versions
> 

I think "free fonts" should be included in `_required`.

@mkoeppe any tip on how to organize the spkg?

mkoeppe:
Is this just for the PDF docbuild? As sagemath_doc_pdf is an optional package, I think the fonts package should just be a dependency of that

kwankyu:
Main concern is PDF build of sage documentation.

I am also making xelatex the default pdf engine for say `view(object)` (which renders the object by latex to pdf).

dimpase:
> Main concern is PDF build of sage documentation.
> 
> I am also making xelatex the default pdf engine for say `view(object)` (which renders the object by latex to pdf).

in this case it would be nice to have better fonts for `view`, so it's better be standard.

mkoeppe:
For that, the package should be added to `build/pkgs/_recommended/dependencies` (where `texlive` also appears).

kwankyu:
> For that, the package should be added to `build/pkgs/_recommended/dependencies` (where `texlive` also appears).

So I did it right here. But perhaps I did things wrong in writing the spkg for free fonts. It seems not working for the build & test workflow...

build/pkgs/free_fonts/SPKG.rst
```diff
@@ -0,0 +1,10 @@
+free fonts
+==========
+
+Description
+-----------
+
+This dummy package represents the free fonts.
+
+We do not have an SPKG for it. The purpose of this dummy package is to
+associate system package lists with it.
```
mkoeppe, line 13: Add section `Upstream Contact`?

build/pkgs/free_fonts/SPKG.rst
```diff
@@ -0,0 +1,10 @@
+free fonts
```
mkoeppe:
```suggestion
free_fonts: SOME DESCRIPTION HERE
```

build/pkgs/free_fonts/SPKG.rst
kwankyu: Done, hopefully correctly.

build/pkgs/free_fonts/SPKG.rst
kwankyu: Done.

kwankyu: Changed "free fonts: ..." to "free_fonts: ..."

kwankyu:
Should I keep (or modify) `build/pkgs/free_fonts/spkg-configure.m4`? This seems useless in this case.

kwankyu:
> > Main concern is PDF build of sage documentation.
> > I am also making xelatex the default pdf engine for say `view(object)` (which renders the object by latex to pdf).
> 
> in this case it would be nice to have better fonts for `view`, so it's better be standard.

`view` doesn't work without texlive anyway. As texlive is optional, it seems right to keep `free_fonts` as optional too.

kwankyu:
I omitted `distros/homebrew.txt` because I don't know what to add here. There is https://github.com/Homebrew/homebrew-cask-fonts though...

build/pkgs/free_fonts/SPKG.rst
```diff
@@ -0,0 +1,23 @@
...
+publishing. It is Unicode-encoded for compatibility with all modern operating
+systems.
+
+We do not have an SPKG for it. The purpose of this dummy package is to
+associate system package lists with it.
+
+License
+-------
+
+GNU General Public License
```
mkoeppe: version?

mkoeppe:
> Should I keep (or modify) `build/pkgs/free_fonts/spkg-configure.m4`? This seems useless in this case.

I think there should be check for this package. Probably by depending on texlive and using xelatex (or lualatex) to check for the fonts

Also add the package to `build/pkgs/sagemath_doc_pdf/dependencies` (if it's now required for that)

build/pkgs/free_fonts/SPKG.rst
kwankyu: Done.

kwankyu:
> > Should I keep (or modify) `build/pkgs/free_fonts/spkg-configure.m4`? This seems useless in this case.
> 
> I think there should be check for this package. Probably by depending on texlive and using xelatex (or lualatex) to check for the fonts

We may use
```
kpsewhich FreeSerif.ttf
```
But I don't speak autotools...

kwankyu:
Is there a similar spkg that I can consult to write the `spkg-configure.m4`?

mkoeppe:
Just adapt from build/pkgs/texlive/spkg-configure.m4

mkoeppe:
In fact, maybe the whole new package can just be replaced by adding a check in that file?

kwankyu:
> In fact, maybe the whole new package can just be replaced by adding a check in that file?

Perhaps not a good idea, since gnu free fonts is independent from texlive.

mkoeppe:
> I omitted `distros/homebrew.txt` because I don't know what to add here.

I have the fonts as part of my homebrew-installed texlive installation (`mactex` cask)

mkoeppe:
> Perhaps not a good idea, since gnu free fonts is independent from texlive.

OK, I agree

kwankyu:
> Just adapt from build/pkgs/texlive/spkg-configure.m4

Thanks. Done, hopefully correctly :-)

mkoeppe:
The package should probably be installed by `.github/workflows/doc-build-pdf.yml`

kwankyu:
Regarding that, currently we have
```
          make doc-clean doc-uninstall; make sagemath_doc_html-build-deps sagemath_doc_pdf-no-deps    
```
Should we change it to 
```
          make doc-clean doc-uninstall; make sagemath_doc_pdf 
```
?

mkoeppe:
That doesn't help. Look instead for the line where `texlive` is installed

kwankyu:
Because it is optional?

mkoeppe:
Yes, neither `texlive` nor this new package are installed in our "standard" package configurations. So they are installed into the image as part of the PDF-build workflow

.github/workflows/doc-build-pdf.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           export PATH="build/bin:$PATH"
           eval $(sage-print-system-package-command auto update)
           eval $(sage-print-system-package-command auto --yes --no-install-recommends install zip)
-          eval $(sage-print-system-package-command auto --spkg --yes --no-install-recommends install git texlive)
+          eval $(sage-print-system-package-command auto --spkg --yes --no-install-recommends install git texlive fonts-freefont)
```
mkoeppe: the spkg name needs to be used here

kwankyu:
OK. Thanks. Done.

.github/workflows/doc-build-pdf.yml
kwankyu: Ah. I see

kwankyu: Fixed.

mkoeppe:
I see in the log:
```
2023-12-13T00:18:37.5449035Z [sagemath_doc_pdf-none] Latexmk: applying rule 'makeindex manifolds.idx'...
2023-12-13T00:18:37.5449691Z [sagemath_doc_pdf-none] Can't exec "xindy": No such file or directory at (eval 13) line 14.
2023-12-13T00:18:37.5450289Z [sagemath_doc_pdf-none] Latexmk: Summary of warnings from last run of (pdf)latex:
2023-12-13T00:18:37.5450917Z [sagemath_doc_pdf-none]   Latex failed to resolve 68 reference(s)
```

kwankyu:
Yes. Proper unicode handling seems to require advanced index processor: xindy. Let's see.

kwankyu:
Again texlive on mac includes xindy. texlive on linux seems smaller...

mkoeppe:
I think xindy needs to be a dummy package too, so it can be a dependency of sagemath_doc_pdf.

kwankyu:
Yes. xindy seems to be included in the texlive package on most linux platforms as well as mac. But unfortunately it is not included in ubuntu texlive package.

kwankyu:
Checks passed on baseline. I am testing with #36857.

.github/workflows/doc-build-pdf.yml
```diff
@@ -55,9 +55,10 @@ jobs:
       - name: Update system packages
         run: |
           export PATH="build/bin:$PATH"
+          # Install free fonts and xindy along with texlive to build pdf docs by [xe|lua]latex
           eval $(sage-print-system-package-command auto update)
-          eval $(sage-print-system-package-command auto --yes --no-install-recommends install zip)
-          eval $(sage-print-system-package-command auto --spkg --yes --no-install-recommends install git texlive)
+          eval $(sage-print-system-package-command auto --yes --no-install-recommends install zip xindy)
+          eval $(sage-print-system-package-command auto --spkg --yes --no-install-recommends install git texlive free_fonts)
```
mkoeppe:
```suggestion
          eval $(sage-print-system-package-command auto --yes --no-install-recommends install zip)
          eval $(sage-print-system-package-command auto --spkg --yes --no-install-recommends install git texlive free_fonts xindy)
```

kwankyu:
Thanks.

kwankyu:
This PR works well with #36857.

kwankyu:
> ... Certainly then all these `\DeclareUnicodeCharacter` in `\ifPDFTeX` can be removed.

Yes. Done here.

mkoeppe:
I've looked through manifolds.pdf, which uses quite a bit of Unicode, and didn't see any defects.
So I think this is good to go.

mkoeppe (approved):

mkoeppe:
I haven't tested anything with `sage.misc.latex` though, so perhaps another reviewer should take a look

kwankyu:
I will also post on sage-devel to invite testers.

kwankyu:
> I haven't tested anything with `sage.misc.latex` though, so perhaps another reviewer should take a look

Actually the many `\DeclareUnicodeCharacter` commands deleted from `conf.py` were used only for building documentation. They are not related with the `sage.misc.latex` module. Except cosmetic edits, the only change I made in the module is switching the default engine from pdflatex to xelatex.

Hence, in theory, there should be no regression in `sage.misc.latex`.

kwankyu:
> luatex is at least the official successor of pdftex in TexLive: https://en.wikipedia.org/wiki/LuaTeX
> 
> I use lualatex, it's quite nice, and its unicode support is better in some ways. Also, xetex is known not to work very well with TikZ and in bidirectional typesetting.

I found a good reason that we should go with lualatex right away skipping xelatex. 

That is, on linux (at least on ubuntu), texlive does not contain xelatex! xelatex is installed by another package: texlive-xetex

Perhaps I should do all this again with lualatex.

dimpase:
¯\_(ツ)_/¯

kwankyu:
Now reborn with lualatex.

build/pkgs/_recommended/dependencies
```diff
@@ -1 +1 @@
-pandoc ffmpeg imagemagick texlive git libjpeg
+pandoc ffmpeg imagemagick texlive git libjpeg free_fonts xindy
```
mkoeppe: Should `texlive_luatex` be added here?

build/pkgs/sagemath_doc_pdf/dependencies
```diff
@@ -1 +1 @@
-sagemath_doc_html texlive
+sagemath_doc_html texlive texlive-luatex free_fonts xindy
```
mkoeppe:
```suggestion
sagemath_doc_html texlive texlive_luatex free_fonts xindy
```

build/pkgs/texlive_luatex/SPKG.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Upstream Contact
+----------------
+
+https://www.luatex.org/
+
+Dependencies
+------------
+
+texlive
```
mkoeppe: Move this to a `dependencies` file

src/sage/misc/latex.py
```diff
@@ -664,49 +557,48 @@ def latex_extra_preamble():
...
-    engine is latex or ``None``, then a dvi file is created, but if there
-    appear to be problems with it (because of PS special commands, for
-    example), then a pdf file is created instead.  The function
-    returns 'dvi' or 'pdf' to indicate which type of file is created.
-    (Detecting problems requires that dvipng be installed; if it is
-    not, then the dvi file is not checked for problems and 'dvi' is
-    returned.)  If engine is pdflatex or xelatex and there are no errors, then
-    'pdf' is returned.
+    A string which could be a string starting with ``'Error'`` (if there was a
+    problem), or it could be ``'pdf'`` or ``'dvi'``.  If engine is latex or ``None``,
```
mkoeppe:
```suggestion
    problem), or it could be ``'pdf'`` or ``'dvi'``.  If ``engine`` is ``'latex'`` or ``None``,
```

src/sage/misc/latex.py
```diff
@@ -1046,41 +941,36 @@ def _latex_preparse(self, s, locals):
...
-           evaluating Sage code in ``x``.
+        - ``locals`` - extra local variables used when evaluating Sage code in ``x``
 
         .. WARNING::
 
-           When using latex (the default), you must have 'dvipng' (or
-           'dvips' and 'convert') installed on your operating system,
-           or this command will not work.  When using pdflatex or xelatex, you
-           must have 'convert' installed.
+            When using ``latex`` (the default), you must have ``dvipng`` (or
```
mkoeppe:
```suggestion
            When using ``'latex'`` (the default), you must have ``dvipng`` (or
```

src/sage/misc/latex.py
```diff
@@ -1046,41 +941,36 @@ def _latex_preparse(self, s, locals):
...
+            ``dvips`` and ``convert``) installed on your operating system, or
+            this command will not work.  When using ``pdflatex``, ``xelatex``
+            or ``lualatex``, you must have ``convert`` installed.
```
mkoeppe:
```suggestion
            this command will not work.  When using ``'pdflatex'``, ``'xelatex'``
            or ``'lualatex'``, you must have ``convert`` installed.
```

src/sage/misc/latex.py
```diff
@@ -1766,50 +1659,49 @@ def view(objects, title='Sage', debug=False, sep='', tiny=False,
...
 
-       - ``'latex'`` -- compilation first tries tex -> dvi -> png and if an
-         error occurs then tries dvi -> ps -> pdf. This is slower than
-         ``'pdflatex'`` and known to be broken when overfull hbox are detected.
+      - ``'lualatex'`` -- compilation does ``tex`` -> ``pdf``
 
-    -  ``viewer`` -- string or ``None`` (default: ``None``): specify a viewer
+      - ``'latex'`` -- compilation first tries ``tex`` -> ``dvi`` -> ``png`` and if an
+        error occurs then tries ``dvi`` -> ``ps`` -> ``pdf``. This is slower than
+        ``'pdflatex'`` and known to be broken when overfull hbox are detected.
```
mkoeppe:
```suggestion
        ``'pdflatex'`` and known to be broken when overfull hboxes are detected.
```

src/sage/misc/latex.py
```diff
@@ -1829,20 +1721,17 @@ def view(objects, title='Sage', debug=False, sep='', tiny=False,
...
-    If the ``engine`` is either ``pdflatex`` or ``xelatex``,  it produces
-    a pdf file. Otherwise, it produces a dvi file, and if the program dvipng is
-    installed, it checks the dvi file by trying to convert it to a png
-    file.  If this conversion fails, the dvi file probably contains
-    some postscript special commands or it has other issues which
-    might make displaying it a problem; in this case, the file is
-    converted to a pdf file, which is then displayed.
+    If the ``engine`` is either ``'pdflatex'``, ``'xelatex'``, or ``'lualatex'``,
+    it produces a pdf file. Otherwise, it produces a dvi file, and if the program
+    dvipng is installed, it checks the dvi file by trying to convert it to a
```
mkoeppe:
```suggestion
    ``dvipng`` is installed, it checks the dvi file by trying to convert it to a
```

src/sage/misc/latex.py
```diff
@@ -664,49 +557,48 @@ def latex_extra_preamble():
 def _run_latex_(filename, debug=False, density=150, engine=None, png=False, do_in_background=False):
     """
     This runs LaTeX on the TeX file "filename.tex".  It produces files
-    "filename.dvi" (or "filename.pdf"` if engine is either ``pdflatex``
-    or ``xelatex``) and if ``png`` is ``True``, "filename.png".  If ``png``
-    is ``True`` and dvipng cannot convert the dvi file to png (because of
-    postscript specials or other issues), then dvips is called, and the
-    PS file is converted to a png file.
+    ``filename.dvi`` (or ``filename.pdf``` if engine is either ``'pdflatex'``,
```
mkoeppe:
```suggestion
    ``filename.dvi`` (or ``filename.pdf``` if ``engine`` is either ``'pdflatex'``,
```

mkoeppe:
> To test `view` via lualatex, for example,
> 
> ```
> sage: view('αβγ')
> ```
> 
> you should build sage on your local platform with this PR. You may need to install `texlive-full` before testing. It seems that `texlive` (2019) is not enough even for the `view` via pdftex.

Could you clarify if ``lualatex`` is now required for `view`? Does it fall back to ``pdflatex`` if ``lualatex`` is not available in the system?

build/pkgs/_recommended/dependencies
kwankyu: Yes. Done.

build/pkgs/sagemath_doc_pdf/dependencies
kwankyu: Fixed.

build/pkgs/texlive_luatex/SPKG.rst
kwankyu: Thanks. Done.

src/sage/misc/latex.py
kwankyu: Done.

src/sage/misc/latex.py
kwankyu: Done.

src/sage/misc/latex.py
kwankyu: Done.

src/sage/misc/latex.py
kwankyu: Done.

src/sage/misc/latex.py
kwankyu: Done.

kwankyu:
> > To test `view` via lualatex, for example,
> > ```
> > sage: view('αβγ')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > you should build sage on your local platform with this PR. You may need to install `texlive-full` before testing. It seems that `texlive` (2019) is not enough even for the `view` via pdftex.
> 
> Could you clarify if `lualatex` is now required for `view`? 

Yes. I changed the default `engine` to `lualatex` from `pdflatex` everywhere. 

If the user set `viewer` to `pdf`, `pdflatex` is used if the engine is `latex`. (I had changed `pdflatex` to `lualatex` in this case, but changed back to `pdflatex` in the last commit)

When the `engine` is `latex` and it fails to build pdf file via dvi file, it tries again with `pdflatex`. (I had changed `pdflatex` to `lualatex` in this case, but changed back to `pdflatex` in the last commit)

> Does it fall back to `pdflatex` if `lualatex` is not available in the system?

No.

kwankyu:
> Does it fall back to `pdflatex` if `lualatex` is not available in the system?

Do you think we need to implement this mechanism? 

Perhaps we may do that easily with the features mechanism.

src/sage/misc/latex.py
kwankyu: Done.

kwankyu:
> > Does it fall back to `pdflatex` if `lualatex` is not available in the system?
> 
> Do you think we need to implement this mechanism?
> 
> Perhaps we may do that easily with the features mechanism.

Done.

src/sage/misc/latex.py
```diff
@@ -605,8 +520,8 @@ def __init__(self, bb=False, delimiters=["(", ")"],
         self._option["matrix_column_alignment"] = matrix_column_alignment
         self._option["macros"] = ""
         self._option["preamble"] = ""
-        self._option["engine"] = "pdflatex"
-        self._option["engine_name"] = "LaTeX"
+        self._option["engine"] = default_engine()[0]
```
mkoeppe: does this get run on each startup of Sage?

kwankyu: Yes. Problem?

kwankyu: But as usual, the checks will not run until requested. That is, it won't delay the startup time.

kwankyu: Hmm. It does. I will find a fix.

kwankyu: Now done.

kwankyu: Would it make a sense that `sage_docbuild/conf.py` also uses the default engine detected in `sage/misc/latex.py`?

dimpase: only if it detects xelatex or lualatex,  otherwise we can't build docs.

I would say this detection should be delegated to ./configure

kwankyu: OK. Then I leave it as a later work to one who knows `./configure`.

kwankyu: Or no work. Perhaps it is not worth while to detect xelatex, as we are recommending to install lualatex for doc.

mkoeppe: Right, for sagemath_doc_pdf, it's declared as a dependency; so nothing to do.

mkoeppe:
Test failure https://github.com/sagemath/sage/actions/runs/7238519979/job/19719457454?pr=36861#step:14:17
```
sage -t --random-seed=26151039727556449658509468572536179411 src/sage/misc/latex.py
File "src/sage/misc/latex.py", line 1527, in sage.misc.latex.Latex.engine
Failed example:
    latex.engine()
Expected:
    'lualatex'
Got:
    'latex'
```

mkoeppe (approved):

mkoeppe:
I think this is ready for merging

kwankyu:
Yes. Thank Matthias. Better to be further tested out there.

Thank Dima for suggesting lualatex. I am happy with it. Perhaps I have unhappy memories with it from when it was still unstable.

vbraun:
merge conflict

kwankyu:
Fixed.

vbraun:
I'm getting
```
[xindy-none] Error: xindy is a dummy package and 
[xindy-none] cannot be installed using the Sage distribution.
make[2]: *** [Makefile:3570: xindy-SAGE_LOCAL-no-deps] Error 1
make[1]: *** [Makefile:3570: /home/release/Sage/local/var/lib/sage/installed/xindy-none] Error 2
```

kwankyu:
The message is correct. Under what commands, do you get the message?

mkoeppe:
@vbraun PDF building has now increased system package requirements

kwankyu:
On Ubuntu, PDF building requires 
```
sudo apt install texlive texlive-luatex fonts-freefont-otf xindy
```

vbraun:
OK thanks for telling me on the ticket, but can this perhaps be added to the installation manual? The error message is also hardy self-explanatory.

dimpase:
Does ./configure tell you to install these?

kwankyu:
>  but can this perhaps be added to the installation manual? 

Already added. See the recommended package list after

https://deploy-preview-36861--sagemath-tobias.netlify.app/html/en/installation/source#:~:text=client%20pkg%2Dconfig-,For%20all%20users%2C%20we%20recommend%20that%20you%20install%20the%20following%20system%20packages%2C%20which%20provide%20additional%20functionality%20and%20cannot%20be%20installed%20by%20Sage%3A,-%24%20sudo%20apt

Packages required for doc build, such as "texlive", are only recommended.

kwankyu:
But I agree that there is a problem here.

We tell users who want to install sage from source to `make` sage, and the `make` builds doc, but building doc requires "the recommended" packages including texlive, texlive-luatex, etc. Hence "the recommended" packages are required by default. 

If we want to keep texlive only "recommended", perhaps `make` should not build doc...

dimpase:
`make` does not build pdf docs. Is xindy needed for html too?

kwankyu:
> OK thanks for telling me on the ticket, but can this perhaps be added to the installation manual? 

In addition, I may write some guide later in the release tour. Would that be enough?

kwankyu:
> `make` does not build pdf docs. Is xindy needed for html too?

No.

kiwifb:
I am currently getting one error in building the pdf doc with this
```
$ tail -20 /home/portage/sci-mathematics/sage-doc-9999/work/build_doc/latex/en/reference/coding/coding.log

LaTeX Warning: Hyper reference `sage/coding/decoder:sage.coding.decoder.Decoder
.decode_to_code' on page 28 undefined on input line 2667.


Overfull \hbox (11.0582pt too wide) in paragraph at lines 2666--2669
[]\TU/FreeSerif(0)/m/n/10 This is a de-fault im-ple-mentmen-ta-tion, which as-su
mes that the method \TU/FreeMono(0)/m/it/10 de-code_to_code() \TU/FreeSerif(0)/
m/n/10 has been im-ple-mented,
 []


LaTeX Warning: Hyper reference `sage/coding/decoder:sage.coding.decoder.Decoder
.message_space' on page 28 undefined on input line 2684.

[28]
! error:  (nodes): fuzzy token cleanup in whatsit node with type whatsit and sub
type 28
!  ==> Fatal error occurred, no output PDF file produced!
```
Am I missing some package?

dimpase:
is it with lualatex ?

kiwifb:
Most definitely lualatex.

dimpase:
do you see this warning in CI logs too? I.e., is the error consequence of the warning? 

might potentially be some locale nonsense...

kiwifb:
I see some of the warnings in the CI, with slightly different line number. Although I do not see the `Overfull` bit in that location, could be A4 vs Letter for that one. I should dig my log to see if there is some more meaningful clues earlier. Locale non-sense could be a possibility, but I am thinking a missing font or texlive module that is not obvious.

dimpase:
maybe you can post various .log files?

kiwifb:
A bit of the folder where things fails. Includes the full `coding.log` and also `coding.tex` for those who want to inspect it.
http://202.36.178.9/sage/coding.tar.xz

kwankyu:
Sorry for not responding. I am out of town and come back after holidays.

kwankyu:
> I am currently getting one error in building the pdf doc with this
> 
> ```
> $ tail -20 /home/portage/sci-mathematics/sage-doc-9999/work/build_doc/latex/en/reference/coding/coding.log
> 
> LaTeX Warning: Hyper reference `sage/coding/decoder:sage.coding.decoder.Decoder
> .decode_to_code' on page 28 undefined on input line 2667.
> 
> 
> Overfull \hbox (11.0582pt too wide) in paragraph at lines 2666--2669
> []\TU/FreeSerif(0)/m/n/10 This is a de-fault im-ple-mentmen-ta-tion, which as-su
> mes that the method \TU/FreeMono(0)/m/it/10 de-code_to_code() \TU/FreeSerif(0)/
> m/n/10 has been im-ple-mented,
>  []
> 
> 
> LaTeX Warning: Hyper reference `sage/coding/decoder:sage.coding.decoder.Decoder
> .message_space' on page 28 undefined on input line 2684.
> 
> [28]
> ! error:  (nodes): fuzzy token cleanup in whatsit node with type whatsit and sub
> type 28
> !  ==> Fatal error occurred, no output PDF file produced!
> ```
> 
> Am I missing some package?

I don't know. I don't see any hint from your logs about what went wrong. If you still have it, I would just recommend to upgrade your texlive or reinstall gnu free fonts and wish a good luck.

kiwifb:
Yes, in the end, texlive from the gentoo main tree proved too old. Moving to texlive 2023 fixed the issue.

kwankyu:
Good to hear that. texlive 2023 seems to also include `xindy` while texlive 2019 does not. If we could recommend texlive 2023 instead of just any texlive, then we can get rid of `xindy` from the list of recommended packages! It seems that the free fonts pack is in the same situation...

kiwifb:
I was on texlive 2021 for info.

chriswuthrich:
Now that I see this, out of curiosity, doesn't this slow down all LaTeX compilations a lot? LuLaTeX is much slower than pdflatex.

kwankyu:
That was noted here: https://groups.google.com/g/sage-devel/c/tG2LK6Jvw0I/m/TlMGEkCwAAAJ

dimpase:
lualatex, in its default configuration, is slower to start up. There are, however, various ways to speed it up.
E.g. there is a way to build a custom format, there is a Lua JIT option, there is luametatex (the current actively developed luatex branch).
Perhaps we can also use some Lua in Sage to speed typesetting things up (?).

All these need extra investigation. The big plus here is that we finally sorted our Unicode troubles.

PR #36862, opened 2023-12-11
Fix one doctest for giac 1.9.0-73

tornaria:
With giac 1.9.0-73 the output of a `symbolic_sum()` test changes. Note that both answers are correct, and it's not clear that one is "better" than the other...

I managed to find a way to avoid trouble (by sending the output through `factor()`) so the output is the same with latest giac and earlier giac.

However, fixing doctests keeps getting harder (I mean: making sure the same doctest works for different releases of dependencies). I don't have a good idea on how to improve the situation.

### :memo: Checklist


mkoeppe (approved):
LGTM

PR #36351, opened 2023-09-28
`gmp` 6.3.0, `mpc` 1.3.1, `mpfr` 4.2.1, `mpfrcx` 0.6.3, `gmpy2` 2.2.0a1

mkoeppe:
The unreleased 2.2.x series has work on support for Python 3.12

https://github.com/aleaxit/gmpy

The new version makes it necessary to tighten the requirements for MPC and MPFR.
We do the version check without pkg-config as before; changing it to pkg-config is #31348.

(The upgrade of the packages was previously done in #36775, but has now been merged here.)

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36868, opened 2023-12-12
Update python3 spkg-configure.m4 to allow system Python 3.12.x

dimpase:
When the necessary package upgrades (see #36181) are completed, we can allow system Python 3.12.x.

### DIFF
--- a/build/pkgs/python3/spkg-configure.m4
+++ b/build/pkgs/python3/spkg-configure.m4
@@ -1,8 +1,8 @@
 SAGE_SPKG_CONFIGURE([python3], [
    m4_pushdef([MIN_VERSION],               [3.9.0])
    m4_pushdef([MIN_NONDEPRECATED_VERSION], [3.9.0])
-   m4_pushdef([LT_STABLE_VERSION],         [3.12.0])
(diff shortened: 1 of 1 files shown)


dimpase:
> Rule: First port it, then claim to support it

No idea what you mean by "port" here. It's running other people Sage's.

mkoeppe:
A key package, gmpy2, does not yet support Python 3.12. That's #36351

dimpase:
This is silly. It must be that @tornaria bought you a lot of beer so that you let that PR of his through, no? So now what, Fedora people must buy you more beer so that you let this PR through, too ?

mkoeppe:
@dimpase This is unwelcome speech.

dimpase:
look, you obviously made a mistake by approving a PR with the title "Support Python 3.12" but not actually allowing ./configure to accept it.

Now instead of admitting it 1st of all, you come up with tasks for me to perform before you might allow an appropriate correction. Thus sucks.

mkoeppe:
Read https://github.com/sagemath/sage/pull/36407#issuecomment-1852740639

mkoeppe:
The commit is merged into #36181, closing.

PR #36873, opened 2023-12-13
add is_trivial method for groups

mantepse:
We add a method to check whether a permutation group is trivial, that is, consists only of the identity element.

### DIFF
--- a/src/sage/groups/perm_gps/permgroup.py
+++ b/src/sage/groups/perm_gps/permgroup.py
@@ -1365,6 +1365,29 @@ class PermutationGroup_generic(FiniteGroup):
         """
         return len(self.gens())
 
+    def is_trivial(self):
+        r"""
+        Return ``True`` if this group is the trivial group.
+
+        A permutation group is trivial, if it consists only of the
+        identity element, that is, if it has no generators.
+
+        EXAMPLES::
+
+            sage: G = PermutationGroup([], domain=["a", "b", "c"])
+            sage: G.is_trivial()
+            True
+            sage: SymmetricGroup(0).is_trivial()
+            True
+            sage: SymmetricGroup(1).is_trivial()
+            True
+            sage: SymmetricGroup(2).is_trivial()
+            False
+            sage: DihedralGroup(1).is_trivial()
+            False
+        """
+        return not self._gens or (len(self._gens) == 1 and self._gens[0].is_one())
+    
     @cached_method
     def one(self):
         """
--- a/src/sage/groups/perm_gps/permgroup.py
+++ b/src/sage/groups/perm_gps/permgroup.py
@@ -1371,7 +1371,7 @@ class PermutationGroup_generic(FiniteGroup):
 
         A permutation group is trivial, if it consists only of the
         identity element, that is, if it has no generators or only
-        the trivial generator.
+        trivial generators.
 
         EXAMPLES::
 
@@ -1387,9 +1387,16 @@ class PermutationGroup_generic(FiniteGroup):
             sage: DihedralGroup(1).is_trivial()
             False
 
+        TESTS::
+
+            sage: G = PermutationGroup([[], []], canonicalize=False)
+            sage: G.ngens()
+            2
+            sage: G.is_trivial()
+            True
         """
-        return not self._gens or (len(self._gens) == 1 and self._gens[0].is_one())
-    
+        return all(g.is_one() for g in self._gens)
+
     @cached_method
     def one(self):
         """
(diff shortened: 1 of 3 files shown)


src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1365,6 +1365,29 @@ def ngens(self):
         """
         return len(self.gens())
 
+    def is_trivial(self):
+        r"""
+        Return ``True`` if this group is the trivial group.
+
+        A permutation group is trivial, if it consists only of the
+        identity element, that is, if it has no generators.
```
jukkakohonen: This method is a very welcome addition, thanks. The code is good as far as I understand. A small complaint about the docstring: In Sage a trivial group may in fact have a (trivial) generator. To my surprise, here we specify no generators but end up having one generator anyway:
```
sage: PermutationGroup([],domain=[1,2]).gens()
((),)
```
If the docstring mentions generators (and I think it is a good idea, since it clarifies to the user what the code will actually be checking), it should mention this surprising possibility.

(I believe/hope it is not possible to make a group with *several* trivial generators, at least not easily. If that is a possibility, we would need to check that *all* generators are trivial.)

mantepse: It is, actually.

jukkakohonen: Oh well, it is possible and not that difficult...
```
sage: P=PermutationGroup([(),()], canonicalize=False)
sage: P.gens()
((), ())
sage: P.order()
1
```

A human would likely not create such a group on purpose, but I imagine some code might. Anyway it is possible. Mathematically of course this is a trivial group, so Sage should recognize it as one.

mantepse: crossed.

jukkakohonen: Looks good now!

videlec:
Why is this new method restricted to permutation groups and not made available for groups in general? To my mind, such code would better be in the `Groups` or possibly `Monoids` category.

mantepse:
> Why is this new method restricted to permutation groups and not made available for groups in general? To my mind, such code would better be in the `Groups` or possibly `Monoids` category.

We certainly need finitely generated for this code, although, we can certainly use `cardinality` for a generic implementation.

It seems that the same code might work well for `MatrixGroup`, I can add it there, if you want to.

I don't know the category code well enough to add generic code, but if you help me, I'd do it.

jukkakohonen:
> we can certainly use `cardinality` for a generic implementation.

Interestingly, `cardinality` can fail in some corner cases, e.g. a special linear group of degree 1:
```
sage: SL(1, QQ).cardinality()
...
RecursionError: maximum recursion depth exceeded
```
(Should probably file an issue on this.)

dimpase:
> > we can certainly use `cardinality` for a generic implementation.
> 
> Interestingly, `cardinality` can fail in some corner cases, e.g. a special linear group of degree 1:
> 
> ```
> sage: SL(1, QQ).cardinality()
> ...
> RecursionError: maximum recursion depth exceeded
> ```
> 
> (Should probably file an issue on this.)

yes, this is definitely a bug (it should return NotImplementedError I think). Please file an issue

jukkakohonen:
> Please file an issue

#36876

mantepse:
I am not sure what to do in the case of finitely presented group.  Being trivial is undecidable (https://en.wikipedia.org/wiki/Adian%E2%80%93Rabin_theorem), but of course we can (and should) attempt to decide it.  Computing the cardinality seems to be wasteful, however.

dimpase:
> I am not sure what to do in the case of finitely presented group. Being trivial is undecidable (https://en.wikipedia.org/wiki/Adian%E2%80%93Rabin_theorem), but of course we can (and should) attempt to decide it. Computing the cardinality seems to be wasteful, however.

Throw NotImplementedError ?

mantepse:
> > I am not sure what to do in the case of finitely presented group. Being trivial is undecidable (https://en.wikipedia.org/wiki/Adian%E2%80%93Rabin_theorem), but of course we can (and should) attempt to decide it. Computing the cardinality seems to be wasteful, however.
> 
> Throw NotImplementedError ?

Isn't that worse to checking cardinality?

mantepse:
Done, please review.

dimpase:
> It appears that automatic actions triggering is working for you. So it is possible that you not seeing the "Approve and run" button on other PRs is due to your browser/auth, not the setup on GitHub.

unless someone pushed "Approve and run" button here. It was not me, anyone else?

dimpase:
For solvable groups the world problem is decidable. OK, we can talk about this somewhere later

dimpase (approved):
lgtm

mantepse:
sorry, I pressed the wrong button

PR #36877, opened 2023-12-13
Improve code style for matrix2.pyx file

RuchitJagodara:
- Correct pycodestyle and remove check condition for columns in prod_of_row_sums() function
- Improve code style and remove an unwanted if condition in restrict() function
- Improve code style of matrix2.pyx file
- Improve code style for matrix2.pyx

This patch improves the code style as mentioned in the sage documentation and removes 
some unnecessary lines which were commented before and was not removed.

### :memo: Checklist

### :hourglass: Dependencies

None


fchapoton:
Adding the missing whitespace after commas is much too intrusive to be done systematically, and in particular in huge files such as the one touched here. To say that otherwise, we are not doing that, except just one in passing when modifying the code for other reasons.

RuchitJagodara:
Hmm.. got it. Actually, I was modifying the code of this file when I saw this, so I thought it would be better to correct that.

RuchitJagodara:
@fchapoton, can you please help me find out what the problem is with the build and test check?

kwankyu (approved):
LGTM.

kwankyu:
> @fchapoton, can you please help me find out what the problem is with the build and test check?

There is no problem. Ignore conda tests.

RuchitJagodara:
Ok, thank you !

PR #36881, opened 2023-12-14
Fix cardinality of special linear group

### LINKED ISSUE #35490: Infinite recursion in `LinearMatrixGroup_generic`

### Did you read the documentation and troubleshoot guide?

### Environment

```markdown
- **OS**: ubuntu
- **Sage Version**: 9.5
```

### Steps To Reproduce

In some situation, the special linear groups `sage.groups.matrix_gps.linear.LinearMatrixGroup_generic` is  such that the two methods `order` (inherited from `sage/groups/group.pyx`) and `cardinality` (inherited from `sage/categories/finite_groups.py`) are calling each other
```python
q = 7
FqT.<T> = GF(q)[]
N = T^2+1
FqTN = QuotientRing(FqT, N*FqT)
S = SL(2,FqTN)
S.is_finite()
```
In particular it breaks `S.is_finite()` which should work out of the box given that the category of `S` is the category of finite groups (obtained by `S.category()`).

### Expected Behavior

```python
True
```

### Actual Behavior

```python
<ipython-input-11-397d936442b9> in <module>
----> 1 S.is_finite()

.../sage/groups/group.pyx in sage.groups.group.Group.is_finite (build/cythonized/sage/groups/group.c:3068)()
    188             NotImplementedError
    189         """
--> 190         return self.order() != infinity
    191 
    192     def is_multiplicative(self):

.../sage/groups/group.pyx in sage.groups.group.Group.order (build/cythonized/sage/groups/group.c:2924)()
    171         """
    172         try:
--> 173             return self.cardinality()
    174         except AttributeError:
    175             raise NotImplementedError

.../sage/categories/finite_groups.py in cardinality(self)
    106                 return self._cardinality_from_iterator()
    107             else:
--> 108                 return o()
    109 
    110         def some_elements(self):

... last 2 frames repeated, from the frame below ...

.../sage/groups/group.pyx in sage.groups.group.Group.order (build/cythonized/sage/groups/group.c:2924)()
    171         """
    172         try:
--> 173             return self.cardinality()
    174         except AttributeError:
    175             raise NotImplementedError

RecursionError: maximum recursion depth exceeded while calling a Python object
```

### Additional Information

_No response_

### LINKED ISSUE #36876: Cardinality fails in small special linear groups
### Steps To Reproduce

Using SageMath version 10.3.beta1, Release Date: 2023-12-10, run:
```
sage: SL(1, QQ).cardinality()
```

### Expected Behavior

Mathematically I would expect it to be `1`, in fact for any special linear group of degree 1 regardless of the field, because the only $1 \times 1$ matrix with determinant one is $[1]$, where $1$ is the multiplicative unity.

If not implemented, at least it should raise a `NotImplementedError`.

### Actual Behavior

It goes into an infinite recursion.

```
sage: SL(1, QQ).cardinality()
...
RecursionError: maximum recursion depth exceeded
```

### Additional Information

Same happens with `SL(1,RR)`, and also with `is_finite`.

### Environment

```markdown
- **OS**: Ubuntu 20.04.5 LTS
- **Sage Version**: SageMath version 10.3.beta1, Release Date: 2023-12-10

Also happens on SageMath version 9.0, Release Date: 2020-01-01, with a slightly different error message `RecursionError: maximum recursion depth exceeded while calling a Python object`.
```

### Checklist

RuchitJagodara:
- Fixed infinite recursion of the cardinality and order functions
- Improved pycodestyle
- Add more doctests to cover all cases

This patch fixes #36876 and fixes #35490. In this PR, I have implemented a method to fix the
answer given by cardinality function for finite special linear groups
(before it was going into infinite recursion). This also fixes the answer given by is_finite()
function for finite special linear groups.

### :memo: Checklist

### :hourglass: Dependencies

None

### DIFF
--- a/src/sage/categories/finite_groups.py
+++ b/src/sage/categories/finite_groups.py
@@ -101,12 +101,19 @@ class FiniteGroups(CategoryWithAxiom):
                 sage: G.cardinality()
                 384
             """
+
+            # Check whether the object has _deg attribute or not 
+            # Note for developers - It will have _deg attribute if it is derived from  MatrixGroup_generic class
+            # And if it has _deg attribute then return the value of that attribute
+            if hasattr(self, '_deg'):
+                return self._deg
+
+            # If object does not have _deg attribute then try to find its order with the help of order function.
             try:
                 o = self.order
+                return o()
             except AttributeError:
                 return self._cardinality_from_iterator()
-            else:
-                return o()
 
         def some_elements(self):
             """
--- a/src/sage/categories/finite_groups.py
+++ b/src/sage/categories/finite_groups.py
@@ -102,7 +102,7 @@ class FiniteGroups(CategoryWithAxiom):
                 384
             """
 
-            # Check whether the object has _deg attribute or not 
+            # Check whether the object has _deg attribute or not
             # Note for developers - It will have _deg attribute if it is derived from  MatrixGroup_generic class
             # And if it has _deg attribute then return the value of that attribute
             if hasattr(self, '_deg'):
--- a/src/sage/categories/finite_groups.py
+++ b/src/sage/categories/finite_groups.py
@@ -101,10 +101,6 @@ class FiniteGroups(CategoryWithAxiom):
                 sage: G.cardinality()
                 384
             """
-            # If the group is special group and finite then there is only one possibility that the degree
-            # and cardinality will be one.
-            if hasattr(self,'_special') and self._special==True:
-                return 1
             try:
                 o = self.order
                 return o()
--- a/src/sage/categories/finite_groups.py
+++ b/src/sage/categories/finite_groups.py
@@ -102,13 +102,10 @@ class FiniteGroups(CategoryWithAxiom):
                 384
             """
 
-            # Check whether the object has _deg attribute or not
-            # Note for developers - It will have _deg attribute if it is derived from  MatrixGroup_generic class
-            # And if it has _deg attribute then return the value of that attribute
+            # Note for developers - self will have _deg attribute if it is derived from  MatrixGroup_generic class
             if hasattr(self, '_deg'):
                 return self._deg
 
-            # If object does not have _deg attribute then try to find its order with the help of order function.
             try:
                 o = self.order
                 return o()
--- a/src/sage/categories/finite_groups.py
+++ b/src/sage/categories/finite_groups.py
@@ -101,11 +101,10 @@ class FiniteGroups(CategoryWithAxiom):
                 sage: G.cardinality()
                 384
             """
-
-            # Note for developers - self will have _deg attribute if it is derived from  MatrixGroup_generic class
-            if hasattr(self, '_deg'):
-                return self._deg
-
+            # If the group is special group and finite then there is only one possibility that the degree
+            # and cardinality will be one.
+            if hasattr(self,'_special') and self._special==True:
+                return 1
             try:
                 o = self.order
                 return o()
--- a/src/sage/categories/finite_groups.py
+++ b/src/sage/categories/finite_groups.py
@@ -101,11 +101,9 @@ class FiniteGroups(CategoryWithAxiom):
                 sage: G.cardinality()
                 384
             """
-            try:
-                o = self.order
-                return o()
-            except AttributeError:
-                return self._cardinality_from_iterator()
+            if hasattr(self, 'order'):
+                return self.order()
+            return self._cardinality_from_iterator()
 
         def some_elements(self):
             """
--- a/src/sage/groups/matrix_gps/linear.py
+++ b/src/sage/groups/matrix_gps/linear.py
@@ -277,6 +277,18 @@ def SL(n, R, var='a'):
         sage: S = SL(2,FqTN)
         sage: S.is_finite()
         True
+
+        Check if is_finite() function is giving correct answer for the special linear group which is not finite::
+
+        sage: SL(2, QQ).is_finite()
+        False
+
+        Check if the cardinality or order of infinite special linear group is +Infinity or not::
+
+        sage: SL(2, QQ).cardinality()
+        +Infinity
+        sage: SL(2, QQ).order()
+        +Infinity
     """
     degree, ring = normalize_args_vectorspace(n, R, var='a')
     try:
--- a/src/sage/groups/matrix_gps/linear.py
+++ b/src/sage/groups/matrix_gps/linear.py
@@ -256,6 +256,27 @@ def SL(n, R, var='a'):
 
         sage: groups.matrix.SL(2, 3)
         Special Linear Group of degree 2 over Finite Field of size 3
+
+        Check if :trac:`36876` is fixed: The cardinality or order of a special linear group with degree=1 over any field is 1::
+
+        sage: SL(1, QQ).cardinality()
+        1
+        sage: SL(1, RR).cardinality()
+        1
+        sage: SL(1, QQ).order()
+        1
+        sage: SL(1, CC).order()
+        1
+
+        Check if :trac:`35490` is fixed: The special linear group with degree=1 is finite::
+
+        sage: q = 7
+        sage: FqT.<T> = GF(q)[]
+        sage: N = T^2+1
+        sage: FqTN = QuotientRing(FqT, N*FqT)
+        sage: S = SL(2,FqTN)
+        sage: S.is_finite()
+        True
     """
     degree, ring = normalize_args_vectorspace(n, R, var='a')
     try:
--- a/src/sage/groups/matrix_gps/linear.py
+++ b/src/sage/groups/matrix_gps/linear.py
@@ -323,36 +323,34 @@ class LinearMatrixGroup_generic(NamedMatrixGroup_generic):
         Check if :trac:`36876` is fixed::
             sage: SL(1, QQ).order()
             1
-            sage: SL(1, ZZ).cardinality()
-            1
+            sage: SL(2, ZZ).cardinality()
+            +Infinity
 
         Check if :trac:`35490` is fixed::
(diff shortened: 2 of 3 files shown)


RuchitJagodara:
@jukkakohonen and @videlec  can you please review this PR ?

src/sage/categories/finite_groups.py
```diff
@@ -101,12 +101,19 @@ def cardinality(self):
                 sage: G.cardinality()
                 384
             """
+
+            # Check whether the object has _deg attribute or not
```
dimpase: this line is not needed - it's clear from the code

RuchitJagodara: Done ! I have removed all unnecessary comments.

src/sage/categories/finite_groups.py
```diff
@@ -101,12 +101,16 @@ def cardinality(self):
...
                 384
             """
+
+            # Note for developers - self will have _deg attribute if it is derived from  MatrixGroup_generic class
+            if hasattr(self, '_deg'):
+                return self._deg
+
             try:
                 o = self.order
+                return o()
```
mantepse: This looks wrong.  The idea of the try except is to check whether `self` has an `order` method.  However, this line of code is from 2010, so it might well be that things could be done better now.

src/sage/categories/finite_groups.py
```diff
@@ -101,12 +101,16 @@ def cardinality(self):
                 sage: G.cardinality()
                 384
             """
+
+            # Note for developers - self will have _deg attribute if it is derived from  MatrixGroup_generic class
+            if hasattr(self, '_deg'):
+                return self._deg
```
mantepse: Why do you return the degree of the group for its cardinality?

RuchitJagodara: Thank you for pointing that out! I just saw that it is a parent method for all finite groups, so doing this won't work. But I propose implementing a cardinality function specifically for the special linear group, that will fix all the issues.

RuchitJagodara: So for now, I am converting this PR to a draft until I implement the method.

mantepse:
Have you located the reason for the infinite recursion?

RuchitJagodara:
Yes, the is_finite() function is calling this cardinality function, and it seems that it is calling the order function (Considering that other classes have their own order function). I didn't find any special function that calculates the cardinality of these special linear groups. So, maybe it is not implemented yet, so it is calling the order function of some class from which it is derived, but that function is calling back this cardinality function.

dimpase:
Orders of special linear groups over fields  are known, you can just implement formulae.

Less sure for more general cases

RuchitJagodara:
> Orders of special linear groups over fields are known, you can just implement formulae.

I checked and found that order function is implemented only for gap-based groups and there is only one function available for other groups which is going in infinite recursion. Here below is a example of a gap-based special linear group which is working perfect.

```python
sage: F = FiniteField(3)
sage: SL(2,F).order()
24
```

But when it comes to other groups which are not gap-based then it is going in infinite recursion like below 

```python
sage: q = 7
....: FqT.<T> = GF(q)[]
....: N = T^2+1
....: FqTN = QuotientRing(FqT, N*FqT)
....: S = SL(2,FqTN)
....: S.order()
RecursionError                            Traceback (most recent call last)
...
```
This is happening because in below code it is first trying to make an object of LinearMatrixGroup_gap but if it fails then it is simply making an object of LinearMatrixGroup_generic and that class does not have properties of gap and because of that it is not able to calculate cardinality of the gorup.
https://github.com/sagemath/sage/blob/879ac37acaaf2a564c15ea4d1d50546830a25cdd/src/sage/groups/matrix_gps/linear.py#L308-L316

mantepse:
> But when it comes to other groups which are not gap-based then it is going in infinite recursion like below
> 
> ```python
> sage: q = 7
> ....: FqT.<T> = GF(q)[]
> ....: N = T^2+1
> ....: FqTN = QuotientRing(FqT, N*FqT)
> ....: S = SL(2,FqTN)
> ```

Should we be able to iterate over such a group?

dimpase:
There should be a special `.order()` method for `SL`. For finite fields $\mathbb{F}_q$ one has

$|SL(n,q)|={\frac{1}{q-1}} \Pi_{k=0}^{n-1} (q^n-q^k).$

Here `FqTN` is isomorphic to $\mathbb{F}_{49}$, so just use the forrmula, with $n=2$, $q=49$.

RuchitJagodara:
I have implemented the order function for linear matrix groups. I discovered that GAP does not support linear matrix groups over the rational field (and some other fields), which was the main cause of the error. Consequently, I attempted to address this issue in GAP, but it proved to be very non-trivial. Therefore, I implemented the function in Sage. Additionally, I have another question: Why do we use functions from GAP when we can implement all of those functions from scratch in Sage as well? In this case, we would have the flexibility to construct functions according to our needs. I understand it may take considerable time, but we can proceed step by step. The same can be applied to Magma, which was also causing some errors, as can be found in the [Discussion](https://github.com/sagemath/sage/issues/36410#issuecomment-1763476272) and in issue #36841.

RuchitJagodara:
@dimpase and @mantepse, can you please review this pr...

src/sage/groups/matrix_gps/linear.py
```diff
@@ -307,3 +307,53 @@ def _check_matrix(self, x, *args):
...
+        Check if :trac:`35490` is fixed::
+            sage: q = 7
+            sage: FqT.<T> = GF(q)[]
+            sage: N = T^2+1
+            sage: FqTN = QuotientRing(FqT, N*FqT)
+            sage: S = SL(2,FqTN)
+            sage: S.is_finite()
+            True
+            sage: S.order()
+            117600
```
mantepse, line 342: It would be good to have tests for the infinite cases.

src/sage/groups/matrix_gps/linear.py
```diff
@@ -307,3 +307,53 @@ def _check_matrix(self, x, *args):
...
+            1
+            sage: SL(1, ZZ).cardinality()
+            1
+
+        Check if :trac:`35490` is fixed::
+            sage: q = 7
+            sage: FqT.<T> = GF(q)[]
+            sage: N = T^2+1
+            sage: FqTN = QuotientRing(FqT, N*FqT)
+            sage: S = SL(2,FqTN)
```
mantepse: Please add a whitespace after the comma

src/sage/groups/matrix_gps/linear.py
```diff
@@ -307,3 +307,53 @@ def _check_matrix(self, x, *args):
...
+            if self._special:
+                return reduce(operator.mul, (q**n - q**i for i in range(0, n)), 1) / (q-1)
+            return reduce(operator.mul, (q**n - q**i for i in range(0, n)), 1)
+
+        if self._special:
+            if n == 1:
+                return 1
+            return Infinity
+
+        return Infinity
```
mantepse, line 355: I think that 352-357 can be simplified to
```
if self._special and n == 1:
    return 1
return Infinity
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -307,3 +307,53 @@ def _check_matrix(self, x, *args):
...
+        import operator
+        from sage.rings.infinity import Infinity
+
+        n = self.degree()
+
+        if self.base_ring().is_finite():
+            q = self.base_ring().order()
+            if self._special:
+                return reduce(operator.mul, (q**n - q**i for i in range(0, n)), 1) / (q-1)
+            return reduce(operator.mul, (q**n - q**i for i in range(0, n)), 1)
```
mantepse: In 349 and 350 you can use `prod`, that should be more readable.  Also, `range(0, n)` is the same as `range(n)`.

It might also make sense to first compute `prod(q**n - q**i for i in range(n))`, and only divide by `q-1` if `self._special`.

src/sage/groups/matrix_gps/linear.py
```diff
@@ -307,3 +307,50 @@ def _check_matrix(self, x, *args):
...
+            sage: N = T^2+1
+            sage: FqTN = QuotientRing(FqT, N*FqT)
+            sage: S = SL(2, FqTN)
+            sage: S.is_finite()
+            True
+            sage: S.order()
+            117600
+        """
+        from sage.rings.infinity import Infinity
+        from sage.all import prod
```
mantepse: This should be `from sage.misc.misc_c import prod`, and it might be better (if it works) to move both import statements to line 69 or so.

mantepse (changes requested):

mantepse (approved):
Thank you!

src/sage/groups/matrix_gps/linear.py
```diff
@@ -307,3 +309,49 @@ def _check_matrix(self, x, *args):
...
+            sage: S.order()
+            117600
+        """
+        n = self.degree()
+
+        if self.base_ring().is_finite():
+            q = self.base_ring().order()
+            ord = prod(q**n - q**i for i in range(n))
+            if self._special:
+                return ord / (q-1)
```
grhkm21, line 350: This should be `//`.

mantepse: Possibly yes, but it doesn't look serious to me.  Since this ticket is about to be merged, please open a new one!

grhkm21: Yes but also no, rational numbers behave weirdly, e.g. `is_prime(7 / 1)` is `False`. But I will open a new PR

RuchitJagodara: @grhkm21, have you made another PR for the changes you suggested? If not then, should I create a new PR for this?

PR #36884, opened 2023-12-14
Method braid of class Link loops when the Link contains loops

soehms:
The following link `L` is ambient isotopic to the treefoil knot but has one additional loop.

grafik

```
sage: L = Link([[1, 7, 2, 6], [3, 1, 4, 8], [5, 5, 6, 4], [7, 3, 8, 2]])
sage: L.regions()`
[[-5], [8, 4, 6, 2], [7, -2], [3, -8], [1, -6, 5, -4], [-1, -3, -7]]
sage: L.seifert_circles()
[[5], [1, 7, 3], [2, 8, 4, 6]]
``` 

Invoking `L.braid()` does not terminate. The problem already occurs on the first application of a Vogel move which leads to an extended `pd_code = [[10, 7, 2, 6], [3, 1, 4, 8], [11, 5, 6, 4], [7, 3, 8, 2], [12, 9, 5, 1], [11, 9, 12, 10]]`. A diagram of the corresponding link looks like this:

grafik

The labels `C1`, `C2 = [3, 1, 4, 8]`, `D = [11, 9, 12, 10]` and `E = [12, 9, 5, 1]` correspond to the variables in the code and the numbers to the edges. In case of `C1`the correct list of edges would be `[5, 11, 6, 4]` but the code produces `[11, 5, 6, 4]`. This is caused by usage of the `index` method of the `list` class which always returns the first occurrence of a value in the list, i.e. by the line:

```
C2[C2.index(b)] = newedge + 2
```

To fix it I replace the method by a local function:

```
        def idx(cross, edge):
            r"""
            Return the index of an edge in a crossing taking loops into account.
            A loop appears as an edge which occurs twice in the crossing.
            In all cases the second occurrence is the correct one needed in
            the Vogel algorithm.
            """
            i = cross.index(edge)
            if cross.count(edge) > 1:
                return cross.index(edge, i+1)
            else:
                return i
```

> In all cases the second occurrence is the correct one needed in the Vogel algorithm.

The according argumentation goes as follows:

If the loop is clockwise oriented the edge has a positive sign in the region, else a negative sign. In the first case the outgoing side of the edge always occurs after the incoming one in the PD-code (edge 1 in the first picture below and edge 2 in the second). In the second case the incoming side of the edge always occurs after the outgoing one in the PD-code (edge 2 in the first picture below and edge 1 in the second).

grafik grafik

Thus, this corresponds to the fact that in the first case `a` and `b` leave the crossings `C1` and `C2` (in the original link) whereas they arrive there in the second case.

Another way to fix the issue would be to remove all loops from the diagram before starting the Vogel algorithm. But this would harm regular isotopy which might violate the expectation of the user. Thus, I implement this as an optional feature. This gives the user the possibility to obtain a braid of smaller index if he is fine with ambient isotopy.

### :memo: Checklist


soehms:
None of the failures from the [bot-logfile](https://github.com/sagemath/sage/actions/runs/7219053632/job/19669427934?pr=36884#step:17:19944) appear to be related to this PR.

src/sage/knots/link.py
```diff
@@ -648,7 +663,20 @@ def braid(self):
             sage: A = Link([[[1, 2, -2, -1, -3, -4, 4, 3]], [1, 1, 1, 1]])
             sage: A.braid()
             s0*s1*s2*s3
+
+        Check that `PR 36884 <https://github.com/sagemath/sage/pull/36884>`__ is solved::
+
```
tscrim:
```suggestion
        Check that :issue:`36884` is solved::

```

src/sage/knots/link.py
```diff
@@ -2362,6 +2403,48 @@ def regions(self):
             regions.append(region)
         return regions
 
+    def remove_loops(self):
+        r"""
+        Return an ambient isotopic link in which all loops are removed.
+
+        EXAMPLES::
+
+          sage: b = BraidGroup(4)((3, 2, -1, -1))
```
tscrim:
```suggestion
        EXAMPLES::

            sage: b = BraidGroup(4)((3, 2, -1, -1))
```
The spacing of this block needs to be fixed.

src/sage/knots/link.py
```diff
@@ -2362,6 +2403,48 @@ def regions(self):
...
+          sage: K3 = K4.remove_loops()
+          sage: K3.pd_code()
+          [[1, 7, 2, 4], [3, 1, 4, 8], [7, 3, 8, 2]]
+          sage: U = Link([[1, 2, 2, 1]])
+          sage: U.remove_loops()
+          Link with 1 component represented by 0 crossings
+        """
+        pd = self.pd_code()
+        from copy import copy
+        new_pd = [copy(cr) for cr in pd if len(set(cr)) == 4]
```
tscrim:
```suggestion
        new_pd = [list(cr) for cr in pd if len(set(cr)) == 4]
```
This can be done to make a (shallow) copy of a list.

src/sage/knots/link.py
```diff
@@ -2362,6 +2403,48 @@ def regions(self):
...
+            # trivial knot
+            return type(self)([])
+        elif pd == new_pd:
+            # no loops detected
+            return self
+        from sage.sets.set import Set
+        new_edges = Set(flatten(new_pd))
+        for cr in Set(pd).difference(Set(new_pd)):
+            # cr is a loop crossing
+            rem = list(Set(cr).intersection(new_edges))
```
tscrim: Wouldn't it also be better to simply split the PD code into the two parts when constructing `new_pd` code? That way you could avoid having to use Sage `Set`'s and only need to pass over the PD code once.

tscrim (changes requested):

src/sage/knots/link.py
soehms: Travis, will this cover pull requests, too?

src/sage/knots/link.py
soehms: Good to know!

src/sage/knots/link.py
soehms: I will try to do this later on!

src/sage/knots/link.py
tscrim: Yes it will.

src/sage/knots/link.py
soehms: Done in https://github.com/sagemath/sage/pull/36884/commits/e0b5e55a624c30a4f8d9582b8e9601ac7c93f749.

src/sage/knots/link.py
```diff
@@ -2362,6 +2403,50 @@ def regions(self):
...
+        for cr in pd:
+            if len(set(cr)) == 4:
+                new_pd.append(list(cr))
+            else:
+                loop_crossings.append(cr)
+        if not new_pd:
+            # trivial knot
+            return type(self)([])
+        elif not loop_crossings:
+            return self
```
tscrim:
```suggestion
        if not loop_crossings:
            return self
        if not new_pd:
            # trivial knot
            return type(self)([])
```
This way you don't create multiple copies of the unknot.

tscrim (approved):
Thank you.

Just one small additional change; once done, you can set a positive review.

tscrim:
Thank you.

src/sage/knots/link.py
soehms: Good point!

soehms:
> Thank you.
> 
> Just one small additional change; once done, you can set a positive review.

Thanks!

PR #36889, opened 2023-12-15
more libgap in permutation groups ; get rid of _libgap_init_

fchapoton:
Moving forward from gap pexpect interface to libgap interface :

- using more libgap in the permutation groups code

- get rid of the barely used `_libgap_init_` method in favor of the unique `_gap_init_` method.

### :memo: Checklist


src/sage/groups/class_function.py
```diff
@@ -123,20 +125,20 @@ def __init__(self, G, values):
         if isinstance(values, GapElement) and gap.IsClassFunction(values):
             self._gap_classfunction = values
         else:
-            self._gap_classfunction = gap.ClassFunction(G, list(values))
+            self._gap_classfunction = gap.ClassFunction(G, gap(values))
```
dimpase: What is the point of calling `gap()` here? This is a call to pexpect GAP, what were getting rid of, right?

fchapoton: Well, this is inside `Classfunction_gap`. I have made so that `Classfunction` now calls `ClassFunction_libgap` by default, unless the second argument is a GapPexpect element.

Of course, it would be better to get rid completely of `Classfunction_gap` but maybe in another ticket ?

Do you have any idea about the failing doctests in class_function.py ?

dimpase: > Do you have any idea about the failing doctests in class_function.py ?

it comes from the fact that GAP operates on characters associated to a character table,
not to a group, and is unable to identify these tables for $\chi_1$ and $\chi_2$ in the 1st failing test. 
Not clear how to fix this easily.

```
sage: chi1 = ClassFunction(G, [3, 1, -1, 0, -1])
sage: isinstance(chi1, ClassFunction_libgap)
True
sage: chi2 = ClassFunction(G, [1, -1, 1, 1, -1])
sage: isinstance(chi2, ClassFunction_libgap)
True
sage: chi1._gap_classfunction
ClassFunction( CharacterTable( Sym( [ 1 .. 4 ] ) ), [ 3, 1, -1, 0, -1 ] )
sage: chi2._gap_classfunction
ClassFunction( CharacterTable( Sym( [ 1 .. 4 ] ) ), [ 1, -1, 1, 1, -1 ] )
sage: chi1._gap_classfunction*chi2._gap_classfunction
Cell In[11], line 1
----> 1 chi1._gap_classfunction*chi2._gap_classfunction

File /mnt/opt/Sage/sage-clang/src/sage/structure/element.pyx:1513, in sage.structure.element.Element.__mul__()
   1511 cdef int cl = classify_elements(left, right)
   1512 if HAVE_SAME_PARENT(cl):
-> 1513     return (<Element>left)._mul_(right)
   1514 if BOTH_ARE_ELEMENT(cl):
   1515     return coercion_model.bin_op(left, right, mul)

File /mnt/opt/Sage/sage-clang/src/sage/libs/gap/element.pyx:1056, in sage.libs.gap.element.GapElement._mul_()
   1054 try:
   1055     sig_GAP_Enter()
-> 1056     sig_on()
   1057     result = GAP_PROD(self.value, (<GapElement>right).value)
   1058     sig_off()

GAPError: Error, no product of class functions of different tables
```

dimpase:
there also seems to be a disconnect between Sage's conjugacy classes of $S_n$ and underlying (lib)GAP.
One has to tell the underlying GAP that these is the classes to use, otherwise it seems it's just by luck no discrepancies come in here. 
E.g. in this failing test for character multiplication it seems it's just a coincidence that GAP uses the same ordering of conjugacy classes as in `G.conjugacy_classes()`.

fchapoton:
the failure can be traced to the fact that the underlying character tables are not *identical*:

sage: ct1 = chi1._libgap_().UnderlyingCharacterTable()
sage: ct2 = chi2._libgap_().UnderlyingCharacterTable()
sage: ct1.IsIdenticalObj(ct2)
false

although they share the same group..

fchapoton:
and one can find that

```
sage: S = SymmetricGroup(4)
sage: g1=S._libgap_()
sage: g2=S._libgap_()
sage: g1.IsIdenticalObj(g2)
false
```

dimpase:
this is very strange. I would expect the 1st call to _libgap_() to create a record in S, and subsequent calls to reuse it

fchapoton:
the good old expect interface behaves better :
```
sage: S = SymmetricGroup(4)
sage: g1 = S._gap_()
sage: g2 = S._gap_()
sage: g1.IsIdenticalObj(g2)
true
```

dimpase:
I don't know `permgroup.py` code.
How about
```diff
--- a/src/sage/groups/perm_gps/permgroup.py
+++ b/src/sage/groups/perm_gps/permgroup.py
@@ -684,7 +684,8 @@ class PermutationGroup_generic(FiniteGroup):
         """
         if self._libgap is not None:
             return self._libgap
-        return super()._libgap_()
+        self._libgap = super()._libgap_()
+        return self._libgap
 
     # Override the default _libgap_ to use the caching as self._libgap
     _libgap_ = gap
```

This fixes this weird `false`.

dimpase:
now all tests in class_function.py pass. Some noise in doctests of permutation_group.py (same groups, different sets of generators, will fix them now)

dimpase:
OK, I'm not 100% sure about the fix
```
             sage: G = PermutationGroup([[(1,2,3,4)],[(1,2)]])
             sage: G.strong_generating_system()
             [[(), (1,2)(3,4), (1,3)(2,4), (1,4)(2,3)],
-             [(), (2,4), (2,3,4)], [(), (3,4)], [()]]
+            [(), (2,4,3), (2,3,4)],
+            [(), (3,4)],
+            [()]]
```

originally one had the sequence of subgroups $A_4$,  $S_3$, $S_2$, and now it's
$A_4$, $A_3$, $S_2$. Note that in both cases it's not a sequence of sugroups, each subgroup of the previous - and this  contracts to what I know about such systems.
(cf https://en.wikipedia.org/wiki/Strong_generating_set too)

It seems it's a bug, which now manifests itself differently. Something funny is going on, but I don't think it's due to my change.

fchapoton:
still one broken doctest in `src/sage/algebras/fusion_rings/fusion_double.py` that needs to be investigated

fchapoton:
The issue in fusion rings may be caused by the fact that the trivial character no longer comes first in the list :
```
sage: G = groups.permutation.Cyclic(2)
sage: [list(t) for t in G.irreducible_characters()]
[[1, -1], [1, 1]]
```
but I am not sure.

fchapoton:
there is indeed a difference of behaviour :
```

sage: [list(ClassFunction(G,irr)) for irr in G._gap_().Irr()]
[[1, 1], [1, -1]]
sage: [list(ClassFunction(G,irr)) for irr in G._libgap_().Irr()]
[[1, -1], [1, 1]]

```

dimpase:
I can try to look at it tonight.

dimpase:
Yes, that's in line 426 of `src/sage/algebras/fusion_rings/fusion_double.py`.

I don't see what they do there, let's ask @dwbump @tscrim 

Do you assume any fixed order of the irreducible characters in the character tables?
(e.g. that the trivial character is always 1st ?)

dimpase:
I tried modifying `irreducible_characters()` and `character_table()` so that the trivial character always comes 1st, and it allows this doctest to pass. So, indeed, this appears to be the root cause here.

dimpase:
However, doing such a change in SageMath is bound to cause a discrepancy, as we really don't want `G.character_table()` and `G._libgap_().Irr()` to give different orders.

fchapoton:
There method `_one_basis_` is assuming the 0 is the index of the unit in the fusion ring.

tscrim:
I don’t think there is anything else besides the `one_basis` method that uses that the trivial character is the first one. So it is easy enough to fix the issue by just searching in the list of characters for it.

dimpase:
OK, thanks.

fchapoton:
here is a tentative

dimpase:
as a side job, can we unify `irreducible_characters` and `character_table` for permutation groups? Both use libgap's `Irr()`, probably the latter should use the former instead.

dimpase:
Let us do the following change:

```diff
--- a/src/sage/structure/sage_object.pyx
+++ b/src/sage/structure/sage_object.pyx
@@ -755,7 +755,7 @@ cdef class SageObject:
 
     def _libgap_(self):
         from sage.libs.gap.libgap import libgap
-        return libgap.eval(self._gap_init_())
+        return libgap.eval(self)
 
     def _gp_(self, G=None):
         if G is None:
```

Your `libgap.eval(self._gap_init_())` creates a pexpect GAP handle for every libgap object used in Sage, and then passes it in text form to libgap. This is very inefficient.

dimpase (approved):
looks good to me

PR #36837, opened 2023-12-07
update qepcad to B 1.74

dimpase:
As noted in #36835, we missed an update to qepcad done in 2021. It has made it into Debian and Fedora,
and we use Debian's tarball and patches (most of the latter are updated versions of our B 1.72 patches)

We removed 'B' from the version number to accommodate the tarfile name from Debian.

We also update saclib to the latest version

Fixes #36831
Fixes #36835

PR #36895, opened 2023-12-16
fix qepcad doctest

dimpase:
This test started to fail on Sage 10.3.beta2 after the merged update of qepcad to 1.74 in https://github.com/sagemath/sage/pull/36837


dcoudert (approved):
Tested on Fedora 39 and macOS M1.

LGTM.

PR #36912, opened 2023-12-18
Take domain into account when computing with permutation groups

### LINKED ISSUE #36527: IntegerVectorsModPermutationGroup confuses groups whose domains are different

mantepse:
Fix #36527.

A permutation group really consists of a group and a domain.  After this pull request, 

* the domain is mentioned in the repr, if it is not natural, ie., `{1,\dots,n}`,
* conjugation of groups works for all domains
* permutation groups with different domain are different

### :memo: Checklist


tscrim:
I am not quite convinced this is the correct fix for #36527. I think there, that class should explicitly take the domain as input as it is part of the defining data and it might be something the user wants to pass in.

One subtle issue is converting the elements of a permutation group written as cycles (equivalently bijections) into one-line notation, which is definitely used in Sage's `Permutation` implementation and I believe is used by GAP's implementation. In particular, the domain needs to be a totally ordered set for this as that defines the identity permutation in one-line notation. I remember there being a trac ticket about this at some point.

mantepse:
> I am not quite convinced this is the correct fix for #36527. I think there, that class should explicitly take the domain as input as it is part of the defining data and it might be something the user wants to pass in.

Could you please clarify: with "that class", do you refer to "IntegerVectorsModPermutationGroup" or to "PermutationGroup"?

If the former, do you think that the domain should not be part of the definition of a permutation group?

> One subtle issue is converting the elements of a permutation group written as cycles (equivalently bijections) into one-line notation, which is definitely used in Sage's `Permutation` implementation and I believe is used by GAP's implementation. In particular, the domain needs to be a totally ordered set for this as that defines the identity permutation in one-line notation.

Indeed, but this will not affect equality of permutation groups, and it also does not affect the display (or meaning) of elements, it only affects the *creation* of elements, because `PermutationGroup` allows you to create elements in one-line notation.

mantepse:
> is the domain set to the moved points by default, or it could be None, too ?

`PermutationsGroup_generic.domain` is never `None`.  If you pass `domain=None`, it is either the set of moved points, or, if all moved points are integers, the smallest interval containing all moved points and 1.  For example, with this branch:
```
sage: PermutationGroup([(2,4)]) # domain will be "natural", that is {1, 2, 3, 4}
Permutation Group with generators [(2,4)]
sage: PermutationGroup([(0,2,4)])
Permutation Group with generators [(0,2,4)] and domain {0, 1, 2, 3, 4}
sage: PermutationGroup([("a",2,4)])
Permutation Group with generators [("a",2,4)] and domain {"a", 2, 4}
```

tscrim:
> > I am not quite convinced this is the correct fix for #36527. I think there, that class should explicitly take the domain as input as it is part of the defining data and it might be something the user wants to pass in.
> 
> Could you please clarify: with "that class", do you refer to "IntegerVectorsModPermutationGroup" or to "PermutationGroup"?
> 
> If the former, do you think that the domain should not be part of the definition of a permutation group?

`IntegerVectorsModPermutationGroup`. It can (and likely should) be part of a `PermutationGroup`, but the integer vectors are not so closely linked with the permutation group I feel.

> > One subtle issue is converting the elements of a permutation group written as cycles (equivalently bijections) into one-line notation, which is definitely used in Sage's `Permutation` implementation and I believe is used by GAP's implementation. In particular, the domain needs to be a totally ordered set for this as that defines the identity permutation in one-line notation.
> 
> Indeed, but this will not affect equality of permutation groups, and it also does not affect the display (or meaning) of elements, it only affects the _creation_ of elements, because `PermutationGroup` allows you to create elements in one-line notation. The output is always as a list of cycles.

It might affect the internal representation of the elements, which might need some care in other places (like equality of elements). We also have to be careful that anything we do doesn't interfere with those permutation groups that also inherit from `UniqueRepresentation`. This isn't anything mathematical, but some technical things to be aware of.

dimpase:
Yes, internally GAP uses 1-line, dense representation. 

I think we should stop pretending that the present implementation of PermutationGroup is correct in the way domains are dealt with.

We can either do the GAP-style, and declare that the domain is always the set of moved points, and only permutations  involved matter, or do the full-blown correct implementation of domains and
perm. groups as pairs (group, domain).

The former is easier, but it leaves the question of ordering of the domain open.

dimpase:
> ```
> sage: PermutationGroup([("a",2,4)])
> Permutation Group with generators [("a",2,4)] and domain {"a", 2, 4}
> ```

how does Sage divine the order on the domain? E.g.
```
sage: PermutationGroup([("a",4,2)])
Permutation Group with generators [(2,'a',4)] and domain {2, 4, 'a'}

```
Does it do the `UniqueRepresentation` thing on the domain as a set?

dimpase:
```
> sage: G
```
it is
```
Permutation Group with generators [((0,1,1),(1,1,1))((0,1,0),(1,1,0)), ((0,1,1),(0,1,0))((1,1,1),...
```
so this allows to get the domain from the permutations (under the assumption that this is what we want)

-----------
Also note that actually, for `repr()` in this case (if we keep domain in `repr()`), there is no need to print elements of the domain in the permutations, just print positions in there.

mantepse:
> > ```
> > sage: PermutationGroup([("a",2,4)])
> > Permutation Group with generators [("a",2,4)] and domain {"a", 2, 4}
> > ```
> how does Sage divine the order on the domain? E.g.
> ```
> sage: PermutationGroup([("a",4,2)])
> Permutation Group with generators [(2,'a',4)] and domain {2, 4, 'a'}
> ```

It sorts by string representation, so I don't quite understand how the above happened.

However, for equality, the order should not be taken into account.  

> Does it do the `UniqueRepresentation` thing on the domain as a set?

There is currently no subclass of `PermutationGroup` with `UniqueRepresentation` as far as I know.

mantepse:
> Also note that actually, for `repr()` in this case (if we keep domain in `repr()`), there is no need to print elements of the domain in the permutations, just print positions in there.

Unfortunately, the "natural" domain is {1,2,...,n} rather than {0,1,...,n-1}, so I guess this might be quite confusing.

mantepse:
> ```
> Permutation Group with generators [((0,1,1),(1,1,1))((0,1,0),(1,1,0)), ((0,1,1),(0,1,0))((1,1,1),...
> ```
> 
> so this allows to get the domain from the permutations (under the assumption that this is what we want)

I guess we could also omit the domain in `repr` if the group was created with `domain=None` (and store this fact as eg., `self._no_domain`).

We cannot omit the domain in `repr` if it consists of the moved points, because then we cannot distinguish between
`PermutationGroup([(2, 3)])` and `PermutationGroup([(2,3)], domain=[2,3])` anymore (unless we do this instead of omitting the domain if it is `{1,...,n}`).

However, I think it would be important to decide on equality first - although, the only option I can see currently is that two permutation groups are equal if the groups are equal and the domains are equal as sets.

mantepse:
Possibly we can do the following for equality:

if the domain keyword argument is not provided  *and* we have a "natural domain", that is, {1,2,...,n}, we treat the domain as $\mathbb N$.

Then, for comparability, we require that the domains coincide.

mantepse:
It seems to me that we could also take the opportunity to remove the arbitrary ordering of groups, in case they are not subgroups of each other.

jukkakohonen:
> if the domain keyword argument is not provided _and_ we have a "natural domain", that is, {1,2,...,n}, we treat the domain as

This sounds good. I mean, if the user has explicitly specified domains ``[1,2,3]`` and ``[1,2,3,4,5]`` for two trivial permutation groups, it would seem odd to assume that the user wants them to be the "same" group.  Treating them equal would anyway keep the integer vectors broken (issue `#36527` which was the original motivation for this whole thing).

But if the user has *not* specified a domain, perhaps they don't care about it that much. Or, they really are thinking of permutations on any arbitrary positive integers.

The mathematically clean way to do this is to say that permutation groups on "natural domains" are, in fact, permutation groups on the *infinite* domain $\mathbb{N}$, no matter which elements they happen to move. Very similar to how GAP treats them. Of course, the current implementation of `PermutationGroups` does not support infinite domains at all, so supporting $\mathbb{N}$ would need to be treated as a special case. (A full-fledged implementation of permutation groups on arbitrary infinite domains would be very different fish to fry.)

If implemented and documented very cleanly, this can clear up much of the confusion about permutation group domains that Sage is currently having. I think it is much clearer to say aloud that we treat such groups "acting on $\mathbb{N}$" than to make all kinds of confusing special cases like "yeah, this group is acting on $[1,2,3]$ and that other one is acting on $[1,2,3,4,5]$ so let's pretend the domains are equal, at least in this case, even though only one of them is equal to the domain of that third group which is also acting on $[1,2,3]$".

jukkakohonen:
Whatever the equality definition ends up being, I would like it to be mathematically well-behaved equality, in particular symmetric, reflexive and transitive. (We all know the current implementation is not transitive.)

For instance, if
```
A = PermutationGroup([(2,3)])
B = PermutationGroup([(2,3)], domain=[1,2,3])
C = PermutationGroup([(2,3)], domain=[1,2,3,4,5])
```
it should be clear from the definition (whatever it is) which of these three groups are equal, and which ones are not. If `B != C` (which seems desirable), then for example `A` cannot be equal to both of them. If we define that `A` implicitly has the smallest domain that contains all elements mentioned in the generators, then the domain is `[2,3]` and `A != B`. If we define that `A` implicitly has the "natural" domain `[1,2,3]`, then perhaps `A == B`. If we define that `A` implicitly has the infinite domain, then should it compare equal to `B`? To `C`? Both?

mantepse:
The last push is a bit experimental, in particular, I did not do the (necessary) conjugation yet if the orders of `self._domain` and `right._domain` differ.

However, it turned up a problem, which I don't know how to solve.  It might be unrelated, but I don't know.  In `src/sage/geometry/polyhedron/ppl_lattice_polytope.py`, line 979, we have (slightly edited to make reading easier)
```
            sage: from sage.geometry.polyhedron.ppl_lattice_polytope import LatticePolytope_PPL
            sage: Z3square = LatticePolytope_PPL((0,0), (1,2), (2,1), (3,3))
            sage: G1234 = Z3square.restricted_automorphism_group(vertex_labels=(1,2,3,4))
            sage: G1234 == PermutationGroup([[(2,3)], [(1,2),(3,4)]])
            True
            sage: G = Z3square.restricted_automorphism_group()
            sage: H = PermutationGroup([[((1,2),(2,1))], [((0,0),(1,2)),((2,1),(3,3))], [((0,0),(3,3))]])
            sage: G == H
            True
```
With the current PR, the last test gives `False`, because the domains differ.  The domain of `H` consists of tuples, whereas the domain of `G` is a subset of `Ambient free module of rank 2 over the principal ideal domain Integer Ring`.

@tscrim, can you help with this?  It seems to me that `PermutationGroup_generic.__richcmp__` should try to make the domains match, but I don't know how.  I am guessing that this is what you meant in your earlier comment.

So, essentially, I want to find a common parent of all the elements and coerce them into this parent - except that `tuple` is not a parent of course.

jukkakohonen:
> `PermutationsGroup_generic.domain` is never `None`. If you pass `domain=None`, it is either the set of moved points, or, if all moved points are integers, the smallest interval containing all moved points and 1. 

Not exactly. There is a fine distinction between "the moved points" and "the points mentioned in the generators it was created with", because a generator might explicitly mention a fixed point.

As of version 10.3.beta2:
```
sage: G=PermutationGroup([(4,5),(7,)])
sage: G.gens()
((), (4,5))
sage: G.domain()
{1, 2, 3, 4, 5, 6, 7}
sage: G.non_fixed_points()
[4, 5]
```

And if not all points are integers:
```
sage: G=PermutationGroup([(4,5),(7,),('a',)])
sage: G.domain()
{4, 5, 7, 'a'}
```

dimpase:
:facepalm: - another reason to always use explicit domains.

jukkakohonen:
> another reason to always use explicit domains.

At least if one needs to *care* about the domain. I'm sure there are situations where one does not care, as in "just give me a permutation group with these generators, and enough domain that I can work with them". But then one should not really expect to obtain a particular domain. And I'm not sure how much one should even expect of "equality" or "inequality" of such groups.

jukkakohonen:
> how does Sage divine the order on the domain? E.g.
> 
> ```
> sage: PermutationGroup([("a",4,2)])
> Permutation Group with generators [(2,'a',4)] and domain {2, 4, 'a'}
> ```

This is in `permgroup.py`:

```
        if domain is None:
            gens = [standardize_generator(x, as_cycles=True) for x in gens]
            domain = set()
            for x in gens:
                for cycle in x:
                    domain = domain.union(cycle)
            try:
                domain = sorted(domain)
            except TypeError:
                # Sorting the domain will sometimes fail with Python 3.
                # Fallback (not ideal: find a better solution?)
                domain = sorted(domain, key=str)
```

In your case, before sorting the domain is just a `set` containing the elements in some unspecified order. Then it tries to sort the elements as objects, but fails because `<` is not defined between integers and strings. Then it resorts to string representation, which determines the order as `'2'  < '4' < 'a'`.

tscrim:
One other important distinction we need to be careful with our wording is between a bug in the implementation and strongly suggesting a definition is ill-defined.

I can fix the inequality issue I noticed on #36527 by making the `PermutationGroup_generic.__richcmp__` lines starting at 799 into:
```python
        if gSelf.IsSubgroup(gRight):
            if op == op_LE or op == op_GE:  # check equality for <= and >=
                if gRight.IsSubgroup(gSelf):
                    return True
            return rich_to_bool(op, 1)
        if gRight.IsSubgroup(gSelf):
            # We have already verified that self <= right is not true
            return rich_to_bool(op, -1)
```
GAP is responsible for handling the equality (although our translation to GAP might have a bug), but this seems to make things consistent with one of the examples we have since everything is now depending on GAP (correctly). However, that is not answering @dimpase's question of matching GAP or doing our own with proper domain comparisons.

mantepse:
> Then it resorts to string representation, which determines the order as `'2' < '4' < 'a'`.

Didn't I write exactly this before?

jukkakohonen:
> Didn't I write exactly this before?

I clarified that it first tries to sort as objects, *then* resorts to strings. You are welcome.

mantepse:
> You also seemed to express some confusion by writing "so I don't quite understand how the above happened".

What I found confusing was:
```
sage: PermutationGroup([("a",2,4)])
Permutation Group with generators [("a",2,4)] and domain {"a", 2, 4}
```
in the same comment.

mantepse:
> The last push is a bit experimental

unfortunately, I forgot to push, so this will have to wait until tomorrow morning :-(

jukkakohonen:
> What I found confusing was:
> 
> ```
> sage: PermutationGroup([("a",2,4)])
> Permutation Group with generators [("a",2,4)] and domain {"a", 2, 4}
> ```

I see. I don't know what code you were using then, but with the code currently in this branch I get

```
sage: PermutationGroup([("a",2,4)])
Permutation Group with generators [(2,4,'a')] and domain {2, 4, 'a'}
```

dimpase:
might depend on the machine, position of planets, etc.

dimpase:
it might be that within one Python session a set created from the same elements has the same ordering, but I don't think that Python guarantees this. So this is not something that can be relied on.

dimpase:
> The last push is a bit experimental, in particular, I did not do the (necessary) conjugation yet if the orders of `self._domain` and `right._domain` differ.
> 
> However, it turned up a problem, which I don't know how to solve. It might be unrelated, but I don't know. In `src/sage/geometry/polyhedron/ppl_lattice_polytope.py`, line 979, we have (slightly edited to make reading easier)
> 
> ```
>             sage: from sage.geometry.polyhedron.ppl_lattice_polytope import LatticePolytope_PPL
>             sage: Z3square = LatticePolytope_PPL((0,0), (1,2), (2,1), (3,3))
>             sage: G1234 = Z3square.restricted_automorphism_group(vertex_labels=(1,2,3,4))
>             sage: G1234 == PermutationGroup([[(2,3)], [(1,2),(3,4)]])
>             True
>             sage: G = Z3square.restricted_automorphism_group()
>             sage: H = PermutationGroup([[((1,2),(2,1))], [((0,0),(1,2)),((2,1),(3,3))], [((0,0),(3,3))]])
>             sage: G == H
>             True
> ```
> 
> With the current PR, the last test gives `False`, because the domains differ. The domain of `H` consists of tuples, whereas the domain of `G` is a subset of `Ambient free module of rank 2 over the principal ideal domain Integer Ring`.

I get 
```
sage: type(H.domain()[0])
<class 'tuple'>
sage: type(G.domain()[0])
<class 'sage.modules.vector_integer_dense.Vector_integer_dense'>
```
still, `G==H` is false. Isn't it because of non-sensical
```
sage: isinstance(G.domain()[0],tuple)
False
```
Shouldn't the dense vectors inherit from `tuple` ?

tscrim:
> Shouldn't the dense vectors inherit from `tuple` ?

No, they definitely should not since they are mutable and often times have internal storage quite different from a tuple (e.g., you can pack an $F_2$ vector using bitsets).

tscrim:
That being said, we could convert everything list-like into a `tuple` for consistency since the domain must be finite and allows for non-hashable elements. Subsequently, sorting using the natural order with a fallback to `str` would make it have set-like behavior. I think this would be better than trying to find a common parent (the coercion framework can take care of the `==` between different parents when comparing the tuples).

mantepse:
I think that only hashable labels should be allowed.

Can you explain how to convert everything list like to tuples? What precisely is list like?

mantepse:
I'm afraid there is a misunderstanding: the problem I have is in the doctest of ppl_polytope, where the labels of self are morally equal to the labels of right but compare different.

mantepse:
Possibly, the doctest in question (reproduced below) simply shouldn't pass.  
```
            sage: from sage.geometry.polyhedron.ppl_lattice_polytope import LatticePolytope_PPL
            sage: Z3square = LatticePolytope_PPL((0,0), (1,2), (2,1), (3,3))
            sage: G1234 = Z3square.restricted_automorphism_group(vertex_labels=(1,2,3,4))
            sage: G1234 == PermutationGroup([[(2,3)], [(1,2),(3,4)]])
            True
            sage: G = Z3square.restricted_automorphism_group()
            sage: H = PermutationGroup([[((1,2),(2,1))], [((0,0),(1,2)),((2,1),(3,3))], [((0,0),(3,3))]])
            sage: G == H
            True
```
The domains are different: in `G` we have vectors, in `H` we have tuples, and these are different things.

mantepse:
With the last push, you can do
```
sage: G = Z3square.restricted_automorphism_group()
sage: H = PermutationGroup([[((1,2),(2,1))], [((0,0),(1,2)),((2,1),(3,3))], [((0,0),(3,3))]])
sage: G.conjugate(lambda g: tuple(g)) == H
True
```

tscrim:
Ah, I see. Yes, sorry for the misunderstanding.

I don't think there is a sensible way to automatically say they are the same inputs (as you said, tuples are not vectors). They are isomorphic as groups but not equal.

dimpase:
What is `conjugate` doing there? Does it act on domains? 
I'd not reuse `conjugate` for this..`conjugate(g)` is doing $gxg^{-1}$ for $g,x\in G$

mantepse:
Relabelling is a most natural généralisation of conjugation. But if you want a separate method that's fine with me. if I understand correctly, conjugate already does a bit more than conjugation with a group element

mantepse:
In principle, this should work now.  However, I have not yet adapted doctests and the docstring of `PermutationGroup_generic`,  and I have not yet checked when `domain` should be passed as a keyword argument and when it shouldn't.  For example, in `conjugate`, we should probably pass `domain` if and only if `not self._domain_is_N`.

Finally, maybe it makes sense to streamline `_has_natural_domain` and the new `_domain_is_N` to a single attribute.

I'd be grateful for testing and comments.  Most importantly, whether we want this modification in principle, or not so.

mantepse:
I think we should discuss getting rid of the idea of an infinite domain.

If I made no mistake, there wouldn't be any doctest failures either way.

I think that the following (which would be a consequence of having an infinite domain in the special case of positive integers) is a bit dangerous, even if documented:
```
            sage: H1 = PermutationGroup([[(1,2)],[(5,6)]])
            sage: H3 = PermutationGroup([[(1,2)]])
            sage: H3 < H1
            True

            sage: G1 = PermutationGroup([[(0,1)],[(4,5)]])
            sage: G3 = PermutationGroup([[(0,1)]])
            sage: G3 < G1
            False
```

jukkakohonen:
> I think we should discuss getting rid of the idea of an infinite domain.

I wouldn't mind. I always work with explicit finite domains anyway.

I don't know how many people are actually performing equality tests between permutation groups anyway (in Sage). Given how broken the current implementation is, I would expect we would have heard of them.

So, would the idea then be that equality is the mathematical equality (same domains, same permutations)? And if domain is left unspecified when creating a permutation group, then just some finite default domain is filled in (just like now)?

mantepse:
Yes

One more data point: if I made no mistake, `is_primitive` never uses the natural numbers as domain, but the smallest interval, if the domain is not specified.

dimpase:
primitivity implies transitivity, so it can only happen on moved by the group points.

mantepse:
Yes.

mantepse:
Shall we vote or can we reach a conclusion in some way?

dimpase:
I don't see a way to reach maths consistency here, without introducing explicit (finite) domains

tscrim:
@dimpase Can you be more explicit? I believe everything is consistent if we simply consider subgroups as `<=` by a subset of bijections (closed under composition) and equality from the corresponding poset. However, this is agnostic to the domain. I think we should allow the user to not have to specify the domain as it complicates the construction for a more casual user.

dimpase:
Do you suggest to ignore domains completely? Otherwise, we're stuck with  $A\leq B\leq A$  not implying $A=B$, see https://github.com/sagemath/sage/issues/36527#issuecomment-1848404613

tscrim:
No, that's not what I'm saying. We have one mathematically correct way to define equality, which is the documentation says, but the side effect is that equality does not include the domains at all. I am not saying this is desirable, but this appears to be what was intended and how GAP is comparing the groups. I don't think GAP allows specifying the domain as it seems to only consider permutations as bijections $\mathbb{N} \to \mathbb{N}$. Hence, we run into subtleties with how we convert a general domain to GAP. For example:
```
sage: G = PermutationGroup([(1,3)])
sage: H = PermutationGroup([(1,3)], domain=[1,3])
sage: G == H
sage: G._libgap_()
Group([ (1,3) ])
sage: H._libgap_()
Group([ (1,2) ])
sage: K = PermutationGroup([(1,3)], domain=[1,3,2])
sage: G == K
False
sage: K._libgap_()
Group([ (1,2) ])
sage: H == K
True
```
So to have permutations of arbitrary sets, we need to have a lot more care with subgroups and equality to fix the bugs unless things are consistently totally ordered (which is generally not true for partially ordered objects). This is independent of having perm groups including its domain or not.

tscrim:
Note also that our `PermutationGroup` documentation explicitly says that domain can be given as a list in order to make sense of one-line notation input.

mantepse:
> I think we should allow the user to not have to specify the domain as it complicates the construction for a more casual user.

I agree with this, but this is orthogonal to equality. The current pr, but also the alternative which I would actually favour, allows this.

Having only an infinite domain makes it difficult to interpret is_primitive.

mantepse:
@dimpase, what do you think, can we have this?

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -740,19 +750,36 @@ def _magma_init_(self, magma):
...
-        The comparison extends the subgroup relation. Hence, it is first checked
-        whether one of the groups is subgroup of the other. If this is not the
-        case then the ordering is whatever it is in GAP.
+        Two permutation groups are equal if and only if their domains
+        coincide as sets and their underlying groups are equal.
 
-        .. NOTE::
+        The relation `G < H` means that the domains of `G` and `H`
+        coincide as sets and that the underlying group of `G` is a
+        strict subgroup of `H`.
```
dimpase: proper subgroup

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -278,12 +280,31 @@ def PermutationGroup(gens=None, *args, **kwds):
 
     - a permutation group
 
+    .. NOTE::
+
+        If the domain is not specified, and all generators act on
+        integers, the domain is the smallest interval of integers
```
dimpase, line 287: be more precise about the type of integers. `ZZ` ? `int` ?

dimpase (changes requested):
should we also have tests for $G\leq H$, not only $G\lt H$ ?

src/sage/groups/perm_gps/permgroup.py
mantepse: could you propose language?  Would you prefer `act on integers (``ZZ`` or ``int``),...`?

src/sage/groups/perm_gps/permgroup.py
mantepse: done

mantepse:
I tested Jukkas examples, and found one instance which I do not completely understand:
```
sage: S = SymmetricGroup(Set("ac"))
sage: T = SymmetricGroup(Set("ce"))
sage: S < T
True
```
The reason is that there is a separate implementation of `_richcmp_` for `SymmetricGroup`
```
    def __richcmp__(self, x, op):
        """
        Fast comparison for SymmetricGroups.

        EXAMPLES::

            sage: S8 = SymmetricGroup(8)
            sage: S3 = SymmetricGroup(3)
            sage: S8 > S3
            True
        """
        if isinstance(x, SymmetricGroup):
            return richcmp((self._deg, self._domain), (x._deg, x._domain), op)
        return super().__richcmp__(x, op)
```
Can somebody spell out for me what the second line does - that is, in the case that `x` is a `SymmetricGroup`?

tscrim:
> I tested Jukkas examples, and found one instance which I do not completely understand:
> 
> ```
> sage: S = SymmetricGroup(Set("ac"))
> sage: T = SymmetricGroup(Set("ce"))
> sage: S < T
> True
> ```
> 
> The reason is that there is a separate implementation of `_richcmp_` for `SymmetricGroup`
> 
> ```
>     def __richcmp__(self, x, op):
>         if isinstance(x, SymmetricGroup):
>             return richcmp((self._deg, self._domain), (x._deg, x._domain), op)
>         return super().__richcmp__(x, op)
> ```
> 
> Can somebody spell out for me what the second line does - that is, in the case that `x` is a `SymmetricGroup`?

It is simply comparing the two tuples `(degree, domain)`, which is done "lexicographically". So it always has `A < B` whenever the degree of `A` is less than the degree of `B`. Next it compares the domains. I believe this is done to get things closer to a total ordering on the symmetric groups. It is just not necessarily consistent with subgroup testing.

mantepse:
Since `SymmetricGroup` is a `PermutationGroup`, to make things consistent, `SymmetricGroup(2) < SymmetricGroup(3)` should return `False`, because the domains differ.

Do we have to add a deprecation warning to do this?  If so, how?

mantepse:
This is the implementation I would do.  Previously, `S8 > S3` would give `True`.
```python
    def __richcmp__(self, x, op):
        """
        Fast comparison for SymmetricGroups.

        This compares ``self`` and ``x`` as permutation groups.
        Thus, `G == H`` if and only if their domains coincide as
        sets.  Otherwise, they are different but incomparable.

        EXAMPLES::

            sage: S8 = SymmetricGroup(8)
            sage: S3 = SymmetricGroup(3)
            sage: S8 > S3
            False

        TESTS::

            sage: A = SymmetricGroup(2)
            sage: B = SymmetricGroup(["a", "b"])
            sage: C = SymmetricGroup(["b", "a"])
            sage: D = SymmetricGroup(["a", "c"])
            sage: E = SymmetricGroup(["a", "b", "c"])
            sage: groups = [A, B, C, D, E]
            sage: not any(X > Y for X in groups for Y in groups)
            True
            sage: [[ZZ(X == Y) for X in groups] for Y in groups]
            [[1, 0, 0, 0, 0],
             [0, 1, 1, 0, 0],
             [0, 1, 1, 0, 0],
             [0, 0, 0, 1, 0],
             [0, 0, 0, 0, 1]]

        """
        if isinstance(x, SymmetricGroup):
            if self._deg == x._deg and set(self._domain) == set(x._domain):
                return op is op_EQ or op is op_LE or op is op_GE
            return op is op_NE
            
        return super().__richcmp__(x, op)
```
There are no other doctest failures in the library, which makes me hope that not too many people use it as shortcut for subgroup.

mantepse:
I found one more thing to check: there is a special `__richcmp__` implementation for `PermutationGroup_subgroup`.  I am not sure whether this is affected by the new philosophy.  Note that subgroups of permutation groups inherit the domain of the ambient group.

dimpase:
> It is simply comparing the two tuples `(degree, domain)`, which is done "lexicographically". So it always has `A < B` whenever the degree of `A` is less than the degree of `B`. Next it compares the domains. I believe this is done to get things closer to a total ordering on the symmetric groups.

but this has very little to do with symmetric groups being permutation groups. It might make sense in some weird category, but "total ordering on the symmetric groups" isn't something I  know anything about.

mantepse:
@dimpase, it would be great if you could comment on https://github.com/sagemath/sage/pull/36912#issuecomment-1878714566!

I'd like to finish this before sage days in Le Teich end of January.

dimpase:
we're just back from vacations, let me put out few fires...

src/sage/groups/perm_gps/permgroup.py
dimpase: I'd  prefer a full detail, using Sphinx `:class:` tag to specify the type completely.

dimpase:
Other than my comment on the doc for "integers", it's good to go.

mantepse:
> Other than my comment on the doc for "integers", it's good to go.

Well, not from my part: there is a big question in comment https://github.com/sagemath/sage/pull/36912#issuecomment-1878714566.  It seems to me that this is the natural category.  Am I mistaken?

dimpase:
> > Other than my comment on the doc for "integers", it's good to go.
> 
> Well, not from my part: there is a big question in comment [#36912 (comment)](https://github.com/sagemath/sage/pull/36912#issuecomment-1878714566). It seems to me that this is the natural category. Am I mistaken?

It's a bit unusual, in the sense that one can have $X\subset Y$ and $S=T$, giving $(X,S)<(Y,T)$, but on the other hand, why not?

mantepse:
Is there any argument which might favour one category over the other?

dimpase:
we either have one category for a fixed domain,
of subgroups of the full symmetric group on it, or we have your category, which has more objects.

The former are subcategories of the latter.

mantepse:
Why is the former not a category? I would have thought that we just remove all morphisms between permutation groups that have different domains. Or did I misunderstand you?

dimpase:
> Why is the former not a category? I would have thought that we just remove all morphisms between permutation groups that have different domains. Or did I misunderstand you?

I meant to say that you either have a category _per domain_, or you have the one you suggest. I like your proposal better.

mantepse:
@dimpase, I just realized that things are a bit more subtle than I thought.

Consider the two groups $G=\langle (1, 2)\rangle$ and $H=\langle((1,2),(3,4))\rangle$, both having a single generator (!).  Thus, as abstract groups they are in fact isomorphic.  Do we want that $(\\{1,2\\}, G) < (\\{1,2,3,4\\}, H)$?

I think that the definition inspired by the category above would say that this is the case, because there is a set-preserving homomorphism from $G$ to $H$.  However, $G$ is not a subgroup of $H$, of course.

But I am not at all firm in category theory.  I thought I would be using https://en.wikipedia.org/wiki/Subobject, where I additionally require that the morphism, restricted to the domain $X$ of $G$, is the identity.  But I am not sure whether I interpret the definition correctly.

dimpase:
It seems that this definition is not good for us. It might be coming from some formal methods CS area ("Nominal sets" ?).
We want to have the maps $\phi^g$ to be a monomorphism of groups of permutations, and $\phi^s$ to be
a restriction of the identity map.

mantepse:
> It seems that this definition is not good for us. It might be coming from some formal methods CS area ("Nominal sets" ?). We want to have the maps $ϕ^g$ to be a monomorphism of groups of permutations, and $ϕ^s$ to be a restriction of the identity map.

Isn't, in my example, both the case?  That is, I would have thought that the map sending $(1,2)$ to $(1,2)(3,4)$ is a monomorphism, no?

dimpase:
monomorphism of groups of permutations, not of abstract groups

mantepse:
> monomorphism of groups of permutations, not of abstract groups

Please expand, I am not familiar enough with the definitions.  Why isn't that a monomorphism of groups of permutations?

dimpase:
> > monomorphism of groups of permutations, not of abstract groups
> 
> Please expand, I am not familiar enough with the definitions. Why isn't that a monomorphism of groups of permutations?

a permutation should go to an equal permutation.

I think all this category stuff is leading us up a garden path. Surely the permutation groups on a set form a category, with the usual group theory definitions for morphisms. (This category is very boring, has no non-obvious morphisms, etc)

mantepse:
> a permutation should go to an equal permutation.

I think I have now understood my mistake.  I misunderstood the notion of subobject - I thought it would be an object, which is nonsense.  A subobject in a category is (an isomorphism class) of monomorphisms.

My idea was to model `_richcmp_` after the notion of subobject, by replacing 'homomorphism' with 'restriction of the identity'.

So, maybe the question boils down to: are the permutations $(1,2)\in\mathfrak S_2$ and $(1,2)(3)(4)\in\mathfrak S_4$ equal?

> I think all this category stuff is leading us up a garden path

Quite possible, but a decision must be taken.  What should `<` mean?

dimpase:
> > a permutation should go to an equal permutation.
> 
> I think I have now understood my mistake.  I misunderstood the notion of subobject - I thought it would be an object, which is nonsense.  A subobject in a category is (an isomorphism class) of monomorphisms.
> 
> My idea was to model `_richcmp_` after the notion of subobject, by replacing 'homomorphism' with 'restriction of the identity'.
> 
> So, maybe the question boils down to: are the permutations $(1,2)\in\mathfrak S_2$ and $(1,2)(3)(4)\in\mathfrak S_4$ equal?


no, not in the usual group theory. maybe in some fancy-shmancy category they are, but we don't do this here.


> 
> > I think all this category stuff is leading us up a garden path
> 
> Quite possible, but a decision must be taken.  What should `<` mean?

$H<G$ if $H$ is a subgroup of $G$ (in particular, a subset of $G$).
And the domain of $H$ must be a subset of the domain of $G$

mantepse:
> > So, maybe the question boils down to: are the permutations $(1,2)\in\mathfrak S_2$ and $(1,2)(3)(4)\in\mathfrak S_4$ equal?
> 
> no, not in the usual group theory. maybe in some fancy-shmancy category they are, but we don't do this here.

> > What should `<` mean?
> 
> H<G if H is a subgroup of G (in particular, a subset of G). Same for ≤, but allowing equality. And the domain of H must be a subset of the domain of G

Fine with me!  It seems to me, however, that this is effectively saying that $(1,2)\in\mathfrak S_2$ and $(1,2)(3)(4)\in\mathfrak S_4$ are equal, because otherwise $(1,2)$ would not be in the group generated by $(1,2)(3)(4)$.

dimpase:
these are two representations of the same permutation.
(unless you encode the domains with length 1 cycles).

it's akin to asking whether $1=x^0$.

dimpase:
The simplest, and it seems to me the only posdwould be to say that we can only compare if domains are equal (as sets).
This is to avoid precisely this issue.

We can perhaps introduce "groups of permutations" - this is how GAP calls their permutation groups-there the domains are implicit, they are sets of moved points.
To quote GAP manual:
---
A permutation group is a group of permutations on a finite set Ω of positive integers. 
---
I

mantepse:
> The simplest would be to say that we can only compare if domains are equal (as sets). This is to avoid precisely this issue.

I agree that this is the simplest solution.  

But I am completely happy also with the alternative, that is, $(X, G) \leq (Y, H)$ if $X \subseteq Y$ and $\bar G\leq H$, where $\bar G$ is obtained from $G$ by keeping all points in $Y\setminus X$ fixed.  I think that this is a consistent definition.  I don't know whether this definition is any better than the other.

src/sage/combinat/designs/incidence_structures.py
```diff
@@ -1840,6 +1840,8 @@ def automorphism_group(self):
             True
             sage: G
             Permutation Group with generators [...]
+            and domain {(0, 0, 1), (1, 0, 1), (0, 1, 1), (1, 1, 1), (0, 1, 0), (1, 1, 0), (1, 0, 0)}
+
```
mkoeppe, line 1844:
```suggestion
```
(deleting this empty line will fix the dataflow warning)

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -586,7 +596,7 @@ def construction(self):
             (1,2)(3,4,5)('a','b')
             sage: P = parent(p)
             sage: P
-            Permutation Group with generators [('a','b'), (1,2), (1,2,3,4,5)]
+            Permutation Group with generators [('a','b'), (1,2), (1,2,3,4,5)] and domain {1, 2, 3, 4, 5, 'a', 'b'}
```
mkoeppe, line 621:
```suggestion
            Permutation Group with generators [('a','b'), (1,2), (1,2,3,4,5)] 
             and domain {1, 2, 3, 4, 5, 'a', 'b'}
```

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -774,12 +801,22 @@ def __richcmp__(self, right, op):
...
             sage: H1 = PermutationGroup([[(1,2)],[(5,6)]])
             sage: H2 = PermutationGroup([[(3,4)]])
+            sage: H1 < H2
+            False
+            sage: H2 < H1
+            False
+
+        The domain of `H1` is `\{1,2,3,4,5,6\}`, whereas the domain
+        of the permutation group `H3` below is `\{1,2\}`.  Therefore
+        `H3` is not a permutation subgroup of `H1`::
```
mkoeppe, line 833:
```suggestion
        The domain of ``H1`` is `\{1,2,3,4,5,6\}`, whereas the domain
        of the permutation group ``H3`` below is `\{1,2\}`.  Therefore
        ``H3`` is not a permutation subgroup of ``H1``::
```

src/sage/graphs/generic_graph.py
```diff
@@ -23695,8 +23695,10 @@ def automorphism_group(self, partition=None, verbosity=0,
             sage: G1 = Graph(':H`ECw@HGXGAGUG`e')
             sage: G = G1.automorphism_group()                                           # needs sage.groups
             sage: G.subgroups()                                                         # needs sage.groups
-            [Subgroup generated by [()] of (Permutation Group with generators [(0,7)(1,4)(2,3)(6,8)]),
-             Subgroup generated by [(0,7)(1,4)(2,3)(6,8)] of (Permutation Group with generators [(0,7)(1,4)(2,3)(6,8)])]
+            [Subgroup generated by [()] of (Permutation Group with generators [(0,7)(1,4)(2,3)(6,8)]
+             and domain {0, 1, 2, 3, 4, 5, 6, 7, 8}),
+             Subgroup generated by [(0,7)(1,4)(2,3)(6,8)] of (Permutation Group with generators [(0,7)(1,4)(2,3)(6,8)]
+             and domain {0, 1, 2, 3, 4, 5, 6, 7, 8})]
```
mkoeppe:
```suggestion
            [Subgroup generated by [()] of 
               (Permutation Group with generators [(0,7)(1,4)(2,3)(6,8)]
                and domain {0, 1, 2, 3, 4, 5, 6, 7, 8}),
             Subgroup generated by [(0,7)(1,4)(2,3)(6,8)] of 
               (Permutation Group with generators [(0,7)(1,4)(2,3)(6,8)]
                and domain {0, 1, 2, 3, 4, 5, 6, 7, 8})]
```

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -2118,8 +2193,32 @@ def _repr_(self):
...
+
+        TESTS::
+
+            sage: PermutationGroup([], domain=[])
+            Permutation Group with generators [()]
+
         """
-        return "Permutation Group with generators %s" % list(self.gens())
+        if (self._has_natural_domain()
+            and self.largest_moved_point() == max(self.domain(), default=0)):
```
mkoeppe, line 2241:
```suggestion
                and self.largest_moved_point() == max(self.domain(), default=0)):
```
(need to indent because of the following line)

mkoeppe (commented):
Otherwise LGTM

mantepse:
Thank you for the input!

I met Travis in person and we were discussing this issue.  The conclusion is that we will propose a variation of the implemented change which is hopefully more useful and at the same time more logical.  I have not done this yet because I'm focusing on the lazy solver.

mantepse:
Note to myself:
```
sage: G = PermutationGroup([(1,2)], domain=[1,2,3])
sage: H = PermutationGroup([(1,3)], domain=[1,2,3])
sage: G.is_subgroup(H)
False
sage: G = PermutationGroup([(1,2)], domain=[1,2,3])
sage: H = PermutationGroup([(1,3)], domain=[1,3,2])
sage: G.is_subgroup(H)
True
```

PR #36901, opened 2023-12-17
`build/pkgs/e_antic`: Update to 2.0.0

mkoeppe:
The new release adds support for FLINT 3. https://github.com/flatsurf/e-antic/releases/tag/2.0.0 Thanks @saraedum for preparing this release!

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36930, opened 2023-12-20
`build/pkgs/normaliz`: Add patch for FLINT 3

mkoeppe:
### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36901


tscrim:
Shouldn't we bump the patch level?

Also, even with this patch and #36901 I cannot build normaliz:
```
checking for fmpz_poly_set_coeff_mpz in -lflint... no
configure: error: flint is not available but was requested
```

mkoeppe:
> Shouldn't we bump the patch level?

Really only necessary when we want to replace something that builds by something else that builds.

mkoeppe:
This builds for me now

tscrim (approved):
This now builds and works for me (as far as I can tell).

mkoeppe:
Thanks! I've thrown in the patch level bump for you

PR #36558, opened 2023-10-28
`sage -t --format github` (illustration)

mkoeppe:
This prints doctest failures as github annotations. 

The branch on this PR here provokes some errors for illustration purposes and should not be merged. See https://github.com/sagemath/sage/pull/36938 for the actual PR.

As seen in the "Files changed" tab (https://github.com/sagemath/sage/pull/36558/files):
- Doctest failures show as errors next to the doctest in source code
- Doctest warnings show as "warnings"
- Doctest failures from files that "failed in baseline" show as "notices"

These messages can also be seen in the Summary page of the workflow (https://github.com/sagemath/sage/actions/runs/7282457873?pr=36558). Clicking on a message leads to the source code location.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36936

PR #36936, opened 2023-12-20
`src/sage/doctest/forker.py`: Show '# [failed in baseline]' earlier

mkoeppe:
The note "[failed in baseline]" already appears in the doctest failure summary.

Here we add the note also to the `sage -t` line during doctesting:

https://github.com/sagemath/sage/actions/runs/7281940815/job/19843412065?pr=36936#step:11:2744
```
sage -t --random-seed=256963700569517996050547927079750684585 src/sage/combinat/combinatorial_map.py
    [75 tests, 0.05 s]
sage -t --random-seed=256963700569517996050547927079750684585 src/sage/combinat/cluster_algebra_quiver/quiver.py  # [failed in baseline]
    [320 tests, 3.73 s]
sage -t --random-seed=256963700569517996050547927079750684585 src/sage/combinat/composition_signed.py
    [20 tests, 0.16 s]
```

https://github.com/sagemath/sage/actions/runs/7281940815/job/19843412065?pr=36936#step:11:9192
```
sage -t --random-seed=256963700569517996050547927079750684585 src/sage_setup/setenv.py
    [0 tests, 0.00 s]
sage -t --random-seed=256963700569517996050547927079750684585 src/sage_setup/clean.py  # [failed in baseline]
File "src/sage_setup/clean.py", line 104, in sage_setup.clean._find_stale_files
Failed example:
    for f in stale_iter:
        if f.endswith(skip_extensions): continue
        if '/ext_data/' in f: continue
        print('Found stale file: ' + f)
Expected nothing
Got:
    Found stale file: sage/tests/books/judson-abstract-algebra/homomorph-sage-exercises.py
    Found stale file: sage/tests/books/judson-abstract-algebra/actions-sage-exercises.py
    Found stale file: sage/tests/books/judson-abstract-algebra/homomorph-sage.py
```

The changes in the code that implement this are also preparation for:
- #36558.

### :memo: Checklist

### DIFF
--- a/src/sage/doctest/reporting.py
+++ b/src/sage/doctest/reporting.py
@@ -162,13 +162,15 @@ class DocTestReporter(SageObject):
             return True
         return False
 
-    def report_head(self, source):
+    def report_head(self, source, fail_msg=None):
         """
-        Return the "sage -t [options] file.py" line as string.
+        Return the ``sage -t [options] file.py`` line as string.
 
         INPUT:
 
         - ``source`` -- a source from :mod:`sage.doctest.sources`
+
+        - ``fail_msg`` -- ``None`` or a string
 
         EXAMPLES::
 
@@ -190,6 +192,8 @@ class DocTestReporter(SageObject):
             sage: DD.long = True
             sage: print(DTR.report_head(FDS))
             sage -t --long .../sage/doctest/reporting.py
+            sage: print(DTR.report_head(FDS, "Failed by self-sabotage"))
+            sage -t --long .../sage/doctest/reporting.py  # Failed by self-sabotage
         """
         cmd = "sage -t"
         if self.controller.options.long:
@@ -206,6 +210,13 @@ class DocTestReporter(SageObject):
         if environment != "sage.repl.ipython_kernel.all_jupyter":
             cmd += f" --environment={environment}"
         cmd += " " + source.printpath
+        baseline = self.controller.source_baseline(source)
+        if fail_msg:
+            cmd += "  # " + fail_msg
+        if baseline.get('failed', False):
+            if not fail_msg:
+                cmd += "  #"
+            cmd += " [failed in baseline]"
         return cmd
 
     def report(self, source, timeout, return_code, results, output, pid=None):
@@ -420,12 +431,10 @@ class DocTestReporter(SageObject):
                         fail_msg += " (and interrupt failed)"
                     else:
                         fail_msg += " (with %s after interrupt)" % signal_name(sig)
-                if baseline.get('failed', False):
-                    fail_msg += " [failed in baseline]"
                 log("    %s\n%s\nTests run before %s timed out:" % (fail_msg, "*"*70, process_name))
                 log(output)
                 log("*"*70)
-                postscript['lines'].append(cmd + "  # %s" % fail_msg)
+                postscript['lines'].append(self.report_head(source, fail_msg))
                 stats[basename] = {"failed": True, "walltime": 1e6, "ntests": ntests}
                 if not baseline.get('failed', False):
                     self.error_status |= 4
@@ -436,12 +445,10 @@ class DocTestReporter(SageObject):
                     fail_msg = "Killed due to %s" % signal_name(-return_code)
                 if ntests > 0:
                     fail_msg += " after testing finished"
-                if baseline.get('failed', False):
-                    fail_msg += " [failed in baseline]"
                 log("    %s\n%s\nTests run before %s failed:" % (fail_msg,"*"*70, process_name))
                 log(output)
                 log("*"*70)
-                postscript['lines'].append(cmd + "  # %s" % fail_msg)
+                postscript['lines'].append(self.report_head(source, fail_msg))
                 stats[basename] = {"failed": True, "walltime": 1e6, "ntests": ntests}
                 if not baseline.get('failed', False):
                     self.error_status |= (8 if return_code > 0 else 16)
@@ -458,13 +465,13 @@ class DocTestReporter(SageObject):
                     log("    Error in doctesting framework (bad result returned)\n%s\nTests run before error:" % ("*"*70))
                     log(output)
                     log("*"*70)
-                    postscript['lines'].append(cmd + "  # Testing error: bad result")
+                    postscript['lines'].append(self.report_head(source, "Testing error: bad result"))
                     self.error_status |= 64
                 elif result_dict.err == 'noresult':
                     log("    Error in doctesting framework (no result returned)\n%s\nTests run before error:" % ("*"*70))
                     log(output)
                     log("*"*70)
-                    postscript['lines'].append(cmd + "  # Testing error: no result")
+                    postscript['lines'].append(self.report_head(source, "Testing error: no result"))
                     self.error_status |= 64
                 elif result_dict.err == 'tab':
                     if len(result_dict.tab_linenos) > 5:
@@ -473,11 +480,11 @@ class DocTestReporter(SageObject):
                     if len(result_dict.tab_linenos) > 1:
                         tabs = "s" + tabs
                     log("    Error: TAB character found at line%s" % (tabs))
-                    postscript['lines'].append(cmd + "  # Tab character found")
+                    postscript['lines'].append(self.report_head(source, "Tab character found"))
                     self.error_status |= 32
                 elif result_dict.err == 'line_number':
                     log("    Error: Source line number found")
-                    postscript['lines'].append(cmd + "  # Source line number found")
+                    postscript['lines'].append(self.report_head(source, "Source line number found"))
                     self.error_status |= 256
                 elif result_dict.err is not None:
                     # This case should not occur
@@ -494,7 +501,7 @@ class DocTestReporter(SageObject):
                     if output:
                         log("Tests run before doctest exception:\n" + output)
                         log("*"*70)
-                    postscript['lines'].append(cmd + "  # %s" % fail_msg)
+                    postscript['lines'].append(self.report_head(source, fail_msg))
                     if hasattr(result_dict, 'tb'):
                         log(result_dict.tb)
                     if hasattr(result_dict, 'walltime'):
@@ -506,9 +513,7 @@ class DocTestReporter(SageObject):
                     f = result_dict.failures
                     if f:
                         fail_msg = "%s failed" % (count_noun(f, "doctest"))
-                        if baseline.get('failed', False):
-                            fail_msg += " [failed in baseline]"
-                        postscript['lines'].append(cmd + "  # %s" % fail_msg)
+                        postscript['lines'].append(self.report_head(source, fail_msg))
                         if not baseline.get('failed', False):
                             self.error_status |= 1
                     if f or result_dict.err == 'tab':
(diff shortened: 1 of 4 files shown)


kwankyu:
How about `sage -t ....  # fails in baseline`?

The test is not yet run, so we should say that "`sage -t ...` is expected to fail but this failure is in the baseline". Hence no past tense.

mkoeppe:
Not sure about the wording. Would it help if we include the filename (`.github/workflows/ci-conda-known-test-failures.json`) in the messages somehow? This name is clearer than the very neutral "baseline".

kwankyu:
> Would it help if we include the filename (`.github/workflows/ci-conda-known-test-failures.json`) in the messages somehow? This name is clearer than the very neutral "baseline".

Better to be short. Anyway, developers should learn about the precise meaning of the message if they are interested. We cannot explain everything in the small space without cluttering.

kwankyu:
May be "conda baseline"?

mkoeppe:
> Would it help if we include the filename (`.github/workflows/ci-conda-known-test-failures.json`) in the messages somehow?

OK, I'm logging it just once at the beginning

mkoeppe:
> The test is not yet run, so we should say that "`sage -t ...` is expected to fail but this failure is in the baseline". Hence no past tense.

In my view, the past tense refers to the time when the baseline was established

kwankyu:
Then I don't understand the phrase "failed in the baseline" here. 

By the dictionary definition, "the baseline" is the line from which we measure quality of someone or something. 

We collected a list of files that fail persistently (or randomly persistently), and regard the list as part of "the baseline".  So if the file fails again in this run, then it is not a failure measured from "the baseline". You express this by "Failed (but it is) in the baseline". No?

Now you use the same phrase in "sage -t ...  # failed in the baseline". The comment is about "sage -t ...". At this moment, nothing yet failed or it may not fail at all. I cannot understand the same phrase "failed in the baseline"  as above. You want to say that "this may fail but it is in the baseline". No?

kwankyu:
> In my view, the past tense refers to the time when the baseline was established

I see. So you mean "[This file] failed when the baseline was established" by "Failed in the baseline".

mkoeppe:
Yes. And in the summary line, when it says
```
sage -t --random-seed=271115389197827456996254478912680781530 src/ sage/doctest/sources.py  # 2 doctests failed [failed in baseline]
```
... it means it failed just now (and also it failed when the baseline was established).

kwankyu:
OK. It makes sense to me now.

kwankyu:
> ... it means it failed just now (and also it failed when the baseline was established).

You may want to use this explanation when you explain the new facility in the developer manual or in the release tour.

mkoeppe:
bottom of this subsection: https://doc.sagemath.org/html/en/developer/doctesting.html#auxiliary-files

kwankyu:
Yes. The phrase "[failed in baseline]" is explained there. Good enough.

Here 

https://doc.sagemath.org/html/en/developer/doctesting.html#:~:text=%2D%2Dbaseline%2Dstats%2Dpath%20known%2Dtest%2Dfailures.json

"=" is necessary.

Otherwise looks good to me. It works well.

mkoeppe:
> "=" is necessary.

Both ways of writing are actually fine with the Python argparse. See examples just above that do `--stats-path ~/.sage/timings2.json` or `--logfile test1.log `

kwankyu:
Instead of `timing2.json`, `.github/workflows/ci-conda-known-test-failures.json` may be a better example.

There's no `"failed": true` in `timing2.json`.

kwankyu:
> > "=" is necessary.
> 
> Both ways of writing are actually fine with the Python argparse. See examples just above that do `--stats-path ~/.sage/timings2.json` or `--logfile test1.log `

But 
```
$ sage -t --baseline-stats-path .github/workflows/ci-conda-known-test-failures.json src/sage_setup/clean.py
usage: sage -t [options] filenames
sage-runtests: error: argument --baseline_stats_path/--baseline-stats-path: expected one argument
```

mkoeppe:
hm... let me check

kwankyu:
> Instead of `timing2.json`, `.github/workflows/ci-conda-known-test-failures.json` may be a better example.
> 
> There's no `"failed": true` in `timing2.json`.

Yes, there is one:
```
    "sage.graphs.generators.families": {
        "failed": true,
        "ntests": 446,
        "walltime": 11.227901220321655
    }, 
```

mkoeppe:
Likely this custom code https://github.com/sagemath/sage/blob/develop/src/bin/sage-runtests#L131 is to blame

mkoeppe:
I'll take another look at this tomorrow.

mkoeppe:
> $ sage -t --baseline-stats-path .github/workflows/ci-conda-known-test-failures.json src/sage_setup/clean.py
> usage: sage -t [options] filenames
> sage-runtests: error: argument --baseline_stats_path/--baseline-stats-path: expected one argument

Fixed now

kwankyu:
Thanks. 

This is an example run:
```
Using --baseline-stats-path=.github/workflows/ci-conda-known-test-failures.json
Doctesting 1 file.
sage -t --warn-long 59.9 --random-seed=30564486128159533757106732980543283332 src/sage_setup/clean.py  # [failed in baseline: _find_stale_files finds some stale files under sage/tests when executed under conda]
File "src/sage_setup/clean.py", line 48, in sage_setup.clean._remove
Failed example:
    files
Expected:
    {'a/b/c.py'}
Got:
    {'a/b/c.pyx'}
1 item had failures:
   1 of   8 in sage_setup.clean._remove
    [21 tests, 1 failure, 0.44 s]
sage -t --warn-long 59.9 --random-seed=30564486128159533757106732980543283332 src/sage_setup/clean.py  # 1 doctest failed [failed in baseline: _find_stale_files finds some stale files under sage/tests when executed under conda]
```

mkoeppe:
This last commit takes care of https://github.com/sagemath/sage/actions/runs/7306041656/job/19910309897?pr=36938#step:11:8651
```
sage -t --random-seed=239832805085423255575990885703796854524 src/sage/structure/coerce_actions.pyx
File "src/sage/structure/coerce_actions.pyx", line 784, in sage.structure.coerce_actions.IntegerMulAction.noexcept
Failed example:
    alarm(0.001); 2^(10^7) * P
Expected:
    ...
    AlarmInterrupt
Got:
    (0 : 1 : 0)
    AlarmInterrupt in doctesting framework
```

kwankyu:
I am educating myself about the baseline-stats mechanism. It was introduced in #33233. Reading that, I learn that the mechanism was file-scoped from the start. Why was it so? The basic idea that known doctest failures should be ignored is good. But the idea of keeping records of failed files instead of failed doctests seems inherently flawed. Does it make a better sense for the target application, such as modularization? 

How is the mechanism used for testing a distribution package? Where should I look to learn about that?

mkoeppe:
`failed` True or False for each doctest module -- that's simply all that the Sage doctester records in the stats. When I implemented baseline-stats, I kept this to keep the scope of the PR limited.

Recording failures with finer granularity would be an improvement; noted already in https://github.com/sagemath/sage/issues/33233#issuecomment-1418161492

mkoeppe:
> Does it make a better sense for the target application, such as modularization?
> 
> How is the mechanism used for testing a distribution package?

In #35095, I maintain `known-test-failures.json` files in `pkgs/*/sagemath-*`. 

The file-level mechanism has worked well enough so far, but going forward, it would also benefit from a finer granularity of this mechanism.

kwankyu:
> > Does it make a better sense for the target application, such as modularization?
> > How is the mechanism used for testing a distribution package?
> 
> In #35095, I maintain `known-test-failures.json` files in `pkgs/*/sagemath-*`.

Ah, that thing circling around my head was those files. 

> The file-level mechanism has worked well enough so far, but going forward, it would also benefit from a finer granularity of this mechanism.

I guess that that would be hard to maintain, though that has the benefit of keeping source files not cluttered, compared with the "known bug" mechanism.

Maybe each of both mechanisms has its own purpose, and we can keep both mechanisms.

mkoeppe:
> I guess that that would be hard to maintain

These ones are not manually edited; they are updated by copying over the stats files using `sage --fixdoctests --update-known-test-failures`

kwankyu:
I meant that if the baseline-stats mechanism works in finer granularity, that is, on individual doctests, then it would be complex to keep the info (like line numbers) recorded in the baseline-stats file in sync with the doctests in the source files.

mkoeppe:
I agree that going to the granularity of line numbers will not be a good idea.
But going to the granularity of functions could work well.

kwankyu:
OK. Thanks for explanation.

kwankyu:
> This last commit takes care of https://github.com/sagemath/sage/actions/runs/7306041656/job/19910309897?pr=36938#step:11:8651
> 
> ```
> sage -t --random-seed=239832805085423255575990885703796854524 src/sage/structure/coerce_actions.pyx
> **********************************************************************
> File "src/sage/structure/coerce_actions.pyx", line 784, in sage.structure.coerce_actions.IntegerMulAction.noexcept
> Failed example:
>     alarm(0.001); 2^(10^7) * P
> Expected:
>     Traceback (most recent call last):
>     ...
>     AlarmInterrupt
> Got:
>     (0 : 1 : 0)
>     AlarmInterrupt in doctesting framework
> ```

What is the proper way to deal with this? That looks just an unfortunate case...

How does the commit deal with this failure? Why this failure is related with this PR?

kwankyu:
Do you want the doctester to ignore the failure by adding the file to some known-failures.json ?

mkoeppe:
> How does the commit deal with this failure? Why this failure is related with this PR?

It's just a codepath in the doctester that so far has not been subject to the baseline mechanism. Various codepaths that are for internal failures of the doctesting framework are like this; but this particular codepath is triggered by doctests that test some timeout functionality.

The failure itself is already added to the known failures in https://github.com/sagemath/sage/pull/36937/files#diff-c6591fddad6cbbb6477ab5e8cc342b6c7b89b09f2707c7cc7aa818b52693448dR77

src/sage/doctest/reporting.py
```diff
@@ -497,22 +501,21 @@ def report(self, source, timeout, return_code, results, output, pid=None):
...
+                    postscript['lines'].append(self.report_head(source, fail_msg))
                     if hasattr(result_dict, 'tb'):
                         log(result_dict.tb)
                     if hasattr(result_dict, 'walltime'):
                         stats[basename] = {"failed": True, "walltime": wall, "ntests": ntests}
                     else:
                         stats[basename] = {"failed": True, "walltime": 1e6, "ntests": ntests}
-                    self.error_status |= 64
+                    if not baseline.get('failed', False):  # e.g. AlarmInterrupt in doctesting framework
+                        self.error_status |= 64
```
kwankyu, line 516: The comment is hard to understand. How about
```python
# We may get here by e.g. AlarmInterrupt in doctesting framework
if not baseline.get('failed', False):
    self.error_status |= 64
```
?

kwankyu:
Otherwise, lgtm.

src/sage/doctest/reporting.py
mkoeppe: Thanks; improved in f61cb9a17bc

kwankyu: Sorry for nitpicking. But one reason why the comment is hard to understand is that the comment is inside the if statement. The comment is about the whole if statement, I think.

kwankyu: We add a comment for the thing which is below or to the left. No? (Sometimes what I take as granted is not really granted...)

mkoeppe: You are right, I've fixed it in cf9024f816c

kwankyu (approved):
Thanks.

mkoeppe:
Thank you!

Issue #36860, opened 2023-12-11
`sagemath` is not available in Fedora 39


FilipLaurentiu:
### Steps To Reproduce

`sudo yum install sagemath`

### Expected Behavior

Install `sagemath`

### Actual Behavior

Sagemath is not available in Fedora 39 repository

`Last metadata expiration check: 0:07:16 ago on Mon 11 Dec 2023 01:01:27 PM EET.
No match for argument: sagemath
Error: Unable to find a match: sagemath`

### Environment

```markdown
➜  ~ lsb_release -a
LSB Version:	
Distributor ID:	Fedora
Description:	Fedora release 39 (Thirty Nine)
Release:	39
Codename:	ThirtyNine
➜  ~ uname -r
6.6.4-200.fc39.x86_64
➜  ~ sudo yum install sagemath 
Last metadata expiration check: 0:07:16 ago on Mon 11 Dec 2023 01:01:27 PM EET.
No match for argument: sagemath
Error: Unable to find a match: sagemath
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

enriqueartal:
When I told fedora developers that there was a problem with their nauty package, they pointed out that sagemath packager had abandoned the project, they wanted to know if some sagemath developer would be interested. I do not know rpm techniques, and actually it is simpler to compile it.

vincentmacri:
A couple people (including myself) are considering taking over as package maintainers for Sagemath on Fedora (https://discussion.fedoraproject.org/t/will-sagemath-return-to-the-fedora-39-repository/95662/17), but we're currently blocked by Sagemath not supporting Python 3.12, which is what ships with Fedora 39 (and Fedora doesn't allow packages to depend on older versions of Python than what ships with the OS by default).

vincentmacri:
I think #36023 and #36181 (as well as all the people interested in taking over maintaining it being new to package maintenance) are the main blockers.

dimpase:
Are you saying that Fedora 39 doesn't carry Python 3.11? Python 3.12.0 is quite buggy. Here is my favourite: https://github.com/python/cpython/issues/112215

OTOH #36023 should not be a big problem - there is a new polymake release which does support newer perl.
https://github.com/polymake/polymake/releases/tag/V4.11
Although polymake is an optional package of Sage, it need not be there for Sage to function otherwise.

dimpase:
[...]

PR #36937, opened 2023-12-20
CI Conda: Add to known-test-failures

mkoeppe:
Not all of the added failures are conda-specific, but the 6 instances of CI Conda amplify the error probability (https://github.com/sagemath/sage/pull/36694#issuecomment-1824871551)

We also add info (ideally a link to the GitHub Issue where the failure is reported) to the failures and arrange for it to be printed by the doctester as part of the `[failed in baseline]` message.

Marked as blocker so it takes effect in CI runs -- to reduce the noise from failures of the Conda CI that PR authors and reviewers find distracting.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36936 (which refactors the `[failed in baseline]` printing)

### DIFF
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -62,6 +62,9 @@
     "sage.rings.polynomial.skew_polynomial_finite_field": {
         "failed": true
     },
+    "sage.schemes.elliptic_curves.descent_two_isogeny": {
+        "failed": "random segfault (macOS) https://github.com/sagemath/sage/issues/36949"
+    },
     "sage.schemes.elliptic_curves.ell_field": {
         "failed": "random failure https://github.com/sagemath/sage/issues/36832"
     },
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -53,6 +53,9 @@
     "sage.schemes.elliptic_curves.ell_field": {
         "failed": true
     },
+    "sage.structure.coerce_actions": {
+        "failed": true
+    },
     "sage.tests.gap_packages": {
         "failed": true
     }
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -3,13 +3,13 @@
         "failed": true
     },
     "sage.algebras.fusion_rings.fusion_ring": {
-        "failed": true
+        "failed": "see https://github.com/sagemath/sage/pull/36372"
     },
     "sage.combinat.cluster_algebra_quiver.quiver": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6836592771/job/18591690058#step:11:10059"
     },
     "sage.geometry.cone": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6827937663/job/18571052628#step:11:12362"
     },
     "sage.groups.matrix_gps.finitely_generated_gap": {
         "failed": true
@@ -18,31 +18,31 @@
         "failed": true
     },
     "sage.libs.gap.element": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6840579851/job/18600012965#step:11:13016"
     },
     "sage.libs.singular.singular": {
         "failed": true
     },
     "sage.matrix.matrix2": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6835143781/job/18588599649#step:11:16939"
     },
     "sage.matrix.matrix_integer_sparse": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6827937663/job/18571052628#step:11:12362"
     },
     "sage.misc.lazy_import": {
         "failed": true
     },
     "sage.misc.weak_dict": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6870325919/job/18684964234#step:11:10059"
     },
     "sage.modular.modform.l_series_gross_zagier": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6859244281/job/18651257775#step:11:10102"
     },
     "sage.parallel.map_reduce": {
-        "failed": true
+        "failed": "random failure https://github.com/sagemath/sage/issues/36939"
     },
     "sage.rings.function_field.drinfeld_modules.morphism": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6840502530/job/18599835766#step:11:10107"
     },
     "sage.rings.polynomial.multi_polynomial_ideal": {
         "failed": true
@@ -51,19 +51,19 @@
         "failed": true
     },
     "sage.rings.polynomial.polynomial_element": {
-        "failed": true
+        "failed": "see https://github.com/sagemath/sage/pull/36372"
     },
     "sage.rings.polynomial.skew_polynomial_finite_field": {
         "failed": true
     },
     "sage.schemes.elliptic_curves.ell_field": {
-        "failed": true
+        "failed": "random failure https://github.com/sagemath/sage/issues/36832"
     },
     "sage.sets.recursively_enumerated_set": {
-        "failed": true
+        "failed": "see https://github.com/sagemath/sage/pull/36372"
     },
     "sage.structure.coerce_actions": {
-        "failed": true
+        "failed": "random failure https://github.com/sagemath/sage/issues/35973"
     },
     "sage.tests.gap_packages": {
         "failed": true
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -2,6 +2,9 @@
     "sage_setup.clean": {
         "failed": true
     },
+    "sage.algebras.fusion_rings.fusion_ring": {
+        "failed": true
+    },
     "sage.combinat.cluster_algebra_quiver.quiver": {
         "failed": true
     },
@@ -47,10 +50,16 @@
     "sage.rings.polynomial.multi_polynomial_libsingular": {
         "failed": true
     },
+    "sage.rings.polynomial.polynomial_element": {
+        "failed": true
+    },
     "sage.rings.polynomial.skew_polynomial_finite_field": {
         "failed": true
     },
     "sage.schemes.elliptic_curves.ell_field": {
+        "failed": true
+    },
+    "sage.sets.recursively_enumerated_set": {
         "failed": true
     },
     "sage.structure.coerce_actions": {
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -47,6 +47,9 @@
     "sage.rings.polynomial.skew_polynomial_finite_field": {
         "failed": true
     },
+    "sage.schemes.elliptic_curves.ell_field": {
+        "failed": true
+    },
     "sage.tests.gap_packages": {
         "failed": true
     }
--- a/.github/workflows/ci-conda-known-test-failures.json
+++ b/.github/workflows/ci-conda-known-test-failures.json
@@ -1,15 +1,18 @@
 {
+    "doc.en.constructions.calculus": {
+        "failed": "unreported random failures in plotting"
+    },
     "sage_setup.clean": {
-        "failed": true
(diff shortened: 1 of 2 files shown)


tobiasdiez:
Ignoring the whole file just because there is a single failing test runs against the goal of making the conda tests more stable. Hence motion to close. For the same reason this is not suitable as blocker.

mkoeppe:
> Ignoring the whole file just because there is a single failing test

That's not at all what it does, as we have explained to you at length in #36678.

We have also explained to you at length why your approach of #36372 is not suitable.
It simply has no support by any reviewer. 
It is highly inappropriate to use the blocker label to have it take effect in every PR.

mkoeppe:
As I said in https://github.com/sagemath/sage/pull/36372#issuecomment-1865326473 and before, Tobias, these continued abuses of your Triage team privileges have to stop.

kwankyu:
I think that for each record of a file in `.github/workflows/ci-conda-known-test-failures.json`, there should be a github issue associated with it so that we track the history of the failures in the file.

Is there an issue for 
```
    "sage.parallel.map_reduce": {
        "failed": true
    },
```
?

kwankyu:
and for 
```
    "sage.structure.coerce_actions": {
        "failed": true
    },
```
?

kwankyu:
Since json does not allow comments, we may add an "issue" field so that "issue: 12345" will point to the github issue with the number.

mkoeppe:
fine with me, but I think i'd prefer a free-form text that can contain an URL or whatever for flexibility.

kwankyu:
Yes, URL would work better.

mkoeppe:
i added this info where available to the commit messages, so you can see it via "git blame"

kwankyu:
> fine with me, but I think i'd prefer a free-form text that can contain an URL or whatever for flexibility.

Then
```
    "sage.structure.coerce_actions": {
        "failed": true
        "comment": "some text, preferably an URL"
    },
```
for example.

kwankyu:
> i added this info where available to the commit messages, so you can see it via "git blame"

My `git blame` only shows the sha and the committer's name and the date. But I can see them in the git log.

kwankyu:
@tobiasdiez Let's go this way. You may want to add those comments.

Matthias is working in https://github.com/sagemath/sage/pull/36558 to make these failures in baseline more visible to developers (without failing checks).

tobiasdiez:
Yeah sure, why wouldn't we want to exclude even more files from the conda tests? For consistency, you should also add a similar file-wide exclusion mechanism for the tests that are known to fail sometimes in the Build & Test.

So stupid, instead of viewing the conda environment as a means to harden the test suit, we now massively swap the failures under the carpet. And that in a way that doesn't even allow to answer the simple question "is this the only example that fails, or are other examples failing as well"!

I also don't understand why I should add these comments. My solution keeps very well track of what example fails. It's one of the major shortcomings of Mathias solution, as I've pointed out multiple times. So at the very least he should do the work himself.

mkoeppe:
You are repeating points that have already been refuted in the original discussion.

mkoeppe:
> For consistency, you should also add a similar file-wide exclusion mechanism for the tests that are known to fail sometimes in the Build & Test.

Yes, we may want to do this at some point. But not in this PR.

mkoeppe:
> Then
> 
> ```
>     "sage.structure.coerce_actions": {
>         "failed": true
>         "comment": "some text, preferably an URL"
>     },
> ```
> 
> for example.

I'm instead just putting it in `"failed"` instead of `true`.

mkoeppe:
Tobias, stop this nonsense.

.github/workflows/ci-conda-known-test-failures.json
```diff
@@ -1,12 +1,18 @@
...
     "sage_setup.clean": {
-        "failed": true
+        "failed": "_find_stale_files finds some stale files under sage/tests when executed under conda"
+    },
+    "sage.algebras.fusion_rings.fusion_ring": {
+        "failed": "unreported random timeouts"
     },
     "sage.combinat.cluster_algebra_quiver.quiver": {
-        "failed": true
+        "failed": "failure seen in https://github.com/sagemath/sage/actions/runs/6836592771/job/18591690058#step:11:10059"
```
tobiasdiez, line 12: These comments are pretty useless as the logs are deleted quite quickly (90 days?)

tobiasdiez:
Removing "blocker" in favor of https://github.com/sagemath/sage/pull/36372 as this PR here still contains a lot of outdated exclusions.

Moreover, I would like to ask you to add these exceptions less literately. With #36372 I only added them once I had two independent PRs that show a given error. Here is my list of failures that I've seen only once:
```
sage -t --random-seed=214436596912198344130626509908513694044 src/sage/misc/latex.py  # 1 doctest failed

sage -t --random-seed=295417759961273984682997709208968154147 src/sage/rings/finite_rings/integer_mod.pyx  # Timed out (macos?)

sage -t --random-seed=329096063982609013558495029873741838049 src/sage/interfaces/giac.py  # 1 doctest failed

 sage -t --random-seed=128826216443334456408110059374142917558 src/sage/rings/number_field/number_field_element_quadratic.pyx  # 1 doctest failed

sage -t --random-seed=92781704964877147605342482066320289765 src/sage/schemes/elliptic_curves/ell_rational_field.py  # 1 doctest failed

sage -t --random-seed=142732370142697541280655968131944483835 src/sage/graphs/generic_graph.py  # Timed out (on macos)

sage -t --random-seed=145162806241242408148129167207410356033 src/sage/matrix/matrix2.pyx  # Timed out

sage -t --random-seed=114726989967833111405585804986834164322 src/doc/en/constructions/calculus.rst  # Timed out (on macos)

 sage -t --random-seed=3369460376730320sage -t --random-seed=73515522912035769254887028411671928367 src/sage/schemes/elliptic_curves/ell_field.py
File "src/sage/schemes/elliptic_curves/ell_field.py", line 1029, in sage.schemes.elliptic_curves.ell_field.EllipticCurve_field.division_field
Failed example:
    K = E.division_field(l)13006113016408769185426 src/sage/rings/polynomial/pbori/pbori.pyx  # 1 doctest failed
on macos
```

mkoeppe:
Tobias, stop this nonsense with the labels.

orlitzky:
Are these heisenbugs? Why fight over how to ignore them rather than fix them?

mkoeppe:
Yes, many of them are sporadic random failures, some of them probably known for many years. Individual developers may not be familiar enough with the code in the affected modules to fix them.

Seasoned developers are of course already trained to ignore such failures. But with the proliferation of many new ways that Sage is tested automatically on GH Actions, there appears to be a lot of uncertainty among developers where to look, which of the workflows and which of the failures matter etc., see for example https://groups.google.com/g/sage-devel/c/xjsqw-uundU

Hence the priority of this PR to reduce the noise that comes from the Conda CI with its 6 runs, and thus hopefully to remove the extra friction for developers.

kwankyu:
@mkoeppe @tobiasdiez

I will play an arbiter role here. I hope that both of you agree on the following steps to resolve this deadlock situation.

1. @tobiasdiez remove "blocker" label from #36960.
2. @tobiasdiez set this PR #36937 as "blocker" (yes, already it is), and review the PR (double check the failed files list without assuming #36960).
3. @tobiasdiez close PR #36372.
4. @volker merge #36937 to the next beta.
5. @tobiasdiez rebase PR #36960 on #36937, and edits items in `.github/workflows/ci-conda-known-test-failures.json` instead of removing the file itself.
6. @mkoeppe review PR #36960, and set the PR as "blocker".
7.  @volker merge #36960 to the next beta.
8. @mkoeppe make a new PR removing `.github/workflows/ci-conda-known-test-failures.json` if it is and will be empty.
9. We all live happily ever after.

Let me know if both of you agree on the plan.

tobiasdiez:
>     3. @tobiasdiez close PR [Fix/ignore all remaining test failures when run under conda #36372](https://github.com/sagemath/sage/pull/36372).

Okay, done.

>     5. @tobiasdiez rebase PR [Introduce os-dependent doctest tags `# known bug: macos`, `# known bug: linux` #36960](https://github.com/sagemath/sage/pull/36960) on [CI Conda: Add to known-test-failures #36937](https://github.com/sagemath/sage/pull/36937), and edits items in `.github/workflows/ci-conda-known-test-failures.json` instead of removing the file itself.

I did that. The file was empty afterwards, that's why I've removed it (more details at https://github.com/sagemath/sage/pull/36960#issuecomment-1871121215).

kwankyu:
@tobiasdiez Do you agree with the plan, that is, the steps from 1 to 9?

tobiasdiez:
> @tobiasdiez Do you agree with the plan, that is, the steps from 1 to 9?

I think we can directly jump to 6, as this PR here is made obsolete by https://github.com/sagemath/sage/pull/36960

kwankyu:
> > @tobiasdiez Do you agree with the plan, that is, the steps from 1 to 9?
> 
> I think we can directly jump to 6, as this PR here is made obsolete by #36960

OK. No deal.

tobiasdiez:
> > > @tobiasdiez Do you agree with the plan, that is, the steps from 1 to 9?
> > 
> > 
> > I think we can directly jump to 6, as this PR here is made obsolete by #36960
> 
> OK. No deal.

I'm okay if you feel like spending the time to make this PR ready first. My reviewer comments here are:
- https://github.com/sagemath/sage/pull/36937#discussion_r1435641278: replace the links to the gh runs by links to issues (that need to be created first in some cases)
- only exclude the files touched in https://github.com/sagemath/sage/pull/36960. All other files are passing (in my experience).

Once the second point is done, I'm okay with removing the "blocker" label from https://github.com/sagemath/sage/pull/36960, and will wait for the completion of step 1 to 5.

.github/workflows/ci-conda-known-test-failures.json
mkoeppe: Yes, 90 days. Enough time for people to investigate and open Issues / PRs.

Not helpful to demand that it's done here.

mkoeppe:
> * only exclude the files touched in [Introduce os-dependent doctest tags `# known bug: macos`, `# known bug: linux` #36960](https://github.com/sagemath/sage/pull/36960). All other files are passing (in my experience).

The point with these random errors is that no single developer's "experience" is sufficient to say that; we want to eliminate the noise from sporadic, annoying failures.

tobiasdiez:
> > * only exclude the files touched in [Introduce os-dependent doctest tags `# known bug: macos`, `# known bug: linux` #36960](https://github.com/sagemath/sage/pull/36960). All other files are passing (in my experience).
> 
> The point with these random errors is that no single developer's "experience" is sufficient to say that; we want to eliminate the noise from sporadic, annoying failures.

Right, and there never have been any sporadic failures in sage.misc.lazy_import or sage.libs.singular.singular (among others); moreover a few other sporadic failures likely have been fixed in the mean time - but since you constantly remove the blocker label from #36960 it is hard to confirm this. So please either don't touch the blocker label or change this PR accordingly
(or go through all conda logs and see if the errors still appear but are suppressed).

mkoeppe:
The point here is that we reduce the noise for **all** developers on unrelated PRs from the conda workflows. 
Developers currently do not get any value from the conda workflows, only noise. 
Marking things as baseline failures, even it marks more than necessary, is a mild and unproblematic solution; milder than just administratively disabling the conda workflow.
That's why this PR here is a blocker, and it should be merged ASAP.

The concern of **one** developer to confirm whether the failures marked up in https://github.com/sagemath/sage/pull/36960 cover everything is secondary. Monitoring this can be done just using the logs and does not have to rely on signaling test failures to all other developers.

tobiasdiez:
> The concern of **one** developer to confirm whether the failures marked up in #36960 cover everything is secondary. Monitoring this can be done just using the logs and does not have to rely on signaling test failures to all other developers.

I now went through the logs and added a few more exceptions to #36960. So every failure (that occurred more than once) is now excluded in #36860. Could you please mirror the changes here. Thanks!

.github/workflows/ci-conda-known-test-failures.json
tobiasdiez: You missed step 2 of your own procedure (https://github.com/sagemath/sage/pull/36678#issuecomment-1837386861):
>
    Test failure is noticed
    You record it in a GitHub Issue. There you include the necessary information so that it does not take 2 hours to reconstruct it.
    One adds stuff to the exclusion list to eliminate noise in the tests. It's not for humans to read, and definitely not for humans to use 2 hours of guessing time to reconstruct the failure.
    The exclusion list is maintained like everything else in Sage -- by PRs, not by specific people in charge.

mkoeppe:
Completeness is not part of the scope of this PR.

mkoeppe:
> I now went through the logs and added a few more exceptions to #36960.

Thanks for doing this work. Do you have script for this - retrieving and parsing logs? 

You may be interested in:
- #36970

mkoeppe:
Let's merge this please.

kwankyu (approved):
OK. Hopefully save us from chronic conda failures in CI.

PR #36938, opened 2023-12-21
CI Build & Test: Show doctest failures, warnings as annotations in the 'Files changed' tab

mkoeppe:
This prints doctest failures as github annotations. 

For illustrating this new feature, the branch on #36558 provokes some errors.

As seen in the "Files changed" tab over there (https://github.com/sagemath/sage/pull/36558/files):
- Doctest failures show as errors next to the doctest in source code
- Doctest warnings show as "warnings"
- Doctest failures from files that "failed in baseline" show as "notices"

These messages can also be seen in the Summary page of the workflow (https://github.com/sagemath/sage/actions/runs/7359388426?pr=36558). Clicking on a message leads to the source code location.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36936


tscrim:
In the sample, errors appear twice.

Is there also some way to disable this when viewing the diff online? This can get annoying when wanting to look at the code as a whole (without having to do it locally; just like the codecov reports).

mkoeppe:
> In the sample, errors appear twice.

That's probably from first testing the changed files. I'll disable the annotations from there.

mkoeppe:
> Is there also some way to disable this when viewing the diff online? 

No, it doesn't seem so.

> This can get annoying when wanting to look at the code as a whole (without having to do it locally; just like the codecov reports).

That's a valid concern.
Still worth trying?

tscrim:
> > This can get annoying when wanting to look at the code as a whole (without having to do it locally; just like the codecov reports).
> 
> That's a valid concern. Still worth trying?

Probably since we already have the codecov cluttering things up. Although a quick notice to sage-devel should be done in case this would cause controversy as it has potential workflow implications.

.github/workflows/build.yml
```diff
@@ -185,14 +185,14 @@ jobs:
       - name: Test all files (sage -t --all --long)
         if: (success() || failure()) && steps.build.outcome == 'success'
         run: |
-          ../sage -python -m pip install coverage
-          ../sage -python -m coverage run ./bin/sage-runtests --all --long -p2 --random-seed=286735480429121101562228604801325644303
-        working-directory: ./worktree-image/src
+          ./sage -python -m pip install coverage
+          ./sage -python -m coverage run ./src/bin/sage-runtests --all --long -p2 --format github --random-seed=286735480429121101562228604801325644303
+        working-directory: ./worktree-image
```
tobiasdiez, line 202: That change doesn't seem to work, and is also unrelated to the actual goal of the PR.

.github/workflows/build.yml
```diff
@@ -185,14 +185,14 @@ jobs:
       - name: Test all files (sage -t --all --long)
         if: (success() || failure()) && steps.build.outcome == 'success'
         run: |
-          ../sage -python -m pip install coverage
-          ../sage -python -m coverage run ./bin/sage-runtests --all --long -p2 --random-seed=286735480429121101562228604801325644303
-        working-directory: ./worktree-image/src
+          ./sage -python -m pip install coverage
+          ./sage -python -m coverage run ./src/bin/sage-runtests --all --long -p2 --format github --random-seed=286735480429121101562228604801325644303
```
tobiasdiez: Would it be possible to autodetect that the tests are run in a GH environment, and then automatically format them correctly?

tobiasdiez (changes requested):
Nice, happy to see my idea works.

Could you please implement the hack of https://github.com/actions/toolkit/issues/193#issuecomment-605394935 to properly display multiline strings.

.github/workflows/build.yml
mkoeppe: > unrelated to the actual goal of the PR.

It's necessary for the printed paths to be correct. Otherwise annotations don't work.

mkoeppe: > That change doesn't seem to work

Please, Tobias, include the necessary information when you say that something doesn't work.

.github/workflows/build.yml
mkoeppe: That wouldn't be an improvement. The workflow is in charge of requesting these annotations; I don't want to bake this policy into the doctester itself. It would also probably break the doctests of `sage.doctest`.

mkoeppe:
> Could you please implement the hack of [actions/toolkit#193 (comment)](https://github.com/actions/toolkit/issues/193#issuecomment-605394935) to properly display multiline strings.

Thanks for finding this. Implemented in b83dcf3; tests running in https://github.com/sagemath/sage/pull/36558, let's see what it looks like.

mkoeppe:
Unfortunately, it only affects the Summary view, not the diff view.

mkoeppe:
> Unfortunately, it only affects the Summary view, not the diff view.

No, with the latest changes, it works also in the diff view! https://github.com/sagemath/sage/pull/36558/files

mkoeppe:
coverage.py still broken

.github/workflows/build.yml
mkoeppe: I assume you meant that the next step, "Prepare coverage results", failed. That's fixed now.

mkoeppe:
Ready for review. (Failure in the conda test is unrelated.)

alexjbest:
I think this is a very nice and useful feature, and I am glad that Sage is enabling it.

I think it can be more simply implemented using problem matchers (https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) though, rather than baking in github specific output formats into Sage itself.
For example https://github.com/actions/setup-python/blob/main/.github/python.json should work almost out of the box if I understand correctly.

Did you consider this approach already?

mkoeppe:
> Did you consider this approach already?

Yes, in #36545; I gave up on it because the multi-line format of our doctest failures.

.github/workflows/build.yml
```diff
@@ -187,25 +187,25 @@ jobs:
       - name: Pytest
         if: contains(github.ref, 'pytest')
         run: |
-          ../sage -python -m pip install coverage pytest-xdist
-          ../sage -python -m coverage run -m pytest -c tox.ini --doctest || true
-        working-directory: ./worktree-image/src
+          ./sage -python -m pip install coverage pytest-xdist
```
tobiasdiez: What's the reason for these changes to pytest and did you test them?

mkoeppe: I don't remember. I've backed out the change now in 0c52169.

alexjbest (approved):

mkoeppe:
Thanks, Alex!

tobiasdiez:
At the end of the changed files list, there are now a few annotated warnings that are not related to PR (can also be seen here). Can you please fix this?

mkoeppe:
That doctest *failures* in unchanged files are annotated is a feature, not a bug. 

The doctest failure in `sage_setup.clean` is fixed in #37185.

To limit annotation of the doctester's *warnings* (for `# optional/needs` tags) to the files modified in the PR would be possible, by putting the run of `sage -t --new` in charge of emitting them. I've noted this as #37235, a wishlist item.
But it's easier to just fix these warnings.

tobiasdiez:
> But it's easier to just fix these warnings.

I agree. Could you please open a PR for this?

PR #36940, opened 2023-12-21
new method "fraction" in integer-valued polynomials

fchapoton:
This is adding a new useful method to this class.

Also fixing the h-polynomial method by adding a reversal there.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/polynomial/integer_valued_polynomials.py
+++ b/src/sage/rings/polynomial/integer_valued_polynomials.py
@@ -860,13 +860,25 @@ class IntegerValuedPolynomialRing(UniqueRepresentation, Parent):
                     sage: f = ex.fraction();f
                     -1/(t^5 - 5*t^4 + 10*t^3 - 10*t^2 + 5*t - 1)
 
-                    sage: F = QQ[['t']]
-                    sage: F(f).O(6)
-                    1 + 5*t + 15*t^2 + 35*t^3 + 70*t^4 + 126*t^5 + O(t^6)
+                    sage: F = LazyPowerSeriesRing(QQ, 't')
+                    sage: F(f)
+                    1 + 5*t + 15*t^2 + 35*t^3 + 70*t^4 + 126*t^5 + 210*t^6 + O(t^7)
 
                     sage: poly = ex.polynomial()
                     sage: [poly(i) for i in range(6)]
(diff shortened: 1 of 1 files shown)


src/sage/rings/polynomial/integer_valued_polynomials.py
```diff
@@ -832,18 +831,48 @@ def h_polynomial(self):
...
+                Return the generating series of values as a fraction.
+
+                In the case of Ehrhart polynomials, this is known as
+                the Ehrhart series.
+
+                .. SEEALSO:: :meth:`h_vector`, :meth:`h_polynomial`
+
+                EXAMPLES::
+
+                    sage: A = IntegerValuedPolynomialRing(ZZ).S()
```
tscrim, line 858: It's slightly weird to have this be called `A` instead of `S`. Nothing to you need to change, but it caught my eye.

src/sage/rings/polynomial/integer_valued_polynomials.py
```diff
@@ -832,18 +831,48 @@ def h_polynomial(self):
...
+                    sage: ex = A.monomial(4)
+                    sage: f = ex.fraction();f
+                    -1/(t^5 - 5*t^4 + 10*t^3 - 10*t^2 + 5*t - 1)
```
tscrim, line 861: Can we also add a test where the numerator is not a constant?

Can you check the corner case for the h-vector being empty for 0 and returning an element in the same parent?

src/sage/rings/polynomial/integer_valued_polynomials.py
```diff
@@ -832,18 +831,48 @@ def h_polynomial(self):
...
+
+                    sage: F = QQ[['t']]
+                    sage: F(f).O(6)
```
tscrim: Shameless plug, but I would use the lazy power series ring. :)

tscrim (changes requested):

src/sage/rings/polynomial/integer_valued_polynomials.py
fchapoton: this is for A like algebra ; it's like that everywhere else in the file

tscrim (approved):
Thank you. LGTM.

Sorry for letting this fall off my radar!

PR #36955, opened 2023-12-24
build/pkgs/ninja_build: support samurai version scheme

orlitzky:
Samurai is a C99 ninja implementation with an almost-compatible version scheme, except that it has only two version components instead of the three that ninja has. We update the "sed" call used to parse the version number out of `ninja --version` so that it can parse a samurai version too.

This should only matter on systems where (for example) /usr/bin/ninja points to samurai. That's not typical, but it recently became possible to do on Gentoo in an "official" way.


orlitzky:
I've been abusing the `ac_cv` cache variable to do this for a long time, but now `app-alternatives/ninja` exists to manage the `/usr/bin/ninja` symlink.

mkoeppe (approved):
LGTM

PR #36956, opened 2023-12-24
Cache matroid bases and circuits upon computation

### LINKED ISSUE #36820: Cache matroid circuit computation
### Problem Description

The current implementation of matroid circuit computations does not cache the result. 

      sage: M = matroids.CompleteGraphic(7)
      ....: 
      ....: %time C1 = M.circuits();
      ....: 
      ....: %time C2 = M.circuits();
      CPU times: user 5.65 s, sys: 26.4 ms, total: 5.68 s
      Wall time: 5.72 s
      CPU times: user 5.81 s, sys: 47 ms, total: 5.86 s
      Wall time: 6.14 s

### Proposed Solution

It would be nice to cache the result rather than recomputing it. 

### Alternatives Considered

An alternative is to require the user to know that this is not cached and instead manually store it wherever makes sense when doing a relevant computation.

### Additional Information

_No response_

gmou3:
By adding two empty SetSystem placeholders ._B (for bases) and ._C (for circuits), the bases and circuits can be cached upon 1st computation. This significantly accelerates future calls to the .bases() or .circuits() functions.

This PR fixes #36820:

```
sage: M = matroids.CompleteGraphic(7)
sage: %time C1 = M.circuits()
CPU times: user 12 s, sys: 0 ns, total: 12 s
Wall time: 12 s
sage: %time C2 = M.circuits()
CPU times: user 24 µs, sys: 1 µs, total: 25 µs
Wall time: 37.9 µs
```

### :memo: Checklist

### DIFF
--- a/src/sage/matroids/basis_exchange_matroid.pyx
+++ b/src/sage/matroids/basis_exchange_matroid.pyx
@@ -1655,7 +1655,9 @@ cdef class BasisExchangeMatroid(Matroid):
             sage: len([B for B in M.bases()])
             184
         """
-        return self.independent_r_sets(self.full_rank())
+        if not self._B:
+            self._B = self.independent_r_sets(self.full_rank())
+        return self._B
 
     cpdef dependent_r_sets(self, long r) noexcept:
         """
@@ -1893,31 +1895,33 @@ cdef class BasisExchangeMatroid(Matroid):
              ['c', 'f', 'g'], ['d', 'e', 'f', 'g']]
         """
         cdef SetSystem NSC
-        NSC = SetSystem(self._E)
-        if self._groundset_size == 0:
-            return NSC
-        bitset_clear(self._input)
-        bitset_set_first_n(self._input, self._matroid_rank)
         cdef long e, f
-        repeat = True
-        while repeat:
-            if self.__is_independent(self._input):
-                bitset_complement(self._input2, self._current_basis)
-                e = bitset_first(self._current_basis)
-                while e >= 0:
-                    self.__fundamental_cocircuit(self._output, e)
-                    if e > bitset_first(self._output):
-                        bitset_intersection(self._input2, self._input2, self._output)
-                    e = bitset_next(self._current_basis, e + 1)
-                f = bitset_first(self._input2)
-                while f >= 0:
-                    self.__fundamental_circuit(self._output, f)
-                    if f == bitset_first(self._output):
-                        NSC._append(self._output)
-                    f = bitset_next(self._input2, f + 1)
-            repeat = nxksrd(self._input, self._groundset_size, self._matroid_rank, True)
-        NSC.resize()
-        return NSC
+        if not self._C:
+            NSC = SetSystem(self._E)
+            if self._groundset_size == 0:
+                return NSC
+            bitset_clear(self._input)
+            bitset_set_first_n(self._input, self._matroid_rank)
+            repeat = True
+            while repeat:
+                if self.__is_independent(self._input):
+                    bitset_complement(self._input2, self._current_basis)
+                    e = bitset_first(self._current_basis)
+                    while e >= 0:
+                        self.__fundamental_cocircuit(self._output, e)
+                        if e > bitset_first(self._output):
+                            bitset_intersection(self._input2, self._input2, self._output)
+                        e = bitset_next(self._current_basis, e + 1)
+                    f = bitset_first(self._input2)
+                    while f >= 0:
+                        self.__fundamental_circuit(self._output, f)
+                        if f == bitset_first(self._output):
+                            NSC._append(self._output)
+                        f = bitset_next(self._input2, f + 1)
+                repeat = nxksrd(self._input, self._groundset_size, self._matroid_rank, True)
+            NSC.resize()
+            self._C = NSC
+        return self._C
 
     # isomorphism
 
--- a/src/sage/matroids/basis_matroid.pyx
+++ b/src/sage/matroids/basis_matroid.pyx
@@ -601,17 +601,19 @@ cdef class BasisMatroid(BasisExchangeMatroid):
             28
         """
         cdef long r, n
-        r = self.full_rank()
-        n = len(self)
         cdef SetSystem BB
-        BB = SetSystem(self._E, capacity=bitset_len(self._bb))
         cdef long b
-        b = bitset_first(self._bb)
-        while b >= 0:
-            index_to_set(self._b, b, r, n)
-            BB._append(self._b)
-            b = bitset_next(self._bb, b + 1)
-        return BB
+        if not self._B:
+            r = self.full_rank()
+            n = len(self)
+            BB = SetSystem(self._E, capacity=bitset_len(self._bb))
+            b = bitset_first(self._bb)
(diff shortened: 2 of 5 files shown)


tscrim:
I would store it internally as a `SetSystem` but always return a `list` copy for the corresponding methods. This will keep backwards compatibility, should remain fast, and can never mutate the output (as a `SetSystem` is mutable).

gmou3:
Indeed it was a list, although the bases are returned as a `SetSystem`, which is a weird inconsistency. I can easily revert this if you wish.

More generally on this note, I was thinking of having `bases()`, `circuits()`, etc., always return a `list` of `sets` (**not** `frozensets`) sorted first by length and then by lexicographic ordering (i.e., like a human would probably like to read them). So, when doing something like `M.bases()` you would get a direct and more user-friendly output (rather than having to do `list(M.bases())` to see them in an unordered fashion and with the word `frozenset` everywhere). In order to avoid overhead (e.g. from sorting) we can always include _bases(), _circuits(), etc., which would be following the current implementation.

tscrim:
It is a weird inconsistency, but the change might break some code in the wild and (more importantly) it is not returning an immutable output.

I am generally opposed to such return orders as they can be used internally by things that do not care about such orderings. I would let the user decide (with perhaps a doctest showing how to sort). I don't see a sufficiently good reason to make things more complicated with such hidden methods.

gmou3:
> It is a weird inconsistency, but the change might break some code in the wild and (more importantly) it is not returning an immutable output.
> 
Aren't bases mutable though, being a `SetSystem`? What principle should we follow?

> I am generally opposed to such return orders as they can be used internally by things that do not care about such orderings. I would let the user decide (with perhaps a doctest showing how to sort). I don't see a sufficiently good reason to make things more complicated with such hidden methods.
>
Personally, as a user of these functions, I have already wasted a significant amount of time typing `list()` and `sorted()` after getting the output `Iterator over a system of subsets`; and that only to get an unordered list filled with the sheer ugliness and noise of the omnipresence of `frozenset`. (I know I basically repeated the same point, but I am doing so for emphasis.) Appropriately sorting is not so trivial and easily typable each time one wants to see the bases or circuits. For example, for the circuits one first needs to sort based on length and then based on the lexicographic order of the (already internally sorted) set. It requires something like this:

`sorted([sorted(C) for C in M.circuits()], key=lambda S: (len(S), S))`.

By doing this in `bases()` and `circuits()` we make sure to always get the same easy-on-the-eyes output. And if you can compute `n`, then you can compute `nlogn` in my opinion (:P). These would be considered `print` functions, and for internal use we would only use `_bases()` and `_circuits()`. Admittedly, this is an extra complication for the programmer, but - first - this is less important than the associated benefit (or so I claim), and - secondly - I am willing to implement this.

In any case, please excuse my persistence. Most importantly for this PR I would like your opinion on the first issue about the type of the circuits.

tscrim:
> > It is a weird inconsistency, but the change might break some code in the wild and (more importantly) it is not returning an immutable output.
> 
> Aren't bases mutable though, being a `SetSystem`? What principle should we follow?

The output is mutable but those for the matroid are not. By caching them, you can easily create an inconsistency. You get the bases `B` of a matroid $M$, then you modify `B` for some computation you are doing, then with the caching, $M$ now thinks it has different bases (precisely those in the modified `B`) than it should have (say, via its maximal independent sets).

> > I am generally opposed to such return orders as they can be used internally by things that do not care about such orderings. I would let the user decide (with perhaps a doctest showing how to sort). I don't see a sufficiently good reason to make things more complicated with such hidden methods.
> 
> Personally, as a user of these functions, I have already wasted a significant amount of time typing `list()` and `sorted()` after getting the output `Iterator over a system of subsets`; and that only to get an unordered list filled with the sheer ugliness and noise of the omnipresence of `frozenset`. (I know I basically repeated the same point, but I am doing so for emphasis.) Appropriately sorting is not so trivial and easily typable each time one wants to see the bases or circuits. For example, for the circuits one first needs to sort based on length and then based on the lexicographic order of the (already internally sorted) set. It requires something like this:
> 
> `sorted([sorted(C) for C in M.circuits()], key=lambda S: (len(S), S))`.
> 
> By doing this in `bases()` and `circuits()` we make sure to always get the same easy-on-the-eyes output.

Remember that beauty is in the eye of the beholder.

> And if you can compute `n`, then you can compute `nlogn` in my opinion (:P).

Yes, but I think we are both aware that is not actually true. `;)`

> These would be considered `print` functions, and for internal use we would only use `_bases()` and `_circuits()`

That is not the case. The user would *not* know about the internal use methods, but for users who want to do fast computations because they are not looking at the output, they would not want to spend extra CPU cycles on sorting.

>. Admittedly, this is an extra complication for the programmer, but - first - this is less important than the associated benefit (or so I claim), and - secondly - I am willing to implement this.

It's not about the implementation but the maintenance and general-purpose utility. I mostly don't agree with your benefit claim here. However, as a middle ground, I would not be opposed to having an additional `sort` argument whose default is `False`.

> In any case, please excuse my persistence. Most importantly for this PR I would like your opinion on the first issue about the type of the circuits.

Actually, there are some reasons to go in the opposite direction for the types: where we instead just iterate over the output to reduce the memory footprint (this is especially important for matroids with large number of bases). This is what the graph theory portion has been moving towards. Of course, there is a speed/memory trade-off happening here. When you peer into the implementation, it is generally (perhaps always) done by iterators. That being said, some implementations of circuits do return a `SetSystem` anyways, so +1 on making that uniform.

gmou3:
So, it seems that internally we would like something immutable, like a `frozenset` of `frozensets`. When outputting, and in order to have consistency, we can at every call of circuits() create a `SetSystem` to output.

As for the comments regarding whether it should be an iterator, it seems like this was indeed the ultimate intention. Quoting from `set_system.pyx`:

> The only guaranteed behavior of this class is that it is iterable. It
> is expected that M.circuits(), M.bases(), and so on will in the near
> future return actual iterators.

gmou3:
I will close this PR. Caching is a bad idea in general and should be left to the user to do consciously. Every enumerator function (bases(), circuits(), etc.) should become an actual iterator. A new CircuitsMatroid class is needed because now there is no matroid subclass internally defined by circuits (!). I have mostly implemented these already and I am waiting for other PRs to be reviewed.

tscrim:
While these methods return a `SetSystem`, I don't think it is necessarily a bad thing to cache the results. It is just a trade-off. I hope you didn't think I was against this PR.

gmou3:
No worries, I was actually convinced that caching is to be left to the user. Each matroid class should hold only its particular description that defines the matroid. That seems to be the general architecture of the code, and it is wise in terms of memory consumption, as moving from one description to another can be "explosive". If you define a matroid succinctly, e.g. in terms of circuit closures, and then you attempt to call bases(), the code must return an iterator. That seems to be the most robust approach.

tscrim:
Sounds like a good plan.

On a related note: One way to avoid backwards incompatible changes is to have a `foo_iterator()` method that `foo()` simply `return list(self.foo_iterator())`. Although it might just be better to accept the pain with apologies to the matroid users...

gmou3:
Let there be pain.

PR #36923, opened 2023-12-19
Revert "gh-36678: CI conda: Ignore baseline test failures "

tobiasdiez:
Reverts #36678, which was marked as "positive review" although not all reviewer concern had been addressed.

I'll set this directly as "positive review" since this is a trivial git revert.

### :memo: Checklist

- [ ] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #36769, opened 2023-11-25
Do not fail when `java` is available but `jmol` is not.

tornaria:
If java is not available, sagemath will fallback to tachyon gracefully.

When java is available, sagemath will assume jmol is available and error if not. This commit fixes the issue by implementing a method `is_jmol_available()` to replace `is_jvm_available()`.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have created tests covering the changes.

Issue #35715, opened 2023-06-03
Random testing has revealed a problem in test_karatsuba_multiplication


vbraun:
### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

### Did you read the documentation and troubleshoot guide?

- [X] I have read the documentation and troubleshoot guide

### Environment

```markdown
- **OS**: Ubuntu 18.04 32 bit
- **Sage Version**: 10.1.beta1
```

### Steps To Reproduce

Random fail when running the tests, just saw it for the first time on a build bot:
```
**********************************************************************
File "src/sage/rings/tests.py", line 449, in sage.rings.tests.?
Failed example:
    test_karatsuba_multiplication(QQbar, 3, 3, numtests=2)    # long time     # optional - sage.rings.number_field
Expected nothing
Got:
    Random testing has revealed a problem in test_karatsuba_multiplication
    Please report this bug!  You may be the first
    person in the world to have seen this problem.
    Please include this random seed in your bug report:
    Random seed: 164480309984640964737929113640888706047
    ValueError('Multiplication failed')
**********************************************************************
1 item had failures:
   1 of  13 in sage.rings.tests.?
    [62 tests, 1 failure, 25.59 s]
----------------------------------------------------------------------
sage -t --long --random-seed=0 src/sage/rings/tests.py  # 1 doctest failed
----------------------------------------------------------------------
```

### Expected Behavior

Doctest should not fail

### Actual Behavior

Doctest fails (randomly, seed is provided above)

### Additional Information

probably due to #35443

vbraun:
@tornaria FYI

tobiasdiez:
More info on how to reproduce this (or a very closely related error:: https://github.com/sagemath/sage/issues/36961

mezzarobba:
Also here: https://github.com/sagemath/sage/pull/37193#issuecomment-1917520429

GiacomoPope:
Another example of this:

https://github.com/sagemath/sage/pull/37441#issuecomment-2045370332

```
sage: from sage.rings.tests import test_karatsuba_multiplication ## line 429 ##
sage: test_karatsuba_multiplication(ZZ, 6, 5, verbose=True, seed=42) ## line 430 ##
test_karatsuba_multiplication: ring=Univariate Polynomial Ring in x over Integer Ring, threshold=2
[...]

PR #36960, opened 2023-12-25
Introduce os-dependent doctest tags `# known bug: macos`, `# known bug: linux`

tobiasdiez:
Extend the `known bug` doctest tag by the possibility to hide these doctest errors only a certain platform. Use this to hide a few known (mostly randomly occurring) bugs. Also convert a few 'random' tags to 'known bug' if random was used in the sense "on a certain architecture, or with a certain dependency, this fails"`.

Considered alternatives:
- A new (perhaps os-dependent) `skip` tag. In most cases, known bug is however better name since these failures point to actual bugs. But some output is merely architecture dependent and in this case a generic `skip` tag would be better.
- A possibility to only skip these failing tests on ci (e.g. by detecting the `CI` env variable, or by explicitly specifying a baseline as in #36678). This has the disadvantage that these known bugs may randomly popup on developers machines without adding any value in this case.
- A new `random failure` or `known failure` tag with essentially the same meaning as `known bug` but limiting it to random occurrence. This idea didn't got much support in #36372, and the randomness can easily be specified by an additional comment.

Setting to blocker to hide #35715 and #36992 which annoyed devs before https://github.com/sagemath/sage/pull/36769#issuecomment-1826973821. Also hides another unreported test failure in `sage_setup/clean.py`. 

### :memo: Checklist

### :hourglass: Dependencies

- #36923 as the baseline stats are no longer needed with this PR


mkoeppe:
Please adjust the scope of this PR. There is no need to back out `.github/workflows/ci-conda-known-test-failures.json` for the goals of this PR, so it should not be done here.

src/doc/en/constructions/calculus.rst
```diff
@@ -387,7 +387,7 @@ are obtained via ``fourier_series_partial_sum``::
     -6/5/pi
     sage: s5 = f.fourier_series_partial_sum(5); s5
     -6/5*sin(10*x)/pi - 2*sin(6*x)/pi - 6*sin(2*x)/pi + 1/2
-    sage: plot(f, (0,pi)) + plot(s5, (x,0,pi), color='red')
+    sage: plot(f, (0,pi)) + plot(s5, (x,0,pi), color='red')  # known bug: macos, random failures
```
mkoeppe: We should avoid the word "random" in these comments because it is a keyword in the doctest code with semantics that are orthogonal to what is denoted here.

@kwankyu suggested the word "sporadic" in https://github.com/sagemath/sage/pull/36372#issuecomment-1868124429; I would support that.

src/sage/calculus/transforms/dft.py
```diff
@@ -798,10 +798,10 @@ def idwt(self, other="haar", wavelet_k=2):
             sage: A = [RR(1) for i in J]
             sage: s = IndexedSequence(A,J)
             sage: t = s.dwt()
-            sage: t            # random arch dependent output
+            sage: t            # known bug, arch dependent output
```
mkoeppe: I don't think this is a "bug"

src/sage/structure/coerce_actions.pyx
```diff
@@ -781,7 +781,7 @@ cdef class IntegerMulAction(IntegerAction):
         Check that large multiplications can be interrupted::
 
             sage: # needs sage.schemes
-            sage: alarm(0.001); 2^(10^7) * P
+            sage: alarm(0.001); 2^(10^7) * P # known bug: macos, random failures, see #35973
```
mkoeppe: As I explained in https://github.com/sagemath/sage/pull/36372#issuecomment-1868342056 (see also https://github.com/sagemath/sage/pull/36372#issuecomment-1868386854, https://github.com/sagemath/sage/pull/36372#issuecomment-1742417444), explanations should go into parentheses because this is what our parsing/unparsing code knows to handle.

mkoeppe:
I have edited the title to use the terminology that we use in the doctest framework.

mkoeppe:
I would suggest to simplify and generalize the code as follows:
- add `linux` and `macos` to `auto_optional_tags` in https://github.com/sagemath/sage/blob/develop/src/sage/doctest/control.py#L57, 
- generalize the code to handle `known bug: TAGS...` with the semantics that if ALL of the given tags are present, the example is treated as a known bug.

src/sage/doctest/parsing.py
```diff
@@ -1210,11 +1242,27 @@ def check_and_clear_tag_counts():
                             continue
 
                     if self.optional_tags is not True:
-                        extra = {tag
-                                 for tag in optional_tags
-                                 if (tag not in self.optional_tags
-                                     and tag not in available_software)}
-                        if extra:
+                        extra = {
+                            tag
```
mkoeppe: this reformatting seems excessive

mkoeppe:
Overall I'm still skeptical that introducing platform-specific "known bug" semantics is a good idea, but I don't object to creating this mechanism. As I wrote in https://github.com/sagemath/sage/pull/36372#issuecomment-1824956436: "Introducing platform-specific skip flags in the source code is a departure from our long-standing practice. Our doctests are enabled by the presence of features, not disabled by an excuse. Where necessary, our downstream distribution developers have been externally maintaining their patches for tests that cannot pass on their platforms; and have been helping rewrite doctests so that they are suitable both within the Sage distribution and in downstream deployments."

kwankyu:
Hopefully this PR should produce something that makes everyone happy. One principle for this is to build based on what is already in sage and change things little by little as everyone agrees.

Though this is a PR, we may start with discussing things from ground up. We want to renew existing tag `# known bug` to cover doctests that are known to fail sometimes or on some platforms but not expected to be fixed soon. By the way, if a doctest always fail but for some reason we want to keep it, then there's `# not tested` tag for it.

Currently a doctest annotated with `# known bug` is skipped. As I suggested before and Matthias said ("it is a design flaw"), we can change its semantics. Thus the doctester runs the doctest but does not exit with error code when it fails, since it is expected. With this semantics, it is not an "optional tag" (which skips the doctest).

We want the tag `# known bug` to be conditional depending on platforms. If a doctest may fail (always or sometimes) only on a certain platform (say on conda, on macos conda, or on macos, etc.). So we check the condition. Modularization efforts introduced `# needs` tag and Matthias worked a lot with cleaning doctests annotations. Thus, I think, it is already established to use parentheses to complement or to give a reason for, a tag.  Doctest annotations can be complex because several tags can be given to a doctest. Parentheses is easier to parse. So I suggest the syntax `# known bug (randomly on macos conda)` for example.

What should happen if a doctest is annotated with `# known bug (randomly on macos conda)`? Since the doctester sees "macos conda" in the parenthesized content, it checks if we are on such a platform. (We may implement "platform features" like distribution packages features. Then "macos conda" may be the conjunction of two features "macos" and "conda") If the condition check fails, that is we are not on "macos conda", the doctester runs the doctest as like any other doctest. The interesting part is when we are on "macos conda". Then the doctester runs the doctest. If the doctest fails (as expected), it does report the failure but does not exit with error code. If the doctest does not fail (because it randomly fails), the doctester reports the success but does not exit. 

For a doctest with `# known bug` without parenthesized content, the doctester runs the doctest and reports the result but does not exit with error code. 

The renewed `# known bug` can be used  for conda-related doctest failures instead of @tobiasdiez `# skip_conda` (platform-specific tag) and @mkoeppe baseline-stats mechanism (file-scoped).

mkoeppe:
Note that `# known bug - macos (sporadic failures)` matches the format `# optional - bliss (because it needs it)`.

The explanations in parentheses should never be parsed - they are for humans to read.

kwankyu:
That is okay with me too.

kwankyu:
> ... Where necessary, our downstream distribution developers have been externally maintaining their patches for tests that cannot pass on their platforms

What does a patch do with the tests? Does it remove them?

mkoeppe:
> > ... Where necessary, our downstream distribution developers have been externally maintaining their patches for tests that cannot pass on their platforms
> 
> What does a patch do with the tests? Does it remove them?

Maybe @kiwifb can comment

kwankyu:
> ... Introducing platform-specific skip flags in the source code is a departure from our long-standing practice. Our doctests are enabled by the presence of features, not disabled by an excuse. 

As you know, we rather intend to run (previously skipped) doctests with "# known bug" empowered with platform check. Adding platform check means that we understand the nature of the bug more. So it is not so bad as "platform-specific tag".

kiwifb:
I have had a policy of mostly try to deal with broken doctest through issue ticket or enforcement of dependencies when appropriate. I sometimes adopt patches for a dependency update early on but they rarely last more than one release these days. But really I only patch or remove stuff if I think it is completely broken otherwise.

I do remove some doctest for stuff that I removed from sage. Stuff related to package management.
https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sage/files/sage-10.3-neutering.patch

Recently there has been some instance where I would have liked a tag to cover for stuff that is not always shipped when you use properly split sage. The tags we have may be sufficient provided we use them properly. I am thinking about test about `tdlib` source presence in particular in `sage/misc/package_dir.py` which is passing now because `sage/graphs/graph_decompositions/tdlib.pyx` has somehow made it back in sagemath-standard sdist instead of being only in `sagemath-tdlib`. I do remember 2 other instances of similar doctest failing recently.

mkoeppe:
This Issue has links to other downstream packaging:
- https://github.com/sagemath/sage/issues/31016

kwankyu:
> Please adjust the scope of this PR. There is no need to back out `.github/workflows/ci-conda-known-test-failures.json` for the goals of this PR, so it should not be done here.

In this PR, we may focus on implementing the "# known bug" mechanism. Adding it to doctests may be done elsewhere with the "backout out" if necessary.

src/sage/calculus/transforms/dft.py
tobiasdiez: See PR description.

src/sage/structure/coerce_actions.pyx
tobiasdiez: Done

src/sage/calculus/transforms/dft.py
mkoeppe: Yes, this is not a case that the PR description covers.

src/doc/en/constructions/calculus.rst
tobiasdiez: Done

tobiasdiez:
> Currently a doctest annotated with `# known bug` is skipped. As I suggested before and Matthias said ("it is a design flaw"), we can change its semantics. Thus the doctester runs the doctest but does not exit with error code when it fails, since it is expected. With this semantics, it is not an "optional tag" (which skips the doctest).

In this PR, I didn't change the semantics of "known bug". I also don't see any real advantage of running the tests and then just hiding its output (people are even ignoring red ci checks, I doubt they will study the log for 'known bug' annotations). But changing the semantics would be fine with me. That can be done in a follow-up PR.

tobiasdiez:
> The renewed `# known bug` can be used for conda-related doctest failures instead of @tobiasdiez `# skip_conda` (platform-specific tag) and @mkoeppe baseline-stats mechanism (file-scoped).

Note that there is only one bug left that apparently is specific to conda. Hence, I didn't included any conda-specific exclusion mechanism here.

src/sage/rings/function_field/drinfeld_modules/morphism.py
```diff
@@ -540,12 +540,12 @@ def __invert__(self):
             sage: K.<z> = Fq.extension(3)
             sage: coeffs = [z] + [K.random_element() for _ in range(10)]
             sage: phi = DrinfeldModule(A, coeffs)
-            sage: f = phi.hom(K.random_element())
-            sage: g = ~f
+            sage: f = phi.hom(K.random_element()) # known bug: macos (sporadic ValueError: the input does not define an isogeny)
+            sage: g = ~f # known bug: macos (sporadic ValueError: the input does not define an isogeny)eny
```
mkoeppe:
```suggestion
            sage: f = phi.hom(K.random_element())  # known bug: macos (sporadic ValueError: the input does not define an isogeny)
            sage: g = ~f  # known bug: macos (sporadic ValueError: the input does not define an isogeny)
```

per codestyle, 2 spaces before the `#` (please fix throughout)

kwankyu:
About syntax, as Matthias said, better to use hyphen  `-` instead of colon `:`, for uniformity across tags.

kwankyu:
> > The renewed `# known bug` can be used for conda-related doctest failures instead of @tobiasdiez `# skip_conda` (platform-specific tag) and @mkoeppe baseline-stats mechanism (file-scoped).
> 
> Note that there is only one bug left that apparently is specific to conda. Hence, I didn't included any conda-specific exclusion mechanism here.

OK. Good to hear that.

kwankyu:
By the way, does colon `:` work in `# optional - bliss` instead of hyphen `-`, currently?

kwankyu:
Just deleting `.github/workflows/ci-conda-known-test-failures.json` is not the proper way to remove the baseline-stats-based mechanism for known failures, which is already set in place in github workflows. 

The proper way is to prepare a PR just for that as a sequel to this PR. Please don't rush.

tobiasdiez:
> By the way, does colon `:` work in `# optional - bliss` instead of hyphen `-`, currently?

Yes, colon is also supported - it is actually the first example: https://github.com/sagemath/sage/blob/e249befd47ad8610f74372533a2943ff7b2dde94/src/sage/doctest/parsing.py#L137-L139. I didn't change anything in that respect.

I think its better to be consistent with other standard tools than with sage's own custom conventions. Also `known bug: linux (#23)` reads nicer than `known bug -- linux (#23)` in my opinion.

mkoeppe:
> About syntax, as Matthias said, better to use hyphen `-` instead of colon `:`, for uniformity across tags.

Actually I don't care too much about this. 

Note that we have standardized on `# needs FEATURES` (no interpunction) but `# optional - FEATURES` (single dash); so having the third form `known bug: FEATURES` is not a big deal. But we should keep it consistent within these tags.

Also we need to make sure that `sage --fixdoctests --no-test` normalizes these tags correctly (haven't checked)

kwankyu:
> > By the way, does colon `:` work in `# optional - bliss` instead of hyphen `-`, currently?
> 
> Yes, colon is also supported - it is actually the first example:
> 
> https://github.com/sagemath/sage/blob/e249befd47ad8610f74372533a2943ff7b2dde94/src/sage/doctest/parsing.py#L137-L139
> 
> . I didn't change anything in that respect.

OK.

> I think its better to be consistent with other standard tools than with sage's own custom conventions. 

I don't think so. Consistency across sage library is important, if there is already consistency.

If we want to follow other standard tools in python, we should do it across sage library. If this is a big change, then we should discuss on sage-devel.

> Also `known bug: linux (#23)` reads nicer than `known bug -- linux (#23)` in my opinion.

In my opinion too.

mkoeppe:
@tobiasdiez As I wrote in https://github.com/sagemath/sage/pull/36960#issuecomment-1869093146: adjust the scope of this PR. There is no need to back out `.github/workflows/ci-conda-known-test-failures.json` for the goals of this PR, so it should not be done here.
And certainly it will not be forced via the "blocker" label.

tobiasdiez:
> @tobiasdiez As I wrote in [#36960 (comment)](https://github.com/sagemath/sage/pull/36960#issuecomment-1869093146): adjust the scope of this PR. There is no need to back out `.github/workflows/ci-conda-known-test-failures.json` for the goals of this PR, so it should not be done here. 

Yes there is. With the changes in this PR, all tests executed in the files that are marked as "known test failures" are actually passing (at least to the best of my knowledge). This is verified by the fact that all checks are green. Thus, these exclusions are no longer necessary.

> And certainly it will not be forced via the "blocker" label.

The blocker label is needed to hide an annoying randomly failing test from the "Build & Test" workflow, as described in the PR description. I would kindly ask you to respect that and not remove the label again. Thanks!

mkoeppe:
> > @tobiasdiez As I wrote in [#36960 (comment)](https://github.com/sagemath/sage/pull/36960#issuecomment-1869093146): adjust the scope of this PR. There is no need to back out `.github/workflows/ci-conda-known-test-failures.json` for the goals of this PR, so it should not be done here.
> 
> Yes there is. With the changes in this PR, all tests executed in the files that are marked as "known test failures" are actually passing (at least to the best of my knowledge). This is verified by the fact that all checks are green. Thus, these exclusions are no longer necessary.

You are describing it correctly. They may no longer be needed, perhaps to the best of your knowledge.
But to back it out is not necessary for the goals of the PR. So it should not be done, and certainly not by blocker.

> > And certainly it will not be forced via the "blocker" label.
> 
> The blocker label is needed to hide an annoying randomly failing test from the "Build & Test" workflow, as described in the PR description. 

Can't have it both: (1) apply as a blocker (2) wider scope than necessary.

Tobias, when I introduced the "CI fixes via blocker label", I explained that restraint is necessary when using them. This is also documented in https://github.com/sagemath/sage/wiki/Sage-10.2-Release-Tour#open-blocker-prs-are-applied-automatically-in-ci-workflows

> I would kindly ask you to respect that and not remove the label again.

When I introduced the "CI fixes via blocker label", you were very concerned that it could be abused. I held that common sense and requiring the Triage privileges would be sufficient. Turns out you're the only developer who's abusing it. It's disruptive; stop it.

tobiasdiez:
> > > @tobiasdiez As I wrote in [#36960 (comment)](https://github.com/sagemath/sage/pull/36960#issuecomment-1869093146): adjust the scope of this PR. There is no need to back out `.github/workflows/ci-conda-known-test-failures.json` for the goals of this PR, so it should not be done here.
> > 
> > 
> > Yes there is. With the changes in this PR, all tests executed in the files that are marked as "known test failures" are actually passing (at least to the best of my knowledge). This is verified by the fact that all checks are green. Thus, these exclusions are no longer necessary.
> 
> You are describing it correctly. 

Awesome, then we are finally on the same page. Can I consider this as a positive review?

> But to back it out is not necessary for the goals of the PR. 

The goal of this PR is to provide an alternative for both https://github.com/sagemath/sage/pull/36678 and https://github.com/sagemath/sage/pull/36372.
So until you give concrete reasons why this change should not be done in this PR, I don't see why I should remove it.

> > The blocker label is needed to hide an annoying randomly failing test from the "Build & Test" workflow, as described in the PR description.
> 
> Can't have it both: (1) apply as a blocker (2) wider scope than necessary.

Let me know if you encounter any issues with this PR being blocker, and I'll happily remove it for you.

> When I introduced the "CI fixes via blocker label", you were very concerned that it could be abused. I held that common sense and requiring the Triage privileges would be sufficient. Turns out you're the only developer who's abusing it. It's disruptive; stop it.

I find it also disruptive that you remove the label constantly from my PRs. Moreover, no one else reported any issues so far.

mkoeppe:
> > You are describing it correctly.
> 
> Awesome, then we are finally on the same page. Can I consider this as a positive review?

You are deliberately misquoting by omission.

mkoeppe:
Any reaction to https://github.com/sagemath/sage/pull/36960#issuecomment-1869105228?

tobiasdiez:
> Any reaction to [#36960 (comment)](https://github.com/sagemath/sage/pull/36960#issuecomment-1869105228)?

I don't see any need at the moment to allow multiple tags. Can be done in the future if such need arises.

mkoeppe:
Happy new year, all.

As a maintainer action, I'll open a PR that cherry-picks from here according to the scope suggested in https://github.com/sagemath/sage/pull/36960#issuecomment-1869158986

tobiasdiez:
Test pass and all known errors are excluded, so this can safely be set as blocker. I'll monitor the runs, but in case I missed something please let me know and I will add a further known bug annotation.

tobiasdiez:
@roed314 I indeed made a small mistake while rebasing. Fixed now, so could you please read the blocker label. Thanks!

roed314:
Adding blocker back at @tobiasdiez's request in order to prevent failures such as https://groups.google.com/g/sage-devel/c/mLtXcaypU3Q.

mkoeppe:
@roed314 This blocks the blocker #36937, as explained in https://github.com/sagemath/sage/pull/36960#issuecomment-1869093146, https://github.com/sagemath/sage/pull/36960#issuecomment-1869984596

roed314:
@tobiasdiez claims that this PR will help resolve failures such as that in https://groups.google.com/g/sage-devel/c/mLtXcaypU3Q.  Do you disagree?  Will #36937 help with that?

I understand from the two comments that you're referencing that you don't agree with backing out `.github/workflows/ci-conda-known-test-failures.json`; can you explain the concrete harm that having both this and #36937 as blockers will be?  For example, auto-merging blockers is a fairly new feature, and I don't know what happens with merge conflicts.

mkoeppe:
1. Yes, marking doctests as `# known bug` as done here will hide failures in CI.
2. Deleting the file `.github/workflows/ci-conda-known-test-failures.json` is not needed for that.
3. Per https://github.com/sagemath/sage/wiki/Sage-10.2-Release-Tour#open-blocker-prs-are-applied-automatically-in-ci-workflows, "caution and restraint are necessary when using the label. Only narrowly focused fixes are suitable to avoid surprises and merge conflicts".
4. https://github.com/sagemath/sage/pull/36937 suppresses failures by updating the `.github/workflows/ci-conda-known-test-failures.json` file. 
5. The gratuitous deletion of `.github/workflows/ci-conda-known-test-failures.json` creates the conflict that stops #36937 from being applied. This can be seen in https://github.com/sagemath/sage/actions/runs/7672419399/job/20912863829?pr=37172#step:3:12

roed314:
@tobiasdiez I'm going to remove the blocker tag again.  I understand that you and @mkoeppe disagree about this ticket; hopefully Volker will be able to make a decision about how to proceed soon.

tobiasdiez:
I went through the conda logs of the past month with the following result:
- During the two weeks were this PR was marked as blocker there were exactly two failures due to missing `known bug` annotations which would have been covered by the `known-failures.json` mechanism.
- There were a couple of more failures which were not observed before (and hence are also not in `known-failures.json`)

Based on this analysis, I've now opened a couple of issues that track the new failures I've observed (if there were at least two instances of that failure on two different PRs), and added these via `known bug` annotations. Moreover, as a compromise, I've now readded the `known-failures.json` file and only removed the entries were I'm pretty sure that they are a) outdated or b) covered by the new `known bug` annotations.

Since this PR now includes quite a few annotations of known bugs that are not conda specific and are also not in `known-failures.json`, and since the above mentioned compromise is hopefully good for everyone, I'll remove the disputed label now and set it again as blocker. Hopefully we can get this is now.

roed314:
@tobiasdiez Thanks for looking through the logs, and I'm glad that this PR helped find a few bugs (perhaps it would be good to link to the issue numbers).  But please don't unilaterally change tickets to blockers, especially if there is disagreement about the approach you've taken.  Doing so further may lead to your removal from Triage again.

roed314:
In addition, you can't unilaterally decide that a ticket is no longer disputed.  Removing that tag is not appropriate.

tobiasdiez:
> @tobiasdiez Thanks for looking through the logs, and I'm glad that this PR helped find a few bugs (perhaps it would be good to link to the issue numbers). 

I "discovered" these issues by going through the logs and not via this PR (see the last 10 or 15 of https://github.com/sagemath/sage/issues/created_by/tobiasdiez). The point of this PR, and why it should be a "blocker", is that these issues show up randomly as test failures for both the Build & Test and the conda workflow (as well as various portability runs) and thus impacting everyone.  

So far nobody mentioned any concrete negative consequences of this PR being a blocker (except of course of my rebase mistake) - on the other hand, there are clear benefits. 

> especially if there is disagreement about the approach you've taken. 

I've reacted to the reviewer's concerns, so there shouldn't be any disagreement anymore.

roed314:
> I've reacted to the reviewer's concerns, so there shouldn't be any disagreement anymore.

Even if that's the case, it's better to give the reviewer a chance to weigh in themselves.

saraedum:
The description of this PR seems to be outdated. At least I cannot find any functional change to how doctests are parsed in the changeset.

Generally, the changes here look fine to me though.

So, maybe you could clarify what this PR is supposed to do exactly and then I am happy to cast a vote here. Or maybe, this PR has diverged so much from the original plan that it's worth to create a new PR and close this one.

mkoeppe:
@roed314 @saraedum @jhpalmieri I'm removing the "disputed" label, as the PR no longer sabotages the known-test-failures.json mechanism

mkoeppe:
Setting to "needs work" -- indeed the ticket description is outdated, as the implementation of the platform-dependent doctest tag mechanism was split out from here and merged long ago. 
The branch also needs updating to remove conflicts.
And then further review of the actual doctest changes is neded.

mkoeppe:
My general concern, as discussed previously in #36372, is that marking doctests `known bug` hides the bug too thoroughly: They no longer get tested at all. In particular for random failures, it eliminates our source of data on how frequently this bug occurs.

mkoeppe:
So before deploying more `# known bug` tags as done here, I would think we should do what I suggested in https://github.com/sagemath/sage/pull/36372#issuecomment-1867964422: "change the semantics of it to running the test but handling the failure in the same way as the baseline failures. For tests that really should not be run, this can be combined with # not tested (some doctests already do this)."

PR #36965, opened 2023-12-26
Update system package info for slackware

mkoeppe:
@cduston44

### :memo: Checklist

### DIFF
--- a/build/pkgs/_prereq/distros/slackware.txt
+++ b/build/pkgs/_prereq/distros/slackware.txt
@@ -13,10 +13,8 @@ make
 guile
 gc
 libffi
-gcc-[0-9]      # So that slackpkg pattern matching does not pull in all gcc-* packages
-gcc-11           # on slackware-current
-gcc-g++
-gcc-g++-11       # on slackware-current
+gcc-13
+gcc-g++-13
 # dependencies of gcc
 libmpc
 glibc
@@ -24,8 +22,7 @@ kernel-headers
 perl
 m4
 bc
-python-2.7       # on slackware-14.2
-python3          # on slackware-current
+python3
 flex
 # for https upstream_url downloads
 ca-certificates


kwankyu:
If this was tested at least once (since I can't test) and @cduston44 does not respond, then I will set this positive review.

kwankyu:
Could this 

https://github.com/aclemons/slackware-dockerfiles

be used to test this?

By the way, do we have many sage users on slackware linux distro?

mkoeppe:
It can be tested like our other Linux platforms, using `tox -e docker-slackware-current-standard` etc.; see tox.ini

(Note that the goal of this PR is not to make everything work. It just makes it testable.)

kwankyu:
> ... since I can't test

Then what did you mean?

kwankyu:
> If this was tested at least once (since I can't test) and @cduston44 does not respond, then I will set this positive review.

I forgot that I made this comment :-)

OK. With this PR, slackware become testable in the sense that the test by tox proceeds further than before this PR.

For me, the test fails in building sage even with this PR.

kwankyu (approved):

kwankyu:
> By the way, do we have many sage users on slackware linux distro? Or is @cduston44 the only one as far as known?

Perhaps non answerable...  I asked  because I guess that it is an unpopular linux distro.

mkoeppe:
> do we have many sage users on slackware linux distro?

I don't know any. When I did the initial porting to slackware, I did it only as an illustration of how to do the porting work for Sage using the Docker technology.

mkoeppe:
Thanks for the review!

cduston44:
As far as I know, I am the only Slackware user of Sage, but I don't really use docker so I don't think I can contribute to this discussion. Sorry!

Sent with [Proton Mail](https://proton.me/) secure email.

PR #36970, opened 2023-12-26
CI: Upload test stats as artifacts, improve output of "List Docker images"

mkoeppe:
- `tox -e docker-....` now copies logs and stats from the image to the tox environment directory
- CI Linux and CI Linux Incremental (via docker.yml) prepare the logs artifact from there (they used to be in charge of copying the logs from the image)
- We switch from the legacy Docker builder to the current one ("buildkit"), avoiding deprecation warnings (e.g. https://github.com/sagemath/sage/actions/runs/7342841283/job/19992633064?pr=36977#step:11:29)
- We also improve the output of the step "List Docker images": 
   - It no longer advertises Docker images if they could not be pushed to the registry - see https://github.com/sagemath/sage/actions/runs/7335489109/job/19973517180?pr=36970#step:15:60
   - When it is able to push, it now shows instructions for developers what to do with them - see  https://github.com/mkoeppe/sage/actions/runs/7345058426/job/19997709977#step:15:61 - and also creates an Annotation - see https://github.com/mkoeppe/sage/actions/runs/7345058426
- Just to trigger the CI Linux Incremental workflow here on the PR, we include the upgrade of the `info` package.

In a follow-up:
- We could store the stats in a persistent database

### :memo: Checklist


build/bin/write-dockerfile.sh
```diff
@@ -214,6 +214,18 @@ EOF
...
+case ${DOCKER_BUILDKIT-0} in
+    1)
+        # With buildkit we cannot retrieve failed builds.
+        # So we do not allow the main step of a build stage to fail.
+        # Instead we record the exit code in the file STATUS.
+        THEN_SAVE_STATUS='; echo $? > STATUS'
+        # ... and at the beginning of the next build stage,
+        # we check the status and exit with an error status.
+        CHECK_STATUS_THEN='STATUS=$(cat STATUS 2>/dev/null); case "$STATUS" in ""|0) ;; *) exit $STATUS;; esac; '
+esac
```
kwankyu, line 227: Out of curiosity, how does this trick avoid the problem of "With buildkit we cannot retrieve failed builds."?

How is "failed build" different from "exit" with nonzero status?

kwankyu:
Otherwise lgtm as far as I can understand. 

Is this change fully tested (in your fork)? Do you need to test again after 2 months?

build/bin/write-dockerfile.sh
mkoeppe: > Out of curiosity, how does this trick avoid the problem of "With buildkit we cannot retrieve failed builds."?

By having the next build stage fail at the beginning, instead of failing the current one.

mkoeppe:
> Is this change fully tested (in your fork)? 

Yes

> Do you need to test again after 2 months?

Well, I've started another run now: https://github.com/mkoeppe/sage/actions/runs/8033793985

mkoeppe:
Yes, looking good. e.g. https://github.com/mkoeppe/sage/actions/runs/8033793985/job/21944652261#step:15:61

build/bin/write-dockerfile.sh
kwankyu: so that the previous stage remains in the container and the current one discarded...

OK.

kwankyu (approved):
LGTM.

kwankyu:
OK. Let's see it in the battle field.

mkoeppe:
Thank you!

PR #36972, opened 2023-12-26
add a specialized parent object for elliptic-curve morphisms

yyyyx4:
Currently, the set of morphisms between elliptic curves is a generic `SchemeHomset`. However, the morphisms between abelian varieties such as elliptic curves have a lot more structure, which should be reflected in Sage by a specialized parent object.
This patch adds such a parent. The current implementation does not really do much except to enable convenience syntax for scalar multiplications. More advanced functionality (structure computation, finding elements, ...) can be added later.

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1113,7 +1113,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             L = L, F_to_K.post_compose(K_to_L)
         return L
 
-    def _Hom_(*args, **kwds):
+    def _Hom_(self, other, category=None):
         r"""
         Hook to make :class:`~sage.categories.homset.Hom`
         set the correct parent
@@ -1128,8 +1128,11 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             sage: type(E._Hom_(E))
             <class 'sage.schemes.elliptic_curves.homset.EllipticCurveHomset_with_category'>
         """
-        from . import homset
-        return homset.EllipticCurveHomset(*args, **kwds)
+        if isinstance(other, ell_generic.EllipticCurve_generic) and self.base_ring() == other.base_ring():
+            from . import homset
+            return homset.EllipticCurveHomset(self, other, category=category)
+        from sage.schemes.generic.homset import SchemeHomset_generic
+        return SchemeHomset_generic(self, other, category=category)
 
     def isogeny(self, kernel, codomain=None, degree=None, model=None, check=True, algorithm=None):
         r"""
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1113,6 +1113,24 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             L = L, F_to_K.post_compose(K_to_L)
         return L
 
+    def _Hom_(*args, **kwds):
+        r"""
+        Hook to make :class:`~sage.categories.homset.Hom`
+        set the correct parent
+        :class:`~sage.schemes.elliptic_curves.homset.EllipticCurveHomset`
+        for
+        :class:`~sage.schemes.elliptic_curves.hom.EllipticCurveHom`
+        objects.
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve(GF(19), [1,0])
+            sage: type(E._Hom_(E))
+            <class 'sage.schemes.elliptic_curves.homset.EllipticCurveHomset_with_category'>
+        """
+        from . import homset
+        return homset.EllipticCurveHomset(*args, **kwds)
+
     def isogeny(self, kernel, codomain=None, degree=None, model=None, check=True, algorithm=None):
         r"""
         Return an elliptic-curve isogeny from this elliptic curve.
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -2069,6 +2069,45 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
         GL = G.relabel(labels, inplace=False)
         return GL
 
+    def endomorphism_ring_is_commutative(self):
+        r"""
+        Check whether the endomorphism ring of this elliptic curve
+        *over its base field* is commutative.
+
+        ALGORITHM: The endomorphism ring is always commutative in
+        characteristic zero. Over finite fields, it is commutative
+        if and only if the Frobenius endomorphism is not in `\ZZ`.
+        All elliptic curves with non-commutative endomorphism ring
+        are supersingular. (The converse holds over the algebraic
+        closure, but here we consider endomorphisms *over the field
+        of definition*.)
+
+        EXAMPLES::
+
+            sage: EllipticCurve(QQ, [1,1]).endomorphism_ring_is_commutative()
+            True
+            sage: EllipticCurve(QQ, [1,0]).endomorphism_ring_is_commutative()
+            True
+            sage: EllipticCurve(GF(19), [1,1]).endomorphism_ring_is_commutative()
+            True
+            sage: EllipticCurve(GF(19^2), [1,1]).endomorphism_ring_is_commutative()
+            True
+            sage: EllipticCurve(GF(19), [1,0]).endomorphism_ring_is_commutative()
+            True
+            sage: EllipticCurve(GF(19^2), [1,0]).endomorphism_ring_is_commutative()
+            False
+            sage: EllipticCurve(GF(19^3), [1,0]).endomorphism_ring_is_commutative()
+            True
+        """
+        k = self.base()
+        if k.characteristic() == 0 or self.is_ordinary():
+            return True
+
+        if not k.is_finite():
+            raise NotImplementedError
+
+        return self.frobenius() not in ZZ
+
 
 def compute_model(E, name):
     r"""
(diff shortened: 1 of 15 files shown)


GiacomoPope (approved):
Seems like a sensible parent class to create to allow for more complex and interesting code to be included in the future. I cannot think of a reason *not* to add this parent class

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1113,6 +1113,24 @@ def division_field(self, n, names='t', map=False, **kwds):
...
+        objects.
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve(GF(19), [1,0])
+            sage: type(E._Hom_(E))
+            <class 'sage.schemes.elliptic_curves.homset.EllipticCurveHomset_with_category'>
+        """
+        from . import homset
+        return homset.EllipticCurveHomset(*args, **kwds)
```
tscrim: You should have some more restrictions on the input as the `_Hom_` method is called very generically IIRC. Try, e.g., `Hom(E, R['x','y'])`. I would be more specific about the inputs allowed too.

src/sage/schemes/elliptic_curves/homset.py
```diff
@@ -0,0 +1,151 @@
...
+        ::
+
+            sage: E1 = EllipticCurve(j=42)
+            sage: E2 = EllipticCurve(j=43)
+            sage: Hom(E1, E2)
+            Group of elliptic-curve morphisms
+              From: Elliptic Curve defined by y^2 = x^3 + 5901*x + 1105454 over Rational Field
+              To:   Elliptic Curve defined by y^2 + x*y = x^3 + x^2 + 1510*x - 140675 over Rational Field
+        """
+        super().__init__(*args, **kwds)
```
tscrim: It would be good to manipulate the category to know that the homset is also in `Groups()` if that is indeed the case. Although I have trouble believe that, in which case it should be the "Set of...".

tscrim (commented):
I would say there is a more natural way to set up the map from $\mathbb{Z}$ by using `_coerce_map_from_`, which could return `True` and then let `_element_constructor_` handle constructing the map, which seems like the more natural case here. The other option would be to have `_coerce_map_from_` simply return the map. The class for the morphism should be on its own with doctests too (if it survives this change).

yyyyx4:
Thanks for your comments, I think I've addressed them all with the new commits.

src/sage/schemes/elliptic_curves/homset.py
```diff
@@ -0,0 +1,234 @@
...
+
+            assert E1 == E2
+
+            # The endomorphism ring of an elliptic curve over a finite field
+            # is non-commutative if and only if the curve is supersingular and
+            # the Frobenius endomorphism equals a scalar.
+            if (base.characteristic() == 0 or E1.is_ordinary() or
+                    (base.is_finite() and E1.frobenius() not in ZZ)):
+                from sage.categories.commutative_rings import CommutativeRings
+                self._refine_category_(CommutativeRings())
```
tscrim: Rather than refine the category afterwards, it would be better to modify the `category` (for example, by `category & Rings()`) variable before passing it off to `super().__init__()`.

src/sage/schemes/elliptic_curves/homset.py
```diff
@@ -0,0 +1,234 @@
...
+            sage: E1 = EllipticCurve(j=42)
+            sage: E2 = EllipticCurve(j=43)
+            sage: Hom(E1, E1)(5)
+            Scalar-multiplication endomorphism [5] of Elliptic Curve defined by y^2 = x^3 + 5901*x + 1105454 over Rational Field
+            sage: Hom(E1, E2)(5)
+            Traceback (most recent call last):
+            ...
+            ValueError: domain and codomain must be equal
+        """
+        if self.codomain() != self.domain():
```
tscrim:
```suggestion
        if not self.is_endomorphism_set():
```
For consistency, plus some protections for accidental deallocations.

src/sage/schemes/elliptic_curves/homset.py
```diff
@@ -0,0 +1,234 @@
...
+            sage: Hom(E1, E1)
+            Ring of elliptic-curve endomorphisms
+              From: Elliptic Curve defined by y^2 = x^3 + x + 1 over Rational Field
+              To:   Elliptic Curve defined by y^2 = x^3 + x + 1 over Rational Field
+            sage: Hom(E1, E2)
+            Additive group of elliptic-curve morphisms
+              From: Elliptic Curve defined by y^2 = x^3 + x + 1 over Rational Field
+              To:   Elliptic Curve defined by y^2 = x^3 + 2*x + 2 over Rational Field
+        """
+        if self.domain() == self.codomain():
```
tscrim:
```suggestion
        if self.is_endomorphism_set():
```
Similarly.

src/sage/schemes/elliptic_curves/homset.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(self, E1, E2, category=None):
+        r"""
+        Construct the homset for a given pair of elliptic curves
+        defined over the same base ring.
+
+        TESTS::
+
+            sage: E = EllipticCurve('37a1')                                             # needs sage.schemes
+            sage: Hom(E, E).__class__                                                   # needs sage.schemes
+            <class 'sage.schemes.elliptic_curves.homset.EllipticCurveHomset_with_category'>
```
tscrim:
```suggestion
            sage: H = Hom(E, E)
            sage: TestSuite(H).run()
```
I don't find setting the class a very useful test personally. I would rather see a test like this (although I think you need to have `skip="_test_category"` by the nature of how homsets work) and similarly for `Hom(E1, E2)`.

tscrim (changes requested):
Thank you for the changes. Sorry for a more incremental review (I am doing more parenting right now and having limited time because of that). There are a few smaller things, but I also have one bigger comment:

I am not sure if we want to instead have the endomorphisms as a separate subclass. I guess since the `identity()` method is implemented for general homsets, we would want to keep things together. Likewise, I am wondering if there is a use in overriding `identity()` to return a `EllipticCurveHom_scalar`.

After these comments are addressed, then it will be a positive review.

src/sage/schemes/elliptic_curves/homset.py
yyyyx4: I believe `category` here is the category of the *objects* for which this homset is defined, but `._refine_category_()` changes the category of the homset itself. I don't see an obvious way to pass a category for the homset itself to the parent constructor.

tscrim: You're right. It seems like the proper way to do this would be to implement the appropriate `Homsets` category to `Schemes_over_base` (or perhaps a more refined category is needed?). For an example, see the `Modules` category in `category/modules.py`. Refining categories can be somewhat brittle IIRC, although within an `__init__` is safer I believe.

yyyyx4: The correct category for this would probably be abelian varieties. I'd personally leave that for the future (when we hopefully have more examples than just elliptic curves!) unless there are real problems with the current approach. Do you insist on doing it the proper way?

tscrim: For this I feel a bit more insistent since there have been many subtle issues that have come up with using `_refine_category_`. To be fair, most of those involve refining something like `Zmod(5)` once a primality test is done, where elements could already exist. That being said, there are already multiple proper mechanisms for setting the correct categories for morphisms before the objects are created.

yyyyx4: Alright, I gave it a try. I'm not sure of anything I'm doing here, though, since it's my first time dealing with categories in Sage...

tscrim: From first glance at what is pushed, it looks good and tests seem to be passing.

tscrim:
So right now we get a really annoying but trivial (and dumb) doctest failures:
```
sage -t --long --random-seed=286735480429121101562228604801325644303 sage/doctest/forker.py
File "sage/doctest/forker.py", line 1560, in sage.doctest.forker.SageDocTestRunner.report_unexpected_exception
Failed example:
    sage0.eval("u")
Expected:
    '...EllipticCurve_field.__init__(self, K, ainvs)'
Got:
    '> /sage/src/sage/schemes/elliptic_curves/ell_field.py(51)__init__()\n-> super().__init__(R, data, category=category)'
File "sage/doctest/forker.py", line 1562, in sage.doctest.forker.SageDocTestRunner.report_unexpected_exception
Failed example:
    sage0.eval("p ainvs")
Expected:
    '(0, 0, 0, 0, 0)'
Got:
    "*** NameError: name 'ainvs' is not defined"
```
This comes from the fact that the construction is done differently, so the steps needed by `Pbd` are different. At least, I believe that to be the problem. Basically, this is a very brittle test, but we will have to fix it. `:/

Some additional comments on the category:

- `_endomorphism_ring_is_commutative` should be a method of the elliptic curve since it 
- Sorry it this is mathematically wrong; I am not an expert in this area. You could at least check if the characteristic is 0 for the endset's `extra_super_categories()` since that would be a quick check. Of course you can do the full check there, but it seems from your code comments that it would be (relatively) expensive to do on construction (without enough benefits). Of course, you could refine the category in `is_commutative()`, but I don't think we would get any benefits from doing that.

yyyyx4:
Fixed the doctest (we have to do `u` twice now) and turned `_endomorphism_ring_is_commutative()` into a method.

It's true that characteristic 0 implies commutative endomorphism rings for elliptic curves, but this is no longer true for higher-dimensional abelian varieties: For example, a product of an elliptic curve with itself has $\mathbb Z^{2\times 2}$ in its endomorphism ring. So we cannot set this as a sub*category*; it really depends on the object of which we take the `End()`, just like the `.endomorphism_ring_is_commutative()` method for elliptic curves over finite fields does. Anyway, I'd argue that this is a non‑issue for the moment: We have no obligation to register something with the finest possible category in which it lies, and I find it hard to imagine that anyone would run `End(X) in CommutativeRings()` rather than just `End(X).is_commutative()` as things stand.

tscrim (approved):
Thank you for fixing that and the explanations.

Yes, I agree that we do not need to give the most refined category. It could be done lazily by refining the category when `is_commutative()` is called.

Thanks for all the fixes. This LGTM now; positive review. We can always make further improvements later on.

tscrim:
PS - Sorry for the delay in getting the final review done.

yyyyx4:
No worries, thanks for the review!

PR #36951, opened 2023-12-23
`pkgs/sagemath-standard`: Move metadata from `setup.cfg` to `pyproject.toml`

mkoeppe:
Modernize the metadata. 

This is a trivial "chore" PR. It updates Python metadata to the latest format. No controversies about the current format are known about in the Python community. In a typical open source project, someone in a Maintainer role would open a PR and then immediately merge it, or when receiving such a PR from the outside, quickly review and merge it (examples: my PRs https://github.com/pytest-dev/pytest-mock/pull/410 (merged in within 1 day), https://github.com/pyodide/pyodide/pull/4472, https://github.com/pytest-dev/pytest-xdist/pull/1020,
https://github.com/sagemath/cypari2/pull/158, https://github.com/fplll/fpylll/pull/258, https://github.com/polymake/JuPyMake/pull/2, https://github.com/cvxpy/cvxpy/pull/2276, https://github.com/sagemath/cysignals/pull/177). 

Also fixes the `pyproject.toml` build requirements of **sagemath-standard**, broken since 10.2, hence "critical".

- Part of https://github.com/sagemath/sage/issues/33577

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on https://github.com/sagemath/sage/pull/36885 (merged here)
- Depends on #35986 (merged here to resolve merge conflict)
- Depends on #37546 (merged here to resolve merge conflict)

PR #36957, opened 2023-12-25
tox.ini: Update voidlinux, update installation guide

mkoeppe:
To test: Use `export DOCKER_TARGETS="with-system-packages configured with-targets-pre with-targets"`, then one of:
- `tox -e docker-voidlinux-standard-python3.11`
- `tox -e docker-voidlinux-musl-standard-python3.11`

Support for `tox -e docker-voidlinux-standard-sitepackages` will have to wait for:
- https://github.com/sagemath/sage/pull/36181 

This
- Fixes #33084

### :memo: Checklist

- [ ] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

Issue #35890, opened 2023-07-03
Complete our GitHub dependency graph (javascript help needed)


mkoeppe:
https://github.com/sagemath/sage/network/dependencies?page=1 knows about some our Python dependencies -- because it happens to find the `build/pkgs/*/requirements.txt` files of our "pip" packages.

This is, of course, highly incomplete. To add all of SageMath's dependencies, we should use the [GitHub Dependency submission API](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api). This will also make SageMath visible in our upstream projects' GitHub front pages (in the "Used by" section.)

This will need some basic Javascript coding, using https://github.com/github/dependency-submission-toolkit in a GH Actions workflow and reading in our package metadata from `build/pkgs`.

Bonus points for making this workflow reusable by our upstream projects, reusing the dependency information that we have for them in `build/pkgs`.

Depends on:
- #37500 (because the Dependency API uses PURLs)

ydv-ankit:
I've knowledge of js, how can i help?

mkoeppe:
Learn about the GH Dependency submission API (link above), the dependency-submission-toolkit (link above), then write a JS program that reads the package metadata at https://github.com/sagemath/sage/tree/develop/build/pkgs

ydv-ankit:
okay.

ydv-ankit:
Can you tell me which one is metadata file(s) in these directories
https://github.com/sagemath/sage/tree/develop/build/pkgs

mkoeppe:
The contents of these directories are documented in https://doc.sagemath.org/html/en/developer/packaging.html#directory-structure

What information will you need for each package?

ydv-ankit:
Nice, this can help.

PR #36982, opened 2023-12-31
Make pyproject.toml the source for build dependencies

tobiasdiez:
As noted in https://github.com/sagemath/sage/pull/36951#issuecomment-1872637925, the fact that the dependency declaration of sage-the-library depends on information contain in sage-the-distro's build folder is problematic for the separation of concerns. Sage-the-distro should rely on information of sage-the-library, not the other way around. Here we fix this for the build requirements by making `src/pyproject.toml` the single source of thruth. The corresponding `install-requires.txt` are demoted to mere indicator files (to be removed later). Moreover, this change allows us to make `pyproject.toml` a static file (no longer generated by configure), which should help with downstream packaging.

---

Happy new year!

### :memo: Checklist

### :hourglass: Dependencies

- https://github.com/sagemath/sage/pull/37796


mkoeppe:
`pyproject.toml` is not generated by `configure`. It's generated by `bootstrap`. The best way to think about `bootstrap` is as a script of the monorepo, not of the Sage distribution.

I expect `src/pyproject.toml` to go away with the modularization. And the modularized distributions still have to go through `bootstrap`, so I don't think this approach helps with anything.

Anyway, some change away from the `install-requires.txt` files should probably happen at some point, as this name has become outdated with the transition to the new names set by the `pyproject.toml` format. But this clearly needs more thought.

orlitzky:
Well, this is pretty much what I had in mind when I said "but we also don't need to redesign it right now." Distro-from-library has a few benefits over library-from-distro:

1. Having the dependencies in `pyproject.toml` is the standard thing and is what people expect
2. While the library (or modular packages) may still need `./bootstrap`, this reduces how much they need it. Some sub-packages could potentially become git submodules that can be checked out and built on their own without `./bootstrap`; it's one fewer thing tying us to the monorepo design.

We do however still need the install-requires.txt with the versions in them (the python package `./configure` check uses them). To keep everything working, we'd have to generate the `build/.../install-requires.txt` from the toml file(s). The `sage-get-system-packages` script is already parsing the toml file though so maybe that's not too hard.

mkoeppe:
> 2\. Some sub-packages could potentially become git submodules that can be checked out and built on their own without `./bootstrap`; it's one fewer thing tying us to the monorepo design.

My solution for this is #31662, packaging sage-bootstrap with the metadata as a pip-installable distribution

orlitzky:
> > 2. Some sub-packages could potentially become git submodules that can be checked out and built on their own without `./bootstrap`; it's one fewer thing tying us to the monorepo design.
> 
> My solution for this is #31662, packaging sage-bootstrap with the metadata as a pip-installable distribution

The example I have in mind (if _sage-foo_ is pure python) is,

```
$ git clone https://github.com/sagemath/sage-foo.git
$ cd sage-foo
$ emacs src/sage/example.py
$ pip install .
```

That's just how people expect to work on a python project. If cython is involved, there would also be a `meson setup/compile` in there somewhere, but the process is pretty familiar to people. You can declare `sage-bootstrap` to be a dependency of `sage-foo` but that still adds an extra step, and it's likely that the sage-bootstrap version on pypi will be outdated relative to your git checkout of the subproject.

Anyway, details aside, it's just nicer to _not_ have to solve this extra problem.

mkoeppe:
The syncing of versions has to happen one way or the other. I think we all know the various possible designs, which differ by the time that the syncing happens and the actor that does the syncing.

In the end, it will depend on the actual developer & user community of a "subpackage" what is the best way to maintain it. 

Right now we have the luxury of the monorepo, where all version info is central and syncing is done by developers, not users, by the bootstrap script. 

I'm all for making our system agnostic with respect to the syncing mechanism so that *when* it is useful for a community that takes care of a "subpackage" finds it convenient or more productive to develop in a separate repo, we are ready for it. 

But at the moment there is no evidence that we have a viable community for any subpackage that would need this.

mkoeppe:
One design consideration:
- #35890

mkoeppe:
The built-in stuff in GitHub (obviously) does not know about out install-requires.txt files. The table https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems shows for Python:

Package manager | Languages | Recommended formats | All supported formats
-- | -- | -- | --
pip | Python | requirements.txt, pipfile.lock | requirements.txt, pipfile, pipfile.lock, setup.py
Python Poetry | Python | poetry.lock | poetry.lock, pyproject.toml

orlitzky:
> The syncing of versions has to happen one way or the other. I think we all know the various possible designs, which differ by the time that the syncing happens and the actor that does the syncing.

If a subpackage can be made to not depend on sage-bootstrap, then we don't need to worry about the version of sage-bootstrap. All else being equal, that's an improvement.


> In the end, it will depend on the actual developer & user community of a "subpackage" what is the best way to maintain it.
> 
> Right now we have the luxury of the monorepo, where all version info is central and syncing is done by developers, not users, by running the `bootstrap` script; [Edit:] and the release manager by running the `sage-update-version` script.
> 
> I'm all for making our system agnostic with respect to the syncing mechanism so that _when_ it is useful for a community that takes care of a "subpackage" finds it convenient or more productive to develop in a separate repo, we are ready for it.
> 
> But at the moment there is no evidence that we have a viable community for any subpackage that would need this.

We'll never _need_ it, but it's the simple standard solution to an easy problem. The Github dependency graph is a great example of how such a solution can save us time. Putting the dependencies in the standard location lets Github just... find them, easily, in contrast with the amount of work you've outlined under `build/` to achieve the same thing.

mkoeppe:
> Putting the dependencies in the standard location

But there is no "standard location"; we have multiple packages in `pkgs/`, and it's the very point that they are synchronized. (As I wrote in https://github.com/sagemath/sage/pull/36982#issuecomment-1872776919, the monolithic src/pyproject.toml is going to go away.)

orlitzky:
> > Putting the dependencies in the standard location
> 
> But there is no "standard location"; we have multiple packages in `pkgs/`, and it's the very point that they are synchronized. (As I wrote in [#36982 (comment)](https://github.com/sagemath/sage/pull/36982#issuecomment-1872776919), the monolithic src/pyproject.toml is going to go away.)

Inside of a pyproject.toml file is the standard location. When the monolithic one goes away, we'll have multiple pyproject.toml files, one for each subproject, having its own list of dependencies and constraints. Github will still be able to find them. and we'll still be able to figure out a version that satisfies them all to put in package-version.txt. There's even an automated tool called a package manager than can compute a feasible set of versions for us :)

mkoeppe:
> When the monolithic one goes away, we'll have multiple pyproject.toml files, one for each subproject, having its own list of dependencies and constraints.

That's synchronizing ... by not synchronizing?

orlitzky:
> > When the monolithic one goes away, we'll have multiple pyproject.toml files, one for each subproject, having its own list of dependencies and constraints.
> 
> That's synchronizing ... by not synchronizing?

The best kind of synchronizing! (Why would we manually synchronize them?)

mkoeppe:
Hm? The synchronization of version constraints through bootstrap simplifies maintenance. I think there should be a compelling reason before we start giving different version constraints to different distribution packages in pkgs/.

orlitzky:
> Hm? The synchronization of version constraints through bootstrap simplifies maintenance. I think there should be a compelling reason before we start giving different version constraints to different distribution packages in pkgs/.

Well, presumably you've spent years working on the modularization so that users will be able to install only some parts of sage. Why would we require them to satisfy the constraints of the pieces that they _haven't_ installed?

mkoeppe:
For the convenience of developers. When users need more flexibility, that can be a compelling reason to deviate from it.

orlitzky:
For this developer at least, not having to run `./bootstrap` is pretty convenient. Not having to reverse engineer the correct dependencies for our distro packages would be nice as well.

dimpase:
We already constrain users and developers --- by not using rust-dependent packages.
(Arguing that it's not feasible to ship rust - which defies the logic of not shipping C++ compilers on macOS, as we just can't.)

We already hear requests from.developers (@tornaria for instance) for not putting upper limits on package versions.

It may well be that there are platforms on which sage the distro doesn't work, but sagelib does work.

tobiasdiez:
> We do however still need the install-requires.txt with the versions in them (the python package `./configure` check uses them). To keep everything working, we'd have to generate the `build/.../install-requires.txt` from the toml file(s). The `sage-get-system-packages` script is already parsing the toml file though so maybe that's not too hard.

Right, I forgot about that. Was indeed very easy to fix and the necessary install_requires files are now automatically generated. In the future, we may consider extracting the version from the pyproject.toml file directly also for the configure checks.

tobiasdiez:
> The syncing of versions has to happen one way or the other. I think we all know the various possible designs, which differ by the time that the syncing happens and the actor that does the syncing.

And none of this is changed in this PR.

tobiasdiez:
The failing test is not related to this PR https://github.com/sagemath/sage/issues/36992.

bootstrap
```diff
@@ -45,6 +45,16 @@ bootstrap () {
     for a in m4/sage_spkg_versions.m4 m4/sage_spkg_versions_toml.m4; do
         echo 'changequote(`>>>'"'"', `<<<'"')dnl" >> $a
     done
+
+    for pkgname in cypari cysignals cython gmpy2 jupyter_core memory_allocator numpy pkgconfig pplpy setuptools wheel; do
```
dimpase: what does this list of packages consist of?

src/pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+    # Some version of sage-conf is required.
+    # Note that PEP517/518 have no notion of optional sage_spkg dependencies:
+    # https://github.com/pypa/pip/issues/6144
+    'sage-conf ~= 10.3b4',
+    # 68.1.0 Promote pyproject.toml's [tool.setuptools] out of beta.
+    # 68.1.1 Fix editable install finder handling of nested packages
+    'setuptools >= 68.1.1',
+    # version constraint for macOS Big Sur support (see https://github.com/sagemath/sage/issues/31050)
+    'wheel >=0.36.2',
+    'sage-setup ~= 10.3b4',
```
dimpase: can this be generated from Sage version?

tornaria:
> For the convenience of developers. When users need more flexibility, that can be a compelling reason to deviate from it.

I'm starting to believe that when you say "modularity" you mean something different than what I understand by "modularity".

For the convenience of developers, IMO modularity means independent modules that can be built, tested, installed, updated, independently of each other. This includes having each its own dependencies, and trying to "synchronize up" or "automate" is against this.

I'm 100% in favour of static pyproject.toml files that allow me to build sage-the-lib with standard pep-517 builder, no bootstrap, no sage-setup if possible, no mandatory sage-conf, etc. Just find stuff on standard locations. Of course this only works if I have all the necessary bits already installed, but so what? The compiler will tell me if I'm missing some bits and I can correct that.

Modularity also means separation of concerns, e.g. that my distro of choice takes care of other software needed to build sagemath.

> It may well be that there are platforms on which sage the distro doesn't work, but sagelib does work.

Indeed, trying #36957 I learned that "sage the distro" doesn't work on void linux (I couldn't make `tox -e docker-voidlinux-standard` work after spending a whole afternoon trying -- most of the time waiting between attempts which goes to show how useless it is since building sagelib from scratch only takes a few minutes).

bootstrap
tobiasdiez: These packages are required at build time and as of this PR have an `install requires` file that is automatically generated. In the future, we can expand this mechanism and include all python packages (with their version defined in the `project.toml` file).

src/pyproject.toml
tobiasdiez: It has to be static. I'm not sure why we need a version constraint at all, and actually hope we can get ride of the setup and conf packages soon.

tornaria: Re: sage-setup version. It's actually a problem that it requires a prerelease version.

If sage-setup is meant to be a separate python package, then building sage-the-lib should not require a prerelease version of sage-setup.

This way sage-setup can actually be packaged in distros, and used to build sagelib. But I only want stable / released packages installed in my system, while still being able to build sagelib.

mkoeppe: > If sage-setup is meant to be a separate python package, then building sage-the-lib should not require a prerelease version of sage-setup.

A release version of sagelib never requires a prerelease version of sage-setup.

mkoeppe:
> I'm starting to believe that when you say "modularity" you mean something different than what I understand by "modularity".
> 
> For the convenience of developers, IMO modularity means independent modules that can be built, tested, installed, updated, independently of each other. This includes having each its own dependencies, and trying to "synchronize up" or "automate" is against this.

Who's going to maintain it? As I explained above in https://github.com/sagemath/sage/pull/36982#issuecomment-1873020654, we don't have separate developer communities for the independent modules. Right we have distributions in `pkgs/` than developers who touch any of these version constraints. And even more with the modularization in #35095 adding more distributions.

mkoeppe:
> trying #36957 I learned that "sage the distro" doesn't work on void linux (I couldn't make `tox -e docker-voidlinux-standard` work after spending a whole afternoon trying

It's not tested, so it's broken. #36957 fixes that so it can be tested.

dimpase (approved):
perhaps add a comment on the list of packages above. Version generation may be done on a follow up PR.

mkoeppe:
This needs more discussion.

tornaria:
> > I'm starting to believe that when you say "modularity" you mean something different than what I understand by "modularity".
> > For the convenience of developers, IMO modularity means independent modules that can be built, tested, installed, updated, independently of each other. This includes having each its own dependencies, and trying to "synchronize up" or "automate" is against this.
> 
> Who's going to maintain it? As I explained above in [#36982 (comment)](https://github.com/sagemath/sage/pull/36982#issuecomment-1873020654), we don't have separate developer communities for the independent modules. Right we have distributions in `pkgs/` than developers who touch any of these version constraints. And even more with the modularization in #35095 adding more distributions.

Then sagemath is not ready for this type of modularization. It seems too much to me too...

mkoeppe:
> Then sagemath is not ready for this type of modularization.

No, it's not ready. I'm making it ready.

dimpase:
the task would have been much easier if it only involved sagelib proper, and not the scaffolding you have put around and insist on keeping.

jupyter, sphinx, compilers, patch, python3 don't belong here, it's a historic burden you insist on maintaining

mkoeppe:
> the task would have been much easier if it only involved sagelib proper

Nonsense, Dima.

tornaria:
> > Then sagemath is not ready for this type of modularization.
> 
> No, it's not ready. I'm making it ready.

You said "...we don't have separate developer communities for the independent modules...".

Thus sagemath is not ready for all this added complexity.

mkoeppe:
> You said "...we don't have separate developer communities for the independent modules...".
> 
> Thus sagemath is not ready for all this added complexity.

It's a vicious cycle, and I'm breaking it.

mkoeppe:
> this added complexity

Keeping the complexity controlled is exactly why I am keeping versions synchronized -- until there is a clear need for more flexibility.

tornaria:
> > You said "...we don't have separate developer communities for the independent modules...".
> > Thus sagemath is not ready for all this added complexity.
> 
> It's a vicious cycle, and I'm breaking it.

The point is: I don't think you are building separate developer communities. Not that you are expected to. But adding 50 different distributions is just making it harder. Not even `sagemath-standard` is working fine, let's get *that* working first. It's only one, and it's the main one. It sure can have its own *static* pyproject.toml, and build without preinstalling sage_setup and/or sage_conf.

dimpase:
it's a vicious circle only because you are forcing your own vision on everyone.

compilers, python, Jupyter, matplotlib is something only you want to keep under control - at a huge expense.

mkoeppe:
> The point is: I don't think you are building separate developer communities.

No, not yet. That can only happen after the modularization is merged.

> But adding 50 different distributions is just making it harder. 

Nope.

tornaria:
> > The point is: I don't think you are building separate developer communities.
> 
> No, not yet. That can only happen after the modularization is merged.
>
> > But adding 50 different distributions is just making it harder.
> 
> Nope.

Whatever...

dimpase:
we have a vicious circle of maintenance  of all that  stuff in sage distro we don't need to maintain

dimpase:
> > the task would have been much easier if it only involved sagelib proper
> 
> Nonsense, Dima.
> 
Huh?
How is not maintaining 300+ packages (which do not belong) make our task harder?

mkoeppe:
> [...] maintenance of all that stuff in sage distro [... ] does not go away with Matthias' modularisation

That's right, Dima, you got it! These two topics are entirely unrelated, which makes https://github.com/sagemath/sage/pull/36982#issuecomment-1874285785 off-topic.

dimpase:
> > > You said "...we don't have separate developer communities for the independent modules...".
> > > Thus sagemath is not ready for all this added complexity.
> > 
> > It's a vicious cycle, and I'm breaking it.
> 
> The point is: I don't think you are building separate developer communities. Not that you are expected to. 

Indeed, Matthias makes us slave away on sage distro we don't need. 
For the convenience of the developer, 
named Matthias

> But adding 50 different distributions is just making it harder. Not even `sagemath-standard` is working fine, let's get *that* working first. It's only one, and it's the main one. It sure can have its own *static* pyproject.toml, and build without preinstalling sage_setup and/or sage_conf.

dimpase:
> > [...] maintenance of all that stuff in sage distro [... ] does not go away with Matthias' modularisation
> 
> That's right, Dima, you got it! These two topics are entirely unrelated, which makes https://github.com/sagemath/sage/pull/36982#issuecomment-1874285785 off-topic.
> 
> 

the topics are related, as you insist that 300+ packages in Sage distro are needed for success of your modularisation.

but they are not, you are really delusional about it.

mkoeppe:
> I doesn't say in the documentation what actual filenames it scans. But https://github.com/sagemath/sage/network/dependencies shows that it finds files named `pyproject.toml` and `requirements.txt`.

Looking around in other repos, I see that it also accepts files named `*_requirements.txt` (https://github.com/numpy/numpy/network/dependencies).
So an easy solution for SBOM will be to just rename `install-requires.txt` to perhaps `version_requirements.txt`.

tobiasdiez:
> This needs more discussion.

@mkoeppe Could you then please share with us your reasons why you don't agree that the standard location `pyproject.toml` is a good place to store the build requirements for sage-the-library.

mkoeppe:
You are once more requesting info that I have already shared.

tobiasdiez:
> You are once more requesting info that I have already shared.

Since there was a fair amount of off topic discussion, I would indeed ask you for a summary of your points. Thanks

mkoeppe:
By the way, the content of `install-requires.txt` is the only think that links the SPKG name to the PyPI name; see #31136.

jhpalmieri:
Please do not change the label to "positive review". Users should not change labels on disputed tickets. Please wait for action by the Code of Conduct Committee or their designees.

orlitzky:
Well that didn't take long.

dimpase:
in other communications in the past 20 hours we are told that changing positively reviewed tickets back to needs review is not in accordance with the current Sage practices...

roed314:
> in other communications in the past 20 hours we are told that changing positively reviewed tickets back to needs review is not in accordance with the current Sage practices...

This PR was given positive review by @dimpase, though in this case after being given a disputed label.  In preparation for voting on disputed tickets, I'm restoring the positive review.  I will make a post on sage-devel shortly about voting.

mkoeppe:
I confirm my vote of -1 here. 

This PR is not an improvement.

orlitzky:
+1

We're going to keep fighting over the same things for as long as people who don't want to work on sage-the-distribution are forced to work on sage-the-distribution. Making the distribution a consumer of the library (and not the other way around) is the right move, for that and other reasons.

roed314:
Based on @kwankyu's [suggestion](https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/4Hls5k3MCgAJ) and discussion by the new Code of Conduct Committee, I'm setting this ticket to Needs Review so that all the disputed tickets start with a clean slate prior to being update based on voting.  Starting at midnight US/Pacific time tonight anyone involved should feel free to set the status to positive review or needs review in accordance with the [new policy](https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/kvmOlVb1AQAJ).

mkoeppe:
> > I doesn't say in the documentation what actual filenames it scans. But https://github.com/sagemath/sage/network/dependencies shows that it finds files named `pyproject.toml` and `requirements.txt`.
> 
> Looking around in other repos, I see that it also accepts files named `*_requirements.txt` (https://github.com/numpy/numpy/network/dependencies).

Another data point (thanks to @tobiasdiez via #37461): SciPy uses files `requirements/*.txt` since https://github.com/scipy/scipy/pull/19750; and GitHub recognizes these files too, as can be seen in https://github.com/scipy/scipy/network/dependencies?q=pydata-sphinx-theme

mkoeppe:
> We're going to keep fighting over the same things for as long as people who don't want to work on sage-the-distribution are forced to work on sage-the-distribution. Making the distribution a consumer of the library (and not the other way around) is the right move, for that and other reasons.

@orlitzky The `build/pkgs/*/install-requires.txt` files were NOT introduced as infrastructure of Sage-the-distribution. I introduced them in 2020 (#30719, #30913) to single-source the version constraints that would otherwise be duplicated in various different Python metadata files that were experiencing proliferation at the time: setup.cfg, pyproject.toml, requirements.txt, Pipfile (per @tobiasdiez's interest at that time), and anticipating more of this duplication as we were starting to preparing the modularized distributions.

mkoeppe:
> > The syncing of versions has to happen one way or the other. I think we all know the various possible designs, which differ by the time that the syncing happens and the actor that does the syncing.
> 
> And none of this is changed in this PR.

@tobiasdiez Well, sourcing the versions from the monolithic source tree in part from `SAGE_ROOT/src/`, as proposed here, *is* problematic from the angle of the monorepo/multirepo discussion. 
- It introduces a mutual dependency between sage-bootstrap and the (source code of) the Sage library: After the change proposed here, several metadata files (`requirements.txt`, `setup.cfg`, `Pipfile`) are to be generated by `bootstrap` from their .m4 templates. 
- If any part of the Sage library was to be split out to be maintained in a separate repository and its maintainers wish to use sage-bootstrap to obtain synchronized version dependencies (see #31662), it now has to depend on both sage-bootstrap and on the monolithic Sage source tree.

orlitzky:
> @orlitzky The `build/pkgs/*/install-requires.txt` files were NOT introduced as infrastructure of Sage-the-distribution.

I'm not objecting to it because they're fruit of the poisoned tree or anything like that. Today, they make the sage library depend on the sage distribution, and I would prefer if that dependency were inverted. As discussed earlier, I also think it's silly to purposely encode overly-strict dependency constraints (from the one source) in several submodules who a priori may have different requirements.

mkoeppe:
> Today, they make the sage library depend on the sage distribution

May I offer the alternative viewpoint that it depends on the scripts shipped by "sage-bootstrap"? after https://github.com/sagemath/sage/issues/31662, this would be installable from PyPI and bring the necessary scripts, and also suitable for use in multiple repos.

orlitzky:
> > Today, they make the sage library depend on the sage distribution
> 
> May I offer the alternative viewpoint that it depends on the scripts shipped by "sage-bootstrap"? after #31662, this would be installable from PyPI and bring the necessary scripts, and also suitable for use in multiple repos.

No, because I know where sage-bootstrap comes from. It starts with 's', ends with 'n', and isn't "spoon".

mkoeppe:
You can't get no ... satisfaction?

mkoeppe:
@orlitzky Have you actually looked at @vbraun's sage-bootstrap code? There's nothing wrong with it.

dimpase:
> No, because I know where sage-bootstrap comes from. It starts with 's', ends with 'n', and isn't "spoon".

spaghettification?

orlitzky:
> @orlitzky Have you actually looked at @vbraun's sage-bootstrap code? There's nothing wrong with it.

It's not the code I have a problem with, it's the design that requires me to download sage-the-distribution and bootstrap the sage library before I can work on it. (Or, OK, someone else to do it for me and upload the result to pypi.)

orlitzky:
> That's automatic.

Not in the scenario I mentioned, where I'm trying to work on sagelib. And even if it were (because, sure, it could be done), it would still be extremely complicated and just plain weird compared to... putting the dependencies in the file where the dependencies go, like everybody else.

tobiasdiez:
Setting this to positive review based on the following counting:
👍 me (author)
👍 @orlitzky 
👍 @dimpase 
👎 @mkoeppe

vbraun:
merge conflict

tobiasdiez:
> merge conflict

Rebased on top of https://github.com/sagemath/sage/pull/37796

vbraun:
merge conflict

mkoeppe:
This is broken, as `./bootstrap && cat m4/sage_spkg_versions_toml.m4` shows.

tobiasdiez:
> This is broken, as `./bootstrap && cat m4/sage_spkg_versions_toml.m4` shows.

For me that file doesn't show any difference to the one generated on the develop branch. I've moved the generation of the version requirements file further up in the bootstrap phase. Hard to say if that makes a difference for you since you (again) don't give any details of what is actually broken (in your opinion)...

dimpase:
> m4/sage_spkg_versions_toml.m4

I can confirm that I don't see a difference between m4/sage_spkg_versions_toml.m4
 and the one generated on 10.4.beta4 branch.

mkoeppe:
In 10.4.beta4:
```
$ grep cysignals m4/sage_spkg_versions_toml.m4
define(>>>SPKG_INSTALL_REQUIRES_cysignals<<<, >>>'cysignals >=1.10.2',<<<)dnl
```

Here:
```
$ grep cysignals m4/sage_spkg_versions_toml.m4
define(>>>SPKG_INSTALL_REQUIRES_cysignals<<<, >>><<<)dnl
```

mkoeppe:
That's on macOS, FWIW.

build/bin/sage-get-system-packages
```diff
@@ -42,11 +45,25 @@ case "$SYSTEM" in
...
 esac
-for PKG_BASE in $SPKGS; do
 
+for PKG_BASE in $SPKGS; do
+    if [ $FROM_PYPROJECT_TOML -eq 1 ]; then
+        if [ -f "$SAGE_ROOT/src/pyproject.toml" ]; then
+            # Extract from the "requires" block in src/pyproject.toml
+            REQUIRES=$(sed -n '/requires = \[/,/^\]/p' "$SAGE_ROOT/src/pyproject.toml")
+            # Packages are in the format "'sage-conf ~= 10.3b3',"
+            PACKAGE_INFO=$(echo "$REQUIRES" | sed -n "s/^\s*'\($PKG_BASE.*\)',/\1/p")
```
mkoeppe: This looks highly suspicious. Is this really how you want us to parse a complex file format like toml?

Makefile
```diff
@@ -167,7 +167,6 @@ bootstrap-clean:
 	rm -rf src/doc/en/reference/spkg/*.rst
 	for a in environment environment-optional src/environment src/environment-dev src/environment-optional; do rm -f $$a.yml $$a-3.[89].yml $$a-3.1[0-9].yml; done
 	rm -f src/Pipfile
-	rm -f src/pyproject.toml
 	rm -f src/requirements.txt
 	rm -f src/setup.cfg
```
mkoeppe, line 171: This also forgets to update `make bootstrap-clean` to remove the files generated in `bootstrap`.

bootstrap
```diff
@@ -35,6 +35,15 @@ CONFVERSION=$(cat $PKG/package-version.txt)
...
 
 bootstrap () {
+    for pkgname in cypari cysignals cython gmpy2 jupyter_core memory_allocator numpy pkgconfig pplpy setuptools wheel; do
+        # Write the version_requirements.txt files for dependencies declared in pyproject.toml
+        target=build/pkgs/${pkgname}/version_requirements.txt
+        if [ "${BOOTSTRAP_QUIET}" = "no" ]; then
+            echo "bootstrap:$LINENO: installing '"$target"'"
+        fi
+        echo "# Generated by SAGE_ROOT/bootstrap based on src/pyproject.toml; do not edit directly" > $target
+        sage-get-system-packages install-requires ${pkgname} >> $target
```
mkoeppe, line 45: This also forgets to put these files in the `configure` tarball.

mkoeppe:
> since you (again) don't give any details

@tobiasdiez The "again" is inappropriate, as you well know. And I have provided all necessary details, namely the file name of the file that is not generated correctly.

dimpase:
> > since you (again) don't give any details
> 
> @tobiasdiez The "again" is inappropriate, as you well know. And I have provided all necessary details, namely the file name of the file that is not generated correctly.

@mkoeppe - you have not provided details on how to achieve this discrepancy. On Fedora 38 the files are identical, and
on this branch I have
```
$ grep cysignals m4/sage_spkg_versions_toml.m4
define(>>>SPKG_INSTALL_REQUIRES_cysignals<<<, >>>'cysignals >=1.10.2',<<<)dnl
```

Maybe you are not testing the correct branch? Needless to say, version of macOS, hardware, whether or not Homebrew is there, is needed.

mkoeppe:
> Maybe you are not testing the correct branch?

You got to me joking.

dimpase:
> > Maybe you are not testing the correct branch?
> 
> You got to me joking.

no, it's the cursed macOS which is a joke. You uncovered a macOS-specific (or zsh, I don't know) bug.
Here is what  I see on our M1 box - the versions are all eaten up. 
```
dima@studio sage % diff m4/sage_spkg_versions_toml.m4 /tmp/sage_spkg_versions_toml.m4
36,38c36,38
< define(>>>SPKG_INSTALL_REQUIRES_cypari<<<, >>><<<)dnl
< define(>>>SPKG_INSTALL_REQUIRES_cysignals<<<, >>><<<)dnl
< define(>>>SPKG_INSTALL_REQUIRES_cython<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_cypari<<<, >>>'cypari2 >=2.1.1',<<<)dnl
> define(>>>SPKG_INSTALL_REQUIRES_cysignals<<<, >>>'cysignals >=1.10.2',<<<)dnl
> define(>>>SPKG_INSTALL_REQUIRES_cython<<<, >>>'cython >=3.0, != 3.0.3, <4.0',<<<)dnl
63c63
< define(>>>SPKG_INSTALL_REQUIRES_gmpy2<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_gmpy2<<<, >>>'gmpy2 ~=2.1.b999',<<<)dnl
90c90
< define(>>>SPKG_INSTALL_REQUIRES_jupyter_core<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_jupyter_core<<<, >>>'jupyter_core >=4.6.3',<<<)dnl
110c110
< define(>>>SPKG_INSTALL_REQUIRES_memory_allocator<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_memory_allocator<<<, >>>'memory_allocator',<<<)dnl
124c124
< define(>>>SPKG_INSTALL_REQUIRES_numpy<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_numpy<<<, >>>'numpy >=1.19',<<<)dnl
142c142
< define(>>>SPKG_INSTALL_REQUIRES_pkgconfig<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_pkgconfig<<<, >>>'pkgconfig',<<<)dnl
146c146
< define(>>>SPKG_INSTALL_REQUIRES_pplpy<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_pplpy<<<, >>>'pplpy >=0.8.6',<<<)dnl
214c214
< define(>>>SPKG_INSTALL_REQUIRES_setuptools<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_setuptools<<<, >>>'setuptools >= 68.1.1',<<<)dnl
258c258
< define(>>>SPKG_INSTALL_REQUIRES_wheel<<<, >>><<<)dnl
---
> define(>>>SPKG_INSTALL_REQUIRES_wheel<<<, >>>'wheel >=0.36.2',<<<)dnl
```

mkoeppe:
Not a bug. The sed expressions are underquoted.

dimpase:
in case:
```
$ checkbashisms bootstrap
possible bashism in bootstrap line 55 ('$(< foo)' should be '$(cat foo)'):
        echo 'changequote(`>>>'"'"', `<<<'"')dnl" >> $a
possible bashism in bootstrap line 55 (<<< here string):
        echo 'changequote(`>>>'"'"', `<<<'"')dnl" >> $a
possible bashism in bootstrap line 120 (<<< here string):
                echo "define(>>>SPKG_INSTALL_REQUIRES_${pkgname}<<<, >>>$(echo $(sage-get-system-packages install-requires ${pkgname}))<<<)dnl" >> m4/sage_spkg_versions.m4
possible bashism in bootstrap line 121 (<<< here string):
                echo "define(>>>SPKG_INSTALL_REQUIRES_${pkgname}<<<, >>>$(echo $(sage-get-system-packages install-requires-toml ${pkgname}))<<<)dnl" >> m4/sage_spkg_versions_toml.m4
possible bashism in bootstrap line 129 (<<< here string):
        echo 'changequote(>>>`<<<, >>>'"'"'<<<)dnl' >> $a
```

mkoeppe:
I've pushed a fix.

But really this kind of processing looks like a terrible idea. 

We are using our shell scripts when we process files that have a deliberately simple syntax.
Using it for processing toml files is clearly a bad idea.

mkoeppe:
> in case:
> 
> ```
> $ checkbashisms bootstrap
> possible bashism in bootstrap line 55 ('$(< foo)' should be '$(cat foo)'):
>         echo 'changequote(`>>>'"'"', `<<<'"')dnl" >> $a
> ```

All of these are false positives.

mkoeppe:
More problems with design and execution of this idea:

Removing the `version_requirements.txt` files from the repo -- bad idea because repology.org uses the presence of these files to distinguish Python from non-Python packages

Removing the `version_requirements.txt` files from the repo is also a broken design because (as I said already in https://github.com/sagemath/sage/pull/36982#issuecomment-1889772165 -- which got no reaction) these files are the only way that SPKG names are tied to PyPI names. The code here fakes it by accepting "cypari2" as `cypari` but will fail as soon as any package with dashes or underscores shows up, or one of the packages that have been renamed for uniqueness (such as "igraph" -> `python_igraph`).

dimpase:
> More problems with design and execution of this idea:
> 
> Removing the `version_requirements.txt` files from the repo -- bad idea because repology.org uses the presence of these files to distinguish Python from non-Python packages

 repology.org has nothing to do with  `version_requirements.txt` - or anything else - in vendored by Sage packages.

> 
> Removing the `version_requirements.txt` files from the repo is also a broken design because (as I said already in [#36982 (comment)](https://github.com/sagemath/sage/pull/36982#issuecomment-1889772165) -- which got no reaction) these files are the only way that SPKG names are tied to PyPI names.

One should use `install-requires.txt` for this, what's the point of multiplying instances?

mkoeppe:
> > Removing the `version_requirements.txt` files from the repo -- bad idea because repology.org uses the presence of these files to distinguish Python from non-Python packages
> 
> repology.org has nothing to do with `version_requirements.txt`

You declare this based on what, Dima?

mkoeppe:
> > Removing the `version_requirements.txt` files from the repo is also a broken design because (as I said already in [#36982 (comment)](https://github.com/sagemath/sage/pull/36982#issuecomment-1889772165) -- which got no reaction) these files are the only way that SPKG names are tied to PyPI names.
> 
> One should use `install-requires.txt` for this

You must have missed that `install-requires.txt` has been renamed as `version_requirements.txt`

Makefile
tobiasdiez: Done now.

bootstrap
tobiasdiez: Done now.

tobiasdiez:
> > since you (again) don't give any details
> 
> @tobiasdiez The "again" is inappropriate, as you well know. 

No, sadly the "again" is highly appropriate. Another example where you claim something is broken without giving enough details: https://github.com/sagemath/sage/pull/36753#issuecomment-1852312655 

> I've pushed a fix.

I thought we closed the topic of you abusing your admin right to push to my branches. Github provides you enough ways to suggest changes in more appropriate ways. But thank you for the fix.

> We are using our shell scripts when we process files that have a deliberately simple syntax. Using it for processing toml files is clearly a bad idea.

Toml files are not very complex, and only light parsing is needed here to extract the version information. But I do would welcome to completely replace the bash-bootstrap scripts by proper python scripts (and then use one of the proper toml parser). This is clearly out of the scope of this PR. 

> The code here fakes it by accepting "cypari2" as `cypari` but will fail as soon as any package with dashes or underscores shows up

Clearly not the case as `memory_allocator` is handled correctly.

Setting this back to positive review as the changes were trivial.

mkoeppe:
https://github.com/sagemath/sage/pull/36982#issuecomment-2081677839 is still not addressed.

mkoeppe:
> I thought we closed the topic of you abusing your admin right to push to my branches.

Also this rhetoric is inappropriate; don't do it, Tobias.

tobiasdiez:
Matthias, please stop removing the positive review label from this disputed PR. There are clear rules under which conditions this is should be done, you repeating your objections is not one of them.

To be explicit: I'm setting this to positive review based on the following counting:
👍 me (author)
👍 @orlitzky
👍 @dimpase
👎 @mkoeppe

mkoeppe:
> > Removing the `version_requirements.txt` files from the repo -- bad idea because repology.org uses the presence of these files to distinguish Python from non-Python packages
> 
> repology.org has nothing to do with `version_requirements.txt` - or anything else - in vendored by Sage packages.

@dimpase 
- https://github.com/repology/repology-updater/issues/1118
- https://github.com/repology/repology-updater/pull/1396

mkoeppe:
The dispute resolution process is intended to resolve disagreements in the review process.

It is not a replacement for the review process; and it's certainly not an invitation for authors to ignore review comments.

dimpase:
@mkoeppe - why do we care about whatever repology.org is doing in this case? It has no value, who cares about the version of jupiter-foobar you are bloating sage distro with?

mkoeppe:
Here's what a correct version of the idea of this PR would look like:
- Keep the `version_requirements.txt` files in the repository, but remove the version constraints from them.
  - This gives the mapping from SPKG names to PyPI distro names. (This is the only place that gives this mapping; see also #37500)
  - Their secondary function: Mark the packages as Python packages for [repology.org's index of the SageMath distribution](https://repology.org/projects/?search=&maintainer=&category=&inrepo=sagemath_develop&notinrepo=&repos=&families=&repos_newest=&families_newest=&outdated=on).
- Map to the PyPI name using the `version_requirements.txt` file, then read the version constraint from the `src/pyproject.toml` file.

mkoeppe:
> @mkoeppe - why do we care about whatever repology.org is doing in this case? It has no value, who cares about the version of jupiter-foobar you are bloating sage distro with?

@dimpase, necessary reminder: "How we talk about past, current, future contributions" (https://groups.google.com/g/sage-devel/c/OeN8o14s6Jc/m/ChnpijP3AgAJ). The denunciations that you are using, the expressions of disrespect, etc., have no place in the Sage community.

roed314:
@tobiasdiez is correct about our procedure for disputed tickets.  @mkoeppe pointed out some changes that he thinks are necessary, but the author and reviewer disagree and the vote is currently 3-1 in favor of this ticket.  Others are, of course welcome to weigh in, but for now I'm setting this back to positive review.

mkoeppe:
@roed314, thanks for the fast clarification about our procedures! Reminder though that I'm waiting for your response on some related matters since January.

mkoeppe:
The broken repology index can now be seen in https://repology.org/projects/?inrepo=sagemath_develop

dimpase:
> This has broken the CI.
> https://github.com/sagemath/sage/actions/runs/8931531493/job/24533776262#step:11:3704

congratulations, you managed to create a very fragile thing noone understands.
No idea what docker is doing there, and why.

PR #36983, opened 2023-12-31
Prepare python3 `spkg-configure.m4` for Python 3.12 (setuptools instead of distutils)

mkoeppe:
- Cherry-picked from #36181

This PR does not change what versions are accepted!
To test with Python 3.12, still need to adjust the version bound set in `build/pkgs/python3/spkg-configure.m4`.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #36957 (merged here to resolve merge conflict)


tornaria:
Looks ok to me, but I don't have the means to test this. There's too many in-flight stuff going on. I only got chatting with you about this since I wanted to have a try at #36957 just to see how it works, but not even this PR would break it (I don't have the inclination to work or review 10 prs just to test a thing I don't really care about and don't use / won't use).

I'd be tempted to positive review this PR on the basis of checks passing and you knowing what you are doing.

The situation brings to mind this: https://martinfowler.com/articles/ship-show-ask.html

mkoeppe:
> I'd be tempted to positive review this PR on the basis of checks passing and you knowing what you are doing.

Good enough for me.

mkoeppe:
Thanks!

vbraun:
Merge conflict

m4/sage_check_python_for_venv.m4
```diff
@@ -16,11 +16,23 @@ AC_DEFUN([SAGE_CHECK_PYTHON_FOR_VENV], [
...
                         AX_COMPARE_VERSION([$python3_version], [lt], LT_VERSION, [
                             dnl Because the system python is not used directly but rather in a venv without site-packages,
-                            dnl we test whether the module will be available in a venv.
+                            dnl we should test whether the module will be available in a venv.
                             dnl Otherwise, some system site-package may be providing this module to the system python.
+                            dnl However, on Python >= 3.12, we need setuptools to run our extension compilation tests
+                            dnl because distutils has been removed from the standard library.
+                            AX_COMPARE_VERSION([$python3_version], [ge], [3.12.0], [
+                                conftest_venv_options="--system-site-packages"
+                                distutils_core="setuptools"
```
tornaria, line 25: Can we move to setuptools unconditionally, instead of only for python 3.12?

mkoeppe: Not really, that would require people to have an additional system package installed that isn't going to be used other than for this configure check.

tornaria: It's just setuptools. Besides, what makes it ok to require setuptools for python 3.12 but not for python 3.11?

In fact, it seems you don't really need to require the presence of neither distutils nor setuptools at all. If distutils or setuptools is available from system, it can be used, otherwise setuptools will be installed in the venv and everything will be fine.

The only reason one checks for a few modules here is because they are built-in python modules that can't be installed after the fact on top of a system python. But setuptools can (and will) be installed.

mkoeppe: > what makes it ok to require setuptools for python 3.12

You know the answer. It's OK because without it we cannot test what we want to test.

mkoeppe: > If distutils or setuptools is available from system, it can be used, otherwise setuptools will be installed in the venv and everything will be fine.

Not really. We stopped using distutils from the system a long time ago (by setting `SETUPTOOLS_USE_DISTUTILS=local`), and we do not use setuptools from the system (except in the experimental system-site mode, which I'll not comment on).

mkoeppe: > The only reason one checks for a few modules here is because they are built-in python modules that can't be installed after the fact on top of a system python.

No, that's not the reason for the check of distutils at all.

The checks with distutils are there to catch toolchain misconfigurations and to work around some of them.

tornaria: > > what makes it ok to require setuptools for python 3.12
> 
> You know the answer. It's OK because without it we cannot test what we want to test.

My question (that you deleted) was: "why is it not ok to require setuptools for python 3.11". Point is, whatever works for 3.12 should work for 3.11 (and 3.10 and 3.9).

Also, no, I have no idea what we need to test and why, given that sage-the-distro will install setuptools in any case.


> The checks with distutils are there to catch toolchain misconfigurations and to work around some of them.

What misconfigurations? 


> > If distutils or setuptools is available from system, it can be used, otherwise setuptools will be installed in the venv and everything will be fine.
> 
> Not really. We stopped using distutils from the system a long time ago (by setting `SETUPTOOLS_USE_DISTUTILS=local`), and we do not use setuptools from the system (except in the experimental system-site mode, which I'll not comment on).

Ok, then I'm confused. We do not use setuptools from the system, why are we still checking if it's available?

mkoeppe: As I said, we use system distutils (and with python 3.12, system setuptools) to catch toolchain misconfigurations and to work around some of them.

See https://github.com/sagemath/sage/blob/develop/build/pkgs/python3/spkg-configure.m4 for what is checked

tornaria: I found the origin for this here: https://github.com/sagemath/sage/issues/27824#issuecomment-1418033698

It seems to me that a simple AC_CHECK_HEADER would suffice to see if python headers are installed. It doesn't matter if the system distutils can compile since building is going to use setuptools anyway.

Other forms of misconfigured toolchain would break a spkg-installed python as much as a system python.

mkoeppe: Nope, it's unfortunately more complicated than you would think.

Source: I wrote and debugged it on the whole range of supported systems.

tornaria: :shrug: *broken* supported systems (that's what "misconfigured" means but ymmv and not a native speaker).

In any case I included this PR verbatim in #37270, what is missing there?

mkoeppe: What's missing is a review of the PR here, and listing it on #37270 as a dependency.

mkoeppe: > _broken_ supported systems (that's what "misconfigured" means but ymmv and not a native speaker).

It seems like you're eager to explain something to me, just what?

tornaria: https://github.com/sagemath/sage/pull/36983#issuecomment-1874240195

tornaria: > > _broken_ supported systems (that's what "misconfigured" means but ymmv and not a native speaker).
> 
> It seems like you're eager to explain something to me, just what?

I read "misconfigured toolchain" as "broken system". Does "the whole range of supported systems" includes "broken systems" for you? Where is an example of this?

mkoeppe: > #36983 (comment)

Thanks, I forgot about that.

mkoeppe: > I read "misconfigured toolchain" as "broken system". Does "the whole range of supported systems" includes "broken systems" for you? Where is an example of this?

I would say the way that we are using system Python installations is pretty much at the edge of their specifications, and there is a lot of ambiguity about what should be expected to work. 

distutils has always been underspecified and underdeveloped, historically there was not a good practice of upstreaming the platform customizations of distutils, and often it's not even clear whether the system Python is even intended to be used by Python developers or only by the distribution itself as a deployment vehicle for applications written in Python. This situation is often acknowledged with the mantra "distro Python is not for you", and people just avoid touching it.

One of the underspecified behaviors is how distutils should behave when runtime environment variables modify the sysconfig variables that come from the compile time of CPython. We had quite a bit of trouble with this on macOS (both plain Xcode and homebrew).
An example related to the role of distro Python is https://github.com/python/cpython/issues/90539#issuecomment-1153274032, with problems with Python from multiple vendors.

tornaria: You avoid the question. Where's an example of this *in sagemath*?

mkoeppe: I don't understand. All of what I wrote is about SageMath

tornaria: "We had quite a bit of trouble with this on macOS (both plain Xcode and homebrew)."

I was asking for an example of this, i.e. broken systems that cause trouble *with sagemath*.

mkoeppe: macOS Xcode + homebrew was pretty problematic in #31227.

And the link to https://github.com/python/cpython/issues/90539#issuecomment-1093942277 that I sent earlier has the pointers to Sage tickets where Python was unusable on Homebrew, Cygwin, and pyenv.

mkoeppe: Also openSUSE was affected by the same (#33153).

tornaria: Ok, thanks.

PR #36986, opened 2024-01-01
Fixed a bug in `AbelianGroup.Subgroup.gens_orders()`, which returns the reduced order of the subgroup

### LINKED ISSUE #36974: Bug in `AbelianGroup.Subgroup`
### Steps To Reproduce

```python
sage: G, (g0, g1) = AbelianGroup(2, [48, 0]).objgens()
....: print(G, g0, g1)
....: assert g0.order() == 48 and g1.order() == oo
....: 
....: G0 = G.subgroup([g0])
....: print(G0, G0.gens(), G0.gens_orders())
....: print(len(G0.gens()) == len(G0.gens_orders()))
Multiplicative Abelian group isomorphic to C48 x Z f0 f1
Multiplicative Abelian subgroup isomorphic to C3 x C16 generated by {f0} (f0,) (3, 16)
False
```

Problem: The length of `G0.gens()` and `G0.gens_orders()` should be the same.

### Additional Information

The `Subgroup` is computed [here](https://github.com/sagemath/sage/blob/develop/src/sage/groups/abelian_gps/abelian_group.py#L1710). The issue seems to be that the GAP functions `AbelianInvariants` doesn't actually return the orders of the generators in `GeneratorsOfGroup`, as demonstrated below:
```python
sage: G, (g0, g1) = AbelianGroup(2, [48, 0]).objgens()
....: H = libgap(G).Subgroup(gens)
....: H_invs = H.AbelianInvariants().sage()
....: H_gens = H.GeneratorsOfGroup()
....: print(H, H_invs, H_gens)
Pcp-group with orders [ 48 ] [3, 16] [ g1 ]
```

The obvious fix would be to replace `.AbelianInvariants` on line 1710 with, well, `[g.Order().sage() for g in H.GeneratorsOfGroup()]` or something like that. If any GAP users have a better solution that returns the order of all (finite?) generators, please suggest it, since I haven't used GAP before.

amanmoon:
The `gens_orders()` method will returns the correct order after these changes. 

Created a new variable `gens_orders`:
```python
gens_orders = [order_sage for g in H.GeneratorsOfGroup() if (order_sage := g.Order().sage()) is not infinity]
```
Modified the `AbelianGroup_class` class to include an additional argument, from which gens_orders is passed and assigned to `self.gens_orders`:
```python
def __init__(self, generator_orders, names, gens_orders=None, category=None):
self._gens_orders = generator_orders if gens_orders is None else gens_orders
```
Also, tests have been written for the `gens_orders` method:
```sage
sage: G, (g0, g1) = AbelianGroup(2, [48, 0]).objgens()
sage: G0 = G.subgroup([g0])
sage: len(G0.gens()) == len(G0.gens_orders())
True
```
Additionally, performed code cleanup.

Fixes #36974

Changing the `invs` variable directly to `[order_sage for g in H.GeneratorsOfGroup() if (order_sage := g.Order().sage()) is not infinity]` [here](https://github.com/sagemath/sage/blob/e249befd47ad8610f74372533a2943ff7b2dde94/src/sage/groups/abelian_gps/abelian_group.py#L1710) causes series of errors, as other variables depend on it.
### :memo: Checklist


src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -976,13 +978,15 @@ def gens_orders(self):
             sage: Z2xZ3 is Z6
             False
 
-        TESTS::
 
+        TESTS::
+            sage: G, (g0, g1) = AbelianGroup(2, [48, 0]).objgens()
```
tscrim:
```suggestion
        TESTS::

            sage: G, (g0, g1) = AbelianGroup(2, [48, 0]).objgens()
```
This needs a blank line between as it was before the change.

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -976,13 +978,15 @@ def gens_orders(self):
...
+            sage: G0 = G.subgroup([g0])
+            sage: len(G0.gens()) == len(G0.gens_orders())
+            True
             sage: F = AbelianGroup(3, [2], names='abc')
             sage: list(map(type, F.gens_orders()))
-            [<class 'sage.rings.integer.Integer'>,
-             <class 'sage.rings.integer.Integer'>,
-             <class 'sage.rings.integer.Integer'>]
+            [<class 'sage.rings.integer.Integer'>, <class 'sage.rings.integer.Integer'>, <class 'sage.rings.integer.Integer'>]
```
tscrim: This change is both unnecessary and make the output harder to read. Please revert.

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1770,8 +1777,7 @@ def __contains__(self, x):
                 ZZ, len(self._gens), len(amb_inv),
                 [g.list() for g in self._gens]
             )
-            return (vector(ZZ, x.list())
-                in inv_basis.stack(gens_basis).row_module())
+            return (vector(ZZ, x.list()) in inv_basis.stack(gens_basis).row_module())
```
tscrim: Similarly this change should be reverted as it is unnecessary.

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1627,7 +1631,7 @@ def __init__(self, ambient, gens, names="f", category=None):
             Multiplicative Abelian subgroup isomorphic to Z x Z generated by {a^3, b}
             sage: F.subgroup([c])
             Multiplicative Abelian subgroup isomorphic to C2 x C3 x C5 generated by {c}
-            sage: F.subgroup([a, c])
+            sage: S = F.subgroup([a, c]); S
```
tscrim: I don't see why you need to hold onto this subgroup. I would revert this change.

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1175,7 +1179,7 @@ def random_element(self):
             order = g.order()
             if order is infinity:
                 order = 42  # infinite order; randomly chosen maximum
-            result *= g**(randint(0,order))
+            result *= g**(randint(0, order))
```
tscrim:
```suggestion
            result *= g ** (randint(0, order))
```
If you are going through and making PEP8 changes (although I generally advise against that to avoid merge conflicts and noise in a change; here it is fairly minimal, so IMO it isn't a problem).

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1708,16 +1712,19 @@ def __init__(self, ambient, gens, names="f", category=None):
         H = libgap(ambient).Subgroup(H_gens)
 
         invs = H.TorsionSubgroup().AbelianInvariants().sage()
+        gens_orders = [order_sage for g in H.GeneratorsOfGroup() if (order_sage := g.Order().sage()) is not infinity]
```
tscrim:
```suggestion
        gens_orders = tuple([ZZ(order_sage) for g in H.GeneratorsOfGroup()
                             if (order_sage := g.Order().sage()) is not infinity])
```

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1708,16 +1712,19 @@ def __init__(self, ambient, gens, names="f", category=None):
...
+
         rank = len([1 for g in H.GeneratorsOfGroup()
                     if g.Order().sage() is infinity])
         invs += [0] * rank
-
+        gens_orders += [0] * rank
```
tscrim:
```suggestion
        gens_orders += (ZZ.zero(),) * rank
```

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1708,16 +1712,19 @@ def __init__(self, ambient, gens, names="f", category=None):
...
         self._abinvs = invs
         invs = tuple(ZZ(i) for i in invs)
+        gens_orders = tuple(ZZ(i) for i in gens_orders)
```
tscrim:
```suggestion
```
This has become unnecessary as the above suggestion should be faster (less transient objects).

tscrim (changes requested):
I think the general idea of not passing `invs` as the generators of the subgroup is the correct idea. However, you are effectively creating redundant information with the additional argument. Plus there becomes an inconsistency with the variable names being created. You can see this by
```
sage: S.inject_variables()
Defining f0, f1
sage: f1
NameError                                 Traceback (most recent call last)
Cell In[13], line 1
----> 1 f1

NameError: name 'f1' is not defined
```
Instead, you should simply pass the `gens_orders` up as the first argument to `AbelianGroup_class.__init__` (after `self` of course).

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -976,13 +978,18 @@ def gens_orders(self):
             sage: Z2xZ3 is Z6
             False
 
+
```
tscrim: This blankline is not needed.

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -976,13 +978,18 @@ def gens_orders(self):
...
+            sage: G0 = G.subgroup([g0])
+            sage: len(G0.gens()) == len(G0.gens_orders())
+            True
             sage: F = AbelianGroup(3, [2], names='abc')
             sage: list(map(type, F.gens_orders()))
             [<class 'sage.rings.integer.Integer'>,
-             <class 'sage.rings.integer.Integer'>,
-             <class 'sage.rings.integer.Integer'>]
+            <class 'sage.rings.integer.Integer'>,
+            <class 'sage.rings.integer.Integer'>]
```
tscrim: The extra space was good here to indicate the list output. Again, please revert back to what was there.

src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1708,16 +1715,19 @@ def __init__(self, ambient, gens, names="f", category=None):
         H = libgap(ambient).Subgroup(H_gens)
 
         invs = H.TorsionSubgroup().AbelianInvariants().sage()
+        gens_orders = tuple([ZZ(order_sage) for g in H.GeneratorsOfGroup()
+                    if (order_sage := g.Order().sage()) is not infinity])
```
tscrim:
```suggestion
        gens_orders = tuple([ZZ(order_sage) for g in H.GeneratorsOfGroup()
                             if (order_sage := g.Order().sage()) is not infinity])
```
Better alignment IMO (and we generally do this (now) in the Sage source from PEP8).

tscrim (commented):
Thank you. Just a few more minor details.

tscrim (approved):
Thank you. LGTM.

vbraun:
merge conflict

amanmoon:
@vbraun

tscrim:
@amanmoon That's not what the automated checker is saying. Please try again by either merging or rebasing (I recommend the latter) over the most recent develop branch.

tscrim:
Thanks. This is back to a positive review.

dimpase:
This has broken `.subgroup(...)` rather badly, see [sage-devel](https://groups.google.com/g/sage-devel/c/OpwSERIr4aM/m/z-_6p4ZrAwAJ)
e.g. with this one gets 
```
sage: g=AbelianGroup([2,2])
sage: g.cardinality().factor()
2^2
sage: gg=g.subgroup(g.list())
sage: gg.cardinality().factor()
2^3
```

PR #36988, opened 2024-01-01
Adding support for sums, intersection, and equality of SubmodulesWithBasis

tscrim:
Common vector subspace operations are taking sums, intersections, and checking equality, but these are not implemented (inclusion already is). We provide generic implementations of these for objects in `ModulesWithBasis` and make a few other minor improvements.

### :memo: Checklist


src/sage/modules/with_basis/subquotient.py
```diff
@@ -377,15 +377,217 @@ def is_submodule(self, other):
...
+            sage: F.is_equal_subspace(F)
+            True
+            sage: F == G
+            False
+            sage: F.is_equal_subspace(G)
+            True
+            sage: F == H  # different support orders
+            False
+            sage: F.is_equal_subspace(G)
+            True
```
mantepse, line 517: I think you meant to test against `H` here, no?

src/sage/modules/with_basis/subquotient.py
```diff
@@ -377,15 +377,217 @@ def is_submodule(self, other):
...
+
+    def is_equal_subspace(self, other):
+        r"""
+        Return whether ``self`` is an equal submodule to ``other``.
+
+        .. NOTE::
+
+            This is the mathematical notion of equality (as sets that are
+            isomorphic as vector spaces), which is weaker than the `==`
+            which takes into account things like the support order.
```
mantepse, line 496: I must admit that I fail to understand the notion of equality.  If I understand correctly, instances of `SubmoduleWithBasis` are equal if and only if they are identical (as `CombinatorialFreeModule`).  This means, that they have the same basis, the same prefix, and even the same LaTeX options:
```
            sage: F = CombinatorialFreeModule(QQ, ['a','b','c'])
            sage: G = CombinatorialFreeModule(QQ, ('a','b','c'))
            sage: F is G
            True

            sage: F = CombinatorialFreeModule(QQ, ['a','b','c'], latex_bracket=['LEFT', 'RIGHT'])
            sage: F.print_options()['latex_bracket']
            ('LEFT', 'RIGHT')

            sage: F is G
            False
```
(I think that the latter is *extremely* strange!)

I would have expected that 'mathematical equality' means that they have the same elements, i.e., their bases span the same space.  Am I mistaken?  The reason for my doubts is in lines 463 and 464:
```
if set(self._support_order) != set(other._support_order):  # different supports
    return False
```
Please clarify!

tscrim: > I must admit that I fail to understand the notion of equality. If I understand correctly, instances of `SubmoduleWithBasis` are equal if and only if they are identical (as `CombinatorialFreeModule`). This means, that they have the same basis, the same prefix, and even the same LaTeX options:

Yes, this is correct. We want parents to be unique; equal-but-not-identitical parents can be a pain to work with. For the most part, a user who is passing such options is likely meaning for them to be different. Yes, mathematically it is slightly strange, but I think it is quite natural from the programming side.

> I would have expected that 'mathematical equality' means that they have the same elements, i.e., their bases span the same space. Am I mistaken? The reason for my doubts is in lines 463 and 464:
> 
> ```
> if set(self._support_order) != set(other._support_order):  # different supports
>     return False
> ```
> 
> Please clarify!

I don't understand why you doubt this. This is a very quick check that the support of the subspace (the union of the supports of all of the vectors in the subspace) are equal. This is clearly a necessary condition for the subspaces to be equal, otherwise there is a vector in one subspace that uses a basis vector in the ambient space that never appears in the other.

mantepse: Ah, now it makes sense, thank you. I somehow thought that an element in self._support would be a list of coefficient, basis vector pairs. Sorry for the noise.

mantepse (approved):
Lgtm

tscrim:
Thank you, although I realized that `_common_submodules` needs a doctest (I forgot to add it, whoops!). I will push another change once I test, which will happen after Sage finishes upgrading.

mantepse:
Oh dear, I missed it too. Thank You!

tscrim:
Okay, I added the tests.

mantepse (approved):
:-)

tscrim:
Thank you!

PR #36994, opened 2024-01-02
Use PARI when computing the Tate pairing

GiacomoPope:
To match `weil_pairing()` this pull request now uses PARI to compute the non-reduced Tate pairing for elliptic curves over finite fields. 

This has been included for efficiency, with PARI offering a significant performance speed up:

```py
sage: GF(65537^2).inject_variables()
Defining z2
sage: E = EllipticCurve(GF(65537^2), [0,1])
sage: P = E(22, 28891)
sage: Q = E.random_point()
sage: # Old timing
sage: %timeit P.tate_pairing(Q, 7282, 2)
1.4 ms ± 66 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
sage: # New timing
sage: %timeit P.tate_pairing(Q, 7282, 2)
50.8 µs ± 674 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

### :memo: Checklist

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_point.py
+++ b/src/sage/schemes/elliptic_curves/ell_point.py
@@ -1917,12 +1917,32 @@ class EllipticCurvePoint_field(SchemeMorphism_point_abelian_variety_field):
 
         TESTS:
 
+        Check that the PARI output matches the original Sage implementation::
+
+            sage: # needs sage.rings.finite_rings
+            sage: GF(65537^2).inject_variables()
+            Defining z2
+            sage: E = EllipticCurve(GF(65537^2), [0,1])
+            sage: P = E(22, 28891)
+            sage: Q = E(-93, 40438*z2 + 31573)
+            sage: P.tate_pairing(Q, 7282, 2)
+            34585*z2 + 4063
+
         The point ``P (self)`` must have ``n`` torsion::
 
             P.tate_pairing(Q, 163, 2)
             Traceback (most recent call last):
             ...
-            ValueError:  The point P must be in the n-torsion
+            ValueError: The point P must be in the n-torsion
+
+        The Tate pairing is only defined for points on curves defined over finite fields::
+
+            sage: E = EllipticCurve([0,1])
+            sage: P = E(2,3)
+            sage: P.tate_pairing(P, 6, 1)
+            Traceback (most recent call last):
+            ...
+            NotImplementedError: Reduced Tate pairing is currently only implemented for fintie fields
 
         ALGORITHM:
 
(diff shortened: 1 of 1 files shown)


GiacomoPope:
I'm now not actually convinced that the `tate_pairing()` is ever expected to be called over non-finite fields... I need to read more literature but if we only consider `E/K` with `characteristic(K) > 0`, then we can simply always use a pari call and simplify this function

GiacomoPope:
Rather than allow Pari to be used when the field is finite, we now assume that the field is finite for the reduced tate pairing, return a NotImplemented error when the field is not finite and always use Pari for the performance boost.

GiacomoPope:
I'm a bit confused about my doctests here... `codecov/patch` is highlighting that I didn't test for the check on line 1978, however this value error is precisely what I check in the `TESTs:` on lines 1931-1936, so I'm not sure what I'm doing wrong here.

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -1944,21 +1974,17 @@ def tate_pairing(self, Q, n, k, q=None):
...
-        try:
-            ret = self._miller_(Q, n)
-            e = Integer((q**k - 1)/n)
-            ret = ret**e
-        except (ZeroDivisionError, ValueError):
-            R = E.random_point()
-            ret = self.tate_pairing(Q + R, n, k)/self.tate_pairing(R, n, k)
-        return ret
+        if pari.ellmul(E, P, n) != [0]:
+            raise ValueError("The point P must be in the n-torsion")
```
JohnCremona: Delete "in the" for better English

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -1944,21 +1974,17 @@ def tate_pairing(self, Q, n, k, q=None):
...
+
+        # NOTE: Pari returns the non-reduced Tate pairing, so we
+        # must perform the exponentation ourselves using the supplied
+        # k value
+        ePQ = pari.elltatepairing(E, P, Q, n)
+        # TODO: if n or k is chosen badly, this could error, should we
+        # handle this explicitly by ensuring n divides q^k - 1?
+        exp = Integer((q**k - 1)/n)
```
JohnCremona, line 1995: I think it is a good idea to check that n divides q^k-1, for example by computing mod(q,n)^k, but only if q has been explicitly supplied and with a suitable error message if the condition fails.  So do this test in the code block which computes q perhaps.

JohnCremona (changes requested):
It's a good idea to use pari where it has the functionality, as here (but was not the case back in 2011 I think).

I only made two suggestions, one a trivial one about wording and the other in response to the TODO.

GiacomoPope:
I have fixed the English in one of the ValueErrors and also added a new check that `n` divides `(q^k - 1)` which is only checked when `q` is not `None` (The case where a user supplies this information).

JohnCremona (approved):
Good, thanks!

PR #36998, opened 2024-01-03
Append (not symlink) `config.log` to `logs/pkgs/config.log`

mkoeppe:
### :memo: Checklist

We change the handling of `config.log`, previously set up as a symlink into `logs/pkgs/`. We sometimes get reports about build problems with a `config.log` attached that just shows "... already installed as an SPKG", so we cannot see what caused the SPKG to be installed in the first place. We fix this problem by appending the current contents of `config.log` (whenever it has changed) to `logs/pkgs/config.log`.

Split out from #33262.


mkoeppe:
Not clear yet if this is a complete solution. If people just run `./configure` and it fails, then `logs/pkgs/config.log` is of course not created yet. So perhaps we'd have to ask  users to send both `config.log` and `logs/pkgs/config.log`....

kwankyu:
How about creating `config.log` at sage root directory, and after sage is built, the `config.log` is copied to `logs/` (instead of `logs/pkgs`) with the timestamp appended to the name?

PR #36999, opened 2024-01-03
Rename `install-requires.txt` to `version_requirements.txt`

mkoeppe:
As discussed in #36982:
- the name "install-requires"  has become outdated with the transition to the new names set by the `pyproject.toml` format (see https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#declaring-required-dependency, compare the tabs "pyproject.toml" vs. "setup.cfg")
- this will help with https://github.com/sagemath/sage/issues/35890, as GitHub will be able to just read the `version_requirements.txt` files 

https://github.com/sagemath/sage/pull/36982#issuecomment-1874857075

Notes for reviewers:
- **This is a simple, limited-scope improvement and not intended as a long-term commitment to the format of the files in build/pkgs/....** 
- PRs #37430, #37350, #36740 remove direct access to build/pkgs in favor of using the sage_bootstrap API (sage --package).

### :memo: Checklist

### :hourglass: Dependencies

- Depends on https://github.com/sagemath/sage/pull/37401


kwankyu:
> * the name "install-requires"  has become outdated with the transition to the new names set by the `pyproject.toml` format (see https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#declaring-required-dependency, compare the tabs "pyproject.toml" vs. "setup.cfg")

There it is "dependencies"?

mkoeppe:
That's right. The term "install-requires" has been eliminated.

kwankyu:
Then you made up the name "version_requirements.txt", instead of using "dependencies"?

mkoeppe:
That's right.

kwankyu:
(we may come up with a name better than both, but I have no idea now...)

mkoeppe:
it needs to end with "_requirements.txt" so that GitHub finds it (see link in PR description). I didn't have a better idea what to stick in front.

kwankyu:
Simply "install_requirements.txt"?

mkoeppe:
For inspiration: https://packaging.python.org/en/latest/specifications/dependency-specifiers/

kwankyu:
I cannot come up with an obviously better name than "version_requirements.txt". This file is about version requirements on dependencies (pip packages on pypi). So other possible names are "dependency_requirements.txt" or "pip_requirements". No strong opinion.

dimpase:
I still don't get why we cannot just use `requirements.txt`. 

Same format, essentially the same purpose, ease of conversion to pip package.

kwankyu:
Perhaps because an spkg is not a python package?

kwankyu:
Another suggestion: "package_requirements".

dimpase:
> Perhaps because an spkg is not a python package?

These are Python packages we are talking about, these which are vendored by Sage, i.e. made into spkg packaged as tarfiles or wheels. Package names in `install-requires.txt` are not spkg names, these are Python package names.
You can rename  `install-requires.txt` to `requirements.txt`, remove `package_version.txt` and `checksums.ini`, and usually 
such a package can still be installed, as a PyPI package, by-passing Sage's tarballs/wheels.

If it's a standard package, it's an illegal operation under the current policy; as you know I called a sage-devel vote on changing this policy...

kwankyu:
I agree that "requirements.txt" seems the best choice. According to the developer guide, presence of "requirements.txt" is used to differentiate "pip" spkg from "normal" and "wheel" spkg. How about using presence of "checksums.ini" for this purpose? Do I miss something? Is there other reason that "requirements.txt" cannot be a choice?

kwankyu:
I agree that "requirements.txt" is the best choice. According to the developer guide, presence of "requirements.txt" is used to differentiate "pip" spkg from "normal" and "wheel" packages (in which "install-requires" is used). But "checksums.ini" may be used for the purpose. Is there other technical reason that  "requirements.txt" cannot be a choice?

mkoeppe:
> I still don't get why we cannot just use `requirements.txt`.

I explained it to you in https://github.com/sagemath/sage/pull/37401#issuecomment-1954727912; complete the review there, please

mkoeppe:
> According to the [developer guide](https://doc.sagemath.org/html/en/developer/packaging.html#:~:text=To%20summarize%3A%20the,a%20dummy%20package.), presence of "requirements.txt" is used to differentiate "pip" spkg from "normal" and "wheel" spkg. 

Yes, that's right.

> How about using presence of "checksums.ini" for this purpose?

This could be done (best after the refactoring in #36740, which needs review), but I would still not use `requirements.txt` for the acceptable version ranges of normal/wheel packages.

kwankyu:
> Because `requirements.txt` means something else. It is the name one uses for version pinnings, as in `pip install -r requirements.txt`.

But you can use version ranges too in the file.  

> `install-requires` is the traditional name (in setuptools) for declaring dependencies.

Yes. But as far as I understand, `install-requires` in an spkg specifies the "dependency" of the spkg, which is the pip package that the spkg installs. The "dependency" can be well described by the "requirements.txt" file. (the spkg "requires" the specified versions.)

mkoeppe:
> > Because `requirements.txt` means something else. It is the name one uses for version pinnings, as in `pip install -r requirements.txt`.
> 
> But you can use version ranges too in the file.

I should have written: `requirements.txt` ... is the **file name** one uses for version pinnings, as in `pip install -r requirements.txt`.

mkoeppe:
In Python packaging, one distinguishes "abstract dependencies" (install-requires) vs "concrete dependencies" (requirements.txt). See e.g. https://martin-thoma.com/python-requirements/#abstract-dependencies

That's why using the name `requirements.txt` for declaring our acceptable version ranges would be bad -- it would confuse anyone who is familiar with Python packaging practices.

kwankyu:
> In Python packaging, one distinguishes "abstract dependencies" (install-requires) vs "concrete dependencies" (requirements.txt). See e.g. https://martin-thoma.com/python-requirements/#abstract-dependencies
> 
> That's why using the name `requirements.txt` for declaring our acceptable version ranges would be bad -- it would confuse anyone who is familiar with Python packaging practices.

Thanks for the read. I see now why you reject the name `requirements.txt`: In python packagaing, `install-requires` and `requirements.txt` have (loosely) separate uses.

> I chose the name "install-requires.txt" to mean: If a Python distribution needs to declare this package as an "install-requires", use the contents of this file for it. And that's what is done when we create `pkgs/*/setup.cfg` from the corresponding `setup.cfg.m4`.

This explains why I got confused. The name "install-requires.txt" cannot be explained within spkg context. You see the spkg as a "dependency" of the Python distribution, and the content of `install-requires.txt` is supposed to be used by the Python distribution. I was thinking within spkg context.

Now the Python distribution uses `pyproject.toml`, and the file uses the name "dependencies" instead of old "install-requires".  Then logically, you wan to replace "install-requires.txt" with "dependencies.txt", but "dependencies" is already being used for spkg dependencies. Hence we need a new name. Am I right?

mkoeppe:
> logically, you wan to replace "install-requires.txt" with "dependencies.txt", but "dependencies" is already being used for spkg dependencies. Hence we need a new name. Am I right?

Exactly that. "install-requires" was an idiosyncratic name only used for one purpose in Python packaging, and hence it was unambiguous. "dependencies" is not just already used, it's also too generic.

kwankyu:
> No, that's not a good way of viewing the relation of the spkg and the pip package. The spkg does not "depend" on it. It "represents" it.

OK. Now I agree that "version_requirements.txt" is the best choice under the constraints that we have.

Would singular "version_requirement.txt" work?

kwankyu:
Perhaps you want to add #37401 as a dependency of this PR.

You may want to rebase.

mkoeppe:
> Would singular "version_requirement.txt" work? I hope so.

I don't think so, but I can't be sure, this is reverse-engineered only (https://github.com/sagemath/sage/pull/36982#issuecomment-1874857075)

kwankyu:
I may experiment with my fork.

kwankyu:
> > Would singular "version_requirement.txt" work? I hope so.
> 
> I don't think so, but I can't be sure, this is reverse-engineered only ([#36982 (comment)](https://github.com/sagemath/sage/pull/36982#issuecomment-1874857075))

Right, according to https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems

mkoeppe:
> Perhaps you may want to add #37401 as a dependency of this PR.

Thanks, I've listed it as a dependency for now.

kwankyu:
> I may experiment with my fork.

`version_requirement.txt` does not work.

`version_requirements.txt` works well.

OK. Let's go with `version_requirements.txt`.

dimpase:
I still don't get why we cannot just use `requirements.txt`.

dimpase:
abstract vs concrete dependencies argument does not fly, as at present install-requires.txt may have versions too.

mkoeppe:
@dimpase Let's do this step by step. https://github.com/sagemath/sage/pull/37401 - does it explain everything that you need to know? Then complete the review please.

kwankyu (approved):
I agree with the change. LGTM.

dimpase (changes requested):
just name them requirements.txt

mkoeppe:
> I still don't get why we cannot just use `requirements.txt`.

@dimpase I think you should at least tell us what you have done since you wrote  "I still don't get why we cannot just use requirements.txt." the previous time (https://github.com/sagemath/sage/pull/36999#issuecomment-1959070326 where). 

Let's take this back to https://github.com/sagemath/sage/pull/37401. Obviously more explanation is needed there.

dimpase:
you have not explained the reason there. 

besides it's just extra confusion stemming from different names for files with the same format and semantics.

it would also shorten the docs, one wouldn't have to describe the same thing twice.

mkoeppe:
> besides it's just extra confusion stemming from different names for files with the same format and semantics.
> 
> it would also shorten the docs, one wouldn't have to describe the same thing twice.

@dimpase This really has no value here. You really need to do your reading before it makes to discuss anything.

dimpase:
no, it's your "explanations" - which are concealing facts - which are useless and harmful.

mkoeppe:
<blocked>

roed314:
This PR was approved by @kwankyu.  If the vote [here](https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ) succeeds, those involved may act in accordance with the new policy.  In the meantime I'm setting this to positive review.  @dimpase, your remarks here are not appropriate.

Given the status of CI and the amount of time that this ticket has been stalled, updates may need to be made.  I ask those involved with this ticket to take reasonable actions in preparation for it getting merged, and @vbraun to give them sufficient time to do so.

mkoeppe:
No, this PR is not "pending".

jhpalmieri:
Dima, please do not change the labels on this PR.

kiwifb:
It seems to me the dispute is really about the naming. In any case we need to move from the current name +1 for me.

kwankyu:
Still +1 from me. `version_requirement.txt` is to replace `install-requires.txt`, which is different from `requirement.txt`.

NathanDunfield (approved):
Looks good to me.  I agree @kwankyu in https://github.com/sagemath/sage/pull/36999/#issuecomment-1960762895 that `version_requirements.txt` is the best choice of name here.

NathanDunfield:
Per the comment by @roed314, I have set to positive review counting:

+1: @mkoeppe, @kwankyu, @kiwifb, and myself
-1: @tobiasdiez, @dimpase

vbraun:
Please try building from a clean checkout, I'm getting lots of `sage-spkg-info: unknown package xz` for all packages

PR #37004, opened 2024-01-03
Fixes in save_session()

### LINKED ISSUE #37002: save_session() is broken when interactive
I've never used this before, so I don't know if it used to work, but:
```
$ sage
│ Using Python 3.12.0. Type "help()" for help.                       │
sage: save_session("misc")
<ipython-input-1-3181853d66d8>:1: DeprecationWarning: 
Importing Euler_Phi from here is deprecated; please use "from sage.arith.misc import Euler_Phi" instead.
See https://github.com/sagemath/sage/issues/30322 for details.
  save_session("misc")
Cell In[1], line 1
----> 1 save_session("misc")

File /usr/lib/python3.12/site-packages/sage/misc/session.pyx:313, in sage.misc.session.save_session (build/cythonized/sage/misc/session.c:4112)()
    311 # We iterate only over the new variables that were defined in this
    312 # session, since those are the only ones we will save.
--> 313 for k in show_identifiers(hidden = True):
    314     try:
    315         x = state[k]

File /usr/lib/python3.12/site-packages/sage/misc/session.pyx:232, in sage.misc.session.show_identifiers (build/cythonized/sage/misc/session.c:3769)()
    230     return val == DocTestTask.extra_globals.get(name, _none)
    231 
--> 232 return sorted([x for x, v in state.items() if _is_new_var(x, v, hidden)
    233                and not _in_extra_globals(x, v)])
    234 

RuntimeError: dictionary changed size during iteration
```

The curious thing is that when doing this as part of a doctest I don't get this failure.

btw, why should I get a deprecation warning for a variable I did not mean to use?

### LINKED ISSUE #37003: save_session() broken when cython changes
I can't test this interactively due to #37002, but I found the following doctest failure:
```
File "src/sage/misc/session.pyx", line 303, in sage.misc.session.save_session
Failed example:
    save_session(tmp_f, verbose=True)
Expected:
    Saving...
    Not saving g: g is a function, method, class or type
    ...
Got:
    Saving __cached__
    Saving __doc__
    Saving __file__
    Saving __loader__
    Saving __name__
    Saving __spec__
    Saving a
    Not saving checkbox: checkbox is a function, method, class or type
    Not saving color_selector: color_selector is a function, method, class or type
    Not saving f: f is a function, method, class or type
    Saving g
    Not saving input_box: input_box is a function, method, class or type
    Not saving input_grid: input_grid is a function, method, class or type
    Saving interact
    Not saving range_slider: range_slider is a function, method, class or type
    Not saving selector: selector is a function, method, class or type
    Not saving slider: slider is a function, method, class or type
    Not saving text_control: text_control is a function, method, class or type
    Saving tmp_f
```
The issue here is that `g` is a cython function and according to the doctest should *not* be saved, but it is.

Further investigation reveals the following:
```
$ sage
│ Using Python 3.12.1. Type "help()" for help.                       │
sage: g = cython_lambda('double x', 'x*x + 1.5')
sage: type(g)
<class '_cython_3_0_7.cython_function_or_method'>
sage: sage.misc.session.CythonFunctionType
<class '_cython_3_0_6.cython_function_or_method'>
sage: type(g) == sage.misc.session.CythonFunctionType
False
```

IOW, sagemath was built with cython 3.0.6; later cython was updated to 3.0.7.

Maybe this can be fixed by changing, in `src/sage/misc/session.pyx`, the line
```python
CythonFunctionType = type(lambda: None)
```
by something like
```python
CythonFunctionType = type(cython_lambda([], None))
```

Maybe better, use a test as in `src/sage/misc/abstract_method.py`, namely:
```
getattr(type(f), '__name__', None) == 'cython_function_or_method')
```

tornaria:
1. fix test for cython functions in save_session()
   This fixes #37003 
2. fix handling of lazy import variables in save_session()
   This fixes #37002 
3. remove unused `extra_globals` attribute in class DocTestTask
   This was introduced for the transition from python 2 to 3, no longer used
4. don't sort the output of `sage.misc.session.show_identifiers()`
   There's no need to sort here... sorting was introduced in baedc413164 in order to fix doctest errors, but dictionaries now keep the insertion order. Thus `show_identifiers()` will return the variables in the order they were introduced for the first time which seems reasonable and it's fine to test.

### :memo: Checklist

Fixes: #37002, #37003

### DIFF
--- a/src/sage/doctest/forker.py
+++ b/src/sage/doctest/forker.py
@@ -2477,19 +2477,6 @@ class DocTestTask():
         ['cputime', 'err', 'failures', 'optionals', 'tests', 'walltime', 'walltime_skips']
     """
 
-    extra_globals = {}
-    """
-    Extra objects to place in the global namespace in which tests are run.
-    Normally this should be empty but there are special cases where it may
-    be useful.
-
-    For example, in Sage versions 9.1 and earlier, on Python 3 add
-    ``long`` as an alias for ``int`` so that tests that use the
-    ``long`` built-in (of which there are many) still pass.  We did
-    this so that the test suite could run on Python 3 while Python 2
-    was still the default.
-    """
-
     def __init__(self, source):
         """
         Initialization.
@@ -2614,10 +2601,6 @@ class DocTestTask():
         # Remove '__package__' item from the globals since it is not
         # always in the globals in an actual Sage session.
         dict_all.pop('__package__', None)
-
-        # Add any other special globals for testing purposes only
-        dict_all.update(self.extra_globals)
-
         sage_namespace = RecordingDict(dict_all)
         sage_namespace['__name__'] = '__main__'
         doctests, extras = self.source.create_doctests(sage_namespace)
--- a/src/sage/misc/session.pyx
+++ b/src/sage/misc/session.pyx
@@ -80,7 +80,6 @@ from sage.misc.persist import load, save, loads, dumps
 
 state_at_init = None
 
-CythonFunctionType = type(lambda: None)
 
 def init(state=None):
     """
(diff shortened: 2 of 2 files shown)


fchapoton (approved):
looks good

PR #37007, opened 2024-01-04
normalize projective points over (quotients of) Euclidean domains

yyyyx4:
Currently, projective points *over fields* are normalized to always have $1$ as their last non-zero coordinate, but projective points over non-fields are not normalized at all.

In this patch, we add similar normalization functionality for base rings which are (1) Euclidean domains, or (2) quotients of Euclidean domains. The most important examples are probably $\mathbb Z$ and $\mathbb Z/n$ for composite $n$.

### DIFF
--- a/src/sage/dynamics/arithmetic_dynamics/projective_ds.py
+++ b/src/sage/dynamics/arithmetic_dynamics/projective_ds.py
@@ -1081,8 +1081,7 @@ class DynamicalSystem_projective(SchemeMorphism_polynomial_projective_space,
             sage: f.nth_iterate(P(0, 1), 3)
             Traceback (most recent call last):
             ...
-            ValueError: [0, 0] does not define a point in Projective Space of
-            dimension 1 over Rational Field since all entries are zero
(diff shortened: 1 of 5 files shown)


yyyyx4:
The current version of this patch seems to cause some performance regressions with elliptic-curve points over finite fields.

yyyyx4:
> The current version of this patch seems to cause some performance regressions with elliptic-curve points over finite fields.

Fixed in the new version by lifting the faster version of coordinate normalization over fields from `EllipticCurvePoint_field` to `SchemeMorphism_point_projective_field`.

GiacomoPope (approved):
If the performance degradation mentioned above is totally gone, then this seems to only improve the code. I have not benchmarked these changes myself but trust Lorenz to have checked thoroughly. 

Call to `SchemeMorphism_point_abelian_variety_field` to simplify initialisation of points is good code clean up and the error handling and normalisation in this patch are more natural than what we had previously. 

Positive review. 

NOTE: failing test is orthogonal to this patch.

PR #37008, opened 2024-01-04
Add `sage-generate-meson` script to sage-setup

mkoeppe:
The script was taken from:
- #36524 

... where it was called `generate-meson.py` and intended to be removed before merging. https://github.com/sagemath/sage/pull/36524#discussion_r1436164787

Here we make it available as a script installed by `sage_setup` because what was useful once for generating code will be useful later for updating code, in particular as will be necessary for modularization concerns, as discussed in https://github.com/sagemath/sage/pull/36524#issuecomment-1873442413

Various improvements have been made. When invoked with the switch `--monolithic`, the output maintains compatibility with #36524.

Author: @tobiasdiez, @mkoeppe 

### :memo: Checklist


tobiasdiez:
As it has been explained before, the point of this script was to create the initial boilerplate code. I see value in creating a script that adds missing python and cython files. But that would require to parse the existing meson files instead of overwriting them. Something similar is already supported for other targets, just not for python. https://mesonbuild.com/Rewriter.html So I am sure the meson team would be happy to accept your PR.

mkoeppe:
> that would require

It wouldn't. Even just overwriting + then editing often has value.

orlitzky:
With the `.in` files, does this now avoid overwriting handwritten `meson.build`? @tobiasdiez should really be the one to OK this in any case.

mkoeppe:
> With the `.in` files, does this now avoid overwriting handwritten `meson.build`? 

Yes, the idea is to separate handwritten bits from autogenerated bits.

kiwifb:
I think we now have pressing needs to move to meson. This stuff looks OK, but it cannot be used in isolation. Can it even be tested without other bits for sagemath-standard?

mkoeppe:
> Can it even be tested without other bits for sagemath-standard?

Well, it generates files that can be compared with what has been committed to #36524.

orlitzky:
Having now accrued a years' worth of new options, bug fixes, and random whitespace changes, the original generation script is less useful than ever. The meson port is "done" and further changes can be made incrementally.

PR #37009, opened 2024-01-04
`src/bin/sage-cython`: Repurpose as PEP 420 fixer

mkoeppe:
We ship it with the distribution **sagemath-environment**.

This is needed for building the modularized distributions of the Sage library using meson-python.

With meson >= 1.4.0 (https://github.com/mesonbuild/meson/pull/12674), it then suffices to set `CYTHON=sage-cython`.

With earlier versions of meson, one can use a native file 
```
[binaries]
cython = 'sage-cython'
```

... as done in #37012 (54c95d60183) and #35095.

### :memo: Checklist

### DIFF
--- a/src/bin/sage-cython
+++ b/src/bin/sage-cython
@@ -1,39 +1,26 @@
-#!/usr/bin/env sage-python
+#!/usr/bin/env python
 
-# This script is a deprecated wrapper around the "cython" program.
-# It is deprecated since Issue #27041 (Sage 8.7) because one should
+# This script is a wrapper around the "cython" program.
+# It was deprecated in Issue #27041 (Sage 8.7) because one should
 # simply use "cython" directly. We display deprecation messages whenever
 # "sage-cython" does something different from plain "cython".
 #
 # A stronger deprecation warning was added in #29923 (Sage 9.2).
+#
+# In Sage 10.3, the deprecated functionality is removed permanently.
+# However, the script is un-deprecated and re-purposed as a temporary workaround
+# for defects in Cython 3.0.x support for PEP 420 implicit namespace packages,
+# see https://github.com/sagemath/sage/pull/36228
+#
+# This script can be used as a replacement for "cython".
+# For example, to have meson use it, set the environment variable
+# CYTHON to sage-cython./
+# https://github.com/mesonbuild/meson/blob/e4bbc630b67ef97ad842badd00855e64cff12e13/mesonbuild/envconfig.py#L91
 
-import os
-import sys
-args = sys.argv[1:]
+from sage.misc.package_dir import cython_namespace_package_support
 
-sys.stderr.write("WARNING: the script sage-cython is deprecated; use cython instead.\n")
+# console scripts declared in https://github.com/cython/cython/blob/master/setup.py#L68
+from Cython.Compiler.Main import setuptools_main
 
-from sage.env import SAGE_SRC
-
-# args can have length 0, in case we're printing a usage message (see trac 12207)
-pyx_file = os.path.abspath(args[-1]) if len(args) else None
-include_sage_flags = False
-
-if pyx_file and pyx_file.startswith(SAGE_SRC):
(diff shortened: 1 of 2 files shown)


src/bin/sage-cython
```diff
@@ -1,39 +1,26 @@
...
+# It was deprecated in Issue #27041 (Sage 8.7) because one should
 # simply use "cython" directly. We display deprecation messages whenever
 # "sage-cython" does something different from plain "cython".
 #
 # A stronger deprecation warning was added in #29923 (Sage 9.2).
+#
+# In Sage 10.3, the deprecated functionality is removed permanently.
+# However, the script is un-deprecated and re-purposed as a temporary workaround
+# for defects in Cython 3.0.x support for PEP 420 implicit namespace packages,
+# see https://github.com/sagemath/sage/pull/36228
```
tornaria, line 13: Please don't un-deprecate this. Instead it should be removed in favor of using an unpatched version of cython.

Right now, compiling cython files that use sage modules is broken because of namespace packages, and patching cython here and in sage_setup gives the misleading impression that it works.

This is the opposite of "fitting in the python/cython ecosystem".

mkoeppe: Note that it is a "temporary workaround".

Anyone report the namespace problem to Cython upstream yet?

tornaria: Temporary hack. And when/if cython fixes it we still have to ship the hack since distros are slow updating cython, etc.

Let's stick to stable features, not the bleeding edge.

mkoeppe: @tornaria Kindly limit the use of fighting words such as "hack".

> Let's stick to stable features, not the bleeding edge.

We have been successfully using Cython with namespace packages ever since 2021.

If anything, it is that our porting effort to Cython 3 is incomplete. 
- https://github.com/sagemath/sage/pull/36228 tracks a remaining task, that's all.

mkoeppe:
@tornaria Is there a dispute here in the sense of the "disputed PRs" process? https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/MV6LXkqRCgAJ

jhpalmieri:

orlitzky:
It's hard to coherently vote on all of these conflicting proposals at once. I lean towards re-adding the init files in the short term, but if the consensus is that we shouldn't do it, then I don't object to this. On the other hand, if I vote to approve this AND to re-add the init files, then we may wind up with two approved but conflicting branches.

mkoeppe:
This PR here only adds a useful command and does not conflict with anything.

mkoeppe:
This PR is prepared exactly how PRs should be prepared. It is not a test of ideological consistency.

It simply a useful tool for the build of where there are no `__init__.py` files -- even if the "re-add init files" PRs is merged. I'm using it in #35095 to build the modularized distributions using meson-python.

orlitzky:
> It simply a useful tool for the build of where there are no `__init__.py` files -- even if the "re-add init files" PRs is merged. I'm using it in #35095 to build the modularized distributions using meson-python.

OK that's helpful to know.

mkoeppe:
@tornaria Please check https://github.com/sagemath/sage/pull/37009#issuecomment-1996635272

tornaria:
> @tornaria Please check [#37009 (comment)](https://github.com/sagemath/sage/pull/37009#issuecomment-1996635272)

-1 vote from me here.

mkoeppe:
@tornaria Would you be able to express in technical terms what your concern about this PR is, if any? 

I am using it in a follow-up for building the modularized distributions with meson-python; it has no effect on building the monolithic Sage library.

tornaria:
https://github.com/sagemath/sage/pull/37009#discussion_r1494643682

mkoeppe:
There's nothing there, Gonzalo, just declarations.

mkoeppe:
@roed314 @saraedum I think the "disputed" tag should be removed; a review has not even been started.

orlitzky:
Seeing as the __init__.py files were replaced and we are no longer producing modular sagemath- distributions, there is no longer any use for this.

PR #37017, opened 2024-01-06
cosmetic little change in topology/ (some ruff UP and PERF)

fchapoton:
a set of small fixes there

### :memo: Checklist

### DIFF
--- a/src/sage/topology/cubical_complex.py
+++ b/src/sage/topology/cubical_complex.py
@@ -545,13 +545,12 @@ class Cube(SageObject):
         if self.dimension() < 0:  # the empty cube
             return [Simplex(())]  # the empty simplex
         v = tuple([max(j) for j in self.tuple()])
+        Sv = Simplex((v,))
         if self.dimension() == 0:  # just v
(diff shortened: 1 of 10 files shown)


src/sage/topology/cubical_complex.py
```diff
@@ -719,7 +718,7 @@ def _repr_(self):
             sage: C1._repr_()
             '[1,1] x [2,3] x [4,5]'
         """
-        s = ("[%s,%s]" % (str(x), str(y)) for x, y in self.__tuple)
+        s = ("[{},{}]".format(str(x), str(y)) for x, y in self.__tuple)
```
tobiasdiez, line 721: Any reason to not use f-strings here? Ruff should be able to do this automatically (UP032).

fchapoton: I have deliberately skipped UP032 for the moment.

tobiasdiez: Okay, it's already an improvement anyway.

tobiasdiez (approved):

PR #37029, opened 2024-01-07
Add devcontainer configs for sagemath/sagemath and sagemathinc/sagemath-core, add Docker section to installation manual

mkoeppe:
In https://groups.google.com/g/sage-announce/c/z83jqg9oQJE/m/4iePjwNvAQAJ, @williamstein announced the availability of  https://hub.docker.com/r/sagemathinc/sagemath-core
Here we add a devcontainer configuration for it, based on the existing configuration for cocalc.

Likewise we add devcontainer configurations for the project's official Docker images (sagemath/sagemath:latest and :develop). We also update the README regarding the available tags.

New Docker section in the installation guide:
- https://deploy-preview-37029--sagemath.netlify.app/html/en/installation/#in-a-container

### :memo: Checklist

### DIFF
--- a/.devcontainer/downstream-docker-sagemath-sagemath-develop/devcontainer.json
+++ b/.devcontainer/downstream-docker-sagemath-sagemath-develop/devcontainer.json
@@ -1,0 +1,18 @@
+// See https://aka.ms/devcontainer.json for format details.
+{
+    "name": "sagemath/sagemath:develop Docker",
+    "image": "sagemath/sagemath:develop",
+    "containerEnv": {
+        "MAKE": "make -j4"
+    },
+    "onCreateCommand": ".devcontainer/onCreate.sh",
+    // * If the workspace directory looks like a copy of the Sage source tree (SAGE_ROOT):
+    //   - it bootstraps the Sage distribution,
+    //   - sets the symlink ``venv`` as expected,
+    // * Otherwise, it does nothing. This is so that users can copy this devcontainer.json file as is
+    //   into their projects.
+    "updateContentCommand": "if [ -d pkgs/sagemath-standard ]; then make configure && ./configure --prefix=/home/sage/sage/local --with-sage-venv; else echo 'Edit .devcontainer/devcontainer.json (updateContentCommand) to run project-specific startup commands'; fi",
+    "extensions": [
+        "ms-python.python"
+    ]
+}
--- a/.devcontainer/downstream-docker-sagemathinc-sagemath-core/devcontainer.json
+++ b/.devcontainer/downstream-docker-sagemathinc-sagemath-core/devcontainer.json
@@ -1,0 +1,19 @@
+// See https://aka.ms/devcontainer.json for format details.
+{
+    "name": "sagemathinc/sagemath-core Docker",
+    "image": "sagemathinc/sagemath-core",
+    "containerEnv": {
+        "MAKE": "make -j4"
+    },
+    // libgmp.a is broken and leads to a build failure of ecm.
+    "onCreateCommand": ".devcontainer/onCreate.sh && rm -f /usr/local/sage/local/lib/libgmp.a",
+    // * If the workspace directory looks like a copy of the Sage source tree (SAGE_ROOT):
+    //   - it bootstraps the Sage distribution,
+    //   - sets the symlink ``venv`` as expected,
+    // * Otherwise, it does nothing. This is so that users can copy this devcontainer.json file as is
+    //   into their projects.
+    "updateContentCommand": "if [ -d pkgs/sagemath-standard ]; then make configure && ./configure --enable-build-as-root --prefix=/usr/local/sage/local --with-sage-venv; else echo 'Edit .devcontainer/devcontainer.json (updateContentCommand) to run project-specific startup commands'; fi",
+    "extensions": [
+        "ms-python.python"
+    ]
+}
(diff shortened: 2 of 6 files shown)


williamstein:
Are there any specific assumptions here about my docker container that I should be aware of and document?  Thanks.

.devcontainer/downstream-docker-sagemathinc-sagemath-core/devcontainer.json
```diff
@@ -0,0 +1,19 @@
...
+        "MAKE": "make -j4"
+    },
+    // libgmp.a is broken and leads to a build failure of ecm.
+    "onCreateCommand": ".devcontainer/onCreate.sh && rm -f /usr/local/sage/local/lib/libgmp.a",
+    // * If the workspace directory looks like a copy of the Sage source tree (SAGE_ROOT):
+    //   - it bootstraps the Sage distribution,
+    //   - sets the symlink ``venv`` as expected,
+    // * Otherwise, it does nothing. This is so that users can copy this devcontainer.json file as is
+    //   into their projects.
+    "updateContentCommand": "if [ -d pkgs/sagemath-standard ]; then make configure && ./configure --enable-build-as-root --prefix=/usr/local/sage/local --with-sage-venv; else echo 'Edit .devcontainer/devcontainer.json (updateContentCommand) to run project-specific startup commands'; fi",
```
mkoeppe, line 15: @williamstein This line is where an assumption is made: That there is a SAGE_ROOT in `/usr/local/sage/` without customization of `--prefix` or `--with-sage-venv`.

williamstein:
Thanks.  I've added a README here to at least reduce the chances we break things upstream:  https://github.com/sagemathinc/cocalc-compute-docker/tree/main/src/sagemath/core

soehms:
I already asked a similar question in https://github.com/sagemath/sage/issues/37304 (see https://github.com/sagemath/sage/issues/37304#issuecomment-1948997886):

Is there anything wrong with pointing to our CI images on `ghcr.io` as another option

This would give another chance to users who fail with some technical details in any of the Docker images (as in the case of the issue above).

soehms:
Doesn't it make sense to add the new devcontainers `downstream-docker-sagemath-sagemath` and `downstream-docker-sagemath-sagemath-develop` to `portability_testing.rst`, as well?

I've tried to test the new `devcontainer.json` files on Windows. But you should know that I don't have much experiences with VS Code (nor other modern IDE's) since I still prefere to work in a bash terminal editing with `vi` where I can always scroll up to see what I did before in correct chronolgy. Maybe that's because my brain is like a sieve and the gaps in it get bigger as I get older. Thus, I don't like that the IDE's constantly erase my traces (and if they are not erased they are diffucult to find).

Anyway, after long and hard experimentation following the instructions I found in #37534 I was able to obtain a working container in a local Windows folder with the new `devcontainer.json`-files. But in all cases I had to replace the `onCreateCommand` (I replaced it in all cases by `sage` which opened sage imediately after creating the terminal in the container). Without that change I got errors not only for the new `devcontainer` files but also in the `computop` case. I think the error is caused by a subfolder of a Samba-drive (`V:`) which occurs in the path variable of the system:

```
[1138 ms] <3>WSL (841) ERROR: UtilTranslatePathList:2866: Failed to translate V:\Environ
```

BTW: On the Windows systems to which I have access there is [Snagit](https://en.wikipedia.org/wiki/Snagit) installed (due to the fact that the snipping-tool that comes with Windows isn't really useful). Many of the VS Code keyboard shortcuts are overlaid by Snagit functionality, for example `Ctrl+Shift+P` or `Ctrl+Shift+X`. It took me some time to find out that the *command palette* is just the input box in the top middle where the input has to start with `>`. Thus, I would suggest to explain this more explicitly in our documentation concering VS Code.

kwankyu:
`sagemath/sagemath:develop` devcontainer does not work for me. I get
```
This codespace is currently running in recovery mode due to a container error.
1. Use Cmd/Ctrl + Shift + P -> "Codespaces: View Creation Log" to see full logs
2. Update your devcontainer configuration as needed
3. Use Cmd/Ctrl + Shift + P -> "Codespaces: Rebuild Container" to retry
4. For help, read more about custom configuration: https://aka.ms/ghcs-custom-configuration
mkdir: cannot create directory ‘/home/sage’: Permission denied
codespaces-87b8a4:/workspaces/sage$ touch: cannot touch '/home/sage/.config/vscode-dev-containers/first-run-notice-already-displayed': No such file or directory
```
I tried it in codespaces.

.devcontainer/downstream-docker-sagemath-sagemath-develop/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+        "MAKE": "make -j4"
+    },
+    "onCreateCommand": ".devcontainer/onCreate.sh",
+    // * If the workspace directory looks like a copy of the Sage source tree (SAGE_ROOT):
+    //   - it bootstraps the Sage distribution,
+    //   - sets the symlink ``venv`` as expected,
+    // * Otherwise, it does nothing. This is so that users can copy this devcontainer.json file as is
+    //   into their projects.
+    "updateContentCommand": "if [ -d pkgs/sagemath-standard ]; then make configure && ./configure --prefix=/home/sage/sage/local --with-sage-venv; else echo 'Edit .devcontainer/devcontainer.json (updateContentCommand) to run project-specific startup commands'; fi",
+    "extensions": [
```
kwankyu, line 15: `--prefix=/usr/local/sage/local` ?

PR #36073, opened 2023-08-11
Stop sorting Graph vertices and edges by default

dcoudert:
Following #22349 and #27408, we finally set parameter `sort` to `False` by default for methods `vertices` and `edges`, and remove the corresponding deprecation warnings.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

PR #37035, opened 2024-01-08
Return a vector of monomial instead of a matrix in PolynomialSequence.coefficient_matrix()

### LINKED ISSUE #37027: `PolynomialSequence.coefficient_matrix()` returns a matrix instead of a vector as second element
### Steps To Reproduce

```
sage: from sage.rings.polynomial.multi_polynomial_sequence import PolynomialSequence
sage: R.<x,y> = QQ[]
sage: PolynomialSequence([x+2*y+3, 4*x, 5]).coefficient_matrix()
(
[1 2 3]  [x]
[4 0 0]  [y]
[0 0 5], [1]
)
sage: _[1].parent()
Full MatrixSpace of 3 by 1 dense matrices over Multivariate Polynomial Ring in x, y over Rational Field
```

### Expected Behavior

The second element should be a vector - it is the vector of monomials appearing in the sequence.

### Actual Behavior

It returns a 1-column matrix.

### Additional Information

The problem is that iterating over a vector gives the elements, whereas iterating over a 1-column matrix gives vectors.

### Environment

```markdown
irrelevant.
```

### Checklist

RuchitJagodara:
- Final change to the function
- Added a deprecation message
This patch fixes #37027 by returning a vector of monomials instead of a matrix,
which follows the documentation.
I have also added a deprecation message.

### :memo: Checklist

### :hourglass: Dependencies

Does not matter.

### DIFF
--- a/src/sage/groups/matrix_gps/finitely_generated_gap.py
+++ b/src/sage/groups/matrix_gps/finitely_generated_gap.py
@@ -936,11 +936,11 @@ def _new_invariant_is_linearly_independent(F, invariants):
         sage: gens = [matrix(QQ, [[-1,1],[-1,0]]), matrix(QQ, [[0,1],[1,0]])]
         sage: G = MatrixGroup(gens)
         sage: s = Sequence(G.invariants_of_degree(14))                                  # needs sage.rings.number_field
-        sage: s.coefficient_matrix()[0].rank()                                          # needs sage.rings.number_field
+        sage: s.coefficients_monomials()[0].rank()                                      # needs sage.rings.number_field
         3
         sage: len(s)                                                                    # needs sage.rings.number_field
         3
     """
     if len(invariants) == 0:
         return True
-    return PolynomialSequence(invariants).coefficient_matrix()[0].rank() != PolynomialSequence(list(invariants)+[F]).coefficient_matrix()[0].rank()
+    return PolynomialSequence(invariants).coefficients_monomials()[0].rank() != PolynomialSequence(list(invariants)+[F]).coefficients_monomials()[0].rank()
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -707,6 +707,74 @@ class PolynomialSequence_generic(Sequence_generic):
         RRR = PolynomialRing(K,r,tuple(Ts),order='negdeglex')
         return RRR.ideal(JJ.gens())
 
+    def coefficients_monomials(self, order=None, sparse=True):
+        """
+        Return the matrix of coefficients ``A`` and
+        the matching vector of monomials ``v``, such that ``A*v == vector(self)``.
+
+        Thus value of ``A[i,j]`` corresponds the coefficient of the
+        monomial ``v[j]`` in the ``i``-th polynomial in this system.
+
+        Monomials are ordered w.r.t. the term ordering of ``order``
+        if given; otherwise, they are ordered w.r.t. ``self.ring()``
+        in reverse order, i.e., such that the smallest entry comes last.
+
+        INPUT:
+
+        - ``sparse`` - construct a sparse matrix (default: ``True``)
+        - ``order`` - a list or tuple specifying the order of monomials (default: ``None``)
+
+        EXAMPLES::
+
+            sage: # needs sage.libs.singular
+            sage: P.<a,b,c,d> = PolynomialRing(GF(127), 4)
+            sage: I = sage.rings.ideal.Katsura(P)
+            sage: I.gens()
+            [a + 2*b + 2*c + 2*d - 1,
+             a^2 + 2*b^2 + 2*c^2 + 2*d^2 - a,
+             2*a*b + 2*b*c + 2*c*d - b,
+             b^2 + 2*a*c + 2*b*d - c]
+            sage: F = Sequence(I)
+            sage: A,v = F.coefficients_monomials()
+            sage: A
+            [  0   0   0   0   0   0   0   0   0   1   2   2   2 126]
+            [  1   0   2   0   0   2   0   0   2 126   0   0   0   0]
+            [  0   2   0   0   2   0   0   2   0   0 126   0   0   0]
+            [  0   0   1   2   0   0   2   0   0   0   0 126   0   0]
+            sage: v
+            (a^2, a*b, b^2, a*c, b*c, c^2, b*d, c*d, d^2, a, b, c, d, 1)
+            sage: A*v
+            (a + 2*b + 2*c + 2*d - 1, a^2 + 2*b^2 + 2*c^2 + 2*d^2 - a,
+             2*a*b + 2*b*c + 2*c*d - b, b^2 + 2*a*c + 2*b*d - c)
+        """
+        from sage.modules.free_module_element import vector
+        from sage.matrix.constructor import Matrix
+
+        R = self.ring()
+        f = tuple(self)
+
+        if order is None:
+            v = sorted(self.monomials(), reverse=True)
+        else:
+            if isinstance(order, (list, tuple)):
+                v = order
+            else:
+                raise ValueError("order argument can only accept list or tuple")
+        nv = len(v)
+
+        # construct dictionary for fast lookups
+        y = dict(zip(v, range(nv)))
+
+        A = Matrix(R.base_ring(), len(f), nv, sparse=sparse)
+
+        for x, poly in enumerate(f):
+            for m in poly.monomials():
+                try:
+                    A[x, y[m]] = poly.monomial_coefficient(m)
+                except KeyError:
+                    raise ValueError("order argument does not contain all monomials")
+        return A, vector(v)
+
     def coefficient_matrix(self, sparse=True):
         """
         Return tuple ``(A,v)`` where ``A`` is the coefficient matrix
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -768,9 +768,9 @@ class PolynomialSequence_generic(Sequence_generic):
         A = Matrix(R.base_ring(), len(f), nv, sparse=sparse)
 
         for x, poly in enumerate(f):
-            for m in poly.monomials():
+            for c, m in poly:
                 try:
-                    A[x, y[m]] = poly.monomial_coefficient(m)
+                    A[x, y[m]] = c
                 except KeyError:
                     raise ValueError("order argument does not contain all monomials")
         return A, vector(v)
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -141,7 +141,7 @@ easily::
     sage: A.rank()
     4056
     sage: A[4055] * v
-    (k001*k003)
+    k001*k003
 
 TESTS::
 
@@ -803,6 +803,9 @@ class PolynomialSequence_generic(Sequence_generic):
              b^2 + 2*a*c + 2*b*d - c]
             sage: F = Sequence(I)
             sage: A,v = F.coefficient_matrix()
+            DeprecationWarning: the function coefficient_matrix is deprecated; use coefficients_monomials instead
+            See https://github.com/sagemath/sage/issues/37035 for details.
+              A,v = F.coefficient_matrix()
             sage: A
             [  0   0   0   0   0   0   0   0   0   1   2   2   2 126]
             [  1   0   2   0   0   2   0   0   2 126   0   0   0   0]
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -803,9 +803,9 @@ class PolynomialSequence_generic(Sequence_generic):
              b^2 + 2*a*c + 2*b*d - c]
             sage: F = Sequence(I)
             sage: A,v = F.coefficient_matrix()
+            doctest:warning...
             DeprecationWarning: the function coefficient_matrix is deprecated; use coefficients_monomials instead
             See https://github.com/sagemath/sage/issues/37035 for details.
-              A,v = F.coefficient_matrix()
             sage: A
             [  0   0   0   0   0   0   0   0   0   1   2   2   2 126]
             [  1   0   2   0   0   2   0   0   2 126   0   0   0   0]
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -114,7 +114,7 @@ We separate the system in independent subsystems::
 
 and compute the coefficient matrix::
 
-    sage: A,v = Sequence(r2).coefficient_matrix()                                       # needs sage.rings.polynomial.pbori
+    sage: A,v = Sequence(r2).coefficients_monomials()                                   # needs sage.rings.polynomial.pbori
     sage: A.rank()                                                                      # needs sage.rings.polynomial.pbori
     32
 
@@ -134,7 +134,7 @@ easily::
     sage: len(monomials)
     190
     sage: F2 = Sequence(map(mul, cartesian_product_iterator((monomials, F))))
-    sage: A, v = F2.coefficient_matrix(sparse=False)
+    sage: A, v = F2.coefficients_monomials(sparse=False)
     sage: A.echelonize()
     sage: A
     6840 x 4474 dense matrix over Finite Field of size 2...
@@ -829,26 +829,13 @@ class PolynomialSequence_generic(Sequence_generic):
             [      2*a*b + 2*b*c + 2*c*d - b]
             [        b^2 + 2*a*c + 2*b*d - c]
         """
-        R = self.ring()
-
-        m = sorted(self.monomials(),reverse=True)
-        nm = len(m)
-        f = tuple(self)
-        nf = len(f)
-
-        #construct dictionary for fast lookups
-        v = dict( zip( m , range(len(m)) ) )
-
         from sage.matrix.constructor import Matrix
+        from sage.misc.superseded import deprecation
+        deprecation(37035, "the function coefficient_matrix is deprecated; use coefficients_monomials instead")
 
-        A = Matrix( R.base_ring(), nf, nm, sparse=sparse )
-
-        for x in range( nf ):
-            poly = f[x]
-            for y in poly.monomials():
-                A[ x , v[y] ] = poly.monomial_coefficient(y)
-
-        return A, Matrix(R,nm,1,m)
+        R = self.ring()
+        A, v = self.coefficients_monomials(sparse=sparse)
+        return A, Matrix(R,len(v),1,v)
 
     def subs(self, *args, **kwargs):
         """
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -750,7 +750,6 @@ class PolynomialSequence_generic(Sequence_generic):
         from sage.modules.free_module_element import vector
         from sage.matrix.constructor import Matrix
 
-        f = tuple(self)
         if order is None:
             v = sorted(self.monomials(), reverse=True)
         else:
@@ -760,8 +759,8 @@ class PolynomialSequence_generic(Sequence_generic):
                 raise ValueError("order argument can only accept list or tuple")
 
         y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
-        A = Matrix(self.ring().base_ring(), len(f), len(v), sparse=sparse)
-        for x, poly in enumerate(f):
+        A = Matrix(self.ring().base_ring(), len(self), len(v), sparse=sparse)
+        for x, poly in enumerate(self):
             for c, m in poly:
                 try:
                     A[x, y[m]] = c
@@ -1694,6 +1693,60 @@ class PolynomialSequence_gf2(PolynomialSequence_generic):
             return PolynomialSequence(l, R, immutable=True)
         else:
             return PolynomialSequence_generic.reduced(self)
+
+    def coefficients_monomials(self, order=None, sparse=True):
+        """
+        Return the matrix of coefficients ``A`` and
+        the matching vector of monomials ``v``, such that ``A*v == vector(self)``.
+
+        Thus value of ``A[i,j]`` corresponds the coefficient of the
+        monomial ``v[j]`` in the ``i``-th polynomial in this system.
+
+        Monomials are ordered w.r.t. the term ordering of ``order``
+        if given; otherwise, they are ordered w.r.t. ``self.ring()``
+        in reverse order, i.e., such that the smallest entry comes last.
+
+        INPUT:
+
+        - ``sparse`` - construct a sparse matrix (default: ``True``)
+        - ``order`` - a list or tuple specifying the order of monomials (default: ``None``)
+
+        EXAMPLES::
+
+            sage: # needs sage.rings.polynomial.pbori
+            sage: B.<x,y,z> = BooleanPolynomialRing()
+            sage: F = Sequence([x*y + y + 1, z + 1])
+            sage: A, v = F.coefficients_monomials()
+            sage: A
+            [1 1 0 1]
+            [0 0 1 1]
+            sage: v
+            (x*y, y, z, 1)
+            sage: A*v
+            (x*y + y + 1, z + 1)
+        """
+        from sage.modules.free_module_element import vector
+        from sage.matrix.constructor import Matrix
+
+        if order is None:
+            v = sorted(self.monomials(), reverse=True)
+        else:
+            if isinstance(order, (list, tuple)):
+                v = order
+            else:
+                raise ValueError("order argument can only accept list or tuple")
+
+        R = self.ring().base_ring()
+        one = R.one()
+        y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
+        A = Matrix(R, len(self), len(v), sparse=sparse)
+        for x, poly in enumerate(self):
+            for m in poly:
+                try:
+                    A[x, y[m]] = one
+                except KeyError:
+                    raise ValueError("order argument does not contain all monomials")
+        return A, vector(v)
 
 
 class PolynomialSequence_gf2e(PolynomialSequence_generic):
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -750,9 +750,7 @@ class PolynomialSequence_generic(Sequence_generic):
         from sage.modules.free_module_element import vector
         from sage.matrix.constructor import Matrix
 
-        R = self.ring()
         f = tuple(self)
-
         if order is None:
             v = sorted(self.monomials(), reverse=True)
         else:
@@ -760,13 +758,9 @@ class PolynomialSequence_generic(Sequence_generic):
                 v = order
             else:
                 raise ValueError("order argument can only accept list or tuple")
-        nv = len(v)
 
-        # construct dictionary for fast lookups
-        y = dict(zip(v, range(nv)))
-
-        A = Matrix(R.base_ring(), len(f), nv, sparse=sparse)
-
+        y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
+        A = Matrix(self.ring().base_ring(), len(f), len(v), sparse=sparse)
         for x, poly in enumerate(f):
             for c, m in poly:
                 try:
--- a/src/sage/rings/polynomial/multi_polynomial_ideal.py
+++ b/src/sage/rings/polynomial/multi_polynomial_ideal.py
@@ -5333,7 +5333,7 @@ class MPolynomialIdeal(MPolynomialIdeal_singular_repr,
 
         2. We linearize and compute the echelon form::
 
-               sage: A, v = F.coefficient_matrix()
+               sage: A, v = F.coefficients_monomials()
                sage: A.echelonize()
 
         3. The result is the desired Gröbner basis::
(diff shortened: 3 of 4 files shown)


src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -771,16 +759,16 @@ def coefficient_matrix(self, sparse=True):
...
-
         A = Matrix( R.base_ring(), nf, nm, sparse=sparse )
 
         for x in range( nf ):
             poly = f[x]
             for y in poly.monomials():
                 A[ x , v[y] ] = poly.monomial_coefficient(y)
 
-        return A, Matrix(R,nm,1,m)
+        deprecation(37027,"The PolynomialSequence.coeffient_matrix() now returns a vector of monomials instead of a 3 by 1 matrix")
```
mantepse: You certainly do not want to hardcode the size of the matrix.

mantepse:
The `qqbar.py` and `expression.pyx` changes do not belong into this pull request.

I am not sure about the deprecation.  I think that it is required that old code works for about one year, and then the deprecated functionality is removed.

I don't know how to best do it, but one way is to introduce an optional keyword like `monomials_as_vector=False`, the deprecation is triggered only if we have `monomials_as_vector == False`, but in this case a matrix is returned (as before).  If, however, `monomials_as_vector == True`, then a vector is returned and no warning is triggered.

The downside is that both seeing the deprecation warning but also always having to provide the keyword is extremely annoying to users.  See #36073.  Thus, if we follow this model, we have to think hard about a good name for the keyword argument.

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: Ohh ! Ya, I think I have not pushed latest commits, let me do it. Actually, this is from my previous commit and I forgot to push latest one.

RuchitJagodara:
How about below implementation, I think we should not show deprecation message when user is using vector_form(I don't whether this name works or not but for now I am using this for reference) attribute have some value then we should not show deprecation message, so what if we make vector_form by default None, and do below changes ?

```python
        if vector_form == True:
            return A, vector(m)
        else:
            if vector_form == None:
                deprecation(37027,'the function PolynomialSequence.coeffient_matrix() will'
                                  ' return a vector of monomilas instead of a matrix as second element in tuple in the future,'
                                  ' please use vector_form argument to get desired type.')
            return A, Matrix(Matrix(R,nm,1,m))

```

But still I am not sure whether this is a good idea or not?

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -766,9 +798,13 @@ def coefficient_matrix(self, sparse=True):
             for y in poly.monomials():
                 A[ x , v[y] ] = poly.monomial_coefficient(y)
 
-        deprecation(37027,"The PolynomialSequence.coeffient_matrix() now returns a vector of monomials instead of a matrix")
-
-        return A, vector(m)
+        if vector_form == True:
```
mantepse: `== True` can be removed

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -766,9 +798,13 @@ def coefficient_matrix(self, sparse=True):
...
+            return A, vector(m)
+        else:
```
mantepse: `else:` can be removed if you remove the indentation of the rest

mantepse:
To make the linter happy, you also need to remove trailing whitespace.

After that, I think we should ask on sage-devel whether this is the way to go.  One problem I see (and do not know how to solve) is that, eventually, we want to remove the `vector_form` keyword argument again.  It would be nice if we could do this in one deprecation.  Do you want to ask on sage-devel, or should I do it?

RuchitJagodara:
Ok, I'll ask. And thank you for your help !

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -771,16 +791,20 @@ def coefficient_matrix(self, sparse=True):
...
             for y in poly.monomials():
                 A[ x , v[y] ] = poly.monomial_coefficient(y)
 
-        return A, Matrix(R,nm,1,m)
+        if vector_form:
+            return A, vector(m)
+
+        deprecation(37027,'the function PolynomialSequence.coeffient_matrix() will'
+                          ' return a vector of monomilas instead of a matrix as second element in tuple in the future,'
+                          ' please use vector_form argument to get vector form of monomilas.')
```
mantepse: `monomilas` -> `monomials`
remove `in tuple`
I'd remove the third line (`, please ... `) completely, to make it shorter.

RuchitJagodara:
I have done all necessary changes, can you please review those once ?

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -745,66 +744,104 @@ def coefficient_matrix(self, sparse=True, vector_form=False):
...
         from sage.matrix.constructor import Matrix
 
         R = self.ring()
-
-        m = sorted(self.monomials(),reverse=True)
-        nm = len(m)
         f = tuple(self)
         nf = len(f)
 
+        if order==None:
```
mantepse, line 757: better: `order is None`

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,10 +707,81 @@ def algebraic_dependence(self):
         RRR = PolynomialRing(K,r,tuple(Ts),order='negdeglex')
         return RRR.ideal(JJ.gens())
 
+    def coefficients_monomials(self, order=None, sparse=True):
+        """
+        Return tuple ``(A,v)`` where ``A`` is the coefficient matrix
```
mantepse: Return *the* tuple...

In fact, maybe:

Return the matrix of coefficients ``A`` and the vector of monomials ``v``, such that ``A v == self``.

RuchitJagodara: I think it is returning a tuple only.

```python
sage: F.coefficients_monomials()
(
[  0   0   0   0   0   0   0   0   0   1   2   2   2 126]
[  1   0   2   0   0   2   0   0   2 126   0   0   0   0]
[  0   2   0   0   2   0   0   2   0   0 126   0   0   0]
[  0   0   1   2   0   0   2   0   0   0   0 126   0   0],

(a^2, a*b, b^2, a*c, b*c, c^2, b*d, c*d, d^2, a, b, c, d, 1)
)
sage: type(F.coefficients_monomials())
<class 'tuple'>
```
> Return the matrix of coefficients A and the vector of monomials v, such that $A * v == self$

And, I am not able to figure out how I can do this because A * v and self belong to different classes, so I think the == operator will return False (Although I am not sure about it, but I think operator overloading can be done for this, but I am not able to find a way to do so). Do you have any idea how this can be done?, e.g., 

```python
sage: (F.coefficient_matrix()[0]*F.coefficient_matrix()[1]).__class__
<class 'sage.matrix.matrix_mpolynomial_dense.Matrix_mpolynomial_dense'>
sage: F.__class__
<class 'sage.rings.polynomial.multi_polynomial_sequence.PolynomialSequence_generic'>
```

mantepse: Yes, to be more precise the sentence should probably read
```
Return the matrix of coefficients ``A`` and the vector of monomials ``v``, such that ``A*v == vector(self)``.
```
However, as I mathematician, I'd ignore the difference.  I am not sure what is better.

("Return x and y, such that..." is saying, I think, the same as "Return the tuple ``x, y`` where x is ... and y is ..., such that ...", but it is shorter and easier to read, at least for me.)

RuchitJagodara: Oh... I thought you were saying that it is not returning a tuple. I am really sorry for the misunderstanding.

I have also updated this description. However, I have not changed this in the coefficient_matrix function because I was not sure whether I should change it or not.

src/sage/groups/matrix_gps/finitely_generated_gap.py
```diff
@@ -936,11 +936,11 @@ def _new_invariant_is_linearly_independent(F, invariants):
         sage: gens = [matrix(QQ, [[-1,1],[-1,0]]), matrix(QQ, [[0,1],[1,0]])]
         sage: G = MatrixGroup(gens)
         sage: s = Sequence(G.invariants_of_degree(14))                                  # needs sage.rings.number_field
-        sage: s.coefficient_matrix()[0].rank()                                          # needs sage.rings.number_field
+        sage: s.coefficients_monomials()[0].rank()                                          # needs sage.rings.number_field
```
mantepse: I'd suggest to line up the comment - annotations ("# needs...") again, it's silly, but I find it visually less destracting.

RuchitJagodara: Done !

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,10 +707,80 @@ def algebraic_dependence(self):
...
+        f = tuple(self)
+        nf = len(f)
+
+        if order is None:
+            m = sorted(self.monomials(), reverse=True)
+        else:
+            if isinstance(order, (list, tuple)):
+                m = order
+            else:
+                raise ValueError("order argument can only accept list or tuple.")
```
mantepse: maybe "order should be a list or tuple of the monomials appearing in the sequence" (and no "." at the end) is better.  The user might thing that an ordering, like "lex" would also be supported.

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,10 +707,80 @@ def algebraic_dependence(self):
...
+        nm = len(m)
+
+        #construct dictionary for fast lookups
```
mantepse: I think a space after "#" would be good

RuchitJagodara: Done !

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: Done !

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,10 +707,80 @@ def algebraic_dependence(self):
...
+
+        A = Matrix(R.base_ring(), nf, nm, sparse=sparse)
+
+        for x in range(nf):
+            poly = f[x]
+            for y in poly.monomials():
+                try:
+                    A[x, v[y]] = poly.monomial_coefficient(y)
+                except KeyError:
+                    raise ValueError("order argument does not contain all monomials")
```
mantepse, line 768: I think that this could be slightly simplified by iterating over `poly` itself:
```
for c, m in f[x]:
    try:
        A[c, v[m]] = c
    except KeyError:
        ...
```

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -754,33 +828,18 @@ def coefficient_matrix(self, sparse=True):
...
-        A = Matrix( R.base_ring(), nf, nm, sparse=sparse )
-
-        for x in range( nf ):
-            poly = f[x]
-            for y in poly.monomials():
-                A[ x , v[y] ] = poly.monomial_coefficient(y)
-
-        return A, Matrix(R,nm,1,m)
+        deprecation(37035,"the function 'coefficient_matrix' is deprecated; use 'coefficients_monomials' instead")
+        cm = self.coefficients_monomials()
```
mantepse: It is really minor, but I think `A, v = self.coefficients_monomials()` would be clearer.

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: But then what will be x ?

mantepse: Oh, yes, even that can be simplified!  So the whole thing, beginning with line 771 becomes:
```
for poly in f:
    for c, m in poly:
        try:
          A[c, v[m]] = c
        except KeyError:
          ...
```
`nf` and `nm` can be removed and their only occurrence be replaced by `len(f), len(m)`.  Possibly even `f = tuple(self)` can be removed, and you iterate `for poly in self` instead of `for poly in f`.

RuchitJagodara: But here, A[c, v[m]] will throw an error because in the first index, instead of the coefficient (i.e., c) of a monomial, there should be an index that represents the row of the coefficient matrix.

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: Done !

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -754,33 +828,18 @@ def coefficient_matrix(self, sparse=True):
             [  b]
             [  c]
             [  d]
-            [  1]
```
mantepse: why does the `1` vanish?  It shouldn't!

src/sage/rings/polynomial/multi_polynomial_sequence.py
mantepse: You are quite right, I mistook the `x` for a `c`.  However, the following might be right:
```
y = dict(zip(m, range(len(m))))
for x, poly in enumerate(f):
    for c, m in poly:
        try:
          A[x, y[m]] = c
        except KeyError:
          ...
```
Could you check?  I also renamed `v` to `y` for clarity.

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: Done ! Thank you for pointing out : )

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: I have changed the code according to your suggestions, please review it once, and sorry of delay, I was finding a way to make A*v == self but I didn't find a way to do so. :disappointed:

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,10 +707,78 @@ def algebraic_dependence(self):
...
+            sage: v
+            (a^2, a*b, b^2, a*c, b*c, c^2, b*d, c*d, d^2, a, b, c, d, 1)
+            sage: A*v
+            (a + 2*b + 2*c + 2*d - 1, a^2 + 2*b^2 + 2*c^2 + 2*d^2 - a,
+             2*a*b + 2*b*c + 2*c*d - b, b^2 + 2*a*c + 2*b*d - c)
+        """
+        from sage.modules.free_module_element import vector
+        from sage.matrix.constructor import Matrix
+
+        R = self.ring()
```
mantepse: very minor: `R` is now used only once (in line 768), so you could inline it.

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,10 +707,78 @@ def algebraic_dependence(self):
...
+            else:
+                raise ValueError("order argument can only accept list or tuple")
+        nv = len(v)
+
+        # construct dictionary for fast lookups
+        y = dict(zip(v, range(nv)))
+
+        A = Matrix(R.base_ring(), len(f), nv, sparse=sparse)
+
+        for x, poly in enumerate(f):
```
mantepse: extremely minor: change the whitespace as follows:
```
                raise ValueError("order argument can only accept list or tuple")

        nv = len(v)
        A = Matrix(self.ring().base_ring(), len(f), nv, sparse=sparse)
        y = dict(zip(v, range(nv)))  # construct dictionary for fast lookups
        for x, poly in enumerate(f):
```
or possibly, if you prefer
```
                raise ValueError("order argument can only accept list or tuple")

        A = Matrix(self.ring().base_ring(), len(f), len(v), sparse=sparse)
        y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
        for x, poly in enumerate(f):
```

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -761,26 +833,12 @@ def coefficient_matrix(self, sparse=True):
...
 
-        A = Matrix( R.base_ring(), nf, nm, sparse=sparse )
-
-        for x in range( nf ):
-            poly = f[x]
-            for y in poly.monomials():
-                A[ x , v[y] ] = poly.monomial_coefficient(y)
-
-        return A, Matrix(R,nm,1,m)
+        deprecation(37035,"the function 'coefficient_matrix' is deprecated; use 'coefficients_monomials' instead")
```
mantepse: there should be a single space after the comma.

src/sage/rings/polynomial/toy_variety.py
```diff
@@ -108,7 +108,7 @@ def coefficient_matrix(polys):
     .. NOTE::
 
         This function may be merged with
-        :meth:`sage.rings.polynomial.multi_polynomial_sequence.PolynomialSequence_generic.coefficient_matrix()` in the future.
+        :meth:`sage.rings.polynomial.multi_polynomial_sequence.PolynomialSequence_generic.coefficients_monomials()` in the future.
```
mantepse: If you have time, this would be an excellent next project, together with #37075.

src/sage/rings/polynomial/multi_polynomial_sequence.py
RuchitJagodara: Ohh ! I just realised that below line may through an error in some cases 

```python
    for c, m in poly:
```

It will happen in case of constant terms, because there will not be enough values to unpack in poly. So to handle this problem, I have changed the code as follows.

```python
        for x, poly in enumerate(f):
            for m in poly.monomials():
                try:
                    A[x, y[m]] = poly.monomial_coefficient(m)
                except KeyError:
                    raise ValueError("order argument does not contain all monomials")
```

mantepse: Sorry, could you clarify the problem?  It is not clear to me why `for c, m in poly` might fail.  That looks like a bug to me.  I think that the members of `self` are guaranteed to be multivariate polynomials.  In any case, there should be a doctest that fails with the simpler code - I cannot reproduce the problem.

RuchitJagodara: @mantepse can you please review these changes.

mantepse: I'm sorry,I don't understand, did I miss a commit?

mantepse:
Hi @RuchitJagodara !  I am now a bit confused.  Could you let me now what the state of the PR is now?  Was the problem with `for c, m in poly` not a problem after all?

You squashed the commits, right?

RuchitJagodara:
> Hi @RuchitJagodara ! I am now a bit confused. Could you let me now what the state of the PR is now? Was the problem with `for c, m in poly` not a problem after all?

No, there was no problem with that.

> You squashed the commits, right?

Yes.

I am really sorry for the confusion caused by this change. I reset the branch because there were some conflicts in my branches due to which my one PR got merged without any positive review, so I was just sorting out that issue. Therefore, I had to reset this branch, but now I have included all commits that were missing in this branch. I am really sorry for the inconvenience caused by this. :disappointed:

mantepse:
> > Hi @RuchitJagodara ! I am now a bit confused. Could you let me now what the state of the PR is now? Was the problem with `for c, m in poly` not a problem after all?
> 
> No, there was no problem with that.

Oh, that's good.

> I am really sorry for the confusion caused by this change. I reset the branch because there were some conflicts in my branches due to which my one PR got merged without any positive review, so I was just sorting out that issue. Therefore, I had to reset this branch, but now I have included all commits that were missing in this branch. I am really sorry for the inconvenience caused because of this. 😞

Absolutely no problem at all!  Could you just do the minor modifications?  I would then wait for a green bot and set it to "positive review".

I am so glad that you did this, by the way!

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -707,6 +707,74 @@ def algebraic_dependence(self):
...
+
+        # construct dictionary for fast lookups
+        y = dict(zip(v, range(nv)))
+
+        A = Matrix(R.base_ring(), len(f), nv, sparse=sparse)
+
+        for x, poly in enumerate(f):
+            for m in poly.monomials():
+                try:
+                    A[x, y[m]] = poly.monomial_coefficient(m)
```
mantepse: This looks backwards and is possibly quite inefficient.  Could you please indicate what's going wrong with the other version?

RuchitJagodara:
Does this PR need any further changes becuase I have done all necessary changes. 

And should I add some tests for new function becuase codecov/patch check is failing ?

mantepse:
@RuchitJagodara, I think there is a misunderstanding: there are still comments which have not been adressed, most importantly the question about line 771 we discussed before.  The last answer I saw I responded to in comment https://github.com/sagemath/sage/pull/37035#issuecomment-1911793242, but then the code changed back.

RuchitJagodara:
I have addreseed that in [comment](https://github.com/sagemath/sage/pull/37035#discussion_r1459553008). 

Actually, there is a problem with `for c, m in poly:`, because this line will give an error when there is a constant term appearing in the polynomial. 

When you asked `Was the problem with for c, m in poly not a problem after all?`, I thought you were asking that if that was the cause of the change. So I just replied no on that. But ya that code won't work for some cases but I have modified it accordingly and it should work now.

mantepse:
> Actually, there is a problem with `for c, m in poly:`, because this line will give an error when there is a constant term appearing in the polynomial.

Could you please give me a simple example for that?  If I do 
```
sage: P.<x,y> = QQ[]
sage: p = x^2*y + 1
sage: [(c, m) for c, m in p]
[(1, x^2*y), (1, 1)]
```
it works.

RuchitJagodara:
Hmm... It is working fine now. I think something should have changed because the last time I tried this method, it was not working; it was giving me a ValueError, saying there were not enough values to unpack. Because of that, I changed the code at that time after leaving a comment. However, I think now it is safe to go this way because I have checked the code in different test cases, and it is working fine.

RuchitJagodara:
@mantepse, I don't know why, but the checks are not running; they are queued. Do you know what the problem is?

mantepse:
That's indeed very confusing.

Anyhow, could you please also do the 2 whitespace comments (which you can see on "Files changed" tab)?

mantepse:
> @mantepse, I don't know why, but the checks are not running; they are queued. Do you know what the problem is?

I think that's just because too many people are testing right now.

(So it really would be good to do the final changes now, in order not to waste resources)

RuchitJagodara:
> Anyhow, could you please also do the 2 whitespace comments (which you can see on "Files changed" tab)?

Are you talking about these lines ?

```
        Return the matrix of coefficients ``A`` and
        the matching vector of monomials ``v``, such that ``A*v == vector(self)``.

        Thus value of ``A[i,j]`` corresponds the coefficient of the
        monomial ``v[j]`` in the ``i``-th polynomial in this system.

        Monomials are ordered w.r.t. the term ordering of ``order``
        if given; otherwise, they are ordered w.r.t. ``self.ring()``
        in reverse order, i.e., such that the smallest entry comes last.

```

RuchitJagodara:
> Can you see these comments? github is quite tricky at times...

I am not able to see those whitespace comments but I have removed the R from the code.

mantepse:
Here is a screenshot, you should see this in the "Files changed" tab.

image

RuchitJagodara:
Oh, I see. I am not able to see this comments. Could you please submit the review, I think after submitting the review I will be able to see those comments.

RuchitJagodara:
Although, I have updated the code with suggested changes. Please let me know if there is anything else.

mantepse:
> Oh, I see. I am not able to see this comments. Could you please submit the review, I think after submitting the review I will be able to see those comments. Because it is showing pending tag on the comments.

No, the pending tag means that the code was not yet changed. It would be really important that you can see these comments for developing sage

RuchitJagodara:
Hmmm.... It's very strange because I am not able to see those comments.

Screenshot from 2024-01-31 00-22-20


And the resolved conversation that you can see in the above screenshot is [review](https://github.com/sagemath/sage/pull/37035#discussion_r1452623651)

mantepse:
Dear @RuchitJagodara, I now see the problem: the BooleanPolynomialRing is a polynomial ring which does not adhere to the polynomial protocol :-1:   Here is what I would suggest to do:
```diff
diff --git a/src/sage/rings/polynomial/multi_polynomial_sequence.py b/src/sage/rings/polynomial/multi_polynomial_sequence.py
index 72ce72f748..c8e6f624ba 100644
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -294,8 +294,10 @@ def PolynomialSequence(arg1, arg2=None, immutable=False, cr=False, cr_str=None):
     from sage.structure.element import is_Matrix
     try:
         from sage.rings.polynomial.pbori.pbori import BooleanMonomialMonoid
+        from sage.rings.polynomial.pbori.pbori import BooleanPolynomialRing
     except ImportError:
         BooleanMonomialMonoid = ()
+        BooleanPolynomialRing = ()
 
     is_ring = lambda r: is_MPolynomialRing(r) or isinstance(r, BooleanMonomialMonoid) or (is_QuotientRing(r) and is_MPolynomialRing(r.cover_ring()))
 
@@ -369,6 +371,8 @@ def PolynomialSequence(arg1, arg2=None, immutable=False, cr=False, cr_str=None):
     if not isinstance(K, FiniteField) or K.characteristic() != 2:
         return PolynomialSequence_generic(parts, ring, immutable=immutable, cr=cr, cr_str=cr_str)
     elif K.degree() == 1:
+        if isinstance(ring, BooleanPolynomialRing):
+            return BooleanPolynomialSequence(parts, ring, immutable=immutable, cr=cr, cr_str=cr_str)
         return PolynomialSequence_gf2(parts, ring, immutable=immutable, cr=cr, cr_str=cr_str)
     elif K.degree() > 1:
         return PolynomialSequence_gf2e(parts, ring, immutable=immutable, cr=cr, cr_str=cr_str)
@@ -750,7 +754,6 @@ class PolynomialSequence_generic(Sequence_generic):
         from sage.modules.free_module_element import vector
         from sage.matrix.constructor import Matrix
 
-        f = tuple(self)
         if order is None:
             v = sorted(self.monomials(), reverse=True)
         else:
@@ -760,8 +763,8 @@ class PolynomialSequence_generic(Sequence_generic):
                 raise ValueError("order argument can only accept list or tuple")
 
         y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
-        A = Matrix(self.ring().base_ring(), len(f), len(v), sparse=sparse)
-        for x, poly in enumerate(f):
+        A = Matrix(self.ring().base_ring(), len(self), len(v), sparse=sparse)
+        for x, poly in enumerate(self):
             for c, m in poly:
                 try:
                     A[x, y[m]] = c
@@ -1695,6 +1698,62 @@ class PolynomialSequence_gf2(PolynomialSequence_generic):
         else:
             return PolynomialSequence_generic.reduced(self)
 
+class BooleanPolynomialSequence(PolynomialSequence_gf2):
+    r"""
+    A sequence over the Boolean polynomial ring.
+    """
+    def coefficients_monomials(self, order=None, sparse=True):
+        """
+        Return the matrix of coefficients ``A`` and
+        the matching vector of monomials ``v``, such that ``A*v == vector(self)``.
+
+        Thus value of ``A[i,j]`` corresponds the coefficient of the
+        monomial ``v[j]`` in the ``i``-th polynomial in this system.
+
+        Monomials are ordered w.r.t. the term ordering of ``order``
+        if given; otherwise, they are ordered w.r.t. ``self.ring()``
+        in reverse order, i.e., such that the smallest entry comes last.
+
+        INPUT:
+
+        - ``sparse`` - construct a sparse matrix (default: ``True``)
+        - ``order`` - a list or tuple specifying the order of monomials (default: ``None``)
+
+        EXAMPLES::
+
+            sage: # needs sage.rings.polynomial.pbori
+            sage: B.<x,y,z> = BooleanPolynomialRing()
+            sage: F = Sequence([x*y + y + 1, z + 1])
+            sage: A, v = F.coefficients_monomials()
+            sage: A
+            [1 1 0 1]
+            [0 0 1 1]
+            sage: v
+            (x*y, y, z, 1)
+            sage: A*v
+            (x*y + y + 1, z + 1)
+        """
+        from sage.modules.free_module_element import vector
+        from sage.matrix.constructor import Matrix
+
+        if order is None:
+            v = sorted(self.monomials(), reverse=True)
+        else:
+            if isinstance(order, (list, tuple)):
+                v = order
+            else:
+                raise ValueError("order argument can only accept list or tuple")
+
+        y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
+        A = Matrix(self.ring().base_ring(), len(self), len(v), sparse=sparse)
+        for x, poly in enumerate(self):
+            for m in poly:
+                try:
+                    A[x, y[m]] = 1
+                except KeyError:
+                    raise ValueError("order argument does not contain all monomials")
+        return A, vector(v)
+
 
 class PolynomialSequence_gf2e(PolynomialSequence_generic):
     r"""

```

RuchitJagodara:
I have updated the code according to your suggestions.

mantepse (approved):
Thank you so much for your patience!

RuchitJagodara:
Now, I am able to see those comments to which you were referring earlier.

> If you have time, this would be an excellent next project, together with https://github.com/sagemath/sage/issues/37075.

I think that issue has been resolved, but it seems that there is a function coeffcient_matrix in toy_variety.py. In which I found a note...

```
    .. NOTE::

        This function may be merged with
        :meth:`sage.rings.polynomial.multi_polynomial_sequence.PolynomialSequence_generic.coefficient_matrix()` in the future.
```

So, should I merge this function with this coefficients_monomials function because this note was added two years ago, so I think it is okay if we merge this. What do you say, @mantepse?

PR #37038, opened 2024-01-09
Add minimal_normal_subgroups and maximal_normal_subgroups functions

RuchitJagodara:
- Added the function
This patch implements a minimal_normal_subgroups function in PermutationGroup_generic class.

### :memo: Checklist

### :hourglass: Dependencies

NA


tscrim (changes requested):
The result should be converted back into Sage (permutation) (sub)groups.

Also, I don't see why $S_8$ should be a test (which are hidden in the final html doc) rather than a normal example.

RuchitJagodara:
I have done some necessary changes. But I just found that these methods can be implemented for all groups as the functions MaximalNormalSubgroups and MinimalNormalSubgroups works for all groups in gap, so I am considering implementing this in Sage for all groups, but I am having difficulty in finding the appropriate location to define these functions. Can you help me with this?

tscrim:
You can add it to the `Groups()` category, but that wouldn't work because not all groups in Sage have a corresponding GAP version (see, e.g., right-angled Artin groups). However, there is a base class for all GAP groups, so you can implement it there.

RuchitJagodara:
I have relocated the function, but I believe the implementation is suboptimal. This is because it will consume a significant amount of time, as it needs to generate the generators first and then convert them into a subgroup (as demonstrated below). I attempted to explore alternative options but was unsuccessful. Do you have any suggestions for the fix for this?

https://github.com/sagemath/sage/blob/297e8b2e5769d703741abba46edaf5d5a254f60e/src/sage/groups/libgap_wrapper.pyx#L359

Also, were you referring to this particular Class? Because now for permutation groups it is giving an AttributeError (seems like permutationgroups are not derived from this class).

tscrim:
Within `ParentLibGAP`, there is the `_subgroup_constructor()` method that you should use since you already have the GAP group (and always takes a GAP group object from doing a grep search).

RuchitJagodara:
I have implemented the modifications; however, I remain uncertain about the location of these functions. I am unable to invoke these functions when I have a permutation group(It is giving me AttributeError). Because this class is never inherited in the PermutationGroup class. So what should I do ? Because I am not sure whether inheriting that function in PermutationGroup class will be good or not.

tscrim:
Something that is annoying is that permutation groups don't seem to use libgap. I think you need to have your previous implement in `permgroup`. Then that should cover the most number of cases I think...

src/sage/groups/libgap_wrapper.pyx
```diff
@@ -345,6 +345,35 @@ class ParentLibGAP(SageObject):
...
+             [6 0]
+             [0 6]
+             ) of Special Linear Group of degree 2 over Finite Field in z2 of size 7^2]
+        """
+        return [self._subgroup_constructor(gap_subgroup) for gap_subgroup in self.gap().MinimalNormalSubgroups()]
+
+    def maximal_normal_subgroups(self):
+        """
+        Return a list containing those proper normal subgroups of the group G that are maximal among the proper normal subgroups.
+        Gives error if G/G' is infinite, yielding infinitely many maximal normal subgroups.
```
tscrim:
```suggestion
        Return the maximal proper normal subgroups of ``self``.

        This raises an error if `G/[G, G]` is infinite, yielding infinitely
        many maximal normal subgroups.
```

src/sage/groups/libgap_wrapper.pyx
```diff
@@ -345,6 +345,35 @@ class ParentLibGAP(SageObject):
         """
         return self._libgap._repr_()
 
+    def minimal_normal_subgroups(self):
+        """
+        Return a list containing those nontrivial normal subgroups of the group that are minimal among the nontrivial normal subgroups.
```
tscrim:
```suggestion
        Return the nontrivial minimal normal subgroups ``self``.
```

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -4106,6 +4106,33 @@ def isomorphism_type_info_simple_group(self):
         else:
             raise TypeError("group must be simple")
 
+    def minimal_normal_subgroups(self):
+        """
+        Return a list containing those nontrivial normal subgroups of the group that are minimal among the nontrivial normal subgroups.
```
tscrim: Same as in the libgap file.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -4106,6 +4106,33 @@ def isomorphism_type_info_simple_group(self):
...
+            sage: G = PermutationGroup([(1,2,3),(4,5)])
+            sage: G.minimal_normal_subgroups()
+            [Subgroup generated by [(4,5)] of (Permutation Group with generators [(4,5), (1,2,3)]),
+             Subgroup generated by [(1,2,3)] of (Permutation Group with generators [(4,5), (1,2,3)])]
+        """
+        return [self.subgroup(gap_group=gap_subgroup) for gap_subgroup in self.gap().MinimalNormalSubgroups()]
+
+    def maximal_normal_subgroups(self):
+        """
+        Return a list containing those proper normal subgroups of the group G that are maximal among the proper normal subgroups.
```
tscrim: Same as in the libgap file.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -4106,6 +4106,33 @@ def isomorphism_type_info_simple_group(self):
...
+        Return a list containing those nontrivial normal subgroups of the group that are minimal among the nontrivial normal subgroups.
+
+        EXAMPLES::
+
+            sage: G = PermutationGroup([(1,2,3),(4,5)])
+            sage: G.minimal_normal_subgroups()
+            [Subgroup generated by [(4,5)] of (Permutation Group with generators [(4,5), (1,2,3)]),
+             Subgroup generated by [(1,2,3)] of (Permutation Group with generators [(4,5), (1,2,3)])]
+        """
+        return [self.subgroup(gap_group=gap_subgroup) for gap_subgroup in self.gap().MinimalNormalSubgroups()]
```
tscrim:
```suggestion
        return [self.subgroup(gap_group=gap_subgroup)
                for gap_subgroup in self.libgap().MinimalNormalSubgroups()]
```
We want to move things to libgap (direct library interface), and if it doesn't use libgap, this redirects to a gap call.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -4106,6 +4106,33 @@ def isomorphism_type_info_simple_group(self):
...
+        Gives error if G/G' is infinite, yielding infinitely many maximal normal subgroups.
+
+        EXAMPLES::
+
+            sage: G = PermutationGroup([(1,2,3),(4,5)])
+            sage: G.maximal_normal_subgroups()
+            [Subgroup generated by [(1,2,3)] of (Permutation Group with generators [(4,5), (1,2,3)]),
+             Subgroup generated by [(4,5)] of (Permutation Group with generators [(4,5), (1,2,3)])]
+        """
+        return [self.subgroup(gap_group=gap_subgroup) for gap_subgroup in self.gap().MaximalNormalSubgroups()]
```
tscrim:
```suggestion
        return [self.subgroup(gap_group=gap_subgroup)
                for gap_subgroup in self.libgap().MaximalNormalSubgroups()]
```

src/sage/groups/libgap_wrapper.pyx
```diff
@@ -345,6 +345,35 @@ class ParentLibGAP(SageObject):
...
+
+        EXAMPLES::
+
+            sage: SL(2,GF(49)).minimal_normal_subgroups()
+            [Subgroup with 1 generators (
+             [6 0]
+             [0 6]
+             ) of Special Linear Group of degree 2 over Finite Field in z2 of size 7^2]
+        """
+        return [self._subgroup_constructor(gap_subgroup) for gap_subgroup in self.gap().MinimalNormalSubgroups()]
```
tscrim:
```suggestion
        return [self._subgroup_constructor(gap_subgroup)
                for gap_subgroup in self._libgap.MinimalNormalSubgroups()]
```
No need for the redirect.

src/sage/groups/libgap_wrapper.pyx
```diff
@@ -345,6 +345,35 @@ class ParentLibGAP(SageObject):
...
+
+        EXAMPLES::
+
+            sage: SL(2,GF(49)).minimal_normal_subgroups()
+            [Subgroup with 1 generators (
+             [6 0]
+             [0 6]
+             ) of Special Linear Group of degree 2 over Finite Field in z2 of size 7^2]
+        """
+        return [self._subgroup_constructor(gap_subgroup) for gap_subgroup in self.gap().MaximalNormalSubgroups()]
```
tscrim:
```suggestion
        return [self._subgroup_constructor(gap_subgroup)
                for gap_subgroup in self._libgap.MaximalNormalSubgroups()]
```

tscrim (changes requested):

src/sage/groups/perm_gps/permgroup.py
RuchitJagodara: Here, libgap() is not working, it is giving an AttributeError. But I found that other functions similar to this one uses _libgap_() so I have implemented that, is it okay ?

tscrim: Yes, that's right. Sorry, I forgot the underscores (`_`).

src/sage/groups/libgap_wrapper.pyx
tscrim: This should not be `_libgap()` but the attribute `_libgap` as suggested.

tscrim:
Please also squash and give at least a little bit better commit messages for the first two. These automated ones from github are not good (I want to fire that "feature" into the sun).

RuchitJagodara:
Done! I have applied all the changes that you suggested. Thank you for your help, @tscrim  throughout :) . I learned a lot while working on this, and I will make sure that I won't make these types of mistakes ahead.

tscrim (approved):
No problem. Thank you for the changes. This is just dealing with some trickier parts of Sage with how things are implemented because of the interfaces.

I am cc-ing @dimpase in case he has some comments.

dimpase:
@fchapoton did some work on removing pexpect GAP and replacing it with libgap just last week.

Maybe it should be taken into account here

tscrim:
Good point. I rebased this off the latest develop, which should have #36889 included. Since there are no merge conflicts and @RuchitJagodara wrote the code with future-proofing in mind, there shouldn’t be any issues (which will be reflected in the tests all passing). I also don’t see any way to make additional _improvements at this point either.

PR #37039, opened 2024-01-09
some details in Galois group as permutation group

fchapoton:
just a few minor changes in the modified file

removing the unused attribute `_base` 

### :memo: Checklist


tscrim (commented):
LGTM overall, but I don't quite agree with changing the docstrings for the `@lazy_attribute`s since they don't quite return something. It isn't too strong of an opinion, but it feels a bit more natural to declare what it is. If you feel a bit more strongly about changing them, then we can leave it.

tscrim (approved):
Thank you. LGTM.

Issue #31761, opened 2021-05-01
Stop loading all installed GAP packages


antonio-rojas:
<div id="comment:0"></div>

On a fresh run of the GAP interface, Sage caches a GAP workspace which loads every single GAP package on the system. While this might be OK with Sage's trimmed-down version of gap_packages, if using a system GAP install 
with all optional packages installed this can take a massive amount of memory (~1GB with all packages included in the 4.11.1 tarball), which in many casses surpasses the default gap_memory_pool_size value, causing crashes. This will become a common issue once #29644 is merged.

Possible solutions:

1) Don't preload packages in Sage and let GAP load its default list of packages

2) Load a hardcoded list of packages (such as those installed by the gap and gap_packages spkg's)

CC:  @dimpase

Component: **packages: standard**

_Issue created by migration from https://trac.sagemath.org/ticket/31761_

antonio-rojas:
Description changed:
``````diff
--- 
+++ 
@@ -2,6 +2,8 @@
 with all optional packages installed this can take a massive amount of memory (~1GB with all packages included in the 4.11.1 tarball), which in many casses surpasses the default gap_memory_pool_size value, causing crashes. This will become a common issue once #29644 is merged.
 
 Possible solutions:
+
 1) Don't preload packages in Sage and let GAP load its default list of packages
+
 2) Load a hardcoded list of packages (such as those installed by the gap and gap_packages spkg's)
 
``````

**Comment by orlitzky** [comment]
This is partially done in https://github.com/sagemath/sage/pull/36792, where the GAP interface and libgap are both now initialized with `-A`, and each has its own explicit list of auto-loaded packages. For example, `src/sage/ext_data/gap/sage.g ` now ends with,

```
# Load the GAP packages that GAP itself tries to autoload in the
# default configuration (see "PackagesToLoad" in lib/package.gi). The
# combination of passing -A to gap and these LoadPackage statements
# allows us to load the usual set of packages, but only if they are
# installed. So most people will get exactly the default behavior,
# but minimal installations won't throw warnings and fail tests.
_autoloads := [ "autpgrp", "alnuth", "crisp", "ctbllib", "factint", "fga",
                "irredsol", "laguna", "polenta", "polycyclic", "resclasses",
                "sophus", "tomlib" ];
for p in _autoloads do
  LoadPackage(p);
od;
```

I say "partially" though because I missed `gap_reset_workspace()`. It's still trying to load all installed packages.

[...]

PR #37049, opened 2024-01-12
Dont load all GAP packages when resetting a workspace

orlitzky:
The `gap_reset_workspace()` function currently loops through all installed GAP packages (both by sage and on the system) and loads them before saving the new workspace. This is no longer necessary or desired: the two GAP initialization files already load a specific set of packages used in sage, and loading more can cause problems including wasted memory and unexpected changes of behavior. Moreover, the need for _users_ to load additional packages interactively has been documented.

Fixes:

- https://github.com/sagemath/sage/issues/31761

### DIFF
--- a/src/sage/ext_data/gap/sage.g
+++ b/src/sage/ext_data/gap/sage.g
@@ -2,6 +2,9 @@
 # Sage support utilities to read into the GAP session.
 #
 
+# Disable color in the prompt, it interferes with parsing.
+ColorPrompt(false);
+
 # Prevent loading the xgap package; we use the -p flag to GAP in order to
 # communicate with it via the pexpect interface; this is normally used by
 # for an xgap window to communicate with GAP, so unfortunately setting this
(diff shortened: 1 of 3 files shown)


orlitzky:
@antonio-rojas does this help your memory usage?

antonio-rojas:
> @antonio-rojas does this help your memory usage?

Yes, the only remaining issue with all gap-packages installed (besides the Browse issue) is the test in `sage/tests/gap_packages.py` which explicitly tries to load all installed packages (and blows up for me). I wonder if it would make sense to replace it with testing just the default preloaded list of packages.

orlitzky:
> I wonder if it would make sense to replace it with testing just the default preloaded list of packages.

We already load-test those as soon as we use GAP in any other doctest. Now that `all_installed_packages()` is not used in `gap_reset_workspace()`, we could just `git rm` the whole thing? GAP packages have their own test suites. If sage-the-distro wants to test them, I think a better place would be in `build/gap_packages/spkg-check`.

PR #36449, opened 2023-10-11
Autogenerate flint headers

videlec:
We
- write down a script to auto-generate flint header files (currently at https://github.com/videlec/flint-header-sage-autogen)
- replace most of the files in `src/sage/libs/flint` by auto-generated files. Doing so, the custom sage code in `src/sage/libs/flint` is always contained in files suffixed by `_sage.pxd`/`_sage.pyx`.
- deprecate the files in `src/sage/libs/arb/`

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

#35848 

#### Upstream issues and patches

- https://github.com/flintlib/flint/pull/1523
- https://github.com/flintlib/flint/issues/1529
- https://github.com/flintlib/flint/pull/1532
- https://github.com/flintlib/flint/issues/1535
- https://github.com/flintlib/flint/pull/1536
- https://github.com/cython/cython/issues/5779
- https://github.com/flintlib/flint/issues/1653
- https://github.com/flintlib/flint/pull/1655

PR #37064, opened 2024-01-15
Fix build when linbox is build with fplll support

antonio-rojas:
Undefine flint macros that conflict with fplll variables and break build when both are included

See https://github.com/sagemath/sage/pull/36449#issuecomment-1891095354

### DIFF
--- a/src/sage/libs/flint/flint_wrap.h
+++ b/src/sage/libs/flint/flint_wrap.h
@@ -168,4 +168,10 @@
 
 #pragma pop_macro("ulong")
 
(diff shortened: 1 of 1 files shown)


videlec (approved):
Looks good to me.

tobiasdiez (approved):
At https://github.com/sagemath/sage/pull/37077, I've encountered the same error and your PR fixed it. Thanks!

antonio-rojas:
Test failure looks unrelated, setting to positive review

PR #37069, opened 2024-01-16
Use Parent in number fields and QQ

fchapoton:
Towards the removal of `ParentWithGens` : use parent in number fields and in QQ

### :memo: Checklist

### DIFF
--- a/src/sage/rings/rational_field.py
+++ b/src/sage/rings/rational_field.py
@@ -62,7 +62,7 @@ ZZ = None
 import sage.rings.number_field.number_field_base as number_field_base
 from sage.misc.fast_methods import Singleton
 from sage.misc.superseded import deprecated_function_alias
-from sage.structure.parent_gens import ParentWithGens
+from sage.structure.parent import Parent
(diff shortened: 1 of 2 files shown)


src/sage/rings/rational_field.py
```diff
@@ -1663,7 +1664,7 @@ def _factor_univariate_polynomial(self, f):
         G = list(f._pari_with_name().factor())
 
         # normalize the leading coefficients
-        F = [(f.parent()(g).monic(), int(e)) for (g,e) in zip(*G)]
+        F = [(f.parent()(g).monic(), int(e)) for (g, e) in zip(*G)]
```
tscrim:
```suggestion
        P = f.parent()
        F = [(P(g).monic(), int(e)) for (g, e) in zip(*G)]
```
Can we do one micro-optimization here since we are already modifying this line?

tscrim (approved):
Sorry; just one small while-we-are-at-it thing. You can set a positive review once changed or decided to ignore.

fchapoton:
ok, setting to positive, thanks

PR #37070, opened 2024-01-16
use directly OSError everywhere

fchapoton:
fixed by `ruff --fix --select=UP024 src/`

OSError has replaced several errors since python 3.3:

https://stackoverflow.com/questions/29347790/difference-between-ioerror-and-oserror

https://peps.python.org/pep-3151/#confusing-set-of-os-related-exceptions

### :memo: Checklist


dcoudert (approved):
LGTM.

fchapoton:
merci beaucoup David. Et bonne année !

PR #37072, opened 2024-01-16
some cleanup in overconvergent and quaternion

fchapoton:
Two different things:

- cleaning the file `overconvergent/genus0`, adding a `TestSuite` and making it pass ; getting rid of non-sensical `gens_dict` there

- making quaternion algebra return a tuple of generators

### :memo: Checklist

### DIFF
--- a/src/sage/modular/overconvergent/genus0.py
+++ b/src/sage/modular/overconvergent/genus0.py
@@ -194,11 +194,10 @@ from sage.rings.power_series_ring import PowerSeriesRing
 from sage.rings.rational_field import QQ
 from sage.structure.element import Vector, ModuleElement
 from sage.structure.richcmp import richcmp
+from .weightspace import WeightSpace_constructor as WeightSpace, WeightCharacter
 
 lazy_import('sage.modular.modform.j_invariant', 'j_invariant_qexp')
 lazy_import('sage.rings.padics.factory', 'Qp', as_='pAdicField')
-
-from .weightspace import WeightSpace_constructor as WeightSpace, WeightCharacter
 
 
 __ocmfdict = {}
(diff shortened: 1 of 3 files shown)


src/sage/modular/overconvergent/genus0.py
```diff
@@ -882,23 +883,6 @@ def ngens(self):
         """
         return Infinity
 
-    def gens_dict(self):
```
dcoudert, line 885: Shouldn't we first deprecate this method ?

src/sage/modular/overconvergent/genus0.py
```diff
@@ -1638,7 +1623,7 @@ def governing_term(self, r):
         p = self.prime()
         F = self.parent().base_ring()
         if not isinstance(F, sage.rings.abc.pAdicField):
-            F = pAdicField(p)
+            F = pAdicField(p)  # noqa:F821
```
dcoudert, line 1626: What's the meaning of this annotation, used in several places ?

src/sage/modular/overconvergent/genus0.py
fchapoton: given that it returns not implemented error, and explains itself to be meaningless I think we can just remove it

src/sage/modular/overconvergent/genus0.py
fchapoton: this is to remove the pycodestyle warning about pAdicField not being defined (because lazy-imported)

src/sage/modular/overconvergent/genus0.py
dcoudert: ok, I agree

src/sage/modular/overconvergent/genus0.py
dcoudert: I didn't know that.

dcoudert (approved):
LGTM.

src/sage/modular/overconvergent/genus0.py
fchapoton: noqa = no quality assessment, I think ; then one or several of the standard codes

dcoudert:
Yes.

PR #37073, opened 2024-01-16
`build/pkgs/free_fonts`: Fix fedora system package information

mkoeppe:
The package `gnu-free-fonts` does not exist, as seen for example in 
- `fedora-36-maximal` (https://github.com/sagemath/sage/actions/runs/7521668027/job/20497322866).
- `fedora-39-maximal` (https://github.com/sagemath/sage/actions/runs/7521668027/job/20497323960#step:11:20270)

`gnu-free-fonts` is merely the name of the source package, the actual rpms are called `gnu-free-mono-fonts`, `gnu-free-sans-fonts `, `gnu-free-serif-fonts` https://koji.fedoraproject.org/koji/buildinfo?buildID=2308657

Here we adjust for it.

According to https://rpmfind.net/linux/rpm2html/search.php?query=gnu-free-sans-fonts, the packages are available for fedora >=37, EPEL 9, centos-7, centos-stream-8. Accordingly we remove the setting `IGNORE_MISSING_SYSTEM_PACKAGES=no` from fedora < 37.

Also adding fedora-40, the new rawhide.

### :memo: Checklist


mkoeppe:
Actually, it looks like `gnu-free-fonts` is merely the name of the source package - https://src.fedoraproject.org/rpms/gnu-free-fonts

The actual rpms are called `gnu-free-mono-fonts`, `gnu-free-sans-fonts `, `gnu-free-serif-fonts` https://koji.fedoraproject.org/koji/buildinfo?buildID=2308657

dimpase:
OK for Fedora 38 and OpenBSD - tested on real systems

kwankyu:
Thanks for fixing this. 

Did this pass the "CI Linux Incremental" test?

kwankyu:
I tested with some fedora versions. It works well.

I read this:
```
    # Set this to 'yes' instead of 'no' to ignore missing system packages - by installing them one by one
    # and ignoring errors.  We use that to take care of old versions of distributions.
    # For -maximal environments, the default is 'yes' but later we override it for rolling distributions
    # (but not for unstable distributions that often have intermittent issues).
                                               IGNORE_MISSING_SYSTEM_PACKAGES=no
    maximal:                                   IGNORE_MISSING_SYSTEM_PACKAGES=yes
```
but I still don't understand the purpose of `IGNORE_MISSING_SYSTEM_PACKAGE`. So would you explain this statement

> According to https://rpmfind.net/linux/rpm2html/search.php?query=gnu-free-sans-fonts, the packages are available for fedora >=37, EPEL 9, centos-7, centos-stream-8. Accordingly we remove the setting `IGNORE_MISSING_SYSTEM_PACKAGES=no` from fedora < 37.

?

mkoeppe:
> I tested with some fedora versions. It works well.

Thanks for testing!

> I read this:
> 
> ```
>     # Set this to 'yes' instead of 'no' to ignore missing system packages - by installing them one by one
>     # and ignoring errors.  We use that to take care of old versions of distributions.
>     # For -maximal environments, the default is 'yes' but later we override it for rolling distributions
>     # (but not for unstable distributions that often have intermittent issues).
>                                                IGNORE_MISSING_SYSTEM_PACKAGES=no
>     maximal:                                   IGNORE_MISSING_SYSTEM_PACKAGES=yes
> ```
> 
> but I still don't understand the purpose of `IGNORE_MISSING_SYSTEM_PACKAGES`.

It's all about balancing the effective use of system packages vs. the burden of maintenance:
- If system packages that we refer to are renamed or become unavailable otherwise, we want the CI to tell us; we don't want to find out from user complaints. So for standard packages on reasonable versions of distributions, we make it a hard error if the listed system packages cannot be installed.
- To avoid a much larger burden of maintenance, we only have one set of distro package information for each large family of distributions. All of fedora, centos, rhel is recorded in `fedora.txt`. All of debian, ubuntu, and the various value-added or sugar-added variants is recorded in `debian.txt` etc. So if a variant misses one of our standard packages – tough luck, we'll just set `IGNORE_MISSING_SYSTEM_PACKAGES=yes` for this variant. This means that when testing that variant, we determine the available subset of packages dynamically. When a package disappears, we can see it in the log, but the test of the platform is still supposed to go through. 
- Testing the `-maximal` package configuration has two main purposes: 1) To test that system packages that we don't know how to use (no spkg-configure.m4) don't break the build by their sheer presence on the system. 2) For tests of optional packages that depend on other optional packages.

> So would you explain this statement
> 
> > According to https://rpmfind.net/linux/rpm2html/search.php?query=gnu-free-sans-fonts, the packages are available for fedora >=37, EPEL 9, centos-7, centos-stream-8. Accordingly we remove the setting `IGNORE_MISSING_SYSTEM_PACKAGES=no` from fedora < 37.
> 
> ?

Before this PR, we had the line
```
fedora-34:                                 IGNORE_MISSING_SYSTEM_PACKAGES=no
```
This marked `fedora-34` as a very reasonable distro variant: We require that all system packages listed in `fedora.txt`, whether standard or optional, whether we know how to use them or not, must be installable. And running the `fedora-34-maximal` tested that, giving a hard error otherwise.

The comment from the ticket description says that because gnu-free-sans-fonts is not available on `fedora-34`, we should no longer demand that `fedora-34-maximal` has all listed packages. By removing the line setting `IGNORE_MISSING_SYSTEM_PACKAGES=no`, we are back to using the default: `IGNORE_MISSING_SYSTEM_PACKAGES=no` for `-minimal` and `-standard`, but `IGNORE_MISSING_SYSTEM_PACKAGES=yes` for `-maximal`.

It's admittedly a bit twisty...

kwankyu:
Thanks for the detailed explanation.

kwankyu (approved):

mkoeppe:
Thanks!

PR #36184, opened 2023-09-03
add class groups of binary quadratic forms

yyyyx4:
This patch adds an implementation of the form class group for positive definite binary quadratic forms. (We restrict to definite forms for now since it is the easier case: Every class contains a unique reduced form.)

Most of the required functionality was already available in PARI or Sage, but the types introduced in this patch make it easier to manipulate form classes and interface with existing algorithms for generic abelian groups.

PR #37074, opened 2024-01-16
implement natural morphism from cl(f²D) to cl(D)

yyyyx4:
...for class groups of binary quadratic forms, which we have since #36184.

This map is defined by finding a class representative $[a,b,c]$ for which $f^2 \mid a$ (and $f \mid b$) and then applying the substitution $x \mapsto x/f$. In the code, we do it in prime steps, since that makes finding a suitable representative a little easier.

This can be used, among other things, for determining the kernel of the surjection $\mathrm{cl}(f^2D) \twoheadrightarrow \mathrm{cl}(D)$.

### DIFF
--- a/src/sage/quadratic_forms/bqf_class_group.py
+++ b/src/sage/quadratic_forms/bqf_class_group.py
@@ -82,11 +82,15 @@ from sage.misc.cachefunc import cached_method
 from sage.structure.parent import Parent
 from sage.structure.unique_representation import UniqueRepresentation
 from sage.structure.element import AdditiveGroupElement
+from sage.categories.morphism import Morphism
 
 from sage.misc.prandom import randrange
 from sage.rings.integer_ring import ZZ
+from sage.rings.finite_rings.integer_mod_ring import Zmod
 from sage.rings.finite_rings.integer_mod import Mod
+from sage.rings.polynomial.polynomial_ring import polygen
 from sage.arith.misc import random_prime
+from sage.matrix.constructor import matrix
 from sage.groups.generic import order_from_multiple, multiple
 from sage.groups.additive_abelian.additive_abelian_wrapper import AdditiveAbelianGroupWrapper
 from sage.quadratic_forms.binary_qf import BinaryQF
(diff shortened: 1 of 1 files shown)


src/sage/quadratic_forms/bqf_class_group.py
```diff
@@ -330,6 +334,34 @@ def gens(self):
...
+            Class of 333*x^2 + 306*x*y + 422*y^2
+            sage: proj(elt)
+            Class of 9*x^2 + 4*x*y + 9*y^2
+        """
+        if not isinstance(other, BQFClassGroup):
+            return None
+        try:
+            proj = BQFClassGroupQuotientMorphism(other, self)
+        except (TypeError, ValueError):
+            return None
```
tscrim:
```suggestion
            return super()._coerce_map_from_()
        try:
            proj = BQFClassGroupQuotientMorphism(other, self)
        except (TypeError, ValueError):
            return super()._coerce_map_from_()
```
This is better in case some super class decides to implement more general stuff. (Unlikely, but one never knows...)

src/sage/quadratic_forms/bqf_class_group.py
```diff
@@ -611,3 +643,178 @@ def order(self):
...
+        sage: f1 = BinaryQF([109, 92, 113])
+        sage: f2 = _project_bqf(f1, 101); f2
+        53*x^2 - 152*x*y + 109*y^2
+        sage: f1.discriminant().factor()
+        -1 * 2^2 * 101^2
+        sage: f2.discriminant().factor()
+        -1 * 2^2
+
+    ALGORITHM: Find a class representative with `q^2 \mid a`
+    (and `q\mid b`) and substitute `x\mapsto x/q`.
```
tscrim:
```suggestion
    ALGORITHM:

    Find a class representative with `q^2 \divides a`
    (and `q \divides b`) and substitute `x\mapsto x/q`.
```
For usual latex `\divides` is better (operator and clearly differentiates between a set separator), but I am not sure for mathjax.

Can we also move this before the `EXAMPLES::`?

src/sage/quadratic_forms/bqf_class_group.py
```diff
@@ -611,3 +643,178 @@ def order(self):
...
+
+        EXAMPLES::
+
+            sage: from sage.quadratic_forms.bqf_class_group import BQFClassGroupQuotientMorphism
+            sage: G = BQFClassGroup(-4*117117)
+            sage: H = BQFClassGroup(-4*77)
+            sage: BQFClassGroupQuotientMorphism(G, H)
+            Generic morphism:
+              From: Form Class Group of Discriminant -468468
+              To:   Form Class Group of Discriminant -308
```
tscrim: A `TestSuite(f).run(skip="_test_category")` would be better. It fails the category test because we don't have a good way yet to have parents with multiple element classes (much less ones that are created separate from the parent). Yet, the rest should pass IIRC.

src/sage/quadratic_forms/bqf_class_group.py
```diff
@@ -611,3 +643,178 @@ def order(self):
...
+        """
+        if not isinstance(G, BQFClassGroup):
+            try:
+                disc = ZZ(G)
+            except Exception:
+                raise TypeError('G needs to be a BQFClassGroup')
```
tscrim: This error message is not quite correct as it could also be data used to construct a `BFQClassGroup`. Also it would be good to have a check for at least some part (if not all) of these code paths.

On the other hand, this should follow the standard element-parent model with the user always constructing the elements (morphisms) through a parent (homset) or by code. So I would consider removing this path altogether.

tscrim (changes requested):

src/sage/quadratic_forms/bqf_class_group.py
yyyyx4: According to a quick test, `\divides` seems undefined in MathJax.

tscrim: Thanks. That's what I was thinking, but I appreciate you checking and confirming.

tscrim (approved):
Thanks. LGTM.

yyyyx4:
Thank you!

PR #37076, opened 2024-01-17
src/sage/tests/gap_packages.py: remove

orlitzky:
These tests for the installed GAP packages are problematic in a world where Sage can use GAP from the system. Particularly, loading _all_ installed GAP packages is not a safe operation when there might be an enormous number of them.

Moreover, with the system GAP, it's not really Sage's reponsibility to ensure that the installed packages load. It _would_ be nice to know that any extra packages from the gap_packages SPKG are installed correctly; however, that type of test would be more appropriate in that SPKG's spkg-check phase.

Dependencies:

- https://github.com/sagemath/sage/pull/37049


orlitzky:
On top of the other PR, just removes the whole problematic file.

@tobiasdiez I also dropped the "known failure" entry from your CI thingy.

antonio-rojas (approved):

vbraun:
merge conflict

orlitzky:
rebased onto develop

PR #37078, opened 2024-01-17
use parent in weyl_algebra.py

fchapoton:
Using `Parent` instead of `Algebra`

also full pep8 cleanup of the modified file, plus some typing annotations

### :memo: Checklist


tscrim (approved):
LGTM.

PR #37079, opened 2024-01-17
use Parent in Monsky-Washnitzer

fchapoton:
use `Parent` in the modified file

also removing `__neg__` method

and some other details

### :memo: Checklist

### DIFF
--- a/src/sage/schemes/hyperelliptic_curves/monsky_washnitzer.py
+++ b/src/sage/schemes/hyperelliptic_curves/monsky_washnitzer.py
@@ -48,6 +48,7 @@ AUTHORS:
 
 from sage.arith.misc import integer_ceil as ceil
 from sage.arith.misc import binomial
+from sage.categories.algebras import Algebras
 from sage.functions.log import log
 from sage.matrix.constructor import matrix
 from sage.misc.cachefunc import cached_method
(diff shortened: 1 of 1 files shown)


fchapoton:
The `TestSuite` is not there, as it fails on pickling..

src/sage/schemes/hyperelliptic_curves/monsky_washnitzer.py
```diff
@@ -513,12 +522,12 @@ def __init__(self, Q, laurent_series=False):
...
                                   "coefficient ring (=%s) of Q" % base_ring)
 
-        # CommutativeAlgebra.__init__ tries to establish a coercion
+        # .__init__ tries to establish a coercion
         # from the base ring, by github issue #9138. The corresponding
         # hom set is cached.  In order to use self as cache key, its
         # string representation is used. In otder to get the string
         # representation, we need to know the attributes _a and
-        # _b. Hence, in #9138, we have to move CommutativeAlgebra.__init__
+        # _b. Hence, in #9138, we have to move .__init__
```
tscrim: This is somewhat of a strange comment. It would be good to fix this here by just making this class a `UniqueRepresentation`. If you want to leave the comment for now, it would be better to say `Parent.__init__` here.

src/sage/schemes/hyperelliptic_curves/monsky_washnitzer.py
```diff
@@ -2385,12 +2396,12 @@ def __init__(self, Q, R=None, invert_y=True):
...
             R = Q.base_ring()
 
-        # Github issue #9138: CommutativeAlgebra.__init__ must not be
+        # Github issue #9138: .__init__ must not be
         # done so early.  It tries to register a coercion, but that
         # requires the hash being available.  But the hash, in its
         # default implementation, relies on the string representation,
         # which is not available at this point.
-        # CommutativeAlgebra.__init__(self, R)  # moved to below.
+        # .__init__(self, R)  # moved to below.
```
tscrim: This comment does not apply because of the `UniqueRepresentation` (the hash is the id); so we can remove it.

src/sage/schemes/hyperelliptic_curves/monsky_washnitzer.py
```diff
@@ -2428,10 +2439,10 @@ def __init__(self, Q, R=None, invert_y=True):
         self._series_ring_y = self._series_ring.gen(0)
         self._series_ring_0 = self._series_ring.zero()
 
-        # Github issue #9138: Initialise the commutative algebra here!
+        # Github issue #9138: Initialise the parent here!
         # Below, we do self(self._poly_ring.gen(0)), which requires
         # the initialisation being finished.
```
tscrim: Similarly we can remove this comment.

tscrim (changes requested):
In addition to the comments below, can you explain a bit about why you removed the `__neg__`? The generic implementation seems like it would be slower (I don't know how much that would matter though).

Should we also try and fix the pickling issue to? Could you say at least a bit more about the failure? In particular, on which class(es)?

fchapoton:
ok, so I have added back `__neg__` and it works.

Sorry for the unanswered mail. I will do that now.

tscrim:
> ok, so I have added back `__neg__` and it works.

That is just as strange. Well, I am happy that it is working as I would have expected it to. Thank you. This is now a positive review.

> Sorry for the unanswered mail. I will do that now.

No problem. Thank you. I wasn't sure if it ended up in a spam folder or was some outdated email (Japanese universities seem to enjoy changing email addresses every few years in some cases...).

tscrim (approved):

PR #37081, opened 2024-01-17
convenience method to construct elliptic-curve isomorphisms from u,r,s,t

yyyyx4:
The `EllipticCurve_generic` class has methods for computing an isomorphism between two given curves as well as computing the codomain of a particular isomorphism (specified by its coefficients $u,r,s,t$).

What it does not currently have is a method for computing an isomorphism from its (co)domain and the coefficients $u,r,s,t$. The best way to construct such an isomorphism is currently to `from sage.schemes.elliptic_curves.weierstrass_morphism import WeierstrassIsomorphism`, which is less than convenient. In this patch we add the desired method.

#sd123


tscrim (changes requested):
You should also say something about the default values of `rst`. (Or perhaps this is obvious to an expert?) Otherwise LGTM.

yyyyx4:
Thanks, done. (I'd say zero is definitely the single obvious choice for $r,s,t$, but it can't hurt to state it explicitly anyway.)

tscrim (approved):
Thank you. LGTM.

PR #37082, opened 2024-01-17
Fix NumberFieldElement_quadratic.is_integral() when D is not square-free

### LINKED ISSUE #34800: wrong result for .is_integral() in quadratic field
<div id="comment:0"></div>

```sage
sage: K.<t> = QuadraticField(-10007^2)
sage: a = t / 10007
sage: a.minpoly()
x^2 + 1
sage: a.minpoly() in ZZ['x']
True
sage: a.is_integral()
False
```

Component: **number fields**

_Issue created by migration from https://trac.sagemath.org/ticket/34800_

pjbruin:
Fixes #34800. The current implementation of is_integral() assumes that D is square-free. We implement the check for integrality of the trace and norm in a way that works for general D.

### DIFF
--- a/src/sage/rings/number_field/number_field_element_quadratic.pyx
+++ b/src/sage/rings/number_field/number_field_element_quadratic.pyx
@@ -2137,19 +2137,57 @@ cdef class NumberFieldElement_quadratic(NumberFieldElement_absolute):
             True
             sage: for _ in range(20):
             ....:     assert O.random_element().is_integral()
+
+        Check that :issue:`34800` is fixed::
+
+            sage: K.<t> = QuadraticField(-10007^2)
+            sage: (t/10007).is_integral()
+            True
         """
+        cdef mpz_t m, n, q, r, s, t, u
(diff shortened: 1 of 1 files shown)


tscrim:
Does it make sense to keep the old test for the square-free case (which, of course, would include a check in the method for that)?

pjbruin:
> Does it make sense to keep the old test for the square-free case (which, of course, would include a check in the method for that)?

The problem with that is that knowing whether `D` is square-free requires factoring `D`, which we should certainly avoid. In my implementation, we only need to check one specific candidate for a square divisor, namely `denom` or `u = denom/2` depending on a parity condition.

tscrim (approved):
Thank you for the explanation. I wasn’t sure if this was already tested somewhere during the initialization (or could have been cached by some other computation) and if the new test would have been (relatively) slower than the old in the case it was already known.

Anyways, positive review.

pjbruin:
Thank you!

PR #37065, opened 2024-01-16
Installation guide: Use `.. envvar::` to define environment variables

mkoeppe:
This turns references ``:envvar:`....` `` into hyperlinks to the place of the definition. https://deploy-preview-37065--sagemath.netlify.app/html/en/installation/source#environment-variables

Also improving the structure of the section, and some smaller updates.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37056, opened 2024-01-14
Installation guide: Show system package commands using sphinx-inline-tabs

mkoeppe:
Preview: https://deploy-preview-37056--sagemath.netlify.app/html/en/installation/source#linux-system-package-installation

Same for the SPKG pages in the reference manual. https://deploy-preview-37056--sagemath.netlify.app/html/en/reference/spkg/gfortran#equivalent-system-packages

Same for the git instructions in the developer guide, using 2 tabs for HTTPS and SSH: https://deploy-preview-37056--sagemath.netlify.app/html/en/developer/github#forking-the-sage-repository

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36957 (merged here)

PR #37083, opened 2024-01-18
HTML documentation: Show preparsed doctests using inline tabs

### LINKED ISSUE #35791: Auto-translate doctests to pure Python, with import statements
I think we would be able to generate `import` statements from the `# optional` (or `# needs` - #35779) annotations.

The necessary cross-referencing information is going to be in sage.features.sagemath.
See:
- #35749 (needs review)
(doc preview: https://deploy-preview-35749--sagemath-tobias.netlify.app/reference/misc/sage/features/sagemath.html#sage.features.sagemath.all_features)

> This would also have the huge advantage to support users that want to use sage as a normal python package without loading the sage environment.

+1. 

I could imagine that the code blocks in the formatted HTML could offers tabs for different views, as is done for example in https://setuptools.pypa.io/en/latest/userguide/package_discovery.html: The example in monolithic Sage (goes through preparser, no imports), and the example in plain Python (some prettified version of the preparsed example + imports). This would also address @williamstein's comment (https://groups.google.com/g/sage-devel/c/utA0N1it0Eo/m/79aD4-RQAAAJ)

_Originally posted by @mkoeppe in https://github.com/sagemath/sage/issues/35750#issuecomment-1594840729_

mkoeppe:
Every doctest block is decorated with tabs, by default showing the original doctest in **Sage** syntax, and offering a preparsed, pure **Python** version of it. The tabs are synchronized across the page.

If @kwankyu's live documentation is being built, it is offered as another tab **Sage (live)**. When this tab is selected, it automatically starts the Thebe/Binder; it is not necessary to find and push the "Make live" button.

tabs-anim

[Preview](https://deploy-preview-37083--sagemath.netlify.app/html/en/reference/function_fields/sage/rings/function_field/function_field_rational)

Fixes #35791

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37056
- Depends on #37065


.ci/merge-fixes.sh
```diff
@@ -67,7 +68,7 @@ for REPO in ${SAGE_CI_FIXES_FROM_REPOSITORIES:-sagemath/sage}; do
                         git am --signoff --show-current-patch=diff
                         echo "--------------------------------------------------------------------8<-----------------------------"
                         echo "::endgroup::"
-                        echo "Failure applying $PULL_URL as a patch, resetting"
+                        echo "Failure applying $PULL_SHORT as a patch, resetting"
```
tobiasdiez, line 71: I like the full URL since it easy to open the corresponding PR. Either way, these changes are unrelated to the actual goal of the PR and thus should be moved to somewhere else.

build/pkgs/sphinx_inline_tabs/type
```diff
@@ -0,0 +1 @@
+standard
```
tobiasdiez: Adding a new standard package needs a period where it is optional and then a vote on the mailing list first, right?

src/sage_docbuild/__main__.py
```diff
@@ -288,6 +288,9 @@ def setup_parser():
     standard.add_argument("--no-plot", dest="no_plot",
                           action="store_true",
                           help="do not include graphics auto-generated using the '.. plot' markup")
+    standard.add_argument("--no-preparsed-examples", dest="no_preparsed_examples",
```
tobiasdiez, line 292: What's the use case for this option?

tobiasdiez (changes requested):

.ci/merge-fixes.sh
mkoeppe: The full URL appears just above. The improvement is that the repeated one is shortened so that there is not a second, duplicate link.

I'll not open a separate PR for this.

build/pkgs/sphinx_inline_tabs/type
mkoeppe: We don't do that for trivialities.

src/sage_docbuild/__main__.py
mkoeppe: It's faster and some people may not want to see the tabs.

tobiasdiez: Speedwise it doesn't seem to make a change (by looking at the github workflow execution time). Thus, to reduce complexity I prefer to just remove the option here and always build the docs with the parsed python tab.

.ci/merge-fixes.sh
tobiasdiez: Why not? This is clearly not connected to main theme of this PR.

src/sage_docbuild/__main__.py
mkoeppe: Noted that this is your preference. I won't remove it.

.ci/merge-fixes.sh
mkoeppe: You've just finished the review of this one line change, no?

tobiasdiez: No, and there are also more changes here that seem unrelated.

src/sage_docbuild/__main__.py
tobiasdiez: Why do you want to make this feature configurable?

mkoeppe: Asked and answered already.

.ci/merge-fixes.sh
mkoeppe: You commented, I responded to your comment. What's missing for the review of this one line change.

mkoeppe: The rules of review, spelled out for you: 
- Provide the necessary information. 
- Do not use vagueness as a tool.
- Do not ask for information that has already been provided.

kwankyu:
There are overlapped tabs like

Screen Shot 2024-02-06 at 1 40 37 PM
Screen Shot 2024-02-06 at 1 40 48 PM

This is from

```rst
Sage can handle very large numbers, even numbers with millions or billions of
digits.

::

    sage: factorial(100)
    93326215443944152681699238856266700490715968264381621468592963895217599993229915608941463976156518286253697920827223758251185210916864000000000000000000000000

::

    sage: n = factorial(1000000)  # about 1 second
    sage: len(n.digits())
    5565709

This computes at least 100 digits of :math:`\pi`.  
```

It would be better to separate tabs for the two blocks of examples. Is this possible?

mkoeppe:
Thanks for finding this bug. I'll investigate

src/sage_docbuild/__main__.py
tobiasdiez: Not really. I don't think we should start to make various features of the documentation configurable (except for the live docs that need additional setup non of the features can be toggled off), it just bloats the code and increases the maintenance effort without any real benefits. So please remove this.

mkoeppe: You're repeating yourself.

kwankyu:
I tested with live documentation. Very nice.

One suggestion. How about "Live" instead of "Sage (live)"? Short and easy to read, and without parentheses. 

Sample view:

Screen Shot 2024-02-07 at 11 46 21 PM

mkoeppe:
I think I prefer "Sage (live)" for its clearer semantics ("same language (but a variant)") and extensibility (we can add "Python (live)" later). But I also do have to admit that I generally like parentheses, which is probably related to the amount of time I spent with Lisp in the past~

src/sage_docbuild/__main__.py
mkoeppe: More rules for review
- Check the facts about an unfamiliar domain before making claims about it

Specifically here: Any developer who has worked with the Sage docbuild system knows that we have long had a range of options that configure global aspects of it. There's `SAGE_SKIP_PLOT_DIRECTIVE` (which you see even in the diff of the present PR...), `SAGE_DOC_UNDERSCORE`, `SAGE_USE_CDNS`, and we used to have `SAGE_DOC_MATHJAX`, `SAGE_DOC_JSMATH` until recently.

kwankyu:
> I think I prefer "Sage (live)" for its clearer semantics ("same language (but a variant)") 

OK. This is just a matter of personal preference, where the author's preference takes precedence. 

> and extensibility (we can add "Python (live)" later). 

That is interesting, though I cannot imagine it now.

> But I also do have to admit that I generally like parentheses, which is probably related to the amount of time I spent with Lisp in the past~

I don't like parentheses, as they add branches (obstructions) to the (linear) flow of sentences.

I think I learned it from Knuth's advices on writing: one of the last things he does after writing a sentence is to remove unnecessary parentheses from it.

src/sage_docbuild/__main__.py
tobiasdiez: In each of these cases there is a good reason for why the switch exists beyond "some people may not want to" have xyz.

mkoeppe: Just a moment ago you denied that the switches even exist.
Now you say they exist but they have a better reason to exist.

This is called "moving the goalposts", a well-known standard technique.

tobiasdiez: The goalpost stayed the same: we should not add config options to toggle default documentation features off just because "some people may not want to see [xyz]".

mkoeppe: Yes, you're repeating yourself. 

Just because you have this opinion does not mean that this PR needs work.

kwankyu:
Building doc with
```
export SAGE_PREPARSED_DOC=no
export SAGE_LIVE_DOC=yes
```
seems to activate preparsed doc.

kwankyu:
How to turn off tabs?

mkoeppe:
> It seems that
> 
> ```
> export SAGE_PREPARSED_DOC=no
> export SAGE_LIVE_DOC=yes
> ```
> 
> turns on tabs.

Yes, that's right. Would you prefer the tabless version for this case generally, or as an option?

kwankyu:
I am not sure...

What is the use of the preparsed doc? I guess that primarily advanced users of distribution packages may find it useful. I can imagine that a beginner may become aware of the preparsing mechanism and see how sage syntax is different from python syntax, which is an essential knowledge for effective use of sage.

Users in the middle ground may find tabs (with preparsed doc) just take spaces for information of little use. Having tabs for each example that synchronize anyway seems a waste of space. 

I feel that somehow we should have configuration option or switch on pages to support the users in the middle ground. And whatever option or switch we implement should get along with the configuration option and switch for live doc. I am not sure how we could do this without making things too complicated...

mkoeppe:
> What is the use of the preparsed doc? I guess that primarily advanced users of distribution packages may find it useful. I can imagine that a beginner may become aware of the preparsing mechanism and see how sage syntax is different from python syntax, which is an essential knowledge for effective use of sage.

Yes, all of that. But it also helps anyone who is already an experienced Python programmer and comes across Sage. It makes it obvious and immediately testable that Sage is actual Python and usable from Python as a library, not some kind of quasi-Python, Python-like, Python-inspired kind of specialty language, or Python with training wheels.

> Users in the middle ground may find tabs (with preparsed doc) just take spaces for information of little use. Having tabs for each example that synchronize anyway seems a waste of space.

Yes, the vertical space is precious, which is exactly why I made it disablable. 

> I feel that somehow we should have configuration option or switch on pages to support the users in the middle ground. And whatever option or switch we implement should get along with the configuration option and switch for live doc. I am not sure how we could do this without making things too complicated...

A persistent switch (is the light/dark selector persistent? using cookies?) for users who only want to see one version and save the vertical space could be an interesting follow-up. But that needs to be done by someone other than me; I stopped following web technologies around 1997;)

kwankyu:
I think I can suggest these now:

- Only two tabs ("original" and "preparsed").  Keep the live doc feature orthogonal to the tabs feature. 
- Instead of `--no-preparsed-examples`, use `--preparsed-examples` to turn on the tabs feature. That is, the default is no tabs. It may get an argument `parsed` and `preparsed`  to choose the default for the tabs.
- Tabs do not synchronize (vertically). Let users be able to switch to preparsed (or original) version only for a particular tab.

kwankyu:
> .. I stopped following web technologies around 1997;)

Nowadays, AI helps humans to catch up :-)

mkoeppe:
> * Tabs do not synchronize (vertically). Let users be able to switch to preparsed (or original) version only for a particular tab.

Why would that possibly be an improvement?

mkoeppe:
> 1. Only two tabs ("original" and "preparsed").  Keep the live doc feature orthogonal to the tabs feature.

Why would that be better? I much prefer to click the middle tab instead of having to chase down that shy "Make live" button that lives in hiding near the window frame

kwankyu:
> > * Tabs do not synchronize (vertically). Let users be able to switch to preparsed (or original) version only for a particular tab.
> 
> Why would that possibly be an improvement?

I assumed all tabs are already in either "parsed" or "preparsed" by default as the user want. Then the user would want to switch to the other version temporarily for a particular tab.

But right... in my proposal, it is not handy for the user to choose the the default, unless to make a full doc build.

OK. I am not strong about this.

kwankyu:
> > 1. Only two tabs ("original" and "preparsed").  Keep the live doc feature orthogonal to the tabs feature.
> 
> Why would that be better? I much prefer to click the middle tab instead of having to chase down that shy "Make live" button that lives in hiding near the window frame

By having the two features orthogonal, we don't need to add an additional tab for "live preparsed".

But I agree that the button is not so visible...

kwankyu (approved):
OK.

Now I am convinced that having live tab is better than inserting live cell below.

kwankyu:
By the way, live doc seems broken (as seen in [the doc preview](https://deploy-livedoc--sagemath.netlify.app/html/en/reference/plotting/sage/plot/arrow#sage.plot.arrow.arrow))...

mkoeppe:
> By the way, live doc seems broken

Are we not pinning the versions of the components there?

kwankyu:
We are using https://unpkg.com/thebe@latest/lib/index.js via patched jupyter-sphinx.

kwankyu:
Fixing in https://github.com/sagemath/sage/pull/37637

kwankyu:
You may want these refinements:
```diff
diff --git a/src/doc/common/static/custom-jupyter-sphinx.css b/src/doc/common/static/custom-jupyter-sphinx.css
index a68a5cb05aa..e79b47b4539 100644
--- a/src/doc/common/static/custom-jupyter-sphinx.css
+++ b/src/doc/common/static/custom-jupyter-sphinx.css
@@ -1,5 +1,5 @@
 div.jupyter_container {
-  margin: .5rem 0;
+  border: 0;
 }
 
 div.jupyter_container + div.jupyter_container {
diff --git a/src/sage_docbuild/conf.py b/src/sage_docbuild/conf.py
index 145105bb814..1c705c68d1f 100644
--- a/src/sage_docbuild/conf.py
+++ b/src/sage_docbuild/conf.py
@@ -903,7 +903,7 @@ class SagecodeTransform(SphinxTransform):
                             linenostart=1)
                         cell_node += cell_input
                         container = TabContainer("", type="tab", new_set=False)
-                        textnodes = [Text('Sage (live)')]
+                        textnodes = [Text('Sage Live')]
                         label = Label("", "", *textnodes)
                         container += label
                         content = Container("", is_div=True, classes=["tab-content"])
```
"Sage Live"  or "Sage live" (no parentheses please).

(before) (see the double border)
Screen Shot 2024-03-20 at 7 41 29 PM
(after)
Screen Shot 2024-03-20 at 7 21 52 PM

mkoeppe:
Thank you, committed as dea37e7.

PR #37087, opened 2024-01-18
use parent in Ore polynomials

fchapoton:
this is using `Parent` in Ore polynomial and fractions

also some typing annotations and other little details in the 2 modified files.

also moving a method to the category of rings (could go higher in the hierarchy maybe)

### :memo: Checklist


xcaruso:
Looks good to me!

xcaruso (approved):

PR #37088, opened 2024-01-18
Remove deprecated methods from QuaternionFractionalIdeal_rational

pjbruin:
This removes two methods of `QuaternionFractionalIdeal_rational` deprecated in #31583 (because of ambiguous behaviour; both could return either the left or the right order):
- `quaternion_order()`; users should call `left_order()` or `right_order()` as appropriate
- `ring()`; calling `ring()` still works but calls the method from the parent class, which returns the ambient quaternion algebra.


pjbruin:
#sd123

vbraun:
merge conflict

PR #37089, opened 2024-01-18
Implemented order computation and identity checking for automorphisms of elliptic curves

r98inver:
The WeierstrassIsomorphism class is also used to represent automorphisms of elliptic curves. Sometimes one is looking for a particular automorphism (say, of order 3 or 4), and this check is simplified with the .order() method for elliptic-curve automorphisms.
The is_identity() method is added for convenience.

#sd123

### DIFF
--- a/src/sage/schemes/elliptic_curves/weierstrass_morphism.py
+++ b/src/sage/schemes/elliptic_curves/weierstrass_morphism.py
@@ -1003,9 +1003,9 @@ class WeierstrassIsomorphism(EllipticCurveHom, baseWI):
         r"""
         Compute the order of this Weierstrass isomorphism if it is an automorphism.
 
-        A ``ValueError`` is raised if the domain is not equal to the codomain.
+        A :class:`ValueError` is raised if the domain is not equal to the codomain.
 
-        A ``NotImplementedError`` is raised if the order of the automorphism is not 1, 2, 3, 4 or 6.
+        A :class:`NotImplementedError` is raised if the order of the automorphism is not 1, 2, 3, 4 or 6.
 
         EXAMPLES::
 
--- a/src/sage/schemes/elliptic_curves/weierstrass_morphism.py
+++ b/src/sage/schemes/elliptic_curves/weierstrass_morphism.py
@@ -973,6 +973,105 @@ class WeierstrassIsomorphism(EllipticCurveHom, baseWI):
         """
         return Integer(1)
 
+    def is_identity(self):
+        r"""
+        Returns if the Weierstrass isomorphims is the identity.
+
+        EXAMPLES::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import *
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, None, E)
+            sage: ws.is_identity()
+            False
+
+        ::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import *
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, (1, 0, 0, 0), None)
+            sage: ws.is_identity()
+            True
+        """
+        return self.tuple() == (1, 0, 0, 0)
+
+    def order(self):
+        r"""
+        If the Weierstrass isomorphism is an automorphism, computes the order of the automorphism.
+
+        If the domain is not equal to the codomain a `ValueError` is returned.
+
+        If the order of the automorphism is not 2,3,4 or 6 a `NotImplementedError` is returned.
+
+        EXAMPLES::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import *
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, None, E)
+            sage: ws.order()
+            2
+
+        TESTS::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import *
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, None, E)
+            sage: ws.order()
+            2
+            sage: E1 = EllipticCurve(Fp, [1, 69])
+            sage: ws = E.isomorphism_to(E1)
+            sage: ws.order()
+            Traceback (most recent call last):
+            ...
+            ValueError: the domain is different from the codomain
+
+        ::
+
+            sage: E = EllipticCurve_from_j(Fp(0))
+            sage: ws = WeierstrassIsomorphism(E, (Fp(36), 0, 0, 0), None)
+            sage: ws.order()
+            6
+            sage: ws2 = ws*ws
+            sage: ws2.order()
+            3
+            sage: F2_bar = GF(2).algebraic_closure()
+            sage: E = EllipticCurve_from_j(F2_bar(0))
+            sage: ws = WeierstrassIsomorphism(E, None, E)
+            sage: ws.order()
+            3
+        """
+        # Check if it is an actual endomorphism
+        if self._domain != self._codomain:
+            raise ValueError("the domain is different from the codomain")
+
+        if self.is_identity():
+            return Integer(1)
+
+        ws2 = WeierstrassIsomorphism._composition_impl(self, self)
+        if ws2.is_identity():
+            return Integer(2)
+
+        ws3 = WeierstrassIsomorphism._composition_impl(self, ws2)
+        if ws3.is_identity():
+            return Integer(3)
+
+        ws4 = WeierstrassIsomorphism._composition_impl(ws2, ws2)
+        if ws4.is_identity():
+            return Integer(4)
+
+        ws6 = WeierstrassIsomorphism._composition_impl(ws2, ws4)
+        if ws6.is_identity():
+            return Integer(6)
+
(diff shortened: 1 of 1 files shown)


grhkm21:
(Please correct me if I'm wrong!) If the elliptic curve is not over a field, e.g. over `Zmod(35)`, and when the automorphism has order $3$ when restricted to over $E / \mathbb{F}_5$ and $4$ when restricted to over $E / \mathbb{F}_7$, then it will have order $12$ over $E / \mathbb{F}_{35}$. This can be a doctest

grhkm21:
Just for completeness sake, here's an update. Currently it can't be a doctest since `WeierstrassIsomorphism` assumes the base ring is a field, which is not really necessary. However, this demonstrates the idea:
```python
sage: E = EllipticCurve(Zmod(13 * 19), [19, 13])
....: assert E.change_ring(GF(13)).j_invariant() == 1728 # % 13
....: assert E.change_ring(GF(19)).j_invariant() == 0
....: 
....: # E.automorphisms() # this fails
....: aut_13 = E.change_ring(GF(13)).automorphisms()
....: aut_19 = E.change_ring(GF(19)).automorphisms()
....: 
....: # Since it's an isomorphism, only the *u* matters
....: import itertools
....: for phi_13, phi_19 in itertools.product(aut_13, aut_19):
....:     u = Mod(CRT([ZZ(phi_13.u), ZZ(phi_19.u)], [13, 19]), 13 * 19)
....:     order = u.multiplicative_order()
....:     if order > 6:
....:         print(phi_13, phi_19, order, sep="\n")
....:         break
....: 
Elliptic-curve endomorphism of Elliptic Curve defined by y^2 = x^3 + 6*x over Finite Field of size 13
  Via:  (u,r,s,t) = (5, 0, 0, 0)
Elliptic-curve endomorphism of Elliptic Curve defined by y^2 = x^3 + 13 over Finite Field of size 19
  Via:  (u,r,s,t) = (7, 0, 0, 0)
12
```

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
         """
         return Integer(1)
 
+    def is_identity(self):
+        r"""
+        Returns if the Weierstrass isomorphims is the identity.
```
grhkm21: Docstrings start with "Return" (no s)

grhkm21: Also typo. I recommend rewriting as "Check if this Weierstrass isomorphism is the identity." ("check" and "return" are common verbs in Sage docs)

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+
+        EXAMPLES::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import *
```
grhkm21: Don't `import *` please!

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, None, E)
+            sage: ws.is_identity()
+            False
+
+        ::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import *
```
grhkm21: see above

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, (1, 0, 0, 0), None)
+            sage: ws.is_identity()
+            True
+        """
+        return self.tuple() == (1, 0, 0, 0)
+
+    def order(self):
+        r"""
+        If the Weierstrass isomorphism is an automorphism, computes the order of the automorphism.
```
grhkm21: As above, it should start with a verb. Something like "Compute the order of this Weierstrass isomorphism if it is an automorphism. A `ValueError` is raised otherwise."

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+
+        If the domain is not equal to the codomain a `ValueError` is returned.
```
grhkm21: In docstrings, single backtick stands for dollar signs for $\LaTeX$. To write code blocks, use double backticks "a \`\`ValueError`` is ..."

You can see the current documentations [here](https://deploy-preview-37089--sagemath.netlify.app/html/en/reference/arithmetic_curves/sage/schemes/elliptic_curves/weierstrass_morphism#sage.schemes.elliptic_curves.weierstrass_morphism.WeierstrassIsomorphism.order), which includes the error :)

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+
+        ws4 = WeierstrassIsomorphism._composition_impl(ws2, ws2)
+        if ws4.is_identity():
+            return Integer(4)
+
+        ws6 = WeierstrassIsomorphism._composition_impl(ws2, ws4)
+        if ws6.is_identity():
+            return Integer(6)
+
+        raise NotImplementedError("the order of the endomorphism is not 2,3,4 or 6")
```
grhkm21: I guess you are missing `1` in the error string?

grhkm21 (changes requested):
Thanks for contributing #sd123 :rocket:

styling issues

grhkm21:
Hi, is there any update on this? :)

r98inver:
Forgot to push, should be ok now

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+            sage: ws.is_identity()
+            True
+        """
+        return self.tuple() == (1, 0, 0, 0)
+
+    def order(self):
+        r"""
+        Compute the order of this Weierstrass isomorphism if it is an automorphism.
+
+        A ``ValueError`` is raised if the domain is not equal to the codomain.
```
grhkm21:
```suggestion
        A :class:`ValueError` is raised if the domain is not equal to the codomain.
```

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+
+        A ``NotImplementedError`` is raised if the order of the automorphism is not 1, 2, 3, 4 or 6.
```
grhkm21:
```suggestion
        A :class:`NotImplementedError` is raised if the order of the automorphism is not 1, 2, 3, 4 or 6.
```

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+        r"""
+        Check if this Weierstrass isomorphism is the identity.
+
+        EXAMPLES::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import WeierstrassIsomorphism
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, None, E)
```
grhkm21, line 986:
```suggestion
            sage: ws = E.isomorphism_to(E); ws
            # add the output here
```
I think this will make it more clear, since it might be *slightly* confusing why the isomorphism from $E$ to $E$ is not the identity.

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
```diff
@@ -973,6 +973,105 @@ def inseparable_degree(self):
...
+            sage: ws.is_identity()
+            False
+
+        ::
+
+            sage: from sage.schemes.elliptic_curves.weierstrass_morphism import WeierstrassIsomorphism
+            sage: p = 97
+            sage: Fp = GF(p)
+            sage: E = EllipticCurve(Fp, [1, 28])
+            sage: ws = WeierstrassIsomorphism(E, (1, 0, 0, 0), None)
```
grhkm21, line 996:
```suggestion
            sage: ws = E.isomorphism(1, 0, 0, 0)
```
Do you mind changing the constructors to use `.isomorphism` or `.isomorphism_to` for the rest as well?

grhkm21 (changes requested):
Very sorry for not spotting these earlier. After these changes it should be ready.

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
r98inver: This I don't get. My purpose was to show that it could happen

grhkm21: Ohh right, nevermind then

src/sage/schemes/elliptic_curves/weierstrass_morphism.py
r98inver: For this and also the above one: I think that since this is a "helper" class (i.e. people usually don't/shouldn't call it directly) it is better to put direct tests here. `.isomorphism` and `.isomorphism_to` are already tested and showcased elsewhere

grhkm21 (approved):
Looks good

r98inver:
Thanks for the review and the suggestions!

grhkm21:
No problem! Great work for a first contribution :D

PR #37092, opened 2024-01-18
Adding methods .x() and .y() for EllipticCurvePoint

r98inver:
Added the more natural notation P.x() and P.y() to access respectively the x and y coordinates of an EllipticCurvePoint. The behaviour mimics P.xy() and aims to replace P.xy()[0] and P.xy()[1]. 
Multiple instances of P.xy()[0] and P.xy()[1] are also replaced in related modules.

#sd123

Done with @gioella

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_point.py
+++ b/src/sage/schemes/elliptic_curves/ell_point.py
@@ -778,6 +778,52 @@ class EllipticCurvePoint_field(SchemeMorphism_point_abelian_variety_field):
         else:
             return self[0]/self[2], self[1]/self[2]
 
+    def x(self):
+        """
+        Return the `x` coordinate of this point, as an element of the base field.
+        If this is the point at infinity, a :class:`ZeroDivisionError` is raised.
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve('389a')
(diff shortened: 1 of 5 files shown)


src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -778,6 +778,52 @@ def xy(self):
...
+            sage: P.x()
+            -1
+            sage: Q = E(0); Q
+            (0 : 1 : 0)
+            sage: Q.x()
+            Traceback (most recent call last):
+            ...
+            ZeroDivisionError: rational division by zero
+        """
+        if self[2] == 1:
```
yyyyx4: The call `a.is_one()` is quite a lot faster than the test `a == 1`: The reason is that the latter spends time on converting the Python integer `1` into the parent of `a`.

r98inver: Fixed. I also modified the method `.xy()` both to keep them equal and for the same speed reason

yyyyx4 (approved):
Looks good to me. Thank you!

(The tests seem to be failing for unrelated reasons.)

r98inver:
Thanks!

PR #36190, opened 2023-09-04
establish interface for instantiated classical modular polynomials

yyyyx4:
Generic modular polynomials (i.e., as elements of $\mathbb Z[X,Y]$) grow very big very quickly. However, most algorithmic applications only require evaluations $\Phi_\ell(j,Y)$ of those polynomials. There are algorithms to compute such evaluations faster than computing-then-evaluating the full modular polynomial, for example due to [Sutherland](https://arxiv.org/abs/1202.3985). PARI implements both kinds of algorithms in `polmodular()`.

This patch adds a common interface for (1) accessing Kohel's database of modular polynomials if available, (2) calling PARI's `polmodular()` to compute generic modular polynomials, and (3) calling PARI's `polmodular()` to compute instantiated modular polynomials. Results are opportunistically cached and reused whenever this seems to make sense.

PR #37094, opened 2024-01-18
Fix bug with pari interface in `classical_modular_polynomial`

drago-96:
There is a bug in the new `classical_modular_polynomial` function created in https://github.com/sagemath/sage/pull/36190.

A simple example to reproduce the bug is the following

```
p = 2^216 * 3^137 - 1
F.<i> = GF(p^2, modulus=[1,0,1])
E = EllipticCurve(F, [0, 6, 0, 1, 0])
classical_modular_polynomial(2, E.j_invariant())
```

This will fail with a `TypeError`.

The bug is due to the interface with the pari function `polmodular`. In particular, contrary to the [documentation](http://pari.math.u-bordeaux.fr/dochtml/ref/Polynomials_and_power_series.html#polmodular), the `polmodular` function will only evaluate $\Phi_\ell(j, Y)$ for $j$-invariants that are defined over $\mathbb F_p$, and not over any generic finite field.
If however `j.parent()` is a generic finite field, but `j` itself is defined over the prime field, pari will evaluate that and the current implementation fails to convert back to a sage polynomial.

The proposed fix is to cast the result of the pari call to `ZZ['Y']`, since the result of `polmodular` currently returns a polynomial in the correct $\mathbb Z/n\mathbb Z[Y]$.

#sd123

### DIFF
--- a/src/sage/schemes/elliptic_curves/mod_poly.py
+++ b/src/sage/schemes/elliptic_curves/mod_poly.py
@@ -85,21 +85,23 @@ def classical_modular_polynomial(l, j=None):
         sage: j = GF(q).random_element()
         sage: l = random_prime(50)
         sage: Y = polygen(parent(j), 'Y')
-        sage: classical_modular_polynomial(l,j) == classical_modular_polynomial(l)(j,Y)
+        sage: classical_modular_polynomial(l, j) == classical_modular_polynomial(l)(j, Y)
         True
         sage: p = 2^216 * 3^137 - 1
-        sage: F.<i> = GF(p^2, modulus=[1,0,1])
+        sage: F.<i> = GF((p,2), modulus=[1,0,1])
         sage: l = random_prime(50)
         sage: j = F.random_element()
         sage: Y = polygen(parent(j), 'Y')
-        sage: classical_modular_polynomial(l,j) == classical_modular_polynomial(l)(j,Y)
+        sage: classical_modular_polynomial(l, j) == classical_modular_polynomial(l)(j, Y)
         True
         sage: E = EllipticCurve(F, [0, 6, 0, 1, 0])
         sage: j = E.j_invariant()
-        sage: classical_modular_polynomial(l,j) == classical_modular_polynomial(l)(j,Y)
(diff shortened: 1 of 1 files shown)


src/sage/schemes/elliptic_curves/mod_poly.py
```diff
@@ -87,6 +87,21 @@ def classical_modular_polynomial(l, j=None):
         sage: Y = polygen(parent(j), 'Y')
         sage: classical_modular_polynomial(l,j) == classical_modular_polynomial(l)(j,Y)
         True
+        sage: p = 2^216 * 3^137 - 1
+        sage: F.<i> = GF(p^2, modulus=[1,0,1])
+        sage: l = random_prime(50)
```
fchapoton, line 92: you could use the syntax `GF((p,2),...)` to avoid factoring.

drago-96: Thanks, I didn't know this constructor!

pjbruin:
Shouldn't the result be converted into a polynomial over **Z**/_p_**Z** where _p_ is the characteristic of the finite field?

grhkm21:
> Shouldn't the result be converted into a polynomial over **Z**/_p_**Z** where _p_ is the characteristic of the finite field?

Why? I think the PR is fine, it returns a polynomial over `parent(j)`, which is what you want normally anyways (to compute its roots later etc.).

pjbruin:
You are right. I was puzzled by the conversion to **Z**[Y], and did not notice that changing Z to Z/pZ would not make a difference, due to the final conversion to R.

The fact that `polmodular` returns a polynomial over a different field than the one in which _j_ lies does seem to be a possible PARI bug and should probably also be reported there.

pjbruin (approved):
Likely a PARI bug, the patch looks like a good workaround to me

pjbruin:
Now fixed in PARI: https://pari.math.u-bordeaux.fr/cgi-bin/bugreport.cgi?bug=2533

drago-96:
Thanks for the reporting and fixing in PARI!

PR #37096, opened 2024-01-18
Fix `.multiplication_by_m` when m is not coprime to characteristic (#6413)

### LINKED ISSUE #6413: Retrieving the multiplication-by-p isogeny on elliptic curves over fields of characteristic p fails.
<div id="comment:0"></div>

```
sage: p = 11
sage: E = EllipticCurve(GF(p), [1,1])
sage: E.multiplication_by_m(p)
ZeroDivisionError                         Traceback (most recent call last)

/Users/hlaw/.sage/temp/resid_tg082.upc.es/80890/_Users_hlaw__sage_init_sage_0.py in <module>()

/Users/hlaw/src/sage-4.0.1/local/lib/python2.5/site-packages/sage/schemes/elliptic_curves/ell_generic.pyc in multiplication_by_m(self, m, x_only)
   2133         #  and hence 2*my+a1*mx+a3 = (1/m)*(2*y+a1*x+a3)*d(mx)/dx
   2134 
-> 2135         my = ((2*y+a1*x+a3)*mx.derivative(x)/m - a1*mx-a3)/2
   2136 
   2137         return mx, my

/Users/hlaw/src/sage-4.0.1/local/lib/python2.5/site-packages/sage/structure/element.so in sage.structure.element.RingElement.__div__ (sage/structure/element.c:10361)()

/Users/hlaw/src/sage-4.0.1/local/lib/python2.5/site-packages/sage/structure/coerce.so in sage.structure.coerce.CoercionModel_cache_maps.bin_op (sage/structure/coerce.c:6105)()

/Users/hlaw/src/sage-4.0.1/local/lib/python2.5/site-packages/sage/structure/element.so in sage.structure.element.RingElement.__div__ (sage/structure/element.c:10342)()

/Users/hlaw/src/sage-4.0.1/local/lib/python2.5/site-packages/sage/rings/fraction_field_element.so in sage.rings.fraction_field_element.FractionFieldElement._div_ (sage/rings/fraction_field_element.c:5805)()

/Users/hlaw/src/sage-4.0.1/local/lib/python2.5/site-packages/sage/rings/fraction_field_element.so in sage.rings.fraction_field_element.FractionFieldElement.__init__ (sage/rings/fraction_field_element.c:1954)()

ZeroDivisionError: fraction field element division by zero
```
One can see from the line that causes the error that the calculation of the *y*-component of the isogeny involves a division by *m*, which in this case is the characteristic of the ground field.

CC:  @defeo @jpflori @yyyyx4

Component: **elliptic curves**

_Issue created by migration from https://trac.sagemath.org/ticket/6413_

grhkm21:
*Almost* fixes #6413 - doesn't work for function fields

Compute pth multiplication coordinate maps by using isogenies.

The multivariate substitution step is VERY slow, and I don't know how to fix it. I think the reason might also be the elements live in the fraction field.

Thanks @yyyyx4 for the idea! #sd123 🚀


tscrim:
@yyyyx4 Can you be reviewing this (and other such related PRs from sd123)?

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2352,126 +2389,51 @@ def multiplication_by_m(self, m, x_only=False):
...
-        r"""
-        Return the ``EllipticCurveIsogeny`` object associated to the
-        multiplication-by-`m` map on this elliptic curve.
-
-        The resulting isogeny will
-        have the associated rational maps (i.e., those returned by
-        :meth:`multiplication_by_m`) already computed.
+        if v_p > 0:
+            frob = self.frobenius_isogeny()
+            isog = (frob.dual() * frob)**v_p
```
yyyyx4: Is there any reason not to call the (cached) `.multiplication_by_p_isogeny()` here?

yyyyx4:
> @yyyyx4 Can you be reviewing this (and other such related PRs from sd123)?

I wasn't sure if I'm "allowed" to review many of these since I suggested quite a few of them as possible #sd123 projects myself (and in some cases even contributed to the patch with ideas and advice). Is there a policy about this?

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2019,6 +2018,30 @@ def division_polynomial(self, m, x=None, two_torsion_multiplicity=2, force_evalu
 
     torsion_polynomial = division_polynomial
 
+    @cached_method
+    def multiplication_by_p_isogeny(self):
```
yyyyx4: This currently rejects infinite base fields: Should it be moved to `EllipticCurve_finite_field`? Alternatively, have you checked if this works for infinite fields of positive characteristic?

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2352,126 +2389,51 @@ def multiplication_by_m(self, m, x_only=False):
...
         if m == -1:
             if not x_only:
                 return (x, -y-a1*x-a3)
             else:
                 return x
 
-        # the x-coordinate does not depend on the sign of m.  The work
+        # If we only require the x coordinate, it is faster to use the recursive formula
+        # since substituting polynomials is quite slow.
+        p = Integer(self.base_ring().characteristic())
```
yyyyx4: I think `.characteristic()` should already return an `Integer` in all cases: If not, that's a bug.

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2352,126 +2389,51 @@ def multiplication_by_m(self, m, x_only=False):
...
+        v_p = 0 if p == 0 else valuation(m.abs(), p)
```
yyyyx4: The `valuation()` is invariant under multiplication by units, so taking the `.abs()` is unnecessary.

tscrim:
> > @yyyyx4 Can you be reviewing this (and other such related PRs from sd123)?
> 
> I wasn't sure if I'm "allowed" to review many of these since I suggested quite a few of them as possible #sd123 projects myself (and in some cases even contributed to the patch with ideas and advice). Is there a policy about this?

You are definitely allowed to review this. It can even be 2 authors who are the same reviewers but they cross-check each other. Although with the new GH system, this has become more troublesome to do in practice. I can still do a final double-check if you prefer too.

src/sage/schemes/elliptic_curves/ell_generic.py
grhkm21: Nope, it doesn't work for function fields. I will mark that as a TODO. I guess the 15-year-old issue isn't resolved yet...

grhkm21:
Thanks for the review and sorry for the mess, I didn't really think through the code and just pushed what worked for the first example :P

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -610,6 +610,24 @@ def cardinality(self, algorithm=None, extension_degree=1):
 
     order = cardinality  # alias
 
+    @cached_method
+    def multiplication_by_p_isogeny(self):
+        r"""
+        Return the multiplication-by-\(p\) isogeny.
```
yyyyx4: Looking at the [documentation preview](https://deploy-preview-37096--sagemath.netlify.app/html/en/reference/arithmetic_curves/sage/schemes/elliptic_curves/ell_finite_field#sage.schemes.elliptic_curves.ell_finite_field.EllipticCurve_finite_field.multiplication_by_p_isogeny), this syntax appears to be unsupported.

```suggestion
        Return the multiplication-by-`p` isogeny.
```

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2421,15 +2420,53 @@ def multiplication_by_m(self, m, x_only=False):
...
+            sage: f, g = E.multiplication_by_m(k)
+            sage: for _ in range(100):
+            ....:     P = E.random_point()
+            ....:     if P * k == 0:
+            ....:         continue
+            ....:     Qx = f.subs(x=P[0])
+            ....:     Qy = g.subs(x=P[0], y=P[1])
+            ....:     assert (P * k).xy() == (Qx, Qy)
+
+        However, it still fails for elliptic curves over positive characteristics fields::
```
yyyyx4:
```suggestion
        However, it still fails for elliptic curves over positive-characteristic fields::
```

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2421,15 +2420,53 @@ def multiplication_by_m(self, m, x_only=False):
...
             raise ValueError("m must be a non-zero integer")
 
         if x_only:
             x = polygen(self.base_ring(), 'x')
         else:
+            from sage.rings.finite_rings.finite_field_base import FiniteField as FiniteField_generic
+            if m % p == 0 and not isinstance(self.base_ring(), FiniteField_generic):
+                # TODO: Implement the correct formula?
+                raise NotImplementedError("multiplication by integer not coprime to p"
+                                          "is only implemented for curves over finite fields")
```
yyyyx4:
```suggestion
                raise NotImplementedError("multiplication by integer not coprime to p "
                                          "is only implemented for curves over finite fields")
```

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2440,126 +2477,42 @@ def multiplication_by_m(self, m, x_only=False):
...
-        #  and hence 2*my+a1*mx+a3 = (1/m)*(2*y+a1*x+a3)*d(mx)/dx
-
+        # Consideration of the invariant differential
+        # w=dx/(2*y+a1*x+a3) shows that m*w = d(mx)/(2*my+a1*mx+a3)
+        # and hence 2*my+a1*mx+a3 = (1/m)*(2*y+a1*x+a3)*d(mx)/dx
         my = ((2*y+a1*x+a3)*mx.derivative(x)/m - a1*mx-a3)/2
 
-        return mx, my
-
-    def multiplication_by_m_isogeny(self, m):
```
yyyyx4, line 2469: I acknowledge that we are technically "allowed" to remove this method now, but I'm wondering if it's a good idea at this point: Debian and Ubuntu are still on Sage 9.5, so most of their users have never even seen the deprecation warning. Since keeping this method in Sage a little longer is essentially free of harm, I'd suggest to do so.

grhkm21:
I've left the deprecated `multiplication_by_m_isogeny`. Hopefully the package installers update their versions soon, ha

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2421,15 +2420,53 @@ def multiplication_by_m(self, m, x_only=False):
...
+        p = self.base_ring().characteristic()
+
         if m == 0:
             raise ValueError("m must be a non-zero integer")
 
         if x_only:
             x = polygen(self.base_ring(), 'x')
         else:
+            from sage.rings.finite_rings.finite_field_base import FiniteField as FiniteField_generic
+            if m % p == 0 and not isinstance(self.base_ring(), FiniteField_generic):
```
yyyyx4: Plenty of doctests are failing in this line when $p=0$.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -610,6 +610,24 @@ def cardinality(self, algorithm=None, extension_degree=1):
...
+    @cached_method
+    def multiplication_by_p_isogeny(self):
+        r"""
+        Return the multiplication-by-`p` isogeny.
+
+        EXAMPLES::
+
+            sage: p = 23
+            sage: K.<a> = GF(p^3)
+            sage: E = EllipticCurve(K, [K.random_element(), K.random_element()])
```
yyyyx4: This will throw an `ArithmeticError` whenever the discriminant of the elliptic curve happens to be zero; same issue in some more places below. (Sorry for not catching this earlier...)

yyyyx4 (approved):
Looks good to me, thanks!

src/sage/schemes/elliptic_curves/ell_generic.py
```diff
@@ -2421,15 +2420,53 @@ def multiplication_by_m(self, m, x_only=False):
...
+            ...
+            NotImplementedError: multiplication by integer not coprime to p is only implemented for curves over finite fields
+
+        ::
+
+            sage: p = 7
+            sage: K.<a> = GF(p^2)
+            sage: E = EllipticCurve(K, [K.random_element(), K.random_element()])
+            sage: E.multiplication_by_m(p * 2)[0] == E.multiplication_by_m(p * 2, x_only=True)
+            True
```
tornaria, line 2453: This test fails at random with probability ~ 1/49 or something like that (i.e. when the discriminant is 0).

grhkm21: I will fix this in a separate PR since this is merged already. Sorry for not spotting this earlier (there's one similar above)

PR #37098, opened 2024-01-18
Fixed `.random_element()` for fractional ideals in rational quaternion algebras

S17A05:
Added `QuaternionFractionalIdeal_rational.random_element()` to fix random sampling in fractional ideals of quaternion algebras.

The method `.random_element()` of the ideal class refers to the ring over which the ideal is defined, but for fractional ideals Sage often (incorrectly) gives the base field of the algebra as a base ring. To work around this, we implement a new function `QuaternionFractionalIdeal_rational.random_element()` that hardcodes $\mathbb{Z}$ to return a random $\mathbb{Z}$-linear combination of the (lattice) generators of the fractional ideal.

Note: The deeper issue of separating ideals and fractional ideals still remains - see #3680.

#sd123

### DIFF
--- a/src/sage/algebras/quatalg/quaternion_algebra.py
+++ b/src/sage/algebras/quatalg/quaternion_algebra.py
@@ -2295,6 +2295,20 @@ class QuaternionFractionalIdeal_rational(QuaternionFractionalIdeal):
         """
         return 'Fractional ideal %s' % (self.gens(),)
 
+    def random_element(self, *args, **kwds):
+        """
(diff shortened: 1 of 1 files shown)


S17A05:
Removed a wrongly placed blank line of code, which probably caused Lint to fail.

S17A05:
Removed the whitespaces in blank lines detected by Lint.

pjbruin (approved):
Looks good to me!

PR #34976, opened 2023-02-06
compute isomorphisms between quaternion orders

yyyyx4:
Fixes #34861.

PR #37100, opened 2024-01-19
Quaternion Algebra Fractional Ideal improvements - equivalence and reduced bases

### LINKED ISSUE #37080: Ideal equivalence in indefinite quaternion algebras
### Steps To Reproduce

Define an indefinite quaternion algebra. Define two equivalent right ideals `I` and `J`. Then `I.is_equivalent(J)` throws an error.
```python
B.<i,j,k> = QuaternionAlgebra(-2, 5)
O0 = B.maximal_order()
I = O0.unit_ideal()
I.is_equivalent(I)
```

### Expected Behavior

Should return `True`.

### Actual Behavior

Throws error `PariError: domain error in minim0: form is not positive definite`

### Additional Information

The `is_equivalent` function calls `I.theta_series_vector` and a theta series only makes sense in a definite quaternion algebra when the norm form of the ideal is positive definite. In turn it calls Pari's `qfrep` which does a positive definite check, and throws the error. An alternative method to test equivalence would be to implement Minkowski reduced basis, and look for the smallest norm element of `J.conjugate() * I`

@S17A05

### Environment

```markdown
- **OS**: Windows 10 (WSL)
- **Sage Version**: 10.3.beta4
```

### Checklist

### LINKED ISSUE #37337: `isomorphism_to` of `quaternion_algebra.py` does not find an isomorphism of isomorphic maximal orders
### Steps To Reproduce

```
p = 419
B.<i,j,k> = QuaternionAlgebra(p)
O = B.quaternion_order([1/2 + 3/2*j + k, 1/18*i + 25/9*j + 5/6*k, 3*j + 2*k, 3*k])
Oconj = j.inverse() * O * j
Oconj = B.quaternion_order(Oconj.basis())
O.isomorphism_to(Oconj)
```

### Expected Behavior

The method should return an isomorphism between the maximal orders O and Oconj.

### Actual Behavior

`ValueError: quaternion orders not isomorphic`

### Additional Information

Originally discovered by @jtcc2, I'm only posting the issue here. Further discussion in https://github.com/sagemath/sage/pull/34976#issuecomment-1937007360.

### Environment

```markdown
- **OS**: Windows 11 (WSL Ubuntu 22.04)
- **Sage Version**: 10.3beta8
```

### Checklist

jtcc2:
Includes the following changes
1. Added `.is_definite()` to quaternion algebras, which performs a check on the invariants.
2. Added `.reduced_basis()` to quaternion fractional ideals which uses LLL to find the basis of shortest elements
3. Modified quaternion fractional ideal `.is_equivalent` to include a `side` parameter which can be set to `'left'` or `'right'` to  do left-ideal or right-ideal equivalence, and a `certificate` parameter which if true, returns the element a such that I = J*a. Throws a `NotImplementedError` for indefinite quaternion algebras which fixes #37080 . Backwards compatible.
4. Also added `.is_left_equivalent` and `.is_right_equivalent`.
5. Added `.is_principal` to quaternion fractional ideals of rank 4 in definite quaternion algebras. Includes `certificate` parameter for finding generator. Finds the shortest norm element in the lattice and checks its norm.
6. Fixes #37337, bug in quaternion order `.isomorphism_to` where it cannot find the isomorphism between $\mathcal{O}$ and $\alpha^{-1} \mathcal{O} \alpha$ if $nrd(\alpha)$ is a ramified prime times a square. Fixed this by repeating the previous `isomorphism_to` method multiple times on inputs $\mathcal{O}$ and $\alpha^{-1} \mathcal{O} \alpha$ with $\alpha = 1, i, j, k$. Throws a `NotImplementedError` in cases not covered.

@S17A05 @Jonathke

#sd123


S17A05:
@yyyyx4 Can you have a look at this since you worked on the closely related #34976?

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2477,6 +2500,43 @@ def basis_matrix(self):
...
+            sage: I = O*alpha + O*3089622859
+            sage: I.reduced_basis()[0]
+            1/2*i + j + 5/2*k
+        """
+        if len(self.basis()) < 4:
+            raise ValueError("basis must have rank 4")
+
+        if not self.quaternion_algebra().is_definite():
+            if not self.quadratic_form().is_positive_definite():
+                raise TypeError("The quaternion algebra must be definite")
```
yyyyx4: Aren't these two tests equivalent? Were they duplicated by accident?

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+        - ``B`` -- a bound to compute and compare theta series before
+          doing the full equivalence test
+
+        - ``certificate`` -- if ``True`` returns an element alpha such that alpha*J=I
+
+        OUTPUT: bool, or (bool, alpha) if ``certificate`` is ``True``
         """
-        # shorthand: let I be self
-        if not isinstance(self, QuaternionFractionalIdeal_rational):
+        if not isinstance(J, QuaternionFractionalIdeal_rational):
```
yyyyx4, line 3293: I'd say in this case we should probably throw a `TypeError`. Returning `False` is arguably correct, but I can't see any situation in which users will rely on this behaviour wilfully, hence failing visibly is better.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2845,9 +2905,11 @@ def multiply_by_conjugate(self, J):
         R = self.quaternion_algebra()
         return R.ideal(basis, check=False)
 
-    def is_equivalent(self, J, B=10) -> bool:
-        """
-        Return ``True`` if ``self`` and ``J`` are equivalent as right ideals.
+    def is_equivalent(self, J, B=10, certificate=False, side=None):
```
yyyyx4, line 3185: Personally I would simply deprecate this method in favor of the two new explicitly named one-sided methods: Having it default to `.is_right_equivalent()` is arbitrary and ambiguous.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+                return False, None
             return False
 
         if self.right_order() != J.right_order():
-            raise ValueError("self and J must be right ideals")
+            raise ValueError("self and J must be right ideals over the same order")
+
+        if not self.quaternion_algebra().is_definite():
+            raise NotImplementedError('equivalence test of ideals not implemented'
+                                    'for indefinite quaternion algebras')
```
yyyyx4:
```suggestion
            raise NotImplementedError('equivalence test of ideals not implemented'
                                    ' for indefinite quaternion algebras')
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+
+        c = self.theta_series_vector(2)[1]
+        if not certificate:
+            return c != 0
+        if certificate and c == 0:
+            return False, None
+
+        # Use Pari's qfminim with flag = 1 to find alpha with N(alpha) = N(I)
+        _,v = self.quadratic_form().__pari__().qfminim(None, None, 1)
+        return True, sum(ZZ(c)*g for c,g in zip(v, self.basis()))
```
yyyyx4: Same here re: the `.minimal_element()` method from #34976.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+            sage: I = alpha * O
+            sage: I.is_principal(True)
+            (True, 1 + 59*i - 1/2*j - 154/9*k)
+        """
+        if not self.quaternion_algebra().is_definite():
+            raise NotImplementedError('principality test not'
+                                    'implemented in indefinite'
+                                    'quaternion algebras')
+
+        if len(self.basis()) < 4:
```
yyyyx4: Can this fail?

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+            sage: O = B.quaternion_order([1/2 + 3/2*j, 1/6*i + 2/3*j + 1/2*k, 3*j, k])
+            sage: alpha = -1 - 59*i + 1/2*j + 154/9*k
+            sage: I = alpha * O
+            sage: I.is_principal(True)
+            (True, 1 + 59*i - 1/2*j - 154/9*k)
+        """
+        if not self.quaternion_algebra().is_definite():
+            raise NotImplementedError('principality test not'
+                                    'implemented in indefinite'
+                                    'quaternion algebras')
```
yyyyx4:
```suggestion
            raise NotImplementedError('principality test not'
                                    ' implemented in indefinite'
                                    ' quaternion algebras')
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+            if certificate:
+                return False, None
+            return False
+
+        if not certificate:
+            return True
+
+        # Use Pari's qfminim with flag = 1 to find alpha as above
+        _,v = IJbar.quadratic_form().__pari__().qfminim(None, None, 1)
+        alpha = sum(ZZ(c)*g for c,g in zip(v, IJbar.basis()))
```
yyyyx4: This one could be `alpha = IJbar.minimal_element()` once #34976 is merged, but pulling that in as a dependency seems a bit overkill just to deduplicate these two lines. Maybe add a TODO so that eventually someone can replace it later?

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -2870,30 +2938,142 @@ def is_equivalent(self, J, B=10) -> bool:
...
+
+        # As explained by Pizer, we now only need to rescale alpha by 1/N(J)
+        return True, alpha * (1/J.norm())
+
+    def is_principal(self, certificate=False):
```
yyyyx4, line 3322: This method duplicates quite some code from the equivalence tests. Perhaps the equivalence tests could/should just end up calling `.is_principal()` on `IJbar` instead of repeating all the steps on that special case?

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Yes, these two should be equivalent - I believe the latter test was originally put in place for the method to also work for "lattices" of non-maximal rank, until it came up that `.gram_matrix()` needs a full rank lattice anyway. I'll remove the second test.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: So should I add a deprecation warning for now to set it up for future removal, or should I remove it immediately?

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: On faulty input, I believe, since calling `B.ideal([i,j])` for a quaternion algebra `B` gives a "fractional ideal", but it really is just a two-dimensional sublattice with basis `(i,j)`.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Since it might be a good idea to test definiteness with this PR's `.is_definite()`-method as well, it might actually be a good idea to put a dependency on this PR to yours (see also [this review comment](https://github.com/sagemath/sage/pull/34976#discussion_r1476246835)).

src/sage/algebras/quatalg/quaternion_algebra.py
yyyyx4: According to The Rules, we're only allowed to remove things once they've been deprecated for one year, so it has to be the former: Add a deprecation warning. (If you agree, that is: I don't have a strong opinion about this either way.)

src/sage/algebras/quatalg/quaternion_algebra.py
yyyyx4: Interesting, I was expecting the constructor for fractional ideals to check the rank. That should probably be fixed eventually, but it doesn't necessarily have to be done here and now.

S17A05:
@yyyyx4 Once #34976 is merged I will modify the definiteness check in your code and will integrate your `.minimal_element()`-method into our code, and then I'll re-request a review; for now I'll just leave this PR labelled as "needs work".

S17A05:
Since beta 7 of Sage 10.3 was just released, I have instead decided to merge #34976 into this branch. In the progress of figuring this out, however, I cause some chaos in the PR; in particular, GitHub now thinks that a lot of files in `.devcontainer` have been modified, even though they should be the same files - if someone knows how to fix this please let me know, so that I can repair this.
Otherwise this should be ready for a second review, so I'll update the labels.

fchapoton:
it was not a good idea to merge with the latest develop. One better way if you really want that is to use github's `update branch` button that does the merge in the correct direction.

To fix the merge, you could try to rebase your branch on top of develop using `git rebase -i develop`, but this may trigger conflicts.

S17A05:
> it was not a good idea to merge with the latest develop. One better way if you really want that is to use github's `update branch` button that does the merge in the correct direction.
> 
> To fix the merge, you could try to rebase your branch on top of develop using `git rebase -i develop`, but this may trigger conflicts.

That's how I did it originally, but then I wrongly assumed that I had to revert it to merge with the other PR's branch, and after that I couldn't do it this way again.

Thanks for the suggestion, I'll try to do that.

S17A05:
I rebased the branch on upstream/develop, but it still lists the `.devcontainer`-files under the made changes.

yyyyx4:
The merge still seems broken. I've attempted to fix it [on a new branch](https://github.com/yyyyx4/sage/tree/quatfracidl) by cherry-picking all non-merge commits from here on top of the current `develop` branch.

You can get my version by adding my fork as a remote and overwriting your version of this branch with mine:
```
git remote add yyyyx4 https://github.com/yyyyx4/sage.git
git fetch yyyyx4 quatfracidl:this-branch-name -f
```
After that, you should be able to `git push -f` the new version of the branch to this pull request as usual.

S17A05:
> The merge still seems broken. I've attempted to fix it [on a new branch](https://github.com/yyyyx4/sage/tree/quatfracidl) by cherry-picking all non-merge commits from here on top of the current `develop` branch.
> 
> You can get my version by adding my fork as a remote and overwriting your version of this branch with mine:
> 
> ```
> git remote add yyyyx4 https://github.com/yyyyx4/sage.git
> git fetch yyyyx4 quatfracidl:this-branch-name -f
> ```
> 
> After that, you should be able to `git push -f` the new version of the branch to this pull request as usual.

Perfect, that worked - thanks for the help!

yyyyx4 (approved):
Looks good to me now, thanks!

PR #37102, opened 2024-01-19
fix random doctest failure in generic has_order() function

yyyyx4:
In this simple patch we fix [a random test failure](https://github.com/sagemath/sage/actions/runs/7577410381/job/20638145383#step:10:4436) in the `has_order()` function for generic groups.

#sd123


tscrim (changes requested):
Can we also have an explicit test for this? Otherwise LGTM.

tscrim (approved):
Thanks. LGTM.

PR #37104, opened 2024-01-19
do not use xargs -P 0 - not all systems allow 0

dimpase:
E.g. on OpenBSD this fails, they need something >0


dimpase:
@orlitzky - could you review this?

orlitzky:
Sure

PR #37106, opened 2024-01-19
Additive identity of EllipticCurve should always be minimum

grhkm21:
This is inspired by how `.automorphisms()` always return the identity morphism as the first automorphism. So now `min(E.points()) == 0` always holds (when `E.points()` is nonempty and finite).

#sd123 🚀


src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -308,6 +308,7 @@ def __init__(self, curve, v, check=True):
                 raise TypeError("Coordinates %s do not define a point on %s" % (list(v), curve))
 
         SchemeMorphism_point_abelian_variety_field.__init__(self, point_homset, v, check=False)
+        self._zxy_coords = (self._coords[2], self._coords[0], self._coords[1])
```
yyyyx4: Creating a second copy of all the coordinates seems a bit wasteful. I'd suggest constructing this tuple on the fly in `._richcmp_()`.

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -386,13 +387,19 @@ def _richcmp_(self, other, op):
...
+            sage: min(E.points()) == 0
+            True
         """
         if not isinstance(other, EllipticCurvePoint_field):
             try:
                 other = self.codomain().ambient_space()(other)
             except TypeError:
                 return NotImplemented
-        return richcmp(self._coords, other._coords, op)
+        return richcmp(self._zxy_coords, other._zxy_coords, op)
```
yyyyx4: If you do construct the $(z,x,y)$ tuple on the fly, it would make sense to have a faster path for equality testing (where $(x,y,z)$ is of course fine).

src/sage/schemes/elliptic_curves/saturation.py
```diff
@@ -383,9 +383,54 @@ def p_saturation(self, Plist, p, sieve=True):
...
+             --> [(6 : 42 : 1), (0 : 1 : 0)]
+             --rank is now 1
+            E has 2-torsion over Finite Field of size 73, projecting points
+             --> [(15 : 62 : 1), (0 : 1 : 0)]
+             --rank is now 1
+            ! rank same for 15 steps, checking kernel...
+            kernel vectors: [
+            (0, 1)
+            ]
+            points generating kernel: [(0 : 1 : 0)]
```
yyyyx4: Are we sure that this output is always the same? I'd assume there's a reason it was `...` previously?

grhkm21: Sorry, I think it's not. I used `sage --fixdoctests` to perform this automatically, sorry for that!

PR #37107, opened 2024-01-19
Added `.is_isomorphic` for rational quaternion algebras

S17A05:
Added `QuaternionAlgebra_ab.is_isomorphic()` to conveniently check whether two rational quaternion algebras are isomorphic.

TODO (future work):
- [] Extend isomorphism check to quaternion algebras over number fields.

#sd123

### DIFF
--- a/src/sage/algebras/quatalg/quaternion_algebra.py
+++ b/src/sage/algebras/quatalg/quaternion_algebra.py
@@ -1091,6 +1091,9 @@ class QuaternionAlgebra_ab(QuaternionAlgebra_abstract):
             sage: A == B
             False
         """
+        if not isinstance(A, QuaternionAlgebra_ab):
+            raise TypeError("A must be a quaternion algebra of the form (a,b)_K")
+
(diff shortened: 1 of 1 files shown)


pjbruin:
Could you add a type check to verify that `A` is a quaternion algebra?

S17A05:
> Could you add a type check to verify that `A` is a quaternion algebra?

Added a check to make sure $A$ is a quaternion algebra of the form $A = (a,b)_K$ since `.discriminant()` depends on this type of quaternion algebra representation - thanks for bringing this up!

S17A05:
> > Could you add a type check to verify that `A` is a quaternion algebra?
> 
> Added a check to make sure A is a quaternion algebra of the form A=(a,b)K since `.discriminant()` depends on this type of quaternion algebra representation - thanks for bringing this up!

Modified the error message to exclude the word _rational_ since the base field of both quaternion algebras is only checked afterwards.

pjbruin (approved):
Looks good to me.

PR #37110, opened 2024-01-19
Precompute orders of twists of EllipticCurve

grhkm21:
Let `E` be an EllipticCurve. If its order is known, then so is the order of its quadratic twist, as given by $$|E| + |E^t| = 2p + 2$$. This can also be done for the quartic twists and sextic twists. This PR sets the order of the returned elliptic curve.

For example, the second call to `.order` below is faster:
```python
E = EllipticCurve(GF(2**255 - 19), [3, 5])
assert E.order() == 57896044618658097711785492504343953926540591650149308728870074001824607516706
Et = E.quadratic_twist()
print(Et.order()) # fast
```

TODO List:
- [ ] Precompute orders of quartic twists
- [ ] Precompute orders of sextic twists
- [ ] Write documentations
- [ ] Make twists deterministic

#sd123

### DIFF
--- a/src/doc/en/reference/references/index.rst
+++ b/src/doc/en/reference/references/index.rst
@@ -5463,6 +5463,10 @@ REFERENCES:
              imaginary quadratic fields. Invent. Math. 103 (1991),
              no. 1, 25--68.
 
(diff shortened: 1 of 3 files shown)


grhkm21:
Wait I'm confused, this PR isn't ready yet since it's draft, but I also see that not all commits are merged. So what is the status of this PR now?

PR #37111, opened 2024-01-19
Add test if quaternion order is maximal

syndrakon:
Add a method to sage.algebras.quatalg.quaternion_algebra.QuaternionOrder which allows to check whether the order is maximal.
This method compares the discriminant of the order to the discriminant of the algebra. According to Voight's book on quaternion algebras (chapter 15), this test is valid in number fields (but not over any field), so the method fails with a notImplementedError in  other cases.

Currently, there is no method to test maximality of orders, so implementing it for quaternion algebras over number fields is some progress.

#sd123


src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1758,9 +1759,48 @@ def discriminant(self):
 
         return (MatrixSpace(QQ, 4, 4)(L)).determinant().sqrt()
 
+    def is_maximal(self):
+        r"""
+        Return ``True`` if the order is maximal in the ambient quaternion algebra, ``False`` otherwise
```
dcoudert: The first line should be short and usually we start such description with `Check whether...`
```diff
-        Return ``True`` if the order is maximal in the ambient quaternion algebra, ``False`` otherwise
-        Only works in quaternion algebras over number fields
+        Check whether the order of ``self`` is maximal in the ambient
+        quaternion algebra.
+
+        Only works in quaternion algebras over number fields
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1758,9 +1759,48 @@ def discriminant(self):
...
+            sage: B = QuaternionAlgebra(GF(13),-1,-11)
+            sage: i,j,k = B.gens()
+            sage: O0_basis = (1,i,j, k)
+            sage: O0 = B.quaternion_order(O0_basis)
+            sage: try:
+            ....:     O0.is_maximal()
+            ....:     print("This should fail")
+            ....: except NotImplementedError:
+            ....:     print("ok")
+            ....:
```
dcoudert: remove this line

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1758,9 +1759,48 @@ def discriminant(self):
...
+            ok
+        """
+        if not(self.quaternion_algebra().base_ring() in NumberFields()):
```
dcoudert: It is better to use "X not in Y". Furthermore, we don't start error messages with capital letters and the dot at the end is not needed.
```diff
-        if not(self.quaternion_algebra().base_ring() in NumberFields()):
-            raise NotImplementedError("Check for maximality is only implemented for quaternion algebras over number fields).")
-        return(self.discriminant() == self.quaternion_algebra().discriminant())
+        if self.quaternion_algebra().base_ring() not in NumberFields():
+            raise NotImplementedError("check for maximality is only implemented for "
+                                      "quaternion algebras over number fields")
+        return self.discriminant() == self.quaternion_algebra().discriminant()
```

dcoudert (changes requested):
a few comments on the coding style

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1758,9 +1759,48 @@ def discriminant(self):
...
+            raise NotImplementedError("Check for maximality is only implemented for quaternion algebras over number fields).")
+        return(self.discriminant() == self.quaternion_algebra().discriminant())
+
     def left_ideal(self, gens, check=True, *, is_basis=False):
         r"""
         Return the left ideal of this order generated by the given generators.
+        Only works in quaternion algebras over number fields.
```
pjbruin: Please add an empty line after the first sentence of the docstring (see dcoudert's comment above).

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1758,9 +1759,48 @@ def discriminant(self):
...
+            sage: O1.is_maximal()
+            False
+
+        TESTS::
+
+            sage: B = QuaternionAlgebra(GF(13),-1,-11)
+            sage: i,j,k = B.gens()
+            sage: O0_basis = (1,i,j, k)
+            sage: O0 = B.quaternion_order(O0_basis)
+            sage: try:
```
pjbruin: The recommended way to write doctests for error messages is
```
sage: O0.is_maximal()
...
NotImplementedError: check for maximality is only implemented for quaternion algebras over number fields
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1758,9 +1759,48 @@ def discriminant(self):
...
+        Only works in quaternion algebras over number fields
+
+        OUTPUT: Boolean
+
+        EXAMPLES::
+
+            sage: p = 11
+            sage: B = QuaternionAlgebra(QQ,-1,-p)
+            sage: i,j,k = B.gens()
+            sage: O0_basis = (1,i,(i+j)/2, (1+i*j)/2)
```
pjbruin: fix spacing (after each comma there should be a space except in some specific cases)

src/sage/algebras/quatalg/quaternion_algebra.py
syndrakon: Without object now since the test was completely rewritten following @pjbruin's remarks.

src/sage/algebras/quatalg/quaternion_algebra.py
syndrakon: I tried, hopefully I did not miss any comma and never met such a specific case.

pjbruin: The main example is in one-element tuples, such as `(1,)`. I also tend to omit spaces when declaring multiple symbolic variables (`var('x,y')`), but I'm not sure if this is generally accepted.

pjbruin (approved):

PR #37114, opened 2024-01-19
Raise errors when Cantor arithmetic will fail due to curve model

### LINKED ISSUE #37109: Arithmetic on Jacobians of Hyperelliptic curves fails when there are two points at infinity
## Description

This is a meta issue made from three encountered during Sage Days 123. It was noticed that there was ambiguous representation and failing group laws in certain scenarios. Issue #37093 noticed ambiguous representation and #37101 highlighted that the even degree model **could not** support the case for the even degree model due to a lack of data in the elements of the Jacobian.

## Expected Behaviour

The proper fix for this is to rewrite the mumford coordinates and the arithmetic to allow for proper handling when there are two points at infinity. The Magma implementation of this follows: https://www.math.auckland.ac.nz/~sgal018/crypto-book/ch10.pdf and I think this is what SageMath should do too.

### Group law fails when the model is even

For the even degree model, there are two points at infinity and the current implementation has no way of distinguishing between $P - (\infty_+)$ and $P - (\infty_-)$.

A proper fix for this would require refactoring how `src/sage/schemes/hyperelliptic_curves/jacobian_morphism.py` currently behaves for the even degree case. For instance some tuple of the form `(u, v, ±)`. Magma handles this with a tuple `(a, b, d)`.

I don't currently have a *good* idea on how to fix this, but I'll try and come up with a reasonable fix to propose. At the moment this issue is really a place holder.

Thanks to @sabrinakunzweiler for bringing this to my attention.

At the moment there is no way to represent both $P - (\infty_+)$ and $P - (\infty_-)$. This means that if we randomly sample divisors on the Jacobian we will have `H.count_points()` elements missing.

```py
sage: F = GF(7)
sage: R.<x> = PolynomialRing(F)
sage: f = x^6 + x + 1
sage: H = HyperellipticCurve(f)
sage: J = H.jacobian()
sage: H.count_points()
[9]
sage: H.zeta_function().numerator()(1) # order of the Jacobian
67
sage: len(set(tuple(J.random_element()) for _ in range(2000)))
58
sage: assert 58 + 9 == 67
```

### Group law fails when degree of h is too large

```python
sage: F = GF(11)
....: R.<x> = PolynomialRing(F)
....: f = 6*x^5 + 6*x^4 + 9*x^3 + 9*x + 8
....: h = 5*x^3 + 3*x^2 + 5*x + 4
....: H = HyperellipticCurve(f, h)
....: J = H.jacobian()
....: D = J(H.lift_x(1))
....: o = H.zeta_function().numerator()(1) # order of Jacobian
....: o * D == J(0)
....: o * D
False
(x^2 + 4*x + 4, y + 4*x + 3)
```

### Mumford coordinates are not uniquely represented

When summing divisors for even degree models, the cantor reduction sometimes does not finish the reduction process. This seems to be due to a TODO left in the code.

```python
sage: F = GF(163)
sage: R.<x> = PolynomialRing(F)
sage: f = x^6 + x^2 + 1
sage: H = HyperellipticCurve(f)
sage: J = H.jacobian()
sage: Ps = H.points()
sage: Ds = [J(P) for P in Ps]
sage: sum(choice(Ds) for _ in range(7))
Returning ambiguous form of degree genus+1.
Returning ambiguous form of degree genus+1.
Returning ambiguous form of degree genus+1.
(x^3 + 11*x^2 + 104*x + 136, y + 95*x^2 + 144*x + 156)
```

### Environment

```markdown
- **OS**: MacOS
- **Sage Version**: 10.2
```

### Checklist

GiacomoPope:
Old code did not check whether the polynomials $f, h$ were of the expected degree for Cantor's algorithm when computing arithmetic with divisors on Jacobians of hyperelliptic curves. Now the code does! This fixes #37109 with minimal changes.

### :memo: Checklist

### DIFF
--- a/src/sage/schemes/hyperelliptic_curves/jacobian_morphism.py
+++ b/src/sage/schemes/hyperelliptic_curves/jacobian_morphism.py
@@ -106,11 +106,11 @@ on the Jacobian of this reduction and the order of the Jacobian is
     True
 """
 
-#*****************************************************************************
+# *****************************************************************************
 #  Copyright (C) 2005 David Kohel <kohel@maths.usyd.edu.au>
 #  Distributed under the terms of the GNU General Public License (GPL)
 #                  http://www.gnu.org/licenses/
-#*****************************************************************************
+# *****************************************************************************
 
 from sage.misc.latex import latex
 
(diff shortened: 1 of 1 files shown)


GiacomoPope:
Although this comment patches over the issue, I would prefer to fix it by rewriting the arithmetic. This *could* be merged, but hopefully I'll have a mathematical fix for this problem. The solution is probably to use https://www.math.auckland.ac.nz/~sgal018/crypto-book/ch10.pdf, which is what (roughly) magma uses.

GiacomoPope:
I think the fix for this needs to be more comprehensive. The real issue is that models with two points at infinity are not properly handled, so we should either fix the arithmetic or error when the model is of the form:

1. Even degree f(x)
2. Degree h(x) > g

However, I'm not totally convinced this is exactly the check which always breaks, so I'm going to close this and try and figure out something better.

PR #37115, opened 2024-01-19
Remove check for deprecated attribute _no_generic_basering_coercion

pjbruin:
In `sage.categories.unital_algebras.UnitalAlgebras`, remove the check for the attribute `_no_generic_basering_coercion`, which was deprecated in #19225. This means that this attribute is now completely ignored by the coercion framework.


mezzarobba (approved):
thanks!

PR #37118, opened 2024-01-20
Fix random polynomial bias

grhkm21:
For a polynomial ring `R` say `GF(11)["x"]`, before `R.random_element` is very far from random, now it is uniformly random.


src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1370,6 +1370,10 @@ def random_element(self, degree=(-1,2), *args, **kwds):
            or a tuple of minimum and maximum degrees. By default set to
            ``(-1,2)``.
 
+        -  ``monic`` - optional boolean to indicate whether the sample
+           polynomial should be monic or not. If degree includes -1, it is still
+           possible for the algorithm to return ``0``, which is not monic.
```
tscrim: This is not quite correct. Indeed, we do not consider `0` to be monic, but it is $1 \cdot 0$, where $0$ is the "basis" element of degree -1. I think you should simply adjust the degree to be nonnegative. It is valid input and the output matches it, provided `is_monic()` is always `True`.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1466,64 @@ def random_element(self, degree=(-1,2), *args, **kwds):
             else:
                 raise ValueError("No polynomial of degree >= 0 has all coefficients zero")
 
-        # Pick a random degree
-        d = randint(degree[0], degree[1])
+        if degree == (-1, -1):
+            return 0
```
tscrim:
```suggestion
        if degree == (-1, -1):
            return self.zero()
```
The Python `int` `0` is not the same as the zero in the polynomial ring; it has different methods.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1466,64 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
+
+        if degree[0] == -1:
+            allow_zero = True
+            degree = (0, degree[1])
+        else:
+            allow_zero = False
+
+        while True:
+            # Pick random coefficients
+            coefs = []
```
tscrim: Rather than constantly appending to a list, it is faster to populate the list with `None` and then modify the entries. Since you know exactly how many elements will be in the list, you can do this.

src/sage/rings/polynomial/polynomial_ring.py
grhkm21: I don’t understand this comment. So what is your suggestion? That if monic is set then degree can’t be -1? (I think it’s commented out below, can’t really check on phone.)

I decided against since for my own use case, I actually want to sample all polynomials with degree -1 to k, and if it is not degree -1, then make it monic. I think in general this use case is common too, and if this is undesirable ie degree -1 shouldn’t be returned, then the user can just write degree=(0,k)

tscrim: Allowing degree -1 is fine, but when `monic=True`, it should never return `0` as that is not a monic polynomial (by Sage's definition/convention). Basically, you should always have `P.random_polynomial(monic=True).is_monic()` return `True` no matter what the degree range is.

Thus within the code, it should simply (silently) have `degree[0]` become `max(0, degree[0])` when `monic=True`.

Note that, e.g., `degree=(-5, 10)` should be considered as valid input IMO (although it currently throws an error). Nothing requires that the result be uniform, much less have full support of the input.

grhkm21: > Allowing degree -1 is fine, but when `monic=True`, it should never return `0` as that is not a monic polynomial (by Sage's definition/convention). Basically, you should always have `P.random_polynomial(monic=True).is_monic()` return `True` no matter what the degree range is.

I have reworded `monic` to `monic_or_zero`. If the user doesn't want zero, they can set `degree[0] = 0`, or call `random_monic_element(zero=False)`. I prefer giving user freedom when they *do* want zero.

> Note that, e.g., `degree=(-5, 10)` should be considered as valid input IMO (although it currently throws an error). Nothing requires that the result be uniform, much less have full support of the input.

I wonder if that'll be more vulnerable to errors, but I'll apply that.

grhkm21:
Actually, I think the API is too confusing (it is to myself right now). I will rewrite it tomorrow. Sorry for the confusion.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,41 @@ def ngens(self):
...
+        - ``_allow_zero`` - optional boolean to indicate whether zero should be
+          included. If ``monic`` is not set, or the degree bound does not
+          include `-1`, then this argument does nothing. Otherwise, the
+          algorithm returns either monic or `0` polynomials (with uniform
+          probability over the range of output). This is for internal use, users
+          should use :meth:`random_monic_or_zero_element`.
+
         -  ``*args, **kwds`` - Passed on to the ``random_element`` method for
-           the base ring
+           the base ring.
```
tscrim:
```suggestion
           the base ring
```
Revert; this is against our doc style.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,41 @@ def ngens(self):
...
+    def random_element(self, degree=(-1, 2), monic=False, _allow_zero=False, *args, **kwds):
         r"""
-        Return a random polynomial of given degree or with given degree bounds.
+        Return a random polynomial of given degree (bounds).
 
         INPUT:
 
         -  ``degree`` - optional integer for fixing the degree
            or a tuple of minimum and maximum degrees. By default set to
            ``(-1,2)``.
```
tscrim:
```suggestion
        -  ``degree`` -- (default: ``(-1, 2)``) integer for fixing the degree
           or a tuple of minimum and maximum degrees
```
Let's make the doc into the correct format while modifying it.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,41 @@ def ngens(self):
...
 
+        -  ``monic`` - optional boolean to indicate whether the sampled
+           polynomial should be monic, or not. If this is set, `0` is not a
+           possible output of the method. If monic and `0` are both needed,
+           one should use :meth:`random_monic_or_zero_element`.
```
tscrim:
```suggestion
        -  ``monic`` -- boolean  (default: ``False``); whether the sampled
           polynomial should be monic, or not
```
See topline review comment.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,41 @@ def ngens(self):
...
+        .. SEEALSO::
+
+            :meth:`random_monic_element`, :meth:`random_monic_or_zero_element`
 
         EXAMPLES::
 
             sage: R.<x> = ZZ[]
-            sage: f = R.random_element(10, 5, 10)
+            sage: f = R.random_element(10, x=5, y=10)
+            sage: f # random
```
tscrim:
```suggestion
            sage: f  # random
```
PEP8 thing IIRC; at least we are doing this spacing uniformly. Although rather than having the random polynomial, it is better to check the properties.

tscrim (changes requested):
This is now even more complicated to make something that is actually wrong be a returned value. Again, the zero polynomial is **not** monic. It should **not** be returned when `monic=True`, ever. It is a contradiction. Remove the `_allow_zero`, then apply fire to it. Then repeat for everything related to this.

Additionally, I see no point in having other methods that simply are setting options in another method. That is why we have arguments, right? These methods are creating extra maintenance complications and polluting tab completion IMO.

src/sage/rings/polynomial/polynomial_ring.py
grhkm21: Is the correct format to not include a full stop at the end of "INPUT:" docstrings?

grhkm21:
FYI, I want `monic_or_zero` because of my work on jacobians of hyperelliptic curves. There, a valid "x coordinate" of an element is exactly either a random element or zero. I suppose I will have to write this now?

```python
total_possible = 1 + sum(q^i for i in range(g + 1))
if randrange(total_possible) == 0:
    u = R(0)
else:
    u = R.random_element(degree=(0, g)).monic()
```

src/sage/rings/polynomial/polynomial_ring.py
tscrim: Generally yes. The exception is if you need multiple sentences to describe the input (although there are other ways to do more complicated behaviors).

tscrim:
> FYI, I want `monic_or_zero` because of my work on jacobians of hyperelliptic curves. There, a valid "x coordinate" of an element is exactly either a random element or zero. I suppose I will have to write this now?

Putting aside that calling a random output an exact computation is a bit strange, assuming you meant a random monic polynomial or zero, if `random_element(monic=True)1 does not return zero (again, as I think it should), then it would make sense to have a `random_monic_or_zero()` (not quite sure what the best name would be). There isn’t a method with the appropriate arguments. That being said, it probably is most natural to have an additional argument code-wise as you have. (Note, argument names with leading underscores are *not* hidden). Perhaps the most reasonable thing would be to state that if `allow_monic_zero` (again, not quite sure on the name, but it should be more descriptive than `allow_zero`, which will be confusing) is `True`, then random monic elements can return zero. However, I still find this whole business somewhat strange from an outsiders POV (i.e., without knowing your specific use-case, but that is almost certainly outside my area of expertise).

grhkm21:
I'll make a `_random_element` method instead and hide the `allow_monic_zero` method there, since as you said it's not useful for normal users, and that'll be hidden in tab completion.

tscrim:
I am still somewhat confused by your use-case for having random monic polynomials and zero (with seemingly some prescribed uniform-like (used very loosely here) distribution). I have convinced myself more that there should not be an extra parameter for `monic=True` to also return the zero polynomial, but if there is a good use-case for also returning `0`, then we can add a carefully documented parameter. Can you explain a little bit more detail how you plan to use this?

grhkm21:
Actually, I just discussed this with @yyyyx4 and I think we don't actually need the $0$ case, as I had some concepts mixed up. So I think the suggestion for $0$ can be ignored. The current commits should contain no $0$ options, so it's ready for review.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,32 @@ def ngens(self):
...
-           ``(-1,2)``.
+        -  ``degree`` - (default: ``(-1, 2)``) integer for fixing the degree or
+           a tuple of minimum and maximum degrees
+
+        -  ``monic`` - optional boolean to indicate whether the sampled
+           polynomial should be monic, or not. If this is set, `0` is not a
+           possible output of the method
 
         -  ``*args, **kwds`` - Passed on to the ``random_element`` method for
            the base ring
```
tscrim:
```suggestion
        -  ``degree`` — (default: ``(-1, 2)``) integer for fixing the degree or
           a tuple of minimum and maximum degrees

        -  ``monic`` — boolean (default: ``False``) whether the sampled
           polynomial should be monic

        -  additional (keyword) arguments are passed on to the ``random_element``
           method for the base ring
```
Since `0` is not a monic polynomial, it doesn’t need to be stated that it won’t return `0`.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1481,101 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
+        if degree == (-1, -1):
             return self.zero()
 
-        # If degree is 0, return a random constant term
-        if d == 0:
-            return self(R._random_nonzero_element(*args, **kwds))
+        # If `monic` is set, zero should be ignored
+        if degree[0] == -1:
+            degree = (0, degree[1])
+            has_zero = not monic
```
tscrim: The `has_zero` variable is not serving any purpose anymore and can be removed in favor of `monic`. This will also simplify the non-monic logic.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1481,101 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
 
-        # Add non-zero leading coefficient
-        p += R._random_nonzero_element(*args, **kwds) * self.gen() ** d
+            # If we don't want zero (not has_zero), but coefs is zero (not
+            # nonzero), then reject
+            if not has_zero and not nonzero:
+                continue
 
-        return p
+            coefs.reverse()
```
tscrim: Rather than reversing here, shouldn’t you just construct the list in the correct order?

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1421,10 +1445,17 @@ def random_element(self, degree=(-1,2), *args, **kwds):
             ....:     P = R.random_element(degree=d)
             ....:     assert P.degree() == d, "problem with {} which has not degree {}".format(P,d)
 
-            sage: R.random_element(degree=-2)
-            Traceback (most recent call last):
-            ...
-            ValueError: degree should be an integer greater or equal than -1
+        In :issue:`37118`, ranges including integers below `-1` no longer error::
```
tscrim:
```suggestion
        In :issue:`37118`, ranges including integers below `-1` no longer
        raise an error::
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1401,6 +1410,21 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
 
+        It is possible to sample a monic polynomial, but it is preferable to
+        use :meth:`random_monic_element`::
+
+            sage: R.random_element(degree=(-1, 5), monic=True).is_monic()
+            True
+            sage: R.random_monic_element(degree=(-1, 5)).is_monic()
+            True
+
+        Note that if the degree range includes `-1` and ``monic`` is set, it is silently ignored::
```
tscrim: This could be changed to remind the user that `0` is not a monic polynomial.

tscrim (changes requested):
Thanks. I am glad things are clarified.

Let me remind you about my comment that having methods that simply call another with a default is not good practice. You already have the argument, it makes the code more complicated with more to maintain (and potential to cause conflicts), and pollutes tab completion. `random_monic_polynomial` should be removed.

You also need to include a check if `monic=True` and the only degree(s) possible are `-1`. Related, if `degree[1] < -1` then you also need to error out.

I don’t find it useful to have a `# random` output, but instead to test the desired properties of the output polynomial (e.g., the degree is within the correct range).

grhkm21:
I have made changes according to your review, thanks! Regarding your review

> I don’t find it useful to have a `# random output`, but instead to test the desired properties of the output polynomial (e.g., the degree is within the correct range).

I actually find it to be the opposite. It's useful for the user to have a quick idea what the output looks like as a mental picture, and for me it makes me read the documentations faster. So I kept those.

yyyyx4:
> > I don’t find it useful to have a `# random output`, but instead to test the desired properties of the output polynomial (e.g., the degree is within the correct range).
> 
> I actually find it to be the opposite. It's useful for the user to have a quick idea what the output looks like as a mental picture, and for me it makes me read the documentations faster. So I kept those.

This seems to capture precisely the difference between an example and a test: Examples are for users, tests are for the machine.
(However, there is a non-empty intersection between the concepts — of course examples should *work*, and on the other hand it may be helpful to users to see some expected/guaranteed properties of the output spelled out.)

tscrim:
> > > I don’t find it useful to have a `# random output`, but instead to test the desired properties of the output polynomial (e.g., the degree is within the correct range).
> > 
> > 
> > I actually find it to be the opposite. It's useful for the user to have a quick idea what the output looks like as a mental picture, and for me it makes me read the documentations faster. So I kept those.

IMO, this is a fallacy. If I happen to  get ``0`, this does not look like a degree 10 polynomial. This is different than if the output order is random (in which case, there are generally ways to convert the test into something deterministic; e.g., sorting).

> This seems to capture precisely the difference between an example and a test: Examples are for users, tests are for the machine. (However, there is a non-empty intersection between the concepts — of course examples should _work_, and on the other hand it may be helpful to users to see some expected/guaranteed properties of the output spelled out.)

Indeed, the last part of what you said is the key point, and not only for the test.

tscrim:
> I have made changes according to your review, thanks! Regarding your review

This isn't quite accurate. Let me be very clear (sorry for the directness, don't read this as me being upset): The `random_monic_element` method needs to be removed for this to gain a positive review. I see no benefits to keeping it but lots of demerits.

grhkm21:
Sorry for the delay, I was away for the past week. I've removed the `monic` method, I missed that accidentally.

About the `# random` discussion, yes I agree with what you said, and I don't think it contradicts with what I said. Take this segment of doc:

```sage
        EXAMPLES::

            sage: R.<x> = ZZ[]
            sage: f = R.random_element(10, x=5, y=10)
            sage: f  # random
            5*x^10 + 6*x^9 + 5*x^8 + 8*x^7 + 8*x^6 + 5*x^5 + 7*x^4 + 8*x^3 + 6*x^2 + 9*x + 6
            sage: f.degree()
            10
            sage: f.parent() is R
            True
            sage: all(a in range(5, 10) for a in f.coefficients())
            True
            sage: R.random_element(6).degree()
            6
```

IMO this is useful both for the user and for ensuring correctness. The final 8 lines shows the property of the output so it achieves what you said, and the `# random` call simply shows an example, again for a mental picture. I don't think it should be removed, but if you feel strongly about it I will remove it to avoid further off-topic discussion :D

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1401,6 +1405,14 @@ def random_element(self, degree=(-1,2), *args, **kwds):
             sage: while R.random_element(degree=(-1,2), x=-1, y=1) != R.zero():
             ....:     pass
 
+        Note that if the degree range includes `-1` and ``monic`` is set, it is
+        silently ignored, as `0` is not a monic polynomial::
```
tscrim:
```suggestion
        Monic polynomials are also chosen among all monic polynomials
        with degree between the given  ``degree`` argument::
```
I don't think you need to tell people that `0` is not a monic polynomial. :) Same comment about uniformity.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1433,11 +1463,17 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
         else:
-            degree = (degree,degree)
+            if degree < -1:
+                raise ValueError(f"degree (={degree}) must be at least -1")
+            degree = (degree, degree)
 
         if degree[0] <= -2:
-            raise ValueError("degree should be an integer greater or equal than -1")
+            # This error has been removed in issue #37118.
+            # raise ValueError("degree should be an integer greater or equal than -1")
```
tscrim:
```suggestion
```
The comment is not necessary. We have the git history.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1484,48 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
-        # Pick random coefficients
-        p = self([R.random_element(*args, **kwds) for _ in range(d)])
-
-        # Add non-zero leading coefficient
-        p += R._random_nonzero_element(*args, **kwds) * self.gen() ** d
-
-        return p
+        # If `monic` is set, zero should be ignored
+        if degree[0] == -1 and monic:
+            degree = (0, degree[1])
```
tscrim:
```suggestion
        if degree[0] == -1 and monic:
            if degree[1] == -1:
                raise ValueError("the maximum degree of monic polynomials needs to be at least 0")
            if degree[1] == 0:  # Must be a constant and monic, so 1
                return self.one()
            degree = (0, degree[1])
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,27 @@ def ngens(self):
...
         r"""
-        Return a random polynomial of given degree or with given degree bounds.
+        Return a random polynomial of given degree (bounds).
 
         INPUT:
 
-        -  ``degree`` - optional integer for fixing the degree
-           or a tuple of minimum and maximum degrees. By default set to
-           ``(-1,2)``.
+        -  ``degree`` - (default: ``(-1, 2)``) integer for fixing the degree or
```
tscrim:
```suggestion
        -  ``degree`` -- (default: ``(-1, 2)``) integer for fixing the degree or
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1360,23 +1360,27 @@ def ngens(self):
...
+           a tuple of minimum and maximum degrees
 
-        -  ``*args, **kwds`` - Passed on to the ``random_element`` method for
-           the base ring
+        -  ``monic`` - optional boolean to indicate whether the sampled
+           polynomial should be monic
+
+        -  ``*args, **kwds`` - additional keyword parameters passed on to the
+           ``random_element`` method for the base ring
```
tscrim:
```suggestion
        -  ``monic`` -- boolean (optional); indicate whether the sampled
           polynomial should be monic

        -  ``*args, **kwds`` -- additional keyword parameters passed on to the
           ``random_element`` method for the base ring
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1484,48 @@ def random_element(self, degree=(-1,2), *args, **kwds):
             else:
                 raise ValueError("No polynomial of degree >= 0 has all coefficients zero")
 
-        # Pick a random degree
-        d = randint(degree[0], degree[1])
-
-        # If degree is -1, return the 0 polynomial
-        if d == -1:
+        degree_lb = degree[0]
```
tscrim:
```suggestion
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1484,48 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
-        return p
+        # If `monic` is set, zero should be ignored
+        if degree[0] == -1 and monic:
+            degree = (0, degree[1])
+
+        while True:
+            # Pick random coefficients
+            end = degree[1]
+            coefs = [None] * (degree[1] + 1)
+            nonzero = False
```
tscrim:
```suggestion
        end = degree[1]
        if degree[0] == -1:  # already know it is not monic
            # Since we allow 0 as an output, we can get just a random list of elements
            return self([R.random_element(*args, **kwds) for _ in range(end+1)])

        nonzero = False
        coefs = [None] * (end + 1)  # We will always replace every element each iteration
        while not nonzero:
            # Pick random coefficients
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1448,24 +1484,48 @@ def random_element(self, degree=(-1,2), *args, **kwds):
...
+            # tracked to handle `has_zero`.
+            for i in range(degree[1] - degree[0] + 1, degree[1] + 1):
+                coefs[end - i] = R.random_element(*args, **kwds)
+
+            # If we don't want zero (not has_zero), but coefs is zero (not
+            # nonzero), then reject
+            if degree_lb == -1 and not nonzero:
+                continue
+
+            return self(coefs)
```
tscrim:
```suggestion
        # Now we pick the remaining coefficients
        for i in range(degree[1] - degree[0] + 1, degree[1] + 1):
            coefs[end - i] = R.random_element(*args, **kwds)

        return self(coefs)
```

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1386,8 +1390,8 @@ def random_element(self, degree=(-1,2), *args, **kwds):
             sage: R.random_element(6).degree()
             6
 
-        If a tuple of two integers is given for the ``degree`` argument, a degree
-        is first uniformly chosen, then a polynomial of that degree is given::
+        If a tuple of two integers is given for the ``degree`` argument, a polynomial is chosen
+        uniformly among all polynomials with degree between them::
```
tscrim:
```suggestion
        among all polynomials with degree between them::
```
I am dubious about this uniform distribution claim since it depends highly on the sampling of the coefficient ring (which doesn't have to be uniform).

tscrim (changes requested):
No problem. Thank you for the additional changes.

However, I do feel somewhat strongly about not having the `# random` tests here. Having the output there is useless IMO and could even potentially confuse a user. They were not included here for a reason I believe.

I also have some additional comments to try and improve the code some more.

tscrim:
Also, there are some failing doctests due to the changed random sampling algorithm to fix too.

src/sage/rings/polynomial/polynomial_ring.py
grhkm21: I think the information is still important to note for cases like $\mathbb{F}_{103}[x]$ which is uniform. I now phrase it as

```
If a tuple of two integers is given for the ``degree`` argument, a polynomial is chosen
among all polynomials with degree between them. If the base ring is uniform, so is this
method::
```

is that ok? Not sure if the English is really correct.

grhkm21:
Oh there’s one failing test still oops. I will fix it when I am back.

src/sage/rings/polynomial/polynomial_ring.py
tscrim: I think the last sentence needs to be something like "If the base ring is sampled uniformly, then this is also sampled uniformly::"

I am wondering if we should include a test about the uniformness. It's hard to test it reliably is random, but perhaps we should sample over something like 10000 polynomials of degree 3 over $F_3$ and check the distribution for the first digit? I don't think this is enough for 2 digits in general. It might not be worth it to spend a lot of time on this.

grhkm21: I agree that we should have a test for uniformness, but I don't know how to approach this correctly. I tried a chi-squared test with say $p = 0.05$, but it often fails, c.f. ["Too Big to Fail: Large Samples and the p-Value Problem"](https://www.researchgate.net/publication/270504262_Too_Big_to_Fail_Large_Samples_and_the_p-Value_Problem).

grhkm21:
I have a question about the failing doctest. The output from `sage -t` and from the Sage REPL is different. This is from `sage -t` (also from GitHub actions):

Screenshot 2024-02-10 at 17 30 58

This is from REPL:

Screenshot 2024-02-10 at 17 31 32

tscrim:
> I have a question about the failing doctest. The output from `sage -t` and from the Sage REPL is different. This is from `sage -t` (also from GitHub actions):

This happens for a few things; we sort certain outputs for tests that we wouldn’t otherwise (in order to have the consistency in testing the output). So just take the output of `sage -t`.

grhkm21:
Interesting, I have done that.

src/sage/rings/polynomial/polynomial_ring.py
tscrim: Yea, I couldn't think of a good way to reasonably guarantee this. Let's just leave it.

However, it doesn't make sense to say "the base ring is uniform". You need to discuss about the sampling. My previous suggestion can also be improved by "If we can sample uniformly from the base ring, then ..."

grhkm21:
Just a side note, I recently learned about FLINT, and in particular its `nmod_poly_randtest_monic` method and others, [docs](nmod_poly_randtest_monic). It should be faster and should be used I guess.

tscrim:
It would be faster, but only for certain base rings (which already have specialized implementations IIRC). So it could be used in those specialized classes.

Also there are 2 doctests failing in `polynomial_ring.py` according to the test bot.

grhkm21:
Oh lmao my `samples = [...]` is marked as `# long time`, but the assert check afterwards is not. In fact the `--all --long` tests pass.

Is there a way to mark the entire test as `long time`? Or do I just add the tag after every line?

tscrim:
If you add `sage: # long time` to the start of a block, then all tests within that block are counted as `# long time`, but it is generally better to do it locally IMO. In this case, there is only 2 tests it seems.

src/sage/rings/polynomial/polynomial_ring.py
tscrim: This still needs to be addressed. It does not make sense to say the base ring is uniform.

grhkm21:
@tscrim Do you mind reviewing again? Or is it ready now? CI pass now (before I merged)

src/sage/rings/polynomial/polynomial_ring.py
grhkm21: Oh god, I completely missed this comment. I have addressed it now though the wording is slightly different. (I fixed it before I saw your newest comment two days ago)

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -2414,8 +2477,8 @@ def lagrange_polynomial(self, points, algorithm="divided_difference", previous_r
 #                 P += (F[i] * prod)
 #             return P
 
-        # using Neville's method for recursively generating the
-        # Lagrange interpolation polynomial
+# using Neville's method for recursively generating the
+# Lagrange interpolation polynomial
```
tscrim: Please revert. The indentation is useful and such a change can just create (merge) conflicts.

tscrim (approved):
Thank you. One last little thing to do. It would also be good if the lines you changed were at 80 char/line or less too.

src/sage/rings/polynomial/polynomial_ring.py
grhkm21: Sorry, I blame this on my editor.

grhkm21:
I thought the maximum line width in Sage is 100, and (if I didn't miss any) my changes should wrap at that

tscrim:
For code, we try to do 100 (but there can be reasons for breaking this and this is less enforced), but docstrings we generally try to keep shorter.

grhkm21:
Do test outputs have to be wrapped at 80 too? e.g.
```diff
+        Testing uniformity::
+
+            sage: from collections import Counter
+            sage: R = GF(3)["x"]
+            sage: samples = [R.random_element(degree=(-1, 2)) for _ in range(27000)]    # long time
+            sage: assert all(750 <= f <= 1250 for f in Counter(samples).values())       # long time
+
+            sage: samples = [R.random_element(degree=(-1, 2), monic=True) for _ in range(13000)] # long time
+            sage: assert all(750 <= f <= 1250 for f in Counter(samples).values())       # long time
```

grhkm21:
I will be back and fix the rest in a few hours.

tscrim:
Thanks. This is good with me now.

Testing output can be a bit more lax about it, but it is still a good approximate guide.

PR #37119, opened 2024-01-21
Implement `EllipticCurve_with_order`

### LINKED ISSUE #37131: Incorrect doctest in `set_order`
These lines in the doctest is incorrect:

https://github.com/sagemath/sage/blob/3dd953c3aa6b5143071b6c39208199cf128c8080/src/sage/schemes/elliptic_curves/ell_finite_field.py#L1346-L1351

The elliptic curve actually has order 12, as shown by printing `E.points()`:

```
sage: E.points()
[(0 : 1 : 0), (0 : 1 : 1), (0 : 6 : 1), (1 : 3 : 1), (1 : 4 : 1), (2 : 3 : 1), (2 : 4 : 1), (3 : 0 : 1), (4 : 3 : 1), (4 : 4 : 1), (5 : 0 : 1), (6 : 0 : 1)]
```

### Additional Information

All points have order $6$, but not all groups are abelian!

Fixed in #37119.

grhkm21:
This function returns a curve / an iterator of curves with a given order. These are generated using the CM method.

Also fix #37131 (13 year old bug).

Co-authored-by: @GiacomoPope #sd123 🚀 

```python
sage: from sage.schemes.elliptic_curves.ell_finite_field import EllipticCurve_with_order
sage: E = EllipticCurve_with_order(1234); E
Elliptic Curve defined by y^2 = x^3 + 8*x over Finite Field of size 1229
sage: E.order() == 1234
True
```


grhkm21:
The function can be sped up on some small edge cases when https://github.com/sagemath/sage/pull/37110 is finished, but it works as is.

grhkm21:
I believe this is ready for review :)

src/sage/quadratic_forms/binary_qf.py
```diff
@@ -1596,17 +1596,20 @@ def small_prime_value(self, Bmax=1000):
                 raise ValueError("Unable to find a prime value of %s" % self)
             B += 10
 
-    def solve_integer(self, n, *, algorithm="general"):
+    def solve_integer(self, n, *, algorithm="general", flag=2):
```
yyyyx4: In my opinion we shouldn't expose PARI's non-human-readable APIs: Can `flag` be passed by a sensible choice of keyword arguments instead?

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -185,21 +199,31 @@ def quadratic_twist(self, D=None):
...
+                E0 = self.change_weierstrass_model(a1,a3/a1,0,(a1**2*a4+a3**2)/a1**3)
+                # which has the form = [1,A2,0,0,A6]
+                assert E0.a1() == K(1)
+                assert E0.a3() == K(0)
+                assert E0.a4() == K(0)
+                Et = EllipticCurve(K,[1,E0.a2()+D,0,0,E0.a6()])
+            else:
+                raise ValueError("Quadratic twist not implemented in char 2 when j=0")
+
+        if isinstance(K, FiniteField_generic):
```
yyyyx4: Typically the abstract base class `sage.rings.finite_rings.finite_field_base.FiniteField` is used for such checks.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1537,7 +1550,7 @@ def height_above_floor(self, ell, e):
         from sage.libs.pari.convert_sage import gen_to_sage
         from sage.libs.pari.all import pari
         X,Y = polygens(F,['X', 'Y'],2)
-        phi = gen_to_sage(pari.polmodular(ell),{'x':X, 'y':Y})
+        phi = gen_to_sage(pari.polmodular(ell),{'x': X, 'y': Y})
```
yyyyx4: This could now just be `classical_modular_polynomial(ell)(X,Y)`, which opportunistically caches the result.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1851,7 +1873,7 @@ def curves_with_j_0(K):
          Elliptic Curve defined by y^2 = x^3 + 4 over Finite Field of size 7,
          Elliptic Curve defined by y^2 = x^3 + 5 over Finite Field of size 7,
          Elliptic Curve defined by y^2 = x^3 + 6 over Finite Field of size 7]
-        sage: curves_with_j_0(GF(25))                                                   # needs sage.rings.finite_rings
+        sage: curves_with_j_0(GF(25))  # random                                         # needs sage.rings.finite_rings
```
yyyyx4: Is this really random? (Same elsewhere.)

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1870,6 +1892,8 @@ def curves_with_j_0(K):
         sage: curves_with_j_0(GF(11))
         [Elliptic Curve defined by y^2 = x^3 + 1 over Finite Field of size 11,
          Elliptic Curve defined by y^2 = x^3 + 6 over Finite Field of size 11]
+
+    # TODO (grhkm): Docs for order precomputation
```
yyyyx4: There are a couple of TODOs like this left in the branch; I assume that was unintentional?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2109,9 +2202,30 @@ def curves_with_j_0_char3(K):
...
+    if deg % 2:
+        curves = [
+            EllipticCurve(K, a4a6)
+            for a4a6 in [[1, 0], [-1, 0], [-1, b], [-1, -b]]
+        ]
+
+        # It is a bit annoying here since we know the last two curves have order base +- delta, but
+        # we don't know the sign. I think you can compute it via some analysis on `b`, but that is
+        # slow, so I opt to use `.set_order` with check=True, so that if it fails it means it is the
+        # wrong order :)
```
yyyyx4: The `.set_order()` check can have false positives; is there a reason that can't happen here?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2615,3 +2739,150 @@ def special_supersingular_curve(F, *, endomorphism=False):
...
+        ....:                 assert any(Et.is_isomorphic(E) for E in Es)
+
+    .. NOTE::
+
+        The output curves are not fixed, as :func:`EllipticCurve_finite_field.twists` is not
+        deterministic. However, the order of the j-invariants and base fields is fixed.
+
+    AUTHORS:
+
+     - Gareth Ma and Giacomo Pope (Sage Days 123): Fix bug for small curves
```
yyyyx4: This function was written from scratch at Sage Days 123, so only mentioning a bug fix seems strange. Was this supposed to be in another function?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2615,3 +2739,150 @@ def special_supersingular_curve(F, *, endomorphism=False):
...
+    A `:sage:`~sage.structure.factorization.Factorization` can be passed for ``m``, in which case
+    the algorithm is more efficient.
+
+    If ``D`` is specified, it is used as the discriminant.
+
+    If ``all`` is set, a list of all elliptic curves over finite fields with the given order is
+    returned.
+
+    If ``iter`` is set, an iterator of all elliptic curves over finite fields with the given order
+    is returned. This cannot be set with ``all``.
```
yyyyx4: Always returning an iterator actually covers all three use cases using `next()` and `list()`, so I'm not convinced we need two keyword arguments just for this. (Iterators are not a very common sight in the Sage library, but I think they should be!)

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: I have some heuristics argument, but I think I will comment this out for now, and leave it for another issue (#37110)

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: Yes because it chooses a random twisting factor every time. But I will add a check for `all(E.j() == 0)`

src/sage/schemes/elliptic_curves/ell_field.py
grhkm21: Yes, it's aliased at `import`. I think it's a common pattern in Sage, since `FiniteField` is also a function to construct a `FiniteField` (class)?

grhkm21:
The pari flag review is non-trivial, so give me a day or two :) Thanks once again!

GiacomoPope:
Is this being held up because of the pari flag?

grhkm21:
No, I deferred it to another issue already

GiacomoPope:
I'm a coauthor, so I dont think I should review it further, but if @yyyyx4's only comment left is about the PARI flag and this is being delayed then maybe a positive review is OK?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1818,15 +1839,21 @@ def twists(self):
...
-            while not D or D**q2 == 1:
+        else:
+            # find a nonsquare D. note that computing a multiplicative generator is expensive
+            q2 = (K.cardinality() - 1) // 2
+            while True:
                 D = K.random_element()
+                if D and D**q2 != 1:
+                    break
+            # assert D and D**q2 != 1
+            # assert not D.is_square9)
```
yyyyx4:
```suggestion
            # assert not D.is_square()
```

src/sage/quadratic_forms/binary_qf.py
```diff
@@ -1596,17 +1596,20 @@ def small_prime_value(self, Bmax=1000):
...
 
         INPUT:
 
-        - ``n`` -- a positive integer
+        - ``n`` -- a positive integer or a
+          `:sage:`~sage.structure.factorization.Factorization` object
 
         - ``algorithm`` -- ``"general"`` (default) or ``"cornacchia"``
 
+        - ``_flag`` -- ``1``, ``2`` (default) or ``3``
```
yyyyx4: The underscore suggests this already, but perhaps you could say explicitly that this argument is for internal use only? This way we could skip a deprecation when #37364 eventually gets done.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1709,7 +1721,7 @@ def twists(self):
 
             sage: # needs sage.rings.finite_rings
             sage: K = GF(3**5)
-            sage: [E.ainvs() for E in EllipticCurve(j=K(1)).twists()]
+            sage: [E.ainvs() for E in EllipticCurve(j=K(1)).twists()] # random
```
yyyyx4: If this is just for the ordering, use `sorted()` instead of `# random`. If the curves are actually different, it would be good to add a check that the returned list has the correct length and that the curves all have the correct 𝑗‑invariant and that they are pairwise non-isomorphic over the base field.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1818,15 +1839,21 @@ def twists(self):
...
                 D = K.random_element()
-        else: # find a nonsquare D
-            D = K.gen()
-            q2 = (K.cardinality()-1)//2
-            while not D or D**q2 == 1:
+        else:
+            # find a nonsquare D. note that computing a multiplicative generator is expensive
+            q2 = (K.cardinality() - 1) // 2
+            while True:
                 D = K.random_element()
```
yyyyx4, line 1861: I don't understand the motivation behind this change: `K.gen()` doesn't compute a multiplicative generator.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1851,13 +1878,8 @@ def curves_with_j_0(K):
...
          Elliptic Curve defined by y^2 = x^3 + 6 over Finite Field of size 7]
-        sage: curves_with_j_0(GF(25))                                                   # needs sage.rings.finite_rings
-        [Elliptic Curve defined by y^2 = x^3 + 1 over Finite Field in z2 of size 5^2,
-         Elliptic Curve defined by y^2 = x^3 + z2 over Finite Field in z2 of size 5^2,
-         Elliptic Curve defined by y^2 = x^3 + (z2+3) over Finite Field in z2 of size 5^2,
-         Elliptic Curve defined by y^2 = x^3 + (4*z2+3) over Finite Field in z2 of size 5^2,
-         Elliptic Curve defined by y^2 = x^3 + (2*z2+2) over Finite Field in z2 of size 5^2,
-         Elliptic Curve defined by y^2 = x^3 + (4*z2+1) over Finite Field in z2 of size 5^2]
+        sage: set(E.j_invariant() for E in curves_with_j_0(GF(25)))                     # needs sage.rings.finite_rings
+        {0}
```
yyyyx4: Should we check the length and pairwise non-isomorphism here, too?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1881,14 +1903,23 @@ def curves_with_j_0(K):
...
-    q2 = (q-1)//2
-    q3 = (q-1)//3
-    D = K.gen()
+    q2 = (q - 1) // 2
+    q3 = (q - 1) // 3
+    # Note that computing multiplicative generator is expensive
+    D = K.random_element()
     while not D or D**q2 == 1 or D**q3 == 1:
         D = K.random_element()
-    return [EllipticCurve(K, [0,D**i]) for i in range(6)]
```
yyyyx4: Same here, `K.gen()` is not slow.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1881,14 +1903,23 @@ def curves_with_j_0(K):
...
+
+    curves = [EllipticCurve(K, [0, D**i]) for i in range(6)]
+    # TODO (grhkm): issue 37110, Precompute orders of sextic twists + docs
```
yyyyx4: I don't think it's common (or a good idea) to label TODOs in the code with your name; suggesting to just leave this out.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1937,13 +1968,25 @@ def curves_with_j_1728(K):
...
-    q2 = (q-1)//2
-    D = K.gen()
+    q2 = (q - 1) // 2
+    # Note that computing multiplicative generator is expensive
+    D = K.random_element()
     while not D or D**q2 == 1:
         D = K.random_element()
-    return [EllipticCurve(K, [D**i,0]) for i in range(4)]
+    curves = [EllipticCurve(K, [D**i, 0]) for i in range(4)]
+    # TODO (grhkm): issue 37110, Precompute orders of quartic twists + docs
```
yyyyx4: Same here.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2109,9 +2198,27 @@ def curves_with_j_0_char3(K):
...
+            EllipticCurve(K, a4a6)
+            for a4a6 in [[1, 0], [-1, 0], [-1, b], [-1, -b]]
+        ]
+
+        base = 3**deg + 1
+        delta = 3**((deg + 1) // 2)
+
+        curves[0].set_order(base)
+        curves[1].set_order(base)
+        # TODO (grhkm): issue 37110
```
yyyyx4: Same here.

yyyyx4 (changes requested):
Sorry, the reason this was stuck is simply because I hadn't managed to get back to it yet. Here are some comments.

src/sage/schemes/elliptic_curves/ell_field.py
GiacomoPope: another option is `K in FiniteFields()` using the category

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -207,21 +221,31 @@ def quadratic_twist(self, D=None):
...
+            try:
```
GiacomoPope: Could use `if hasattr(self, "_order"):`?

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -207,21 +221,31 @@ def quadratic_twist(self, D=None):
...
-        # now char==2
-        if self.j_invariant() != 0: # iff a1!=0
-            a1,a2,a3,a4,a6 = self.ainvs()
-            E0 = self.change_weierstrass_model(a1,a3/a1,0,(a1**2*a4+a3**2)/a1**3)
-            # which has the form = [1,A2,0,0,A6]
-            assert E0.a1() == K(1)
-            assert E0.a3() == K(0)
-            assert E0.a4() == K(0)
-            return EllipticCurve(K,[1,E0.a2()+D,0,0,E0.a6()])
+            Et = EllipticCurve(K,[0,b2*D,0,8*b4*D**2,16*b6*D**3])
```
GiacomoPope: Some whitespace around `,` and `8 * b4 `... is normally what linting likes, right?

src/sage/quadratic_forms/binary_qf.py
```diff
@@ -1702,22 +1714,39 @@ def solve_integer(self, n, *, algorithm="general"):
...
+        """
         if self.is_negative_definite():  # not supported by PARI
             return (-self).solve_integer(-n)
 
         if self.is_reducible():  # square discriminant; not supported by PARI
+            try:
+                n = ZZ(n)
+            except TypeError:
+                # Factorization
+                n = ZZ(n.value())
```
GiacomoPope: Should the except here check `n` is of type `Factorisation`?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1765,6 +1777,14 @@ def twists(self):
              (0, 0, z8^6 + z8^3 + z8^2, 0, 0),
              (0, 0, z8^6 + z8^3 + z8^2, 0, z8^3 + z8^2)]
 
+        The order of the quadratic twist is precomputed, so the second
+        computation below is fast::
+
+            sage: E = EllipticCurve(GF(2^255 - 19), [3, 5])
+            sage: E.set_order(2^255 - 94400682670973290858718002131957303262)
+            sage: E.quadratic_twist().order()
```
GiacomoPope: Maybe instead of this you can show `E.quadratic_twist()._order` works (or with a hasattr to show the order is already there?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1741,7 +1753,7 @@ def twists(self):
 
             sage: # needs sage.rings.finite_rings
             sage: K = GF(2**7)
-            sage: [E.ainvs() for E in EllipticCurve(j=K(1)).twists()]
+            sage: [E.ainvs() for E in EllipticCurve(j=K(1)).twists()] # random
```
GiacomoPope: I agree with Lorenz that this could be sorted and random could be removed, or is the issue that the twists themselves are random. Maybe im being stupid here.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2006,13 +2052,32 @@ def curves_with_j_0_char2(K):
...
     """
     if not K.is_finite() or K.characteristic() != 2:
         raise ValueError("field must be finite of characteristic 2")
-    if K.degree() % 2:
-        return [EllipticCurve(K, [0, 0, 1, 0, 0]),
-                EllipticCurve(K, [0, 0, 1, 1, 0]),
-                EllipticCurve(K, [0, 0, 1, 1, 1])]
+
+    deg = K.degree()
+    if deg % 2 == 1:
```
GiacomoPope: this could be

```suggestion
    if deg % 2:
```

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2028,15 +2093,26 @@ def curves_with_j_0_char2(K):
     while not b or not b.trace():
         b = K.random_element()
     c = K.one() # OK if degree is 2 mod 4
-    if K.degree() % 4 == 0:
+    if deg % 4 == 0:
```
GiacomoPope: Similarly,
```suggestion
    if not deg % 4:
```

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2615,3 +2735,120 @@ def special_supersingular_curve(F, *, endomorphism=False):
...
+    AUTHORS:
+
+     - Gareth Ma and Giacomo Pope (Sage Days 123): initial version
+    """
+    from sage.arith.misc import is_prime_power, factor
+    from sage.quadratic_forms.binary_qf import BinaryQF
+    from sage.structure.factorization import Factorization
+    from sage.schemes.elliptic_curves.cm import hilbert_class_polynomial
+
+    def helper(m, m4_fac, D):
```
GiacomoPope: maybe we should have given this a better name? :D

GiacomoPope (commented):
just some comments which could be included / ignored.

grhkm21:
Sorry for the snail pace work. I've reverted the changes related to precomputing order, since they don't affect correctness, only speed. They're at #37110 anyways.

Next, I'll work on your reviews

grhkm21:
Conda failure is a real bug, I will make a new PR

GiacomoPope (approved):
LGTM

grhkm21:
Thanks!

PR #37123, opened 2024-01-21
Compatibility with scipy 1.12

antonio-rojas:
Drop usage of some numpy function aliases which are no longer provided in 1.12

### DIFF
--- a/src/sage/matrix/matrix_double_dense.pyx
+++ b/src/sage/matrix/matrix_double_dense.pyx
@@ -867,7 +867,7 @@ cdef class Matrix_double_dense(Matrix_numpy_dense):
         # set cutoff as RDF element
         if eps == 'auto':
             if scipy is None: import scipy
-            eps = 2*max(self._nrows, self._ncols)*scipy.finfo(float).eps*sv[0]
+            eps = 2*max(self._nrows, self._ncols)*numpy.finfo(float).eps*sv[0]
         eps = RDF(eps)
         # locate non-zero entries
         rank = 0
--- a/src/sage/numerical/optimize.py
+++ b/src/sage/numerical/optimize.py
@@ -426,7 +426,7 @@ def minimize(func, x0, gradient=None, hessian=None, algorithm="default",
                 hess = func.hessian()
                 hess_fast = [ [fast_callable(a, vars=var_names, domain=float) for a in row] for row in hess]
                 hessian = lambda p: [[a(*p) for a in row] for row in hess_fast]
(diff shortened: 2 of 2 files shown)


tornaria:
LGTM, will test later.

BTW, there are some changes in openblas 0.3.26 that cause some regressions on scipy (either scipy 1.12 or 1.11.4 is broken). The fix in https://github.com/scipy/scipy/pull/19909 didn't make it to scipy 1.12.

antonio-rojas:
> BTW, there are some changes in openblas 0.3.26 that cause some regressions on scipy (either scipy 1.12 or 1.11.4 is broken). The fix in [scipy/scipy#19909](https://github.com/scipy/scipy/pull/19909) didn't make it to scipy 1.12.

We've been shipping 0.3.26 for a while, seen no issues on either scipy of sage test suites.

kiwifb:
Definitely needed, and it appears those are the only instances that need changing. At least those functions.

kiwifb (approved):

tornaria:
Everything looks fine: https://github.com/void-linux/void-packages/pull/48317

There's an unrelated failure on i686 there, a timeout in:
```
...
sage: R.<x> = PolynomialRing(GF(65537), implementation="FLINT") ## line 701 ##
sage: f = R.random_element(9973) * R.random_element(10007) ## line 702 ##
sage: alarm(0.5); f.factor() ## line 703 ##
```
Have you seen this? I think these things occasionally happen (around the `alarm()`) calls, not specific to i686. My guess is the alarm signal gets lost. The test looks stupid but I think it's there to make sure `f.factor()` can be interrupted (say, by hitting ctrl-C, if that signal gets lots one can hit ctrl-C again ofc). Not sure what to do about this (my first reflex is to add a second alarm so if the first one is lost the second one gets through, but `alarm(2)` doesn't allow for more than one alarm.

tornaria (approved):

antonio-rojas:
> Have you seen this? 

I haven't, but we only build for x86_64

kiwifb:
> > Have you seen this?
> 
> I haven't, but we only build for x86_64

Same here, I am only building for x86_64 right now.

tornaria:
> > > Have you seen this?
> > 
> > 
> > I haven't, but we only build for x86_64
> 
> Same here, I am only building for x86_64 right now.

I don't think this has anything to do with 32 bit. It seems more like the alarm signal is lost because of some race condition (?), and it triggers in CI since gh runners are sometimes very heavily loaded.

PR #37125, opened 2024-01-21
compute isogeny kernel polynomials from irreducible divisors or (possibly irrational) points

yyyyx4:
In this patch we implement Algorithms 3 and 4 from [this paper](https://ia.cr/2023/106). Evidently, the necessity for such an algorithm arises when converting from endomorphism-ring ideals to isogenies.

One immediate application is that `isogenies_prime_degree_general()` can be simplified, even with a small speedup:

```sage
sage: K.<i> = QuadraticField(-1)
sage: E = EllipticCurve(K,[0,0,0,1,0])
sage: %timeit E.isogenies_prime_degree(37)
````
Before: `2.39 s ± 3.31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
After: `2.29 s ± 5.05 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

#sd123

### DIFF
--- a/src/doc/en/reference/references/index.rst
+++ b/src/doc/en/reference/references/index.rst
@@ -2444,6 +2444,11 @@ REFERENCES:
              J. Graph Algorithms and Applications 15 (2): 269-293, 2011.
              :doi:`10.7155/jgaa.00226`, :arxiv:`0705.1025`.
 
+.. [EPSV2023] Jonathan Komada Eriksen, Lorenz Panny, Jana Sotáková, and Mattia Veroni.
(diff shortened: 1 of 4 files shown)


JohnCremona:
I will review this -- but not before Tuesday.

JohnCremona (approved):
Looks good to me.  (To the release manager: I am the author of the original strategy, and I am slo familiar with the paper on which the new strategy is based.)

yyyyx4:
Thank you!

PR #36768, opened 2023-11-25
Retrieve more information of fundamental groups of plane curves

enriqueartal:
The goal of this PR is to add some features to the computation of fundamental groups of complements of the plane curves introduced in ``Sagemath``  by @miguelmarco and his optional package ``sirocco``. It also corrects a bug introduced in https://github.com/sagemath/sage/pull/35376 and cleans some code.

We add the following information: given a tuple of curves we retrieve the information of meridians of each curve as words in the fundamental group, and also of the strands of the braid monodromy. We also add the possibility of excluding the vertical lines from the computation of braid monodromy (as far as there is no vertical asymptote) which makes its computation faster.

To achieve this goal we introduce new classes `AffinePlaneCurveArrangements` and `ProjectivePlaneCurveArrangements`. For now, they serve only to compute the fundamental groups with the meridian information (and braid monodromy and strand information for the affine curves) but more methods may be constructed in the future. The structure of these classes follows the one of `HyperplaneArrangements`, with one difference, the order is important. This is why we also introduce the class of `OrderedHyperplaneArrangements`, where the fundamental group is computed. We have also introduce a new method for hyperplane arrangements, `jhyperplane_section`, in order to compute the fundamental group also for hyperplane arrangements and not only for line arrangements.

These are the main changes:

- schemes/curves/affine_curve.py
     - Introducion of  some auxiliary new methods: `has_vertical_asymptote`, `is_vertical_line`
     - Small update of `fundamental_group`
- schemes/curves/projective_curve.py. Some trivial cases were not considered in `fundamental_group`.
- schemes/curves/plane_curve_arrangement.py. New file with the new classes.
- schemes/curves/zariski_vankampen.py
    - For consistency some lists are converted into tuples.
    - Some functions take a new keyword to take into account the option of eliminating the vertical lines for the computation of the braid monodromy.
    - In `braid_in_segment`, correction of the bug introduced in the previous PR (the parameters must be in the number field embedded in `QQbar`). 
    - The function `geometric_basis` has been changed for designed and to cover some cases where the former one failed.
    - A new function `vertical_lines_in_braidmon` to isolate vertical lines if needed.
    - The function `braid_monodromy` has been cleaned and now it takes into account the strand information.
    - 
- geometry/hyperplane_arrangement/arrangement.py
     - Introducion of the class of `OrderedHyperplaneArrangements`. The hyperplanes are not sorted in this class and the introduction order is respected. It forces to update some methods.
[...]

PR #37128, opened 2024-01-21
Save and load finitely presented groups coming from libgap groups

### LINKED ISSUE #37061: Save and load a finitely presented group
### Steps To Reproduce

Run the following code:
```
sage: save(FreeGroup(1)/[], 'ab')
sage: load('ab.sobj')
Finitely presented group < x |  >
sage: from sage.groups.finitely_presented import wrap_FpGroup
sage: save(wrap_FpGroup(libgap(FreeGroup(1)/[])), 'ab')
sage: load('ab.sobj')
Cell In[20], line 1
----> 1 load('ab.sobj')

File /usr/local/sage-310/src/sage/misc/persist.pyx:189, in sage.misc.persist.load()
    187 
    188     ## Load file by absolute filename
--> 189     with open(filename, 'rb') as fobj:
    190         X = loads(fobj.read(), compress=compress, **kwargs)
    191     try:

File /usr/local/sage-310/src/sage/misc/persist.pyx:190, in sage.misc.persist.load()
    188 ## Load file by absolute filename
    189 with open(filename, 'rb') as fobj:
--> 190     X = loads(fobj.read(), compress=compress, **kwargs)
    191 try:
    192     X._default_filename = os.path.abspath(filename)

File /usr/local/sage-310/src/sage/misc/persist.pyx:995, in sage.misc.persist.loads()
    993 unpickler = SageUnpickler(io.BytesIO(s), **kwargs)
    994 global already_unpickled
--> 995 ans = unpickler.load()
    996 already_unpickled = { }
    997 return ans

File /usr/local/sage-310/src/sage/libs/gap/element.pyx:390, in sage.libs.gap.element._from_sage()
    388 
    389 
--> 390 cpdef _from_sage(elem) noexcept:
    391     """
    392     Currently just used for unpickling; equivalent to calling ``libgap(elem)``

File /usr/local/sage-310/src/sage/libs/gap/element.pyx:396, in sage.libs.gap.element._from_sage()
    394 """
    395 if isinstance(elem, str):
--> 396     return libgap.eval(elem)
    397 
    398 return libgap(elem)

File /usr/local/sage-310/src/sage/libs/gap/libgap.pyx:406, in sage.libs.gap.libgap.Gap.eval()
    404 
    405         initialize()
--> 406         elem = make_any_gap_element(self, gap_eval(gap_command))
    407 
    408         # If the element is NULL just return None instead

File /usr/local/sage-310/src/sage/libs/gap/util.pyx:364, in sage.libs.gap.util.gap_eval()
    362 try:
    363     GAP_Enter()
--> 364     result = GAP_EvalString(cmd)
    365     # We can assume that the result object is a GAP PList (plain list)
    366     # and we should use functions for PLists directly for now; see

GAPError: Error, Variable: 'x' must have a value
```

### Expected Behavior

It should be possible to save and load any finitely presented group. It is possible if it is created directly.

### Actual Behavior

If the group comes from `wrap_FpGroup` a libgap group, the above error happens.

### Additional Information

_No response_

### Environment

```markdown
- **Fedora 39**:
- **Sage Version*: 10.2*:
```

### Checklist

enriqueartal:
At this point it is not possible to load a saved finitely presented group that comes from a libgap group, see an example in https://github.com/sagemath/sage/issues/37061

One possible cause is the use of a general `__reduce__` method for free groups. At least, adding such a method allows to load free groups or finitely presented groups obtained from a libgap group using `wrapFreeGroup` or `wrapFpGroup`. It fixes #37061 and it would simplify some code in https://github.com/sagemath/sage/pull/36768

With these changes, free and finitely presented groups, included libgap groups, can be pickled.

### :memo: Checklist

- [X] I have updated the documentation accordingly.


tscrim:
A bit more detail about what is going wrong here. The problem is that when using the wrap function for a free group, it already knows what its libgap group is, so it passes that along as part of its construction data. This means it is becomes part of its pickling data as it is part of its input (which defines itself as a unique representation). This also causes two free groups to be constructed on the same generators that are **not** equal (as they are not identical).

IMO, we should remove having the (lib)GAP group as an input option for free groups. Likewise, I think we should remove the `wrap_FreeGroup` and `wrap_FpGroup` functions. The former is only called by the latter, which is only called in one place (`schemes/curves/zariski_vankampen.py`) that doesn't even return anything related to the constructed object! Everything is being done in GAP, so just convert the data from GAP over. Not only will this solve the issue at hand, it will prevent other subtleties from happening if this function gets used elsewhere.

enriqueartal:
> A bit more detail about what is going wrong here. The problem is that when using the wrap function for a free group, it already knows what its libgap group is, so it passes that along as part of its construction data. This means it is becomes part of its pickling data as it is part of its input (which defines itself as a unique representation). This also causes two free groups to be constructed on the same generators that are **not** equal (as they are not identical).
> 
> IMO, we should remove having the (lib)GAP group as an input option for free groups. 

I agree with this, may the opinion of @miguelmarco who created it would be interesting.

> Likewise, I think we should remove the `wrap_FreeGroup` and `wrap_FpGroup` functions. The former is only called by the latter, which is only called in one place (`schemes/curves/zariski_vankampen.py`) that doesn't even return anything related to the constructed object! Everything is being done in GAP, so just convert the data from GAP over. Not only will this solve the issue at hand, it will prevent other subtleties from happening if this function gets used elsewhere.

I am not sure this is a good idea to eliminate `wrap_FpGroup`, it appears also in other functions in `groups/finitely_presented.py` and obtaining presentation for new groups is better down by `gap` and something that converts a finitely presented group  in `libgap` into such a group in sage is necessary. One case that may appear, give the presentation of the kernel of a homomorphism into a finite group. Of course, may be another code for this function may also help.

tscrim:
We should not have as part of the input data of a free group the libgap group. It creates needless complexity with breaking the unique representation. However, we can add a `__classcall_private__` that takes the libgap group input, extracts the necessary data, and sets the `_libgap` to the passed in group after the object is created. Although I don't think there is any benefit for free groups to this and it could subtly modify the unique object in (Sage's) memory; GAP doesn't do this uniqueness.

The `_reduction` and `__dict__` change as part of the `UniqueRepresentation` as I mentioned above.

For FP groups, this could be more useful. Although I don't know how sensitive GAP is to using different (but really equivalent) free groups for the FP groups. However, the methods `wrap_*` is not a good approach; it should be handled by the class's construction (either `__init__` or more likely `__classcall_private__`). However, we need to take care of the uniqueness (possibly downgrading to `CachedRepresentation`, but that feels like a step backwards to me). We might need to think carefully about how to do this if we really want to support this functionality.

enriqueartal:
I made a couple of commits:
- `__init__` of `FreeGroup_class` does not have anymore the keyword `libgap_free_group`; the reconstruction of the libgap free group is straightforward.
- `wrap_FreeGroup` is also gone
- `__reduce__` of `FreeGroup_class` is gone, the former issues were caused by `libgap_free_group`
- The function `wrap_FpGroup` has been accordingly changed. Now `G` and `wrap_FreeGroup(libgap(G))` are equal, and `save`-`load` works. I only see one option to eliminate this function: to create a `sage` method for libgap finitely presented groups, if it is possible. There are many places where the only option to create some f.p. groups is passing through libgap.
If there is a better solution, please make a concrete one.

miguelmarco:
> We should not have as part of the input data of a free group the libgap group. It creates needless complexity with breaking the unique representation. However, we can add a `__classcall_private__` that takes the libgap group input, extracts the necessary data, and sets the `_libgap` to the passed in group after the object is created. Although I don't think there is any benefit for free groups to this and it could subtly modify the unique object in (Sage's) memory; GAP doesn't do this uniqueness.

That sounds interesting, but I didn't know about the ' __classcall_private__` method until now. How would be the interface in this approach?

enriqueartal:
While there are still some issues, with the last changes, there is no more `wrap_FpGroup`, its functionality is done with a `sage` method. After discussions with @miguelmarco, I think there is at least one issue (which basically exists in the develop branch): if a sage f.p. group is created from a libgap one, another libgap group is created in the process. The same happens in this version for free groups, but I think this is not an issue, since a gap free group does not contain a lot of extra information, while the original libgap f.p. group  may have information which will be lost, and the gestion of the memory is not optimal.
I do not know if there are solutions for these issues starting from here, or a new approach is needed. Anyway, the original problem does not happen any more.

enriqueartal:
Should the class `GroupMorphismWithGensImages` be based on `hom` in `sage/structure/parent.pyx`? The information is basically the same but much more methods are available, e.g., `kernel` from `groups/libgap_morphism.py`.

enriqueartal:
I tried some tricks for homomorphisms and it seems that it is related to the problem pointed out here by @tscrim, for `G` a libgap group, `G1 = G.sage()` and `G2 = G1.gap()`, we do not have equality between `G` and `G2`. I have no idea how to do it.

enriqueartal:
Finally I added the homomorphisms to this branch, the change is small and it is related to a couple of bugs that could be solved. The first one concerns some errors in `hom` caused to the different order in `G.gens()` and `G.gap().GeneratorsOfGroup()`; this happens for example for symmetric and alternating groups. Since `hom` uses `GroupHomomorphismByImages` the morphism constructed is not the one expected. Either the output is an Error or worse it silently gives a wrong result.

There is another issue that should be easy to be solved but I do not know exactly where it must be done.  Run
```
sage: SymmetricGroup(4).gap().IsomorphismFpGroup()
[ (1,2), (2,3), (3,4) ] -> [ S_4.1, S_4.2, S_4.3 ]
```
The underscores in the names produce an error for `SymmetricGroup(4).gap().IsomorphismFpGroup().sage()` (it happened already with `wrap_FpGroup`. This is less serious, It seems `SymmetricGroup(4).gap().IsomorphismFpGroup()` does not like underscores.

tscrim:
Quick comments while playing catchup:

- All `__reduce__` methods should not and never be there. `UniqueRrepresentation` handles all of that. If you need to add that, you have a bug in your code or are doing something wrong. There will never be any reason to override this as you could create very subtle problems with pickles in different sessions. I really can't stress this enough.
- `Parent` also has a standard mechanism/argument for handling `names` rather than manually setting them. Although there is no point in storing them when they aren't called anywhere (see the previous point).
- In the `sage()` method, I would put the `IsFpGroup()` test after the `IsList()` test since the latter is much more likely. While this test is fast, it could be called in tight loops. Or does the check not commute?
- I don't understand your claim about the morphisms and why they even need to be changed. In current vanilla Sage, they are defined using GAP on every element. You are likely doing far more work than necessary since you already have the libgap morphism at hand, so you can just pass that to the appropriate `Hom` set. Is this related with equality in gap with FP groups not having any useful equality?

enriqueartal:
> Quick comments while playing catchup:
> 
> * All `__reduce__` methods should not and never be there. `UniqueRrepresentation` handles all of that. If you need to add that, you have a bug in your code or are doing something wrong. There will never be any reason to override this as you could create very subtle problems with pickles in different sessions. I really can't stress this enough.

I have no doubt that my code can be improved. But the original problem came from that `__reduce__`  from `UniqueRepresentation` introduces `GAP` elements and then saving is useless. It is possible that my solution may create problems, what is sure is that the as it is now, actual problems exist. Which is the alternative? Actually, it is suggested in the doc of `ParentLibGAP` in `libgap_wrapper.py` to create special `__reduce__` to avoide unwanted `GAP` objects.
> * `Parent` also has a standard mechanism/argument for handling `names` rather than manually setting them. Although there is no point in storing them when they aren't called anywhere (see the previous point).
I do not know how to do this, my goal was to keep the output of the new `__reduce__` as close to the old ones as possible. Improvements are welcome.
> * In the `sage()` method, I would put the `IsFpGroup()` test after the `IsList()` test since the latter is much more likely. While this test is fast, it could be called in tight loops. Or does the check not commute?
This is on me, since `IsList` was the last one I thought it was designed in purpose. So, is it enough to switch the last ones or it would be better to place `IsList` higher in the list?
> * I don't understand your claim about the morphisms and why they even need to be changed. In current vanilla Sage, they are defined using GAP on every element. You are likely doing far more work than necessary since you already have the libgap morphism at hand, so you can just pass that to the appropriate `Hom` set. Is this related with equality in gap with FP groups not having any useful equality?
I guess I did not come with the best solution, but this is the point. The method `hom` for groups have several useful methods, e.g., `kernel`; the output of `simplification_isomorphism` and `epimorphism` gave generic morphisms in a class with no useful methods. 
There was also some small issue with the method `hom` for groups; it essentally uses `GroupHomomorphismByImages` and if the `GAP` group of a `Sage` group has the generators with a distinct ordering, which is the case for `AlternatingGroups`, can produce wrong mathematical results. 

Coming back to the first comment. I assume the bad quality of my code but I disagree that `UniqueRepresentation` handles correctly the situation, this is in fact the origin of this PR and the `ParentLibGAP` suggests explicitely that `__reduce__` methods should be implemented. If these methods create problems they would be local; i do not dare to change the global `__teduce__` method, which could create global problems.

tscrim:
No, `UniqueRepresentation` is working properly. You are breaking the pickling aspect of it by giving it an object that doesn't pickle well. Likewise, the gap group should *not* be part of the key information, in part because they have horrible uniqueness properties and having two objects that look the same but otherwise differ by some backend object should clearly be a code smell. The key point is that a group with a fixed finite presentation is meant to (and should IMO) be unique within Sage.

Knowing how `UniqueRepresentation` is and what it representations and does (I believe I have explained this, but if not, please read its documentation), you should follow the `ParentLibGAP`'s suggestion by not having the GAP group being part of the `__reduce__` information. Again, this means that it should *not* be part of the construction information. You have misunderstood what `ParentLibGAP` is saying about `__reduce__`.

If you want to graph the GAP group onto a FP group, you need to do something like
```python
def __classcall_private__(cls, data, gap_group):
    G = super().__classcall__(cls, data)
    G._gap = gap_group
    return G
```
However, you might have issues if you overwrite the GAP group with a different one. Hence objects could potentially become inconsistent (i.e., using different GAP groups).

Another option is to use `CachedRepresentation` or not even use this. That opens a potentially different can of worms with equal-but-not-identical parents (which you need to then implement your own comparisons). The other groups in Sage that use GAP tend to use this setup. Since you generally don't have many coercions going on, this might be the safer route. My not-so-involved viewpoint is this seems like a lot of work for a feature that only has a potential benefit.

enriqueartal:
Thanks for the ideas. I will try to use them.

enriqueartal:
I tried this for `FreeGroup_class` (with several other places for `ParentLibGAP.__init__(G, libgap_free_group)`) without success. I guess I miss something.
```
    @staticmethod
    def __classcall_private__(cls, generator_names, libgap_free_group=None):
        G = super().__classcall__(cls, generator_names)
        if libgap_free_group is None:
            libgap_free_group = libgap.FreeGroup(generator_names)
        G._libgap = libgap_free_group
        ParentLibGAP.__init__(G, libgap_free_group)
        return G

    def __init__(self, generator_names):
        self._assign_names(generator_names)
        if not generator_names:
            cat = Groups().Finite()
        else:
            cat = Groups().Infinite()
        Group.__init__(self, category=cat)
```
Another comment from the situation. At this moment in the develop and master the gap free group is part of the key information and this is causing the trouble. Changing the code does not look for a potential benefit but for an actual one. 
On the other side, for both free and f.p. groups, gap groups are created and usually contain more information that sage groups. If it is not stored somewhere the same gap group may be recreated several times with two bad consequences: a lot of occupied memory, and as you said, the same *math* objects but different gap objects leading to issues (I encountered them in the use of `hom`). From my bad code, I got a usable __reduce__ and a unique gap group associated to the free or f.p. group. This is what I wanted to keep with a better code.

tscrim:
I am not quite sure what you mean, but it makes me worried that I need to remind you that no solution should have a custom `__reduce__` for `UniqueRepresentation`. It *will* create very subtle problems.

enriqueartal:
> I am not quite sure what you mean, but it makes me worried that I need to remind you that no solution should have a custom `__reduce__` for `UniqueRepresentation`. It _will_ create very subtle problems.

What I would like is to have attach a unique gap group to a sage group; I understand that it would be possible using `__system_classcall_private__` but I do not know how. For your last statement, at least braid groups have a custom `__reduce__`, say `braid.py`.

tscrim:
> > I am not quite sure what you mean, but it makes me worried that I need to remind you that no solution should have a custom `__reduce__` for `UniqueRepresentation`. It _will_ create very subtle problems.
> 
> What I would like is to have attach a unique gap group to a sage group; I understand that it would be possible using `__system_classcall_private__` but I do not know how.

This is basically impossible with `UniqueRepresentation` and allowing a GAP group as input. You're going to run into issues like `loads(dumps(G)) == G` being false when `G` is constructed with a GAP group if `G` part of its construction key (basically you get unequal but clearly meant to be the same FP group). So you're left having many copies in GAP of the same group (which do not compare as equal even in obvious situations, which is bad for GAP IMO and a key point of our problem here; GAP also seems to not to utilize any uniqueness), but Sage only having one.

So if you really want a one-to-one mapping, you need to drop `UniqueRepresentation` for `CachedRepresentation`. This means you implement your own `__reduce__`, `__richcmp__`, `__hash__`, and maybe other such methods. I am not convinced you get any benefits from having this one-to-one mapping. There is also a separate but potentially related issue of how to keep the GAP FP group's free group in sync with the Sage FP group's free group. (So that might also have to loose its `UniqueRepresentation`!)

> For your last statement, at least braid groups have a custom `__reduce__`, say `braid.py`.

This is bad and should be removed. It is doing *exactly* the same thing as provided by `UniqueRepresentation`; it is completely redundant.

enriqueartal:
If a gap group is not an input of a sage group we have several advantages. No custom `__reduce__` method is needed, one can save and load groups with no problem. But we have also some problems. It is not strange that one constructs a sage group from a gap group. Then, internally sage creates a copy of the gap group which is not the same one. We are duplicating computations, memory and, worse, the objects belonging to the original gap group are not in the new one, breaking, for example the internal use of `GroupHomomorphismByImages`.

If a gap group is part of the input, the above issues do not happen. But, without a custom `__reduce__` pickling becomes impossible. What I did is to define a  `__reduce__` whose output is the same as the former one when the group is not part of the output. I agree with you is not elegant, but I disagree with you that it is useless, it solves a problem. It may create other ones, this is why I tried to reproduce the original output. For braids I think it is useless as it is, but AFAIK it is there for a while without causing issues. Are we sure that switching to `CachedPresentation` will not create more problems? We have to define much more custom methods.

I do not know which is the optimal solution, but I am convinced that it should not be kept as it is.

enriqueartal:
In a separate branch I am trying to check for issues if we replace `UniqueRepresentation` by `CachedRepresentation` for free and f.p. groups. I have two questions:
- Is there a good place to learn how to define the new methods?
- Besides, the methods you proposed to custom, how to figure out if other ones are needed?

Thanks.

tscrim:
> If a gap group is not an input of a sage group we have several advantages. No custom `__reduce__` method is needed, one can save and load groups with no problem.

This is misleading at best. Yet again, if you have `UniqueRepresentation`, you never should have a custom `__reduce__`, but it needs to be associated to a unique GAP group that cannot be part of the pickle (and hence key). However, you don't need to have a group be a `UniqueRepresentation`.

> But we have also some problems. It is not strange that one constructs a sage group from a gap group. Then, internally sage creates a copy of the gap group which is not the same one. We are duplicating computations, memory and, worse, the objects belonging to the original gap group are not in the new one, breaking, for example the internal use of `GroupHomomorphismByImages`.

I am not sure I quite agree with the last part. I don't know how much we are actually duplicating; this needs actual evidence (for things that are expensive to compute). GAP seems to duplicate this stuff anyways (which is a big drawback of GAP for this IMO,). This likely could be improved by doing more by utilizing Sage's uniqueness properties, but more care might be needed with how things are constructed. It might also need improvements on GAP's side with at least equality checking the obvious cases (same generators and relations).

> If a gap group is part of the input, the above issues do not happen. But, without a custom `__reduce__` pickling becomes impossible. What I did is to define a `__reduce__` whose output is the same as the former one when the group is not part of the output. I agree with you is not elegant, but I disagree with you that it is useless, it solves a problem.

Please stop making these type of statements. It is not about elegance. It is just wrong. You introduce (subtle) bugs. Pickling gets broken (I am fairly certain `loads(dumps(G)) == G` would break). You effectively break the unique representation property (especially if the group is built based upon input the user never passed).

> It may create other ones, this is why I tried to reproduce the original output.

Indeed, it just creates them. No problems are actually solved. Moving dust hidden under one rug to under another isn't cleaning.

> For braids I think it is useless as it is, but AFAIK it is there for a while without causing issues.

It should be removed. It doesn't cause issues currently, but all it takes is a change in the input format. In other words, it creates maintenance burden with code duplication.

> Are we sure that switching to `CachedPresentation` will not create more problems? We have to define much more custom methods.

This is misleading. It does not provide these methods. `CachedRepresentation` means that if given the same inputs, then return the same object (in memory), but it does not mean the input has to uniquely specify the object. So different inputs can give equal objects. So computations cannot be passed between objects like how GAP works.

> I do not know which is the optimal solution, but I am convinced that it should not be kept as it is.

I would just get rid of giving a passed libgap group input to build the Sage group unless you can demonstrate a real need for them. Of course, you can take that input and parse it to build the Sage group, but it wouldn't be associated with the Sage object. This involves the minimal amount of code and would work with the uniqueness that we want with free groups.

On the other hand, if you put in the work, using `CachedRepresentation` would be in line with permutation groups (in particular, those that we expect more uniqueness properties for; e.g., symmetric groups). It would allow libgap groups as input. However, I would want an actual use-case for this.

tscrim:
> In a separate branch I am trying to check for issues if we replace `UniqueRepresentation` by `CachedRepresentation` for free and f.p. groups. I have two questions:
> 
>     * Is there a good place to learn how to define the new methods?

All are basic methods that a class would need (`__richcmp__` is slightly special, but elements use it; you might need a class-level decorator though).

>     * Besides, the methods you proposed to custom, how to figure out if other ones are needed?

`TestSuite(G).run()` should pass as a first. The second is that equal but different (say, with different libgap group inputs) behave in the same way that you expect them to, as well as respect Python's requirements (equal objects imply equal hashes).

enriqueartal:
Without defining custom methods, besides `__reduce__`, `TestSuite`'s pass

enriqueartal:
I have just pushed a new version where `FreeGroup_class` and `FinitelyPresentedGroup` come from `CachedRepresentation`. I have created `__hash__` and `_richcmp_` methods but I am quite sure that they are far from optimal.
I saw some differences that were created more probably from `__reduce__`. In this branch:
```
sage: F = FreeGroup(2)
sage: F._reduction
(<function dynamic_class at 0x7fb7d04f1bc0>,
 ('FreeGroup_class_with_category',
  (<class 'sage.groups.free_group.FreeGroup_class'>,
   <class 'sage.categories.category.JoinCategory.parent_class'>),
  None,
  None,
  <class 'sage.groups.free_group.FreeGroup_class'>))
sage: G = F / [(1, 2, 2, 1)]
sage: G._reduction
(<function dynamic_class at 0x7fb7d04f1bc0>,
 ('FinitelyPresentedGroup_with_category',
  (<class 'sage.groups.finitely_presented.FinitelyPresentedGroup'>,
   <class 'sage.categories.groups.Groups.parent_class'>),
  None,
  None,
  <class 'sage.groups.finitely_presented.FinitelyPresentedGroup'>))
```
With the stable branch:
```
sage: F = FreeGroup(2)
sage: F._reduction
(<class 'sage.groups.free_group.FreeGroup_class'>, (('x0', 'x1'),), {})
sage: G = F / [(1, 2, 2, 1)]
sage: G._reduction
(<class 'sage.groups.finitely_presented.FinitelyPresentedGroup'>,
 (Free Group on generators {x0, x1}, (x0*x1^2*x0,)),
 {})
```
There is also something strange. In two computers, the test of `sage.libs.gap.element.pyx` has one failure in the first test of `__call__` in `GapElement_Function`:
```
            sage: a = libgap.NormalSubgroups
            sage: b = libgap.SymmetricGroup(4)
            sage: libgap.collect()
            sage: a
            <Gap function "NormalSubgroups">
            sage: b
            Sym( [ 1 .. 4 ] )
            sage: sorted(a(b))
            [Group(()),
             Sym( [ 1 .. 4 ] ),
             Alt( [ 1 .. 4 ] ),
             Group([ (1,4)(2,3), (1,2)(3,4) ])]
```
The last result yields is `Group([ (1,4)(2,3), (1,3)(2,4) ])` so the result is correct but the systems of generators differ. Next tests are similar but work. What is more strange is that in the command line, those orders give the expected result.

tscrim (changes requested):
We are getting there. I don't really like free groups being a cached representation, but we are stuck with GAP's nonuniqueness choice and for FP groups. There still is a code smell here, but it is basically coming from GAP now. So we would need to change GAP to have better equality/uniqueness behaviors with FP and free groups to get rid of it.

Now some additional comments, but I think this will bring us to a positive review.

The free groups should also take in the libgap group, otherwise there is no point in downgrading them from a `UniqueRepresentation`. Subsequently, the libgap group should be passed in the GAP FPGroup `sage()` conversion code, and an analogous `sage()` conversion should be added for free groups. Although if there are actually no problems with having a unique free group within Sage, then make free groups back into a `UniqueRepresentation` (which I think is better).

Remove all changes in `braid.py`, `cubic_braid.py`, `permgroup_named.py`, and `raag.py` modulo one possible change: add `UniqueRepresentation` as the first class in the MRO. Your changes otherwise are not needed and just making the code more complicated. (Adding `UniqueRepresentation` puts them back with the uniqueness they deserve, and first in the MRO means the `UR.__reduce__` method will be used for the pickling.)

Your `_richcmp_` should be `__richcmp__` and you need a class-level decorator `@richcmp_method`. You should also properly test your comparisons that are not based on the cache: compare things with different libgap groups.

You should very clearly document in the class-level doc that any subclass of FPGroup will need to implement its own `__reduce__` method. I don't really like putting implementation details in public doc, but it is too important to not include it.

enriqueartal:
> We are getting there. I don't really like free groups being a cached representation, but we are stuck with GAP's nonuniqueness choice and for FP groups. There still is a code smell here, but it is basically coming from GAP now. So we would need to change GAP to have better equality/uniqueness behaviors with FP and free groups to get rid of it.
> 
> Now some additional comments, but I think this will bring us to a positive review.
> 
Thanks for the help.

> The free groups should also take in the libgap group, otherwise there is no point in downgrading them from a `UniqueRepresentation`. Subsequently, the libgap group should be passed in the GAP FPGroup `sage()` conversion code, and an analogous `sage()` conversion should be added for free groups. Although if there are actually no problems with having a unique free group within Sage, then make free groups back into a `UniqueRepresentation` (which I think is better).
I think I managed to do this, method `sage` for libgap free groups has been added.
> 
> Remove all changes in `braid.py`, `cubic_braid.py`, `permgroup_named.py`, and `raag.py` modulo one possible change: add `UniqueRepresentation` as the first class in the MRO. Your changes otherwise are not needed and just making the code more complicated. (Adding `UniqueRepresentation` puts them back with the uniqueness they deserve, and first in the MRO means the `UR.__reduce__` method will be used for the pickling.)
I guess you mean `artin.py`, `braid.py` has not been changed. I tried a simple approach for `artin.py`: put `UniqueRepresentation` as the first class for `ArtinGroup` and eliminate the method `__reduce__`; it is strange the result for me, `ArtinGroup.__mro__` gives the expected order (`ArtinGroup`, `UniqueRepresentation`, `FinitelyPresentedGroups`, etc) but `__reduce__` is still the method of `FinitelyPresentedGroups`. So, for now it remains with the particular method.
The change for `permgroup_named.py` was of distinct nature, order of generators of alternating groups in `sage` and `gap` were different, causing errors and wrong results.
> 
> Your `_richcmp_` should be `__richcmp__` and you need a class-level decorator `@richcmp_method`. You should also properly test your comparisons that are not based on the cache: compare things with different libgap groups.
I started to change it, but I think it is far from finished. I add a couple of lines in the method for `FreeGroup_class` to avoid some unexpected errors (some where it looked for a list which does not have `_names` attribute).
> 
> You should very clearly document in the class-level doc that any subclass of FPGroup will need to implement its own `__reduce__` method. I don't really like putting implementation details in public doc, but it is too important to not include it.
I agree but not yet changed.

tscrim:
I think I know what is going wrong with the `UniqueRepresentation`. The `__reduce__` comes from `CachedRepresentation`, leading to the problem. So you need to also put `CachedRepresentation` in the MRO beforehand. The fact we have to do this is part of the code smell.

tscrim:
PS - Your response seems to be mis-formatted.

enriqueartal:
> I think I know what is going wrong with the `UniqueRepresentation`. The `__reduce__` comes from `CachedRepresentation`, leading to the problem. So you need to also put `CachedRepresentation` in the MRO beforehand. The fact we have to do this is part of the code smell.

I will retry but I tried to do it in several orders, it did not work maybe because `FinitelyPresentedGroup` comes from `CachedRepresentation`.


> PS - Your response seems to be mis-formatted.

Now it is formatted, thanks

tscrim:
> > I think I know what is going wrong with the `UniqueRepresentation`. The `__reduce__` comes from `CachedRepresentation`, leading to the problem. So you need to also put `CachedRepresentation` in the MRO beforehand. The fact we have to do this is part of the code smell.
> 
> I will retry but I tried to do it in several orders, it did not work maybe because `FinitelyPresentedGroup` comes from `CachedRepresentation`.

Ah, you're right. It won't come first in the MRO. I think this could simplify things:
```python
__reduce__ = CachedRepresentation.__reduce__
```
This is also running into the same issue with the equality and having to define a `_names` attribute. (Also I did mean `artin.py`.)

We are probably doing this wrong as we are breaking an somewhat implicit contract of `CachedRepresentation`. There are groups that do not have a corresponding GAP group implemented (AFAIK, e.g., the RAAGs), and they should keep their uniqueness properties IMO. The question then becomes deciding which of the following:
1. Are we just abusing being a subclass of FPGroup and change the base class?
2. Do we split the FPGroup into two classes: One into an abstract base class that uses the normal `CachedRepresentation` and the other being the actual `FPGroup` that has the corresponding overrides you have defined.
3. Something else entirely.

> The change for `permgroup_named.py` was of distinct nature, order of generators of alternating groups in `sage` and `gap` were different, causing errors and wrong results.

This should not be happening as the group generators are already set from the libgap group, unless I have misunderstood the code. This should be investigated further.

enriqueartal:
Let us start with the simpler things. I will check later your other remarks.

> Ah, you're right. It won't come first in the MRO. I think this could simplify things:
> 
> ```python
> __reduce__ = CachedRepresentation.__reduce__
> ```

It does!

> This should not be happening as the group generators are already set from the libgap group, unless I have misunderstood the code. This should be investigated further.

That's what happens here:
```
sage: AlternatingGroup(4).gens()
((1,2,3), (2,3,4))
sage: libgap.AlternatingGroup(4).GeneratorsOfGroup()
[ (1,2,3), (2,3,4) ]
```
That's what happens in 10.2:
```
sage: AlternatingGroup(4).gens()
((2,3,4), (1,2,3))
sage: libgap.AlternatingGroup(4).GeneratorsOfGroup()
[ (1,2,3), (2,3,4) ]
```

enriqueartal:
I think that the development of richcmp methods is above my capacity. I will try to learn something more since my attempts create so much problems. 

For the options of creating a new class, I do not know which is the best option. 

For the difference of order in sage and gap groups, I was not able to localize the origin (may some sorted iterators). I do not think so but it may be related with the fact that some tests pass here, but they do not in my computer (only testing, they work in command line, see the last lines in https://github.com/sagemath/sage/pull/37128#issuecomment-1924254134.

Any hint would be welcome.

tscrim:
I will have to investigate further as my quick look at the current code says there shouldn't be any differences.

Some other quick comments:

- Anything that was a `UniqueRepresentation` before should go back to being one. This might also deal with some of the rich comparison issues. It might also be showing us that we were previously abusing inheriting from `FinitelyPresentedGroup`, and we need to change their inheritance. All of these problems we are getting are pointing to this being the correct way to fix this.
- The generic `__richcmp__` you were doing before looked fine. Although perhaps one leading check for the type of the other group also being a `FPGroup` is needed? Comparison methods are definitely necessary without `UniqueRepresentation`.

enriqueartal:
> I will have to investigate further as my quick look at the current code says there shouldn't be any differences.

Actually, there are, it is easily checked.
> 
> Some other quick comments:
> 
> * Anything that was a `UniqueRepresentation` before should go back to being one. This might also deal with some of the rich comparison issues. It might also be showing us that we were previously abusing inheriting from `FinitelyPresentedGroup`, and we need to change their inheritance. All of these problems we are getting are pointing to this being the correct way to fix this.

Apparently one way is to create another class of finitely presented groups. Anyway, even if Artin groups do not have a counterpart in GAP, actually, most computations depend on GAP. So, is it better to use UniqueRepresentation or to have associated a unique GAP group?

I do not know if adding `UniqueRepresentation` as the first class for inheritance is enough or not. 

> * The generic `__richcmp__` you were doing before looked fine. Although perhaps one leading check for the type of the other group also being a `FPGroup` is needed? Comparison methods are definitely necessary without `UniqueRepresentation`.

I am going to retry, but with my first attempts, I think that a free group and its `gap()` were distinct (I mean `==` gives False, not only `is`) so I thought the methods were incomplete and I am not sure how to deal with.

I agree with you that maybe some more structural changes are needed. This is why finally I decided to go back to the use of attributes in #36768

enriqueartal:
Besides some details, the only remaining big point is the `cartan_type` tests. I added some more doctests

enriqueartal:
To summarize the main changes in this pull request. Sage and gap finitely presented and free groups can be pickled and if recreated no new gap objects are created. To do this, the order of inheritance in `UniqueRepresentation` has been changed (idea of @tscrim). Only one test fails for the class `CartanType_simple_finite` in `combinat/root_system/cartan_type.py`. The unpickling problem comes from a procedure to ensure compatibility with sage <=4.0 (by @mwhansen); the tests include some modifications by @jhpalmieri (trac #29348). As @tscrim states in https://groups.google.com/g/sage-devel/c/HJodEn2OMMU other issues may be uncovered by the tests.

i think that this PR means an improvement but it would be nice to hear more experts.

enriqueartal:
In order to go on, I have eliminated the failing tests, it was for compatibility with sage <=4.0

enriqueartal:
This PR has been silent for a while. There are changes about pickling gap objects and it introduces some methods relating gap and sage objects. I wonder if it is has some connections with aspects dealt in #37481 and #37884.

enriqueartal:
@tscrim can you compare the change of ##38097 with the one in this PR for `CachedRepresentation`?

enriqueartal:
This PR has been silent for a while. The only big issue was the failing test for something related with sage<4.0 and I eliminated the test. I wonder if it is worth to try to get to 9.4, and if not, which changes are needed.

src/sage/groups/finitely_presented.py
```diff
@@ -786,6 +745,25 @@ def __init__(self, free_group, relations, category=None):
...
+            sage: A5gapfp
+            <fp group of size 60 on the generators [ A_5.1, A_5.2 ]>
+            sage: A5sage = A5gapfp.sage(); A5sage;
+            Finitely presented group < A_5.1, A_5.2 | A_5.1^5*A_5.2^-5, A_5.1^5*(A_5.2^-1*A_5.1^-1)^2, (A_5.1^-2*A_5.2^2)^2 >
+            sage: A5sage.inject_variables()
+            Traceback (most recent call last):
+            ...
+            ValueError: variable names have not yet been set using self._assign_names(...)
+
+        Check that pickling works
```
tscrim:
```suggestion
        Check that pickling works::
```

src/sage/schemes/curves/zariski_vankampen.py
```diff
@@ -1543,8 +1542,8 @@ def braid2rels(L):
         P.SetTzOptions(dic)
         P.TzGoGo()
         P.TzGoGo()
-        gb = wrap_FpGroup(P.FpGroupPresentation())
-        U = [rel.Tietze() for rel in gb.relations()]
+        gb = P.FpGroupPresentation().sage()
+        U = [_.Tietze() for _ in gb.relations()]
```
tscrim:
```suggestion
        U = [rel.Tietze() for rel in gb.relations()]
```
The `_` is generally meant for cases when the iterating object is not used in the iterator itself, e.g., `[1 for _ in range(5)]`.

src/sage/groups/finitely_presented.py
```diff
@@ -794,11 +772,45 @@ def __init__(self, free_group, relations, category=None):
...
+        """
+        Rich comparison of ``self`` and ``other``.
+
+        EXAMPLES::
+
+            sage: G1 = FreeGroup(2) / [(1, 2, 2, 1, 2, 1)]
+            sage: G2 = FreeGroup(2) / [(1, 2, 2, 1, 2, 1)]
+            sage: G1 == G2
+            True
+        """
```
tscrim:
```suggestion
            True
            sage: G1 is G2
            False
        """
```

src/sage/groups/perm_gps/permgroup_named.py
```diff
@@ -716,6 +716,7 @@ def __init__(self, domain=None):
             sage: h = A.hom(codomain=A, im_gens=A.gens())
         """
         PermutationGroup_symalt.__init__(self, gap_group='AlternatingGroup(%s)' % len(domain), domain=domain)
+        self._gens = tuple(self(g) for g in self.gap().GeneratorsOfGroup())
```
tscrim: Is this still needed?

src/sage/groups/raag.py
```diff
@@ -38,6 +38,7 @@
 from sage.categories.groups import Groups
 from sage.categories.algebras_with_basis import AlgebrasWithBasis
 from sage.algebras.clifford_algebra_element import CohomologyRAAGElement
+from sage.structure.unique_representation import CachedRepresentation
```
tscrim:
```suggestion
```
This is no longer needed.

tscrim (changes requested):
Sorry, I haven't had much time to get to this until now.

I am still really bothered by the change breaking the Cartan type unpickling. I really do not understand why that doesn't work. Hence, I am very worried we could unknowingly be breaking something that isn't tested otherwise. It's probably okay since it is for *very* old pickles. I just wish I knew what was breaking there...

Maybe for better transparency, we should split the `UniqueRepresentation` change off as a separate PR. Would you be okay doing this? Sorry for the extra work.

src/sage/groups/finitely_presented.py
enriqueartal: Actually I got `True` for `G1 is G2`

src/sage/groups/perm_gps/permgroup_named.py
enriqueartal: Apparently not.

enriqueartal:
I am going to prepare a branch with the change in the class. Nevertheless, as far as some more experts take a look it will be difficult to test if some problems appear, they do not in the tests.

src/sage/groups/finitely_presented.py
tscrim: Then currently this test is not good. It should be comparing equal-but-not-identical groups. (Actually, we might want to have a shortcut for identical groups, but that might be a premature optimization.)

enriqueartal: You are right, the groups `G1` and `G2` are defined exactly in the same way. I tried several ways to produce `True` for `==` and `False` for `is` with other definitions but I failed.

tscrim: How about passing the gap group?

enriqueartal: It worked defining `G2 = libgap(G1).sage()`

tscrim:
> I am going to prepare a branch with the change in the class. Nevertheless, as far as some more experts take a look it will be difficult to test if some problems appear, they do not in the tests.

Thank you. Well, there is one test that is failing, but we are removing it. So there is at least something that is breaking in a subtle way. At least with a separate PR, there is a clear point where people can see and comment. We can also clearly explain the reasoning (I can add this into the PR if necessary).

tscrim (approved):
Sorry for the very long wait in getting to this and finishing my review. I have no additional comments. Let's get this in. Again, sorry for not getting this done sooner.

PR #37135, opened 2024-01-22
Compute torsion basis via division polynomials

sferl:
Compute torsion basis via division polynomials

The `torsion_basis(n)` method computes a basis of the $n$-torsion points of an elliptic curve $\mathcal{E}$ over a finite field $\mathbb{F}$, provided that the full $n$-torsion is contained in $\mathbb{F}$.
This function currently supports just one algorithm: compute the full abelian group structure of $\mathcal{E}(\mathbb{F})$, then extract the $n$-torsion subgroup.
This PR implements an alternative algorithm that uses division polynomials.
NOTE#1: documentation, tests, examples yet to write. They will be added shortly.

The algorithm:
Firstly, factor $n$. When $n$ is composite, it reduces to the case where $n$ is a prime power, which in turn reduces to the case where $n$ is prime.
When $n$ is a prime number $q$, the $q$-torsion is computed via `E(0).division_points(n)`, that computes the roots of the $q$-th division polynomial and lifts them to the points on the curve constituting the $q$-torsion, among which we choose a $\mathbb{Z}/q\mathbb{Z}$-basis.
When $n$ is a prime power $q^k$, the algorithm first finds a basis $(P,Q)$ of the $q$-torsion, then iteratively finds $(P', Q')$ such that $q^{k-1}P' = P$, $q^kQ' = Q$.
When $n$ is composite, the algorithm finds a basis $(P_q, Q_q)$ of the $q^k$-torsion for the prime powers $q^k$ that divide $n$ exactly, then sums all $P_q$ into a point $P$ of order $n$, all $Q_q$ into a $Q$ of order $n$, so that $P,Q$ are still independent, hence a basis of the n-torsion points, as desired.

Dependencies:
This PR depends on https://github.com/sagemath/sage/pull/37106: assumes that `min(E(0).division_points(q) == 0` holds.

Contributor's note:
I'm a total beginner in SageMath and GitHub contribution, and I may have missed some (basic) guidelines. If so, please don't hesitate to point them out.


src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1122,14 +1122,53 @@ def torsion_basis(self, n):
...
-        if T.invariants() != (n, n):
-            raise ValueError(f'curve does not have full rational {n}-torsion')
-        return tuple(P.element() for P in T.gens())
+        if algorithm == "abelian_group":
+            T = self.abelian_group().torsion_subgroup(n)
+            if T.invariants() != (n, n):
+                raise ValueError(f'curve does not have full rational {n}-torsion')
+            return tuple(P.element() for P in T.gens())
+        
+        elif algorithm == "division_polynomial":
```
GiacomoPope, line 1171: Currently if a user supplies a wrong algorithm argument this code will fail without proper error handling. See https://github.com/sagemath/sage/blob/3dd953c3aa6b5143071b6c39208199cf128c8080/src/sage/schemes/elliptic_curves/ell_point.py#L1821-L1855 for inspiration on how to handle the user input.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1122,14 +1122,53 @@ def torsion_basis(self, n):
...
+                if len(ltorsion) < l**2:
+                    raise ValueError(f'curve does not have full rational {l}-torsion')
+                
+                # the first element of 0.division_points(l) is 0 itself, the others have order l
+                P = ltorsion[1]
+                for Q in ltorsion[2:]:
+                    if P.weil_pairing(Q, l) != 1:
+                        break
+                else:
+                    assert False # shouldn't happen
```
GiacomoPope: If you are going to do an if, else like this then have an error a user can undersatnd. If False really can never be reached, the code can be removed.

yyyyx4: I disagree with removing it: This kind of assertion is great for bug-catching. It *should* never get executed mathematically speaking, so if this point gets reached anyway, we want some kind of alarm going off to indicate that something's wrong.

I don't disagree that adding a human-readable message might be helpful, though: Something like `assert False, "bug: this code should never be reached"` should do the trick.

(Also, in principle, [a compiler could use such statements for optimizations](https://en.cppreference.com/w/cpp/utility/unreachable), but I don't think CPython does this.)

grhkm21: I have seen this pattern a few times, and I think I have written code like this too. I wonder if this indicates we should have a `UnreachableError` type? (This should be in base Python, but oh well)

yyyyx4: Well, we have that: It's called `AssertionError` 😉.

PR #37137, opened 2024-01-22
some details in infinity "ring"

fchapoton:
some typing annotations, and using the base class `CommutativeRing`

### :memo: Checklist

### DIFF
--- a/src/sage/rings/infinity.py
+++ b/src/sage/rings/infinity.py
@@ -486,8 +486,7 @@ class AnInfinity():
         r"""
         Generate a floating-point infinity.
 
-        The printing of
-        floating-point infinity varies across platforms.
+        The printing of floating-point infinity varies across platforms.
 
         EXAMPLES::
 
--- a/src/sage/rings/infinity.py
+++ b/src/sage/rings/infinity.py
@@ -207,21 +207,22 @@ We check that :trac:`17990` is fixed::
     sage: m.rows()                                                                      # needs sage.modules
     [(+Infinity)]
 """
-#*****************************************************************************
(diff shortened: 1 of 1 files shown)


src/sage/rings/infinity.py
```diff
@@ -481,7 +484,9 @@ def _div_(self, other):
 
     def __float__(self):
         r"""
-        Generate a floating-point infinity.  The printing of
+        Generate a floating-point infinity.
+
+        The printing of
         floating-point infinity varies across platforms.
```
tscrim:
```suggestion
        The printing of floating-point infinity varies across platforms.
```
Looks slightly weird otherwise...

src/sage/rings/infinity.py
```diff
@@ -852,10 +858,10 @@ def _div_(self, other):
             0
         """
         if isinstance(other, UnsignedInfinity):
-            return Integer(0)
+            return Integer(0)  # noqa: F821
```
tscrim, line 860: What does this comment mean?

tscrim (changes requested):
Some small details.

src/sage/rings/infinity.py
fchapoton: this is telling the linter not to warn about `Integer` being not defined (it is lazy imported).

tscrim (approved):
It is totally incomprehensible without an explanation IMO. It suggests to me either we shouldn't use such tools so strictly or we need to tell the tool to ignore something in the tool itself. Its fine in this localized setting for now, but I am sure we will get questions about it later...

Issue #31027, opened 2020-12-08
Make jmol optional


mkoeppe:
<div id="comment:0"></div>

(split out from #30315)

We propose to downgrade jmol to optional.  

It used to be needed for
- ~~HTML docbuild~~ (not any more, as of Sage 10.5.beta2) -> #33507
- ~~sagetex testsuite~~ (not any more, as of Sage 10.5.beta2)
- ~~PDF docbuild~~ (not any more, as of Sage 10.5.beta2)

See discussion: ​https://groups.google.com/g/sage-devel/c/qKqTmLzHAbg/m/6YxQ_qWUBQAJ
 and #33507 for alternative technologies for creating 3D graphics for documentation and sagetex

Depends on #33507
Depends on #33534

CC:  @kcrisman @novoselt @jcamp0x2a @antonio-rojas @kiwifb @seblabbe @egourgoulhon @dimpase

Component: **build**

Author: **Matthias Koeppe**

_Issue created by migration from https://trac.sagemath.org/ticket/31027_

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -8,4 +8,4 @@
 so we make installation of jmol conditional on these.
 
 See discussion: ​https://groups.google.com/g/sage-devel/c/qKqTmLzHAbg/m/6YxQ_qWUBQAJ
-... also for alternative technologies for creating 3D graphics for documentation and sagetex
+ and #33507 for alternative technologies for creating 3D graphics for documentation and sagetex
``````

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -3,8 +3,9 @@
 We propose to downgrade jmol to optional.  
 
 It is still needed for
+- HTML docbuild -> #33507
 - sagetex testsuite
-- PDF docbuild(?)
+- PDF docbuild
 so we make installation of jmol conditional on these.
 
 See discussion: ​https://groups.google.com/g/sage-devel/c/qKqTmLzHAbg/m/6YxQ_qWUBQAJ
``````

orlitzky:
The docbuild is already pseudo-optional via `--disable-doc`. I think both sagetex and jmol are good candidates for being pseuso-optional as well: nothing else in sage needs sagetex at all, and nothing bad happens to a normal user who doesn't have jmol. Some other viewer is used by default, and if you try to _force_ jmol, you get an error. OK.

[...]

PR #37138, opened 2024-01-22
Make sage-conf optional

tobiasdiez:
As noted in #37024 and in #36489, `sage-conf` is actually not needed on a few systems now. For this reason, we make the installation of it optional (as there are no optional install requires as far as I know, this means we remove it from `pyproject.toml` and `setup.cfg`). We also remove it from "install all dependencies via conda" as its not needed there.

### :memo: Checklist

### :hourglass: Dependencies

- https://github.com/sagemath/sage/pull/37796


tobiasdiez (commented):
This is ready for review.

@orlitzky @tornaria another PR you might be interested in it.

mkoeppe:
That sage-conf is a required dependency at build and runtime is a crucial part of the design. It supports the PyPI distribution strategy. This is all explained in the README of sage-conf.

tornaria:
> That sage-conf is a required dependency at build and runtime is a crucial part of the design. It supports the PyPI distribution strategy. This is all explained in the README of sage-conf.

No, it's not. `python -m build -nxw` on unpatched sagemath-standard works just fine, thanks.

mkoeppe:
I'll rephrase. That sage-conf is **declared** a required dependency at build and runtime is a crucial part of the design.

tornaria:
> I'll rephrase. That sage-conf is **declared** a required dependency at build and runtime is a crucial part of the design.

It's **incorrectly** declared a required dependency.

mkoeppe:
It is not meaningful to declare that it is "correct" or "incorrect". We declare this particular dependency as part of a design that supports multiple use cases, not just yours. For installing directly from PyPI, it is crucial that sage-conf is pulled in at build and run time; again, see the README.

tornaria:
> It is not meaningful to declare that it is "correct" or "incorrect". We declare this particular dependency as part of a design that supports multiple use cases, not just yours. For installing directly from PyPI, it is crucial that sage-conf is pulled in at build and run time; again, see the README.

I install directly from pypi. I don't claim you have to not-install sage_conf, you are still free to do that. Of course I'm still free to use `-x` and falsify the assertion that `sage_conf` is a requirement to build or run sagemath.

Maybe there should be a separate distribution on pypi that gets the kitchen sink.

tornaria:
> falsify the assertion that `sage_conf` is a requirement to build or run sagemath.

Just for the record
```shell
$ du -h sage*.tar.gz
85K     sage-setup-10.3b6.tar.gz
19M     sagemath-standard-10.3b6.tar.gz
$ tar xf sagemath-standard-10.3b6.tar.gz 
$ tar -C sagemath-standard-10.3b6 -xf sage-setup-10.3b6.tar.gz sage-setup-10.3b6/sage_setup/ --strip-components=1
$ time SAGE_NUM_THREADS=32 python -m build -nxw sagemath-standard-10.3b6 -o dist 2>&1 | tail -1
Successfully built sagemath_standard-10.3b6-cp312-cp312-linux_x86_64.whl

real    1m25.147s
user    9m40.425s
sys     0m53.292s
$ du -h dist/sagemath_standard-10.3b6-cp312-cp312-linux_x86_64.whl 
307M    dist/sagemath_standard-10.3b6-cp312-cp312-linux_x86_64.whl
```

This wheel works perfectly fine if installed in `/usr`. To work from a venv you need to manually set `SAGE_LOCAL=/usr` because it defaults to `SAGE_VENV` (that should be improved):
```shell
$ python -m venv /tmp/sage-from-pypi --system-site-packages --without-pip
$ source /tmp/sage-from-pypi/bin/activate
(sage-from-pypi) $ time python -m installer dist/sagemath_standard-10.3b6-cp312-cp312-linux_x86_64.whl

real    0m20.339s
user    0m19.484s
sys     0m0.769s
(sage-from-pypi) $ python -m installer dist/conway_polynomials-0.9-py3-none-any.whl
(sage-from-pypi) $ SAGE_LOCAL=/usr sage -tp 32 --all
...
All tests passed!
Total time for all tests: 212.2 seconds
    cpu time: 5378.1 seconds
    cumulative wall time: 5103.6 seconds
Features detected for doctesting: conway_polynomials,cvxopt,database_ellcurves,database_graphs,fpylll,gap_package_atlasrep,gap_package_polycyclic,gfan,imagemagick,lrcalc_python,mpmath,nauty,networkx,numpy,palp,pandoc,pexpect,pillow,polytopes_db,pplpy,primecountpy,ptyprocess,pyparsing,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.schemes,sage.symbolic,scipy,singular,sympy,threejs
```

Appart from installing `conway_polynomials-0.9` by hand in the venv, I didn't have to prepare the system in any way, of course I already had all the dependencies installed, all from distro packages.

Note that a lot of mess can happen if you have a system sage installed because `.../site-packages/sage/__init__.py` is missing, so it's a good idea to `touch /tmp/sage-from-pypi/lib/python3.12/site-packages/sage/__init__.py`.

IMNSHO, sagemath-standard distribution should include `sage/__init__.py`, and also `sage_setup`.

mkoeppe:
> To work from a venv you need to manually set `SAGE_LOCAL=/usr`

Do you know what functionality still depends on SAGE_LOCAL?

tornaria:
> > To work from a venv you need to manually set `SAGE_LOCAL=/usr`
> 
> Do you know what functionality still depends on SAGE_LOCAL?

It's mainly used as PREFIX, so lots of things: gap search paths, database search paths, jmol, threejs, mathjax. Mostly via SAGE_SHARE=SAGE_LOCAL/share but also SAGE_LOCAL/lib (for SAGE_LOCAL/lib/gap/sysinfo.gap).

A problem is that variables are not clearly defined, and most of them were created as part of sage-the-distro (when that was the only thing) and then used for sage-the-lib.

For example: is SAGE_SHARE the place where sage-the-distro install "share" stuff, or is it the place where sage-the-lib goes to search for "share" stuff. Maybe there is an installation where some database lives in /usr/share (used from system) but another database was installed by sage in its own SAGE_LOCAL. Or maybe the user installed it in DOT_SAGE/db, etc.

It seems maybe we need to think a little bit about search paths to make something coherent for different search paths that can accomodate different situations and it's easy to reason about.

Using `sys.prefix` seems the right way to go, except when I'm on a venv the sys.prefix corresponds to the venv, but I want to use stuff from the system prefix (/usr) at least as a fallback (kind of like --system-site-packages, but for other stuff).

mkoeppe:
> It seems maybe we need to think a little bit about search paths to make something coherent for different search paths that can accommodate different situations and it's easy to reason about.

+1

> Using `sys.prefix` seems the right way to go, except when I'm on a venv the sys.prefix corresponds to the venv, but I want to use stuff from the system prefix (/usr) at least as a fallback (kind of like --system-site-packages, but for other stuff).

In a venv, of course there's also `sys.base_prefix`; but it's not clear to me that this would be the right thing to use at all.
For example, suppose I install a python in the prefix `/opt/python3.12`; then probably we do not expect to find stuff other than Python in that prefix. We would likely still want to find things in default search paths in prefixes such as `/usr`, `/usr/local`.

mkoeppe:
On removing this overloading of `SAGE_LOCAL`, see also my previous comment https://github.com/sagemath/sage/pull/36489#discussion_r1375137789

tornaria:
Also: it's NOT OK that `sage-env` is doing this: `export PATH="$SAGE_VENV/bin:$PATH"`.

It's not ok, it's not ok, it's not ok.

Say `SAGE_VENV=/usr`, and `PATH=~/bin:/usr/local/bin:/usr/bin`, then you just broke my whole system (this has nothing to do with sage: it means that if I subshell from sage my environment is screwed up wrt what I expect, which is that binaries run from ~/bin first, then from /usr/local, then from /usr).

Broken, broken, broken.

I hack my way out of that by doing `unset SAGE_VENV` in `sage-env-conf` but I shouldn't have to (worse: nobody tells me to do that, so I learn the hard way).

mkoeppe:
I know, that's #33625

tornaria:
> I agree that this needs a sharper definition. In my view, we should move toward a model where in a typical distro installation (also conda), `SAGE_LOCAL` (and likewise `SAGE_SHARE` as its subdirectory) is _not_ defined. Rather, `SAGE_LOCAL`, when defined, would be exclusively something managed by sage-the-distro (or whatever subset incarnation of it).

This sounds reasonable. We still need some kind of PREFIX, and for some stuff this maybe needs to be a search list of possible prefixes, I don't know.

mkoeppe:
yes, some `SAGE_PREFIX_PATH` or some such

orlitzky:
> yes, some `SAGE_PREFIX_PATH` or some such

Can we please not reinvent the wheel?

```
$ meson setup --help
usage: meson setup [-h] [--prefix PREFIX]
```

mkoeppe:
Hm? My suggestion was for the search list of possible prefixes.

orlitzky:
> Hm? My suggestion was for the search list of possible prefixes.

You don't have to guess though, the user is going to tell you the one right answer when he builds sagelib.

Searching will always get the edge cases wrong. For example, if you're searching for libdir and if you try `/usr/lib` before `/usr/lib64`, then on multilib systems where `/usr/lib` contains 32-bit libraries, you'll pick the wrong one. If I want to use `/home/mjo/usr` instead of `/usr`, you aren't going to guess that, etc.

tornaria:
> > yes, some `SAGE_PREFIX_PATH` or some such
> 
> Can we please not reinvent the wheel?
> 
> ```
> $ meson setup --help
> usage: meson setup [-h] [--prefix PREFIX]
> ```

:+1: 

But how is `PREFIX` used *at runtime* ? This seems ok as a *build time* prefix to search for build time stuff (pkgconfig, headers, libs, etc).

But it seems to me that python wheels are kind of "relocatable". To paraphrase you:
```
$ python -m installer -h
usage: python -m installer [-h] [--destdir path] [--prefix path] [--compile-bytecode level] [--no-compile-bytecode] wheel
```

That is the install `--prefix` (which defaults to `sys.prefix`). But you are not allowed to run meson or anything at install time. In any case, the issue here is what to do when `sys.prefix` is different from the prefix where other packages (gap, databases, etc) are installed.

The question is how much diversity we really need / want to handle in this regard.

orlitzky:
> But how is `PREFIX` used _at runtime_ ? This seems ok as a _build time_ prefix to search for build time stuff (pkgconfig, headers, libs, etc).
> 

It's actually only used at install/runtime. At build time, you're searching the standard system locations or whatever pkg-config tells you. (And at install time, you're really installing to a tree under `--destdir` that _includes_ the prefix.)


> But it seems to me that python wheels are kind of "relocatable". To paraphrase you:
> 
> ```
> $ python -m installer -h
> usage: python -m installer [-h] [--destdir path] [--prefix path] [--compile-bytecode level] [--no-compile-bytecode] wheel
> ```
> 
> That is the install `--prefix` (which defaults to `sys.prefix`). But you are not allowed to run meson or anything at install time. In any case, the issue here is what to do when `sys.prefix` is different from the prefix where other packages (gap, databases, etc) are installed.

The main thing that changed for us with meson-python is exactly that: there's now a real build system that can be invoked from python to handle each of the configuration, build, and install phases. When you `pip install` a meson-python package, it runs meson for you to configure, build, and install the thing. Though you can also run meson yourself, which is much simpler once you get used to it.

The "configure" phase of that would find, say, GAP, via `$PATH` at build time, regardless of the meson `--prefix`. That works because `$PATH` and/or `-lgap` are standard interfaces.

The database packages are trickier, because they're just data files with no build system and that supply no information about themselves via standard interfaces. That's why I packaged conway-polynomials and started on the MOLS data. The end goal is to make them available via _some_ standard interface, such as

1. In python, via a module in the standard python search path
2. In a library, via `-ldatabase`
3. On the CLI, from a tool in `PATH` that can print the data
4. Or via a bare pkg-config file that points to the database

All of those require the databases to be separately packaged with their own build systems. But once they are, they can be found by `meson setup` (or `pip install`), completely independent of where sagelib is going to be installed.

In short, I would stop worrying about the databases because they're just going to mess up your mental model. There's only like three of them left that need to be packaged, and when they're done, it will be a lot easier to see how building sagelib should work.

mkoeppe:
> The main thing that changed for us with meson-python is exactly that: there's now a real build system that can be invoked from python to handle each of the configuration, build, and install phases. 
> When you `pip install` a meson-python package, it runs meson for you to configure, build, and install the thing.

meson-python only installs into a temporary directory and builds a wheel from it. The "prefix" has no meaning, it's just the temporary directory. The wheel is always installable in an arbitrary venv.

orlitzky:
> meson-python only installs into a temporary directory and builds a wheel from it. The "prefix" has no meaning, it's just the temporary directory. The wheel is always installable in an arbitrary venv.

it still invokes meson, though, which uses the `--prefix` to whatever extent you decide. In the context of dependency paths, I think the relevant cases are,

1. If it's not a database package, the sage `--prefix` is irrelevant, because they can all be found via standard interfaces
2. A few databases still can't be found automatically but that's easy to fix by packaging them with a build system

With (2) done, there's no need to coordinate any paths, and `SAGE_SHARE`, `SAGE_LOCAL`, etc. all become irrelevant. We can then use whatever `--prefix`, `--datadir`, etc. the user asks meson to use. And that's one less sage-ism to carry around.

mkoeppe:
> We can then use whatever `--prefix`, `--datadir`, etc. the user asks meson to use.

No, we can't. As I just explained, in meson-python, the prefix is temporary and has no meaning at the runtime of the package.

meson-python is just another PEP-517/660 build backend. If used right, it can have benefits over using setuptools, in particular regarding dependency tracking.

If you are excited about switching to meson because it would allow us to depend on the installation prefix and to do arbitrary things at install time... let me just point out that the packages that have led the development of meson-python (scipy, numpy) don't do that.

tobiasdiez:
> > falsify the assertion that `sage_conf` is a requirement
> 
> Once again, the declaration is there by design; it is not an "assertion", so it cannot be "falsified".

Well, this is nothing for us to design. PEP518 specifies the meaning of the `requires` key as "This key must have a value of a list of strings representing [PEP 508](https://peps.python.org/pep-0508/) dependencies required to execute the build system (currently that means what dependencies are required to execute a setup.py file)." And it's pretty clear that sage_conf is not required in this sense. 

Moreover, nobody prevents you or anyone else to install sage_conf before installing sagelib/sagemath-standard.
I can make this more explicit in the docs if you want. Is the Readme under pkgs the correct place, or is there a better one?

dimpase (approved):

roed314:
I'm undoing @dimpase setting this to positive review; voting on disputed tickets will be available soon.

mkoeppe:
@kcrisman Responding to your suggestion in https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ/m/ciVrZ7x0AQAJ:

To summarize, I confirm my vote of -1.

orlitzky:
+1

The Right Way to generate what is now provided by sage-conf is at build time, using a real build system.

* https://github.com/sagemath/sage/issues/14807
* https://github.com/sagemath/sage/issues/14834
* https://github.com/sagemath/sage/issues/29118
* https://github.com/sagemath/sage/issues/29119

Those earlier efforts failed, but meson now does for python what autotools could not. There's a PR open that would provide the build-time settings from meson while our existing `./configure` checks are being ported:

* https://github.com/sagemath/sage/pull/37136

sage-conf is optional in that case, and very hopefully someday completely irrelevant after all of the checks are ported.

dimpase:
just in case I am still +1 here

mkoeppe:
> The Right Way to generate what is now provided by sage-conf is at build time, using a real build system.

Making declarations about the Right Way and a Real Build System cannot replace a proper discussion.

mkoeppe:
I'm also -1 on the change done here that removes the use of `configure` in the conda configuration. This removes the valuable consistency checking that `configure --enable-system-site-packages` does in this case. 

I mean, this change:
```diff
 - Bootstrap the source tree and install the build prerequisites and the Sage library::

      $ ./bootstrap
      $ pip install --no-build-isolation -v -v --editable ./pkgs/sage-conf_conda ./pkgs/sage-setup
      $ pip install --no-build-isolation -v -v --editable ./pkgs/sage-setup
      $ pip install --no-build-isolation --config-settings editable_mode=compat -v -v --editable ./src

    If you encounter any errors, try to install the ``sage-conf`` package first::

      $ pip install --no-build-isolation -v -v --editable ./pkgs/sage-conf_conda

    and then run the last command again.

  - Verify that Sage has been installed::
```

What is the user supposed to think here? ``sage-conf`` is not needed, but you need it when there are "errors"?

orlitzky:
> > The Right Way to generate what is now provided by sage-conf is at build time, using a real build system.
> 
> Making declarations about the Right Way and a Real Build System cannot replace a proper discussion.

We've been discussing it for over a decade, time's up.

mkoeppe:
@orlitzky I mean, I'm glad that you like my project
- #34630

... but it has nothing to do with sage-conf.

orlitzky:
> ... but it has nothing to do with sage-conf.

Setup.py can't do checks like `./configure` can, and `./configure` doesn't really suport python or cython. Meson does both.

Now, instead of having a third party bootstrap the sage distribution and send me a config file to hand-edit, any part of sage that is using meson can detect the correct configuration itself. Afterwards there is no need for sage-conf. This isn't a crazy proposal -- it's how everything that isn't sage is built.

dimpase:
@saraedum - does this help packaging Sage for conda a bit easier?

it's +1 from me, one way or another.

mkoeppe:
> Setup.py can't do checks like `./configure` can, and `./configure` doesn't really suport python or cython. Meson does both.
> 
> Now, instead of having a third party bootstrap the sage distribution and send me a config file to hand-edit, any part of sage that is using meson can detect the correct configuration itself. Afterwards there is no need for sage-conf.

Perhaps, but that's just talk at the moment. It's premature to make the change to sage-conf here on the PR based on this.

saraedum:
> @saraedum - does this help packaging Sage for conda a bit easier?

It does not make a substantial difference from what I understand.

I tried to make sense of this PR but fail to understand what's going on here. (I had a look at the linked https://github.com/sagemath/sage/pull/37024 and https://github.com/sagemath/sage/pull/36489 but couldn't understand why sage-conf is not actually needed. Also, I checked the README of sage-conf.)

So, sage-conf contains a bunch of environment variables with paths (and some scripts/code to expose them.) When sage-conf is not installed these variables are not set, and most of the time the right thing happens because whatever is on the search paths is doing the right thing. (From a packager point of view, that's how the world should work. You set some generic things, such as PATH, CXXFLAGS, … nothing tries to be smart, and everything works.)

I probably already misunderstood something up to this point. But in which scenario is sage-conf actually needed?

@mkoeppe from your point of view, should sage-conf eventually go away? Or is it needed in the long run?

dimpase:
> > Setup.py can't do checks like `./configure` can, and `./configure` doesn't really suport python or cython. Meson does both.
> > 
> > Now, instead of having a third party bootstrap the sage distribution and send me a config file to hand-edit, any part of sage that is using meson can detect the correct configuration itself. Afterwards there is no need for sage-conf.
> 
> Perhaps, but that's just talk at the moment. It's premature to make the change to sage-conf here on the PR based on this.

It's not just talk. First of all, there are indeed systems for which sage-conf is just not needed, it's not something new.
As well, we have a PR which makes sagelib buildable with Meson, and allows a sage-conf-less install of sagelib.

mkoeppe:
> I tried to make sense of this PR but fail to understand what's going on here. (I had a look at the linked #37024 and #36489 but couldn't understand why sage-conf is not actually needed. Also, I checked the README of sage-conf.)
> 
> So, sage-conf contains a bunch of environment variables with paths (and some scripts/code to expose them.)

That's right. sage-conf gives a way to make such variables persistent in an environment by pure Python packaging means, by just installing (some version of) sage-conf in the environment.

> When sage-conf is not installed these variables are not set, and most of the time the right thing happens because whatever is on the search paths is doing the right thing.

Well, I don't know about "most of the time". The claims that sage-conf is not needed are based on specific simple use cases:
- (Tobias') the all-conda installation method, where all dependencies are installed in the conda-environment, sagelib is built from source and where at runtime (as of today) `SAGE_LOCAL` is the same as the `CONDA_PREFIX`,
- (Michael's and Gonzalo's) building a monolithic Sage library from source on a user's system where everything is installed as a system package at build time and no runtime configuration is needed or even wanted.

All of the above use cases were of course on the radar when we designed sage-conf. But sage-conf (including the fact that it is listed as a build-system requirement and a runtime dependency) is designed to support other important deployment scenarios, in particular PyPI, which have a higher complexity.
- Installing a prebuilt binary sagelib (for example sagemath-* wheels from PyPI) and configuring its runtime environment, by installing a suitable sage-conf next to it.
- Building from an sdist from PyPI and configuring its build-time environment - with a user-friendly default (= configuring and building a SAGE_ROOT + SAGE_LOCAL in ~/.sage).

> (From a packager point of view, that's how the world should work. You set some generic things, such as PATH, CXXFLAGS, … nothing tries to be smart, and everything works.)

Packagers can bring their own sage-conf; which sage-conf is used can be set with standard "pip" techniques, in particular constraints. #37620 (`SAGE_CONF_FILE`, needs review) makes this even easier.

> @mkoeppe from your point of view, should sage-conf eventually go away? Or is it needed in the long run?

It's needed in the long run, but its implementation(s) may be changing.

orlitzky:
>     * (Michael's and Gonzalo's) building a monolithic Sage library from source on a user's system where everything is installed as a system package at build time and no runtime configuration is needed or even wanted.

Runtime configuration is fine, but it shouldn't be required, and it definitely shouldn't be part of a separate package. It's _runtime feature detection_ that I complain about. Nobody likes it when their homework assignments stop working because they uninstalled Inkscape and accidentally removed imagemagick from the system.

> All of the above use cases were of course on the radar when we designed sage-conf. But sage-conf (including the fact that it is listed as a build-system requirement and a runtime dependency) is designed to support other important deployment scenarios, in particular PyPI, which have a higher complexity.
> 
>     * Installing a prebuilt binary sagelib (for example sagemath-* wheels from PyPI) and configuring its runtime environment, by installing a suitable sage-conf next to it.
> 
>     * Building from an sdist from PyPI and configuring its build-time environment - with a user-friendly default (= configuring and building a SAGE_ROOT + SAGE_LOCAL in ~/.sage).

This isn't higher complexity. You want to run your own binary linux distro on pypi. Bad idea, but OK, it will probably help some people. Shipping config files is a solved problem though. With meson, each module can detect its own configuration and write it to a file at build time. Then you ship that file as part of the package, like how all other software does it.

Sage-conf was a band-aid over the fact that our config file comes from sage-the-distro and that autotools doesn't have great python integration. But now, there's no reason to insist on using _one_ config file for the many different sage packages. If one package depends on the config file from another, something's wrong.

mkoeppe:
> You want to run your own binary linux distro on pypi.

Please, Michael, it's a bit undignified to use such polemic characterizations in every single discussion.

Yes, we want to ship binary wheels via PyPI because that's what is expected in the Python "ecosystem".

tornaria:
> I probably already misunderstood something up to this point. But in which scenario is sage-conf actually needed?

@saraedum Here's a sample. I don't have sage-conf and things work (a few unrelated bugs caused by building from sdist):
1. Get sdists from pypi:
```
$ wget -q https://files.pythonhosted.org/packages/source/s/sagemath-standard/sagemath-standard-10.3.tar.gz
$ wget -q https://files.pythonhosted.org/packages/source/s/sage-setup/sage-setup-10.3.tar.gz
$ du -h sage*-10.3.tar.gz
85K	sage-setup-10.3.tar.gz
19M	sagemath-standard-10.3.tar.gz
$ md5sum sage*-10.3.tar.gz
55e0f9bf0d37b23a9e71a4b1f95856d5  sage-setup-10.3.tar.gz
f5289163c328403bea12e4a6b828ffb3  sagemath-standard-10.3.tar.gz
```

2. Untar the sdist; the second line places sage_setup in the correct place (#37287 avoids this step). The third line is a minor patch to solve a bug.
```
$ tar xf sagemath-standard-10.3.tar.gz
$ tar -C sagemath-standard-10.3/ -xf sage-setup-10.3.tar.gz sage-setup-10.3/sage_setup/ --strip-components=1
$ sed -ie '/read_distribution.*tdlib.pyx/s/$/  # needs tdlib/' sagemath-standard-10.3/sage/misc/package_dir.py
```

3. Build from sdist. Note that this is with a hot ccache; without ccache it would be ~ 5m I think.
```
$ time SAGE_NUM_THREADS=32 python -m build -nxw sagemath-standard-10.3/ -o dist 2>&1 | tee -a build.log | tail -1
Successfully built sagemath_standard-10.3-cp312-cp312-linux_x86_64.whl

real	1m26.135s
user	9m20.670s
sys	0m39.099s


$ du -h dist/sagemath_standard-10.3-cp312-cp312-linux_x86_64.whl
308M	dist/sagemath_standard-10.3-cp312-cp312-linux_x86_64.whl
```

4. With the binary wheel we do whatever we want, here we unpack it to a test directory and run it using `PATH` and `PYTHONPATH`. Note that we have to workaround a critical bug #33625.
```
$ python3 -m installer --destdir test/ dist/sagemath_standard-10.3-cp312-cp312-linux_x86_64.whl
$ echo 'unset SAGE_VENV' > test/usr/bin/sage-env-config  ### FIX bug introduced by #30013 see #33625

$ PATH=test/usr/bin:$PATH PYTHONPATH=test/usr/lib/python3.12/site-packages/ sage
│ SageMath version 10.3, Release Date: 2024-03-19                    │
│ Using Python 3.12.1. Type "help()" for help.                       │
sage: quit

$ time PATH=test/usr/bin:$PATH PYTHONPATH=test/usr/lib/python3.12/site-packages/ time sage -tp 64 --all
...
All tests passed!
Total time for all tests: 148.5 seconds
    cpu time: 6737.9 seconds
    cumulative wall time: 6982.5 seconds
Features detected for doctesting: conway_polynomials,cvxopt,database_ellcurves,database_graphs,fpylll,gap_package_atlasrep,gap_package_polycyclic,gfan,graphviz,imagemagick,lrcalc_python,mpmath,nauty,networkx,numpy,palp,pandoc,pexpect,pillow,polytopes_db,pplpy,primecountpy,ptyprocess,pyparsing,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.schemes,sage.symbolic,scipy,singular,sympy,threejs
```

saraedum:
@tornaria thanks for this annotated example. I guess we might not need sage-conf over at conda-forge anymore either then (and can simplify our build script quite a bit. I had already seen that archlinux does much less than we do by now.)

mkoeppe:
@tornaria FWIW, with #37620 (needs review), you will be able to put that `echo 'unset SAGE_VENV'` into SAGE_CONF_ENV_FILE and just use the declared PEP-518 build prerequisites of sagemath-standard, including the stock sage-conf from PyPI.

tornaria:
> @tornaria FWIW, with #37620 (needs review), you will be able to put that `echo 'unset SAGE_VENV'` into SAGE_CONF_ENV_FILE and just use the declared PEP-518 build prerequisites of sagemath-standard, including the stock sage-conf from PyPI.

1. I fail to see the need for a package to be installed to do just nothing
2. sagemath-standard still depends on an exactly matching version of sage-conf, so even if all packages that do nothing should be compatible, they won't be.
3. the `unset SAGE_VENV` is a hack to workaround a bug, let's hope it can be fixed in a different way. Meanwhile, I know how to write a line to a file.

Note that this PR + #37287 would be really good from my point of view (distro packaging) since it will remove (unnecessary for me) dependencies and would work perfectly when one has all dependencies installed in reasonably standard locations (otherwise it will error out giving information of what is missing).

However, I'm not (yet) sure of this PR because it seems it might be a bit disruptive if one uses this workflow and expects sage-conf to actually do something (although I don't know if there are indeed users that depend on this functionality).

mkoeppe:
> 1. I fail to see the need for a package to be installed to do just nothing

I don't know if you will find this helpful, but I have observed that sometimes in distro packaging it is forgotten or dismissed that the metadata of installed Python distribution packages is part of the API. It is relevant for the interop of distro-installed packages and user-installed packages.

tornaria:
> > 1. I fail to see the need for a package to be installed to do just nothing
> 
> I don't know if you will find this helpful, but I have observed that sometimes in distro packaging it is forgotten or dismissed that the metadata of installed Python distribution packages is part of the API. It is relevant for the interop of distro-installed packages and user-installed packages.

Yes, we do ack that, and we do keep the metadata.

We try to respect python packaging conventions as much as possible, but it's also important that python packaging respects distribution. I think it's been getting much better in recent years. Pep 517 is really a step forward, and other decisions (e.g. python ack'ing "externally managed" environments -- it's their recommendation that "pip install" is disabled on distros).

For configuration, though, I don't think having a separate package is good shape. Some configuration options could be "user serviceable" and can go in files in `/etc` or `$HOME` (to be overrided by end user). Other configuration options are tied to the build of the package itself, e.g. `/usr/lib/gap/sysinfo.gap`, this is a way the package has to record information that depends on build options that is relevant for runtime. But this information comes together with the actual package, this is where it makes most sense.

I think another proposal was to turn sage-conf into a `sage.config` package or something like this, to be built and shipped together with `sagemath-standard` (this is not incompatible with modularization: you could have the configuration split in different files shipped by the different distributions).

mkoeppe:
> For configuration, though, I don't think having a separate package is good shape. Some configuration options could be "user serviceable" and can go in files in `/etc` or `$HOME` (to be overrided by end user).

I disagree. The problem with using files is exactly that it is not "venv-able". There is no way to make a configuration apply to an installation in a venv in a declarative way such as `requirements.txt` files or `environment.yml` files etc. This is what sage-conf solves.

> I think another proposal was to turn sage-conf into a `sage.config` package or something like this, to be built and shipped together with `sagemath-standard` (this is not incompatible with modularization: you could have the configuration split in different files shipped by the different distributions).

I agree that there can be value in having multiple partial configurations and some configurations that are prepared directly by a distribution at its build time (see https://github.com/sagemath/sage/issues/21707#issuecomment-1417845042)

But I think you may be missing that there is also possible value in applying a configuration for the runtime to a package that has already been built -- for example if it comes from a binary wheel.

sage-conf is really designed with a plurality of use cases in mind. Granted, in simple situations it may seem as unnecessary decoration or perhaps complexity; but the price to pay is really very small.

mkoeppe:
See also my comment in
- https://github.com/sagemath/sage/pull/37136#issuecomment-1904584943 ("I think we can come up with a solution that aligns with the 0-1-infinity rule")

That PR adds a 2nd configuration module with identical functionality to sage-conf as an "alternative" to sage-conf - in an obvious violation of the 0-1-infinity rule. A proper design would acknowledge the modularization and support an arbitrary number of configuration mix-ins instead of posing a dichotomy.

orlitzky:
> But I think you may be missing that there is also possible value in applying a configuration for the runtime to a package that has already been built -- for example if it comes from a binary wheel.

No one is missing that. At runtime, you are free to edit the configuration that was generated for you at build time. All other software works the same way. At build time, the build system substitutes bindir, libdir, runstatedir, etc. into `foo.conf`. Then, at runtime, you edit `foo.conf` if you want.

sage-conf isn't solving any new problems with its awkward design, only causing new problems.

mkoeppe:
Please, @orlitzky, "awkward design" is an empty fighting word, and "causing new problems" is unsubstantiated. 

It is necessary to raise the level of discourse here.

mkoeppe:
I've heard that downstream packagers would like to be able to use sage-conf from the PyPI sdist, instead of bringing their own version as suggested in https://pypi.org/project/sage-conf/

I've created this PR to address this:
- https://github.com/sagemath/sage/pull/37620

What other problems are there with sage-conf, if any?

orlitzky:
I'm not doing this any more. Let's vote.

mkoeppe:
Doing "what" any more, Michael?

dimpase:
I am for making stuff optional. Especially awkward (which is a good characteristic for design) one. +1

dimpase:
> I've heard that downstream packagers would like to be able to use sage-conf from the PyPI sdist, instead of bringing their own version as suggested in https://pypi.org/project/sage-conf/
> 
> I've created this PR to address this:
> - https://github.com/sagemath/sage/pull/37620
> 
> What other problems are there with sage-conf, if any?
> 

"I've heard" - where, from who?
People here tell you the opposite, quite uniformly - is #37620 just an attempt at gaslighting ?

kiwifb:
> 
> "I've heard" - where, from who? 

Me. I am not brewing my own to replace an upstream package that does not play ball properly.

culler:
Here is my review of this PR.  The PR proposes to make the sage-conf package optional because there are some scenarios when it is not needed.  The existence of such scenarios is not in dispute.  Matthias lists two of them.  I can add a third which is at the opposite extreme: on a macOS system with no dependencies installed except for autoconf (needed since bootstrap -d cannot be used on a PR) this PR builds without errors.

The rationale of the PR is that if the package is not needed in some scenarios then it should be optional.  I disagree with this because making it an option forces people to decide whether they need the option or not.  They will naturally assume not - options are things that are not required.  If in fact it is required in their specific context then they will be in for a very unpleasant experience trying to figure out what went wrong.

The maxim "if it ain't broke, don't fix it" says that the package should not be optional unless there are no scenarios where it is needed, and there are no planned changes that will create scenarios where it is needed.

The non-existence of such scenarios is in dispute, but I did not see any convincing evidence given that they do not exist now, and there was a clear statement that situations where it is needed will exist with planned future changes.

It might be a different story if making this package a requirement were causing real harm.  But I didn't see any evidence for that.  The objections were unsupported claims that it is "awkward" or somehow a bad design.

For these reasons I do not think that the PR should be merged.  Not now, anyway.

dimpase:
Installing unneeded stuff is bad design. Software Engineering 101, if you like.

culler:
Sure.  And when your collaborators tell you that even though there are some cases where it is unneeded there are others where it is needed, you should try to understand what they are talking about.  Collaboration 101, if you will.

dimpase:
Collaboration with @mkoeppe apparently means doing as he wants.

orlitzky:
> Sure. And when your collaborators tell you that even though there are some cases where it is unneeded there are others where it is needed, you should try to understand what they are talking about. Collaboration 101, if you will.

There are no cases where it is needed. In every case, building sage how everyone else builds everything else would be simpler and solves the same problems.

The case we keep hearing about is binary wheels on pypi, but we are not the first project to ship binary packages to end users. In fact, several of the people pushing for the removal of sage-conf are distribution packagers who spend all day shipping binary packages to end users. It's not that we don't understand.

dimpase:
> Please, @orlitzky, "awkward design" is an empty fighting word, and "causing new problems" is unsubstantiated. 
> 
> It is necessary to raise the level of discourse here.

The discourse is pretty clear here, as well as on most other disputed PRs and issues: @mkoeppe wants a sage-distro lock-in, and monorepo lock-in,  for free. 

It's a labour dispute: we don't want to be forced to contribute our free labour to sage-distro and monorepo, and propose technical means to alleviate this tension. If technical means are not acceptable then, unfortunately, more labour disputes-style means will kick in, such as forming a union, strikes, mass resignations, etc

dimpase:
> Sure.  And when your collaborators tell you that even though there are some cases where it is unneeded there are others where it is needed, you should try to understand what they are talking about.  Collaboration 101, if you will.

it's not collaboration, it's squeezing free labour from people unwilling to give it to a particular job.
Enforced by a "majority".

jhpalmieri:
> > Please, @orlitzky, "awkward design" is an empty fighting word, and "causing new problems" is unsubstantiated.
> > It is necessary to raise the level of discourse here.
> 
> The discourse is pretty clear here, as well as on most other disputed PRs and issues: @mkoeppe wants a sage-distro lock-in, and monorepo lock-in, for free.

In my opinion, it is not appropriate to ascribe motives, desires, etc. to another user. You can describe what _you_ want, of course, but do not tell us what you claim someone else wants.

mkoeppe:
Thanks @culler for the clear analysis and @jhpalmieri for the support.

mkoeppe:
People with expertise and/or interest in PyPI packaging are welcome to join the effort coordinated in:
- #31251

tornaria:
> I've heard that downstream packagers would like to be able to use sage-conf from the PyPI sdist, instead of bringing their own version as suggested in https://pypi.org/project/sage-conf/

sagemath should listen to all downstream packagers, and come up with a solution that works for all of them, and that it's simple enough so that there is a chance of getting sagemath into more distributions.

Some desired features:
 - single sdist to build sagelib (e.g. current sagemath-standard)
 - no tight coupling of dependencies, i.e., split packages is ok, but having exact version matching is not
 - strive for a simple pep517 workflow
 - minimize extra steps and configuration spread in different parts of the build system; every build knob should be handled and documented at the top level of the build system
 - (ideally) straight configuration files (pyproject.toml, etc) instead of indirection via m4
 - a directory structure that matches the git repo so that patches can be taken from git and applied to the sdist (e.g. in the current `sagemath-standard` we have `./sage/` corresponding to git repo `./src/sage/`, this is a headache for maintenance.

>     * [pkgs/sage-conf_pypi: Skip configure/make when `SAGE_CONF_FILE` or `SAGE_CONF_ENV_FILE` are set #37620](https://github.com/sagemath/sage/pull/37620)

Not useful, it makes building more complicated instead of simpler.

> What other problems are there with sage-conf, if any?

 - Unnecessary parts of sage-the-distro leaks into sage-as-a-python-package
 - configure
 - exact version matching, not an independent package.

> People with expertise and/or interest in PyPI packaging are welcome to join the effort coordinated in:
> 
>     * [Meta-ticket: Distribution as wheels #31251](https://github.com/sagemath/sage/issues/31251)

We ack that making distributions takes a lot of effort, that's why we only want one distribution, not a lot of them. With #37287 we could get rid of sage-setup and with this PR we could get rid of sage-conf.


@culler: sage-conf is mostly needed for sage-the-distro. I don't think the idea is to make sage-conf optional for that use case, but only for building sagelib from sagemath-standard python distribution. This is not a self-contained version of sagemath: it needs all the dependencies to be available.

culler:
> @culler: sage-conf is mostly needed for sage-the-distro. I don't think the idea is to make sage-conf optional for that use case, but only for building sagelib from sagemath-standard python distribution. This is not a self-contained version of sagemath: it needs all the dependencies to be available.

This sounds like a start towards an explanation of the issue here.  What about making it optional to NOT have sage-conf as a required depencency?  That is, to have a switch that experts could be expected to know about which would remove sage-conf as a required dependency, but have the default be that it is a required dependency.  Would something like that be feasible?

tornaria:
> > @culler: sage-conf is mostly needed for sage-the-distro. I don't think the idea is to make sage-conf optional for that use case, but only for building sagelib from sagemath-standard python distribution. This is not a self-contained version of sagemath: it needs all the dependencies to be available.
> 
> This sounds like a start towards an explanation of the issue here. What about making it optional to NOT have sage-conf as a required depencency? That is, to have a switch that experts could be expected to know about which would remove sage-conf as a required dependency, but have the default be that it is a required dependency. Would something like that be feasible?

Sure, I workaround using `--skip-dependency-check` so I'm not in such a hurry for this PR but I described above what features would be desired from a distribution packager point of view and the current design of `sage-conf` seems incompatible. I might be able to survive if `sagemath-standard` includes a dependency to `sage-conf` that is not really necessary, but if some change were to make it necessary, that would break my package of sagemath for void linux

AFAIK @antonio-rojas package of sagemath for arch linux also does not use sage-conf and uses a `PYTHONPATH` trick to use `sage-setup` when building `sagemath-standard` (if I'm not mistaken he also builds all the optional distributions at the same time in the same arch pkg; I don't build optional distributions)

Here's one idea: we could create a separate package (sdist) that corresponds to a "standalone" version of sagemath (= sagelib). We can start with what currently is `sagemath-standard`+`sage-setup` and work on that (fix our issues, maybe experiment with a meson build, etc). We can give this a different name so `sagemath-standard` is no longer a source of contention.

culler:
There appears to be a very consistent message from linux packagers that this PR would significantly simplify the task of producing SageMath packages for their distro.   @mkoeppe appears to be saying that other tasks, such as producing binary wheels for PyPi, would be made significantly harder if this PR were merged.  Meanwhile, people like me, who have never done either of those things, are being asked to decide whether the costs for one task exceed the benefits to the other.  (And we have to do this amidst sniper fire and threats.)

@mkoeppe and @tobiasdiez: I think we are owed some clear, specific examples of exactly how this PR would simplify some tasks (and what those tasks are) and how it would make other tasks harder (and what those tasks are).

dimpase:
> Meanwhile, people like me, who have never done either of those things, are being asked to decide whether the costs for one task exceed the benefits to the other.

What's the point of this voting contest? I too can summond a bunch of contributors for voting on this, should I?

mkoeppe:
> > * [pkgs/sage-conf_pypi: Skip configure/make when `SAGE_CONF_FILE` or `SAGE_CONF_ENV_FILE` are set #37620](https://github.com/sagemath/sage/pull/37620)
> 
> Not useful, it makes building more complicated instead of simpler.

What exactly is getting more complicated?

mkoeppe:
> > What other problems are there with sage-conf, if any?
> 
> * Unnecessary parts of sage-the-distro leaks into sage-as-a-python-package
> * configure

Both are exactly what you can disable just by setting an envvar.
- https://github.com/sagemath/sage/pull/37620

Why is this not good enough?

mkoeppe:
> Some desired features:
> * a directory structure that matches the git repo so that patches can be taken from git and applied to the sdist (e.g. in the current `sagemath-standard` we have `./sage/` corresponding to git repo `./src/sage/`, this is a headache for maintenance.

Sure, we can do that by switching the distributions defined in `pkgs/*/` from a flat layout to a `src`-layout.

mkoeppe:
> Some desired features:
> 
> * single sdist to build sagelib (e.g. current sagemath-standard)

Incompatible with the modularization. 

> * strive for a simple pep517 workflow

Everything is pure PEP 517/518 already.
What are you looking for?

mkoeppe:
> Here's one idea: we could create a separate package (sdist) that corresponds to a "standalone" version of sagemath (= sagelib). We can start with what currently is `sagemath-standard`+`sage-setup` and work on that (fix our issues, maybe experiment with a meson build, etc). We can give this a different name so `sagemath-standard` is no longer a source of contention.

@tornaria This does not work because of the constraints of contemporary Python packaging.
- User packages need to be able to declare their dependency on a distribution that provides Sage.
- Build-time requirements and runtime dependencies are keyed 1:1 to distribution names, nothing else. There's no such thing as declaring "Provides" in a distribution, or virtual packages, or anything like that. And there's no such thing as an "OR" in build-time requirements and runtime dependency specifications.

kiwifb:
I'll confess to have skipped a lot of the conversation, mostly because I have better things to do with my life. But if as "sage-on-gentoo", I decide not to ship `sage_conf` at all - and I could probably do all of the bits I desire from it from a small `/etc/env.d/` file (or equivalent on your distro of choice), I would in no way require for it to be marked optional in sage-the-distro. For cleanliness, I would prefer it to disappear from `setup.cfg` in such scenario and that's the only thing I would want.

mkoeppe:
> @culler: sage-conf is mostly needed for sage-the-distro.

No.

> I don't think the idea is to make sage-conf optional for that use case, but only for building sagelib from sagemath-standard python distribution.

Building from the sagemath-standard sdist (from PyPI) is exactly what the required build-time and run-time dependency on sage-conf is there for.

mkoeppe:
> > > * single sdist to build sagelib (e.g. current sagemath-standard)
> > 
> > Incompatible with the modularization.
> 
> Here we reach to the Gordian Knot.

Actually, it's an unknot.

> * single sdist is incompatible with the modularization

Yes.

> * multiple sdist has huge maintenance cost,

Nope.

> and with exact matching versions is incompatible with distro packages

Also nope, that's just something unsubstantiated that you repeat.

kiwifb:
> > I'll confess to have skipped a lot of the conversation, mostly because I have better things to do with my life. But if as "sage-on-gentoo", I decide not to ship `sage_conf` at all - and I could probably do all of the bits I desire from it from a small `/etc/env.d/` file (or equivalent on your distro of choice), I would in no way require for it to be marked optional in sage-the-distro. For cleanliness, I would prefer it to disappear from `setup.cfg` in such scenario and that's the only thing I would want.
> 
> @kiwifb why is sage-on-gentoo not listed here: https://repology.org/project/sagemath/versions ?
> 
> Is it not supported by gentoo? What would it take for sage-on-gentoo to be incorporated in official gentoo?

My work is in an overlay, an additional not official repo. Now, a lot of work has been done already. There are very few packages left, relatively speaking, that needs to migrate to the main tree apart from sage itself. @orlitzky has been instrumental in migrating things to the main tree. 

What's left to do? It would be nice to the remaining databases given the conway-polynomials treatment and sagemath-doc-html and sagemath-doc-pdf finished. For the rest it is now in a great shape I think. The real recent biggie was migrating gap to the main tree, it was possibly a more difficult job.

kiwifb:
Ah, yes kick jmol out. That would be massively helpful.

dimpase:
I've requested reviews from two more Sage "distributors"

orlitzky:
> Is it not supported by gentoo? What would it take for sage-on-gentoo to be incorporated in official gentoo?

The bare minimum is for the remaining databases to be converted to python packages.

While Francois is maintaining it, I'm happy to merge his work. But in the long run (if he gets tired), it would sure help if upstream wasn't openly hostile to anything resembling common sense. I maintain around 200 packages, all built from source on random machines by random people, and keep my bug count in the single digits. I have to be fairly selective when it comes to how stupid an upstream build system can be. I would not bother to package sage myself with the current build system or with what @mkoeppe has planned for it.

mkoeppe:
> I would not bother to package sage myself with the current build system

It's all just PEP 517/518. Just do what is declared and you'll be fine.

tornaria:
> > I would not bother to package sage myself with the current build system
> 
> It's all just PEP 517/518. Just do what is declared and you'll be fine.

You know a lot about python packaging, but you seem to fail to understand the constraints of distributions. Pep 517 is a really good step forward for python packaging, but it's not magic.

mkoeppe:
> the constraints of distributions.

Try to explain them to me, but this time kindly with references to actual distribution policies.

kiwifb:
> > I would not bother to package sage myself with the current build system
> 
> It's all just PEP 517/518. Just do what is declared and you'll be fine.

Well, build system wise I still need this little bit which is annoying
https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sagemath-standard/files/sagemath-standard-9.8-build_ext.patch
because when we build wheels with stuff like `python3.11 -m gpep517 build-wheel --prefix=/usr --backend setuptools.build_meta --output-fd 3 --wheel-dir /home/portage/sci-mathematics/sagemath-standard-9999/work/sagemath-standard-9999-python3_11/wheel` cython files do not compiled without it.

But we are getting wildly out of topic and I would suggest we move further discussion of these details on sage-devel.

mkoeppe:
> Well, build system wise I still need this little bit which is annoying
> https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sagemath-standard/files/sagemath-standard-9.8-build_ext.patch

This is triggered by one of Gentoo's custom PEP 517 circumvention codepaths?

No problem, we can merge it in Sage. PR please.

kiwifb:
> > Well, build system wise I still need this little bit which is annoying
> > https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sagemath-standard/files/sagemath-standard-9.8-build_ext.patch
> 
> This is triggered by one of Gentoo's custom PEP 517 circumvention codepaths?
> 

For the record, and I double checked, the Gentoo version gpep517 is totally vanilla flavor. No changes at all. I could accept that we are calling it wrong (which is why I included a real used command).

mkoeppe:
I mean, gpep517 *is* a custom gentoo tool.

mkoeppe:
> kick jmol out. That would be massively helpful.

@kiwifb That's #31027; but despite the progress of using jupyter-sphinx for Live Doc, both our HTML and PDF doc still contain 3D graphics rendered with Jmol. What should we do about this?

kiwifb:
> > kick jmol out. That would be massively helpful.
> 
> @kiwifb That's #31027; but despite the progress of using jupyter-sphinx for Live Doc, both our HTML and PDF doc still contain 3D graphics rendered with Jmol. What should we do about this?

Not sure. The problems we have are:

1. tachyon does not seem to be a viable alternative for doc
2. I think we cannot use or do not how to use jsmol non interactively

mkoeppe:
The plan in #33513 was to use a screenshotting tool with threejs. What is cited there, pyppeteer, is no longer maintained, but there are other options. But they all have some browser package dependence.

mkoeppe:
> > Well, build system wise I still need this little bit which is annoying
> > https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sagemath-standard/files/sagemath-standard-9.8-build_ext.patch
> 
> This is triggered by one of Gentoo's custom PEP 517 circumvention codepaths?
> 
> No problem, we can merge it in Sage. 

@kiwifb That's now #37671, please review.

kcrisman:
> pypp

kcrisman:
Sorry!!!! Browser delayed reaction ... 🤐 my apologies for spam.

mkoeppe:
Welcome @kcrisman. Higher quality content than some here, would recommend.

kcrisman:
> Welcome @kcrisman. Higher quality content than some here, would recommend.

That was unnecessary.  Some other users have used more volatile language than necessary, and it's reported.  Such sarcasm can be taken just as volatile and/or insulting, so saying "kindly with references to actual distribution policies," has now lost much of its force.

----

I think @culler and @kiwifb have the right idea.  Perhaps a "technical committee" comprised of a few people who are not @dimpase or @mkoeppe can indeed look at whether there is any possible compromise here.  (If we can find volunteers.)  It will likely be unpalatable to the main players, but if it allows most people to get on with Sage development, is that not worth it for them to accept such a solution?  (This would be a sage-devel discussion.)

I find it really hard to believe that modularization (as perceived by @mkoeppe ) and efficient Linux packaging (as perceived by @orlitzky and others) are truly completely incompatible.  Note I say "as perceived by"; obviously the main dispute involves assertions that both are compatible in a way that is unacceptable to one or the other (or that the compatibility is unnecessary), but that's not what I mean.

mkoeppe:
> > Welcome @kcrisman. Higher quality content than some here, would recommend.
> 
> That was unnecessary. Some other users have used more volatile language than necessary, and those occurrences have been reported.

@kcrisman Thanks for reporting it.

>  Such sarcasm can be taken just as volatile and/or insulting

... however, I have to reject this description as harmful bothsidesing.

dimpase:
> Sorry!!!! Browser delayed reaction ... 🤐 my apologies for spam.

you can remove your comments btw :-) 😉

kiwifb:
> > > Welcome @kcrisman. Higher quality content than some here, would recommend.
> > 
> > 
> > That was unnecessary. Some other users have used more volatile language than necessary, and those occurrences have been reported.
> 
> @kcrisman Thanks for reporting it.
> 
> > Such sarcasm can be taken just as volatile and/or insulting
> 
> ... however, I have to reject this description as harmful bothsidesing.

Well, I thought pretty much the same side. You may not perceive it this way or even have intended it this way, but in the current atmosphere it is bound to be interpreted this way by some. If the tenseness I currently perceive was much cooler, a bit of boistering of that kind would be fine, but right now I am more mindful.

dimpase:
sagelib modularisation  is just a "nice to have" feature. Maybe 10 years ago it might've been more important than now,  but hardware is getting more capable. It's not going to attract more users.

making Sage easier to install/build in a variety of scenarios  is much more likely to attract more users and developers.

orlitzky:
> It is unfortunate that the downstream packaging and the current modularization seem to conflict to each other. This seems to me inherent and unavoidable.

There is no conflict. There is a solution that works in all scenarios and is also the standard approach used by all other software packages (because it's what works).

Sage-conf is "needed" for the modularization because all of our configuration is generated in sage-the-distro and winds up in a single file. You can't ship that one file in multiple packages, so it needs to be part of it's own package... if you take for granted that all of our configuration needs to be generated in sage-the-distro and stuffed into one file. **It doesn't.**

Each module can detect the correct configuration at build time, and store it either in a separate file or in the source files themselves (`.py.in` -> `.py`substitution). Then there is no conflict. Packagers are happy, and modularization can proceed.

When sage-conf was introduced, no one was seriously working on a build system for the sage library, and autotools would not have been a great option anyway; really, no good options existed. Now, meson is that good option. Tobias invested the time to build sagelib with it, and Matthias is planning to use it for the modular packages, too. There is no longer any reason to keep shipping a config file in a second package that's generated from within a third package. We can just have meson spit them out automatically.

> Would it be a crazy idea or do not make a sense to fork the sagemath/sage repo to another repo, say sagemath/sage-package?

Apropos of nothing else, it would be great if it were possible to develop and package the sage library independent of the sage distribution. That's what all of these disputed PRs are about. If you're going to fix the design to make it easy to fork, then everyone would be happy anyway and we would no longer need to fork it. But if you're going leave the design as-is and ask the packagers to maintain a fork, then that's just a slightly different _kind_ of pointless extra work -- the thing we object to in the first place.

mkoeppe:
> It is unfortunate that the downstream packaging and the current modularization seem to conflict to each other. This seems to me inherent and unavoidable.

No, there is no such conflict.

The modularization of the Sage library is prepared in a way that is fully aware of packaging considerations, following strictly the standard (PEP 517/518). Downstream packaging of Python packages prepared in this way is straightforward and does not require any particular expertise.

mkoeppe:
> Sage-conf is "needed" for the modularization because [...]

Sage-conf has nothing to do with the modularization.

mkoeppe:
> it would be great if it were possible to develop and package the sage library independent of the sage distribution.

@orlitzky Let's find out what you are missing. So we have to run `./bootstrap`. After that, in `pkgs/` you have standard modern Python packages. What are you looking for?

orlitzky:
> > it would be great if it were possible to develop and package the sage library independent of the sage distribution.
> 
> @orlitzky Let's find out what you are missing. So we have to run `./bootstrap`. After that, in `pkgs/` you have standard modern Python packages. What are you looking for?

I'm not missing anything. I have too much, and need:

1. Less downloading of the entire sage distribution to build one litte piece of the library.
2. Less bootstrapping of sage-the-distro before I can work with (what should be) a standard python package.
3. Less hand-editing of a big config file that changes all the time and that could be generated for me automatically.
4. Less confusion. A lot of people here have been working on sage for 10+ years and I think it's telling that no one actually knows what any of this stuff is or what it's doing.
5. Less documentation. Similar to the last item, but, if you build open source software from github all day long, then you should be able to figure out how to build sage without having to spend a week learning about all the ways in which sage is different.  As a reasonably advanced student... Download Octave? `./configure && make && make install`. Download GAP?  `./configure && make && make install`. Download maxima? `./configure && make && make install`. Download flint? `./configure && make && make install`. Download sage? I guess I'll go fuck myself.

The design _is_ the problem. We can't fix it without changing it.

mkoeppe:
> 1. Less downloading of the entire sage distribution to build one litte piece of the library.
> 2. Less bootstrapping of sage-the-distro before I can work with (what should be) a standard python package.

I think you are confusing the Sage distribution with the monorepository.

orlitzky:
> > 1. Less downloading of the entire sage distribution to build one litte piece of the library.
> > 2. Less bootstrapping of sage-the-distro before I can work with (what should be) a standard python package.
> 
> I think you are confusing the Sage distribution with the monorepository.

It does not help that (a) they are the same thing, and (b) tickets to change that are _also_ in dispute. But my complaints about sage-conf remain largely unchanged under separate repos. I still need to download a separate project and/or hand-edit a config file before I can package/develop sage. What that second package is called is irrelevant. That it should not be needed in the first place is what's important.

dimpase:
As long as `build/pkgs/` remains a flat directory, with even tags in form of empty `math`, etc. files placed there (sic! who approved this design? please step forward...)
to sort things into categories, I doubt that sage-distro is in any way prepared to stand on equal footing with other distros.

Anything short of ability to do `git sparse-checkout ...` to get the minimum needed to build and install `sagelib` isn't really acceptable, in case of monorepo remaining in place.

mkoeppe:
> As long as `build/pkgs/` remains a flat directory [...]

Dima, many distributions look like this; I'm sure you actually know that. Example: https://github.com/void-linux/void-packages/tree/master/srcpkgs

And if you are interested in a change of this directory format, for example
- #37566

... then you can help by reviewing the PRs that do the necessary refactoring, in particular
- https://github.com/sagemath/sage/pull/37430

kwankyu:
OK. Sorry for an idea that nobody likes. 

> > Here's one idea: we could create a separate package (sdist) that corresponds to a "standalone" version of sagemath (= sagelib). We can start with what currently is `sagemath-standard`+`sage-setup` and work on that (fix our issues, maybe experiment with a meson build, etc). We can give this a different name so `sagemath-standard` is no longer a source of contention.
> @tornaria This does not work because of the constraints of contemporary Python packaging.
> 
> * User packages need to be able to declare their dependency on a distribution that provides Sage.
> * Build-time requirements and runtime dependencies are keyed 1:1 to distribution names, nothing else. There's no such thing as declaring "Provides" in a distribution, or virtual packages, or anything like that. And there's no such thing as an "OR" in build-time requirements and runtime dependency specifications.

I have no idea about what these two bullet points mean. But Gonzalo's proposal seems to mean simply to create a separate package (say sagemath-standard-base) and keep everything else outside of the package untouched. Right? Then how could this be "does not work"?

Gonzalo, do you understand the bullet points and agree? I am curious.

mkoeppe:
@kwankyu Let me start to explain. Questions and comments are welcome.

 I wrote: "User packages need to be able to declare their dependency on a distribution that provides Sage."

I'm talking about pip-installable packages that depend on Sage (or parts of Sage). Some examples: 
- https://github.com/flatsurf/sage-flatsurf
- https://github.com/mkauers/ore_algebra
- https://github.com/mkoeppe/sage-numerical-interactive-mip

Like any other pip-installable package (distribution package), such user packages are able to (and expected to!) declare their dependencies. That's part of the contract of being part of the Python "ecosystem".

The dependencies come in two main flavors: Build-time dependencies ("build-system requirements") and runtime dependencies ("install-requires").
For example: https://github.com/flatsurf/sage-flatsurf/blob/master/setup.py#L41 declares a dependency on the package **surface-dynamics**; this is a runtime dependency.

Before I made the **sagemath-standard** distribution available, there was actually no consistent way to even declare a dependency on the Sage library. Because of this, still today many user packages do not actually declare this dependency, and so they are not yet first-class members of the Python "ecosystem".

dimpase:
> > As long as `build/pkgs/` remains a flat directory [...]
> 
> Dima, many distributions look like this; I'm sure you actually know that. Example: https://github.com/void-linux/void-packages/tree/master/srcpkgs
> 
> And if you are interested in a change of this directory format, for example
> - #37566
> 
> ... then you can help by reviewing the PRs that do the necessary refactoring, in particular
> - https://github.com/sagemath/sage/pull/37430
> 

I cannot review PRs by someone who's blocked me on GitHub, as I can't leave comments there

dimpase:
> > As long as `build/pkgs/` remains a flat directory [...]
> 
> Dima, many distributions look like this; I'm sure you actually know that. Example: https://github.com/void-linux/void-packages/tree/master/srcpkgs

No, I actually don't - but  I suppose they have indices to group packages according to their purposes.
sage-distro does not have such an index 


Besides, sagelib has a very special role 
in sage-distro - this is not the case for other distros, where all packages are "equal".

mkoeppe:
> > > As long as `build/pkgs/` remains a flat directory [...]
> > 
> > 
> > Dima, many distributions look like this; I'm sure you actually know that. Example: https://github.com/void-linux/void-packages/tree/master/srcpkgs
> 
> No, I actually don't 

Well, then you know what to do.

kwankyu:
@mkoeppe Thanks. I understand that. How is that related with Gonzalo's suggestion? User packages may still declare sagemath-standard, but downstream packagers may use sagemath-standard-base. No? I fear that I misunderstand the situation...

culler:
`Like any other pip-installable package (distribution package), such user packages are able to (and expected to!) declare their dependencies.`

On thing that makes this topic so hard to discuss is that people use so much jargon while assuming (incorrectly) that others know what they mean.  I have never clearly understood what is meant by phrases like "sage-the-distribution", for example.  I finally decided it must mean "things that get installed in sage/local"  but I really don't know.

Now here we have a term "distribution package" being defined as a pip-installable package.  And this is in the middle of a discussion about linux distros, where "distro" is short for "distribution". And I would have assumed that any pip-installable package would be part of what people call "sage-the-library" not "sage-the-distribution".  And the new term is defined in terms of another term "user package" which has no meaning to me.

It would really help to have clearly defined terms when discussing these things.  Preferably the definitions would avoid vague,  ambiguous or overused terms like "distribution" and "vendored" and "flat directory". (What is a "curved directory"?)   It is clear what is meant by a "package on PyPI" or a "tarball on a sage mirror" or a "dynamic library" or a "python package" or a "linux package".  If the terminology were defined in terms of things that have clear meanings we might do better.

mkoeppe:
> One thing that makes this topic so hard to discuss is that people use so much jargon while assuming (incorrectly) that others know what they mean.

Yes, and unfortunately many of the terms of the jargon are heavily overloaded.

> we have a term "distribution package" being defined as a pip-installable package.

Yes. "Distribution package" is the official word for this, according to PyPA https://packaging.python.org/en/latest/glossary/#term-Distribution-Package

tornaria:
> > > Here's one idea: we could create a separate package (sdist) that corresponds to a 
> I have no idea about what these two bullet points mean. But Gonzalo's proposal seems to mean simply to create a separate package (say sagemath-standard-base) and keep everything else outside of the package untouched. Right? Then how could this be "does not work"?

@kwankyu: Precisely, the proposal is to create a separate package *with a different name* that will be a standalone package. Major goals from the onset:
 - all external dependencies would be provided by the environment and it's out of the scope of this package how those are installed,
 - other than that, everything else would be provided in a single self-contained package.

IOW, assuming all the required dependencies are installed in standard locations, it would only take a standard build/install step to have sagemath working on the system (identical to e.g. numpy, scipy, sympy, ipython, etc)

One way it could be implemented is to create it as a new subdirectory inside `pkgs` in the monorepo. This subdirectoory would only contain the build scripts and configuration, but no sagelib code; the latter would be taken from `src/sage` (just like most packages inside `pkgs` do.)


> Gonzalo, do you understand the bullet points and agree? I am curious.

The proposal is not trying to solve "user packages", so the bullet points you refer to are not relevant. Note that we would use a different name to avoid any type of conflict with that. Supporting user packages is a worthwhile goal; having this separate standalone package is also a way to remove some pressure from the current packages so they can move faster towards that goal.


> Now here we have a term "distribution package" being defined as a pip-installable package. And this is in the middle of a discussion about linux distros, where "distro" is short for "distribution".

@culler: I use the shorthand "distro" to mean "software distribution" (as in "a distribution of software packages"). I try to stress the shorthand in an attempt to avoid the confusion with the python term.

In fact, I often write "linux distro" only to make it more clear, but the general concept is not limited to "linux". A clear example would be freebsd, but possibly conda, homebrew, etc. could also match. When we say "sage-the-distro" we refer to the part of sagemath that also matches this (see `build/pkgs` for the list of 400+ pkgs included in sage-the-distro).

The general idea of a distro is that there is some sort of package manager that allows one to install software packages (usually from a binary but sometimes from source, e.g. gentoo and sage-the-distro). The role of the distro is to integrate the different packages together and with the system, and usually they either control the whole stack (from kernel up) or they run in a specific environment (e.g. homebrew only for mac).

Sage-the-distro is special in this regard since it aims to work in many different environments which makes the task more difficult. Other distributions have it easier; for instance, when I make a sagemath package for void linux I only have to test with the versions of the dependencies that we ship on void linux (it helps a lot that I maintain the void linux packages for most of the dependencies of sagemath)


> And I would have assumed that any pip-installable package would be part of what people call "sage-the-library" not "sage-the-distribution". And the new term is defined in terms of another term "user package" which has no meaning to me.

One of the packages available in some "distros" is a python package which in sage-the-distro is called `sagelib` (see `build/pkgs/sagelib`). This is what some may call `sage-the-library` although is not just a "library" since it includes the script frontend `sage`.


> It would really help to have clearly defined terms when discussing these things. Preferably the definitions would avoid vague, ambiguous or overused terms like "distribution" and "vendored" and "flat directory". (What is a "curved directory"?) It is clear what is meant by a "package on PyPI" or a "tarball on a sage mirror" or a "dynamic library" or a "python package" or a "linux package". If the terminology were defined in terms of things that have clear meanings we might do better.

As for the python term "distribution", it's a shortcut for "source distribution package" aka "sdist" which means an individual "python package".  Our monorepo contains several distribution packages in subdirectories under "pkgs". Actually, the subdirectory is not really the "sdist", but it contains the instructions to create the sdist (see the file `MANIFEST.in` for each package). The "sdist" is built and uploaded to pypi as a .tar.gz file that you see for instance in https://pypi.org/project/sagemath-standard/#files.


Flat directory menans "build/pkgs" has 400+ subdirectories, one for each package. As opposed to a tree-like structure. This is not so important IMO, although handling a directory that is too large leads to some inconveniences (the void packages repo has a directory with 13000+ entries and this is really painful sometimes).

kwankyu:
> @kwankyu: Precisely, the proposal is to create a separate package _with a different name_ that will be a standalone package. Major goals from the onset:
> 
> * all external dependencies would be provided by the environment and it's out of the scope of this package how those are installed,
> * other than that, everything else would be provided in a single self-contained package.
> 
> IOW, assuming all the required dependencies are installed in standard locations, it would only take a standard build/install step to have sagemath working on the system (identical to e.g. numpy, scipy, sympy, ipython, etc)
> 
> One way it could be implemented is to create it as a new subdirectory inside `pkgs` in the monorepo. This subdirectoory would only contain the build scripts and configuration, but no sagelib code; the latter would be taken from `src/sage` (just like most packages inside `pkgs` do.)
> 
> > Gonzalo, do you understand the bullet points and agree? I am curious.
> 
> The proposal is not trying to solve "user packages", so the bullet points you refer to are not relevant. Note that we would use a different name to avoid any type of conflict with that. Supporting user packages is a worthwhile goal; having this separate standalone package is also a way to remove some pressure from the current packages so they can move faster towards that goal.

Thanks. So this new distribution package with whatever name `xxx` has zero chance to conflict with the current modularization effort. All would happen inside the directory `pkgs/xxx`. This distribution package would exist solely for the software sage deployment to linux platforms, or to environments like conda or possibly homebrew.

Matthias may regard it as a duplicate since in his opinion, `sagemath-standard` serves for the same purpose. But if some downstream packagers would prefer `xxx` to `sagemath-standard` for whatever reasons, we may keep both distribution packages side by side for users to choose. Matthias?

kwankyu:
> @culler: I use the shorthand "distro" to mean "software distribution" (as in "a distribution of software packages"). I try to stress the shorthand in an attempt to avoid the confusion with the python term.

"distribution" and "package" are too much overloaded. We should qualify the terms when we use them. 

Please do not use the terms for general concepts.

> Sage-the-distro is special in this regard since it aims to work in many different environments which makes the task more difficult. Other distributions have it easier. ...

Please do not use the term "sage-the-distribution" as something equal to a linux distribution.

> One of the packages available in some "distros" is a python package which in sage-the-distro is called `sagelib` (see `build/pkgs/sagelib`). ...

That is a confusing statement. Yes, there is `build/pkgs/sagelib` for the sage installation procedure. But we should not think sage-the-library as part of sage-the-distribution. The main purpose of sage-the-distribution is to provide external dependencies of sage-the-library.

> As for the python term "distribution", it's a shortcut for "source distribution package" aka "sdist" which means an individual "python package". 

For that meaning, let's use only the terms "source distribution" ("sdist" ) and "binary distribution" ("bdist").
 
> Flat directory menans "build/pkgs" has 400+ subdirectories, one for each package. As opposed to a tree-like structure. This is not so important IMO, although handling a directory that is too large leads to some inconveniences (the void packages repo has a directory with 13000+ entries and this is really painful sometimes).

I don't think a tree structure would be useful for "build/pkgs".

dimpase:
> Flat directory menans "build/pkgs" has 400+ subdirectories, one for each package. As opposed to a tree-like structure. This is not so important IMO, although handling a directory that is too large leads to some inconveniences (the void packages repo has a directory with 13000+ entries and this is really painful sometimes).

well, mentioning how Void does things as an excuse is a typical whataboutism. 
Void is a bit of a joke in this sense - all it seems to offer, if one wants a particular package, is keywords search facility. :-(
Our story is rather different, and Void or any other Linux distro are not an excuse for the status quo.

Unlike Void (or other Linux distros), majority of spkgs can't be installed at will
(only optional and experimental can). Also, there are only very few spkg which must be installed, and not substituted by an external analog, already installed on the system - it's sagelib, sage-conf,  sagelib doctesting tools, and  few other, bootstrap-related, things.
In this sense, the rest of standard spkgs is optional.
Let's have a closer look at these standard packages. The rough classification is as follows, and it IMHO should make its way into the docs.

### standard package classification (sans sagelib+sage-conf+bootstrap)
1. python (an - outdated - Python source) and its toolchain - for building wheels etc.
2. compilers (C, C++, Fortran, Lisp, Cython)
3. sagelib deps - what one needs to build and use sagelib - other than 1. and 2.
     * build+runtime
     * runtime-only
4.  "Schroedinger" sagelib deps, such as rpy2 (they are optional by right, but not quite, and so marked at standard)
5. sage preparser+terminal IO - to support sage's  "language" (`^` instead of `**`, etc) + ipython
6. notebook+jupyter(lab) - that's (G)UI. One can build and use sagelib without it
7. sage's Jupyter kernel
8. sphinx & Co. - tools to build docs (they do use a small part of 3.)
9. ... something else?

tornaria:
+1: @tobiasdiez @orlitzky @dimpase @tornaria 
-1: @mkoeppe 

Am I missing anyone?

src/pyproject.toml.m4
```diff
@@ -6,7 +6,6 @@ requires = [
     # Note that PEP517/518 have no notion of optional sage_spkg dependencies:
     # https://github.com/pypa/pip/issues/6144
      esyscmd(`sage-get-system-packages install-requires-toml \
-        sage_conf      \
         setuptools \
```
kwankyu, line 6: Line 5 says "Some version of sage-conf is required". But `sage_conf` is removed from line 9. 

Then would `pip install sagemath-standard`  work?

tobiasdiez: yes, it's still working as can be seen from the conda ci workflow.

Thanks for catching the outdated comment. I've removed it now.

tobiasdiez:
Setting to positive review based on the counting in https://github.com/sagemath/sage/pull/37138#issuecomment-2042619025

vbraun:
merge conflict

tobiasdiez:
> merge conflict

Cleanly rebased on top of https://github.com/sagemath/sage/pull/37796.

vbraun:
merge conflict

PR #37141, opened 2024-01-22
Format graphic_matroid.py

gmou3:
Improve the formatting of graphic_matroid.py.


src/sage/matroids/graphic_matroid.py
```diff
@@ -1357,26 +1397,29 @@ def graphic_extension(self, u, v=None, element=None):
...
 
         EXAMPLES::
 
             sage: M = matroids.CompleteGraphic(4)
             sage: M1 = M.graphic_extension(0,1,'a'); M1
             Graphic matroid of rank 3 on 7 elements
             sage: list(M1.graph().edge_iterator())
-            [(0, 1, 'a'), (0, 1, 0), (0, 2, 1), (0, 3, 2), (1, 2, 3), (1, 3, 4), (2, 3, 5)]
+            [(0, 1, 'a'), (0, 1, 0), (0, 2, 1), (0, 3, 2), (1, 2, 3),
+            (1, 3, 4), (2, 3, 5)]
```
tscrim:
```suggestion
            [(0, 1, 'a'), (0, 1, 0), (0, 2, 1), (0, 3, 2), (1, 2, 3),
             (1, 3, 4), (2, 3, 5)]
```

src/sage/matroids/graphic_matroid.py
```diff
@@ -1357,26 +1397,29 @@ def graphic_extension(self, u, v=None, element=None):
...
 
         ::
 
             sage: M = Matroid(range(10), graphs.PetersenGraph())
-            sage: sorted(M.graphic_extension(0, 'b', 'c').graph().vertex_iterator(), key=str)
+            sage: EX_G = M.graphic_extension(0, 'b', 'c').graph()
+            sage: sorted(EX_G.vertex_iterator(), key=str)
             [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 'b']
-            sage: M.graphic_extension('a', 'b', 'c').graph().vertices(sort=False)
+            sage: EX_G = M.graphic_extension('a', 'b', 'c').graph()
```
tscrim:
```suggestion
            sage: M.graphic_extension('a', 'b', 'c').graph()
```
No need to suggest it is being stored.

src/sage/matroids/graphic_matroid.py
```diff
@@ -1688,19 +1746,21 @@ def graphic_coextensions(self, vertices=None, v=None, element=None, cosimple=Fal
 
             sage: M = Matroid(graphs.WheelGraph(9))
             sage: I = M.graphic_coextensions()
-            sage: sum(1 for N in I)
+            sage: len(list(I))
             136
             sage: I = M.graphic_coextensions(cosimple=True)
-            sage: sum(1 for N in I)
+            sage: len(list(I))
```
tscrim: These actually put everything in memory rather than iterating over it. So it is actually a slightly different doctest. I think the previous behavior is a bit better to suggest to the user.

tscrim (commented):
Thank you for working to clean up the file. However, most of your changes actually make the file harder to understand the code and output. The 80 char/line can often be bad to follow rigidly, which can be seen by line splits with small output for just a few characters over. While we generally follow this more strictly for docstrings, there are good reasons to violate this at times for readability, especially when the html output doesn't require a horizontal scrollbar. So unfortunately I think most of these changes should be reverted.

Also, IMO, the strong avoidance of `l` because it could be confused with `I` thing only is an issue if both appear or there is ambiguity with what symbol it should mean. As such, calling labels `ll` instead of `l` doesn't actually fix the problem (why isn't it `II` instead?) but just makes the names longer and more confusing.

gmou3:
I agree actually, I mostly wanted to remove the warnings that appeared on my editor. I think it is better now.

tscrim:
> I agree actually, I mostly wanted to remove the warnings that appeared on my editor. I think it is better now.

Please change your settings so those warning are disabled or find a new editor please (and launch the current one into the sun :) ).  Most of these changes really are making the code harder to read. PEP8 allows this flexibility (in particular, one of the main points is make the code consistent).

src/sage/matroids/graphic_matroid.py
```diff
@@ -544,8 +533,9 @@ def _minor(self, contractions=frozenset([]), deletions=frozenset([])):
             sage: M._minor(contractions=frozenset([0,1,2]))
             Graphic matroid of rank 1 on 7 elements
             sage: M = Matroid(range(15), graphs.PetersenGraph())
-            sage: N = M._minor(deletions = frozenset([0, 3, 5, 9]), contractions =
-            ....: frozenset([1, 2, 11])); N
+            sage: N = M._minor(deletions = frozenset([0, 3, 5, 9]),
+            ....:              contractions = frozenset([1, 2, 11]))
```
tscrim:
```suggestion
            sage: N = M._minor(deletions=frozenset([0, 3, 5, 9]),
            ....:              contractions=frozenset([1, 2, 11]))
```
PEP8

tscrim (approved):
Thank you. One last little PEP8 thing since you are modifying the line.

tscrim:
Thank you.

PR #37142, opened 2024-01-22
throw ValueError instead of TypeError when logarithm doesn't exist in AdditiveAbelianGroupWrapper

yyyyx4:
Attempting to run the higher-level algorithms in the `additive_abelian_wrapper.py` file on a group with unhashable elements will also throw a `TypeError`, which leads to very strange behavior (rather than failing with a clear error message, as it should).

#sd123

### DIFF
--- a/src/sage/groups/additive_abelian/additive_abelian_wrapper.py
+++ b/src/sage/groups/additive_abelian/additive_abelian_wrapper.py
@@ -365,7 +365,7 @@ class AdditiveAbelianGroupWrapper(addgp.AdditiveAbelianGroup_fixed_gens):
             sage: G.discrete_log(V([6, 4]))
             Traceback (most recent call last):
             ...
-            TypeError: Not in group
+            ValueError: not in group
 
         ::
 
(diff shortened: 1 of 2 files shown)


grhkm21:
LGTM

roed314:
@grhkm21, I've invited you to the Sage organization (and the Triage team).  Once you accept, you can review PRs.

grhkm21:
Doctests are failing

grhkm21 (changes requested):
Doctests are failing. For example, line 110 (P1 in G) throws an error now even though it should return False.

Maybe [this line](https://github.com/yyyyx4/sage/blob/public/throw_ValueError_when_logarithm_does_not_exist/src/sage/modules/fg_pid/fgp_module.py#L705) should be modified to catch `ValueError` too?

yyyyx4:
Thanks, and sorry, I ripped this out of a branch with a larger set of changes and apparently forgot to run the tests again prior to making the pull request...

grhkm21:
LGTM :D

PR #37145, opened 2024-01-22
Add matroid attribute girth()

gmou3:
Added a function `girth()` to the main matroid class. This is a standard notion which corresponds to the size of the smallest circuit of the matroid. A similar notion exists for graphs.

### :memo: Checklist

### DIFF
--- a/src/sage/matroids/matroid.pyx
+++ b/src/sage/matroids/matroid.pyx
@@ -5986,8 +5986,8 @@ cdef class Matroid(SageObject):
 
         EXAMPLES::
 
-            sage: matroids.Uniform(5,5).girth()
-            inf
+            sage: matroids.Uniform(5, 5).girth()
+            +Infinity
             sage: matroids.catalog.K4().girth()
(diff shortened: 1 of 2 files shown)


src/sage/matroids/matroid.pyx
```diff
@@ -5976,6 +5977,33 @@ cdef class Matroid(SageObject):
...
+        REFERENCES:
+
+        [Oxl2011]_, p. 327.
+        """
+        for k in range(self.rank() + 2):
+            for X in combinations(self.groundset(), k):
+                X = frozenset(X)
+                if self._is_circuit(X):
+                    return k
+        return float('inf')
```
tscrim: Rather than use `float(“inf”)`, it would be better to use the Sage `infinity`.

tscrim (changes requested):
One minor thing; otherwise LGTM.

tscrim (approved):
Thank you. LGTM.

PR #37148, opened 2024-01-23
Add class CircuitsMatroid

gmou3:
Added a subclass `CircuitsMatroid` (of the main `Matroid` class). This enables the storage of a matroid based on its list of circuits, the verification using the circuit axioms, etc. The circuits description can be very succinct and it can facilitate efficient implementations of algorithms (one such example is the aforementioned verification).

Up till now, when a matroid was defined via the keyword `circuits` - `Matroid(circuits=[...])` - it was converted to a `BasisMatroid`. This required the computation of the bases upon definition which could be very slow.

For some relevant discussion, have a look at #36820 and #36956. This PR is my proposed resolution of #36820 (Note that this provides a workaround and not a direct implementation of the issue's request. In fact, I deem that a direct implementation is undesirable due to memory considerations; see #36956.)

### :memo: Checklist


gmou3:
@tscrim

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+
+from sage.structure.richcmp cimport rich_to_bool, richcmp
+from sage.matroids.matroid cimport Matroid
+from sage.matroids.set_system cimport SetSystem
+from cpython.object cimport Py_EQ, Py_NE
+
+
+cdef class CircuitsMatroid(Matroid):
+    r"""
+    INPUT:
```
tscrim:
```suggestion
cdef class CircuitsMatroid(Matroid):
    r"""
    A matroid defined by its set of circuits.

    INPUT:
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+    - ``groundset`` -- a list (default: ``None``); the groundset of the matroid
+    - ``circuits`` -- a list (default: ``None``); the collection of circuits of
+      the matroid
+
+    OUTPUT:
+
+    - If the input is a matroid ``M``, return a ``CircuitsMatroid`` instance
+      representing ``M``.
+    - Otherwise, return a ``CircuitsMatroid`` instance based on the
+      ``groundset`` and ``circuits``.
```
tscrim:
```suggestion
```
There is no point as the output is clearly this class. This isn't uniform and very, very in Sage (and likely should be removed in other places).

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+
+        For a more flexible means of input, use the ``Matroid()`` function.
+    """
+
+    # necessary (__init__, groundset, _rank)
+
+    def __init__(self, M=None, groundset=None, circuits=None, nsc_defined=False):
+        """
+        Initialization of the matroid. See class docstring for full
+        documentation.
```
tscrim, line 62: Needs a doctest; typically a `TestSuite(M).run()`.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+        Compute a maximal independent subset.
+
+        INPUT:
+
+        - ``X`` -- An object with Python's ``frozenset`` interface containing
+          a subset of ``self.groundset()``
+
+        OUTPUT: a frozenset; a maximal independent subset of ``X``
+        """
+        I = set(F)
```
tscrim:
```suggestion
        cdef set I = set(F)
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+        for i in self._k_C:
+            for C in self._k_C[i]:
+                if i <= len(I) and i > 0:
+                    if C <= I:
+                        for e in C:
```
tscrim:
```suggestion
                        if e in C:
```
otherwise this is not doing anything.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+
+        INPUT:
+
+        - ``X`` -- An object with Python's ``frozenset`` interface containing
+          a subset of ``self.groundset()``
+
+        OUTPUT: boolean
+        """
+        I = set(F)
+        s = len(F)
```
tscrim:
```suggestion
        cdef set I = set(F)
        cdef int s = len(F)
```

src/sage/matroids/constructor.py
```diff
@@ -302,33 +303,33 @@ def Matroid(groundset=None, data=None, **kwds):
...
+        :class:`CircuitsMatroid <sage.matroids.basis_matroid.CircuitsMatroid>`
         internally::
 
+            sage: from sage.matroids.circuits_matroid import CircuitsMatroid
+            sage: M2 = CircuitsMatroid(Matroid(bases=['ab', 'ac']))
+            sage: M1 == M2
+            True
+
+
+
```
tscrim:
```suggestion
            True

```

src/sage/matroids/unpickling.pyx
```diff
@@ -28,13 +28,15 @@ AUTHORS:
 from sage.data_structures.bitset_base cimport *
 import sage.matroids.matroid
 import sage.matroids.basis_exchange_matroid
-from sage.matroids.minor_matroid import MinorMatroid
-from sage.matroids.dual_matroid import DualMatroid
-from sage.matroids.circuit_closures_matroid cimport CircuitClosuresMatroid
 from sage.matroids.basis_matroid cimport BasisMatroid
-from sage.matroids.linear_matroid cimport LinearMatroid, RegularMatroid, BinaryMatroid, TernaryMatroid, QuaternaryMatroid
+from sage.matroids.circuit_closures_matroid cimport CircuitClosuresMatroid
```
tscrim: Let's not needlessly change the order of things, especially to something just as random.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+            sage: M.is_paving()
+            True
+        """
+        return self.girth() >= self.rank()
+
+    # verification
+
+    cpdef is_valid(self) noexcept:
+        r"""
+        Test if self obeys the matroid axioms.
```
tscrim:
```suggestion
        Test if ``self`` obeys the matroid axioms.
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+        EXAMPLES::
+
+            sage: matroids.Theta(10).girth()
+            3
+
+        REFERENCES:
+
+        [Oxl2011]_, p. 327.
+        """
+        return min([i for i in self._k_C], default=float('inf'))
```
tscrim:
```suggestion
        return min(self._k_C, default=float('inf'))
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+        """
+        cdef SetSystem NSC
+        NSC = SetSystem(list(self.groundset()))
+        for i in self._k_C:
+            if i <= self.rank():
+                for C in self._k_C[i]:
+                    NSC.append(C)
+        return NSC
+
+    def no_broken_circuits_sets(self, order=None):
```
tscrim: This is an iterator, but the other methods (at least by a spot check) return lists. This should be a list for consistency (perhaps good to change to an iterator at some point though).

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,613 @@
...
+        INPUT:
+
+        - ``X`` -- An object with Python's ``frozenset`` interface containing
+          a subset of ``self.groundset()``.
+
+        OUTPUT: a frozenset; a circuit contained in ``X``, if it exists.
+        Otherwise an error is raised.
+        """
+        I = set(F)
+        for C in self.circuits():
```
tscrim:
```suggestion
        cdef set I = set(F)
        for C in self._C:
```
(or perhaps `self.circuits_iterator()`?)

tscrim (changes requested):
In addition to these comments, *all* `def` and `cpdef` methods need doctests.

src/sage/matroids/circuits_matroid.pyx
gmou3: That was at the top of the file. Moved it inside __init__.

src/sage/matroids/circuits_matroid.pyx
gmou3: This is a fast way of picking an arbitrary element from C. (I know it looks weird.)

src/sage/matroids/circuits_matroid.pyx
gmou3: Thanks, this needed some essential changes.

src/sage/matroids/circuits_matroid.pyx
gmou3: For a `CircuitsMatroid` there is no need to use the `circuits_iterator` internally as the circuits are already in memory.

gmou3:
Thanks for the detailed feedback. I was planning to further enrich the examples after the merging of this, and after editing some database matroids to be circuits matroids (the plan is to make the doctests much faster via a few strategic recasting of matroids into `CircuitMatroid`s).

src/sage/matroids/circuits_matroid.pyx
tscrim: You should not rely on scoped variables. If you just want to get an element in `C`, then do
```python
e = next(iter(C))
```

src/sage/matroids/database_matroids.py
```diff
@@ -2116,7 +2116,8 @@ def Spike(r, t=True, C3=[]):
     EXAMPLES::
 
         sage: M = matroids.Spike(3, False); M
-        Free 3-spike\t: Matroid of rank 3 on 6 elements with 20 bases
+        Free 3-spike\t: M \ {'t'}, where M is Matroid of rank 3 on 7 elements
+        with 3 non-spanning circuits
```
tscrim:
```suggestion
         with 3 non-spanning circuits
```
Just as a subtle nod to the fact it is a continuing doctest (this sort of thing is common in Sage's doctests; IMO, it should be more universal).

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+            sage: C = [[1, 2, 3], [3, 4, 5]]
+            sage: M = Matroid(circuits=C)
+            sage: M.is_valid()
+            False
+        """
+        for i in self._k_C:
+            for j in self._k_C:
+                if i <= j:
+                    for C1 in self._k_C[i]:
+                        if len(C1) == 0:
```
tscrim:
```suggestion
                        if not C1:
```
A microoptimization.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+                            return False
+                        for C2 in self._k_C[j]:
+                            if C1 < C2:
+                                return False
+                            if C1 == C2:
+                                break
+                            for e in C1 & C2:
+                                flag = False
+                                S = (set(C1) | set(C2)) - {e}
```
tscrim:
```suggestion
                            UC12 = set(C1) | set(C2))
                            for e in C1 & C2:
                                flag = False
                                S = UC12- {e}
```
So you don't recreate the union every loop.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+            sage: M.is_valid()
+            False
+            sage: C = [[1, 2, 3], [3, 4, 5]]
+            sage: M = Matroid(circuits=C)
+            sage: M.is_valid()
+            False
+        """
+        for i in self._k_C:
+            for j in self._k_C:
+                if i <= j:
```
tscrim:
```suggestion
        sizes = sorted(self._k_C)
        for ii, i in enumerate(sizes):
            for jj,j in enumerate(sizes[:ii+1]):
```
Fewer checks and simplifies the logic IMO. Actually, you probably can check if the first size is zero and then skip it to further simplify the next lines. The `jj` will be used below.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+                    for C1 in self._k_C[i]:
+                        if len(C1) == 0:
+                            return False
+                        for C2 in self._k_C[j]:
+                            if C1 < C2:
+                                return False
+                            if C1 == C2:
+                                break
```
tscrim:
```suggestion
```
This should never happen because `C1` should be smaller than `C2`, right?

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+                            for e in C1 & C2:
+                                flag = False
+                                S = (set(C1) | set(C2)) - {e}
+                                for k in self._k_C:
```
tscrim:
```suggestion
                                for k in sizes[jj+1:]:
```
It doesn't remove the check below, but it does mean you potentially get to skip a lot of checks.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+            sage: M = CircuitsMatroid(matroids.Uniform(2, 4))
+            sage: M.nonspanning_circuits()
+            Iterator over a system of subsets
+        """
+        cdef SetSystem NSC
+        NSC = SetSystem(list(self.groundset()))
+        for i in self._k_C:
+            if i <= self.rank():
+                for C in self._k_C[i]:
+                    NSC.append(C)
```
tscrim:
```suggestion
                NSC.extend(self._k_C[i])
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+
+        EXAMPLES::
+
+            sage: from sage.matroids.circuits_matroid import CircuitsMatroid
+            sage: M = CircuitsMatroid(matroids.Uniform(2, 4))
+            sage: M.nonspanning_circuits()
+            Iterator over a system of subsets
+        """
+        cdef SetSystem NSC
+        NSC = SetSystem(list(self.groundset()))
```
tscrim:
```suggestion
        cdef list NSC = list(self.groundset())
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        for i in self._k_C:
+            if i <= self.rank():
+                for C in self._k_C[i]:
+                    NSC.append(C)
+        return NSC
```
tscrim:
```suggestion
        return SetSystem(NSC)
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        - ``k`` -- an integer (optional); the length of the circuits
+
+        EXAMPLES::
+
+            sage: from sage.matroids.circuits_matroid import CircuitsMatroid
+            sage: M = CircuitsMatroid(matroids.Uniform(2, 4))
+            sage: sum(1 for C in M.circuits_iterator())
+            4
+        """
+        if k:
```
tscrim:
```suggestion
        if k is not None:
```
`k` could be `0`, which would then not fall into this path. Probably good to add a test case for this.

I would actually explicitly list the circuits in the doctest too.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        EXAMPLES::
+
+            sage: from sage.matroids.circuits_matroid import CircuitsMatroid
+            sage: M = CircuitsMatroid(matroids.Uniform(2, 4))
+            sage: M.circuits()
+            Iterator over a system of subsets
+        """
+        cdef SetSystem C
+        C = SetSystem(list(self.groundset()))
+        if k:
```
tscrim:
```suggestion
        if k is not None:
```
Same as for the iterator.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        """
+        from itertools import combinations
+        for B in combinations(self._groundset, self._matroid_rank):
+            flag = True
+            for C in self.nonspanning_circuits():
+                if C <= set(B):
+                    flag = False
+                    break
+            if flag:
+                yield frozenset(B)
```
tscrim:
```suggestion
        cdef SetSystem NSC = self.nonspanning_circuits()
        for B in combinations(self._groundset, self._matroid_rank):
            B = frozenset(B)
            if not any(C <= B for C in NSC)
                yield B
```
For normal `def` functions you can use generators IIRC. Anyways, it is best to not recreate things many times over.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        from itertools import combinations
+        for S in combinations(self._groundset, self._matroid_rank):
+            flag = True
+            for C in self.nonspanning_circuits():
+                if C <= set(S):
+                    flag = False
+                    break
+            if flag:
+                B.append(S)
+        return B
```
tscrim:
```suggestion
        cdef SetSystem B, NSC
        cdef bint flag
        B = SetSystem(list(self.groundset()))
        NSC = self.nonspanning_circuits()
        from itertools import combinations
        for S in combinations(self._groundset, self._matroid_rank):
            flag = True
            S = frozenset(S)
            for C in NSC:
                if C <= S:
                    flag = False
                    break
            if flag:
                B.append(S)
        return B
```

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+
+        EXAMPLES::
+
+            sage: M = matroids.Theta(20)
+            sage: M.full_rank()
+            20
+        """
+        return self._matroid_rank
+
+    cpdef _is_independent(self, F) noexcept:
```
tscrim, line 143: This method needs a doctest.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        cdef set I = set(F)
+        cdef int s = len(F)
+        for i in self._k_C:
+            if i <= s:
+                for C in self._k_C[i]:
+                    if C <= I:
+                        return False
+        return True
+
+    cpdef _max_independent(self, F) noexcept:
```
tscrim, line 171: This method needs a doctest.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+            for C in self._k_C[i]:
+                if i <= len(I) and i > 0:
+                    if C <= I:
+                        for e in C:
+                            break
+                        I.remove(e)
+
+        return frozenset(I)
+
+    cpdef _circuit(self, F) noexcept:
```
tscrim, line 198: This method needs a doctest.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+        OUTPUT: a frozenset; a circuit contained in ``X``, if it exists.
+        Otherwise an error is raised.
+        """
+        cdef set I = set(F)
+        for C in self._C:
+            if C <= I:
+                return C
+        raise ValueError("no circuit in independent set")
+
+    cpdef _is_isomorphic(self, other, certificate=False) noexcept:
```
tscrim, line 226: This method needs a doctest.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,676 @@
...
+            Internal version that does no input checking.
+        """
+        if certificate:
+            return self._is_isomorphic(other), self._isomorphism(other)
+        N = CircuitsMatroid(other)
+        return self._C._isomorphism(N._C) is not None
+
+    # representation
+
+    def _repr_(self):
```
tscrim, line 260: This method needs a doctest.

tscrim (changes requested):
Thank you. It is looking good.

Here are some more comments. Some are necessary (again, all `def` and `cpdef` methods need doctests within their docstring), but others are some optimizations I think are fairly easy.

src/sage/matroids/database_matroids.py
gmou3: I went ahead and did many docstring changes. I changed some example blocks having in mind a 95th column limit, and I made the REFERENCE blocks into one-liners where applicable.

src/sage/matroids/circuits_matroid.pyx
gmou3: Seems more complicated to me. These checks are nothing compared to the algorithm's complexity.

src/sage/matroids/circuits_matroid.pyx
gmou3: The size of `C1` is smaller or equal to that of `C2`.

src/sage/matroids/circuits_matroid.pyx
gmou3: Done. Good catch!

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,737 @@
...
+        """
+        Test if ``self`` is isomorphic to ``other``.
+
+        INPUT:
+
+        - ``other`` -- a matroid
+        - ``certificate`` -- boolean (optional)
+
+        OUTPUT: boolean, and, if certificate = True, a dictionary giving the
+        isomorphism or None
```
tscrim:
```suggestion
        - ``certificate`` -- boolean (default: ``False``)

        OUTPUT: boolean, and, if ``certificate=True``, a dictionary giving the
        isomorphism or ``None``
```

src/sage/matroids/circuits_matroid.pyx
tscrim: If you want to do something by triangularity, then I think you should do it as such. I agree that the checks are likely not large, there are a lot of them (approx $\binom{n}{2}$ here plus more later on). Although I find this code easier to read and understand with less indentation. (I also got the triangular order backwards; I updated my suggestion above to fix this.) If you really disagree, then please change this to
```python
if i > j:
    continue
```

src/sage/matroids/circuits_matroid.pyx
tscrim: I got the triangularity wrong. Yes, you're right.

tscrim (changes requested):
Thank you for the changes.

I think the way you had the references before was better (and it is more in line with how we do other things in Sage generally).

A few more specific comments.

src/sage/matroids/circuits_matroid.pyx
gmou3: This is incorrect; the circuit can be smaller in size than `C2` (`jj`). Consider a cycle graph with one edge replaced by two parallel edges. Let `C1` and `C2` be the two big cycles with all edges in common except one. Then the `2`-cycle is the `C3` that we are looking for, and that is smaller than the size of `C2`.

gmou3:
I think I found a solution for which we can both be happy (easily readable and avoids checks and indents).

As for the references, by having them in one line we save a lot of whitespace. But I don't mind much either way; let me know if you want me to revert this.

tscrim:
A few blanklines isn’t really a big deal, and the previous way was more readable with how it is formatted IMO.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -0,0 +1,740 @@
...
+                    UC12 = set(C1) | set(C2)
+                    for e in C1 & C2:
+                        flag = False
+                        S = UC12 - {e}
+                        for k in self._k_C:
+                            if k <= len(S) and not flag:
+                                for C3 in self._k_C[k]:
+                                    if C3 <= S:
+                                        flag = True
+                                        break
```
tscrim:
```suggestion
                    I12 = C1 & C2
                    if not I12:  # C1 and C2 are disjoint; nothing to test
                        continue
                    if len(I12) == len(C1):
                        if len(C1) == len(C2):  # they are equal
                            break
                        # a circuit can't be a subset of another circuit
                        return False 
                    # check circuit elimination axiom
                    UC12 = C1 | C2
                    Slen = len(UC12) - 1  # will be the size of S below
                    for e in I12:
                        flag = False
                        S = UC12 - {e}
                        for k in self._k_C:
                            if k <= Slen:
                                for C3 in self._k_C[k]:
                                    if C3 <= S:
                                        flag = True
                                        break
                            if flag:
                                break
```
This should short-curcuit properly to avoid extra checks. Since we expect that nearly all checks will not have containment, we can avoid repeating some iterations over the circuits by simply building the intersection first. I also don't see why we needed to create `set` copies of `C!` and `C2`.

Separately, it might be beneficial to
```
cdef int i, j, k, lenS
cdef bint flag
```
so the comparisons are done at the true C level.

src/sage/matroids/circuits_matroid.pyx
tscrim: Right, I was off by 1. The size of `S` is at least the size of `C2` (which is at least the size of `C1`) since it must be at least one larger than `C2` but then subtracting one for removing `e`. Anyways, moot point since we don't have `jj` with your recent change (which I don't oppose).

src/sage/matroids/circuits_matroid.pyx
gmou3: Looks good. Do you know if it is indeed better to add the `cdef i, j, k`, i.e. for variables defined in for loops? If so there are many other places I should do that.

tscrim: It's good for comparing things that are `int`'s or castable into them. However, that involves some extra checks. The advantage is you get (much) faster comparisons between `int` objects. I am skeptical it will have any (net)  non-negligible speed benefits, so let's just leave it be.

tscrim (changes requested):
Thank you. The only thing left is the order change in `unpickling.pyx`. I think you wanted alphabetical, but it is not. I also am necessarily not in favor of these input sorting-type changes; yet, I won't reject it here (provided it is resorted by some reasonable way).

tscrim (approved):
Then let is be so.

For the future, please avoid changing the import orders. Alphabetical is not always best and these unrelated changes can create unnecessary conflicts.

PR #37156, opened 2024-01-24
using Parent in integer rings instead of ParentWithGens

fchapoton:
Let's try to get rid of the last direct inheritance to `ParentWithGens`.

(not quite the last one, see #37158 )

There remains many indirect inheritance, mostly through `Ring`, `Algebra`, `CommutativeRing` and their subclasses. This is a step towards cleaning and simplifying the old coercion framework.

The present changes is close to the heart of sage, so appropriate timings should be checked.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/integer_ring.pyx
+++ b/src/sage/rings/integer_ring.pyx
@@ -61,7 +61,7 @@ from sage.categories.infinite_enumerated_sets import InfiniteEnumeratedSets
 from sage.rings.number_field.number_field_element_base import NumberFieldElement_base
 from sage.structure.coerce cimport is_numpy_type
 from sage.structure.element cimport parent
-from sage.structure.parent_gens import ParentWithGens
+from sage.structure.parent cimport Parent
 from sage.structure.richcmp cimport rich_to_bool
 
 from sage.misc.misc_c import prod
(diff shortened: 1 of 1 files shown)


kwankyu:
> The present changes is close to the heart of sage, so appropriate timings should be checked.

What timings are necessary? Would you be explicit? Did you do the timings?

tscrim:
I think coercions and conversions with common objects (finite fields, constant polynomials, rational numbers, etc.) would be good. Although I wouldn't expect really any changes in the timings...

fchapoton:
I ran the tests of the `rings` folder and the time taken seemed not to change significantly. Not done scientifically..

kwankyu:
I got, where m is a 1000*1000 matrix of integer entries,

Before this PR:
```
sage: m = load('m')
sage: timeit('m^3+m^2+m')
5 loops, best of 3: 1.14 s per loop
sage: timeit('1/3*m')
5 loops, best of 3: 447 ms per loop
```

After this PR:
```
sage: m = load('m')
sage: timeit('m^3+m^2+m')
5 loops, best of 3: 1.14 s per loop
sage: timeit('1/3*m')
5 loops, best of 3: 444 ms per loop
```

Perhaps not enough scientific either, but I don't perceive any performance regression.

kwankyu (approved):
LGTM

PR #37157, opened 2024-01-24
Add additional lattice methods for matrices over QQ

TheBlupper:
LLL has long been implemented for matrices over QQ by clearing the denominator and doing LLL over ZZ, but this was not done for BKZ() or is_LLL_reduced()

This PR adds BKZ() and is_LLL_reduced() to rational matrices, using the same techniques as the current LLL method.

### DIFF
--- a/src/sage/matrix/matrix_rational_dense.pyx
+++ b/src/sage/matrix/matrix_rational_dense.pyx
@@ -2927,6 +2927,24 @@ cdef class Matrix_rational_dense(Matrix_dense):
     # LLL
     # ###############################################
 
+    def BKZ(self, *args, **kwargs):
+        """
+        Block Korkin-Zolotarev reduction.
+
+        For details on input parameters, see
+        :meth:`sage.matrix.matrix_integer_dense.Matrix_integer_dense.BKZ`.
+
+        EXAMPLES::
+
+            sage: A = Matrix(QQ, 3, 3, [1/n for n in range(1, 10)])
+            sage: A.BKZ()
+            [ 1/28 -1/40 -1/18]
+            [ 1/28 -1/40  1/18]
+            [-1/14 -1/40     0]
+        """
+        A, d = self._clear_denom()
+        return A.BKZ(*args, **kwargs) / d
+
     def LLL(self, *args, **kwargs):
         """
         Return an LLL reduced or approximated LLL reduced lattice for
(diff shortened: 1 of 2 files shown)


src/sage/matrix/matrix_rational_dense.pyx
```diff
@@ -2927,6 +2927,24 @@ cdef class Matrix_rational_dense(Matrix_dense):
     # LLL
     # ###############################################
 
+    def BKZ(self, *args, **kwargs):
+        """
+        Block Korkin-Zolotarev reduction.
+
+        For details on input parameters, see
```
grhkm21: I recommend explicitly writing "the arguments `*args` and `**kwargs` are passed onto ..." (this change should be done for the LLL method as well).

src/sage/matrix/matrix_rational_dense.pyx
```diff
@@ -2927,6 +2927,24 @@ cdef class Matrix_rational_dense(Matrix_dense):
...
+        :meth:`sage.matrix.matrix_integer_dense.Matrix_integer_dense.BKZ`.
+
+        EXAMPLES::
+
+            sage: A = Matrix(QQ, 3, 3, [1/n for n in range(1, 10)])
+            sage: A.BKZ()
+            [ 1/28 -1/40 -1/18]
+            [ 1/28 -1/40  1/18]
+            [-1/14 -1/40     0]
```
grhkm21, line 2945: The example should check for the desired property. Here, I suggest adding a line for `A._clear_denom()` showing the output `(..., 2520)`, and a check `A.BKZ() == (A * 2520).change_ring(ZZ).BKZ() / 2520`. A random test would also be nice: Add an example for `random_matrix`.

src/sage/matrix/matrix_rational_dense.pyx
```diff
@@ -2946,6 +2964,23 @@ cdef class Matrix_rational_dense(Matrix_dense):
         A, d = self._clear_denom()
         return A.LLL(*args, **kwargs) / d
 
+    def is_LLL_reduced(self, delta=None, eta=None):
+        """
+        Return ``True`` if this lattice is `(\delta, \eta)`-LLL reduced.
+        For a definition of LLL reduction, see
+        :meth:`sage.matrix.matrix_integer_dense.Matrix_integer_dense.LLL`.
+
+        EXAMPLES:
```
grhkm21: This should be double colon "EXAMPLES::", and a blank line between this and the first `sage: ...` line.

You should run `sage -tox -e rst <file>` before pushing, it should check for these mistakes. There's also `-tox -e {relint, pycodestyle-minimal}`.

You can also try out my setup: https://gist.github.com/grhkm21/6d20ddd7f11b3c00f889349052a43947

src/sage/matrix/matrix_rational_dense.pyx
```diff
@@ -2927,6 +2927,24 @@ cdef class Matrix_rational_dense(Matrix_dense):
     # LLL
     # ###############################################
 
+    def BKZ(self, *args, **kwargs):
+        """
+        Block Korkin-Zolotarev reduction.
```
grhkm21: Docstrings should ideally start with "Return ..." - refer to the LLL docstring

grhkm21:
I think the changes in the review above has all been made (it was from two hours ago), so you can resolve those. I will wait for CI, but otherwise looks fine.

PR #37158, opened 2024-01-24
use Parent in quotient rings too

fchapoton:
instead of using the deprecated `ParentWithGens`

### :memo: Checklist


fchapoton:
strange:

```
sage: FiniteDimensionalAlgebra(k,[],assume_associative=True)
Finite-dimensional algebra of degree 0 over Finite Field of size 5
```

is not in `Rings()`


and also 

```
sage: C = AlgebrasWithBasis(QQ).example()
sage: C in Rings()
True
sage: C = Algebras(QQ).example()
sage: C in Rings()
False

```

fchapoton:
and now we have that
```
sage: from sage.categories.magmatic_algebras import *
sage: cat1 = MagmaticAlgebras(GF(3)).FiniteDimensional().WithBasis()
sage: cat2 = Algebras(GF(3)).FiniteDimensional().WithBasis()
sage: cat1.Associative()
Category of finite dimensional associative algebras with basis over Finite Field of size 3
sage: cat1.Associative() is cat2
False
sage: cat2
Category of finite dimensional algebras with basis over Finite Field of size 3
```

fchapoton:
I have now made a simple first step in #38375.

mantepse:
The only remaining bug is 
```
file "src/sage/rings/finite_rings/finite_field_constructor.py", line 363, in sage.rings.finite_rings.finite_field_constructor.FiniteFieldFactory
```
right?

mantepse:
And the reason for that failure seems to be that, in the implementation of `QuotientRing_nc.gens`, we have
```
    def gens(self) -> tuple:
        return tuple(self(self.__R.gen(i))
                     for i in range(self.cover_ring().ngens()))
```
and in the case at hand, i.e., `k = GF(5, modulus='primitive')`, the ring `k.__R` is the integer ring, with generator `1`.

Note that `FiniteField_prime_modn.gen` has its own implementation, so maybe we should simply implement `FiniteField_prime_modn.gens`, too?

mantepse:
No?

mantepse:
Juhu!

PR #37159, opened 2024-01-25
Fix `change_ring` method of multi-variate polynomials

### LINKED ISSUE #36832: Random failure in ell_field.py with division_field
### Steps To Reproduce

This raises an error:
```
sage: F.<a> = GF(592337197)
sage: E = EllipticCurve([F(341280459), F(489478549), F(203134672), F(127259459), F(366459012)])
sage: E
Elliptic Curve defined by y^2 + 341280459*x*y + 203134672*y = x^3 + 489478549*x^2 + 127259459*x + 366459012 over Finite Field of size 592337197
sage: E.division_field(3)
```
As a result, this doesn't pass for me (OS X):
```
sage -t --long --random-seed=46499345266195225323927627332877868800 src/sage/schemes/elliptic_curves/ell_field.py
```

### Expected Behavior

The test should pass.

### Actual Behavior

An error is raised:
```
TypeError: unable to convert Identity endomorphism of Finite Field of size 592337197 to an element of Finite Field of size 592337197
```

### Additional Information

_No response_

### Environment

```markdown
- OS X 13.6.1, Intel, plus various `homebrew` packages
- Sage version 10.3.beta0
```

### Checklist

kwankyu:
Fixes #36832.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/polynomial/multi_polynomial.pyx
+++ b/src/sage/rings/polynomial/multi_polynomial.pyx
@@ -871,16 +871,23 @@ cdef class MPolynomial(CommutativePolynomial):
 
         TESTS:
 
-        Check that :trac:`25022` is fixed::
+        Check that :issue:`25022` is fixed::
 
             sage: K.<x,y> = ZZ[]
             sage: (x*y).change_ring(SR).monomials()                                     # needs sage.rings.number_field sage.symbolic
             [x*y]
+
+        Check that :issue:`36832` is fixed::
+
+            sage: F = GF(11)
+            sage: phi = Hom(F,F).an_element()
+            sage: R.<x,y> = F[]
+            sage: x.change_ring(phi)
+            x
         """
         if isinstance(R, Map):
-        #if we're given a hom of the base ring extend to a poly hom
             if R.domain() == self.base_ring():
-                R = self.parent().hom(R, self.parent().change_ring(R.codomain()))
+                return self.map_coefficients(R)
             return R(self)
         else:
             return self.parent().change_ring(R)(self.dict())
--- a/src/sage/rings/polynomial/multi_polynomial.pyx
+++ b/src/sage/rings/polynomial/multi_polynomial.pyx
@@ -841,12 +841,14 @@ cdef class MPolynomial(CommutativePolynomial):
 
     def change_ring(self, R):
         r"""
-        Return a copy of this polynomial but with coefficients in ``R``,
-        if at all possible.
+        Return this polynomial with coefficients converted to ``R``.
 
         INPUT:
 
-        - ``R`` -- a ring or morphism.
+        - ``R`` -- a ring or morphism; if a morphism, the coefficients
+          are mapped to the codomain of ``R``
+
+        OUTPUT: a new polynomial with the base ring changed to ``R``.
 
         EXAMPLES::
 
@@ -854,10 +856,14 @@ cdef class MPolynomial(CommutativePolynomial):
             sage: f = x^3 + 3/5*y + 1
             sage: f.change_ring(GF(7))
             x^3 + 2*y + 1
+            sage: g = x^2 + 5*y
+            sage: g.change_ring(GF(5))
+            x^2
 
         ::
 
-            sage: R.<x,y> = GF(9,'a')[]                                                 # needs sage.rings.finite_rings
+            sage: # needs sage.rings.finite_rings
+            sage: R.<x,y> = GF(9,'a')[]
             sage: (x+2*y).change_ring(GF(3))
             x - y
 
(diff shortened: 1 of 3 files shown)


kwankyu:
Sure. Done. Moreover

Before the patch:
```sage
sage: F = GF(11)
sage: phi = Hom(F, F.extension(2)).an_element()
sage: R.<x,y> = F[]
sage: x.change_ring(phi)
sage: timeit('x.change_ring(phi)')
625 loops, best of 3: 148 μs per loop
```

After the patch:
```sage
sage: F = GF(11)
sage: phi = Hom(F, F.extension(2)).an_element()
sage: R.<x,y> = F[]
sage: x.change_ring(phi)
sage: timeit('x.change_ring(phi)')
625 loops, best of 3: 53 μs per loop
```

Hence the new `change_ring` is faster!

mantepse:
> Hence the new `change_ring` is faster!

I don't understand.  Wasn't #36832 about the following?
```
sage: F = GF(11)
sage: phi = Hom(F,F).an_element()
sage: R.<x,y> = F[]
sage: x.change_ring(phi)
```

kwankyu:
> > Hence the new `change_ring` is faster!
> 
> I don't understand. Wasn't #36832 about the following?
> 
> ```
> sage: F = GF(11)
> sage: phi = Hom(F,F).an_element()
> sage: R.<x,y> = F[]
> sage: x.change_ring(phi)
> ```

Yes. That is fixed by this PR. But as a bonus, the PR speeds up already working cases.

mantepse:
> > > Hence the new `change_ring` is faster!
> > 
> > 
> > I don't understand. Wasn't #36832 about the following?
> > ```
> > sage: F = GF(11)
> > sage: phi = Hom(F,F).an_element()
> > sage: R.<x,y> = F[]
> > sage: x.change_ring(phi)
> > ```
> 
> Yes. That (which didn't work before) is fixed by this PR. But as a bonus, the PR speeds up already working cases.

It would be good to make that a doctest, don't you think?

kwankyu:
> > > > Hence the new `change_ring` is faster!
> > > 
> > > 
> > > I don't understand. Wasn't #36832 about the following?
> > > ```
> > > sage: F = GF(11)
> > > sage: phi = Hom(F,F).an_element()
> > > sage: R.<x,y> = F[]
> > > sage: x.change_ring(phi)
> > > ```
> > 
> > 
> > Yes. That (which didn't work before) is fixed by this PR. But as a bonus, the PR speeds up already working cases.
> 
> It would be good to make that a doctest, don't you think?

Above one? A doctest is already added.

src/sage/rings/polynomial/multi_polynomial.pyx
```diff
@@ -871,16 +872,23 @@ cdef class MPolynomial(CommutativePolynomial):
...
+            sage: phi = Hom(F,F).an_element()
+            sage: R.<x,y> = F[]
+            sage: x.change_ring(phi)
+            x
         """
         if isinstance(R, Map):
-        #if we're given a hom of the base ring extend to a poly hom
             if R.domain() == self.base_ring():
-                R = self.parent().hom(R, self.parent().change_ring(R.codomain()))
+                return self.map_coefficients(R, new_base_ring=R.codomain())
```
tornaria:
```suggestion
                return self.map_coefficients(R)
```
The documentation of `map_coefficients` says:
```
If "f" is a :class:sage.categories.map.Map, then the polynomial will be defined over the codomain of "f".
```
Also, can `self.map_coefficients(R)` be used in any case (regardless of `R.domain() == self.base_ring()`) ?

See: #28541 for the corresponding bug and fix in polynomial rings (c615bb68304) and another way to reproduce with non-identity maps (fe1d9daf0ba).

The original code is from #20067 (1e6769852a). Note that the `change_ring()` method is repeated in `src/sage/rings/polynomial/multi_polynomial_element.py`. When is that one used?

tornaria:
Notes:
 - could you squash together the first two commits? The first one makes a change that the second one changes again. That makes history more difficult to use.
 - is it possible to separate the pycodestyle changes from the actual fix?
 - could mention #37159 (and/or #36832) in the commit message
 - maybe it's reasonable to add a test in `ell_field.py` that (deterministically) triggers the original issue.
 - I'm not sure what happens with the same code in `src/sage/rings/polynomial/multi_polynomial_element.py`? Is it ever used? In any case, it's either redundant (should be removed) or buggy (also it checks `isinstance(R, Morhpism)` instead of `isinstance(R, Map)`, this comes also from #20067 and was fixed in the other two occurrences).

src/sage/rings/polynomial/multi_polynomial.pyx
kwankyu: > The documentation of `map_coefficients` says:
> 
> ```
> If "f" is a :class:sage.categories.map.Map, then the polynomial will be defined over the codomain of "f".
> ```

Thanks. Done.

> Also, can `self.map_coefficients(R)` be used in any case (regardless of `R.domain() == self.base_ring()`) ?

Currently for the case `R.domain() != self.base_ring()`,  the code `R(self)` is used, and replacing this with  `self.map_coefficients(R)` would change the semantic. Changing the semantic is not in the scope of this PR.

> The original code is from #20067 ([1e67698](https://github.com/sagemath/sage/commit/1e6769852ac9b88c50832283014615ee8ed2b6ac)). Note that the `change_ring()` method is repeated in `src/sage/rings/polynomial/multi_polynomial_element.py`. When is that one used?

~~I don't understand your question. What is "that one"? Do you mean this one `return self.parent().change_ring(R)(self.dict())`? This one is used when `R` is a ring.~~

kwankyu: Ah, you meant a different file `src/sage/rings/polynomial/multi_polynomial_element.py`. I will see.

kwankyu: > > Also, can `self.map_coefficients(R)` be used in any case (regardless of `R.domain() == self.base_ring()`) ?
> 
> Currently for the case `R.domain() != self.base_ring()`, the code `R(self)` is used, and replacing this with `self.map_coefficients(R)` would change the semantic. Changing the semantic is not in the scope of this PR.

Looking it again, I am not sure if the original code `R(self)` for  the case `R.domain() != self.base_ring()` would really work for any meaningful cases. 

In the last commit, I removed it. Let's see how this would work out.

kwankyu: Build & Test passed!

tornaria:
LGTM

tornaria (approved):

kwankyu:
Thanks!

PR #37161, opened 2024-01-25
Fix cvxopt build with suitesparse homebrew; fixes #37149.

### LINKED ISSUE #37149: Failing to build cvxopt-1.3.2
### Environment

```markdown
- **OS**: macOS Sonoma 14.2.1
- **Sage Version**: 10.3.beta6
```

### Steps To Reproduce

1. `git clone https://github.com/sagemath/sage.git sage-test && cd sage-test`
2. Install all packages suggested [here](https://doc.sagemath.org/html/en/installation/source.html#macos-package-installation)
3. `` export SAGE_ROOT=`pwd` ``
4. `export MAKE="make -j4"`
5. `source ./.homebrew-build-env`
6. `make configure`
7. `./configure`
8. `make`

### Config log

[config.log](https://github.com/sagemath/sage/files/14025402/config.log)

### Package logs

[cvxopt-1.3.2.log](https://github.com/sagemath/sage/files/14025412/cvxopt-1.3.2.log)

### Additional Information

I think that this is closely related to issue #31905, which has been closed, but it seems not to be fully resolved.

I also tried building with the environment variables `CVXOPT_SUITESPARSE_INC_DIR` and `CVXOPT_SUITESPARSE_LIB_DIR` set to homebrew path of suite-sparse, but that didn't solve the issue either.

### Checklist

drago-96:
This PR tries to fix issue #37149.

The problem is that cvxopt doesn't build correctly when using the homebrew-provided version of SuiteSparse.

As suggested in the discussion of the issue, I added the `CVXOPT_SUITESPARSE_INC_DIR` environment variable in the `.homebrew-build-env` file.

This is an ugly workaround, and should probably be reported/discussed also on the cvxopt repo, since the issue seems to be that they do `#include "umfpack.h"` instead of a more proper `#include "suitesparse/umfpack.h"`.
Indeed, for Linux system, their `setup.py` script then has to set the include folder to be `whatever/include/suitesparse` instead of the default system one.

Also, I don't have a way to check if this breaks other installations.

### :memo: Checklist

### DIFF
--- a/.homebrew-build-env
+++ b/.homebrew-build-env
@@ -45,3 +45,9 @@ for l in ; do
     fi
 done
 export ACLOCAL_PATH
+
+# Add cvxopt library path
+if [ -d "$HOMEBREW/include/suitesparse" ]; then
+    export CVXOPT_SUITESPARSE_INC_DIR="$HOMEBREW/include/suitesparse"
+    export CVXOPT_SUITESPARSE_LIB_DIR="$HOMEBREW/lib"
+fi


drago-96:
Another option could be making cvxopt a pip package, if that doesn't break on some systems.

I tried that on mine (macOS ARM), and it works.

jhpalmieri:
This looks okay to me, although I want to do a bit more testing. If you can figure out how to report it to the cvxopt people (the cvxopt Google group?), I encourage you to do so.

jhpalmieri:
> Another option could be making cvxopt a pip package, if that doesn't break on some systems.
> 
> I tried that on mine (macOS ARM), and it works.

That could be done on another PR independently of this one.

dimpase:
> Another option could be making cvxopt a pip package, if that doesn't break on some systems.
> 
> I tried that on mine (macOS ARM), and it works.

indeed, cvxopt is on PyPI already
https://pypi.org/project/cvxopt/
and there is a month-old binary install for arm64 macOS. So this should work.

You can install cvxopt into the Python you're using, and then
`./configure --enable-system-site-packages` should pick up cvxopt from the "system".

jhpalmieri:
Since this works, I think we should merge it at least as a stopgap. Switching to a pip-installable `cvxopt` sounds good, and maybe somehow upgrading SuiteSparse will also fix this problem, but both of those may take longer to implement. Once other solutions are in place, maybe we undo this.

jhpalmieri (approved):

PR #37170, opened 2024-01-26
Total refactor of `any_root()` to solve issue in characteristic two and clean up the code

### LINKED ISSUE #37160: Slow root finding over GF(2^k)[] quotient rings
### Steps To Reproduce

```python
sage: K.<z16> = GF(2^16)
....: R.<x> = K[]
....: u = R.irreducible_element(2)
....: 
....: K_ext = K.extension(modulus=u, names="a")
....: R_ext.<y_ext> = K_ext[]
....: poly = R_ext.random_monic_element(2)
....: print("poly:", poly)
....: 
....: poly.roots()
```

### Expected Behavior

It should take a short time, especially since here `u` is irreducible, so `K_ext` is in theory isomorphic to `GF(2^32)`, and over that field it takes no time to find roots, just do hensel lift or whatever:

```python
sage: K.<x> = GF(2^32)[]
....: %timeit K.random_monic_element(2).roots()
1.37 ms ± 55.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### Actual Behavior

It's slow... Even changing the base field from 2^16 to 2^8 takes forever. Moreover, the time is spent on this weird ` self._roots_from_factorization(self.factor(), multiplicities)` call. I suspect it is because I specify the modulus in the `K_ext` construction, so it has to compute some field isomorphism. Still, I don't see how 8-bit computations will take this long.

For more details, here's the timings:

```python
sage: K.<z16> = GF(2^8)
....: R.<x> = K[]
....: u = R.irreducible_element(2)
....: 
....: K_ext = K.extension(modulus=u, names="a")
....: R_ext.<y_ext> = K_ext[]
....: %time R_ext.random_monic_element(2).roots()
....: %time _ = [R_ext.random_monic_element(2).roots() for _ in range(10^3)]
CPU times: user 43.8 s, sys: 163 ms, total: 44 s
Wall time: 44.4 s
[]
CPU times: user 3.91 s, sys: 15 ms, total: 3.93 s
Wall time: 3.94 s
```

As seen, a ridiculous amount of time is spent on some kind of field isomorphism / data initialisation.

### Additional Information

I tried tracing the calls myself, but as mentioned, it goes into some weird call path `_roots_from_factorization` that I can't understand. Any help with either debugging the above *or* redirect it to use GF(2^32) like I suggested, would be appreciated.

For more additional information, this behaviour only seem to occur for $K = \mathbb{F}\_q$ where $q = 2^k \geq 2^8$. Also once it finishes (which takes several tens of seconds), the root finding itself is very fast. To observe this, run `while True: R_ext.random_monic_element(2).roots()`. This behaviour also doesn't happen, even for weird fields like $K = \mathbb{F}_{103^{13}}$ the whole process takes < 1s.

GiacomoPope:
This PR was inspired by the issue #37160, which found that computing isomorphisms between fields of even characteristic was taking much much longer than expected (20s vs 100ms).

My first attempt (see commit history) were to simply patch up the `any_root()` function, but this yielded some results which left the code in an even more confusing state.

Following the advice of the reviewers, I have now implemented `any_irreducible_factor()` which given a polynomial element computes an irreducible factor of this polynomial with minimal degree. When the optional parameter `degree=None` is set, then an irreducible factor of degree `degree` is computed and a `ValueError` is raise if no such factor exists.

Now, `any_root()` becomes simply a call to `self. any_irreducible_factor(degree=1)` and a root is found from this linear polynomial. We also handle all the other cases of optional arguments handled by the old function, so the function *should* behave as before, but with a cleaner code to read.

### Before PR

```python
sage: set_random_seed(0)
....: K.<z8> = GF(2^8)
....: R.<x> = K[]
....: u = R.irreducible_element(2)
....: K_ext = K.extension(modulus=u, names="a")
....: R_ext.<y_ext> = K_ext[]
....: poly = R_ext.random_element(2).monic()
....: %time poly.roots()
CPU times: user 20.5 s, sys: 19.9 ms, total: 20.5 s
Wall time: 20.5 s
[]
```

### After PR

```py
│ SageMath version 10.3.beta6, Release Date: 2024-01-21              │
sage: set_random_seed(0)
....: K.<z8> = GF(2^8)
....: R.<x> = K[]
....: u = R.irreducible_element(2)
....: K_ext = K.extension(modulus=u, names="a")
....: R_ext.<y_ext> = K_ext[]
....: %time R_ext.random_element(2).roots()
CPU times: user 110 ms, sys: 9.03 ms, total: 119 ms
Wall time: 150 ms
[]
```
This fixes #37160 but i think more feedback and thorough doctests need to be included considering this is a very old function which many projects will be using.


GiacomoPope:
Also, `any_roots()` seems very complicated to me, but maybe it's necessary. I suppose this could be an opportunity to refactor the whole function.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2285,27 +2285,39 @@ cdef class Polynomial(CommutativePolynomial):
...
-                        else:
-                            return (self//h).any_root(ring, -degree, True)
+            if k == 1 or q % 2 == 0:
+                # Is there something better to do here for k = 1?
+                roots = self.roots(ring, multiplicities=False)
+                if roots:
+                    from sage.misc.prandom import choice
+                    return choice(roots)
+                else:
+                    raise ValueError(f"no roots F {self}" % self)
```
grhkm21: Small change, but the `% self` is useless here.

grhkm21 (changes requested):
Request a review again after changing, and I'll approve it

GiacomoPope:
I think I will try and refactor `any_root` as the more I look at it, the less I understand it and think it can be simplified.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
-                    if A != 1:
-                        self = self // A
-                if d == degree:
-                    self = self.gcd(xq-x)
-                    if self.degree() == 0:
-                        raise ValueError("no roots E %s" % self)
-            else:
-                degree = -degree
+        # Currently I can't see a benefit of the degree method, so I think we should
+        # remove it
```
yyyyx4: I also haven't ever used this, but I'm not sure if other people have applications for it. Maybe I'm missing something, but wouldn't it be easy enough to keep the API unchanged? We simply have to divide out the $\gcd$ of $f$ with $x^{q^{n/m}}-x$ for all prime divisors $m$ of the given degree $n$ prior to running the rest of the algorithm?

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+        if degree is not None:
+            from sage.misc.superseded import deprecation
+            deprecation(37170, "'degree' will soon be deprecated as it is no longer needed")
+
+        # When not working over a finite field, do the simple 
+        # thing of factoring for roots and picking the first 
+        # root. If none available, raise an error.
+        from sage.categories.finite_fields import FiniteFields
+        if not self.base_ring() in FiniteFields():
```
yyyyx4, line 2556: `not ... in ...` → `... not in ...`

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+            rs = self.roots(ring=ring, multiplicities=False)
+            if rs:
+                return rs[0]
+            raise ValueError("polynomial {self} has no roots" % self)
+
+        # For finite fields, we find roots in the following three steps:
+        #
+        # 1. Compute the squarefree decomposition of the polynomial
+        # 2. For each squarefree polynomial find the distinct degree = 1
+        #    factorison, F, which is the product of degree one polynomials
```
yyyyx4: "factorison" → "factorisation"?

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
-                while True:
-                    T = R.random_element(2*degree-1)
-                    if T == 0:
-                        continue
+                return -f[0] / f[1]
+            return ring(-f[0] / f[1])
+
+        def _any_root_squarefree(f, ring=None):
+            # Compute the distinct degree factorisation for degree one, 
+            # the output will be the product of #all degree one
```
yyyyx4: `#all` → `all`

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+                raise ValueError(f"no root can be computed for {f}")
+
+            # Otherwise, we can find a root from the CZ splitting of g
+            return _cantor_zassenhaus_split_root(g, 1, ring=ring)
+
+        def _cantor_zassenhaus_split_root(f, degree, ring=None):
+            R = f.parent()
+            q = f.base_ring().order()
+            # Need to handle odd and even charcteristic separately
+            if q % 2 != 0:
```
yyyyx4:
```suggestion
            if q % 2:
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+            elif g.degree().is_zero():
+                raise ValueError(f"no root can be computed for {f}")
+
+            # Otherwise, we can find a root from the CZ splitting of g
+            return _cantor_zassenhaus_split_root(g, 1, ring=ring)
+
+        def _cantor_zassenhaus_split_root(f, degree, ring=None):
+            R = f.parent()
+            q = f.base_ring().order()
+            # Need to handle odd and even charcteristic separately
```
yyyyx4: "charcteristic" → "characteristic"

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
                         else:
-                            return (self//h).any_root(ring, -degree, True)
-        else:
-            rs = self.roots(ring=ring, multiplicities=False)
-            if rs:
-                return rs[0]
-            raise ValueError("no roots (non-field) %s" % self)
+                            return _cantor_zassenhaus_split_root(f//h, degree, ring=ring)
+                # If you are reaching this error, chances are there's a bug in the code.
+                raise ValueError(f"no splitting of degree {degree} found for {f}")
```
yyyyx4: This should be an `AssertionError`: It's something that should never happen unless the code is wrong somehow.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+
+        def _cantor_zassenhaus_split_root(f, degree, ring=None):
+            R = f.parent()
+            q = f.base_ring().order()
+            # Need to handle odd and even charcteristic separately
+            if q % 2 != 0:
+                # We expect to succeed with 1/2 probability, so if we
+                # try 1000 times and fail, there's a bug somewhere.
+                for _ in range(1000):
+                    T = R.random_element(2*degree + 1)
```
yyyyx4: I suppose this should be a uniformly random polynomial of degree less than $\deg(f)$. This works, too, but it's less streamlined than it could be (and it also slightly breaks the `.is_zero()` check below).

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+                    if T.is_zero(): continue # skip T = 0
```
yyyyx4: I think this actually never skips anything because the degree of $T$ is $2\deg(f)+1$ which is $\geq1$, right?

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
                     T = T.monic()
-                    h = self.gcd(pow(T, Integer((q**degree-1)/2), self)-1)
+
+                    # Compute the gcd. 50% chance this is a non-trivial factor of f
```
yyyyx4: It's actually *at least* 50%: The probability of splitting is lower bounded by $1-(1/2)^{\deg(f)-1}$.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
-                return rs[0]
-            raise ValueError("no roots (non-field) %s" % self)
+                            return _cantor_zassenhaus_split_root(f//h, degree, ring=ring)
+                # If you are reaching this error, chances are there's a bug in the code.
+                raise ValueError(f"no splitting of degree {degree} found for {f}")
+
+            # Now handle even charactertistic
+            for _ in range(1000):
+                # Sample a uniformly random element of R
+                T = R.random_element(2*degree + 1)
```
yyyyx4: Same comments here.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+                if T.is_zero(): continue
+                T = T.monic()
+
+                # Compute the trace of T with field of order 2^k
+                # sum T^(2^i) for i in range (degree * k)
+                # We use repeated squaring to avoid redundent multiplications
+                C, TT = T, T
+                for _ in range(degree * f.base_ring().degree() - 1):
+                    TT = pow(TT, 2, f)
```
yyyyx4: Not sure if it makes a difference performance-wise, but this could just be `TT * TT % f`.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+
+                # else check if we have a factor
+                elif not hd.is_zero() and hd != f.degree():
+                    if 2*hd <= f.degree():
+                        return _cantor_zassenhaus_split_root(h, degree, ring)
+                    else:
+                        return _cantor_zassenhaus_split_root(f // h, degree, ring)
+
+            # If you are reaching this error, chances are there's a bug in the code.
+            raise ValueError(f"no splitting of degree {degree} found for {f}")
```
yyyyx4: This should also be an `AssertionError`.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2168,162 +2168,151 @@ cdef class Polynomial(CommutativePolynomial):
...
+        #   End helper functions   #
+        # ======================== #
+
+        # Initial checks for bad input
+        if self.degree() < 0:
+            if ring is None:
+                return self.base_ring()(0)
+            return ring(0)
+        elif self.degree() == 0:
+            raise ValueError(f"no root can be computed for {self}")
```
yyyyx4: We may want to point out to the user *why* we can't compute a root — there aren't any!

src/sage/rings/polynomial/polynomial_element.pyx
GiacomoPope: Yeah I think we can do this. The original code also allowed negative degree which did something optional before taking the absolute sign. The thing is I just don't see what the benefit is. We have some input f we want roots of and allowing degree = 1 means we find all possible linear polynomials. When the degree is larger we just find factors which may not have rational roots?

GiacomoPope: More specifically, the old documentation says:

```py
        - ``degree`` (None or nonzero integer) -- Used for polynomials
          over finite fields.  Return a root of degree
          ``abs(degree)`` over the ground field.  If negative, also
          assumes that all factors of this polynomial are of degree
          ``abs(degree)``.  If None, returns a root of minimal degree
          contained within the given ring.
```

I don't really know what "root of degree n" means, unless what it actually means is that it first finds some polynomial A_d = prod_i^n g_i(x) with g_i(x) of degree n? In this case for a root, we seem to need d = 1 -- unless the idea is we find some polynomial of degree n over the base field which is irreducible and then find a root of this over some extension?

I wish I understood this better.

yyyyx4: Suppose you know a priori that the root you're looking for lies in an extension of degree $k$, but you haven't yet found a suitable irreducible polynomial to construct that extension. Then it should be faster to extract an irreducible factor from your polynomial and construct the extension with that particular polynomial (hence $x$ is the desired root) than to construct the extension using an arbitrary irreducible polynomial and running the root-finding algorithm in the extension.
One particular example where I think this may help is when constructing an ℓ‑torsion basis of an elliptic curve: We often know the required degree from point counting, but we still need to find the actual extension field and points.

yyyyx4: The current behavior is indeed something like that:
```
sage: x = polygen(GF(11^2))
sage: f = x^3+x^2+x+2
sage: f.any_root(degree=3)
9*z6^5 + 5*z6^4 + 5*z6^3 + 10*z6^2
```
...but it looks like it currently just runs `.extension()` on the base field to build the extension (rather than first finding an irreducible factor and making use of it), so there's essentially no point in doing this with the current implementation.

yyyyx4: I think a better interface for this would be `.any_irreducible_factor()` with an optional `degree=` argument. This way the user can do with the returned factor whatever they want (without having the library forcibly construct a field extension for them in case of degree $>1$), and the `.any_root()` method would be a simple wrapper.

GiacomoPope: Ahh ok then this is indeed something which maybe could be included. Maybe I would like the parameter to be something like "extension degree" rather than degree which appears in many places... but that's a whole other thing. 

When fixing everything else from the above review I'll try my hand at this too using your idea

src/sage/rings/polynomial/polynomial_element.pyx
GiacomoPope: I'm anticipating the change in https://github.com/sagemath/sage/pull/37118 which i believe will return 0 with the correct weighting?

grhkm21: No, `degree=d` only returns degree $d$ polynomials, if you want anything between degree $-\infty$ and $d$, write `R.random_element(degree=(-1, 2*degree+1))`. The API is still the same as current `develop` branch.

GiacomoPope: Thank you for the clarification

grhkm21: Or if you want to just skip zero (like you are doing here), you can write `degree=(0, 2*degree+1)` ($0$ is represented as degree=-1)

GiacomoPope:
@yyyyx4 I'm struggling with the logic of the old version of `any_root()` and how the ring and degree parameters seem to want to behave.

```py
sage: R.<x> = GF(11)[]
....: g = (x-1)*(x^2 + 3*x + 9) * (x^5 + 5*x^4 + 8*x^3 + 5*x^2 + 3*x + 5)
....: g.any_root(ring=GF(11^10, 'b'), degree=1)
1
sage: g.any_root(ring=GF(11^10, 'b'), degree=2)
5*b^9 + 4*b^7 + 4*b^6 + 8*b^5 + 10*b^2 + 10*b + 5
sage: g.any_root(ring=GF(11^10, 'b'), degree=5)
5*b^9 + b^8 + 3*b^7 + 2*b^6 + b^5 + 4*b^4 + 3*b^3 + 7*b^2 + 10*b
sage: 
sage: g_ext = g.change_ring(GF(11^10, 'b'))
sage: g_ext.any_root(degree=1)
4*b^9 + 8*b^8 + 8*b^7 + 8*b^6 + 3*b^5 + b^4 + 4*b^3 + 4*b^2 + 2*b
sage: g_ext.any_root(degree=2)
Cell In [6], line 1
----> 1 g_ext.any_root(degree=Integer(2))

File ~/sage/sage-10.2/src/sage/rings/polynomial/polynomial_element.pyx:2257, in sage.rings.polynomial.polynomial_element.Polynomial.any_root()
   2255 if 2*d > e:
   2256     if degree != e:
-> 2257         raise ValueError("no roots D %s" % self)
   2258     break
   2259 d = d+1

ValueError: no roots D 1
sage: g.any_root(ring=GF(11^10, 'b'), degree=2).parent()
Finite Field in b of size 11^10
sage: g.factor()
(x + 10) * (x^2 + 3*x + 9) * (x^5 + 5*x^4 + 8*x^3 + 5*x^2 + 3*x + 5)
sage: g_ext.factor()
(x + 10) * (x + 6*b^9 + 10*b^8 + 8*b^7 + 9*b^6 + 10*b^5 + 7*b^4 + 8*b^3 + 4*b^2 + b) * (x + 7*b^9 + 3*b^8 + 3*b^7 + 3*b^6 + 8*b^5 + 10*b^4 + 7*b^3 + 7*b^2 + 9*b) * (x + 6*b^9 + 7*b^7 + 7*b^6 + 3*b^5 + b^2 + b + 6) * (x + 5*b^9 + 4*b^7 + 4*b^6 + 8*b^5 + 10*b^2 + 10*b + 8) * (x + 4*b^9 + 9*b^8 + 5*b^7 + 8*b^6 + 9*b^5 + 4*b^3 + 4*b^2 + 9) * (x + 9*b^9 + 8*b^8 + 4*b^7 + 9*b^6 + 8*b^5 + 7*b^4 + 5*b^3 + 3*b + 9) * (x + 7*b^9 + 3*b^8 + 2*b^7 + 4*b^6 + 9*b^5 + 9*b^4 + 9*b^3 + 7*b^2 + 9*b + 9)
```

So for my new implementation, if someone adds a `ring` param I assume they want to find factors in this extension, but when the ring and degree params are together, it seems the old one seems to try and find a root in the base field before extension and then casts this the bigger field at the end. This is what allows `g.any_root(degree=1)` to always return `1`, despite the fact that `GF(11^10)` is the splitting field so any of these roots are "Good".

Should I try and do more work to align these ideas? I dont really see their logic and why what they have here is useful but my doctests fail because over the splitting field there are no degree two irreducible factors so my code breaks.

GiacomoPope:
With some small tweaks to the doctests (root finding is not deterministic anymore, which i guess it never was due to using CZ, but the doctests assumed it was for certain cases?)

The `any_root()` now seems to behave as it used to. Although making this happen made the code uglier, so I think it would be good to deprecate certain API. I think the case with `degree` and `ring` both not None is the weirdest of the choices and could be avoided, but i need opinions.

GiacomoPope:
New push should get this "finished" or totally ready for review.

## Summary

- Rewrote `any_root()` as the previous implementation was too complex and had many bugs
- Now isomorphisms for `GF(2^k)` actually work, rather than randomly sampling roots until a factor is found! For GF(2^8) we now can compute isomorphisms in 100ms instead of 20-40 seconds!!
- Deprecated allowing negative degree to make an assumption about the factorisation of the input. Instead we have a new `bool` for this.
- Instead of falling back to factorising and taking roots when `degree > 0`, we now recursively call `any_root()` on the degree `degree` irreducible polynomial over an extension ring.
- Added this new bool throughout the codebase and removed all negative degree inputs.
- New implementation is now no longer deterministic. The old one was "sometimes, depending on optional parameters....". This caused an issue with some doctests which I added # random tags to.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
 
-    def any_root(self, ring=None, degree=None, assume_squarefree=False):
+    def _distinct_degree_factorisation_squarefree(self):
+        """
+        Helper function for any_irreducible_factor which computes
+        the distinct degree factorisation of `self`.
+
+        Creates an iterator for all valid degrees `d`, and return
+        tuples of the form `(a_d, d)` for a polynomial `a_d` the
+        product of irreducible polynomials of degree `d`
```
yyyyx4:
```suggestion
        product of irreducible polynomials of degree `d`.
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
         else:
             return (False, None) if root else False
 
-    def any_root(self, ring=None, degree=None, assume_squarefree=False):
+    def _distinct_degree_factorisation_squarefree(self):
+        """
+        Helper function for any_irreducible_factor which computes
+        the distinct degree factorisation of `self`.
+
+        Creates an iterator for all valid degrees `d`, and return
```
yyyyx4:
```suggestion
        Creates an iterator for all valid degrees `d`, and returns
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        tuples of the form `(a_d, d)` for a polynomial `a_d` the
+        product of irreducible polynomials of degree `d`
+
+        Assumes that self is squarefree.
```
yyyyx4:
```suggestion
        Assumes that this polynomial is squarefree.
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        # Iterate over all possible degrees with degree
+        while 2*(d + 1) <= e:
+            d = d + 1
+            w = pow(w, q, v)
+
+            ad = v.gcd(w - x)
+            yield (ad, d)
+
+            if not ad.is_one():
+                v = v // ad
```
yyyyx4:
```suggestion
            if not ad.is_one():
                yield (ad, d)
                v = v // ad
```

The current version only yields *some* of the pairs `(1, d)`; see for instance the example above with degrees `1,2,3,7`. Suggesting to unify this behaviour.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+                w = w % v
```
yyyyx4: I think this line is not needed: The `pow()` call in the next iteration should reduce the input modulo `v` first thing.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+            e = v.degree()
+
+        # Last case, v itself might be irreducible
+        if e > 0:
+            yield (v, e)
+        return
+
+    def _cantor_zassenhaus_split_to_irreducible(self, degree):
+        """
+        Helper function for any_irreducible_factor which computes
```
yyyyx4:
```suggestion
        Helper function for :meth:`any_irreducible_factor` which computes
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        a factor from a polynomial of the form `self = prod g_i(x)`
+        with all `g_i(x)` having the same degree. Uses the Cantor
+        Zassenhaus splitting method.
```
yyyyx4:
```suggestion
        with all `g_i(x)` irreducible of the same degree. Uses the
        Cantor-Zassenhaus splitting method.
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        """
+        R = self.parent()
+        q = self.base_ring().order()
+
+        # Polynomial is already irreducible by the assumptions of the function
+        if self.degree() == degree:
+            return self
+
+        # We expect to succeed with greater than 1/2 probability,
+        #so if we try 1000 times and fail, there's a bug somewhere.
```
yyyyx4:
```suggestion
        # so if we try 1000 times and fail, there's a bug somewhere.
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        for _ in range(1000):
+            # Sample a polynomial uniformly from R
+            T = R.random_element(2*degree + 1).monic()
```
yyyyx4, line 2172: This way of sampling is essentially a temporary workaround until #37118 is done. Maybe add a comment to that effect?

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+
+            # Need to handle odd and even characteristic separately
+            if q % 2:
+                h = self.gcd(pow(T, (q-1)//2, self)-1)
```
yyyyx4:
```suggestion
                h = self.gcd(pow(T, (q-1)//2, self) - 1)
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+                h = self.gcd(C)
+
+            hd = h.degree()
+
+            # If we found a factor of desired degree, return it
+            if hd == degree:
+                return h
+
+            # Else check if we have a non-trivial factor and keep going
+            if not hd.is_zero() and hd != self.degree():
```
yyyyx4:
```suggestion
            if 0 < hd < self.degree():
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+                if 2*hd <= self.degree():
+                    return h._cantor_zassenhaus_split_to_irreducible(degree)
+                else:
+                    return (self//h)._cantor_zassenhaus_split_to_irreducible(degree)
```
yyyyx4:
```suggestion
                if 2*hd > self.degree():
                    h = self // h
                return h._cantor_zassenhaus_split_to_irreducible(degree)
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        raise AssertionError(f"no splitting of degree {degree} found for {self}")
+
+    def _any_irreducible_factor_squarefree(self, degree=None):
+        """
+        Helper function for any_irreducible_factor which computes
+        an irreducible factor from self, assuming the input is
+        squarefree.
+
+        Does this by first computing the distinct degree factorisations
+        of self and thenfinds a factor with Cantor-Zassenhaus
```
yyyyx4:
```suggestion
        of self and then finds a factor with Cantor-Zassenhaus
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+            x + z4^3 + z4^2 + z4
+            sage: f._any_irreducible_factor_squarefree(degree=2)
+            x^2 + x + z4^3 + 1
+            sage: f._any_irreducible_factor_squarefree(degree=3)
+            x^3 + (z4^3 + z4)*x^2 + z4^2 + 1
+        """
+        # If the degree is not None we only want to check a single polynomial
+        if degree is not None:
+            for (poly, d) in self._distinct_degree_factorisation_squarefree():
+                if d == degree and not poly.is_one():
```
yyyyx4: I suppose if `_distinct_degree_factorisation_squarefree()` is adjusted to never return a tuple with `poly == 1` (see above) then this check can be simplified.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+                elif d > degree:
+                    raise ValueError(f"no irreducible factor of degree {degree} could be computed from {self}")
+            raise ValueError(f"no irreducible factor of degree {degree} could be computed from {self}")
+
+        # Otherwise we check all degrees, starting from the smallest
+        for (poly, d) in self._distinct_degree_factorisation_squarefree():
+            # Skip the split checking if `_distinct_degree_factorisation_squarefree`
+            # has found no elements of degree `d`
+            if poly.is_one():
+                continue
```
yyyyx4: Same here.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+          over finite fields. Attempts to return an irreducible factor of
+          ``self`` of chosen degree ``degree``.
+
+        - ``assume_squarefree`` (bool) -- Used for polynomials over
+          finite fields.  If ``True``, this polynomial is assumed to be
+          squarefree.
+
+        - ``assume_distinct_deg`` (bool) -- Used for polynomials over
+          finite fields.  If ``True``, this polynomial is assumed to be
+          the product of irreducible polynomials of the same degree.
```
yyyyx4: Here it should be mentioned that this argument only makes sense if `degree` is given as well.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+
+        # When not working over a finite field, do the simple thing of factoring.
+        # If degree is none, we return the first factor, otherwise we iterate
+        # through and look for a polynomial with the desired degree.
+        from sage.categories.finite_fields import FiniteFields
+        if self.base_ring() not in FiniteFields():
+            try:
+                factorisation = self.factor()
+            except (NotImplementedError, ValueError):
+                raise ValueError(f"Cannot factor {self} over the base ring {self.base_ring()}")
```
yyyyx4:
```suggestion
                raise ValueError(f"cannot factor {self} over the base ring {self.base_ring()}")
```
All the standard exceptions in Python start with a lowercase letter; we should probably adhere to this convention.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2047,7 +2047,376 @@ cdef class Polynomial(CommutativePolynomial):
...
+        #    dividing the squarefree polynomial
+        # 3. Using Cantor-Zassenhaus splitting with degree `d` to find a
+        #    single linear factor and return the root.
+        #
+        # When degree is None, we check all degrees smaller than the degree of the
+        # squarefree polynomial, otherwise we work with only a single degree set by
+        # the user.
+
+        # Initial checks for bad input
+        if self.degree().is_zero():
```
yyyyx4:
```suggestion
        if self.degree() <= 0:
```
The degree of the zero polynomial is `-1` by convention.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2166,164 +2548,81 @@ cdef class Polynomial(CommutativePolynomial):
...
-                    hd = h.degree()
-                    if hd != 0 and hd != self.degree():
-                        if 2*hd <= self.degree():
-                            return h.any_root(ring, -degree, True)
-                        else:
-                            return (self//h).any_root(ring, -degree, True)
-        else:
+            ValueError: polynomial x^2 + 1 has no roots
+        """
+        # When not working over a finite field, do the simplecthing of factoring for
```
yyyyx4:
```suggestion
        # When not working over a finite field, do the simple thing of factoring for
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2166,164 +2548,81 @@ cdef class Polynomial(CommutativePolynomial):
...
+            return - f[0] / f[1]
+
+        # The old version of `any_root()` allowed degree < 0 to indicate that the input polynomial
+        # had a distinct degree factorisation, we pass this to any_irreducible_factor as a bool and
+        # ensure that the degree is positive.
+        degree = ZZ(degree)
+        if degree < 0:
+            from sage.misc.superseded import deprecation
+            deprecation(37170, "negative ``degree`` will be disallowed. Instead use the bool `assume_distinct_deg`.")
+            degree = ZZ(abs(degree))
```
yyyyx4:
```suggestion
            degree = -degree
```

yyyyx4 (approved):
Looks good to me. Thanks a lot for all the hard work!

GiacomoPope:
Thanks for all the reviews and advice on refactoring. (I think) I even feel happy with how this function works haha!

vbraun:
I'm getting
```
sage -t --warn-long 40.4 --random-seed=123 src/sage/rings/polynomial/polynomial_quotient_ring_element.py
File "src/sage/rings/polynomial/polynomial_quotient_ring_element.py", line 717, in sage.rings.polynomial.polynomial_quotient_ring_element.PolynomialQuotientRingElement.minpoly
Failed example:
    ext(u + 1).minpoly()  # indirect doctest                              # needs sage.modules
Expected:
    x^3 + (396*i + 428)*x^2 + (80*i + 39)*x + 9*i + 178
Got:
    x^3 + (35*i + 428)*x^2 + (351*i + 39)*x + 422*i + 178
1 item had failures:
   1 of  19 in sage.rings.polynomial.polynomial_quotient_ring_element.PolynomialQuotientRingElement.minpoly
    [160 tests, 1 failure, 0.51 s]
sage -t --warn-long 40.4 --random-seed=123 src/sage/rings/polynomial/polynomial_quotient_ring_element.py  # 1 doctest failed
```

src/sage/rings/finite_rings/hom_finite_field_givaro.pyx
```diff
@@ -105,7 +105,7 @@ cdef class SectionFiniteFieldHomomorphism_givaro(SectionFiniteFieldHomomorphism_
             sage: K.<T> = GF(3^4)
             sage: f = FiniteFieldHomomorphism_givaro(Hom(k, K))
             sage: g = f.section()
-            sage: g(f(t+1))
+            sage: g(f(t+1)) # random
```
vbraun: There are a great number of added `# random` tags, are those really random? Isn't this test checking an important identity?

GiacomoPope: Yeah, I am uncomfortable with this, but all these doctests seem to depend on the old, undocumented behaviour of `.any_root()` being deterministic in certain places. Before, the end of the recursion would be a call of `f.roots()[0]`, which fully factors what remains and always returns the same root. 

Now, we recursively call `f.any_root()` and this uses CZ for the splitting and so is not deterministic. 

In my opinion "any_root()" should not be deterministic, but if we want these to be fixed then we need to have that the method in `hom` calls `f.roots()[0]` instead of the current `any_root()`.

I would love to have more advice on this though!

GiacomoPope:
Thank you for reporting the error -- I'll see whether I can find the root of this, I assume it'll be again that `any_root()` is no longer deterministic

src/sage/rings/finite_rings/hom_finite_field_givaro.pyx
yyyyx4: A lot of these tests really are `# random` since field embeddings aren't unique: They were deterministic due to a particular implementation choice, but that wasn't really guaranteed anywhere and @GiacomoPope's new algorithm happens to be (1) better, and (2) randomized.
This particular test, however, seems like it should still be deterministic: Here `g` is supposed to be a left inverse of `f` by construction, and the test tests that by evaluating the composition on a particular input value.

GiacomoPope: Sorry for my confusion above, I did not notice that I had over-zealously added random tags to these special tests. I have removed this in the latest commit.

vbraun:
I can live with it being nondeterministic 

Ideally we'd not just slap `# random` on tests but also check something that is true (like the `g(f(t+1)) == t+1`), both as clarification for the docs and to guard against regressions.

GiacomoPope:
Right, I could look at including checks like these in more places, at least for the `hom_*` files where the `# random` has appeared the most. If it helps this PR I am happy to take some time to rewrite some of the `hom` doctests to be better "tests" in light of the change to non-determinism.

Would changes like this be needed in other files too, or simply `hom_finite_field*`? For the `PCL.polynomial(3)` I am afraid I have no experience here and do not know "good" doctests

vbraun:
If its easy enough to add then go for it, but if not then not.

GiacomoPope:
Modified the tests to check things like `f(a + b) == f(a) + f(b)` and also made the section tests more explicit by picking random `a` and then ensuring `g(f(a)) == a`.

Did not touch the `PCL.polynomial` code again, as I'm just not sure what to do.

PR #37173, opened 2024-01-27
Implemented `.ramified_places` and modified further methods to extend quaternion algebra functionality to number fields

S17A05:
1. Implemented method `.ramified_places` for quaternion algebras over number fields. Integrated `.ramified_primes()` into it in the process.
2. Modified `.is_division_algebra()`, `.is_matrix_ring()` and `.is_isomorphic` to use `.ramified_places` instead of `.discriminant()`, thus extending them to base number fields.
3. Rerouted `.discriminant()` through `.ramified_places` since the original call to `.hilbert_conductor` also computed all finite ramified places.
4. Added `.is_totally_definite()` and moved `is_definite()`.

Some more detail:
1. The new method `.ramified_places` returns all places at which the quaternion algebra `self` ramifies; this includes the infinite places by default, but can be reduced to only the finite places with the optional parameter `inf`. The old version of `.ramified_primes()` from #37164 has been integrated into `.ramified_places`, thus setting the former up for possible future deprecation; currently it calls `self.ramified_places(inf=False)` for backwards compatibility.

2. `.is_division_algebra()` and `.is_matrix_ring()` now instead check whether the list of ramified places (finite and infinite) is trivial. `.is_isomorphic` now compares the set of finite ramified places and, unless working over $\mathbb{Q}$, the list of infinite ramified places of both algebras. The latter can be compared as lists since the real embeddings of the number field are sorted independently of each algebras' invariants, but the former (probably) need to be compared as sets since the order of the list depends on the primes above the respective invariants. The docstring of `.is_isomorphic` (as well as some of the other docstrings) now includes an example of a non-split quaternion algebra with trivial discriminant, namely the algebra with invariants $(-1,-1)$ over the quadratic field $\mathbb{Q}(\sqrt{5})$.

Possible future work:
- Extend functionality to all global fields (of characteristic not equal to $2$) [UPDATE: Will be done once both this PR and #37554 have been merged]


S17A05:
Fixed some whitespaces and blank lines discovered by lint. Also corrected the formatting of references to Voight's book - thanks to @grhkm21 for pointing this out to me!

S17A05:
After fixing some typos and some errors in the docstrings, I believe this should be ready for review now.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -403,50 +403,61 @@ def is_commutative(self) -> bool:
...
-
+            NotImplementedError: base field must be rational numbers or a number field
         """
-        if not is_RationalField(self.base_ring()):
-            raise NotImplementedError("base field must be rational numbers")
-        return self.discriminant() == 1
+        try:
+            return self.ramified_places(inf=True) == ([], [])
+        except ValueError:
+            raise NotImplementedError("base field must be rational numbers or a number field")
```
grhkm21: Correct me if I'm wrong, but isn't this just `not self.is_division_algebra()`?

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
-            return ZZ.prod(self.ramified_primes())
+        F = self.base_ring()
+        if is_RationalField(F):
+            return self.is_definite()
+
+        if F not in NumberFields():
+            raise ValueError("base field must be rational numbers or a number field")
+
+        E = F.real_embeddings()
+        return [e for e in E if F.hilbert_symbol(self._a, self._b, e) == -1] == E
```
grhkm21:
```suggestion
        return all(F.hilbert_symbol(self._a, self._b, e) == -1 for e in E)
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+
+            sage: QuaternionAlgebra(210,-22).ramified_places()
+            ([2, 3, 5, 7], [])
+
+            sage: QuaternionAlgebra(-1, -1).ramified_places()
+            ([2],
+             [Ring morphism:
+                From: Rational Field
+                To:   Real Field with 53 bits of precision
+                Defn: 1 |--> 1.00000000000000])
```
grhkm21, line 1327:
```suggestion
                Defn: 1 |--> 1.00000000000000])
                
            ::
```
You can use `::` to separate these blocks of examples. This is what it looks like right now (it should be in separate blocks)

Screenshot 2024-03-27 at 05 19 40

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+
+        if not inf:
+            return ram_fin
+
+        # At this point the infinite ramified places also need to be computed
+        return ram_fin, [e for e in F.real_embeddings() if F.hilbert_symbol(a, b, e) == -1]
 
     @cached_method
     def ramified_primes(self):
         """
```
grhkm21:
```suggestion
        r"""
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+        Return the (finite) primes of the base number field at
+        which this quaternion algebra ramifies.
 
         OUTPUT:
 
-        The list of prime numbers at which ``self`` ramifies (given as integers), sorted by their
-        magnitude (small to large).
+        The list of finite primes at which ``self`` ramifies; given as
+        integers, sorted small to large, if ``self`` is defined over
+        `\\QQ`, and as fractional ideals in the ring of integers of the
```
grhkm21:
```suggestion
        `\QQ`, and as fractional ideals in the ring of integers of the
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+        if is_RationalField(F):
+            return ZZ.prod(self.ramified_places(inf=False))
 
-        return sorted([p for p in set([2]).union(prime_divisors(self._a.numerator()),
-                prime_divisors(self._a.denominator()), prime_divisors(self._b.numerator()),
-                prime_divisors(self._b.denominator())) if hilbert_symbol(self._a, self._b, p) == -1])
+        return F.ideal(F.prod(self.ramified_places(inf=False)))
 
     def is_isomorphic(self, A) -> bool:
         """
```
grhkm21:
```suggestion
        r"""
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
-        Return ``True`` if (and only if) ``self`` and ``A`` are isomorphic quaternion algebras over Q.
+        Check whether this quaternion algebra is isomorphic to ``A``.
+
+        Currently only implemented over a number field; motivated by
+        Main Theorem 14.6.1 in [Voi2021]_, noting that `\\QQ` has a
```
grhkm21:
```suggestion
        Main Theorem 14.6.1 in [Voi2021]_, noting that `\QQ` has a
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+            sage: B.is_isomorphic(A)
             False
         """
         if not isinstance(A, QuaternionAlgebra_ab):
             raise TypeError("A must be a quaternion algebra of the form (a,b)_K")
 
-        if self.base_ring() != QQ or A.base_ring() != QQ:
-            raise NotImplementedError("isomorphism check only implemented for rational quaternion algebras")
+        F = self.base_ring()
+        if F != A.base_ring():
```
grhkm21:
```suggestion
        if F is not A.base_ring():
```
Rings are `UniqueObject`s

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+
+        OUTPUT:
+
+        The non-Archimedean (AKA finite) places at which the quaternion
+        algebra ramifies (given as elements of `\ZZ`, sorted small to
+        large, if the algebra is defined over the rational field `\QQ`,
+        respectively as fractional ideals of the number field's ring of
+        integers, otherwise) and, if ``inf`` is set to ``True``, also the
+        Archimedean (AKA infinite) places at which the quaternion algebra
+        ramifies (given by real embeddings of the base field).
```
grhkm21: I think this can be phrased more clearly as shorter points, as currently it is a *single* long sentence. Maybe something like

The non-Archimedean (AKA finite) places at which the quaternion algebra ramifies, given as
- $\mathbb{Z}$ elements if the algebra is defined over $\mathbb{Q}$,
- Fractional ideals in the ring of integers of the number field, otherwise;

Additionally, if ``inf`` is set to ``True``, then the Archimedean (AKA infinite) places at which the quaternion algebra ramifies are returned, given by real embeddings of the base field.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1055,312 @@ def inner_product_matrix(self):
...
+    @cached_method
+    def ramified_places(self, inf=True):
+        r"""
+        Return the places of the base number field at which this
+        quaternion algebra ramifies.
+
+        Note: The initial choice of primes (for the base field `\QQ`)
+        respectively of prime ideals (in the number field case) to check
+        ramification for is motivated by 12.4.12(a) in [Voi2021]_. The
+        restriction to real embeddings is due to 14.5.8 in [Voi2021]_.
```
grhkm21:
```suggestion
        .. NOTE::
        
            The initial choice of primes (for the base field `\QQ`)
            respectively of prime ideals (in the number field case) to check
            ramification for is motivated by 12.4.12(a) in [Voi2021]_. The
            restriction to real embeddings is due to 14.5.8 in [Voi2021]_.
```

grhkm21 (changes requested):
some reviews

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Yes, it is - I implemented it independently because that's how it's also done in the code right now. Do you think I should just replace one by a call to the other?

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Is there a reason to prefer this over putting a double backslash for LaTeX commands? I thought that achieved the same result

grhkm21: I think the built documentation shows otherwise? It’s displayed as plain QQ right now.

src/sage/algebras/quatalg/quaternion_algebra.py
grhkm21: Yeah avoid code duplication

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: With the documentation built on the newest commit? It's wrong in the documentation preview because that one is still built on an old commit, I believe.

S17A05: It seems to display correctly in the new build, so I'll resolve all related suggested changes.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: There is a small caveat here, namely that these two methods are probably only opposite to each other as long as the base ring is a field - but since they were never implemented for any non-field, I'll change it as you suggested.

grhkm21: If so, don't replace the entire 4 lines with `not ...`, but only the `try`-ed case (and maybe add a comment noting this ^). This way reduces the chance the next person that implement the remaining cases makes a mistake, I guess

S17A05: In `.is_division_algebra()` I catch a `ValueError` and raise an `NotImplementedError`; but if I would try to catch a `NotImplementedError` in `.is_matrix_ring()`, this might cause errors if the `.is_division_algebra()`-method gets extended first to a case where it is not the opposite of `.is_matrix_ring()` anymore; so maybe it makes sense to leave it as is for now, since it would only replace one line anyway?

grhkm21: Yeah whatever smzall issues

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1073,363 @@ def inner_product_matrix(self):
...
+        return F.ideal(F.prod(self.ramified_places(inf=False)))
 
     def is_isomorphic(self, A) -> bool:
         """
-        Return ``True`` if (and only if) ``self`` and ``A`` are isomorphic quaternion algebras over Q.
+        Check whether this quaternion algebra is isomorphic to ``A``.
+
+        Currently only implemented for quaternion algebras defined over
+        a number field; based on Main Theorem 14.6.1 in [Voi2021]_,
+        noting that `\\QQ` has a unique infinite place.
```
mkoeppe: better to change the docstring to raw and avoid the double backslash

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1073,363 @@ def inner_product_matrix(self):
...
+
+        if F not in NumberFields():
+            raise ValueError("base field must be rational numbers or a number field")
+
+        # Since we need the list of real embeddings of the number field (instead
+        # of just the number of them), we avoid a call of the `is_totally_real()`-
+        # method by directly comparing the embedding list's length to the degree
+        E = F.real_embeddings()
+        return len(E) == F.degree() and all(F.hilbert_symbol(self._a, self._b, e) == -1
+                                                                        for e in E)
```
mkoeppe:
```suggestion
        return len(E) == F.degree() and all(F.hilbert_symbol(self._a, self._b, e) == -1
                                            for e in E)
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1067,86 +1073,363 @@ def inner_product_matrix(self):
...
-            sage: B.discriminant()
-            Fractional ideal (2)
+            sage: QuaternionAlgebra(F, 2*a, F(-1)).is_totally_definite()
+            False
 
-            sage: QuaternionAlgebra(QQ[sqrt(2)], 3, 19).discriminant()                  # needs sage.symbolic
-            Fractional ideal (1)
+        The method does not make sense over an arbitrary base ring::
+
+            sage: QuaternionAlgebra(RR(2.),1).is_totally_definite()
```
mkoeppe:
```suggestion
            sage: QuaternionAlgebra(RR(2.), 1).is_totally_definite()
```

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Alright, I'll change it - is there a reason why one is preferable over the other one, though? The developer guide gives both as valid options; maybe that preference is worth stating there.

mkoeppe: See #37508

S17A05:
The current test failures seem unrelated; also Build&Test/build seems to test all files as new, at least on merging the new beta version.

S17A05:
There's quite a few failures in the Build & Test using Conda (macos, 3.11), but as far as I can tell they are all not related to this PR.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -411,49 +412,79 @@ def is_commutative(self) -> bool:
...
+        By checking ramification, the method is able to recognize that
+        quaternion algebras (defined over a number field) with trivial
+        discriminant need not be matrix rings::
+
+            sage: K = QuadraticField(3)
+            sage: A = QuaternionAlgebra(K, -1, -1)
+            sage: A.discriminant()
+            Fractional ideal (1)
+            sage: A.is_matrix_ring()
+            False
```
AurelPage, line 481: It might be good to add an example that is split but nontrivially so, for instance $(1+\sqrt{3},3-\sqrt{3})_K$ (and similarly in ```is_division_algebra()```)?

AurelPage (approved):
Apart from a small suggestion this looks good to me!

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Thanks for the suggestion and the example! I've added it to both methods

fchapoton (approved):
looks good, thanks

AurelPage (approved):

PR #37175, opened 2024-01-27
search for a pair of pre- and post-isomorphism instead of just a post-isomorphism when computing dual isogeny

### LINKED ISSUE #37168: Taking the dual of an elliptic curve isogeny raises AssertionError
### Steps To Reproduce

We create an isogeny between elliptic curves over a finite field of 23^2 elements:
```
sage: R.<x> = GF(23)[]
sage: F.<a> = FiniteField(23^2, modulus=x^2-x+1)
sage: E0 = EllipticCurve(F, (0, 1))
sage: E1 = EllipticCurve(F, (8, 1))
sage: phi = E0.isogeny(kernel=E0((a, 0)), codomain=E1)
sage: phi.dual()
```

### Expected Behavior

The dual of `phi` should be returned.

### Actual Behavior

The following error is raised:
```
...
AssertionError: bug in dual()
```

### Additional Information

The error is not raised when `codomain=E1` is omitted.

### Environment

```markdown
- **Sage Version**: SageMath 10.3.beta6
```

### Checklist

yyyyx4:
While computing the dual of an isogeny, the current code only searches for a post-isomorphism which gives the correct scaling factor overall. In some cases, as shown by #37168, such an isomorphism may not exist, and we additionally have to compose with a suitable pre-isomorphism.

This resolves #37168.

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_curve_isogeny.py
+++ b/src/sage/schemes/elliptic_curves/ell_curve_isogeny.py
@@ -3197,6 +3197,18 @@ class EllipticCurveIsogeny(EllipticCurveHom):
             sage: phi._EllipticCurveIsogeny__clear_cached_values()  # forget the dual
             sage: -phi.dual() == (-phi).dual()
             True
+
+        Check that :trac:`37168` is fixed::
+
+            sage: R.<x> = GF(23)[]
+            sage: F.<a> = FiniteField(23^2, modulus=x^2-x+1)
+            sage: E0 = EllipticCurve(F, (0, 1))
+            sage: E1 = EllipticCurve(F, (8, 1))
+            sage: phi = E0.isogeny(kernel=E0((a, 0)), codomain=E1)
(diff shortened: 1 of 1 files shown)


src/sage/schemes/elliptic_curves/ell_curve_isogeny.py
```diff
@@ -3197,6 +3197,18 @@ def dual(self):
             sage: phi._EllipticCurveIsogeny__clear_cached_values()  # forget the dual
             sage: -phi.dual() == (-phi).dual()
             True
+
+        Check that :trac:`37168` is fixed::
```
GiacomoPope: is it worth using the :issue: command instead here?

yyyyx4: Why not!

GiacomoPope:
Looks good to me.

GiacomoPope (approved):

Issue #34286, opened 2022-08-06
Cookiecutter for Sage user projects with devcontainer


mkoeppe:
<div id="comment:0"></div>

A cross of
- https://github.com/ilyasotkov/vscode-devcontainer-python-cookiecutter
- https://github.com/mmasdeu/sage_package_template

The devcontainer config would be one of those added in #33671 (sagemath-sage or downstream-...).

Prospective sage users on Windows can use the devcontainer config to automate the Sage install in WSL. VS Code walks you through this.

Also a command `sage --cookiecutter` or `sage --new-project` could be added.

Cookiecutter details:
- https://cookiecutter.readthedocs.io/en/2.6.0/advanced/templates.html
- https://cookiecutter.readthedocs.io/en/2.6.0/advanced/directories.html - symlinks for files are OK
- https://cookiecutter.readthedocs.io/en/2.6.0/advanced/calling_from_python.html

Cookiecutter template inspiration:
- https://github.com/audreyfeldroy/cookiecutter-pypackage/tree/master/%7B%7Bcookiecutter.project_slug%7D%7D
- https://github.com/cjolowicz/cookiecutter-hypermodern-python

Alternative to cookiecutter: 
- https://cruft.github.io/cruft/ (compatible with cookiecutter, has update mechanism)

- https://marketplace.visualstudio.com/items?itemName=cantonios.project-templates

- Getting a devcontainer template for Sage into https://github.com/microsoft/vscode-dev-containers/tree/main/containers ..... this has been replaced by https://github.com/devcontainers/templates

- github's template repo https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository; perhaps combined with https://github.com/marketplace/actions/actions-template-sync

Component: **scripts**

_Issue created by migration from https://trac.sagemath.org/ticket/34286_

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -8,3 +8,4 @@
 
 Also a command `sage --cookiecutter` could be added.
 
+Alternative to cookiecutter: https://marketplace.visualstudio.com/items?itemName=cantonios.project-templates
``````

mkoeppe:
Description changed:
``````diff
--- 
+++ 
@@ -6,6 +6,6 @@
 
 Prospective sage users on Windows can use the devcontainer config to automate the Sage install in WSL. VS Code walks you through this.
 
-Also a command `sage --cookiecutter` could be added.
[...]

PR #37184, opened 2024-01-28
Improve Windows installation instructions

mkoeppe:
We streamline the installation instructions for Windows by including conda-forge instructions.

[Sage is stuck at version 9.5 in Ubuntu](https://repology.org/project/sagemath/versions), the default Linux distribution on WSL, so this should not be our primary recommendation.

Preview: https://deploy-preview-37184--sagemath.netlify.app/html/en/installation/

## Alternatives / follow-ups:
- Include a link to Microsoft Store search for Linux distros - https://www.microsoft.com/en-us/search/shop/apps?price=0&q=linux&devicetype=pc
- Include a link to Arch WSL in Microsoft Store - https://apps.microsoft.com/detail/9MZNMNKSM73X?hl=en-us&gl=US -- this is the only Linux distro in the Microsoft Store that has up-to-date SageMath packaging
- Include a link to "Import any Linux distribution to use with WSL" -- https://learn.microsoft.com/en-us/windows/wsl/use-custom-distro -- these instructions should work with any Docker image, so many options
- Create a WSL launcher app -- https://github.com/Microsoft/WSL-DistroLauncher, and document how users can "sideload" it -- https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro#sideloading-a-custom-linux-distro-package
- Create a WSL launcher app and publish it to the Microsoft Store -- https://github.com/Microsoft/WSL-DistroLauncher?tab=readme-ov-file#publishing
- #34286 
- #37534

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37588 (merged here)

### DIFF
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -146,6 +146,11 @@ Windows
        $ conda activate sage
        $ sage
 
+    This way of starting Sage gives you the most basic way of using
+    Sage in the terminal. See :ref:`sec-launching` for recommended next steps,
+    in particular for setting up the Jupyter notebook, which is required if
+    you want to use graphics.
+
 .. _installation-guide-linux:
 
 Linux
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -63,14 +63,14 @@ Windows
   - **Yes, development:**
 
     Enable `Windows Subsystem for Linux (WSL)
-    <https://learn.microsoft.com/en-us/windows/wsl/>` and install
+    <https://learn.microsoft.com/en-us/windows/wsl/>`_ and install
     Ubuntu as follows.
 
     - Make sure that hardware-assisted virtualization is enabled in
       the EFI or BIOS of your system. If in doubt, refer to your
       system's documentation for instructions on how to do this.
 
-    - `Run the command ``wsl --install`` as administrator.
+    - `Run the WSL install command as administrator.
       <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#get-started>`_
       This will install Ubuntu Linux.
 
@@ -83,7 +83,7 @@ Windows
       <https://learn.microsoft.com/en-us/windows/wsl/install#check-which-version-of-wsl-you-are-running>`_.
 
     - `Set up your Linux username and password.
-      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password>`_.
+      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password>`_
       Do not include any spaces in your username.
 
     - If your computer has less than 10GB of RAM, `change the WSL settings
@@ -96,14 +96,14 @@ Windows
   - **No development:**
 
     Enable `Windows Subsystem for Linux (WSL)
-    <https://learn.microsoft.com/en-us/windows/wsl/>` and install
+    <https://learn.microsoft.com/en-us/windows/wsl/>`_ and install
     Ubuntu as follows.
 
     - Make sure that hardware-assisted virtualization is enabled in
       the EFI or BIOS of your system. If in doubt, refer to your
       system's documentation for instructions on how to do this.
 
-    - `Run the command ``wsl --install`` as administrator.
+    - `Run the WSL install command as administrator.
       <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#get-started>`_
       This will install Ubuntu Linux.
 
@@ -116,7 +116,7 @@ Windows
       <https://learn.microsoft.com/en-us/windows/wsl/install#check-which-version-of-wsl-you-are-running>`_.
 
     - `Set up your Linux username and password.
-      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password>`_.
+      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password>`_
       Do not include any spaces in your username.
 
     - If your computer has less than 8GB of RAM, `change the WSL settings
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -80,7 +80,7 @@ Windows
     Then go to the Microsoft Store and install Ubuntu (or another
     Linux distribution). Start Ubuntu from the start menu.
 
-    In an Ubuntu (WSL) terminal, type the following commands to install
+    Type the following commands to install
     Sage from conda-forge.
 
     .. code-block:: shell
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -124,7 +124,10 @@ Windows
       to make at least 4GB of RAM available to WSL.
 
     Start Ubuntu from the Start menu, and type the following commands
-    to install Sage from conda-forge.
+    to install Sage from conda-forge. (The ``$`` represents the command
+    line prompt, don't type it!) The second step will ask a few questions,
+    and you may need to hit :kbd:`Enter` to confirm or type ``yes``
+    and then hit :kbd:`Enter`.
 
     .. code-block:: shell
 
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -16,6 +16,8 @@ SageMath release.
 
 **Where would you like to run SageMath?** Pick one of the following sections.
 
+.. _installation-guide-macos:
+
 macOS
 =====
 
@@ -51,6 +53,8 @@ macOS
     - Alternatively, build SageMath from source as described in section
       :ref:`sec-installation-from-sources`.
 
+.. _installation-guide-windows:
+
 Windows
 =======
 
@@ -61,6 +65,10 @@ Windows
     Enable `Windows Subsystem for Linux (WSL)
     <https://learn.microsoft.com/en-us/windows/wsl/>` and install
     Ubuntu as follows.
+
+    - Make sure that hardware-assisted virtualization is enabled in
+      the EFI or BIOS of your system. If in doubt, refer to your
+      system's documentation for instructions on how to do this.
 
     - `Run the command ``wsl --install`` as administrator.
       <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#get-started>`_
@@ -90,6 +98,10 @@ Windows
     Enable `Windows Subsystem for Linux (WSL)
     <https://learn.microsoft.com/en-us/windows/wsl/>` and install
     Ubuntu as follows.
+
+    - Make sure that hardware-assisted virtualization is enabled in
+      the EFI or BIOS of your system. If in doubt, refer to your
+      system's documentation for instructions on how to do this.
 
     - `Run the command ``wsl --install`` as administrator.
       <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#get-started>`_
@@ -130,6 +142,8 @@ Windows
 
        $ conda activate sage
        $ sage
+
+.. _installation-guide-linux:
 
 Linux
 =====
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -60,7 +60,7 @@ Windows
 
     Enable Windows Subsystem for Linux (WSL) by following the
     `official WSL setup guide
-    <https://docs.microsoft.com/en-us/windows/wsl/install-win10>`_. Be
+    <https://learn.microsoft.com/en-us/windows/wsl/install>`_. Be
     sure to do the steps to install WSL2 and set it as default.
     Make sure to allocate enough RAM to WSL: 5GB is known to be enough,
     2GB might not allow you to build some packages.
@@ -73,7 +73,7 @@ Windows
 
     Enable Windows Subsystem for Linux (WSL) by following the
     `official WSL setup guide
-    <https://docs.microsoft.com/en-us/windows/wsl/install-win10>`_. Be
+    <https://learn.microsoft.com/en-us/windows/wsl/install>`_. Be
     sure to do the steps to install WSL2 and set it as default.
     Make sure to allocate enough RAM to WSL: 5GB is known to be enough,
     2GB might not allow you to build some packages.
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -58,30 +58,61 @@ Windows
 
   - **Yes, development:**
 
-    Enable Windows Subsystem for Linux (WSL) by following the
-    `official WSL setup guide
-    <https://learn.microsoft.com/en-us/windows/wsl/install>`_. Be
-    sure to do the steps to install WSL2 and set it as default.
-    Make sure to allocate enough RAM to WSL: 5GB is known to be enough,
-    2GB might not allow you to build some packages.
-    Then go to the Microsoft Store and install Ubuntu (or another
-    Linux distribution). Start Ubuntu from the start menu.
+    Enable `Windows Subsystem for Linux (WSL)
+    <https://learn.microsoft.com/en-us/windows/wsl/>` and install
+    Ubuntu as follows.
 
-    Then follow the instructions for development on Linux below.
+    - `Run the command ``wsl --install`` as administrator.
+      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#get-started>`_
+      This will install Ubuntu Linux.
+
+      Note that the basic instructions in the linked article apply to
+      up-to-date installations of Windows 10 and 11, but there are
+      also links to the procedures for older builds of Windows 10.
+
+    - If you had installed WSL previously or installed it using
+      different instructions, `verify that you are running WSL 2
+      <https://learn.microsoft.com/en-us/windows/wsl/install#check-which-version-of-wsl-you-are-running>`_.
+
+    - `Set up your Linux username and password.
+      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password>`_.
+      Do not include any spaces in your username.
+
+    - If your computer has less than 10GB of RAM, `change the WSL settings
+      <https://learn.microsoft.com/en-us/windows/wsl/wsl-config#main-wsl-settings>`_
+      to make at least 5GB of RAM available to WSL.
+
+    Start Ubuntu from the Start menu. Then follow the instructions for
+    development on Linux below.
 
   - **No development:**
 
-    Enable Windows Subsystem for Linux (WSL) by following the
-    `official WSL setup guide
-    <https://learn.microsoft.com/en-us/windows/wsl/install>`_. Be
-    sure to do the steps to install WSL2 and set it as default.
-    Make sure to allocate enough RAM to WSL: 5GB is known to be enough,
-    2GB might not allow you to build some packages.
-    Then go to the Microsoft Store and install Ubuntu (or another
-    Linux distribution). Start Ubuntu from the start menu.
+    Enable `Windows Subsystem for Linux (WSL)
+    <https://learn.microsoft.com/en-us/windows/wsl/>` and install
+    Ubuntu as follows.
 
-    Type the following commands to install
-    Sage from conda-forge.
+    - `Run the command ``wsl --install`` as administrator.
+      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#get-started>`_
+      This will install Ubuntu Linux.
+
+      Note that the basic instructions in the linked article apply to
+      up-to-date installations of Windows 10 and 11, but there are
+      also links to the procedures for older builds of Windows 10.
+
+    - If you had installed WSL previously or installed it using
+      different instructions, `verify that you are running WSL 2
+      <https://learn.microsoft.com/en-us/windows/wsl/install#check-which-version-of-wsl-you-are-running>`_.
+
+    - `Set up your Linux username and password.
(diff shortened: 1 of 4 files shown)


src/doc/en/installation/index.rst
```diff
@@ -80,9 +80,25 @@ Windows
...
 
-      On the Linux running on WSL, you always have root access, so you
-      can use any of the installation methods described below for
-      Linux.
+      In an Ubuntu (WSL) terminal, type the following commands to install
+      Sage from conda-forge.
+
+      .. code-block:: shell
+
+         $ curl -L -O https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-$(uname)-$(uname -m).sh
```
tobiasdiez: Please use miniforge. Usage of mambaforge is discouraged since a few months.

src/doc/en/installation/index.rst
```diff
@@ -80,9 +80,25 @@ Windows
       Then go to the Microsoft Store and install Ubuntu (or another
       Linux distribution). Start Ubuntu from the start menu.
 
-      On the Linux running on WSL, you always have root access, so you
-      can use any of the installation methods described below for
-      Linux.
+      In an Ubuntu (WSL) terminal, type the following commands to install
```
tobiasdiez:
```suggestion
      Type the following commands to install
```

Above it just said to open the WSL terminal.

tobiasdiez (changes requested):
You said that the installation on Ubuntu is currently broken. Is this specific to WSL?

src/doc/en/installation/index.rst
mkoeppe: Reference for that please.

tobiasdiez: I don't remember where I read it. Probably in the mircoforge docs or in some issue in their github repo.

mkoeppe: In the future, when you request a change, please take the moment to back it up with a reference from the beginning. This is gratuitous friction.

The change that you request here is not necessary. https://github.com/conda-forge/miniforge?tab=readme-ov-file#should-i-choose-one-or-another-going-forward-at-the-risk-of-one-of-them-gettting-deprecated is clear: "there are no plans to deprecate Mambaforge"

tobiasdiez: Why? Your link clearly states "if you had to start using one today, we recommend to stick to Miniforge." Why should we not follow this recommendation?

mkoeppe: We are not "starting". 

The instructions are copied from our existing, tested conda-forge instructions in conda.rst

I don't have objections to changing them, but it's not going to happen in this PR. Scope matters.

mkoeppe:
> You said that the installation on Ubuntu is currently broken. Is this specific to WSL?

What I wrote in the PR description was based on watching a student try it on his computer 2 weeks. I can't test it myself. 

Just now I tested the sagemath package on the ubuntu:jammy Docker image. It installs and tests OK as of today.

In any case, the Ubuntu package is stuck at version 9.5, so for this reason alone it should not be our primary recommendation.

src/doc/en/installation/index.rst
tobiasdiez: A user following these instructions would be "starting".

tobiasdiez:
> > You said that the installation on Ubuntu is currently broken. Is this specific to WSL?
> 
> What I wrote in the PR description was based on watching a student try it on his computer 2 weeks. I can't test it myself.
> 
> Just now I tested the sagemath package on the ubuntu:jammy Docker image. It installs and tests OK as of today.
> 
> In any case, the Ubuntu package is stuck at version 9.5, so for this reason alone it should not be our primary recommendation.

Thanks for the details. But this argument (especially your last point) applies to plain ubuntu as well. So I would prefer to simply change the how to install on ubuntu section to give priority to conda, and then leave the wsl section untouched.

mkoeppe:
> change the how to install on ubuntu section to give priority to cond

We do not have such a section.

src/doc/en/installation/index.rst
mkoeppe: That's not an interesting discussion. 

If you are interested in this change, prepare a follow-up PR, change it uniformly, and test it properly.

tobiasdiez:
> > change the how to install on ubuntu section to give priority to cond
> 
> We do not have such a section.

Sorry, I meant the "linux" section.

src/doc/en/installation/index.rst
tobiasdiez: You are adding something suboptimal not me... It's not the task of the reviewer to fix issues introduced in the PR that she is reviewing.

mkoeppe:
> > > change the how to install on ubuntu section to give priority to cond
> > 
> > We do not have such a section.
> 
> Sorry, I meant the "linux" section.

That does not make much more sense either. The Linux section makes different recommendations based on the version of Sage that is found on the distribution.

The problem that I am solving here in this PR is that there are too many indirections, each of which can confuse our potential users. That's what "streamlining" means in the PR description.

tobiasdiez:
> > > > change the how to install on ubuntu section to give priority to cond
> > > 
> > > 
> > > We do not have such a section.
> > 
> > 
> > Sorry, I meant the "linux" section.
> 
> That does not make much more sense either. The Linux section makes different recommendations based on the version of Sage that is found on the distribution.
> 

The same would then also apply to Windows, as it is possible to use many different linux distros via wsl (as we also say in the docs).

mkoeppe:
This is not an interesting discussion.

kcrisman:
Okay, let me see if I can summarize:
* Windows WSL has Ubuntu by default, and the very likely scenario on Windows is that people will not be using anything other than default.
* Apparently the Ubuntu version is stuck for some reason.
* Conda-forge (in some version) works
* So we should recommend this.

It will take me a little bit to test whether this actually works.  But I do have access to a Windows computer if my son lets me on it 😄 

However, I also don't want to get in the middle of the myriad meta-discussions going on around platforms etc.  So I'm going to take the point of view that "Windows users probably need very explicit instructions, and shouldn't be asked to go to a scary Linux section" and not worry about Linux instructions, which however certainly might need updating as well.

Technical question: do we expect that once `conda activate sage` is done once, then `sage` will work (path etc.) from then on in?  I know very little about conda.

mkoeppe:
> So I'm going to take the point of view that "Windows users probably need very explicit instructions, and shouldn't be asked to go to a scary Linux section" and not worry about Linux instructions, which however certainly might need updating as well.

Yes, that's also my viewpoint here. If not for anything else, then certainly for keeping the focus of the PR as narrow as possible. After all, even with this deliberately narrow focus, it has been sitting for 3 weeks.

> Technical question: do we expect that once `conda activate sage` is done once, then `sage` will work (path etc.) from then on in? I know very little about conda.

No, `conda activate` only affects the current shell session. To persist it, one would need to add the command to a shell initialization file such as .bash_profile

mkoeppe:
I'm setting it to blocker because it needs to make its way in the stable release.

kcrisman:
I'll take a look once I get access to a Windows machine.

By the way, any sense on why the Ubuntu is stuck on such an old version?  (Probably this was discussed ad nauseam elsewhere, sorry.)

mkoeppe:
I think maintainer burnout + possibly a problem with brial (hence #36380)

kcrisman:
So this is interesting - at least on the most modern Windows, you actually don't have to install Ubuntu any more either.  It just automatically chooses Ubuntu for you.  (See [here](https://learn.microsoft.com/en-us/windows/wsl/install#install-wsl-command).)  It also should automatically use WSL 2.  So we may want to update that part (we should still remind users of older versions about those things, of course).  I'll see about installing Sage now.

kcrisman:
And then create a UNIX username ... wow, there are a lot of steps.  Anyway, that's documented in the Windows link.

kcrisman:
It works!!!  Though I can't figure out how to get plots to show up; any thoughts?  I tried [this link](https://wiki.sagemath.org/SageWindows) (which should totally be linked on the Windows install page ...) but it didn't seem to work.

Also, I'd still recommend pointing out other Linux options somehow, and streamlining the instructions to point out that most people will not need the Windows store now.  Because your mileage may vary, and we don't want monolith instructions.  Maybe for older Windows versions.  But on this Windows 11 system, it worked.

kcrisman:
Also, how fragile is this line?

    mamba create -n sage sage python=3.11

kcrisman:
> Additionally, you have to enable virtualization in BIOS. 

This is actually documented [here](https://doc.sagemath.org/html/en/installation/source.html#wsl-prerequisites) but not in the basic instructions.

kcrisman:
I'm putting needs work mostly because there are just enough tricky bits that simply saying "do this" isn't enough for a lot of users (notably the BIOS issue), as well as somewhat outdated (for most) like choosing the Ubuntu.  

Also, I'd like to know if there is a good way to avoid the specific Python version reference, or to automatically update it when necessary.

But in principle the conda piece works fine, which is presumably the main point.

mkoeppe:
@kcrisman Many thanks for testing and the detailed suggestions.

> at least on the most modern Windows, you actually don't have to install Ubuntu any more either. It just automatically chooses Ubuntu for you. (See [here](https://learn.microsoft.com/en-us/windows/wsl/install#install-wsl-command).) It also should automatically use WSL 2. So we may want to update that part (we should still remind users of older versions about those things, of course).

I've updated the steps and links to refer to these streamlined instructions, please take a look.

> And then create a UNIX username

I've added this step

mkoeppe:
> > Additionally, you have to enable virtualization in BIOS.
> 
> This is actually documented [here](https://doc.sagemath.org/html/en/installation/source.html#wsl-prerequisites) but not in the basic instructions.

I've added this now (in 23e1b42)

mkoeppe:
> Though I can't figure out how to get plots to show up; any thoughts? I tried [this link](https://wiki.sagemath.org/SageWindows) (which should totally be linked on the Windows install page ...) but it didn't seem to work.

I've just cut out the outdated Docker/WSL stuff from that wiki page and replaced it by a link the Sage installation guide.

I don't know if it makes sense to work on the plotting support instructions that are provided there, or we should just delete them. 
I'd rather guide people to getting Jupyter notebooks running.
A later installation guide section has this:
- https://deploy-livedoc--sagemath.netlify.app/html/en/installation/launching#wsl-post-installation-steps

src/doc/en/installation/index.rst
```diff
@@ -51,38 +53,97 @@ macOS
...
 
-    Enable Windows Subsystem for Linux (WSL) by following the
-    `official WSL setup guide
-    <https://docs.microsoft.com/en-us/windows/wsl/install-win10>`_. Be
-    sure to do the steps to install WSL2 and set it as default.
-    Make sure to allocate enough RAM to WSL: 5GB is known to be enough,
-    2GB might not allow you to build some packages.
-    Then go to the Microsoft Store and install Ubuntu (or another
-    Linux distribution). Start Ubuntu from the start menu.
+    Enable `Windows Subsystem for Linux (WSL)
```
kcrisman, line 65: syntax error in rendering

src/doc/en/installation/index.rst
```diff
@@ -51,38 +53,97 @@ macOS
...
+    <https://learn.microsoft.com/en-us/windows/wsl/>` and install
+    Ubuntu as follows.
+
+    - Make sure that hardware-assisted virtualization is enabled in
+      the EFI or BIOS of your system. If in doubt, refer to your
+      system's documentation for instructions on how to do this.
+
+    - `Run the command ``wsl --install`` as administrator.
```
kcrisman: Syntax error in rendering the first backtick is superfluous

kcrisman: Or maybe not, but something is wrong, anyway.

src/doc/en/installation/index.rst
```diff
@@ -51,38 +53,97 @@ macOS
...
+      up-to-date installations of Windows 10 and 11, but there are
+      also links to the procedures for older builds of Windows 10.
+
+    - If you had installed WSL previously or installed it using
+      different instructions, `verify that you are running WSL 2
+      <https://learn.microsoft.com/en-us/windows/wsl/install#check-which-version-of-wsl-you-are-running>`_.
 
-    Then follow the instructions for development on Linux below.
+    - `Set up your Linux username and password.
+      <https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password>`_.
```
kcrisman: Extra period shows up here.  This syntax for the link is correct, though.

src/doc/en/installation/index.rst
```diff
@@ -51,38 +53,97 @@ macOS
...
-    sure to do the steps to install WSL2 and set it as default.
-    Make sure to allocate enough RAM to WSL: 5GB is known to be enough,
-    2GB might not allow you to build some packages.
-    Then go to the Microsoft Store and install Ubuntu (or another
-    Linux distribution). Start Ubuntu from the start menu.
-
-    On the Linux running on WSL, you always have root access, so you
-    can use any of the installation methods described below for
-    Linux.
+    Enable `Windows Subsystem for Linux (WSL)
```
kcrisman, line 98: Same issues for this subsection, including the extra period.

kcrisman (changes requested):
I don't think I intended to use the GH review process, but anyway needs minor work for links.

kcrisman:
> I don't know if it makes sense to work on the plotting support instructions that are provided there, or we should just delete them. 

Well, ... I mean, not everything needs to live in a notebook.  If there's a way for it to work normally, we should document that.

> I'd rather guide people to getting Jupyter notebooks running. A later installation guide section has this:

Oh, that's useful to know.

----

There are some link syntax errors coming up here, [for instance in the WSL instructions](https://deploy-preview-37184--sagemath.netlify.app/html/en/installation/) - they should be pretty obvious when you look at the live copy.  Missing backticks and other things, mostly, I think.  I'll try to point a few out via comments on the individual lines, but I may miss a few.

kcrisman:
> @kcrisman Many thanks for testing and the detailed suggestions.

No problem, I'd wanted to test this process for quite some time.  It was much more effortless than anticipated, which is nice.  And thanks for all the edits.

src/doc/en/installation/index.rst
mkoeppe: Fixed now

mkoeppe:
Thanks!

roed314:
@tobiasdiez Undoing someone else's positive review with no explanation of a concrete problem is not acceptable.  I know that the sage-abuse committee has been slow to work through the huge pile of disputed tickets, but that doesn't excuse this behavior.

kcrisman:
> @tobiasdiez Undoing someone else's positive review with no explanation of a concrete problem is not acceptable. I know that the sage-abuse committee has been slow to work through the huge pile of disputed tickets, but that doesn't excuse this behavior.

I guess it must be referring to https://github.com/sagemath/sage/pull/37184#discussion_r1469041494 which refers to [this miniforge link](https://github.com/conda-forge/miniforge?tab=readme-ov-file#whats-the-difference-between-mambaforge-and-miniforge).    Far more relevant than that link is [this one](https://mamba.readthedocs.io/en/latest/installation/mamba-installation.html) from the actual Mamba website, though I still don't understand the difference and why they bothered.

I guess I see @tobiasdiez' point, but it also seems hopelessly confusing to someone who is not interested in mamba/conda/whatever anyway, which is the vast majority of our Windows users, I'd wager.  In any case, @roed314's point still stands.  At the very least, even if mini is now better/default, now somebody has to test the whole thing again.  I doubt I can use the same computer to do so - ? The install location will be different, but presumably however PATH on Windows works won't be.  (And it doesn't set path, apparently - does the mambaforge version do that?  I'm not sure how "identical" "identical" is, as a fresh grunt in this conda world.) 

-----

As a final note, it seems that miniforge has a "Windows installer".  Is that better?  Who knows?  It does seem most appropriate, pace @tobiasdiez, to have a new PR which then thoroughly *tests* the option on a Windows box (at least one!), and perhaps the installer as well, since that would be different from simply streamlining.  Howeve, pace @mkoeppe, if the suggestion to switch to miniforge had come from this reviewer (and documented and tested), I hope the suggestion would have been congenial, as it's not completely unrelated.  

[Pace everybody, Sage needs a little peace](https://github.com/sagemath/sage/pull/37534#issuecomment-1980000173), and I think this is achievable with some breaks and flexibility.  We are on the same team.

roed314:
@tobiasdiez, changing tickets from positive review is still not acceptable.

mkoeppe:
I'm setting it to "blocker" so that this improvement is considered for inclusion in 10.3.

jhpalmieri:
I am writing as a representative of the Sage Conduct Committee to record a -1 vote from Tobias.

kcrisman:
> I am writing as a representative of the Sage Conduct Committee to record a -1 vote from Tobias.

Thanks; additional clarification as to exactly how a reviewer might respond to that information will be helpful.  I'm not 100% sure how #37501 fits in with this kind of report, but it's all been in flux so I'm sure it's new to all of us.

mkoeppe:
> I'm not 100% sure how #37501 fits in with this kind of report

@kcrisman I believe recording a "-1" vote is for the process proposed in https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ, which is itself still being voted on until Mar 13.

kcrisman:
>
>
> @kcrisman <https://github.com/kcrisman> I believe recording a "-1" vote
> is for the process proposed in
> https://groups.google.com/g/sage-devel/c/IgBYUJl33SQ, which is itself
> still being voted on until Mar 13.
>
>
> Thanks, I had seen that but not realized it was implemented, and I admit I
didn't read it carefully.   I guess I'm +1 for this, because it's easy to
address the concerns in a new PR based on this.

Update: but it looks like maybe this is resolved in the most recent commit
and this discussion is moot?

mkoeppe:
I have merged #37588 (thanks @williamstein for testing that) and updated the instructions from there.

I'm setting back to "blocker" because this really should be merged in 10.3.

mkoeppe:
@kcrisman With these changes merged, do you want to test again, or can we merge?

jhpalmieri:

mkoeppe:
How can there be a vote if the PR is not "disputed" any more?

jhpalmieri:
In my opinion, if it's not disputed, "-1" votes have no clear impact, other than to let developers know that someone has some objections.

kcrisman:
> @kcrisman With these changes merged, do you want to test again, or can we merge?

Ideally someone should test again, though William's other testing is good.  I just worry about Windows being wonky.    However, I do vote +1 for this change overall, to be clear.

-----

Comments on testing: 

* It does ask a few questions in the install process for miniforge, so one might want to note that in the instructions (specifically when running the bash script), since Windows users may not be used to *text* license questions and typing "yes" 😄 
* Still no one has answered my question about whether the Python 3.11 is brittle.  How often will that have to be updated?  Is there a version-agnostic way to do that?
* Even with an existing mambaforge, it worked fine installing Sage with miniforge instead, it was "smart enough" to ask questions.  So the actual instructions are fine, and 2+2=4.
* Finally, I think the user should be notified in these instructions (again, thinking of absolute newbies) that `sage -n` will give a notebook, but they may have to copy and paste the link given in the command line.  I don't know if that belongs in one of these files (maybe after "You can now start SageMath as follows:"), but I think it's important.   (Until a "WSL launcher app" shows up, that would be cool too!)

-----

Addressing the issue implicitly raised, [this comment](https://github.com/sagemath/sage/pull/37588#issuecomment-1994048645) suggests that, while Tobias is currently voting -1 due to duplication, it might just as easily be possible to have a quick followup PR that then uses some standard rst tool to remove that duplication.  I don't really care if it happens here or there, but it does seem helpful in terms of preventing doc rot.

mkoeppe:
> * no one has answered my question about whether the Python 3.11 is brittle.  How often will that have to be updated?  Is there a version-agnostic way to do that?

Sorry for missing this. 
That it is written there explicitly -- instead of referring the reader to go elsewhere to look for this information -- is to serve the reader.

Keeping the documentation source a simple .rst file (instead of generating it somehow from some template file so that we could refer to a variable) -- this is to keep the structure of our documentation sources as simple and straightforward as possible -- this serves the contributors to the documentation.

The cost of this (= the cost of updating this occurrence) is low, as the update happens only about once a year, and the places to change are easy to find, in particular for those who know how to make the updates of the supported Python versions.

mkoeppe:
> Tobias is currently voting -1 due to duplication

FWIW, I recommend not to take such rationalizations at face value.

kcrisman:
>
>
> Sorry for missing this.
>
No problem.

> The cost of this (= the cost of updating this occurrence) is low, as the
> update happens only about once a year, and the places to change are easy to
> find, in particular for those who know how to make the updates of the
> supported Python versions.Message ID:
> ***@***.***>
>
Okay, I guess we do have to search for that every so often ...

kcrisman:
>
> Tobias is currently voting -1 due to duplication
>
> FWIW, I recommend not to take such rationalizations at face value.
>
Thankfully, I have the margin to do so, though I understand why others
might not.

mkoeppe:
> * I think the user should be notified in these instructions (again, thinking of absolute newbies) that `sage -n` will give a notebook, but they may have to copy and paste the link given in the command line.  I don't know if that belongs in one of these files (maybe after "You can now start SageMath as follows:"), but I think it's important.

Done in 4eacb7be249, please take a look.

kcrisman:
Thanks for addressing all these concerns.  While I still hope for an nth iteration of a double click installer, this is fairly painless as such things go in our Windows ecosystem experience.

mkoeppe:
@kcrisman Thanks for the review!

mkoeppe:
@vbraun Please merge. Can't set to blocker because policy.

PR #37186, opened 2024-01-29
Update debian.txt for various Python packages

### LINKED ISSUE #35259: in ubuntu 22.04 these instructions don't work: "sudo apt-get install  sagemath sagemath-doc-en sagemath-jupyter"
When I found this page

https://doc.sagemath.org/html/en/reference/spkg/_sagemath.html

I immediately thought... "I bet these don't work".  The only one I wanted to try was Ubuntu with the latest version, and indeed it didn't work.  The instructions are 
```
sudo apt-get install  sagemath sagemath-doc-en sagemath-jupyter
```
but "sagemath-doc-en" has to be replaced by "sagemath-doc".  

I wonder if the top of the page should just say "The package names may change slightly or be deprecated and we do not regularly test them, so please pay attention to the output of your package manager."

I also noticed that the sentence at the bottom seem to be missing a period. Also, the trac link is broken, since it is just `https://trac.sagemath.org/27330` instead of `https://trac.sagemath.org/ticket/27330`.  Was that a standard redirect with trac (to not have /ticket" or was it a mistake in this particular link?  I have no idea, but if it is common, then I can add new redirects [here](https://github.com/sagemath/trac.sagemath.org). 

image

mkoeppe:
As reported by @EmmanuelCharpentier in https://groups.google.com/g/sage-support/c/21iZLkesyHk

Also fixes #35259

### :memo: Checklist

### DIFF
--- a/build/bin/sage-spkg-info
+++ b/build/bin/sage-spkg-info
@@ -155,9 +155,17 @@ else
             echo "If the system package is installed, $(code ./configure) will check if it can be used."
         fi
     else
-        echo "However, these system packages will not be used for building Sage"
-        echo "because $(code spkg-configure.m4) has not been written for this package;"
-        echo "see $(issue 27330)"
+        case $PKG_BASE in
+            _*)
+                # Suppress the message, as it makes no sense for these special packages
+                # (_bootstrap, _sagemath)
+                ;;
+            *)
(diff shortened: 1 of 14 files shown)


dimpase (approved):
OK

mkoeppe:
Thanks!

PR #37189, opened 2024-01-29
Constructor quaternion algebra over number fields from ramification

### LINKED ISSUE #16948: Construct a quaternion algebra over a number field by specifying its ramification
<div id="comment:0"></div>

There is a need for the constructor QuaternionAlgebra to take as input a number field F and a ramification (specified by an ideal D in F and a set of real places S of F) and which constructs a quaternion algebra over F with ramified precisely at the primes dividing D and at the places in S.

We have code to add here, which is now being tested and documented.

CC:  @tornaria

Component: **number theory**

Keywords: **quaternion algebra, ramification**

Author: **Marc Masdeu, Xavier Guitart**

Branch/Commit: **[u/mmasdeu/16948-review](https://github.com/sagemath/sagetrac-mirror/tree/u/mmasdeu/16948-review) @ [`fcf1a69`](https://github.com/sagemath/sagetrac-mirror/commit/fcf1a69ad070ee4528138ef95f1d4d22190aa4ae)**

_Issue created by migration from https://trac.sagemath.org/ticket/16948_

Eloitor:
This PR adds a new input format for the constructor of quaternion algebras.

It uses PARI `alginit` function to construct one form the ramification information.

Fixes #16948

- [X] I have updated the documentation accordingly.


S17A05:
I'll review this once the necessary checks have completed.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -178,6 +188,22 @@ class QuaternionAlgebraFactory(UniqueFactory):
...
+        sage: Q = K.prime_above(3)
+        sage: A = QuaternionAlgebra(K, [P,Q], [0,0])
+        sage: A.discriminant()
+        Fractional ideal (6)
+        sage: A = QuaternionAlgebra(K, [P,Q], [1/2,0])
+        Traceback (most recent call last):
+        ...
+        PariError: domain error in checkhasse: sum(Hasse invariants) != 0
+        sage: QuaternionAlgebra(QQ.number_field(), [ZZ.ideal(2),ZZ.ideal(3)], [0])
+        Quaternion Algebra (-1, 3) with base ring Rational Field
```
S17A05: You should catch this PariError and let the user know more clearly that there is no quaternion algebra with the given ramification data (sorry, this is only about lines +201 to +203).

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +283,28 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
             b = K(v[1])
 
-        # QuaternionAlgebra(K, a, b)
         else:
+            # QuaternionAlgebra(K, finite_places, archimedian_places)
             K = arg0
-            a = K(arg1)
-            b = K(arg2)
+            if isinstance(arg1, list) and isinstance(arg2, list):
+                if K == QQ:
```
S17A05: I believe this should be `if is_RationalField(K):` instead (again messed up on the lines, this is only about +290).

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +283,28 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                    D = ZZ.ideal_monoid().prod(arg1).gen()
+                    a, b = hilbert_conductor_inverse(D)
+                    a = Rational(a)
+                    b = Rational(b)
+                else:
+                    if len(arg2) != len(K.real_places()):
```
S17A05: Since this "else"-part only works for NumberFields, you should check/catch this properly here and give an appropriate error warning (only for +295 to +296).

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +283,28 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+            # QuaternionAlgebra(K, finite_places, archimedian_places)
             K = arg0
-            a = K(arg1)
-            b = K(arg2)
+            if isinstance(arg1, list) and isinstance(arg2, list):
+                if K == QQ:
+                    D = ZZ.ideal_monoid().prod(arg1).gen()
+                    a, b = hilbert_conductor_inverse(D)
+                    a = Rational(a)
+                    b = Rational(b)
```
S17A05: This code for QQ doesn't check whether the user actually gave valid input since it doesnt check the second argument, e.g. it doesn't differentiate between the valid input (QQ, [(2), (3)], [0]) and the invalid input (QQ, [(2), (3)], [1/2]).

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +283,28 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                else:
+                    if len(arg2) != len(K.real_places()):
+                        raise ValueError("must specify ramification at the real places of %s" % K)
+                    K_pari = pari(K)
+                    fin_places_pari = [I.pari_prime() for I in arg1]
+                    A = K_pari.alginit([2, [fin_places_pari, [QQ(1/2)] * len(fin_places_pari)], arg2])
```
S17A05: Before passing stuff to pari, you should check that the input is of the correct form; `K` already is a number field here according to the above comment, but you should still check the elements of `arg1` and `arg2`, at least by catching errors appropriately. Also it is probably a bit nicer to not define `K_pari` intermediately here, but to instead directly call `A = pari(K).alginit(...)`.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +283,28 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                    L_pari = A.algsplittingfield()
+                    a = K(L_pari.disc()[1])
```
S17A05: As above, you don't really need to define L_pari intermediately here.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -102,6 +106,12 @@ class QuaternionAlgebraFactory(UniqueFactory):
       `D` over `K = \QQ`.  Suitable nonzero rational numbers `a`, `b`
       as above are deduced from `D`.
 
+    - ``QuaternionAlgebra(K, L_finite, L_archimedean)``, where `L_finite`
+      is a list of prime ideals and `L_archimedean` is a list local
```
S17A05: "L_archimedean is a list _of_ local ..."

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -102,6 +106,12 @@ class QuaternionAlgebraFactory(UniqueFactory):
...
+      invariants (0 or 1/2) specifying the ramification at the real places.
+      This constructs a quaternion algebra ramified exacly at the
+      specified places.
```
S17A05: I would add "at the _(infinite)_ real places _of K_." at the end of the first sentence to be more explicit. Also it wouldn't hurt to make the description of which places will be ramified (exactly those in L_finite plus those in K.real_embeddings() indexed by i with L_archimedean[i]=1/2, if I understood it correctly) a bit more explicit here.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -102,6 +106,12 @@ class QuaternionAlgebraFactory(UniqueFactory):
       `D` over `K = \QQ`.  Suitable nonzero rational numbers `a`, `b`
       as above are deduced from `D`.
 
+    - ``QuaternionAlgebra(K, L_finite, L_archimedean)``, where `L_finite`
```
S17A05: The names "L_finite" and "L_archimedean" don't give the best idea of what input is expected; names such as e.g. "primes"/"places_fin" (or something similar) and "inv_archimedean" would be a bit clearer.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Sorry, this should only be about lines +201 to +203.

S17A05: Oh, and in line +204 you don't really need the `.number_field()` since it just returns QQ (as the Rational Field) again (see #7596).

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: I think it would be nice to redesign it so that the one element list with the correct invariant at infinity as well as the empty list (signaling the constructor to solve for the invariant at infinity on its own) are precisely the valid second arguments; the latter especially makes it nicer to call if the number of primes is not known a priori.
Also a check/error catch for the content of arg1 should be included here as well.

Eloitor: > Also a check/error catch for the content of arg1 should be included here as well.

I now check that the elements of `arg1` are primes. Should I also check that they are unique?

S17A05: That's a good question, I'd say there are arguments both for and against forcing uniqueness. I would not force it, but that means you should make sure to not take one prime factor twice (and similarly for the number field case; Pari throws an appropriate error, but since all Pari errors are of the same type `PariError` you probably want to enforce this in advance to still correctly catch the invariant error), e.g. by converting the list of primes into a set beforehand.

S17A05:
Let me know (e.g. by commenting here) once you think this ready for another full review; in the meantime, Lint recognized some whitespaces in the blank line 302.

Eloitor:
ready for review

S17A05:
> ready for review

Great, I'll review it again when I find the time for it; this will probably not be before Friday though.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +289,37 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                    a, b = hilbert_conductor_inverse(D)
+                    a = QQ(a)
+                    b = QQ(b)
+                else:
+                    if len(arg2) != len(K.real_places()):
+                        raise ValueError("must specify ramification at the real places of %s" % K)
+                    if is_odd(len(arg1) + 2 * sum(arg2)):
+                        raise ValueError("quaternion algebra over the rationals must have an even number of ramified places")
+                    fin_places_pari = [I.pari_prime() for I in arg1]
+                    A = pari(K).alginit([2, [fin_places_pari, [QQ(1/2)] * len(fin_places_pari)], arg2])
```
AurelPage: Hi, I suggest using alginit with flag=0 (4th argument), to prevent computation of a maximal order. Indeed, you are throwing away the information anyway, so you may as well not compute it.

Eloitor: Thank you,

In the future (when orders for quaternion algebras over number fields are more developed), maybe we could offer an optional flag `compute_maximal_order=False`, and store generators of the order if the user wants to work with a maximal order after generating the algebra.

AurelPage:
Did you check that the real embeddings are treated in the same order as in .real_embeddings()? This does not seem automatic as the real embeddings are computed from the roots of the defining polynomial, not from PARI, but you are simply forwarding the second argument to PARI's alginit. Maybe this should be checked, documented, and tested (by checking the signs of the embeddings of a,b).

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +289,37 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
             b = K(v[1])
 
-        # QuaternionAlgebra(K, a, b)
         else:
+            # QuaternionAlgebra(K, primes, inv_archimedean)
             K = arg0
-            a = K(arg1)
-            b = K(arg2)
+            if category(K) is not NumberFields():
+                raise ValueError("quaternion algbera must be defined over a number field")
```
S17A05: The check `category(QQ) is not NumberFields()` returns `True`; you should instead place this check after handling the case of the rational field.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +289,37 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+            if isinstance(arg1, list) and isinstance(arg2, list):
+                if not set(arg2).issubset(set([0,QQ(1/2)])):
+                    raise ValueError("list of local invariants specifying ramification should contain only 0 and 1/2")
+                arg1 = list(set(arg1))
```
S17A05: I get that this is to remove repetitions of prime ideals, but is there a reason why you are converting `arg1` into a list again afterwards?

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +289,37 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                        raise ValueError("quaternion algebra over the rationals must have an even number of ramified places")
+                    D = ZZ.ideal_monoid().prod(arg1).gen()
+                    a, b = hilbert_conductor_inverse(D)
+                    a = QQ(a)
+                    b = QQ(b)
+                else:
+                    if len(arg2) != len(K.real_places()):
+                        raise ValueError("must specify ramification at the real places of %s" % K)
+                    if is_odd(len(arg1) + 2 * sum(arg2)):
+                        raise ValueError("quaternion algebra over the rationals must have an even number of ramified places")
```
S17A05: Here it should be "over a number field" instead of "over the rationals".

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -257,11 +289,37 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                    fin_places_pari = [I.pari_prime() for I in arg1]
+                    A = pari(K).alginit([2, [fin_places_pari, [QQ(1/2)] * len(fin_places_pari)], arg2], maxord=0)
+                    a = K(A.algsplittingfield().disc()[1])
+                    b = K(A.algb())
```
S17A05: Unfortunately there is a major issue here that @AurelPage correctly detected: The order in which the real embeddings are computed in PARI is indeed not the same as the order of embeddings given by `.real_embeddings()`.

Here is an example:
_sage: K.\<v\> = NumberField(-3*x^5 - 11*x^4 - 4*x^3 + 1)
sage: A = QuaternionAlgebra(K, [], [1/2, 0, 1/2])
Quaternion Algebra (-3*v^4 - 14*v^3 - 15*v^2 - 4*v, -8) with base ring Number Field in v with defining polynomial -3*x^5 - 11*x^4 - 4*x^3 + 1_

Using the `ramified_places` code from #37173, one can check that this algebra ramifies at the two embeddings sending $v$ to $-0.738229069310693$ and to $0.467355922387407$. In `K.real_embeddings()` these are the second and the third embeddings, respectively, but the desired output should be ramified at the third embedding as well as the first embedding sending $v$ to $-3.25393276038648$.

This could be fixed by first precomputing the order in which PARI lists the embeddings, and then permuting the list of local Archimedean invariants accordingly. I am not sure if this is feasible, and maybe @AurelPage has a better idea on how to work around this discrepancy between Sage and PARI.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -178,6 +190,26 @@ class QuaternionAlgebraFactory(UniqueFactory):
...
+        sage: K.<w> = NumberField(x^2-x-1)
+        sage: P = K.prime_above(2)
+        sage: Q = K.prime_above(3)
+        sage: A = QuaternionAlgebra(K, [P,Q], [0,0])
+        sage: A.discriminant()
+        Fractional ideal (6)
+        sage: A = QuaternionAlgebra(K, [P,Q], [1/2,0])
+        Traceback (most recent call last):
+        ...
+        ValueError: Quaternion algebra over the rationals must have an even number of ramified places
```
S17A05: The error warning in line 211 needs to be adapted, see the comment on line 314. Also, once #37173 has been reviewed and merged it would be nice to adapt the examples given here to show that the constructor gives an algebra with the desired ramification.

S17A05 (changes requested):
Sorry for the delay. For the future, please don't force push big changes and instead put them into a new commit, otherwise it makes the review process a lot harder because everything has to be re-checked again.

S17A05:
The modifications of the constructor also seem to mess up some examples according to the Build & Test failure.

src/sage/algebras/quatalg/quaternion_algebra.py
AurelPage: @S17A05 It is not difficult to find the bijection. Consider a generator of the field $K$, and take an $\varepsilon$ that separates its real embeddings. Consider the sequences of real embeddings computed by sage and pari respectively, at precision smaller than $\varepsilon$. Sort them both using indirect sort (with the permutations) and compose the permutations in the right way.

However, this procedure should probably not be hard-coded in the middle of the quaternion algebras code, but should be put in a place that ensures it can be reused. Maybe it is already somewhere? I don't know where that should be.

src/sage/algebras/quatalg/quaternion_algebra.py
Eloitor:
```
sage: category(QQ)
Join of Category of number fields and Category of quotient fields and Category of metric spaces
```

Is it possible to use this to check if K is QQ or a number field?

S17A05: It might be possible, but even then I would advise against it: You need to handle the rational field separately anyways (I ran into the same issue in #37173), and up to the check `is_RationalField(K)` I don't see any code that doesn't work for either setting; hence it is probably easier to place the above category check in the else block after the treatment of the rational field case.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: > @S17A05 It is not difficult to find the bijection. Consider a generator of the field K, and take an ε that separates its real embeddings. Consider the sequences of real embeddings computed by sage and pari respectively, at precision smaller than ε. Sort them both using indirect sort (with the permutations) and compose the permutations in the right way.
> 
> However, this procedure should probably not be hard-coded in the middle of the quaternion algebras code, but should be put in a place that ensures it can be reused. Maybe it is already somewhere? I don't know where that should be.

I don't think this is implemented yet. I'm not sure if this is the most elegant solution, but I would suggest adding an optional boolean parameter (e.g. called `pari_perm`, set to `False` by default) to the `.real_embeddings()`-method of a number field such that, if the parameter is set to True, the method returns both the real embeddings and a permutation of the indices that correctly re-indexes the embeddings to be in the order they are computed in in PARI.

S17A05:
The modifications of the constructor also seem to mess up some examples according to the details of the Build & Test / build failure, and on certain inputs there seems to be a `SignalError: Segmentation fault` thrown by `A.algsplittingfield().disc()`; we can look at these things once the code gives the correct mathematical result, though.

S17A05:
@Eloitor I have found a fix for the communication between Sage and PARI - the main problem is that PARI optimizes the defining polynomial, which mixes the embeddings up a bit. By mimicking this optimization in Sage, one can easily compute the permutation there with a simple sorting, though. If you want to, I can implement the solution if you give me the necessary access to your branch.

Eloitor:
This is great news!! Please go ahead. You should have access now :)

S17A05:
Should be fixed; the new test examples aren't deterministic though (it can return different, but necessarily isomorphic algebras), so I've made sure that they are not actually tested right now - I'll make them into proper tests once #37173 has been merged.

S17A05:
The errors in the tests seemed to be outdated, so I updated the branch to re-trigger the tests

S17A05:
> The errors in the tests seemed to be outdated, so I updated the branch to re-trigger the tests

Whoops, turns out I was looking at the wrong thing - the new commit should fix the errors.

S17A05:
The timeout in `src/sage/rings/tests.py` seems unrelated, but should probably be investigated either way.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -263,8 +324,62 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+        elif isinstance(arg1, list) and isinstance(arg2, list):
+            # QuaternionAlgebra(K, primes, inv_archimedean)
+            K = arg0
+            if K not in NumberFields():
+                raise ValueError("quaternion algebra must be defined over a number field")
+            if not set(arg2).issubset(set([0, QQ(1/2)])):
+                raise ValueError("list of local invariants specifying ramification should contain only 0 and 1/2")
+            primes = set(arg1)
+            if not all([p.is_prime() for p in primes]):
+                raise ValueError("quaternion algebra constructor requires a list of primes specifying the ramification")
```
fchapoton, line 380: remove [ and ] inside of all : avoid creating the list inside

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -263,8 +324,62 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                vals_embed = [sigma(from_K(v)) for sigma in L.embeddings(AA)]
+                perm = Word(vals_embed).standard_permutation()
+
+                # Transfer the primes to PARI and permute the local invariants
+                fin_places_pari = [I.pari_prime() for I in primes]
+                inv_arch_pari = [arg2[i-1] for i in perm]
+
+                # Compute the correct quaternion algebra over L in PARI
+                A = L.__pari__().alginit([2, [fin_places_pari, [QQ(1/2)] * len(fin_places_pari)],
+                                          inv_arch_pari], maxord=0)
```
fchapoton: replace QQ(1/2)  by QQ((1, 2)) to avoid going through floats

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -263,8 +324,62 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
             a = K(v[0])
             b = K(v[1])
 
-        # QuaternionAlgebra(K, a, b)
+        elif isinstance(arg1, list) and isinstance(arg2, list):
+            # QuaternionAlgebra(K, primes, inv_archimedean)
+            K = arg0
+            if K not in NumberFields():
+                raise ValueError("quaternion algebra must be defined over a number field")
```
fchapoton: add doctests for every new error message in the TESTS:: section

S17A05:
Since #37173 should be on its way to get merged, I am putting this PR to `s: needs work` for a moment - once the former PR is merged, I'll add the missing tests for error messages and will modify the examples to properly check that the construction works correctly.

S17A05:
I've added the new tests/examples, including the use of `.ramified_places()` to check that the construction works correctly. I've also made some small changes to the input checks to make sure that they reject bad inputs properly. Hopefully everything is covered now :)

fchapoton:
failing doctest

S17A05:
> failing doctest

Whoops, should be fixed now

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -264,8 +361,73 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+
+            else:
+                # Construct the quaternion algebra via ramification over a number field
+                if len(arg2) != len(K.real_places()):
+                    raise ValueError("must specify ramification at all real places of the number field")
+                if is_odd(len(primes) + 2 * sum(arg2)):
+                    raise ValueError("quaternion algebra over a number field must have an even number of ramified places")
+
+                # We want to compute the correct quaternion algebra over K with PARI
+                # As PARI optimizes the polynomial used to define the number field, we need to precompute
```
AurelPage: About the comment: PARI does not optimise the polynomial (by default, it does with `flag=2`), but it forces a defining polynomial that is monic and integral. If it is not, the polynomial gets changed.

AurelPage (approved):
Apart from a tiny nitpick, LGTM.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Thanks for the explanation! I adjusted the comment to be more accurate

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -264,8 +361,74 @@ def create_key(self, arg0, arg1=None, arg2=None, names='i,j,k'):
...
+                vals_embed = [sigma(from_K(v)) for sigma in L.embeddings(AA)]
+                perm = Word(vals_embed).standard_permutation()
+
+                # Transfer the primes to PARI and permute the local invariants
+                fin_places_pari = [I.pari_prime() for I in primes]
+                inv_arch_pari = [arg2[i-1] for i in perm]
+
+                # Compute the correct quaternion algebra over L in PARI
+                A = L.__pari__().alginit([2, [fin_places_pari, [QQ((1,2))] * len(fin_places_pari)],
+                                          inv_arch_pari], maxord=0)
```
AurelPage: I suspect that the new fails are due to a rename of the flag `maxord` to `flag` in pari's `alginit` which must have made its way into cypari2.

S17A05: Thanks for pointing this out! Hopefully everything should work now

PR #37196, opened 2024-01-30
Generalized monomial orders for Laurent polynomials

vilanele:
Adding an implementation of generalized order defined by a conic decomposition and a compatible total order on $\mathbb{Z}^n$. This will be useful to implement the computation of Gröbner basis of ideals over Laurent polynomials.

### :memo: Checklist


src/sage/rings/polynomial/generalized_monomial_order.py
```diff
@@ -0,0 +1,443 @@
...
+        g2 = vector(self.generator(i,L2))
+        m = vector([max(a,b) for a,b in zip(lm1 + g1, lm2 + g2)])
+        return tuple(cone_matrix*m)
+
+def build_cones(n):
+    r"""
+    Return a list of integer matrices such that the colums of the `j`-th matrix in the list
+    are the generators of the `j`-th cone.
+    
+    INPUTS:
```
xcaruso:
```suggestion
    INPUT:
```

src/sage/rings/polynomial/generalized_monomial_order.py
```diff
@@ -0,0 +1,443 @@
...
+
+            sage: GeneralizedMonomialOrder(3, score_function="min")
+            Generalized monomial order in 3 variables using (lex, min)
+
+            sage: GeneralizedMonomialOrder(3, group_order="lex", score_function="degmin")
+            Generalized monomial order in 3 variables using (lex, degmin)
+        """
+        self._n = n
+        self._n_cones = self._n + 1
+        self._cones = build_cones(self._n)
```
xcaruso: It is really useful to have this helper function `build_cones`?

Same remark for `get_group_order` and `get_score_function`.

src/sage/rings/polynomial/generalized_monomial_order.py
```diff
@@ -0,0 +1,443 @@
...
+            ]
+        """
+        return self._cones
+
+    def cone(self,i):
+        r"""
+        Return the matrix whose columns are the generators of the `i`-th cone.
+    
+        INPUTS:
+            - `i` -- a cone index
```
xcaruso:
```suggestion

        INPUT:
        
        - ``i`` -- an integer, the cone index
```

src/sage/rings/polynomial/generalized_monomial_order.py
```diff
@@ -0,0 +1,443 @@
...
+        diff = self._score_function(a) - self._score_function(b)
+        if diff != 0: 
+            return 1 if diff > 0 else -1
+        else: 
+            return 1 if self._group_order.greater_tuple(a,b) == a else -1
+
+    def greatest_tuple(self,a,b=None):
+        r"""
+        Return the greatest of the two tuples ``a`` and ``b`` if ``b`` is not None
+        Return the greatest tuple in the list ``L`` if ``b`` is not a tuple
```
xcaruso:
```suggestion
        Return the greatest tuple in the list ``a`` if ``b`` is not a tuple
```

src/sage/rings/polynomial/generalized_monomial_order.py
```diff
@@ -0,0 +1,443 @@
...
+        if not isinstance(a, tuple) or not isinstance(b, tuple):
+            raise ValueError("Args must be tuples")
+        if a == b: return 0
+        diff = self._score_function(a) - self._score_function(b)
+        if diff != 0: 
+            return 1 if diff > 0 else -1
+        else: 
+            return 1 if self._group_order.greater_tuple(a,b) == a else -1
+
+    def greatest_tuple(self,a,b=None):
```
xcaruso: Why not using the prototype
```
    def greatest_tuple(self, *L):
```

src/sage/rings/polynomial/generalized_monomial_order.py
```diff
@@ -0,0 +1,443 @@
...
+from sage.rings.polynomial.term_order import TermOrder
+from sage.matrix.constructor import matrix
+from sage.structure.sage_object import SageObject
+
+class GeneralizedMonomialOrder(SageObject):
+    def __init__(self,n,group_order="lex",score_function="min"):
+        r"""
+        Create a generalized monomial order.
+
+        INPUTS:
```
xcaruso: I think we should write `INPUTS` without `S` (even if there are multiple inputs).
Moreover, the list of inputs should not be indented.

xcaruso (changes requested):

src/sage/rings/polynomial/generalized_monomial_order.py
vilanele: Corrected

src/sage/rings/polynomial/generalized_monomial_order.py
vilanele: Moved to __init__.

src/sage/rings/polynomial/generalized_monomial_order.py
vilanele: corrected

src/sage/rings/polynomial/generalized_monomial_order.py
vilanele: Ok

src/sage/rings/polynomial/generalized_monomial_order.py
vilanele: Done.

src/sage/rings/polynomial/generalized_order.py
```diff
@@ -0,0 +1,428 @@
...
+Generalized monomial orders on `\ZZ^n` are total orders on `\ZZ` introduced in
+[PU1999]_ and defined by a triple:
+
+- Conic decomposition: a finite covering of the additive monoïd `\ZZ^n` by submonoids.
+  In the current implementation, the conic decomposition of Example 2.2 in 
+  [PU1999]_ is imposed.
+
+- Group order: this can be any class with a 'greatest_tuple' method returning
+  the greatest of two tuples based on a group order on `\ZZ^n`. 
+  Available choices include 'lex'.
```
xcaruso, line 11: It's not really the case.
However, I think that accepting any such class sounds like a good idea.

src/sage/rings/polynomial/generalized_order.py
```diff
@@ -0,0 +1,428 @@
...
+        self._cones = [matrix.identity(ZZ,n)]
+        for i in range(0,n):
+            mat = matrix.identity(ZZ,n)
+            mat.set_column(i,vector(ZZ,[-1]*n))
+            self._cones.append(mat)
+        # Set group order. Add more here.
+        if group_order in ["lex"]:
+            self._group_order = TermOrder(name=group_order)
+        else:
+            raise ValueError("Available group order are: 'lex'")
```
xcaruso, line 97: As I said above, I would add a test like `elif isinstance(group_order, TermOrder):` here.

src/sage/rings/polynomial/generalized_order.py
```diff
@@ -0,0 +1,428 @@
...
+        # Set score function. Add more here.
+        if score_function == "min":
+            self._score_function = min_score_function
+        elif score_function == "degmin":
+            self._score_function = degmin_score_function
+        else:
+            raise ValueError("Available score function are: 'min', 'degmin'")
```
xcaruso, line 104: Again, why not giving the option to the user to pass in any score function ?

src/sage/rings/polynomial/generalized_order.py
```diff
@@ -0,0 +1,428 @@
...
+        # Store names
+        self._group_order_name = group_order
+        self._score_function_name = score_function
+
+    def _repr_(self):
+        r"""
```
xcaruso, line 110: I thnk you should add a description here, something like `Return a  string representation of this order`.
Add you should also add a black line after `TESTS::`.

src/sage/rings/polynomial/generalized_order.py
```diff
@@ -0,0 +1,428 @@
...
+        """
+        n = len(L) 
+        if n == 0:
+           raise ValueError("empty list of tuples")
+
+        if n == 1:
+            return L[0]
+        else:
+            a = L[0]
+            b = self.greatest_tuple(*L[1:])
```
xcaruso, line 284: I don't really like this recursive call because it implies all useless checkings.
Why not programming this imperatively with a for loop?

src/sage/rings/polynomial/generalized_order.py
```diff
@@ -0,0 +1,428 @@
...
+        g2 = vector(self.generator(i,L2))
+        m = vector([max(a,b) for a,b in zip(lm1 + g1, lm2 + g2)])
+        return tuple(cone_matrix*m)
+   
+# Score functions. Add more here and update the __init__ method.
+def min_score_function(t):
+        return -min(0,*t)
+
+def degmin_score_function(t):
+        return sum(t) -(len(t)+1)*min(0,*t)
```
xcaruso, line 425: These two functions should be documented.
You also have a problem with identation.

xcaruso (changes requested):

PR #37197, opened 2024-01-30
Kummer line class for short Weierstrass and Montgomery elliptic curves

nicolassarkis:
This introduce a Kummer line class for elliptic curves in short Weierstrass and Montgomery form, as well as classes to deal with points and isogeny classes.
Also known as x-only arithmetic, Kummer lines are generally used rather than the elliptic curve itself in cryptography as they provide faster arithmetic since there is one less coordinate to deal with.
This is also the natural codomain of the isogeny between elliptic products using the theta model, as the theta model in genus two is the Kummer surface and so evaluation of points under this map only give points on elliptic products on the Kummer line.

```sage
sage: E = EllipticCurve(GF(101), [2, 3])
sage: KummerLine(E)
Kummer Line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
```

This is based on code written by @GiacomoPope and was worked on during I-sage-ny Days in Leuven with @ozbayelif

#sd123

### DIFF
--- a/src/sage/schemes/elliptic_curves/kummer_line.py
+++ b/src/sage/schemes/elliptic_curves/kummer_line.py
@@ -4,7 +4,7 @@ Kummer Line for short Weierstrass elliptic curves
 Sage defines Kummer lines derived from an elliptic curve in short Weierstrass or Montgomery form, known as `x`-only arithmetic.
 Also defines a point on a Kummer line as an element of the projective line as well as isogenies between Kummer lines.
 
-EXAMPLES:
+EXAMPLES::
 
 We construct a Kummer line over an already defined elliptic curve in short Weierstrass form::
 
@@ -17,15 +17,18 @@ We can also give the curve coefficients directly with a base ring::
     sage: KummerLine(QQ, [4, 5/6])
     Kummer line of the elliptic curve y^2 = x^3 + 4*x + 5/6 over Rational Field
 
-TODO KummerPoint examples, as well as many more usage (see integer for instance)
-TODO <Lots and lots of examples>
-
 AUTHORS:
 
 - Giacomo Pope (2023): original code
 
 - Elif Özbay Gürler, Nicolas Sarkis (2024-01): ported code to short Weierstrass and documentation
 """
+
+#
+# ::
+#
+# TODO KummerPoint examples, as well as many more usage (see integer for instance)
+# TODO <Lots and lots of examples>
 
 # ****************************************************************************
 #       Copyright (C) 2023 Giacomo Pope <giacomopope@gmail.com>
@@ -51,6 +54,7 @@ class KummerLine:
     Kummer line class for a short Weierstrass elliptic curve.
 
     EXAMPLES::
+
         sage: E = EllipticCurve(GF(101), [2, 3])
         sage: KummerLine(E)
         Kummer line of the elliptic curve y^2 = x^3 + 2*x + 3 over Finite Field of size 101
@@ -211,7 +215,7 @@ class KummerLine:
 
     def curve(self):
         r"""
-        Returnn the associated elliptic curve.
+        Return the associated elliptic curve.
 
         EXAMPLES::
 
@@ -248,7 +252,7 @@ class KummerLine:
         r"""
         Compute the j-invariant of the Kummer line using the formula
 
-        ..MATH::
+        .. MATH::
 
             j(E) = 1728\frac{4a^3}{4a^3 + 27b^2}.
 
@@ -277,7 +281,7 @@ class KummerLine:
         r"""
         Compute the discriminant of the Kummer line using the formula
 
-        ..MATH::
+        .. MATH::
 
             \Delta(E) = -16(4a^3 + 27b^2).
 
@@ -322,6 +326,7 @@ class KummerLine:
     #     A6 = a6 - 7 * w
     #     K2 = KummerLine(self.base_ring(), [A4, A6])
     #     return K2, K2(alpha)
+    # TODO isogeny call instantiate a KummerLineIsogeny object
 
 
 # class KummerPoint:
--- a/src/sage/schemes/elliptic_curves/kummer_line.py
+++ b/src/sage/schemes/elliptic_curves/kummer_line.py
@@ -95,20 +95,23 @@ from sage.structure.sage_object import SageObject
 
 def KummerLine(curve):
     r"""
-    Wrapper to construct a Kummer line depending on the shape of the curve.
+    Wrapper to construct a Kummer line from an elliptic curve depending on its shape.
+
+    INPUT: ``curve`` -- an elliptic curve in short Weierstrass or Montgomery form
 
     EXAMPLES::
-
-    Can deal with both short Weierstrass and Montgomery curves::
 
         sage: E = EllipticCurve(GF(101), [2, 3])
         sage: K = KummerLine(E); type(K)
         <class 'sage.schemes.elliptic_curves.kummer_line.KummerLine_weierstrass'>
+
+    This can also deal with Montgomery curves::
+
         sage: E = EllipticCurve(GF(101), [0, 3, 0, 1, 0])
         sage: K = KummerLine(E); type(K)
         <class 'sage.schemes.elliptic_curves.kummer_line.KummerLine_montgomery'>
 
-    Raise an error otherwise::
+    Raise an error if not short Weierstrass nor Montgomery::
 
         sage: E = EllipticCurve(GF(101), [1, 1, 2, 3, 5])
         sage: K = KummerLine(E)
--- a/src/sage/schemes/elliptic_curves/kummer_line.py
+++ b/src/sage/schemes/elliptic_curves/kummer_line.py
@@ -24,7 +24,7 @@ It is then possible to define points on this line, derived from points on the cu
     sage: E = EllipticCurve(GF(101), [2, 3])
     sage: K = KummerLine(E)
     sage: P = E(95, 49); xP = K(P)
-    sage: xQ = K([73, 1]) 
+    sage: xQ = K([73, 1])
 
 Several methods helps to recover useful data, such as `XZ`-coordinates or the affine `x`-coordinate::
 
@@ -83,7 +83,7 @@ AUTHORS:
 #                  https://www.gnu.org/licenses/
 # ****************************************************************************
 
-from sage.all import cached_method
+from sage.misc.cachefunc import cached_method
 
 from sage.rings.integer import Integer
 from sage.schemes.elliptic_curves.ell_generic import EllipticCurve_generic
@@ -168,13 +168,11 @@ class KummerLine_generic(SageObject):
 
         EXAMPLES::
 
-        Base ring must be the same::
-
             sage: E1 = EllipticCurve(GF(101), [2, 3])
             sage: E2 = EllipticCurve(QQ, [2, 3])
             sage: K1 = KummerLine(E1)
             sage: K2 = KummerLine(E2)
-            sage: K1 == K2
+            sage: K1 == K2 # Base ring must be the same
             False
 
         Kummer lines are not equals even if the curves are isomorphic, which
--- a/src/sage/schemes/elliptic_curves/kummer_line.py
+++ b/src/sage/schemes/elliptic_curves/kummer_line.py
@@ -12,10 +12,56 @@ We construct a Kummer line over an already defined elliptic curve in short Weier
     sage: KummerLine(E)
     Kummer line of Elliptic curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
 
-We can also give the curve coefficients directly with a base ring::
+If the curve is Montgomery shaped, there is another class to handle it::
 
-    sage: KummerLine(QQ, [4, 5/6])
-    Kummer line of Elliptic curve defined by y^2 = x^3 + 4*x + 5/6 over Rational Field
+    sage: E = EllipticCurve(GF(101), [0, 3, 0, 1, 0])
+    sage: KummerLineMontgomery(E)
+    Kummer line of Elliptic curve defined by y^2 = x^3 + 3*x^2 + x over Finite Field of size 101
+
+It is then possible to define points on this line, derived from points on the curve
+(or by giving it the coordinates)::
+
+    sage: E = EllipticCurve(GF(101), [2, 3])
+    sage: K = KummerLine(E)
+    sage: P = E(95, 49); xP = K(P)
+    sage: xQ = K([73, 1]) 
+
+Several methods helps to recover useful data, such as `XZ`-coordinates or the affine `x`-coordinate::
+
+    sage: xP.XZ()
+    (95, 1)
+    sage: xQ.x()
+    73
+
+Differential addition and scalar multiplication are implemented::
+
+    sage: xPQ = K(P-Q)
+    sage: xQP = xP.add(xQ, xPQ); xQP.x() == (P+Q)[0]
+    True
+    sage: m = 293
+    sage: (m * xP).x() == (m * xP)[0]
+    True
+
+Isogenies between Kummer lines are also available using Vélu formulas::
+
+    sage: xf = K.isogeny(xP); xf
+    Isogeny of degree 48 from Kummer line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101 to Kummer line of Elliptic Curve defined by y^2 = x^3 + 68*x + 29 over Finite Field of size 101
+
+Methods similar to the elliptic curves ones are also implemented::
+
+    sage: xf.degree()
+    48
+    sage: xf.domain()
+    Kummer line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
+    sage: xf.codomain()
+    Kummer line of Elliptic Curve defined by y^2 = x^3 + 68*x + 29 over Finite Field of size 101
+
+They can also be evaluated::
+
+    sage: xf(xQ)
+    (51 : 1)
+    sage: xf(5*xP)
+    (1 : 0)
 
 AUTHORS:
 
@@ -27,7 +73,7 @@ AUTHORS:
 
 # ****************************************************************************
 #       Copyright (C) 2023 Giacomo Pope <giacomopope@gmail.com>
-#       Copyright (C) 2024 Elif Özbay Gürler <TODO>
+#       Copyright (C) 2024 Elif Özbay Gürler <eozbayelif@gmail.com>
 #       Copyright (C) 2024 Nicolas Sarkis <nicolas.sarkis@math.u-bordeaux.fr>
 #
 # This program is free software: you can redistribute it and/or modify
@@ -36,13 +82,6 @@ AUTHORS:
 # (at your option) any later version.
 #                  https://www.gnu.org/licenses/
 # ****************************************************************************
-
-#
-# ::
-#
-# TODO KummerLinePoint examples, as well as many more usage (see integer for instance)
-# TODO <Lots and lots of examples>
-
 
 from sage.all import cached_method
 
@@ -292,7 +331,7 @@ class KummerLine(SageObject):
             sage: P = E(76, 65); xP = K(P)
             sage: Q = E(49, 61); xQ = K(Q)
             sage: KummerLineIsogeny(xP)
-            TODO
+            Isogeny of degree 12 from Kummer line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101 to Kummer line of Elliptic Curve defined by y^2 = x^3 + 23*x + 73 over Finite Field of size 101
         """
 
         if xP._parent != self:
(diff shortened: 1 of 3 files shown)


GiacomoPope:
I don't have time for a proper review at the moment, but a few small things which I think would be good is to break this file into a few smaller ones. For example, elliptic curves are in `ell_generic.py` and then for the case of curves over finite fields the code is in `ell_finite_field.py`, for the above code most of the code looks like a finite field is assumed? Do you have a use case for this code over QQ or number fields?

There's also the question of whether this should inherit something `ProjectiveSpace` or if not this then I think at least allowing the inheritence of `UniqueRepresentation` and `Parent` would help. Similarly, the points could inherit `Element`. Looking at elliptic curves, they inherit:

```py
class EllipticCurve_generic(WithEqualityById, plane_curve.ProjectivePlaneCurve):
```

Something to think about and discuss here I think, as getting this right is important for new objects like this.

nicolassarkis:
You're right, the code is pretty long and I can break it down into several files. As for QQ and number fields, I don't personally know any use case and I did not find any while looking on the Internet (although I did not search very thoroughly), but I don't expect the code to break. I'll try to give more examples over other fields rather than just finite fields.

`ProjectiveSpace` seems to be the correct inheritance, I'll try to look into that too as well as the points. As for the isogenies, I don't know which class would fit best?

GiacomoPope:
I think if you want to restrict to finite fields too, this is fine, but then this should be considered in the code and error handling. 

For isogenies, look at `EllipticCurveHom` for inspiration, which I believe inherits `Morphism`.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+    if not isinstance(curve, EllipticCurve_generic):
+        raise TypeError(
+            "input must be an elliptic curve in short Weierstrass or Montgomery form"
+        )
+
+    ainvs = curve.a_invariants()
+    a, b = ainvs[3], ainvs[4]  # Short Weierstrass
+    AM = ainvs[1]  # Montgomery
+
+    if ainvs == (0, 0, 0, a, b):
```
grhkm21, line 129: Should this be a method of `EllipticCurve`?

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            Kummer Line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
+        """
+        self._curve = curve
+        self._base_ring = curve.base_ring()
+
+    def __eq__(self, other):
+        r"""
+        Test equality of Kummer lines by checking if the above curves are the same.
+
+        EXAMPLES::
```
grhkm21, line 169: There should be a doctest showing when the method succeeds.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            sage: E = EllipticCurve(GF(101), [2, 3])
+            sage: K = KummerLine(E); K
+            Kummer Line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
+        """
+        self._curve = curve
+        self._base_ring = curve.base_ring()
+
+    def __eq__(self, other):
+        r"""
+        Test equality of Kummer lines by checking if the above curves are the same.
```
grhkm21, line 167:
```suggestion
        Test equality of Kummer lines by checking if the associated curves are the same.
```
Not sure if "associated" is correct, but I think "above" is not.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        EXAMPLES::
+
+            sage: E1 = EllipticCurve(GF(101), [2, 3])
+            sage: E2 = EllipticCurve(QQ, [2, 3])
+            sage: K1 = KummerLine(E1)
+            sage: K2 = KummerLine(E2)
+            sage: K1 == K2 # Base ring must be the same
+            False
+
+        Kummer lines are not equals even if the curves are isomorphic, which
```
grhkm21, line 178:
```suggestion
        Kummer lines are not equal even if the curves are isomorphic, which
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            'Kummer Line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101'
+        """
+        return f"Kummer Line of {self._curve}"
+
+    def __call__(self, coords):
+        r"""
+        Create a Kummer point from the coordinates.
+
+        INPUT: ``coords`` -- either a point `P` on the elliptic curve or
+        a valid tuple `(X, Z)` where `P = (X : * : Z)`; `Z` is optional
```
grhkm21, line 209:
```suggestion
        INPUT: ``coords`` -- either a point `P` on the elliptic curve, an `X`
        coordinate, or a valid tuple `(X, Z)` representing `P = (X : * : Z)`
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+        OUTPUT: A Kummer point on the Kummer line
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve(GF(101), [2, 3])
+            sage: P = E(95, 49)
+            sage: K = KummerLine(E)
+
+        Point as a parameter::
```
grhkm21, line 219:
```suggestion
        We can pass in an Elliptic curve point as a parameter::
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+            sage: xP = K(P); xP
+            (95 : 1)
+
+        The output is the same if we just give the `XZ`-coordinates::
+
+            sage: K([95, 1]) == K(P)
+            True
+
+        `Z` is optional::
```
grhkm21, line 229:
```suggestion
        `Z` is set to `1` if not provided::
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
+        """
+        return self._curve
+
+    @cached_method
+    def j_invariant(self):
+        r"""
+        Compute the j-invariant of the associated elliptic curve.
+
+        Cached method.
```
grhkm21, line 279:
```suggestion
        The result of this method is cached.
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        sage: E = EllipticCurve(GF(101), [2, 3])
+        sage: K = KummerLine(E); K
+        Kummer Line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
+    """
+
+    def __init__(self, curve):
+        r"""
+        Constructor for a Kummer line from a short Weierstrass elliptic curve.
+
+        INPUT: ``curve`` -- an elliptic curve in short Weierstrass form `y^2 = x^3 + a*x + b`
```
grhkm21, line 344:
```suggestion
        INPUT: ``curve`` -- an elliptic curve in short Weierstrass form `y^2 = x^3 + ax + b`
```
Single quotations means the content inside is $\LaTeX$.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            sage: K = KummerLine(E); K.extract_weierstrass_constants()
+            (2, 3)
+        """
+        return self._a, self._b
+
+    def xDBL(self, XP, ZP):
+        r"""
+        Doubling formula for `x`-only coordinate on short Weierstrass curve.
+
+        Cost: ``2M + 5S + 3m0 + 8a``
```
grhkm21, line 386: I know this is standard notation, but you might want to document the cost notation somewhere. Also, `xDBL`, `xADD`, `xDBLADD` aren't good method names, as they are not descriptive.

Also, why is `XP` and `ZP` provided separately? Shouldn't this be a point object or a tuple? Or am I confusing something?

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+        - ``XQ``, ``ZQ`` -- the coordinates of the Kummer point `x(Q)`
+
+        - ``XPQ``, ``ZPQ`` -- the coordinates of the Kummer point `x(P-Q)`
+
+        OUTPUT: ``(XQP, ZQP)``, the coordinates of `x(P+Q)`
+
+        .. NOTE::
+
+            The formula does not behave well when `x(P-Q) = (1 : 0)` (should use `xDBL` instead)
```
grhkm21, line 446:
```suggestion
            The formula does not behave well when `x(P-Q) = (1 : 0)`;
            the method :meth:`xDBL` should be used instead.
```

grhkm21: Same for several methods below.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            sage: E = EllipticCurve(GF(101), [2, 3])
+            sage: K = KummerLine(E)
+            sage: P, Q = E(48, 55), E(66, 65)
+            sage: xP, xQ, xPQ = K(P), K(Q), K(P-Q)
+            sage: XQP, ZQP = K.xADD(*xP.XZ(), *xQ.XZ(), *xPQ.XZ())
+            sage: XQP/ZQP
+            11
+            sage: P+Q
+            (11 : 89 : 1)
+
```
grhkm21, line 459:
```suggestion
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        t1 = t1 - t2
+        XQP = ZPQ * t1
+        t1 = C - D
+        t1 = t1 * t1
+        ZQP = XPQ * t1
+
+        return X2P, Z2P, XQP, ZQP
+
+
+class KummerLine_montgomery(KummerLine_generic):
```
grhkm21, line 625: Most of the reviews I gave for `KummerLine_weierstrass` docstrings above apply.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+    def __init__(self, parent, coords):
+        r"""
+        Create a point from the coordinates.
+
+        INPUT:
+
+            - ``parent`` - A Kummer line
+
+            - ``coords`` - either a point P on EllipticCurve or a tuple (X, Z) where `P = (X : * : Z)`; Z is optional
```
grhkm21, line 913:
```suggestion
            - ``coords`` -- either a point `P` on the elliptic curve, an `X`
            coordinate, or a valid tuple `(X, Z)` representing `P = (X : * : Z)`
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        (95 : 1)
+    """
+
+    def __init__(self, parent, coords):
+        r"""
+        Create a point from the coordinates.
+
+        INPUT:
+
+            - ``parent`` - A Kummer line
```
grhkm21, line 911:
```suggestion
            - ``parent`` -- a Kummer line
```
Double dash for INPUT (special syntax for docs)

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        """
+        # Ensure the parent is the right type
+        if not isinstance(parent, KummerLine_generic):
+            raise TypeError("not a Kummer line")
+
+        R = parent.base_ring()
+
+        # Point at infinity
+        if coords is None:
+            coords = (R(1), R(0))
```
grhkm21, line 948: You should use `R.one(), R.zero()` instead, it is faster as it does not go through the coercion model.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        """
+        return self._AM
+
+    def extract_weierstrass_constants(self):
+        r"""
+        Return the short Weierstrass coefficients `(a, b)` as a tuple.
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve(GF(101), [2, 3])
```
grhkm21, line 681: The docs do not test the correct thing.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+                coords = (R(1), R(0))
+            else:
+                a, b = parent.extract_weierstrass_constants()
+                # assert coords.curve().a_invariants() == (0, 0, 0, a, b)
+                # Done this way to handle Montgomery curves
+                assert coords.curve().is_isomorphic(EllipticCurve(R, [a, b]))
+                coords = coords[0], coords[2]
+
+        # Construct from X coordinate only
+        elif isinstance(coords, RingElement):
```
grhkm21, line 963: I think this is better handled if you do a `try` cast into `R`, otherwise say `K(int(1))` will fail. I know this is stupid but I write code in `.py` a lot :(

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        if self.parent() != other.parent():
+            return False
+
+        XP, ZP = self.XZ()
+        XQ, ZQ = other.XZ()
+        return XP * ZQ == XQ * ZP
+
+    def is_zero(self):
+        r"""
+        A Kummer Point is zero if it corresponds to the point at infinity
```
grhkm21, line 1066:
```suggestion
        Return whether this Kummer Point is zero.
        
        A Kummer Point is zero if it corresponds to the point at infinity
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+        OUTPUT: ``False`` if this is the point at infinity, otherwise ``True``
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve(GF(101), [2, 3])
+            sage: P = E(95, 49)
+            sage: K = KummerLine(E)
+            sage: xO = K(None); xO.is_zero()
+            False
```
grhkm21, line 1077: This should be `True`.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        XP, ZP = self.XZ()
+        XQ, ZQ = other.XZ()
+        return XP * ZQ == XQ * ZP
+
+    def is_zero(self):
+        r"""
+        A Kummer Point is zero if it corresponds to the point at infinity
+        on the parent curve.
+
+        OUTPUT: ``False`` if this is the point at infinity, otherwise ``True``
```
grhkm21, line 1069: The description is flipped around.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            sage: K = KummerLine(E)
+            sage: xO = K(None); xO.is_zero()
+            False
+            sage: xP = K(P); xP.is_zero()
+            False
+        """
+
+        return not bool(self)
+
+    def is_x_coord(self):
```
grhkm21, line 1084: Is my understanding correct that this will always return `True` on a valid Kummer Point? If so, this should either be a hidden method (`_is_x_coord`), or there should be a method/constructor that possibly returns an invalid point. For example, `EllipticCurve` has `.point(..., check=False)`.

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+            sage: P = E(95, 49)
+            sage: K = KummerLine(E)
+            sage: xP = K(P); xP.parent()
+            Kummer Line of Elliptic Curve defined by y^2 = x^3 + 2*x + 3 over Finite Field of size 101
+        """
+        return self._parent
+
+    def XZ(self):
+        r"""
+        Return the projective coordinates `(X : Z)` of the point.
```
grhkm21, line 1137:
```suggestion
        Return the projective coordinates `(X, Z)` of the point.
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        X, Z = self.XZ()
+
+        if Z == 1:
+            return self._base_ring(X)
+        return self._base_ring(X / Z)
+
+    @cached_method
+    def curve_point(self):
+        r"""
+        Return a lift of the point on the parent curve.
```
grhkm21, line 1181:
```suggestion
        Return a lift of this Kummer Point onto the parent curve.
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+        B = b - 7 * w
+
+        E2 = EllipticCurve(R, [A, B])
+        K2 = KummerLine(E2)
+
+        return cst_t, cst_u, K2
+
+    def evaluate(self, xP):
+        r"""
+        Evaluate the isogeny at ``xP``.
```
grhkm21, line 1734:
```suggestion
        Evaluate this isogeny at the point given.
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+        EXAMPLES::
+
+            sage: E = EllipticCurve(GF(101), [2, 3]); K = KummerLine(E)
+            sage: P = E(76, 65); xP = K(P)
+            sage: Q = E(49, 61); xQ = K(Q)
+            sage: f = E.isogeny(P)
+            sage: xf = K.isogeny(xP) # Isogeny on K with kernel xP
+            sage: f.codomain() == xf.codomain().curve()
```
grhkm21, line 1743:
```suggestion
            sage: f.codomain() is xf.codomain().curve()
```

src/sage/schemes/elliptic_curves/kummer_line.py
```diff
@@ -0,0 +1,1765 @@
...
+
+It is then possible to define points on this line, derived from points on the curve
+(or by giving it the coordinates)::
+
+    sage: E = EllipticCurve(GF(101), [2, 3])
+    sage: K = KummerLine(E)
+    sage: P = E(95, 49); xP = K(P)
+    sage: xQ = K([73, 1])
+
+Several methods helps to recover useful data, such as `XZ`-coordinates or the affine `x`-coordinate::
```
grhkm21, line 29:
```suggestion
Several methods help to recover useful data, such as `XZ`-coordinates or the affine `x`-coordinate::
```

grhkm21 (changes requested):
I gave some suggestions on the mainly docstrings.

It would also be nice to have a `random_element` and `random_point` method. Note that for those you can't lift `EllipticCurve` directly because that won't be uniform :P (though not a strict requirement anyways).

grhkm21:
Also as Giacomo suggested, it's a good idea to split this into different files. I wonder if this should even live in `schemes/elliptic_curves/*`, or a separate folder.

The code seems to work for number fields, and indeed I don't see why it won't.

```python
sage: E = EllipticCurve(QQ, [2, 3])
....: K = KummerLine(E)
....: P = E(3, 6)
....: K(P) * 10
(22530025406263552494005407484901630647681408671284722145966135457356237300849501743063056857 : 10874884192915767915534771901123061205122377170148511038589335491516561977252134297645806864)
sage: P*10
(531479353783914619632845849151571137189110665429162541878284457/256536613656855871747631938823851386468295946496228420559682064 : -16453886036139809456500625176595598039719127218803897114798879065614104455635946684154276282987/4108885474335556202158650899673783883396708141861736713956421844956502466300564568902035815488 : 1)
```

src/sage/schemes/elliptic_curves/kummer_line.py
yyyyx4: (An elliptic curve lies "above" its Kummer line in the sense that it double-covers it; the choice of word is not uncommon.)

grhkm21: Ahh I misinterpreted as the line above (`self` and `other`) 🥲 my bad

nicolassarkis:
Thank you very much for the suggestions, I'll try to go back to this soon enough!

PR #37201, opened 2024-01-30
Fix univariate polynomial reverse

vneiger:
For univariate polynomials over "non-prime" finite fields, the reverse method had an issue when specifying `degree=0`: it raised an error saying that this optional argument should be... nonnegative (which 0 is).

Example:
```
sage: ring.<x> = GF(9)[]
sage: pol = ring(1)
sage: pol.reverse(degree=2)
x^2
sage: pol.reverse()
1
sage: pol.reverse(degree=0)
Cell In[5], line 1
----> 1 pol.reverse(degree=Integer(0))

File ~/repositories/software/sage/src/sage/rings/polynomial/polynomial_zz_pex.pyx:537, in sage.rings.polynomial.polynomial_zz_pex.Polynomial_ZZ_pEX.reverse()
    535 if degree is not None:
    536     if degree <= 0:
--> 537         raise ValueError("degree argument must be a non-negative integer, got %s" % (degree))
    538     try:
    539         d = degree

ValueError: degree argument must be a non-negative integer, got 0
```

This is fixed in this PR, along with fixes in similar code where the error message was not the expected one ("cannot convert to long", instead of "degree must be nonnegative").

### DIFF
--- a/src/sage/matrix/matrix_polynomial_dense.pyx
+++ b/src/sage/matrix/matrix_polynomial_dense.pyx
@@ -675,7 +675,7 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
             sage: M.reverse([2,3,-1])
             Traceback (most recent call last):
             ...
-            OverflowError: can't convert negative value to unsigned long
+            ValueError: degree argument must be a non-negative integer, got -1
 
         .. SEEALSO::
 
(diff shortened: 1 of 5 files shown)


xcaruso:
Looks good to me.

xcaruso (approved):

vneiger:
Minor fix: the change of the error message (upon negative input value) affects a doctest in matrix_polynomial_dense, this is fixed.

PR #37207, opened 2024-01-31
allow algorithm=pari in LLL

AurelPage:
We allow to call the pari implementation of LLL reduction. The next release of pari will contain an implementation of the FLATTER algorithm for LLL reduction, which has better asymptotic complexity.

### :memo: Checklist


src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3077,14 +3082,40 @@ cdef class Matrix_integer_dense(Matrix_dense):
...
+        Example with a nonzero kernel and rank caching::
+
+            sage: M = matrix(4,3,[1,2,3,2,4,6,7,0,1,-1,-2,-3])
+            sage: M.LLL()[0:2]
+            [0 0 0]
+            [0 0 0]
+            sage: M.rank()
+            2
+
+            sage: M = matrix(4,3,[1,2,3,2,4,6,7,0,1,-1,-2,-3])
```
seblabbe: This repeat the line above: remove.

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3077,14 +3082,40 @@ cdef class Matrix_integer_dense(Matrix_dense):
...
+            sage: M.LLL(algorithm="NTL:LLL")[0:2]
+            [0 0 0]
+            [0 0 0]
+            sage: M.rank()
+            2
+
+            sage: M = matrix(4,3,[1,2,3,2,4,6,7,0,1,-1,-2,-3])
```
seblabbe: This repeat the line above: remove

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3077,14 +3082,40 @@ cdef class Matrix_integer_dense(Matrix_dense):
...
+            sage: U * A == R
+            True
+
+        Example with a nonzero kernel and rank caching::
+
+            sage: M = matrix(4,3,[1,2,3,2,4,6,7,0,1,-1,-2,-3])
+            sage: M.LLL()[0:2]
+            [0 0 0]
+            [0 0 0]
+            sage: M.rank()
```
seblabbe: I would remove the `M.rank()` line from this example.

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3077,14 +3082,40 @@ cdef class Matrix_integer_dense(Matrix_dense):
...
+            2
+
+            sage: M = matrix(4,3,[1,2,3,2,4,6,7,0,1,-1,-2,-3])
+            sage: M.LLL(algorithm="NTL:LLL")[0:2]
+            [0 0 0]
+            [0 0 0]
+            sage: M.rank()
```
seblabbe: I suggest to add a doctest block in the `TESTS::` part which test that the rank is computed correctly by each algo with proper `M._clear_cache()` calls in between.

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3250,6 +3281,18 @@ cdef class Matrix_integer_dense(Matrix_dense):
             R = A.to_matrix(self.new_matrix())
             if transformation:
                 U = U.to_matrix(self.new_matrix(U.nrows, U.ncols))
+
+        elif algorithm == 'pari':
+            # call pari with flag=4: kernel+image
+            # pari uses column convention: need to transpose the matrices
+            from sage.libs.pari import pari
```
seblabbe: if we use `self._gp_().mattranspose()` below, then, we don't need to import pari on this line

Also we would need to use `K.sage().augment(T.sage())` or something like this.

I don't have a strong opinion on this. We may keep it as it is if it is better overall.

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3250,6 +3281,18 @@ cdef class Matrix_integer_dense(Matrix_dense):
...
+            A = pari(self).mattranspose()
+            K, T = A.qflll(4)
+            U = pari.matconcat([K, T]).mattranspose().sage()
+            R = U * self
+            r = ZZ(T.length())
```
seblabbe, line 3301: This is a pari method call done after lines where the sage language is performed. Maybe keep pari computations all together is more readable.

src/sage/matrix/matrix_integer_dense.pyx
AurelPage: I'm not sure about this one. Maybe someone else has an opinion?

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3115,6 +3138,17 @@ cdef class Matrix_integer_dense(Matrix_dense):
             sage: A = random_matrix(ZZ, 0, 0)
             sage: R, U = A.LLL(transformation=True)
 
+            Test rank caching:
```
seblabbe:
```suggestion
        Test rank caching:
```

src/sage/matrix/matrix_integer_dense.pyx
fchapoton: **Never** use `_gp_`, which is going through the old and slow interface ; use `__pari__` 

```
sage: m=matrix(ZZ,4,4,range(16))
sage: type(m)
<class 'sage.matrix.matrix_integer_dense.Matrix_integer_dense'>
sage: m._gp_()
[0, 1, 2, 3; 4, 5, 6, 7; 8, 9, 10, 11; 12, 13, 14, 15]
sage: type(_)
<class 'sage.interfaces.gp.GpElement'>
sage: m.__pari__()
[0, 1, 2, 3; 4, 5, 6, 7; 8, 9, 10, 11; 12, 13, 14, 15]
sage: type(_)
<class 'cypari2.gen.Gen'>

```

AurelPage: Ok, thanks! Would it be better to use __pari__ to avoid the import statement?

yyyyx4: I don't really see a reason to avoid the import statement: It's essentially free since the library has already been loaded anyway.
```
sage: %timeit from sage.libs.pari import pari
182 ns ± 1.79 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

seblabbe (approved):
Looks good to me.

src/sage/matrix/matrix_integer_dense.pyx
dimpase: Shouldn't this be using `cypari2`, not `pari` ?

dimpase:
@videlec

src/sage/matrix/matrix_integer_dense.pyx
videlec: > Shouldn't this be using `cypari2`, not `pari` ?

What do you mean?
```python
>>> from sage.libs.pari import pari
>>> type(pari)
<class 'cypari2.pari_instance.Pari'>
```

dimpase: oops, never mind, sorry

videlec: Though, for the conversion, it is probably faster to go through `M.__pari__()` or even the more direct
```
cdef Gen pari_matrix = integer_matrix(self._matrix, 0)
```
(which is the one-line code of `M.__pari__()`). The second argument of `integer_matrix` allows you to transpose which might be useful here!

videlec: To convert back from pari the most direct at the moment is to call `gen_to_sage_matrix` from `sage.libs.pari.convert_sage_matrix`. Though this function does not take into account the fact that the entries are integral and there is a bunch of irrelevant python objects built for each entry. For huge matrices it could matter.

videlec: I will propose a commit.

videlec:
Hmm... caching in LLL is not very smart
```python
sage: m = matrix(ZZ, 2, 2, [-1,1,1,1])
sage: m.LLL(algorithm="NTL:LLL")
[-1  1]
[ 1  1]
sage: m.det()
2
sage: m = matrix(ZZ, 2, 2, [-1,1,1,1])
sage: m.det()
-2
```

videlec:
I can not push to your branch. github is really bad with collaborations. I added an extra commit to my fork of sage: https://github.com/videlec/sage/tree/pari-LLL

AurelPage:
@videlec Thanks! I am trying your commit. Git made decentralised collaboration easy, and then github ruined it.

AurelPage:
I've included your commit.

dimpase:
> @AurelPage 
> I can not push to your branch. github is really bad with collaborations. I added an extra commit to my fork of sage: https://github.com/videlec/sage/tree/pari-LLL


I think this "cannot push to your branch" only happens if the PR's author disallowed such pushes while creating the PR (not sure if this can be changed later), or if the contributor who wants add commits is not a member of Sagemath Triage team.

In any event it should always be possible to make a PR against the fork of this PR on the PR's author repo.

I agree that's confusing, and could be documented better.

AurelPage:
@dimpase I allowed edits by maintainers, if that's what you mean. And it can be changed after the PR has been created.

dimpase:
perhaps @videlec is not a member of Triage team? (I can't check this myself)

videlec:
Indeed, push are allowed for "maintainers" but not anybody (like me :-)

dimpase:
pushes are not allowed for just anybody, for a good reason - to prevent vandalism from persons unaffiliated with the project.

In this case "maintainer" means being on an appropriate Sagemath (a GitHub organisation) team, with appropriate push access. I thought that members of Triage have such rights - to push to the branches related to the PRs (not to protected  branches) - but I might be wrong here. 

Maybe @vbraun knows better.

In any case, asking the author of the PR to manually merge a commit looks like a suboptimal thing, going via a PR to the fork looks more natural.

PR #37210, opened 2024-01-31
unify alias substitute for subs

mantepse:
We unify the definitions as follows:

* `substitute` is a method in `Element` which calls `self.subs` and passes all arguments along, and not defined anywhere else.
* `subs` is the method that should be overwritten in the various element classes.

Dependencies:

#37143


fchapoton:
I guess this "needs review", right ?

mantepse:
Yes !

fchapoton (approved):
good enough, thanks a lot, danke sehr

PR #37215, opened 2024-02-01
Remove unnecessary bullet point

GiacomoPope:
For the installation guide there was an erroneous bullet point which I have removed.

### DIFF
--- a/src/doc/en/installation/index.rst
+++ b/src/doc/en/installation/index.rst
@@ -71,18 +71,18 @@ Windows
 
   - **No development:**
 
-    - Enable Windows Subsystem for Linux (WSL) by following the
-      `official WSL setup guide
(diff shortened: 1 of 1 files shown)


GiacomoPope:
Arguably, one could "refactor" the section so that

> Enable Windows Subsystem for Linux (WSL) by following the [official WSL setup guide](https://docs.microsoft.com/en-us/windows/wsl/install-win10). Be sure to do the steps to install WSL2 and set it as default. Make sure to allocate enough RAM to WSL: 5GB is known to be enough, 2GB might not allow you to build some packages. Then go to the Microsoft Store and install Ubuntu (or another Linux distribution). Start Ubuntu from the start menu.

Is is before the "Yes, development" and then have smaller pieces for the remaining info as this is repeated, but this would make the section look unlike the others.

kwankyu (approved):
OK.

PR #37218, opened 2024-02-01
Fix to ncsym.ncsym.nesting

lucasgagnon:
The nesting function in combinat.ncsym.ncsym.py does not operate as described in its documentation.  As intended, it should compare all arcs of two set partitions and count the number of times the second arc nests in the first.  As written, there was a break statement caused many (necessary) comparisons to be skipped, particularly for set partitions with large blocks.  

I have re-written the function so that all necessary comparisons are made.  I have also added an example in the docstring that was computed incorrectly by the old code, but is corrected by my fix.  

This bug is not a previously known issue, and I have also confirmed with the original author (@tscrim ) that no one else is currently working on a fix.

### :memo: Checklist

### :hourglass: Dependencies

No dependencies


lucasgagnon:
@tscrim can you review this?

tscrim (approved):
Thank you. LGTM.

PR #37033, opened 2024-01-08
a lazy solver for functional equations

mantepse:
We enable sage to solve systems of functional equations of lazy power series, generalizing the existing possibility of defining series recursively.  For example:
```
sage: L.<z> = LazyPowerSeriesRing(QQ)
sage: A = L.undefined()
sage: B = L.undefined()
sage: FA = A^2 + B^2 - 2 - z*B
sage: FB = B^3 + 2*A^3 - 3 - z*(A + B)
sage: L.define_implicitly([(A, [1]), (B, [1])], [FA, FB])
sage: A
1 + 1/6*z - 7/72*z^2 - 125/1296*z^3 - 3289/31104*z^4 - 23591/186624*z^5 - 1048607/6718464*z^6 + O(z^7)
sage: B
1 + 1/3*z + 7/36*z^2 + 47/324*z^3 + 1903/15552*z^4 + 2959/23328*z^5 + 500663/3359232*z^6 + O(z^7)
```

Dependencies: #37451, #37687,  #37736, #37766, #37761, #37916

PR #37220, opened 2024-02-01
provide a construction functor

mantepse:
We enable sage to find common parents also when symmetric functions are involved.  In particular, we provide a functorial construction that takes a commutative (coefficient) ring and produces a (commutative) ring of symmetric functions over this ring.

For Macdonald polynomials and other symmetric functions that involve parameters we cheat a bit: the categories should actually be commutative rings with distinguished elements (namely, the parameters).  However, having these is very likely overkill.

This might be used, eventually, to fix #37033


mantepse:
I am now trying to fix coercion.  I tried the following:
```diff
diff --cc src/sage/combinat/sf/sfa.py
index 0fd2d890b8,0fd2d890b8..32a8dac3da
--- a/src/sage/combinat/sf/sfa.py
+++ b/src/sage/combinat/sf/sfa.py
@@@ -3013,6 -3013,6 +3013,30 @@@ class SymmetricFunctionAlgebra_generic(
          return self.tensor_square().sum(coeff * tensor([self(s[x]), self(s[y])])
                                          for ((x,y), coeff) in s(elt).coproduct())

++    def _coerce_map_from_(self, S):
++        """
++        Return ``True`` if a coercion from ``S`` exists.
++
++        EXAMPLES::
++
++            sage: S = SymmetricFunctions(QQ).s()
++            sage: S.has_coerce_map_from(QQ)
++            True
++            sage: S.has_coerce_map_from(ZZ)
++            True
++            sage: S.has_coerce_map_from(SymmetricFunctions(ZZ).s())
++            True
++        """
++        R = self.base_ring()
++        if R.has_coerce_map_from(S):
++            return True
++
++        if (isinstance(S, SymmetricFunctionAlgebra_generic)
++            and R.has_coerce_map_from(S.base_ring())):
++            return True
++
++        return None
++
      def construction(self):
          """
          Return a pair ``(F, R)``, where ``F`` is a
```
However, this breaks things like
```
sage: Sym = SymmetricFunctions(QQ)
sage: sp = Sym.sp()
sage: m = Sym.m()
sage: m(sp([4]))
m[1, 1, 1, 1] + m[2, 1, 1] + m[2, 2] + m[3, 1] + m[4]
sage: Sym = SymmetricFunctions(ZZ)
sage: sp = Sym.sp()
sage: m = Sym.m()
sage: m(sp([4]))
File ~/sage-trac/src/sage/combinat/sf/classical.py:305, in SymmetricFunctionAlgebra_classical._element_constructor_(self, x)
    304 try:
--> 305     c = R(x)
    306 except (TypeError, ValueError):

File ~/sage-trac/src/sage/structure/parent.pyx:901, in sage.structure.parent.Parent.__call__()
    900 if no_extra_args:
--> 901     return mor._call_(x)
    902 else:

File ~/sage-trac/src/sage/structure/coerce_maps.pyx:163, in sage.structure.coerce_maps.DefaultConvertMap_unique._call_()
    162     print(type(C._element_constructor), C._element_constructor)
--> 163 raise
    164 

File ~/sage-trac/src/sage/structure/coerce_maps.pyx:158, in sage.structure.coerce_maps.DefaultConvertMap_unique._call_()
    157 try:
--> 158     return C._element_constructor(x)
    159 except Exception:

File ~/sage-trac/src/sage/rings/integer.pyx:712, in sage.rings.integer.Integer.__init__()
    711 
--> 712                 raise TypeError("unable to coerce %s to an integer" % type(x))
    713 

TypeError: unable to coerce <class 'sage.combinat.sf.symplectic.SymmetricFunctionAlgebra_symplectic_with_category.element_class'> to an integer

Cell In[34], line 1
----> 1 m(sp([Integer(4)]))

File ~/sage-trac/src/sage/structure/parent.pyx:901, in sage.structure.parent.Parent.__call__()
    899 if mor is not None:
    900     if no_extra_args:
--> 901         return mor._call_(x)
    902     else:
    903         return mor._call_with_args(x, args, kwds)

File ~/sage-trac/src/sage/structure/coerce_maps.pyx:163, in sage.structure.coerce_maps.DefaultConvertMap_unique._call_()
    161             print(type(C), C)
    162             print(type(C._element_constructor), C._element_constructor)
--> 163         raise
    164 
    165 cpdef Element _call_with_args(self, x, args=(), kwds={}) noexcept:

File ~/sage-trac/src/sage/structure/coerce_maps.pyx:158, in sage.structure.coerce_maps.DefaultConvertMap_unique._call_()
    156 cdef Parent C = self._codomain
    157 try:
--> 158     return C._element_constructor(x)
    159 except Exception:
    160     if print_warnings:

File ~/sage-trac/src/sage/combinat/sf/classical.py:307, in SymmetricFunctionAlgebra_classical._element_constructor_(self, x)
    305     c = R(x)
    306 except (TypeError, ValueError):
--> 307     raise TypeError("do not know how to make x (= {}) an element of self".format(x))
    308 else:
    309     if not c:

TypeError: do not know how to make x (= sp[4]) an element of self
```

mantepse:
This is essentially ready.  There is some documentation missing, and the repr must be polished.  Also, the method `corresponding_basis_over(self, R)` should be deprecated (and rewritten).  However, I'd like to have initial feedback before proceeding, because I am not too familiar with the coercion framework yet.

mantepse:
@nbruin, @tscrim I'd appreciate your feedback enormously!

In particular, I don't really know yet how to make the following work - should I make the `TensorProductFunctor` inherit from `MultivariateConstructionFunctor`?
```
sage: R.<x> = QQ[]
sage: h1 = SymmetricFunctions(QQ).h()
sage: h2 = SymmetricFunctions(R).h()
sage: x = tensor([h1[1], h1([])])
sage: y = tensor([h2([]), h2[1]])
sage: x + y
```
Currently this raises
```
TypeError: unsupported operand parent(s) for +:
'Symmetric Functions over Rational Field in the homogeneous basis # Symmetric Functions over Rational Field in the homogeneous basis'
and 
'Symmetric Functions over Univariate Polynomial Ring in x over Rational Field in the homogeneous basis # Symmetric Functions over Univariate Polynomial Ring in x over Rational Field in the homogeneous basis'
```

tscrim:
As I believe you surmised, the `TensorProductFunctor` does not inherit from the usual `ConstructionFunctor`:
```
sage: tensor.__class__.__mro__
(<class 'sage.categories.tensor.TensorProductFunctor'>,
 <class 'sage.categories.covariant_functorial_construction.CovariantFunctorialConstruction'>,
 <class 'sage.structure.unique_representation.UniqueRepresentation'>,
 <class 'sage.structure.unique_representation.CachedRepresentation'>,
 <class 'sage.misc.fast_methods.WithEqualityById'>,
 <class 'sage.structure.sage_object.SageObject'>,
 <class 'object'>)
```
cf., the `MPoly` functor:
```
sage: MP.__class__.__mro__
(<class 'sage.categories.pushout.MultiPolynomialFunctor'>,
 <class 'sage.categories.pushout.ConstructionFunctor'>,
 <class 'sage.categories.functor.Functor'>,
 <class 'sage.structure.sage_object.SageObject'>,
 <class 'object'>)
```
So it would be natural to inherit from `MultivariateConstructionFunctor` like `cartesian_product` does.

I am slightly worried about allowing tensor products to do pushouts, which secretly changes their base rings, could lead to subtle mathematical issues. Specifically, if $M$ and $N$ are $R$ and $S$ bimodules, resp., then I would think the tensor product $M \otimes N$ would naturally be defined over the smallest ring contained within $R \cap S$. Now we don't have any such functionality currently (the tensor product requires all modules to have the same base ring). I think everything is fine with extending scalars, but I wanted to just check with you that you also think everything is okay.

Some quick comments on the rest of the code:

- `_descriptor` doesn't really describe what the variable is supposed to do. I think `_construction_info` would be better for this.
- Likewise, rather than putting everything into the attribute, it would be better to have overloaded `construction()` methods for, e.g., Macdonald polynomials. This means better testing and more clear specifications. Although it might even be better to have separate functors from a general `WithRealizations`-type construction functor (which could grab information from the class name). This could wait for a subsequent PR if it is too much trouble.
- It probably would be best to have the various `__init__` methods in the common base classes set things based on their inputs. That way things are more uniform with less redundancies.
- It would be good to change all of the `corresponding_basis_over` to `change_ring`.
- It is worrisome that `SymmetricFunctionAlgebra_orthotriang` doesn't have a construction functor, or at least is needing to skip the `_test_construction`.

mantepse:
Thank you for looking into this!

I am still trying to figure out how to get a coercion from a tensor product over a base ring to a tensor product over a larger base ring to work.  The pushout works, after making `TensorProductFunctor` inherit from `MultivariateConstructionFunctor`, and allowing `TensorProductFunctor` in `CombinatorialFreeModule.change_ring`.

I am guessing I need to add a `has_coerce_map` somewhere, but it's not clear to me where.

The missing `SymmetricFunctionAlgebra_orthotriang` is indeed a bit of a worry, I don't really know how to do it. A similar problem exists with `SymmetricFunctionAlgebra_dual`.

The problem is that we would have to store functions (e.g., `scalar`) as part of the data of the construction, and it is ot completely clear to me whether this is OK.  I see, however, that `VectorFunctor` essentially does that: it can be passed a matrix `inner_product_matrix`.

mantepse:
got it, made a mistake :-)  Extension of the base ring when adding or multiplying two tensors now works.

I'm still not quite there: I also need extension of the base ring when multiplying with a scalar.

mantepse:
I am, very slowly, learning how to learn about the coercion model.
```
sage: sage: a = 3/2
sage: h = SymmetricFunctions(ZZ).h()
sage: X = tensor([h[1], h[2]])
sage: cm.record_exceptions()
sage: detect_element_action(X.parent(), a.parent(), False)
sage: print(cm.exception_stack()[0])
  File "/home/martin/sage-trac/src/sage/categories/pushout.py", line 4621, in pushout
    return all(Z)
  File "sage/categories/functor.pyx", line 391, in sage.categories.functor.Functor.__call__
    y = self._apply_functor(self._coerce_into_domain(x))
  File "sage/categories/functor.pyx", line 304, in sage.categories.functor.Functor._coerce_into_domain
    raise TypeError("x (=%s) is not in %s" % (x, self.__domain))
TypeError: x (=Symmetric Functions over Integer Ring in the homogeneous basis # Symmetric Functions over Integer Ring in the homogeneous basis) is not in Category of integral domains

  File "sage/structure/coerce_actions.pyx", line 223, in sage.structure.coerce_actions.detect_element_action
    return (RightModuleAction if X_on_left else LeftModuleAction)(Y, X, y, x)
  File "sage/structure/coerce_actions.pyx", line 334, in sage.structure.coerce_actions.ModuleAction.__init__
    self.extended_base = pushout(G, S)
  File "/home/martin/sage-trac/src/sage/categories/pushout.py", line 4629, in pushout
    raise CoercionException(ex)
sage.structure.coerce_exceptions.CoercionException: x (=Symmetric Functions over Integer Ring in the homogeneous basis # Symmetric Functions over Integer Ring in the homogeneous basis) is not in Category of integral domains
```

mantepse:
I admit that I have no idea why sage is asking for an integral domain.  Moreover, when I brute-force provide `X.parent().is_integral_domain = lambda : True`, then it next asks for the method `fraction_field`.  I don't think I want to provide this one.

mantepse:
Yes, I will open a second one. I was hoping to get a prototype quickly, but could not get scalars to work, see above.

mantepse:
@tscrim commented:
> * `_descriptor` doesn't really describe what the variable is supposed to do. I think `_construction_info` would be better for this.

I agree.

> * Likewise, rather than putting everything into the attribute, it would be better to have overloaded `construction()` methods for, e.g., Macdonald polynomials. This means better testing and more clear specifications. Although it might even be better to have separate functors from a general `WithRealizations`-type construction functor (which could grab information from the class name). This could wait for a subsequent PR if it is too much trouble.

I'm not completely sure.  I doesn't sound right to me to use the class name - although I am relying on class names myself.  But doing essentially the same thing with a class seems like overkill at the moment.  

Hm, perhaps we could somehow pass the info in "_descriptor" down the the `__init__` method of `SymmetricFunctionAlgebra_generic`, which would then call the `WithRealizations` functor?  Is that what you have in mind?

> * It probably would be best to have the various `__init__` methods in the common base classes set things based on their inputs. That way things are more uniform with less redundancies.

I don't quite understand: are you referring to the `__init__` methods of, say, `SymmetricFunctionAlgebra_elementary`?  (I'm guessing not).  Or perhaps `MacdonaldPolynomials_p` and `MacdonaldPolynomials_generic`. and replace

```
class MacdonaldPolynomials_p(MacdonaldPolynomials_generic):
    def __init__(self, macdonald):
...        
        MacdonaldPolynomials_generic.__init__(self, macdonald)
        self._descriptor = (("macdonald", {"q": self.q, "t": self.t}), ("P",))
```

with 

```
class MacdonaldPolynomials_generic(sfa.SymmetricFunctionAlgebra_generic):
    def __init__(self, macdonald):
...
        self._descriptor = (("macdonald", {"q": self.q, "t": self.t}),)
...
class MacdonaldPolynomials_p(MacdonaldPolynomials_generic):
    def __init__(self, macdonald):
...        
        MacdonaldPolynomials_generic.__init__(self, macdonald)
        self._descriptor += (("P",),)
```

> * It would be good to change all of the `corresponding_basis_over` to `change_ring`.

I thought I did, or do you mean the doctest of `corresponding_basis_over`?

> * It is worrisome that `SymmetricFunctionAlgebra_orthotriang` doesn't have a construction functor, or at least is needing to skip the `_test_construction`.

I should try, based on your comment above.

tscrim:
> @tscrim commented:
> > * Likewise, rather than putting everything into the attribute, it would be better to have overloaded `construction()` methods for, e.g., Macdonald polynomials. This means better testing and more clear specifications. Although it might even be better to have separate functors from a general `WithRealizations`-type construction functor (which could grab information from the class name). This could wait for a subsequent PR if it is too much trouble.
> 
> I'm not completely sure. I doesn't sound right to me to use the class name - although I am relying on class names myself. But doing essentially the same thing with a class seems like overkill at the moment.

We use the class names because they are called as such from the "main" algebra object. So it would reduce the amount of hard-coding and make the general case easier to construct. Anyways, it doesn't have to be done here and now.

> Hm, perhaps we could somehow pass the info in "_descriptor" down the the `__init__` method of `SymmetricFunctionAlgebra_generic`, which would then call the `WithRealizations` functor? Is that what you have in mind?

Essentially. More of each `__init__` modifies/sets the `_descriptor` attribute. However, it feels like having different classes for the constructors or at least passing the appropriate data should be done by each class setting its own `construction()` method. This also would stay out of the way of things done in `__init__`.

> > * It probably would be best to have the various `__init__` methods in the common base classes set things based on their inputs. That way things are more uniform with less redundancies.
> 
> I don't quite understand: are you referring to the `__init__` methods of, say, `SymmetricFunctionAlgebra_elementary`? (I'm guessing not). Or perhaps `MacdonaldPolynomials_p` and `MacdonaldPolynomials_generic`. and replace

I am thinking of using the latter ones.

> > * It would be good to change all of the `corresponding_basis_over` to `change_ring`.
> 
> I thought I did, or do you mean the doctest of `corresponding_basis_over`?

Ah, yes. Sorry for not being clear about this. We only need one (quick) test showing the redirect works. The rest are for `change_ring`.

mantepse:
@tscrim, at https://github.com/mantepse/sage/tree/sf/construction-functional there is now an alternative version.  I would prepare the other alternative (as we originally had it) within the next few days, so you can then select the one you like best.

tscrim:
I very much prefer the alternative version we discussed. Let me just say one more time I find it bad form to store an arbitrary amount of input arguments and not have it associated to a subclass, which would know and expect a fixed input. As I said, I won't insist on changing it, but you just have to hear me say it another time. (Alternatively, why have the, e.g., `DualBasisFunctor`?)

One minor comment:, it is `.. WARNING::` with two colons.

mantepse:
As promised: https://github.com/mantepse/sage/tree/sf/construction-many-functors

PR #37223, opened 2024-02-02
Remove invariants commands in global namespace

GiacomoPope:
This PR removes the depreciated invariants from #28064 (2019). It's been long enough and we should declutter the tab-autocomplete from these very specific functions.

As an aside, I'm not sure why we have:

```py
from . import monsky_washnitzer
```

Instead of importing objects from this module, but I have not worked with this myself so am loathed to modify this without further input from the community.

### DIFF
--- a/src/sage/schemes/hyperelliptic_curves/all.py
+++ b/src/sage/schemes/hyperelliptic_curves/all.py
@@ -1,45 +1,4 @@
-"""
-Tests for deprecations of imports in global namespace from :trac:`28064`::
-
-    sage: igusa_clebsch_invariants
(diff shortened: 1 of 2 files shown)


fchapoton:
one failing doctest needs to be fixed

GiacomoPope:
New push fixes the doctest by including an import.

I'm not sure why `absolute_igusa_invariants_wamelen()` is being tested in the doctests of `igusa_clebsch_invariants()` instead of `absolute_igusa_invariants_wamelen()` itself... But rather than change this I just fixed it.

fchapoton (approved):

PR #37225, opened 2024-02-02
splitting_field: polred -> polredbest

AurelPage:
This PR partially addresses https://github.com/sagemath/sage/issues/22165. It removes one (in `splitting_field`) of the two uses of the Pari deprecated function `polred`. The other one is less trivial to remove and should probably be discussed in the issue.

### :memo: Checklist

- [ ] I have created tests covering the changes. [the code is meant to be functionally equivalent]

### :hourglass: Dependencies
None

### DIFF
--- a/src/sage/rings/number_field/splitting_field.py
+++ b/src/sage/rings/number_field/splitting_field.py
@@ -147,7 +147,7 @@ def splitting_field(poly, name, map=False, degree_multiple=None, abort_degree=No
 
     - ``simplify`` -- (default: ``True``) during the algorithm, try
       to find a simpler defining polynomial for the intermediate
-      number fields using PARI's ``polred()``.  This usually speeds
+      number fields using PARI's ``polredbest()``.  This usually speeds
       up the computation but can also considerably slow it down.
       Try and see what works best in the given situation.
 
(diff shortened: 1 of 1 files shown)


videlec:
Wouldn't it be better to use directly pari `nfsplitting`?

AurelPage:
@videlec Yes it would be. But I was planning to do that in a separate change. The reason is that `nfsplitting` does not have all the options that sage's `splitting_field` allows (namely `abort_degree`; in addition, Pari always uses the equivalent of the option `simplify`), so if we don't keep the sage code, the interface will not be backwards compatible. So we may as well fix the sage code first, unless you are happy to change the interface.

videlec:
> @videlec Yes it would be. But I was planning to do that in a separate change. The reason is that `nfsplitting` does not have all the options that sage's `splitting_field` allows (namely `abort_degree`; in addition, Pari always uses the equivalent of the option `simplify`), so if we don't keep the sage code, the interface will not be backwards compatible. So we may as well fix the sage code first, unless you are happy to change the interface.

Indeed, it seems more reasonable to make pari `nfsplitting` an available option to `splitting_field`. And also to do that in a further ticket.

videlec (approved):

PR #37226, opened 2024-02-02
unit group of number field: do not expand product when S is empty

### LINKED ISSUE #36386: Unexpected KeyError and "PariError: the PARI stack overflows" while computing UnitGroup
### Steps To Reproduce

The issue is illustrated with the following code:
```
d = 12936642
K = QuadraticField(d)
K.unit_group(proof=False)
```

### Expected Behavior

The code should compute the unit group without an issue, since directly calling `pari.bnfinit(f'x^2 - {d}', 1)` computes the unit group structure in a fraction of second.

### Actual Behavior

Instead, the code produces a bunch of errors including:

> KeyError: ((<class 'sage.rings.number_field.unit_group.UnitGroup'>, Number Field in a with defining polynomial x^2 - 12936642 with a = 3596.754370262167?, False), ())

and

> PariError: the PARI stack overflows (current size: 1024000000; maximum size: 1073741824)

### Additional Information

The "PariError: the PARI stack overflows" appears to be quite irrelevant to computing the unit group since

- direct call to `pari.bnfinit(f'x^2 - {d}', 1)` runs fine with the same size of PARI stack;
- increasing PARI stack even to 2GB does not help here as the same errors show up.

### Environment

```markdown
- **OS**: Ubuntu 22.04.3 LTS
- **Sage Version**: 10.2.beta5
```

### Checklist

AurelPage:
This PR fixes https://github.com/sagemath/sage/issues/36386, at least for unit group computations (S-units will still suffer from the problem, or when the units are even larger). This PR does **not** address https://github.com/sagemath/sage/issues/31754, which would be a better (but more complicated) solution for the problem.

### :memo: Checklist

### :hourglass: Dependencies
None

### DIFF
--- a/src/sage/rings/number_field/unit_group.py
+++ b/src/sage/rings/number_field/unit_group.py
@@ -305,6 +305,13 @@ class UnitGroup(AbelianGroupWithValues_class):
             sage: tuple(US(K(u)) for u in US.gens()) == US.gens()
             True
 
+        Bug #36386 (pari stack overflow while expanding units)
+            sage: d = 12936642
+            sage: K = QuadraticField(d)
+            sage: K.unit_group(proof=False)
+            Unit group with structure C2 x Z of Number Field in a with defining polynomial x^2 - 12936642 with a = 3596.754370262167?
+
+
         """
         proof = get_flag(proof, "number_field")
         K = number_field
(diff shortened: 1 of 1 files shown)


src/sage/rings/number_field/unit_group.py
```diff
@@ -305,6 +305,13 @@ def __init__(self, number_field, proof=True, S=None):
             sage: tuple(US(K(u)) for u in US.gens()) == US.gens()
             True
 
+        Bug #36386 (pari stack overflow while expanding units)
```
videlec: Should be
```
Bug :issue:`36386` (pari stack overflow while expanding units)::

    sage: d = 12936642
```

videlec: (ie there are three mistakes)

AurelPage: Thanks, fixed!

videlec (approved):

PR #37228, opened 2024-02-03
some pep8 fixes in rings, and activate E221

fchapoton:
This fixes all remaining pycodestyle E221 warnings in py files and activate this check in the pycodestyle-minimal linter.

This is about `E221 multiple spaces before operator`

### :memo: Checklist


tscrim:
Overall, the changes are fine with me, but I am slightly worried about forcing E221 on code as there could be a good reason to have the spaces (demonstrating an alignment in, e..g, a matrix). It might not happen, but I don't want to force it when it could make the code locally harder to parse.

fchapoton:
Well, there is no such case in our code base right now. And pep008 suggest to avoid spaces for alignement:

https://peps.python.org/pep-0008/#pet-peeves

But if you insist, I can remove the activation.

Note also that I have not touched the similar E222, which still gives some room for such visual spacings

https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes

tscrim (approved):
Indeed, but simple cases versus complicated formulas might matter. Well, we can activate it for now and revisit it if a major example comes up.

Issue #37124, opened 2024-01-21
Disable spammer bot: [Documentation preview for this PR](https://deploy-preview-37123--sagemath.netlify.app/html/en) (built with commit 78fee0af4a0ab42b6afff5df72d474154f0cd93d; [changes](https://deploy-preview-37123--sagemath.netlify.app/CHANGES.html)) is ready! :tada:


tornaria:
This bot spams our email with an irrelevant comment for every push on every commit. Please figure out a different way to publish the link to the documentation, or make it optional.

              [Documentation preview for this PR](https://deploy-preview-37123--sagemath.netlify.app/html/en) (built with commit 78fee0af4a0ab42b6afff5df72d474154f0cd93d; [changes](https://deploy-preview-37123--sagemath.netlify.app/CHANGES.html)) is ready! :tada:

_Originally posted by @github-actions[bot] in https://github.com/sagemath/sage/issues/37123#issuecomment-1902638736_

PR #37231, opened 2024-02-03
Use matroid-database package

gmou3:
Remove matroid database files (`ext_data/matroids`) and use newly created `matroid-database` pypi package instead.

Links to `matroid-database`: [pypi](https://pypi.org/project/matroid-database/), [github](https://github.com/gmou3/matroid-database)

Outdated:
> Compress (via `xz -e`) the matroid database files (~75MB to <1MB). The files are now opened with `lzma.open()` and the file paths are located through `sage.ext_data` and not relative to `SAGE_SRC` or `SAGE_EXTCODE`. These changes attempt to address the issues raised by @tornaria.
> 
> Also, the addition of the matroid database files inside the folder `ext_data` needs to be noted in the file `setup.cfg.m4`. This issue was brought to my attention by @antonio-rojas.
> 
> These follow the merging of #37140. 
> 
> I also took the initiative of reordering alphabetically the list of paths inside `setup.cfg.m4` and of removing the line `ext_data/images/*` which seemed redundant.

### :memo: Checklist


antonio-rojas:
Perfect, thanks

tornaria:
I copy below my comment from #37124. IMHO that PR should be reverted, and replaced by an external database package. Also, I understand that `SAGE_EXTCODE` is deprecated (#33037).

---
This PR almost doubles the size of `src/sage`:
`
$ du -sh src/sage
180M	src/sage
$ du -sh src/sage/ext_data/matroids/
75M	src/sage/ext_data/matroids/
`
This is not the way databases are handled in sagemath. The normal way would be to make an *optional* spkg containing the database. The nicer/easier way would be to create a pypi package similar to https://github.com/sagemath/conway-polynomials (btw, in conway-polynomials the actual data files are xz compressed and transparently decompressed as needed)

Also, as @antonio-rojas pointed out, this broke `sagemath-standard`.

tscrim:
@tornaria You mean #37140, correct?

However, I am not sure how much I favor creating additional optional as they are a pain to install unless you have a source build IIRC. Putting things on pypi currently means it goes out into the either and nobody really knows it exists. We (the Sage community) don't really have a good policy for how deal with small databases. (The pypi issue is a bigger thing for me. I am not aware of any good way to find what's available, and I don't have a lot of trust with that code to stay current with changes to Sage.)

In this case, we have a database that is available online but it isn't packaged in any standard form. I agree that this is not the best solution, but without a good way to install optional packages (which would be preference otherwise), I don't see a better option for a somewhat small database like this. Or has something changed with optional package installation?

Lastly,  I wasn't aware of the deprecation. It does not really seem like a formal deprecation, much less have a plan to replace it.

gmou3:
Please have a look at the changes.

tornaria:
> @tornaria You mean #37140, correct?

Indeed

> However, I am not sure how much I favor creating additional optional as they are a pain to install unless you have a source build IIRC.

It should be as simple as `sage -pip install matroid-database` or something like that. Alternatively, you could make a vote in sage-devel to make the package standard instead of optional. This way is only simpler because the vote on sage-devel was omitted.

> Putting things on pypi currently means it goes out into the either and nobody really knows it exists.

For sage-the-distro is just a (standard or optional) spkg as always. In general, if it's optional it can be handled by "features" which means when you want to use the feature, if it's not available, it will explain what is missing and what is the suggested way to install the feature depending on your installation.

> We (the Sage community) don't really have a good policy for how deal with small databases. 

We do. We have a number of databases which are external spkgs, and some work has been started to move them to python packages (I mentioned conway-polynomials, which is a core database which is used to create finite fields, very non-optional I think). The python package is very simple, easy to install in varying circumstances (via pip, as a spkg, as a distro package, conda, etc).

Here's the data packages I have installed (I think they are all the standard data packages):
```
$ du -sch /usr/share/sagemath/* | sort -h
$ du -sch /usr/share/sagemath/* | sort -h
36K	/usr/share/sagemath/combinatorial_designs
188K	/usr/share/sagemath/jones
332K	/usr/share/sagemath/reflexive_polytopes
668K	/usr/share/sagemath/threejs-sage
732K	/usr/share/sagemath/conway_polynomials
2.2M	/usr/share/sagemath/ellcurves
3.5M	/usr/share/sagemath/graphs
8.1M	/usr/share/sagemath/cremona
16M	total
```
These are all smaller than the matroid database, and still they were deemed to be external packages.

> (The pypi issue is a bigger thing for me. I am not aware of any good way to find what's available, and I don't have a lot of trust with that code to stay current with changes to Sage.)

The code would be as a sagemath project in github (e.g. https://github.com/sagemath/conway-polynomials). Having it in pypi means they are easy to install using pip, etc. And sage-the-distro would still have a spkg for this which should be very easy to write and maintain (as are distro packages). Moreover, the databases change very seldom which means all of that data is installed once and doesn't need to change when updating sagemath, etc.

In summary, having a database as a spkg is the usual policy for sagemath. Having a database as a "proper" python package in pypi instead of just a custom spkg has some advantages and there would still be a spkg (based on the pypi package instead of custom spkg).

Anyway, if you prefer to have a "custom spkg" version of the database, that should be fine as well. The current way is that databases are searched for in a few paths given by `SAGE_DATA_PATH` which defaults to something like `~/.sage/db/<NAME>`, `$SAGE_LOCAL/share/sagemath/<NAME>`, `$SAGE_LOCAL/share/<NAME>`, which means a user can easily just put the files in its own home directory if they are not installed in the system.

> In this case, we have a database that is available online but it isn't packaged in any standard form. I agree that this is not the best solution, but without a good way to install optional packages (which would be preference otherwise), I don't see a better option for a somewhat small database like this. Or has something changed with optional package installation?

If you want this database to be standard instead of optional, then a vote has to be held in sage-devel. But the policy for databases (either standard or optional) is to have them as separate packages.

> Lastly, I wasn't aware of the deprecation. It does not really seem like a formal deprecation, much less have a plan to replace it.

No, the idea is that package data should go inside packages and use importlib.resources to access the data. This is for data tied to sage source code.

For databases, the policy AFAICT is to have them as separate spkgs.

tornaria:
> Please have a look at the changes.

I already stated my opinion: I don't think it makes sense to add database files to the repo, even if they are compressed. These files seem to be 10+ years old, we don't need to update them every release (otoh, if they had some bug or were extended, it's nice if they can be updated without having to change sage!)

Having an external spkg is the current policy. Making the database "standard" AFAIK requires a vote in sage-devel.

Note: even if the spkg is only optional, if the database files are searched via `SAGE_DATA_PATH` as done with several other databases (see #37024) it should be possible for a user to place the files in its home dir if not installed in system.

Regarding the compression:
 - for files smaller thank 4k (most of them) it may be better to not compress since data is usually read in 4k blocks anyway.
 -  For the large files I think the best is to use something like `xz -T 2 -e` which gives around half the size than gz.
 - Using xz should not be a problem since lzma is included in python standard library since python 3.3.

With this the largest files are:
```
20K	all_matroids/allr3n10.txt.xz
24K	unorientable_matroids/unorientabler3n11.txt.xz
28K	unorientable_matroids/unorientabler4n09.txt.xz
244K	all_matroids/allr4n09.txt.xz
360K	all_matroids/allr3n11.txt.xz
```
Every other file fits in 4k and the total is:
```
$ du -sch all_matroids/ unorientable_matroids/
768K	all_matroids/
80K	unorientable_matroids/
848K	total
```

src/sage/matroids/database_collections.py
```diff
@@ -195,24 +197,20 @@ def AllMatroids(n, r=None, type="all"):
             rp = min(r, n - r) if (type != "unorientable") else r
             type_file = "all" if (type != "unorientable") else "unorientable"
             file = os.path.join(
-                str(SAGE_EXTCODE), "matroids", "database",
+                str(SAGE_SRC), "sage", "ext_data", "matroids", "database",
```
tornaria: Using `SAGE_SRC` is not the way to avoid using `SAGE_EXTCODE`. The idea of #33037 is that one should not use neither `SAGE_LIB` nor `SAGE_SRC`, and `SAGE_EXTCODE`.

Instead, one is supposed to use `importlib.resources` to access package data (which I guess should be placed in the package that owns the data instead of the generic `src/sage/ext_data` although I guess you could still access files in there via `sage.ext_data`).

gmou3:
> I already stated my opinion: I don't think it makes sense to add database files to the repo, even if they are compressed. These files seem to be 10+ years old, we don't need to update them every release (otoh, if they had some bug or were extended, it's nice if they can be updated without having to change sage!)
> 
> Having an external spkg is the current policy. Making the database "standard" AFAIK requires a vote in sage-devel.
> 
Fair enough. If there is consensus on this or you are confident enough of this requirement let me know and I'll try to make it into a package.
 
> Regarding the compression:
> 
> * for files smaller thank 4k (most of them) it may be better to not compress since data is usually read in 4k blocks anyway.
> * For the large files I think the best is to use something like `xz -T 2 -e` which gives around half the size than gz.
> * Using xz should not be a problem since lzma is included in python standard library since python 3.3.
> 
> With this the largest files are:
> 
> ```
> 20K	all_matroids/allr3n10.txt.xz
> 24K	unorientable_matroids/unorientabler3n11.txt.xz
> 28K	unorientable_matroids/unorientabler4n09.txt.xz
> 244K	all_matroids/allr4n09.txt.xz
> 360K	all_matroids/allr3n11.txt.xz
> ```
> 
> Every other file fits in 4k and the total is:
> 
> ```
> $ du -sch all_matroids/ unorientable_matroids/
> 768K	all_matroids/
> 80K	unorientable_matroids/
> 848K	total
> ```
That's quite cool. I compressed them all for consistency/simplicity.

videlec:
> @tornaria You mean #37140, correct?
> 
> However, I am not sure how much I favor creating additional optional as they are a pain to install unless you have a source build IIRC. Putting things on pypi currently means it goes out into the either and nobody really knows it exists. We (the Sage community) don't really have a good policy for how deal with small databases. (The pypi issue is a bigger thing for me. I am not aware of any good way to find what's available, and I don't have a lot of trust with that code to stay current with changes to Sage.)
> 
> In this case, we have a database that is available online but it isn't packaged in any standard form. I agree that this is not the best solution, but without a good way to install optional packages (which would be preference otherwise), I don't see a better option for a somewhat small database like this. Or has something changed with optional package installation?
> 
> Lastly, I wasn't aware of the deprecation. It does not really seem like a formal deprecation, much less have a plan to replace it.

I do not quite understand your objection. Any package on pypi can be installed on any system with python. The package might have a build-dependency or run-dependency on sage, and this is handled by pypi. For a purely data+python package such as the one discussed here it would only be run-dependency. As an example, you can try
```
$ pip install admcycles
```
which is a pure python library with run-dependency on sage.

What is a pain in the ass are the external packages that requires a recompilation of the sage source code in order to work. For this not only you need a compiled version of sage but also to recompile part of sage to get it work. And this is a nonsense to package.

About advertisement, good projects (as admcycles) tend to survive if people use it and developers maintain it. And I think that dead code in sage is not better than a dead project.

tscrim:
Let me first start with my proposal going forward. A pypi package for the database could work provided we have a clear mechanism for telling a user what they need to do to use the feature (both in code when trying to do it without installing the database and in the documentation with a link to the database). This is assuming that we don't have an auto-download/install feature, but perhaps that becomes a sysadmin security concern?

I am upgrading this to a blocker as I think we all agree that we need to settle this before the next full release.

----

Now let me explain my positions a bit more.

@videlec @tornaria Note there are many differences between Sage optional packages and pypi package.

Optional Sage packages:
- Explicitly listed as part of Sage.
- We say that they should work with the latest version of Sage.
- They are somewhat regularly tested.
- There is no easy way to install them on non-source-builds (AFAIK).

pypi packages:
- Easy enough to install.
- No quality control: there isn't any check or promise that they should work when Sage upgrades.
- No list of packages for optional Sage features or downstream code.
- Similarly, no way to know what is needed for a particular optional feature.

It isn't about the package run-time dependency on Sage, but the other way around: Sage's run-time dependency on a pypi package. This case might be a good place to start setting a policy about this, or at least one reasonable attempt at one.

> These files seem to be 10+ years old, we don't need to update them every release (otoh, if they had some bug or were extended, it's nice if they can be updated without having to change sage!)

I find the argument to be a weak one in general, but a bit stronger for databases. We generally change Sage to push the (minimum) version numbers to not include buggy code. I don't see how we are updating them every release. There are some mild benefits for upgrading one small part independent of our release cycle. Yet we release somewhat frequently and I would tell people just to upgrade Sage.

> About advertisement, good projects (as admcycles) tend to survive if people use it and developers maintain it. And I think that dead code in sage is not better than a dead project.

Who is to judge what counts as dead code? If someone is trying to use it but it is broken, then is it really dead? Frankly, I find this argument completely unconvincing as this doesn't tell someone (who is not in-the-know) what is available.

tornaria:
Some quick comments:

- sage already depends on a lot of pypi packages, both standard and optional
- any pypi package cam be easily added as a spkg, in that sense having the database in pypi is more versatile than having it only as a spkg
- having databases as separate packages has been the policy forever afaik
- moving packages (and databases) from optional to standard requires a vote; adding the database to the source should have the same requirement.

I believe it was a mistake to merge #37140, at least without a wider discussion. The way forward IMO is to revert it and have the discussion or just add the database as a spkg (pypi or not). Even having the database as a standard package would be way, way better than having it in the source.

src/sage/matroids/database_collections.py
mkoeppe: That's right, using `SAGE_SRC` should also be avoided completely because this variable does not have a meaningful definition in the presence of PEP 420 namespace packages.

mkoeppe:
> The pypi issue is a bigger thing for me. I am not aware of any good way to find what's available, and I don't have a lot of trust with that code to stay current with changes to Sage.

@tscrim This is a solved problem. We advertise packages that are on PyPI by creating an SPKG for it. This puts them in our package index https://doc.sagemath.org/html/en/reference/spkg/index.html#optional-packages
We also run automatic tests on them in each release.

mkoeppe:
For @videlec's example package "admcycles", we also have a corresponding SPKG: https://doc.sagemath.org/html/en/reference/spkg/admcycles.html#spkg-admcycles 
Again having it there not only advertises it as part of our documentation, but it also includes it in our automatic tests on GH Actions.

mkoeppe:
Also for reference:
- https://deploy-livedoc--sagemath.netlify.app/html/en/developer/packaging#package-source-types -- note the three options "normal" / "wheel" / "pip" packages
- https://deploy-livedoc--sagemath.netlify.app/html/en/developer/packaging#utility-script-to-create-and-maintain-packages

tscrim:
> > The pypi issue is a bigger thing for me. I am not aware of any good way to find what's available, and I don't have a lot of trust with that code to stay current with changes to Sage.
> 
> @tscrim This is a solved problem. We advertise packages that are on PyPI by creating an SPKG for it. This puts them in our package index https://doc.sagemath.org/html/en/reference/spkg/index.html#optional-packages We also run automatic tests on them in each release.

A solution with duct-tape and glue at best. I don't see any documentation saying someone should do that (or even encouraging it). An unmarked list interspersed within a huge list contained in an even bigger technical-looking list isn't really advertising. Nothing is said about any automated testing; the only way I am aware of testing on optional packages is by us in the community testing in various combinations. I feel like you are trying to sell me a hamburger before it is cooked.

mkoeppe:
> A solution with duct-tape and glue at best.

Really now?

tscrim:
What about my points is wrong?

mkoeppe:
I'll not comment more on the obvious "order of operations" mistake. 

But I'll respond to this point:
> [...] I don't see any documentation saying someone should do that [...]

Generally, our documentation does not tell people what they "should" do. It only tells them "how" to do it when they want to. 

Pretty much the only exception is the [intro to the Developer's Guide](https://deploy-livedoc--sagemath.netlify.app/html/en/developer/), which very briefly "encourages" "everybody who uses Sage" to do one of a short list of things. Right after the last bullet point it reverts to just telling "how" to do things.

Organizing development activities (that includes motivating people to do these activities or perhaps even telling them what they "should" do) is not what our documentation does, but I'd argue that we should (!) do better. We should improve our documentation so that it is much more inviting -- certainly in tone, but also in practical terms of guiding people to things that they "could" do. 

Some time back, I created some meta-tickets for this purpose. The one that I linked above, 
- https://github.com/sagemath/sage/issues/31164

is one of them. Among other things, it does explain why people may be interested in doing this task. 

I planned to add links to such meta-tickets to the documentation in:
- #34526

This may be worth revisiting.

tscrim:
The developer’s guide is full of things that we tell people they should do to contribute to Sage beyond the introduction. That is an essential part of the “how.” So shouldn’t writing external packages and advertising them as part of Sage be a part of that?

When I go out to tell people they should use Sage, I get asked, “Can Sage do X?” I want to be able to do point them to a place to check if I don’t know, and I also want to be able to give them a reasonably high level of guarantee that it will work. I rely on what is tested and advertised to make that promise.

mkoeppe:
> is full of things

[citation needed]

gmou3:
My impression as a new contributor was that the folder `ext_data` was an appropriate place for external data.

I think a matroid-database pypi package would be a nice thing for the following reasons:
- the site https://www-imai.is.s.u-tokyo.ac.jp/~ymatsu/matroid/index.html is often down;
- it could also be useful outside sage as long as we add some processing functions (e.g. decoding the `revlex` encoding into a `list` of matroid bases).

This will require some work though. Having a look at `conway-polynomials`, it seems like a somewhat ornate construction given the simplicity of its functionality.

tornaria:
> My impression as a new contributor was that the folder `ext_data` was an appropriate place for external data.

It sure looks like that, but it's mostly for supporting code or data, not databases. And you don't see anything close to 75M (the whole of `src/sage` is 105M).

> I think a matroid-database pypi package would be a nice thing for the following reasons:
> 
>     * the site https://www-imai.is.s.u-tokyo.ac.jp/~ymatsu/matroid/index.html is often down;

Indeed good reason to have the data mirrored. I think some of the databases in sagemath are only available as tarballs in the sagemath upstream mirror at https://mirrors.mit.edu/sage/spkg/upstream/.

>     * it could also be useful outside sage as long as we add some processing functions (e.g. decoding the `revlex` encoding into a `list` of matroid bases).

Sure, you could start with just a couple of functions to access the "raw" data on the database to be used from sagemath, and later add extra functionality into your package.

> This will require some work though. Having a look at `conway-polynomials`, it seems like a somewhat ornate construction given the simplicity of its functionality.

Should be very simple, e.g.:
```python
def _open_data(module, name):
    import lzma
    from io import TextIOWrapper

    path = files(module).joinpath(name)
    try:
        xz = path.with_suffix(".txt.xz").open("rb")
        return TextIOWrapper(lzma.open(xz))
    except FileNotFoundError:
        return path.with_suffix(".txt").open("r")


def all_matroids_raw(rank, size):
    "Return an iterator over all matroids of given rank and size"
    from . import _all
    with _open_data(_all, f"allr{rank}n{size:02d}") as f:
        while s := f.readline():
            yield s.strip()


def unorientable_matroids_raw(rank, size):
    "Return an iterator over unorientable matroids of given rank and size"
    from . import _unorientable
    with _open_data(_unorientable, f"unorientabler{rank}n{size:02d}") as f:
        while s := f.readline():
            yield s.strip()
```
This would be `matroid_database/__init__.py` with data files placed in `matroid_database/_all/` and `matroid_database/_unorientable/`, and used like:
```
In [1]: from matroid_database import all_matroids_raw

In [2]: for m in all_matroids_raw(2, 5):
   ...:     print(m)
   ...: 
**********
0*********
0****0****
00*0**0***
000*******
000******0
0000**0***
0000**0**0
00000*00**
000000****
0000000***
00000000**
000000000*

In [3]: from matroid_database import unorientable_matroids_raw

In [4]: for m in unorientable_matroids_raw(3, 8):
   ...:     print(m)
   ...: 
0******0******0**********0********0*******0****0**0*****
0******0******0***0******0*0**0*************************
0000************0**********0****0**********0**0***0*****

```

You will need to add some boilerplate for this to become a python package in pypi, but nothing too heavy. For a start, see http://www.cmat.edu.uy/~tornaria/matroid-database-src.tar.gz:
```sh
$ wget http://www.cmat.edu.uy/~tornaria/matroid-database-src.tar.gz
$ tar xf matroid-database-src.tar.gz
$ cd matroid-database-0.1/
$ python -m build 
* Creating venv isolated environment...
[...]
Successfully built matroid-database-0.1.tar.gz and matroid_database-0.1-py3-none-any.whl
```
Testing works:
```
$ pytest
=========================== test session starts ============================
platform linux -- Python 3.12.1, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/tornaria/src/sage/matroid-database/tmp/matroid-database-0.1
configfile: pyproject.toml
plugins: hypothesis-6.97.0, xonsh-0.14.4, timeout-2.2.0, anyio-4.2.0
collected 2 items                                                          

README.rst .                                                         [ 50%]
src/matroid_database/__init__.py .                                   [100%]

============================ 2 passed in 1.15s =============================
```

That may need some more documentation / references, a license, and maybe some more pypi metadata. Feel free to take this and stick whatever license you want, change in whatever way.

gmou3:
@tornaria Thank you, I will build upon your tar.
After I have some decent version I will need to upload to pypi and github, right? Initially on my own accounts I presume.

tornaria:
> @tornaria Thank you, I will build upon your tar. After I have some decent version I will need to upload to pypi and github, right? Initially on my own accounts I presume.

I think you can start making a new github repo under your account, someone can later move it to `sagemath/*` if you want (you keep admin role on the repo). About pypi, I don't know, but maybe someone can create the package under sagemath admin and I think there's a way to have the sdist built and uploaded to pypi directly from github using gh actions (see .github dir in conway_polynomials). Maybe @orlitzky or @mkoeppe can help you with this.

mkoeppe:
> A question for @mkoeppe : suppose there is a "system" install of sagemath (either sage-the-distro or sagemath-standard) which one cannot modify.
> 
> Say a database like this one is a pypi and made an optional spkg. What would be the recommended way to install the database as a user?

Using pip.

But let's see what Sage has to say:
```
$ mv build/bin/sage-site build/bin/sage-site--hidden   # simulate an installation without 'sage -i'
$ chmod -R a-w venv/lib/python3.11/site-packages  # simulate non-writable installation
$ ./sage 
│ SageMath version 10.3.beta7, Release Date: 2024-02-02              │
│ Using Python 3.11.6. Type "help()" for help.                       │
┃ Warning: sage.all is not available; this is a limited REPL.        ┃
sage: from sage.features import PythonModule
sage: PythonModule('admcycles', spkg='admcycles').require()
FeatureNotPresentError                    Traceback (most recent call last)
Cell In[2], line 1
----> 1 PythonModule('admcycles', spkg='admcycles').require()

File ~/s/sage/sage-rebasing/worktree-pristine/src/sage/features/__init__.py:243, in Feature.require(self)
    241 presence = self.is_present()
    242 if not presence:
--> 243     raise FeatureNotPresentError(self, presence.reason, presence.resolution)

FeatureNotPresentError: admcycles is not available.
Failed to import `admcycles`: No module named 'admcycles'
No equivalent system packages for homebrew are known to Sage.
To install admcycles using the pip package manager, you can try to run:
!sage -pip install admcycles
sage: !sage -pip install admcycles
Collecting admcycles
  Downloading admcycles-1.4-py3-none-any.whl (351 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 351.1/351.1 kB 3.4 MB/s eta 0:00:00
Installing collected packages: admcycles
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/lib/python3.11/site-packages/admcycles'
Check the permissions.

sage: !sage -pip install --user admcycles
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

Certainly not great. If the installation had been configured using `./configure --without-system-python3`, then the last command would work. (This is something that the sysadmin who installs the Sage distribution should be made aware of.)

mkoeppe:
> My impression as a new contributor was that the folder `ext_data` was an appropriate place for external data.

@gmou3 Nobody can blame you for that. The developer's guide does not have much to say on `ext_data` (https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_basics.html#files-and-directory-structure), and most of what it does say is outdated. We have to update it based on https://github.com/sagemath/sage/issues/33037.

tornaria:
> > A question for @mkoeppe : suppose there is a "system" install of sagemath (either sage-the-distro or sagemath-standard) which one cannot modify.
> > Say a database like this one is a pypi and made an optional spkg. What would be the recommended way to install the database as a user?
> 
> Using pip.
> 
> But let's see what Sage has to say:

Let me try:

```
$ sage
│ Using Python 3.12.0. Type "help()" for help.                       │
sage: from sage.features import PythonModule
sage: PythonModule('admcycles', spkg='admcycles').require()
FeatureNotPresentError                    Traceback (most recent call last)
Cell In[2], line 1
----> 1 PythonModule('admcycles', spkg='admcycles').require()

File /usr/lib/python3.12/site-packages/sage/features/__init__.py:243, in Feature.require(self)
    241 presence = self.is_present()
    242 if not presence:
--> 243     raise FeatureNotPresentError(self, presence.reason, presence.resolution)

FeatureNotPresentError: admcycles is not available.
Failed to import `admcycles`: No module named 'admcycles'
No equivalent system packages for pip are known to Sage.
sage: !sage -pip install --user admcycles
error: externally-managed-environment

× This environment is externally managed
╰─> 
    This system-wide Python installation is managed by the Void Linux package
    manager, XBPS. Installation of Python packages from other sources is not
    normally allowed.

[... snippet about using a venv or use pipx ...]

```

> Certainly not great. If the installation had been configured using `./configure --without-system-python3`, then the last command would work. (This is something that the sysadmin who installs the Sage distribution should be made aware of.)

Not great = unusable.

We can adjust distro packages if there is a way, but having to use a vendored python is 100% out of reach.

This is strictly worse than having databases installed as plain old files and have sage look for them in `SAGE_DATA_PATH` which allows the user to unpack the database in `~/.sage/db/<DBNAME>` (right now this should work for some of the optional databases, maybe `polytopes_db_4d` or `database_cremona_ellcurve`).

What would take to make the same thing possible with python packages (i.e. have a location inside `~/.sage/, where python packages can be installed for use in sage).

mkoeppe:
> Even completely disregarding a major point of a developer's guide is to explain our coding practices and standards:

@tscrim Clearly we are miscommunicating. Yes, the developer's guide explains the coding practices and standards -- that's explaining "how" to make a contribution (... or else the contribution won't be merged). It does not explain "why" to make a contribution.

> 1. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/github#the-stopgap: "When Sage returns wrong results, an issue and a PR should be created:"

This example could almost be a fair point, but even for this one, the use of the passive voice here leaves it ambiguous who is addressed by it.

> 2. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/github#final-notes: "Every bug fixed should result in a doctest."

That's a "how". It describes what to do when you're preparing a PR that fixes the bug.

> 3. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_basics: "...all Python/Cython code that is included with Sage should adhere to the style conventions discussed in this chapter."

That's a "how". It gives constraints on the code style.

> 4. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_basics#python-code-style: "Follow the standard Python formatting rules when writing code for Sage, as explained at the following URLs:"
> 5. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_basics#files-and-directory-structure: "Python Sage library code goes into src/sage/ and uses the following conventions."; "Adding new top-level packages below sage should be done sparingly."
> 6. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_basics#headings-of-sage-library-code-files: "The top of each Sage code file should follow this format:"
> 7. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_in_python#python-language-standard: "Sage library code needs to be compatible with all versions of Python that Sage supports."
> 8. https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_basics#writing-testable-examples: "All new doctests added to Sage should _pass all tests_"

Likewise, all of these are "how"s.

In https://github.com/sagemath/sage/pull/37231#issuecomment-1928808617, you wrote:
> I don't see any documentation saying someone should [create an SPKG to advertise a package that is on PyPI] (or even encouraging it).

The "how" is documented in the [developer guide chapter "Packaging Third-Party Code for Sage"](https://deploy-livedoc--sagemath.netlify.app/html/en/developer/packaging).

tornaria:
This works with the distro sagemath:
```
$ python -m venv --system-site-packages usersage
$ source usersage/bin/activate
(usersage) $ pip install admcycles
[...]
Successfully installed admcycles-1.4

(usersage) $ sage
│ Using Python 3.12.0. Type "help()" for help.                       │
/usr/lib/python3.12/site-packages/IPython/core/interactiveshell.py:915: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
  warn(
sage: from sage.features import PythonModule
sage: PythonModule('admcycles', spkg='admcycles').is_present()
FeatureTestResult('admcycles', True)
```

But this probably won't work on sage-the-distro.

mkoeppe:
@tornaria Thanks for testing this with your void linux packaging.
> ```
> $ sage
> ┌────────────────────────────────────────────────────────────────────┐
> │ SageMath version 10.2, Release Date: 2023-12-03                    │
> │ Using Python 3.12.0. Type "help()" for help.                       │
> └────────────────────────────────────────────────────────────────────┘
> sage: from sage.features import PythonModule
> sage: PythonModule('admcycles', spkg='admcycles').require()
> ---------------------------------------------------------------------------
> FeatureNotPresentError                    Traceback (most recent call last)
> Cell In[2], line 1
> ----> 1 PythonModule('admcycles', spkg='admcycles').require()
> 
> File /usr/lib/python3.12/site-packages/sage/features/__init__.py:243, in Feature.require(self)
>     241 presence = self.is_present()
>     242 if not presence:
> --> 243     raise FeatureNotPresentError(self, presence.reason, presence.resolution)
> 
> FeatureNotPresentError: admcycles is not available.
> Failed to import `admcycles`: No module named 'admcycles'
> No equivalent system packages for pip are known to Sage.
> ```

I'd say that it does not offer the "pip" installation instructions is a flaw in downstream packaging.


> ```
> sage: !sage -pip install --user admcycles
> error: externally-managed-environment
> 
> × This environment is externally managed
> ╰─> 
>     This system-wide Python installation is managed by the Void Linux package
>     manager, XBPS. Installation of Python packages from other sources is not
>     normally allowed.
> ```

This is surprising to some extent. Is "pip install --user" disabled in general in Void Linux's system python? Do you have a reference that explains this policy, or more generally what users on Void Linux are expected to be able to do with their system Python? (Are there Void Linux users, or does it follow an "every distro user is a distro developer" model?)

> ```
> [... snippet about using a venv or use pipx ...]
> ```

What does it have to say about venv? What you snipped may be the interesting part, I think.

> > Certainly not great. If the installation had been configured using `./configure --without-system-python3`, then the last command would work. (This is something that the sysadmin who installs the Sage distribution should be made aware of.)
> 
> Not great = unusable.
> 
> We can adjust distro packages if there is a way, but having to use a vendored python is 100% out of reach.

As you know, "--without-system-python3" is a configuration option of Sage-the-distribution. It has no relevance for downstream packaging. All I'm saying is: When the Sage distribution sets up its venv, system site packages are normally disabled, and hence user site packages are also disabled. 

> What would take to make the same thing possible with python packages (i.e. have a location inside `~/.sage/, where python packages can be installed for use in sage).

It wouldn't be difficult, but I think it would be much better to direct users to setting up a suitable venv instead.

mkoeppe:
> This works with the distro sagemath:
> 
> ```
> $ python -m venv --system-site-packages usersage
> $ source usersage/bin/activate
> (usersage) $ pip install admcycles
> [...]
> Successfully installed admcycles-1.4
> 
> (usersage) $ sage
> ┌────────────────────────────────────────────────────────────────────┐
> │ SageMath version 10.2, Release Date: 2023-12-03                    │
> │ Using Python 3.12.0. Type "help()" for help.                       │
> └────────────────────────────────────────────────────────────────────┘
> /usr/lib/python3.12/site-packages/IPython/core/interactiveshell.py:915: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
>   warn(
> sage: from sage.features import PythonModule
> sage: PythonModule('admcycles', spkg='admcycles').is_present()
> FeatureTestResult('admcycles', True)
> ```

Yes, that's the way.

This is what the `FeatureNotPresentError` should be recommending in your installation.

> But this probably won't work on sage-the-distro.

Not exactly like this (because you can't make a venv over a venv), but we have the parts ready for it. 
```
$ cat venv/var/lib/sage/scripts/*/spkg-requirements.txt
alabaster @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/alabaster-0.7.12-py2.py3-none-any.whl
anyio @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/anyio-4.0.0-py3-none-any.whl
appnope @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/appnope-0.1.3-py2.py3-none-any.whl
argon2_cffi @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/argon2_cffi-21.3.0-py3-none-any.whl
argon2_cffi_bindings @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/argon2_cffi_bindings-21.2.0-cp311-abi3-macosx_14_0_x86_64.whl
arrow @ file:///Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels/arrow-1.3.0-py3-none-any.whl
...
```
The last missing bit:
- https://github.com/sagemath/sage/issues/32616

orlitzky:
FWIW:

```
$ pip install --user admcycles
error: externally-managed-environment

× This environment is externally managed
╰─> 
    The system-wide Python installation in Gentoo should be maintained
    using the system package manager (e.g. emerge).
    
    If the package in question is not packaged for Gentoo, please
    consider installing it inside a virtual environment, e.g.:
    
    python -m venv /path/to/venv
    . /path/to/venv/bin/activate
    pip install mypackage
    
    To exit the virtual environment, run:
    
    deactivate
    
    The virtual environment is not deleted, and can be re-entered by
    re-sourcing the activate file.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

Otherwise it's too easy for users to break the entire system by following some tutorial.

mkoeppe:
Thanks for checking @orlitzky! 
Let's find a way to detect this situation so that `FeatureNotPresentError` can issue the appropriate recommendation already.

tornaria:
> This is surprising to some extent. Is "pip install --user" disabled in general in Void Linux's system python? Do you have a reference that explains this policy, or more generally what users on Void Linux are expected to be able to do with their system Python? (Are there Void Linux users, or does it follow an "every distro user is a distro developer" model?)

See: https://voidlinux.org/news/2023/10/python3-external.html

But note that this is not a void linux invention: it is recommended in PEP-0668, https://peps.python.org/pep-0668/#recommendations-for-distros

> > ```
> > [... snippet about using a venv or use pipx ...]
> > ```
> 
> What does it have to say about venv? What you snipped may be the interesting part, I think.

The whole message is in the void news post I linked above, but it's roughly speaking what I did.

> As you know, "--without-system-python3" is a configuration option of Sage-the-distribution. It has no relevance for downstream packaging. All I'm saying is: When the Sage distribution sets up its venv, system site packages are normally disabled, and hence user site packages are also disabled.


> 
> > What would take to make the same thing possible with python packages (i.e. have a location inside `~/.sage/, where python packages can be installed for use in sage).
> 
> It wouldn't be difficult, but I think it would be much better to direct users to setting up a suitable venv instead.

Having a location in `~/.sage/` where users can install arbitrary python packages is a recipe for failure (same reason why `pip install --user` is discouraged, I think). But having a location where a curated list of optional sagemath packages (e.g. databases) can be installed, would be great.

I honestly thought that using optional python packages from pypi would be easy. If it's not, 
I'm starting to lean back against moving databases to pypi packages. Just let databases be a directory with files (maybe with some metadata) and have a reasonable mechanism to find the databases in a way that there can be made available as distro packages or in `/usr/local` or in `$HOME`, etc.

mkoeppe:
> > > What would take to make the same thing possible with python packages (i.e. have a location inside `~/.sage/, where python packages can be installed for use in sage).
> > 
> > It wouldn't be difficult, but I think it would be much better to direct users to setting up a suitable venv instead.
> 
> Having a location in `~/.sage/` where users can install arbitrary python packages is a recipe for failure (same reason why `pip install --user` is discouraged, I think).

I fully agree.

> But having a location where a curated list of optional sagemath packages (e.g. databases) can be installed, would be great.
> 
> I honestly thought that using optional python packages from pypi would be easy. 

Let's just fix what makes it not-so-easy. Issuing the venv install instruction is really all it takes.

> If it's not, I'm starting to lean back against moving databases to pypi packages. Just let databases be a directory with files (maybe with some metadata)

I think you are describing a `SAGE_LOCAL`. No, let's please not.

tornaria:
> > But having a location where a curated list of optional sagemath packages (e.g. databases) can be installed, would be great.
> > I honestly thought that using optional python packages from pypi would be easy.

>Let's just fix what makes it not-so-easy. Issuing the venv install instruction is really all it takes.

That's not good enough. I want to "install" the database I need inside `~/.sage/` and forget about it. Then just running `sage` should give me access to the database. And the same instructions should work for sage-the-distro, conda, or distro packages.

> > If it's not, I'm starting to lean back against moving databases to pypi packages. Just let databases be a directory with files (maybe with some metadata)
> 
> I think you are describing a `SAGE_LOCAL`. No, let's please not.

No, I'm not. This is about having a reasonable (small) set of agreed upon locations to install the database packages that suits all common usages (distro package in /usr/share or /usr/share/sagemath, installed by sysadmin in /usr/local/share or /usr/local/share/sagemath, or installed by user in $HOME/.local/share or in $DOT_SAGE). Here /usr is really PREFIX so sage-the-distro would still be using $SAGE_LOCAL/share, etc).

This is a working-now solution. Python packages are not.

mkoeppe:
> > > But having a location where a curated list of optional sagemath packages (e.g. databases) can be installed, would be great.
> > > I honestly thought that using optional python packages from pypi would be easy.
> 
> > Let's just fix what makes it not-so-easy. Issuing the venv install instruction is really all it takes.
> 
> That's not good enough. 
> 
> I want to "install" the database I need inside `~/.sage/` and forget about it. Then just running `sage` should give me access to the database. And the same instructions should work for sage-the-distro, conda, or distro packages.

I disagree. 

First of all, the distinction "database package" vs. "another package" is not useful because it's too arbitrary. Also, databases need versioning just like any other software package.
It's much better to have one method that works both for databases and for other packages on the same system.

If you want uniformity of installation across different deployment methods of Sage -- that's exactly the old SAGE_LOCAL approach.
But we have developed a better way. When we inform users about available system packages, we are deliberately not providing them with a Sage-specific method to install these system packages. Instead we empower users by teaching them how to use the system package tools on their system. (After all, they chose the system that they are using!)

mkoeppe:
Separate from the larger design questions, this fails tests in the conda build:
https://github.com/sagemath/sage/actions/runs/7783168985/job/21221055887?pr=37231#step:10:6143
```
sage -t --random-seed=102093232667646836705241874909477763847 src/sage/matroids/database_collections.py
File "src/sage/matroids/database_collections.py", line 48, in sage.matroids.database_collections.AllMatroids
Failed example:
    for M in matroids.AllMatroids(2):
        M
Exception raised:
      File "/home/runner/work/sage/sage/src/sage/doctest/forker.py", line 712, in _run
      File "/home/runner/work/sage/sage/src/sage/doctest/forker.py", line 1147, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.matroids.database_collections.AllMatroids[0]>", line 1, in <module>
        for M in matroids.AllMatroids(Integer(2)):
      File "/home/runner/work/sage/sage/src/sage/matroids/database_collections.py", line 198, in AllMatroids
        dbfilepath = files(
      File "/usr/share/miniconda3/envs/sage/lib/python3.9/importlib/resources.py", line 147, in files
        return _common.from_package(_get_package(package))
      File "/usr/share/miniconda3/envs/sage/lib/python3.9/importlib/_common.py", line 14, in from_package
        return fallback_resources(package.__spec__)
      File "/usr/share/miniconda3/envs/sage/lib/python3.9/importlib/_common.py", line 18, in fallback_resources
        package_directory = pathlib.Path(spec.origin).parent
      File "/usr/share/miniconda3/envs/sage/lib/python3.9/pathlib.py", line 1082, in __new__
        self = cls._from_parts(args, init=False)
      File "/usr/share/miniconda3/envs/sage/lib/python3.9/pathlib.py", line 707, in _from_parts
        drv, root, parts = self._parse_args(args)
      File "/usr/share/miniconda3/envs/sage/lib/python3.9/pathlib.py", line 691, in _parse_args
        a = os.fspath(a)
    TypeError: expected str, bytes or os.PathLike object, not NoneType
```

Note that as this PR is set to "blocker", it is [applied as a patch on all other PRs](https://github.com/sagemath/sage/wiki/Sage-10.2-Release-Tour#open-blocker-prs-are-applied-automatically-in-ci-workflows), so the conda failure shows up in all other PRs as well.
https://github.com/sagemath/sage/actions/runs/7810776379/job/21304728234?pr=37251#step:10:6143
Because of this mechanism, it's best to set a PR to "blocker" only when it passes its own tests already.
I'm reducing it to "critical" for now, please feel free to set it back when appropriate.

tornaria:
> > > > But having a location where a curated list of optional sagemath packages (e.g. databases) can be installed, would be great.
> > > > I honestly thought that using optional python packages from pypi would be easy.
> > 
> > 
> > > Let's just fix what makes it not-so-easy. Issuing the venv install instruction is really all it takes.
> > 
> > 
> > That's not good enough.
> > I want to "install" the database I need inside `~/.sage/` and forget about it. Then just running `sage` should give me access to the database. And the same instructions should work for sage-the-distro, conda, or distro packages.
> 
> I disagree.

You are welcome to disagree, but please be open to different points of view, and be flexible. The fact that this workflow is not clearly supported is one of the arguments in favor of #37140.

> First of all, the distinction "database package" vs. "another package" is not useful because it's too arbitrary. Also, databases need versioning just like any other software package. It's much better to have one method that works both for databases and for other packages on the same system.

There is a very clear distinction between "code" (especially in python which is very dynamic where monkey-patching is a thing, and with namespace packages) and "support data files".

I think something like I said above is very desirable for databases, particularly for big ones where may be hard to justify a distro package. I see this as a support files that can be installed in different ways, including by the end user.

If something like that can *also* work for general optional packages from pypi, maybe is better (although as a distro packager, I would not be eager to support installing code in such a way, but maybe for a curated list of packages, e.g. databases, without warning, and for arbitrary packages with a fat warning).

> If you want uniformity of installation across different deployment methods of Sage -- that's exactly the old SAGE_LOCAL approach. But we have developed a better way. When we inform users about available system packages, we are deliberately not providing them with a Sage-specific method to install these system packages. Instead we empower users by teaching them how to use the system package tools on their system. (After all, they chose the system that they are using!)

What I say is completely orthogonal to SAGE_LOCAL. All I'm saying is that for some support files (especially very large databases) it should be possible to load them from different locations, including a distro location (/usr or SAGE_LOCAL), a local per-system location (/usr/local) or a end user location (~/.sage/).

Anyway, I discovered that sage *already supports this idea*. That is why `PYTHONUSERBASE` is set:
```
$ sage -pip install --break-system-packages admcycles
Defaulting to user installation because normal site-packages is not writeable
Collecting admcycles
  Using cached admcycles-1.4-py3-none-any.whl (351 kB)
Installing collected packages: admcycles
Successfully installed admcycles-1.4

$ ls ~/.sage/local/lib/python3.12/site-packages/
admcycles  admcycles-1.4.dist-info

$ sage -q
sage: from sage.features import PythonModule
sage: PythonModule("admcycles").is_present()
FeatureTestResult('admcycles', True)


$ python -m admcycles     #### CHECK THAT IT DID NOT MESS WITH MY SYSTEM PYTHON
/usr/bin/python: No module named admcycles
```

So this gives a possible way to move forward.

mkoeppe:
> All I'm saying is that for some support files (especially very large databases) it should be possible to load them from different locations, including a distro location (/usr or SAGE_LOCAL), a local per-system location (/usr/local) or a end user location (~/.sage/).

Users who want to use manually installed stuff can already do so, for example by using ~/.sage/init.sage to add custom directories to their `sys.path` or their `sage.env.SAGE_DATA_PATH`.

I'll also note that currently we don't really have any "very large databases" in Sage. There are some that used to be considered "huge", but by passage of time they are now in "just a few gigabytes" territory.

One simple mechanism for only having one physical copy of a large package in the system is to make it a Python package and use a modern (PEP-660) editable install of it. This is something that we may want to advertise.

> Anyway, I discovered that sage _already supports this idea_. That is why `PYTHONUSERBASE` is set:
> ```
> $ sage -pip install --break-system-packages admcycles
> Defaulting to user installation because normal site-packages is not writeable
> ```
Yes, installing into the `PYTHONUSERBASE` set by `sage-env` does work in this particular use case, where Sage is installed into a system Python environment.
- Should we detect this situation and automatically use `--user --break-system-packages`?

Note that it does not work in a typical installation of the Sage distribution that makes an isolated venv over a system Python -- because disabling system site-packages also disables user site-packages (whether `PYTHONUSERBASE` is set or not).

tornaria:
> I created a `matroid-database` repo on my account and I have made a request for publishing on pypi.

Awesome, I will test it. To summarize my long post above, a very easy way to install into sage:
```
$ git clone https://github.com/gmou3/matroid-database
...
$ cd matroid-database
$ sage -pip install .
...
Successfully installed matroid-database-0.1
```

For the last step, may need to add `--break-system-packages` to the pip options if your python site-packages is externally managed (e.g. running a distro package of sagemath).

Once this is in pypi just doing `sage -pip install matroid-database` should work without the git step.

Now:
```
$ sage -q
sage: from matroid_database import *
sage: for m in all_matroids_bases(5, 2): print(m)
[[0, 1], [0, 2], [1, 2], [0, 3], [1, 3], [2, 3], [0, 4], [1, 4], [2, 4], [3, 4]]
[[0, 2], [1, 2], [0, 3], [1, 3], [2, 3], [0, 4], [1, 4], [2, 4], [3, 4]]
[[0, 2], [1, 2], [0, 3], [1, 3], [0, 4], [1, 4], [2, 4], [3, 4]]
[[1, 2], [1, 3], [2, 3], [1, 4], [2, 4], [3, 4]]
[[0, 3], [1, 3], [2, 3], [0, 4], [1, 4], [2, 4], [3, 4]]
[[0, 3], [1, 3], [2, 3], [0, 4], [1, 4], [2, 4]]
[[1, 3], [2, 3], [1, 4], [2, 4], [3, 4]]
[[1, 3], [2, 3], [1, 4], [2, 4]]
[[2, 3], [2, 4], [3, 4]]
[[0, 4], [1, 4], [2, 4], [3, 4]]
[[1, 4], [2, 4], [3, 4]]
[[2, 4], [3, 4]]
[[3, 4]]
```
:+1:

mkoeppe:
Next step: `sage --package create matroid_database --pypi --type optional`

gmou3:
Should I adapt the present PR?

mkoeppe:
Either that or open a fresh one, it does not matter much

gmou3:
> Next step: `sage --package create matroid_database --pypi --type optional`

I did this and adapted the code to work with the pypi package. However, given that the package is optional, what do I need to do for the tests that rely on it? If I don't install the package manually they all return `ModuleNotFoundError: No module named 'matroid_database'`.

mkoeppe:
> I was able to handle this by using a combination of `sage: # optional - matroid_database` and `# needs matroid_database`

We would use `# optional` here because `matroid_database` is an optional package / feature.
The tag `# needs` is used for standard packages / features.

mkoeppe:
> I think you need to add `PythonModule('matroid_database')` to the list at the end of `src/sage/features/databases.py`. I think that's also necessary for the `# needs matroid_database` tag to work.

That's right. Without declaring the feature, the `# optional - matroid_database` would only work when `matroid_database` is installed as an SPKG, i.e., using `make matroid_database` or `sage -i matroid_database`. But we want it to work also in other deployment situations.

src/sage/features/databases.py
```diff
@@ -263,6 +263,7 @@ def __init__(self, name='polytopes_db'):
 
 def all_features():
     return [PythonModule('conway_polynomials'),
+            PythonModule('matroid_database'),
```
mkoeppe:
```suggestion
    return [PythonModule('conway_polynomials', spkg='conway_polynomials', type='standard'),
            PythonModule('matroid_database', spkg='matroid_database'),
```

src/setup.cfg.m4
```diff
@@ -107,27 +107,26 @@ sage.repl.rich_output =
 
 sage =
     ext_data/*
-    ext_data/kenzo/*
```
mkoeppe: I would suggest to back out the cosmetic change (alphabetic sorting) from this PR because it will conflict with #36951

build/pkgs/matroid_database/SPKG.rst
```diff
@@ -0,0 +1,16 @@
...
+======================================================
+
+Description
+-----------
+
+Python interface to matroid database
+
+License
+-------
+
```
mkoeppe, line 16: please add license and more detailed description -- for example by fixing the metadata of the package on PyPI and re-running `sage --package create matroid_database --pypi`

src/sage/features/databases.py
```diff
@@ -262,7 +262,8 @@ def __init__(self, name='polytopes_db'):
 
 
 def all_features():
-    return [PythonModule('conway_polynomials'),
+    return [PythonModule('conway_polynomials', spkg='conway_polynomials', type='standard'),
+            PythonModule('matroid_database', spkg='matroid_database'),
```
mkoeppe: I'd suggest to make this a separate feature class `DatabaseMatroids`.

Then, just before you `import matroid_database`, use `DatabaseMatroids().require()`. This will give instructions to users how to install it.

build/pkgs/matroid_database/SPKG.rst
gmou3: I did so manually.

mkoeppe:
This is working well. Let's merge it.

PR #37232, opened 2024-02-03
Still less use of isinstance of old-parents

fchapoton:
this is removing many calls to isinstance of old-style parent-like classes

one should use instead the corresponding containment in categories

the case of `Finitefield` is kept for another time

### :memo: Checklist


src/sage/modular/abvar/homology.py
```diff
@@ -56,7 +56,7 @@
 from sage.rings.integer import Integer
 from sage.rings.integer_ring import ZZ
 from sage.rings.rational_field import QQ
-from sage.rings.ring import CommutativeRing
+from sage.categories.commutative_rings import CommutativeRings
```
mkoeppe: best to keep imports sorted when they are sorted

mantepse (approved):
cool, thank you!

PR #37233, opened 2024-02-03
fix small mistake in BQFClassGroupQuotientMorphism documentation

yyyyx4:
The possibility to pass discriminants to the constructor instead of `BQFClassGroup` objects was present for an earlier version of #37074, but it was removed during review. However, it seems I forgot to update the docstring accordingly.


tscrim (approved):
LGTM

PR #37234, opened 2024-02-03
make a category of Dedekind domains, remove code from ring.pyx

fchapoton:
This is creating a `DedekindDomains` category and moving the associated code from `ring.pyx` to the category.

### :memo: Checklist


fchapoton:
maybe one should make ZZ belong to this category ?

src/sage/categories/dedekind_domains.py
```diff
@@ -0,0 +1,155 @@
...
+                sage: K.<i> = QuadraticField(-1)
+                sage: R = K.order(2*i); R
+                Order of conductor 2 generated by 2*i
+                 in Number Field in i with defining polynomial x^2 + 1 with i = 1*I
+                sage: R.is_maximal()
+                False
+                sage: R.krull_dimension()
+                1
+            """
+            return 1
```
tscrim:
```suggestion
            return ZZ.one()
```
To avoid Python `int`s getting out.

tscrim (changes requested):
One tiny thing (it should be fixed more broadly though). However, I think it is a good thing to refine the category of `ZZ` and remove duplicate methods (these should never be a critical part of any algorithm IMO, so downgrading Cython -> Python shouldn't have any real effect...I hope).

fchapoton:
It seems that ZZ prefers to inherit `krull_dimension` from `rings/ring.pyx`. Same for `is_integrally_closed`. Not sure how to fix that.

tscrim:
Ah, right, categories have the lowest precedence. Yes, we can revert removing them. It was just trying to avoid some redundancies.

tscrim (changes requested):
Thank you.

One last thing, but unfortunately it is somewhat annoying. We have an import of `DedekindDomain` in `rings/all.py`, which makes its way into the global namespace. So we need to properly formally deprecate the class. *sigh*

tscrim (approved):
Thank you. Once the bot shows tests pass, you can set a positive review.

PR #37242, opened 2024-02-05
Improved algorithm choice for isogeny computation

r98inver:
Improved the logic for algorithm choice in `EllipticCurve_field.isogeny()`:
- if we have multiple points, `factored` is used;
- if the order is known and composite, `factored` is used;
- if the order is known and (pseudo-)prime, we chose between the traditional Velù formulae and Velù-sqrt, depending on a parameter; this parameter may be supplied by the user with the `velu_sqrt_bound` argument, and if `None` a global parameter is used (see below);
- in all the other cases we fall back to the traditional isogeny algorithm.

The new `velu_sqrt_bound`:
- can be supplied as a parameter when calling `.isogeny()`;
- if supplied, is recursively passed to all calls to `factored`, such that each single isogeny computation has the same bound;
- if not supplied, a global value is used instead; this is stored in the object `_velu_sqrt_bound` from `sage.schemes.elliptic_curves.hom_velusqrt`, and is initially set to 1000 (due to empirical observations) but can be manually set at runtime by the user

#sd123


src/sage/schemes/elliptic_curves/ell_curve_isogeny.py
```diff
@@ -867,9 +867,9 @@ class EllipticCurveIsogeny(EllipticCurveHom):
 
         sage: E = EllipticCurve(j=GF(7)(0))
         sage: phi = E.isogeny([E(0), E((0,1)), E((0,-1))]); phi
-        Isogeny of degree 3
-         from Elliptic Curve defined by y^2 = x^3 + 1 over Finite Field of size 7
-           to Elliptic Curve defined by y^2 = x^3 + 1 over Finite Field of size 7
+        Composite morphism of degree 3 = 3:
```
GiacomoPope: Although this is fine, and obviously works when it's composite, I wonder if we should have a check in the format which does:

```
Composite morphism of degree 3
```

When it's prime and this format when it's composite? This would be a modification of `_repr_` for the composite morphism... 

```
Composite morphism of degree 9 = 3^2:
```

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1349,7 +1454,41 @@ def isogeny(self, kernel, codomain=None, degree=None, model=None, check=True, al
...
+                # Single point on the curve; unpack the list for compatibility with velusqrt
+                if kernel_is_list:
+                    kernel = kernel[0]
+
+                known_order = False
+                try:
+                    single_point_order = kernel._order
+                    known_order = True
+                except AttributeError:
+                    pass
```
GiacomoPope:
```suggestion
                known_order = hasattr(kernel, "_order")
```

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1349,7 +1454,41 @@ def isogeny(self, kernel, codomain=None, degree=None, model=None, check=True, al
...
+
+                if known_order and single_point_order.is_pseudoprime():
+                    if not velu_sqrt_bound:
+                        from sage.schemes.elliptic_curves.hom_velusqrt import _velu_sqrt_bound
+                        velu_sqrt_bound = _velu_sqrt_bound.get()
+
+                    if single_point_order > velu_sqrt_bound:
```
GiacomoPope:
```suggestion
                    if kernel._order > velu_sqrt_bound:
```

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1349,7 +1454,41 @@ def isogeny(self, kernel, codomain=None, degree=None, model=None, check=True, al
...
+                    kernel = kernel[0]
+
+                known_order = False
+                try:
+                    single_point_order = kernel._order
+                    known_order = True
+                except AttributeError:
+                    pass
+
+                if known_order and single_point_order.is_pseudoprime():
```
GiacomoPope:
```suggestion
                if known_order and kernel._order.is_pseudoprime():
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -147,6 +147,10 @@ def _compute_factored_isogeny_prime_power(P, l, n, split=.8):
       multiplications according to the parameter.
       The asymptotic complexity is `O(n \log(n) \ell)`.
 
+    The optional parameter``velu_sqrt_bound`` prescribes the
```
GiacomoPope:
```suggestion
    The optional parameter ``velu_sqrt_bound`` prescribes the
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -219,12 +224,16 @@ def rec(Q, k):
...
 
 
-def _compute_factored_isogeny_single_generator(P):
+def _compute_factored_isogeny_single_generator(P, velu_sqrt_bound=None):
     """
     This method takes a point `P` and returns a sequence of
     prime-degree isogenies whose composition has the subgroup
     generated by `P` as its kernel.
 
+    The optional parameter``velu_sqrt_bound`` prescribes the
```
GiacomoPope:
```suggestion
    The optional parameter ``velu_sqrt_bound`` prescribes the
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -236,23 +245,41 @@ def _compute_factored_isogeny_single_generator(P):
...
 
 
-def _compute_factored_isogeny(kernel):
+def _compute_factored_isogeny(kernel, velu_sqrt_bound=None):
     """
     This method takes a set of points on an elliptic curve
     and returns a sequence of isogenies whose composition
     has the subgroup generated by that subset as its kernel.
 
+    The optional parameter``velu_sqrt_bound`` prescribes the
```
GiacomoPope:
```suggestion
    The optional parameter ``velu_sqrt_bound`` prescribes the
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -289,6 +316,13 @@ def __init__(self, E, kernel, codomain=None, model=None):
         The ``codomain`` and ``model`` parameters have the same meaning
         as for :class:`EllipticCurveIsogeny`.
 
+        The optional parameter``velu_sqrt_bound`` prescribes the point
```
GiacomoPope:
```suggestion
        The optional parameter ``velu_sqrt_bound`` prescribes the point
```

GiacomoPope (commented):
Some small suggestions throughout and a comment about the representation of "composite but prime degree isogenies".

Very nice work, thank you for implementing this.

r98inver:
Thank you! I applied all your suggestions (my `try`-`except` was actually taken from `ell_point.py`, so I modified that as well)

GiacomoPope:
I was about to mark this as completed, but somehow `config.cache` has ended up in this commit? I'm not sure why or how?

r98inver:
Sorry, I keep messing up with git. I deleted it with a new commit, is that fine? Or should I do something different?

GiacomoPope (approved):
Great addition to the isogeny functionality. Thanks for also fixing the small clean up with ell_point. Appreciated.

r98inver:
Thanks!

vbraun:
Test failure, see the CI at https://github.com/sagemath/sage/actions/runs/7805562652

r98inver:
Fixed, I also noticed another failure in the coding style so I fixed that one as well.

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1245,8 +1245,36 @@ def isogeny(self, kernel, codomain=None, degree=None, model=None, check=True, al
...
+
+              - If the order is known and composite, ``"factored"`` is selected.
+
+              - If the order is known and prime, a choice between ``"velusqrt"`` and
+                ``"traditional"`` is done according to the ``velu_sqrt_bound``
+                parameter (see below).
+
+          If none of the previous apply, ``"traditional"`` is selected.
+
+        - ``velu_sqrt_bound``: -- an integer (default: ``None``). Establish the highest
```
grhkm21:
```suggestion
        - ``velu_sqrt_bound`` -- an integer (default: ``None``). Establish the highest
```

src/sage/schemes/elliptic_curves/hom_velusqrt.py
```diff
@@ -135,6 +135,35 @@
...
+    def __init__(self):
+        self.bound = Integer(1000)
+
+    def set(self, b):
+        self.bound = b
+
+    def get(self):
+        return self.bound
+
+    def __str__(self):
```
grhkm21: Should this be `__repr__` instead?

r98inver: added both

src/sage/schemes/elliptic_curves/ell_field.py
r98inver: fixed

src/sage/schemes/elliptic_curves/hom_velusqrt.py
grhkm21: The __str__ method is automatically implemented (more precisely it calls __repr__ if it is not implemented.)

r98inver: ok didn't know that, fixed

grhkm21:
If CI passes then it's all good. This already has `s: positive review`.

PR #37245, opened 2024-02-06
a few details in combinat folder

fchapoton:
namely

- use ruff to fix some containement tests

- add links to some error codes in the doc

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/designs/difference_family.py
+++ b/src/sage/combinat/designs/difference_family.py
@@ -3138,8 +3138,9 @@ def difference_family(v, k, l=1, existence=False, explain_construction=False, ch
 
     See also :wikipedia:`Difference_set`.
 
-    If there is no such difference family, an ``EmptySetError`` is raised and if
-    there is no construction at the moment ``NotImplementedError`` is raised.
+    If there is no such difference family, an ``EmptySetError`` is raised and
+    if there is no construction at the moment :class:`NotImplementedError`
+    is raised.
 
     INPUT:
 
--- a/src/sage/combinat/designs/orthogonal_arrays.py
+++ b/src/sage/combinat/designs/orthogonal_arrays.py
@@ -190,7 +190,7 @@ def transversal_design(k, n, resolvable=False, check=True, existence=False):
         Unknown
 
     If you ask for a transversal design that Sage is not able to build then an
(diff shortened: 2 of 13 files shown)


src/sage/combinat/designs/difference_family.py
```diff
@@ -3138,8 +3138,9 @@ def difference_family(v, k, l=1, existence=False, explain_construction=False, ch
 
     See also :wikipedia:`Difference_set`.
 
-    If there is no such difference family, an ``EmptySetError`` is raised and if
-    there is no construction at the moment ``NotImplementedError`` is raised.
+    If there is no such difference family, an ``EmptySetError`` is raised and
```
dcoudert, line 3141: shouldn't ``EmptySetError`` be ``` :class:`EmtySetError` ``` ?

src/sage/combinat/designs/orthogonal_arrays.py
```diff
@@ -2117,8 +2117,8 @@ class OAMainFunctions:
         6
 
     If you ask for an orthogonal array that does not exist, then you will
-    either obtain an ``EmptySetError`` (if it knows that such an orthogonal array
-    does not exist) or a ``NotImplementedError``::
+    either obtain an ``EmptySetError`` (if it knows that such an orthogonal
```
dcoudert, line 2120: Same here

src/sage/combinat/designs/difference_family.py
fchapoton: This is not a python standard exception, so we cannot link to Python doc.

dcoudert (approved):
Thanks for the explanation.

LGTM.

src/sage/combinat/designs/steiner_quadruple_systems.py
```diff
@@ -709,7 +709,7 @@ def steiner_quadruple_system(n, check=False):
         ....:         sqs = designs.steiner_quadruple_system(n, check=True)
     """
     n = int(n)
-    if not ((n % 6) in [2, 4]):
+    if (n % 6) not in [2, 4]:
```
grhkm21, line 712: (Not mandatory but) should this check be moved to the bottom of the `if`-`else` chain? Especially since there is an error at the end that says "this should never happen".

PR #37251, opened 2024-02-07
ntl 11.5.1

mkoeppe:
### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37252


kwankyu:
Which CI workflow really tests the new ntl? On all platforms that I looked into, ntl is already available from the system or from the previous install.

mkoeppe:
(Should we add something like this to the developer guide?)

kwankyu:
> (Should we add something like this to the developer guide?)

Perhaps no. As you explained, I was confused by the first message in the minimal build workflow. Your explanation seems too much detail to be included in the developer guide.

I tested the package on my system with "--with-system-ntl=no". We may give this tip somewhere in the developer manual, if this is not obvious to others. 

For a package upgrade PR, I imagine Build & Test workflow running with ""--with-system-xxx=no". But I don't know if this is doable...

kwankyu (approved):
It works well. LGTM

mkoeppe:
> For a package upgrade PR, I imagine Build & Test workflow running with ""--with-system-xxx=no". But I don't know if this is doable...

I hope to merge "Build & Test" with "CI Linux Incremental" soon. (Follow-up after #36498)

mkoeppe:
Thanks for the review!

PR #37257, opened 2024-02-08
`sage.groups.generic`: Fix incorrect identity testing

grhkm21:
The `bsgs` function used to check `is_zero`, when it should check `== identity` (which is given). It also allows me to use the method for "ad-hoc classes":

```sage
sage: # modulo arithmetic lmao
....: # I'll just carry around the modulus
....: n = int(12)
....: G_id = (int(0), n)
....: G_gen = (int(1), n)
....: def G_op(P, Q):
....:     return (int((P[0] + Q[0]) % P[1]), P[1])
....: def G_inv(P):
....:     return (int(-P[0] % P[1]), P[1])
....: P = (int(5), n)
....: 
....: from sage.groups.generic import discrete_log
....: discrete_log(P, G_gen, operation=None, ord=n, op=G_op, identity=G_id, inverse=G_inv)
```

Also fixed styling, that's why there are some random edits

### DIFF
--- a/src/sage/groups/generic.py
+++ b/src/sage/groups/generic.py
@@ -119,7 +119,6 @@ from copy import copy
 
 from sage.arith.misc import integer_ceil, integer_floor, xlcm
 from sage.arith.srange import xsrange
-from sage.misc.functional import log
 from sage.misc.misc_c import prod
 import sage.rings.integer_ring as integer_ring
 import sage.rings.integer
@@ -171,9 +170,11 @@ def multiple(a, n, operation='*', identity=None, inverse=None, op=None):
         sage: E = EllipticCurve('389a1')
         sage: P = E(-1,1)
         sage: multiple(P, 10, '+')
-        (645656132358737542773209599489/22817025904944891235367494656 : 525532176124281192881231818644174845702936831/3446581505217248068297884384990762467229696 : 1)
+        (645656132358737542773209599489/22817025904944891235367494656 :
+         525532176124281192881231818644174845702936831/3446581505217248068297884384990762467229696 : 1)
         sage: multiple(P, -10, '+')
-        (645656132358737542773209599489/22817025904944891235367494656 : -528978757629498440949529703029165608170166527/3446581505217248068297884384990762467229696 : 1)
+        (645656132358737542773209599489/22817025904944891235367494656 :
+         -528978757629498440949529703029165608170166527/3446581505217248068297884384990762467229696 : 1)
     """
     from operator import inv, mul, neg, add
 
@@ -332,11 +333,11 @@ class multiples:
                 P0 = P.parent()(0)
             self.op = add
         else:
-            self.op = op
             if P0 is None:
                 raise ValueError("P0 must be supplied when operation is neither addition nor multiplication")
             if op is None:
                 raise ValueError("op() must both be supplied when operation is neither addition nor multiplication")
+            self.op = op
 
         self.P = copy(P)
         self.Q = copy(P0)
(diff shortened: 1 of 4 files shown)


src/sage/groups/generic.py
```diff
@@ -458,6 +459,7 @@ def bsgs(a, b, bounds, operation='*', identity=None, inverse=None, op=None):
         inverse = inv
         op = mul
     elif operation in addition_names:
+        # Should this be replaced with .zero()? With an extra AttributeError handler?
         identity = a.parent()(0)
```
yyyyx4: I think `.zero()` should be faster, but indeed not all additive groups have a `.zero()` method...

src/sage/groups/generic.py
```diff
@@ -479,7 +481,7 @@ def bsgs(a, b, bounds, operation='*', identity=None, inverse=None, op=None):
 
     if ran < 30:    # use simple search for small ranges
         d = c
-#        for i,d in multiples(a,ran,c,indexed=True,operation=operation):
+        #        for i,d in multiples(a,ran,c,indexed=True,operation=operation):
```
yyyyx4:
```suggestion
        # for i,d in multiples(a,ran,c,indexed=True,operation=operation):
```
Did you mean this? Not sure about this change either way: I prefer the `#` at the beginning of the line for commenting things out. (This is a matter of taste.)

src/sage/groups/generic.py
```diff
@@ -1257,14 +1264,23 @@ def order_from_multiple(P, m, plist=None, factorization=None, check=True,
...
+                        inverse=inverse,
+                        op=op)
 
     if P == identity:
         return Z.one()
 
     M = Z(m)
     if check:
-        assert multiple(P, M, operation=operation) == identity
+        assert _multiple(P, M) == identity
```
yyyyx4: Hm, this should probably be a `ValueError` since it can be triggered by invalid user input, right?

src/sage/groups/generic.py
grhkm21: Ah I might have formatted that accidentally *blames editor*. I'll keep the new change since I prefer it :D

src/sage/groups/generic.py
grhkm21: I took the violent approach of replacing it with `.zero()` and fixing the errors. I only found a few use cases of `sage.groups.generic` in the entire Sage, so I fixed those (only required fixing `homset::zero`)

yyyyx4: Thanks, looks promising! Just double-checking: The change from `.codomain()` to `.extended_codomain()` in `src/sage/schemes/generic/homset.py` is an essentially unrelated bugfix, right? And this is what causes the point `(a : a)` in that doctest in `src/sage/schemes/projective/projective_morphism.py` to be normalized properly?

grhkm21: Yes they're two unrelated fixes (sorry!!), the `extended_codomain` is to imitate what `schemes/projective/projective_homset.py` does (and makes more sense). The normalisation is a separate problem.

It is caused by the following stripped code:

```sage
sage: F.<a> = GF(4)
....: ProjectiveSpace(ZZ, 1)._point(S(F), [a, a])
....: ProjectiveSpace(F, 1)._point(S(F), [a, a])
(a : a)
(1 : 1)
```
(This caller difference is introduced by the `extended_codomain` to `codomain` change.) This is because the two lines call different constructors

https://github.com/sagemath/sage/blob/0c390a06f725cfeef11fb7f8430a0f451250af6d/src/sage/schemes/projective/projective_space.py#L823-L836 callls `SchemeMorphism_point_projective_ring`.

https://github.com/sagemath/sage/blob/0c390a06f725cfeef11fb7f8430a0f451250af6d/src/sage/schemes/projective/projective_space.py#L2271-L2284 calls `SchemeMorphism_point_projective_finite_field` which allows division.

I think this is a separate problem, where `._point` shouldn't care about the caller's type but rather the argument (homset)'s type, but again it's a separate problem. It's a separate problem.

yyyyx4 (approved):
Looks good, thanks!

PR #37208, opened 2024-01-31
Fix kernel basis computation for polynomial matrix in corner case

vneiger:
For univariate polynomial matrices, either with zero columns (or zero rows if `row_wise=False`), or which are zero, the computed minimal kernel basis was of the sparse type (due to a mistake in using the identity method). This is fixed.

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

PR #37201  modified a test in the same file `matrix_polynomial_dense.pyx`

PR #37152, opened 2024-01-23
provide Q.log(P) instead of P.discrete_log(Q) for elliptic-curve points

yyyyx4:
Here's an attempt to resolve #37150. I'm not suggesting to remove `.discrete_log()` any time soon, but I do feel this inconsistency should be corrected *at some point*, so we may as well start with it now.

Cc: @JohnCremona

PR #37146, opened 2024-01-23
Fix a bug in reduction of element modulo ideal in exterior algebra

trevorkarn:
This adds a loop to continue doing reduction modulo a Grobner basis for an element modulo an ideal in the exterior algebra. This guarantees a unique representative for each element, and that representative will be $0$ if the element is in the ideal. This fixes #37108.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

### :hourglass: Dependencies

None.

PR #37132, opened 2024-01-22
Fix corners cases of free resolutions

kwankyu:
as seen in
```sage
sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
sage: S = P2.coordinate_ring()
sage: I = S.ideal(0)
sage: C = I.graded_free_resolution(); C
S(0) <-- 0
sage: C[1]
Ambient free module of rank 0 over the integral domain Multivariate Polynomial Ring in x, y, z over Rational Field
sage: C[0]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 C[Integer(0)] 

File ~/GitHub/sage/src/sage/homology/free_resolution.py:426, in FiniteFreeResolution.__getitem__(self, i)
    424     F = FreeModule(self._base_ring, 0)
    425 elif i == self._length:
--> 426     F = FreeModule(self._base_ring, self._maps[i - 1].ncols())
    427 else:
    428     F = FreeModule(self._base_ring, self._maps[i].nrows())

IndexError: list index out of range
sage: C.differential(1)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 C.differential(Integer(1))  

File ~/GitHub/sage/src/sage/homology/free_resolution.py:489, in FiniteFreeResolution.differential(self, i)
    487 elif i == self._length + 1:
    488     s = FreeModule(self._base_ring, 0)
--> 489     t = FreeModule(self._base_ring, self._maps[i - 2].ncols())
    490     m = s.hom(0, t)
    491 elif i > self._length + 1:

IndexError: list index out of range
```

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37262, opened 2024-02-08
Added automatic size labeler

### LINKED ISSUE #37254: CI: Add pull request size labeler / checker
There are a few solutions on GH Actions Marketplace that we may want to use:
- https://github.com/marketplace/actions/pull-request-auto-size-labeler
- https://github.com/marketplace/actions/pr-size-watcher
- https://github.com/marketplace/actions/pr-size-checker

Possible positive effects:
- A size label "tiny" could encourage quick reviews of trivial changes.
- A size label "huge" could help flag problematic PRs.

amanmoon:
I have implemented the `automatic size labeler`, which now assigns labels to pull requests based on the number of lines changed

**Minimal**
Typically involves very small changes, bug fixes, or updates that require only a few lines of code, often less than 50.
#37208 #37146 #37043

**Small**
Involves more substantial changes than minimal, potentially adding new features or making modifications to existing ones. The range is usually between 50 to 100 lines of code.
#37152 #37132

**Moderate**
Represents a significant portion of the codebase being modified, such as adding new features, refactoring, or making extensive changes to existing functionalities. This might involve between 100 to 300 lines of code.
#36919 #37112

**Large**
Involves substantial and complex changes across various parts of the codebase. This could include major architectural changes, the introduction of new modules, or a significant overhaul of existing features, often exceeding 300 lines of code.
#37125 #36977 #36972 

### :memo: Checklist

Fixes: #37254


amanmoon:
there might be better way of getting no of lines changed than doing 
`git diff --stat ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | awk '{changes = strtonum($4) + strtonum($6)} END {print changes}'`

mkoeppe:
> If the changes exceed 200 lines (insertion + deletions), the label will be set to `huge`; otherwise, it will be labeled as `tiny`.

That's a pretty dramatic transition from "tiny" to "huge" in 1 line... 
This may need a bit more thought?

amanmoon:
I was thinking something like
Minimal - less than 50 lines of changes
Small - between 50 to 100 lines of changes
Moderate - between 100 to 300 lines
Large - more than 300 lines of changes

mkoeppe:
It may be useful to illustrate such a proposal by putting examples of PRs of such sizes in the description, to facilitate discussion.

Also consider whether labels are necessary for the sizes in the middle. Labels for the smallest category and for the  largest category have clear functions (which?)

amanmoon:
We can also categorize based on the type of changes made. 
`Tiny` would include documentation, cleanup, refactoring, and small bug fixes (categorize based on size?). 
`Huge` is reserved for large Cython/Python code changes (again based on size?).

soehms:
First some questions:

1. What was the reason for not using the GH actions from the marketplace suggested at https://github.com/sagemath/sage/issues/37254?
2. Shouldn't we group the new labels with a prefix, say `v: ` (for volume), as we did for example for status and priority labels (`s: `, `p: `)?
3. Should it be allowed to add or remove the new labels manually?
4. We already have a bot to sync groups of labels that have a common meaning. It comes as a Python script [.github/sync_labels](https://github.com/sagemath/sage/blob/develop/.github/sync_labels.py) which is called from [.github/workspaces/sync_labes.yml](https://github.com/sagemath/sage/blob/develop/.github/workflows/sync_labels.yml). Wouldn't it make sense to integrate the new functionality there? Then automatic label manipulation would be concentrated in one place and we wouldn't waste resources by launching GitHub workflows twice.

amanmoon:
> What was the reason for not using the GH actions from the marketplace suggested at https://github.com/sagemath/sage/issues/37254? 

Not using the GitHub Action will give us flexibility. I was thinking using regex, we can skip blank lines and also exclude changes in documentation from counting towards the final number of lines changed.

> Should it be allowed to add or remove the new labels manually?

I don't know how can we do this, hitting Github label's endpoint won't return who added those labels.
We can define another set of labels which can only be added manually and if these are detected it will stop the Github label bot.
There might be some better solution for this.
 
>Shouldn't we group the new labels with a prefix, say v:  (for volume), as we did for example for status and priority labels (s: , p: )?
We already have a bot to sync groups of labels that have a common meaning. It comes as a Python script [.github/sync_labels](https://github.com/sagemath/sage/blob/develop/.github/sync_labels.py) which is called from [.github/workspaces/sync_labes.yml](https://github.com/sagemath/sage/blob/develop/.github/workflows/sync_labels.yml). Wouldn't it make sense to integrate the new functionality there? Then automatic label manipulation would be concentrated in one place and we wouldn't waste resources by launching GitHub workflows twice.

I'll address these changes on my next commit.

>Did you notice, the issue with the GitHub web-interface described in [this sage-devel post](https://groups.google.com/g/sage-devel/c/sulCa-6EZRA/m/WsEW35CZAAAJ)?

This worked fine, I tested the code on one of my repo.

soehms:
> > What was the reason for not using the GH actions from the marketplace suggested at #37254?
> 
> Not using the GitHub Action will give us flexibility. I was thinking using regex, we can skip blank lines and also exclude changes in documentation from counting towards the final number of lines changed.
> 

Yes, flexibility is a good point, but I'm not sure we should exclude changes in documentation.


> > Should it be allowed to add or remove the new labels manually?
> 
> I don't know how can we do this, hitting Github label's endpoint won't return who added those labels. We can define another set of labels which can only be added manually and if these are detected it will stop the Github label bot. There might be some better solution for this.
> 

The `sync_labels` bot posts a warning comment that is deleted after a few minutes if the user changes a status or priority label in a non-compliant manner. Maybe we can do something similar here.

> > Shouldn't we group the new labels with a prefix, say v:  (for volume), as we did for example for status and priority labels (s: , p: )?
> > We already have a bot to sync groups of labels that have a common meaning. It comes as a Python script [.github/sync_labels](https://github.com/sagemath/sage/blob/develop/.github/sync_labels.py) which is called from [.github/workspaces/sync_labes.yml](https://github.com/sagemath/sage/blob/develop/.github/workflows/sync_labels.yml). Wouldn't it make sense to integrate the new functionality there? Then automatic label manipulation would be concentrated in one place and we wouldn't waste resources by launching GitHub workflows twice.
> 
> I'll address these changes on my next commit.
> 

Great!

> > Did you notice, the issue with the GitHub web-interface described in [this sage-devel post](https://groups.google.com/g/sage-devel/c/sulCa-6EZRA/m/WsEW35CZAAAJ)?
> 
> This worked fine, I tested the code on one of my repo.

I think it would be less critical here than with state labels, but we should keep attention on this.

FYI: I won't reply for a week as I'm on vacation.

soehms (commented):
Sorry! Maybe I gave a bad hint. I expected that the logic could be included in the Python file, but I overlooked the fact that `sync_labels` doesn't checkout the entire repository (for performance reasons), which is what is needed here. Therefore, I'm no longer sure whether it makes sense to include the code in `sync_labels.yml`. I hope that reverting to the previous version won't cause you too much trouble.

However, I tested your code a bit in my fork https://github.com/soehms/sage/pull/13. So far things are looking good to me. But maybe you should consider making the constants (`SMALL_THRESHOLD` ...) changeable via repository variables.

amanmoon:
Sure!

soehms:
> can we integrate #37373 here?

Sounds reasonable! If you have the resources to do it, please do it!

I wouldn't let the job fail if the repository variables are not set. It's better to just skip it. Otherwise, after the PR is merged but the variables are not set, every open PR would have at least one failed job after each push.

Some notes on code style and structure:

1. I think you should use environment variables for repeatedly used literals (like `v: minimal`...).
2. You should consider implementing more complicated logic in a Python script (or at least a shell script) called in the `yml`. This would make testing much easier.

soehms:
> I wouldn't let the job fail if the repository variables are not set. 

However, even if they are set the job fails (see [this run](https://github.com/soehms/sage/actions/runs/8253622534/job/22575918603?pr=14#step:3:64) in https://github.com/soehms/sage/pull/14).

amanmoon:
Sorry for responding late, i had my midsems.
>You should consider implementing more complicated logic in a Python script (or at least a shell script) called in the yml. This would make testing much easier.

I tried implementing this but we'll have to access the `CHANGES` variable in the shell script, i tried this but it was throwing series of error.
I think the script is not that big so making different file won't be needed.

soehms:
> I tried implementing this but we'll have to access the CHANGES variable in the shell script, i tried this but it was throwing series of error.

It's not clear to me what caused your trouble and I can't believe there were unsolvable problems.

> I think the script is not that big so making different file won't be needed.

This is not just because of the size of the script. More importantly, maintenance will become easier. Once the workflow is burdened by a large number of differently structured branches, we may need to tune the regular expressions in your `grep` and `sed` statements to accommodate special cases. So if you can test the calculation of changed lines locally that would be very helpful.

So it would be nice to have at least this part in a shell scrpit, for example a script named `get_changed_lines.sh` in folder `.github` (although I assume your whole *Add labels based on size* step could be moved into a shell script ). To be clearer about what I mean, your `yml` file could look like this:


```
...
      - name: add labels based on size
        run: |
          git fetch origin ${{ github.base_ref }}

          CHANGES=`.github/get_changed_lines.sh ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }}`
          echo "CHANGES MADE: $CHANGES"

          PR_NUMBER="${{ github.event.number }}"
...
```

and the script `get_changed_lines.sh` like this:

```
#!/usr/bin/env bash

BASE_SHA=$1
HEAD_SHA=$2

# get all the changes made and changed files
CHANGES=$(git diff --ignore-all-space $BASE_SHA $HEAD_SHA)
CHANGED_FILES=$(git diff --name-only  $BASE_SHA $HEAD_SHA| awk -F'src/sage/' '{print $2}' | cut -d'/' -f1 | sed 's/\([^ ]\+\)/c: \1/g')
...
CHANGES=$(echo "$CHANGES" | wc -l)
echo $CHANGES
```


Further comments: If the threshold variables are not set, your current implementation sets the `v: minimal` label. Perhaps this may cause confusion. In order to also save resources I would add a line as shown below to skip the job completely in such a case:

```
...
jobs:
  label-changes:
    if: vars.SMALL_THRESHOLD && vars.MODERATE_THRESHOLD && vars.LARGE_THRESHOLD
    runs-on: ubuntu-latest
...
```

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,85 @@
...
+          echo "CHANGES MADE: $CHANGES"
+
+          PR_NUMBER="${{ github.event.number }}"
+          API_URL="https://api.github.com/repos/${{ github.repository }}/issues/${PR_NUMBER}/labels"
+          AUTH_HEADER="Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}"
+
+          DELETE_LABELS=("v: minimal" "v: small" "v: moderate" "v: large")
+
+          if [ "$CHANGES" -gt "$SMALL_THRESHOLD" ]; then
+            SIZE_LABEL="v: small"
```
soehms: You missed that point from https://github.com/sagemath/sage/pull/37262#issuecomment-1992233557:

> I think you should use environment variables for repeatedly used literals (like v: minimal...).

soehms:
> I have also tested the code https://github.com/amanmoon/sage/pull/1

Thanks! It looks quite good, now. One questions about `CHANGED_DIR`: You restricted it to sub-folders of `src/sage`. This excludes labels like `c: build`. On the other hand your implementation may lead to lables which currently don't exist (for example `data_structure`). Is this intentional?

Further comments:

- I would move the declaration of the literal constants `API_URL`, `MINIMAL`, `SMALL`, `MODERATE` and `LARGE` to the shell script, since they are just used there.
- The name `get_changes.sh` no longer matches its content, as the label setting is now also included. Suggestion for a new name: `set_labels_by_changes.sh`.
- I would add more `echo` statements for debug purpose, for example logging values of interesting variables like `CHANGES` that you set (even on intermediate steps) and one right at the beginning:

```
echo "set_labels_by_changes` called with environment: $PR_BASE_SHA $PR_HEAD_SHA $LARGE_THRESHOLD $MODERATE_THRESHOLD $SMALL_THRESHOLD"
```

amanmoon:
>>You restricted it to sub-folders of src/sage. This >>excludes labels like c: build. On the other hand your implementation may lead to lables which currently don't exist (for example >>data_structure). Is this intentional?

On second thought, I think we should consider some different implementation as we have multiple labels targeting changes in the same directory, and also the author can make changes in multiple directories which will cause unnecessary labels to be added.

amanmoon:
Now when the PR is draft the action won't run.
[test](https://github.com/amanmoon/sage/pull/7)

soehms:
> On second thought, I think we should consider some different implementation as we have multiple labels targeting changes in the same directory, and also the author can make changes in multiple directories which will cause unnecessary labels to be added.

Perhaps we should use a fixed dictionary which maps directories to labels.

> Now when the PR is draft the action won't run.

Yes, that makes sense!

> I also removed the labeling based on the files changed.

Does this mean that you will do #37373 in another PR, now?

amanmoon:
>Does this mean that you will do https://github.com/sagemath/sage/issues/37373 in another PR, now?

Yes

> also the author can make changes in multiple directories which will cause unnecessary labels to be added.

this still remains an issue, I will try to find a solution soon and create a PR.

amanmoon:
>>Perhaps we should use a fixed dictionary which maps directories to labels.

Yes

To solve the multiple-label problem, I was thinking of something like adding a bot-comment containing buttons of labels that can be added and the author can choose which label to add.

soehms:
> To solve the multiple-label problem, I was thinking of something like adding a bot-comment containing buttons of labels that can be added and the author can choose which label to add.

If this is possible, that could be a good solution. However, I wouldn't send such a comment to Triage members as they can set the labels as they want. But let's discuss these things in #37373 or in the new PR.

@kwankyu, would you like to take a look here before I give a positive review?

kwankyu:
No objection from me. Just be ready to react for any complaint after the labeler goes in action.

soehms (approved):

amanmoon:
@soehms, The Actions variables are not set, leading the action to be skipped.

soehms:
> @soehms, The Action variables are not set, leading the action to be skipped.

This must be done by a maintainer of the repository. @mkoeppe, can you enable the feature by setting the repository variables `SMALL_THRESHOLD`, `MODERATE_THRESHOLD` and `LARGE_THRESHOLD`? Do we need to announce this on sage-devel?

mkoeppe:
I've added these variables with values 50, 100, 300, as per ticket description (to my understanding)

mkoeppe:
> Do we need to announce this on sage-devel?

I think that would be a great idea.

Also it would be good to explain it in https://github.com/sagemath/sage/wiki/Sage-10.4-Release-Tour#automatic-pr-size-labeler -- I've only created this section as a stub

soehms:
> > Do we need to announce this on sage-devel?
> 
> I think that would be a great idea.
> 

See https://groups.google.com/g/sage-devel/c/w4IeYgXgVUc.

soehms:
@amanmoon You should have a look at [this workflow run](https://github.com/sagemath/sage/actions/runs/8979068478/job/24660492110#step:3:21).

amanmoon:
The author did not have the latest develop branch (with this auto labeler PR merged) merged with his PR, so the file not found error occurred.
After everyone is on the latest commit this won't cause any issues. We can add a test to check this for now and remove the test after some time, as the test will be irrelevant when everyone is on the latest commit.

The `moderate threshold` label is not been assigned to PR even if the changes are between 100 and 300. we can see this [here](https://github.com/sagemath/sage/actions/runs/8980378300/job/24663908862?pr=37952#step:3:29).
Just to be sure, I tested the code again [here](https://github.com/amanmoon/sage/pull/8).

mkoeppe:
> This looks like a line feed occurs at the end of the `MODERATE_THRESHOLD` value. @mkoeppe can you check this value?

Indeed. I've removed it.

mkoeppe:
> If the author does not have the latest develop branch (with this auto labeler PR merged) merged with his PR, the file not found error will occur.
> After everyone is on the latest commit this won't cause any issues. We can add a test to check this for now and remove the test after some time, as the test will be irrelevant when everyone is on the latest commit.

I think it would be better if the script was taken from a separate checkout of the `develop` branch, not the PR branch:
As the script runs with elevated privileges to make the changes to PRs, it should not execute arbitrary code from a PR.

Also, recently I have switched to putting such helper scripts in the `.ci/` directory instead of the `.github/workflows/` directory, which I think makes the distinction clearer (`.githhub/workflows/` is referred to by actions references, whereas the helper scripts always come from some git checkout).

soehms:
> I think it would be better if the script was taken from a separate checkout of the `develop` branch, not the PR branch: As the script runs with elevated privileges to make the changes to PRs, the workflow should not execute arbitrary code from a PR.
> 

It is also possible to checkout single files (see for example [here](https://github.com/sagemath/sage/blob/744939e037a67193e730d7205e612e2d58197fca/.github/workflows/sync_labels.yml#L28)).
> Also, recently I have switched to putting such helper scripts in the `.ci/` directory instead of the `.github/workflows/` directory, which I think makes the distinction clearer (`.githhub/workflows/` is referred to by actions references, whereas the helper scripts always come from some git checkout).

I agree! There are some more to move:

```
$ LIST='(extract-sage-local.sh|scan-logs.sh|conda-lock-update.py|sync_labels.py|set_labels_by_changes.sh)'
$ egrep $LIST .github/workflows/*.yml
.github/workflows/ci-macos.yml:          .github/workflows/scan-logs.sh "artifacts/$LOGS_ARTIFACT_NAME"
.github/workflows/docker.yml:          .github/workflows/scan-logs.sh "artifacts/$LOGS_ARTIFACT_NAME"
.github/workflows/macos.yml:          .github/workflows/extract-sage-local.sh sage-local-artifact/sage-local-*.tar
.github/workflows/macos.yml:          .github/workflows/scan-logs.sh "artifacts/$LOGS_ARTIFACT_NAME"
.github/workflows/pr-labeler.yml:          chmod a+x .github/workflows/set_labels_by_changes.sh
.github/workflows/pr-labeler.yml:          .github/workflows/set_labels_by_changes.sh
.github/workflows/sync_labels.yml:          files: .github/sync_labels.py
.github/workflows/sync_labels.yml:          chmod a+x .github/sync_labels.py
.github/workflows/sync_labels.yml:          .github/sync_labels.py $ACTION $ISSUE_URL $PR_URL $ACTOR "$LABEL" "$REV_STATE" $LOG_LEVEL
.github/workflows/sync_labels.yml:          chmod a+x .github/sync_labels.py
.github/workflows/sync_labels.yml:          .github/sync_labels.py $REPO_URL $LOG_LEVEL
```

@amanmoon, can you open a corresponding follow-up PR?

mkoeppe:
Note I'm taking care of some of the scripts in:
- https://github.com/sagemath/sage/pull/37921

soehms:
Note that @tscrim started a discussion on the bebefit of the labels in the [sage-devel-thread](https://groups.google.com/g/sage-devel/c/w4IeYgXgVUc).

mkoeppe:
There's a complaint that the size is sometimes wrong: https://groups.google.com/g/sage-devel/c/w4IeYgXgVUc/m/Og7NP-3VAQAJ

soehms:
> It may be useful to illustrate such a proposal by putting examples of PRs of such sizes in the description, to facilitate discussion.
> 
> Also consider whether labels are necessary for the sizes in the middle. Labels for the smallest category and for the largest category have clear functions (which?)

Sorry, I lost track of this hint. At least because of that @vdelecroix is right in speaking of a wrong guidance in [this sage-devel post](https://groups.google.com/g/sage-devel/c/3PLZD-4UFIA/m/q3bfcSV7FgAJ). Another point is that I did not expect this PR to become controversial because I thought the naming of the labels made it clear that they were just quantitative indicators that could help with preselection in the GitHub search function. Now, I agree with @tscrim that we should have involved sage-devel much earlier.

PR #36171, opened 2023-09-01
Implement Chevalley-Eilenberg complexes with module coefficients

tscrim:
We implement the Chevalley-Eilenberg complex of a Lie algebra `L` with coefficients in a general `L` representation.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

### :hourglass: Dependencies

- #36169 uses the representations of Lie algebras

PR #36049, opened 2023-08-08
Implement general coefficient module for Chevalley-Eilenberg complex; add class for super Lie algebras

adityadwarkesh:
We have modified the code for computing the Chevalley-Eilenberg complex of a Lie algebra, so that it allows for generic coefficient modules. In order to do so, it was necessary to set up a general framework for representations of Lie algebras. For this, we have defined a `HomAlgebra` function, which returns the endomorphism Lie algebra ($\frak{gl}_n$) of a given free module, and a `LieAction` function, which returns the action of a module element on a Lie element, given the base module $M$ and Lie algebra homomorphism $f:L\rightarrow \frak{gl}_n$. An example has been provided.

We have also defined a new parent class `SuperLieAlgebra`, which inherits from `CombinatorialFreeModule` and provides a framework for globally finite $\mathbb{Z}$-graded Lie algebras. It has been filed as algebras/lie_algebras/superliealgebra.py. Some basic element methods and tests have been added.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [X] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [X] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37267, opened 2024-02-09
Implement categories of Lie superalgebras and from structure coefficients

tscrim:
We provide the initial implementation of Lie superalgebras defined by structure coefficients from #36049. This also requires an implementation of the corresponding categories.

### :memo: Checklist

### :hourglass: Dependencies

- #36171: Built from this branch. (Could be removed, but already positively reviewed.)


tscrim:
Okay, this is now essentially ready. I refactored things around to prepare for the future and to match more closely what is done in the Lie algebras. Likely things will go even more closely together in the future.

One thing that will need to be addressed in `graded_morphism()`. Right now it returns the image under the morphism, but it should return the morphism itself. This also will mean changing the inputs to remove the `argument`. However, this will allow for more general operations (e.g., compositions) by adding some extra classes that only require very mild implementation details.

@adityadwarkesh Please take a look. Let me know also what additional feature(s) you might want to include at this time. Of course, we can always add more on later PRs.

tscrim:
`graded_morphism` now returns a morphism instance. I was able to extend, where appropriate, the Lie algebra morphism code I had before. This will be better for future implementations and removes duplication. I think all doctests will now pass and we have full coverage, but we will wait for the bots.

adityadwarkesh:
@tscrim Thanks a lot! I also just took a look, and I think this much is good to go for the time being.

mkoeppe:
Failed tests in `lie_algebras`

PR #37268, opened 2024-02-09
Allow custom degree in `random_element` of polynomial quotient ring

emielwiedijk:
Currently, the `random_element` function of the polynomial quotient ring passes all arguments through to the base ring. However, the random_element function itself passes a default degree to the base ring, based on the degree of the quotient ring. So, overwriting the degree in kwargs gives a TypeError. Therefore, detect the `degree` argument manually, such that the user can overrride the default degree.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/polynomial/polynomial_quotient_ring.py
+++ b/src/sage/rings/polynomial/polynomial_quotient_ring.py
@@ -1260,12 +1260,18 @@ class PolynomialQuotientRing_generic(QuotientRing_generic):
 
     cover_ring = polynomial_ring
 
-    def random_element(self, *args, **kwds):
+    def random_element(self, degree=None, *args, **kwds):
         """
         Return a random element of this quotient ring.
 
         INPUT:
 
+        - ``degree`` - Optional argument: either an integer for fixing the
+          degree, or a tuple of the minimum and maximum degree. By default the
+          degree is n - 1 with n the degree of the polynomial ring. Note that
+          the degree of the polynomial is fixed before the modulo calculation.
+          So when `degree` is bigger then the degree of the polynomial ring, the
+          degree of the returned polynomial would be lower than `degree`.
         - ``*args``, ``**kwds`` - Arguments for randomization that are passed
           on to the ``random_element`` method of the polynomial ring, and from
           there to the base ring
@@ -1283,8 +1289,11 @@ class PolynomialQuotientRing_generic(QuotientRing_generic):
             sage: F2.random_element().parent() is F2
             True
         """
+        if degree is None:
+            degree = self.degree() - 1
+
         return self(self.polynomial_ring().random_element(
-            degree=self.degree() - 1, *args, **kwds))
+            degree=degree, *args, **kwds))
 
     @cached_method
     def _S_decomposition(self, S):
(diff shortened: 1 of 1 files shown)


src/sage/rings/polynomial/polynomial_quotient_ring.py
```diff
@@ -1260,12 +1260,15 @@ def polynomial_ring(self):
...
-    def random_element(self, *args, **kwds):
+    def random_element(self, *args, degree=None, **kwds):
         """
         Return a random element of this quotient ring.
 
         INPUT:
 
+        - ``degree`` - Optional argument: either an integer for fixing the
+          degree, or a tuple of the minimum and maximum degree. By default the
+          degree is n - 1 with n the degree of the polynomial ring.
```
S17A05: To be explicit it might be worth it to state here that the degree is only fixed up to modulo calculation, so that the user can check the documentation when the query for a degree 20 element in a quotient ring with modulus of degree 5 returns and element of degree at most 4.

src/sage/rings/polynomial/polynomial_quotient_ring.py
```diff
@@ -1260,12 +1260,15 @@ def polynomial_ring(self):
 
     cover_ring = polynomial_ring
 
-    def random_element(self, *args, **kwds):
+    def random_element(self, *args, degree=None, **kwds):
         """
```
S17A05: Is there a reason why you placed the optional `degree` argument after `*args`, instead of in front as it is done for `.random_element()` in the PolynomialRing_general class and the method call a few lines below?

S17A05 (commented):
Just two small things, otherwise LGTM.

src/sage/rings/polynomial/polynomial_quotient_ring.py
emielwiedijk: I just put the keyword arguments together, but I agree that it's better to put `degree` before `*args`.

src/sage/rings/polynomial/polynomial_quotient_ring.py
emielwiedijk: I would find it hard to predict the behaviour when `degree` is bigger than the degree of the polynomial ring.  I added a note that the degree of the polynomial is fixed before the modulo calculation. Would that be clear enough?

src/sage/rings/polynomial/polynomial_quotient_ring.py
```diff
@@ -1260,12 +1260,18 @@ def polynomial_ring(self):
...
         Return a random element of this quotient ring.
 
         INPUT:
 
+        - ``degree`` - Optional argument: either an integer for fixing the
+          degree, or a tuple of the minimum and maximum degree. By default the
+          degree is n - 1 with n the degree of the polynomial ring. Note that
+          the degree of the polynomial is fixed before the modulo calculation.
+          So when `degree` is bigger then the degree of the polynomial ring, the
+          degree of the returned polynomial would be lower than `degree`.
```
S17A05: Just a small typo left: It should be "than" instead of "then".

src/sage/rings/polynomial/polynomial_quotient_ring.py
S17A05: Yeah, the way you wrote it now is good - my comment was just about the user not being confused when the output representative doesn't have the degree that was given as input.

S17A05 (commented):
For the future, please don't force push big changes and instead put them into a new commit, otherwise it makes the review process harder because everything has to be re-checked again.

S17A05 (approved):
Thanks, LGTM.

Issue #37263, opened 2024-02-08
Using `--enable-system-site-packages` is broken


tornaria:
Using `--enable-system-site-packages` is broken:
 - first, the version of `sage_conf` inside the venv is installed in such a crooked way that *it will not override* the one installed in the system site-packages, in spite the sage venv site-packages being listed before the system site-packages in `sys.path`.
 - the version of `sage_conf.py` I have in my system doesn't define `GAP_ROOT_PATHS` (not a thing in 10.2). This results in an incredibly scary-looking error (to simulate try `GAP_ROOT_PATHS=/ sage -c 'gap(1)'` or worse `GAP_ROOT_PATHS=/ sage -c 'libgap(1)'`).
 - I don't know how to fix the search order for `sage_conf` so I just added `GAP_ROOT_PATHS` to the system `sage_conf` (I can't remove it since I don't want to break sage for other users)
 - With that fixed, almost all tests pass except some in `src/sage_docbuild` and the top comment in `src/sage/env.py`

_Originally posted by @tornaria in https://github.com/sagemath/sage/issues/36181#issuecomment-1935081587_
            
EDIT: some relevant notes that are mentioned in the original post:
 - I'm testing with PR #36181 at 29130ea, since my system python is 3.12
 - I used `./configure --enable-system-site-packages --disable-doc`

tornaria:
More details:
### Failures in `src/sage_docbuild/*`

These are all due to missing `sphinx`. Indeed, it was not installed in my system (nor was it installed in the venv, maybe because of the `--disable-doc`). I guess this can be fixed by adding `# sage.doctest: needs sphinx` at the top of the files as follows:

```diff
diff --git a/src/sage_docbuild/__main__.py b/src/sage_docbuild/__main__.py
index 0d15808a69c..acfcb8392a0 100644
--- a/src/sage_docbuild/__main__.py
+++ b/src/sage_docbuild/__main__.py
@@ -1,3 +1,4 @@
+# sage.doctest: needs sphinx
 r"""
 Sage docbuild main
 
diff --git a/src/sage_docbuild/builders.py b/src/sage_docbuild/builders.py
index a43fdda7350..0be43be307f 100644
--- a/src/sage_docbuild/builders.py
+++ b/src/sage_docbuild/builders.py
@@ -1,3 +1,4 @@
+# sage.doctest: needs sphinx
 """
 Documentation builders
 
[...]

PR #37269, opened 2024-02-09
sage_docbuild: add needs sphinx tags

tornaria:
Add "needs sphinx" tags to some tests in `src/sage_docbuild` so there
is no failure when sphinx is not installed in the system.

See: https://github.com/sagemath/sage/issues/37263#issuecomment-1935113975

### :memo: Checklist

### DIFF
--- a/src/sage_docbuild/__main__.py
+++ b/src/sage_docbuild/__main__.py
@@ -1,3 +1,4 @@
+# sage.doctest: needs sphinx
 r"""
 Sage docbuild main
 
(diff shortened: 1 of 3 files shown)


mkoeppe:
Who's doctesting `src/sage_docbuild` when sage-docbuild is not installed and why? sphinx is a dependency (install-requires) of sage-docbuild

tornaria:
> Who's doctesting `src/sage_docbuild` when sage-docbuild is not installed and why? sphinx is a dependency (install-requires) of sage-docbuild

`./sage -t --all`, as explained in https://github.com/sagemath/sage/issues/37263#issuecomment-1935113975, when using `--disable-doc --enable-editable`

PR #37270, opened 2024-02-09
Allow python 3.12 and scipy 1.12

tornaria:
This is a minimal PR to allow building with python 3.12, given the
circumstances surrounding #36181.

I've tested this works with system python 3.12 by:
```
./bootstrap
./configure --disable-doc --disable-editable --enable-system-site-packages
env 'MAKE=make -j36' make
./sage -tp 36 --all
./sage -tp 36 --all --long
```

It only gave the expected failures in
`src/sage/matroids/database_collections.py` from #37140.

Since scipy 1.12 works ok after #37123 and it's just a one-liner to
change the required version, I included it here.

No attempt is made at upgrading anything in sage-the-distro.

### :memo: Checklist

EDIT: rebased on top of #36983 to address reviewer suggestion.


mkoeppe:
This is missing the changes from #36983, so this will only work with system python 3.12 that has resurrected distuils.

tornaria:
> This is missing the changes from #36983, so this will only work with system python 3.12 that has resurrected distuils.

I don't have "resurrected distutils" and it works.

dimpase:
we're entering a delightful subject of Python (without Monty) theology.

What's resurrection of distutils?

tornaria:
I rebased it on top of #36983. But I don't like this workflow of adding "dependent PRs" and it makes this PR harder to review.

mkoeppe:
> What's resurrection of distutils?

Well, distutils was killed in CPython 3.12.
If you can still import it in a system Python 3.12, it is because it has been resurrected.

mkoeppe:
All joking aside, it is of course `setuptools` that makes its vendored version of distutils (`setuptools._distutils`) available as `distutils`, via a complicated mechanism: https://github.com/pypa/setuptools/blob/main/setup.py#L52, which is slated for removal:
- https://github.com/pypa/setuptools/issues/4137

tornaria:
All joking aside: is this good now?

build/pkgs/python3/SPKG.rst
```diff
@@ -8,16 +8,15 @@ By default, Sage will try to use system's ``python3`` to set up a virtual
 environment, a.k.a. `venv <https://docs.python.org/3.10/library/venv.html>`_
 rather than building a Python 3 installation from scratch.
 
-Sage will accept versions 3.9.x to 3.10.x.
+Sage will accept versions 3.9.x to 3.11.x.
```
Jamim: _Hello @tornaria,_
There is a small thing to update.
```suggestion
Sage will accept versions 3.9.x to 3.12.x.
```
_Best regards!_

Jamim (changes requested):

tobiasdiez (approved):
Thanks! Looks good to me. Feel free to set it to "positive review" once the remark by @Jamim is committed.

build/pkgs/python3/SPKG.rst
tornaria: Good catch, thanks!

Jamim (approved):

PR #37271, opened 2024-02-09
make map_coefficients more general

### LINKED ISSUE #18264: make map_coefficients in modules_with_basis safer or more general
<div id="comment:0"></div>

We would love to be able to, for example, present a symmetric function with factored coefficients.  But:

```
sage: f = SymmetricFunctions(ZZ).s()([2,2,1])
sage: f.map_coefficients(factor)
0
```

The result is not against the specification, because `map_coefficients` requires an endofunction.

There are two issues:

1.) we possibly want to be warned when supplying `map_coefficients` with something that is not an endofunction.  This might be achieved by adding an optional argument `check` with default `True`.

2.) we want some way to factor the coefficients.  Nicolas proposed to make something like

```
sage: f.map_coefficients(factor, codomain=SR)
```
or

```
sage: f.change_base_ring(SR).map_coefficients(factor) 
```
work.

Note that the `Polynomial` class contains something similar, namely:

```
def map_coefficients(self, f, new_base_ring=None)
```

However, there might be another tiny problem: it seems that currently it is not possible to make factored integers into a commutative ring.  This should possibly be a separate ticket.

CC:  @zabrocki @darijgr @sagetrac-sage-combinat @hivert

Component: **algebra**

Keywords: **map_coefficients, modules_with_basis, base_ring, check arguments**

Author: **Martin Rubey**

Branch/Commit: **[public/ticket/18264](https://github.com/sagemath/sagetrac-mirror/tree/public/ticket/18264) @ [`d0d80e7`](https://github.com/sagemath/sagetrac-mirror/commit/d0d80e73ccae7a5823e8692e72e0c31e59800e1b)**

Reviewer: **Nicolas M. Thiéry**

_Issue created by migration from https://trac.sagemath.org/ticket/18264_

mantepse:
The internal documentation of `categories/modules_with_basis.py` mentions
```
        # TODO: Define the appropriate element methods here (instead of in
        # subclasses).  These methods should be consistent with those on
        # polynomials.
```
In line with this, we allow a new keyword `new_base_ring` argument for `map_coefficients`, and allow maps that are not necessarily endofunctions.

In particular, this fixes #18264.  In the end, this is needed for #37033, where we need a generic way to retract a module element to a smaller coefficient ring, if possible.

Unfortunately, I am unable to find the discussion on sage-devel anymore.  The current PR is certainly not optimal.

Dependencies: #37220


mantepse:
Superseded by #37766

PR #36964, opened 2023-12-25
Add `# sage_setup: distribution` directives to all files, remove remaining `# coding: utf-8`

mkoeppe:
These directives at the top of the file inform developers about the intended assignment of modules to pip-installable distributions.

As of this PR, there should be no change to the existing distributions (sagemath-categories...) nor the monolithic build of the Sage library.

- Cherry-picked from #35095

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36951 (merged here)
- Depends on #36676 (merged here)
- Depends on #37667 (merged here)

PR #37286, opened 2024-02-10
sagemath-standard: fix manifest

tornaria:
This PR fixes two issues with the manifest for sagemath-standard:

- src/MANIFEST.in: don't include stuff outside src/sage

The use of `global-include <PATTERN>` is inconvenient since it will
include any file matching the pattern anywhere; replace it with
`recursive-include sage <PATTERN>` which is what we really mean.

- sagemath-standard/setup.py: workaround setuptools_scm

The workaround already there is not good for setuptools_scm >= 8.
Implement a more robust workaround (should work for any version of
setuptools_scm but I have not tested, so I did not remove the old
workaround -- it's harmless to apply both).

### :memo: Checklist

### DIFF
--- a/pkgs/sagemath-standard/setup.py
+++ b/pkgs/sagemath-standard/setup.py
@@ -24,6 +24,11 @@ try:
 except ImportError:
     pass
 
+# Different workaround: disable `walk_revctrl` in setuptools
+# This is needed for setuptools_scm >= 8, should work for any version
+import setuptools.command.egg_info
+setuptools.command.egg_info.walk_revctrl = lambda: ()
+
 #########################################################
 ### Set source directory
 #########################################################
--- a/src/MANIFEST.in
+++ b/src/MANIFEST.in
@@ -1,6 +1,6 @@
 include VERSION.txt
 
-global-include *.pxi *.pxd *.h *.hpp
+recursive-include sage *.pxi *.pxd *.h *.hpp
 
 prune sage/ext/interpreters   # In particular, __init__.py must not be present in the distribution; or sage_setup.autogen.interpreters.rebuild will not generate the code
 prune sage_setup


tornaria:
@mkoeppe I think this is also easy and uncontroversial.

The workaround for setuptools_scm >= 8 is I think much better (and maybe we should also remove the previous workaround, but I'm not sure and it's certainly harmless to apply both).

mkoeppe:
I'll take a look

mkoeppe:
In several other setup.py files, we have another workaround that somehow didn't make it into pkgs/sagemath-standard. Take a look at https://github.com/sagemath/sage/blob/develop/pkgs/sage-docbuild/setup.py#L13
for example

tornaria:
> In several other setup.py files, we have another workaround that somehow didn't make it into pkgs/sagemath-standard. Take a look at https://github.com/sagemath/sage/blob/develop/pkgs/sage-docbuild/setup.py#L13 for example

The workaround here just replaces with a noop the function `walk_revctrl()` in `setuptools.command.egg_info` which is used to list files from scm. The code that uses `walk_revctrl()` is 19 years old so it seems quite safe (https://github.com/pypa/setuptools/blame/main/setuptools/command/egg_info.py#L591)

The other seems a workaround for a different issue, I don't understand how it works and its implications. The warning "LookupError" seemed harmless.

mkoeppe:
This seems to work well, thanks!

The next time I touch the setup.py files in the other disitributions, I'll check if some of the other workarounds for setuptools_scm can be replaced by this.

kiwifb:
Hi guys, this is the only ticket included in 10.4.beta4 that I think may be related to what I just found. When I create a sdist for sagemath-standard-10.4.beta3 there are no python files (.py) in the sage folder. Only cython files. The only python files present in the sdist are at the top level (setup.py) and inside bin (sage-num-threads.py and sage-startuptime.py).

Do you have any ideas?

mkoeppe:
Indeed the 10.4.beta3 sdist (also on PyPI) is broken in this way. I'll investigate.

tornaria:
How do you create the sdist? Is the repo bootstrapped? What version of setuptools?

Can you try reverting each one of the two changes by hand to see which one, if any, is the culprit?

mkoeppe:
This change probably comes from #36951 or #36964, not here.

mkoeppe:
@kiwifb Please review #37434, which adds the missing automatic test for the sagemath-standard distribution built out of pkgs/sagemath-standard....

kiwifb:
I will investigate further the origin, I have my doubts about #36951 but we never know. There could be something in #36964 but that will be difficult to pull out.

mkoeppe:
I have a simple fix in #37804

kiwifb:
Looks like something in #36964 is at fault.

PR #37288, opened 2024-02-10
Directly convert PermutationGroup element into sized Permutation

### LINKED ISSUE #37284: Cannot convert PermutationGroup element into sized Permutation
### Steps To Reproduce

The following code shows that an element of `PermutationGroup` can be converted into a sizeless `Permutation` and then into a sized one, but not directly.

```
PG = PermutationGroup(SymmetricGroup(5).gens())
P5 = Permutations(5)
p = PG.an_element()
print(p.parent())
q = Permutation(p)
print(q.parent())
r = P5(q)
print(r.parent())
s = P5(p)            # <----- this one fails
```

### Expected Behavior

It should be possible to convert a `PermutationGroup` directly into into a sized `Permutation`.

### Actual Behavior

The output is

> Permutation Group with generators [(1,2), (1,2,3,4,5)]
> Standard permutations
> Standard permutations of 5

and then 

> TypeError: object of type 'sage.groups.perm_gps.permgroup_element.PermutationGroupElement' has no len()

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Ubuntu 22.04.3 LTS
- **Sage Version**: 10.2.beta5
```

### Checklist

RuchitJagodara:
This patch fixes #37284 by converting the input into a Permutation class object if 
it is the object of another class.
﻿

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -7101,8 +7101,11 @@ class StandardPermutations_n_abstract(Permutations):
             sage: P5(p)
             [4, 5, 1, 2, 3]
         """
-        if not isinstance(x, Permutation):
-            x = x.domain()
+        try:
+            if not isinstance(x, Permutation):
+                x = x.domain()
+        except AttributeError:
+            pass
         if len(x) < self.n:
             x = list(x) + list(range(len(x) + 1, self.n + 1))
         return self.element_class(self, x, check=check)
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -7102,7 +7102,7 @@ class StandardPermutations_n_abstract(Permutations):
             [4, 5, 1, 2, 3]
         """
         if not isinstance(x, Permutation):
-            x = Permutation(x)
+            x = x.domain()
         if len(x) < self.n:
             x = list(x) + list(range(len(x) + 1, self.n + 1))
         return self.element_class(self, x, check=check)
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -7093,7 +7093,7 @@ class StandardPermutations_n_abstract(Permutations):
             sage: Permutations(6)(x)            # known bug
             [1, 4, 5, 2, 3, 6]
 
-            check if :trac:`37284` is fixed::
+        Ensure that :issue:`37284` is fixed::
 
             sage: S5 = SymmetricGroup(5)
             sage: P5 = Permutations(5)
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -7098,7 +7098,9 @@ class StandardPermutations_n_abstract(Permutations):
             sage: PG = PermutationGroup([[(1,2,3),(5,6)],[(7,8)]])
             sage: P8 = Permutations(8)
             sage: p = PG.an_element()
-            sage: P8(p).parent()
+            sage: q = P8(p); q
+            [2, 3, 1, 4, 6, 5, 8, 7]
+            sage: q.parent()
             Standard permutations of 8
         """
         if isinstance(x, PermutationGroupElement):
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -7101,6 +7101,8 @@ class StandardPermutations_n_abstract(Permutations):
             sage: x = PG([4,3,5,1,2])
             sage: s = P5(x); s
             [4, 3, 5, 1, 2]
+            sage: s.parent()
+            Standard permutations of 5
         """
         if isinstance(x, PermutationGroupElement):
             return self. _from_permutation_group_element(x)
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -7092,7 +7092,17 @@ class StandardPermutations_n_abstract(Permutations):
             [1, 4, 5, 2, 3, 6]
             sage: Permutations(6)(x)            # known bug
             [1, 4, 5, 2, 3, 6]
+
+            check if :trac:`37284` is fixed::
+
+            sage: S5 = SymmetricGroup(5)
+            sage: P5 = Permutations(5)
+            sage: p = S5.list()[3]
+            sage: P5(p)
+            [4, 5, 1, 2, 3]
         """
+        if not isinstance(x, Permutation):
+            x = Permutation(x)
         if len(x) < self.n:
             x = list(x) + list(range(len(x) + 1, self.n + 1))
         return self.element_class(self, x, check=check)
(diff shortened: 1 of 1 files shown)


RuchitJagodara:
Does anybody know any better way so that we can avoid try except block here ?

src/sage/combinat/permutation.py
```diff
@@ -7092,7 +7092,20 @@ def _element_constructor_(self, x, check=True):
             [1, 4, 5, 2, 3, 6]
             sage: Permutations(6)(x)            # known bug
             [1, 4, 5, 2, 3, 6]
+
+            check if :trac:`37284` is fixed::
```
grhkm21:
```suggestion
        Ensure that :issue:`37284` is fixed::
```

src/sage/combinat/permutation.py
```diff
@@ -7092,7 +7092,20 @@ def _element_constructor_(self, x, check=True):
...
+            sage: P5 = Permutations(5)
+            sage: p = S5.list()[3]
+            sage: P5(p)
+            [4, 5, 1, 2, 3]
         """
+        try:
+            if not isinstance(x, Permutation):
+                x = x.domain()
+        except AttributeError:
+            pass
```
grhkm21:
```suggestion
        if not isinstance(x, Permutation) and hasattr(x, "domain"):
            x = x.domain()
```
You can use `hasattr` to check if `domain` exists. However, from what I understand, `try`-`except` is definitely acceptable in Sage.

grhkm21:
Just tiny comments :) looks good to me

src/sage/combinat/permutation.py
tscrim: It is not good to check that something is not a `Permutation`. You should instead check that the input is a `PermutationGroupElement`:
```suggestion
        if isinstance(x, PermutationGroupElement):
            x = x.domain()
```
No accidental catching of things that have a `domain()` that is not a permutation (say, a function from $\{1,2,3\} \to \{0\}$), no `try`-`except` blocks, and simple code.

tscrim:
> Does anybody know any better way so that we can avoid try except block here ?

See my comment. Checking for specific types is usually better than checking for not a specific type when trying to apply methods.

grhkm21:
Oops, learned something myself. I will keep that in mind too :)

RuchitJagodara:
I have done all necessary changes that you suggested, thank you for your suggestions ! : )

tscrim:
Thank you. However, your test is not good because that was working before from the coercion map defined using `P5._from_permutation_group_element`. It would probably be better to go through that method here as well. (Yes, it does the same thing with going back through the `_element_constructor_`, but it might do something smarter/faster in the future. It also makes the code easier to debug should a problem arise later.) Sorry I missed that before.

mantepse:
Is #37318 related?

tscrim:
> Is #37318 related?

From a quick look, doesn’t seem related.

RuchitJagodara:
Then, should I change _coerce_map_from function and redirect the flow to the _from_permutation_group_element whenever G is a PermutationGroup, but I don't know what to include in isinstance function to check wether the group is permutation group or not ?

https://github.com/sagemath/sage/blob/f444d326540464fa9908fcd73fd3ab16511c5410/src/sage/combinat/permutation.py#L7210-L7211

src/sage/combinat/permutation.py
```diff
@@ -7092,7 +7092,17 @@ def _element_constructor_(self, x, check=True):
...
+        Ensure that :issue:`37284` is fixed::
+
+            sage: S5 = SymmetricGroup(5)
+            sage: P5 = Permutations(5)
+            sage: p = S5.list()[3]
+            sage: P5(p)
+            [4, 5, 1, 2, 3]
         """
+        if isinstance(x, PermutationGroupElement):
+            x = x.domain()
```
tscrim:
```suggestion
            return self. _from_permutation_group_element(x)
```
No, you should only make this change here. Do not change any coercions!

tscrim (approved):
The test is not that good (it is better to not depend on the ordering of the group, and a slightly more general group on, say, 4 letters would be better), but it is sufficient. Hence, we can merge this as-is, but it would be nice to have a slightly better test.

RuchitJagodara:
Should I change the test? (Just confirming because @tscrim approved the changes but there is still a 'needs work' label on this PR.)

tscrim:
@RuchitJagodara I’ve left the decision up to you, but it would be better to have a more generic test.

RuchitJagodara:
I have updated the test, @tscrim can you please review that ?

tscrim:
That is essentially the same test. The point is to have a more interesting group than something (trivially) isomorphic to the symmetric group. So something like the group generated by `[(1,2),(2,3), (4,5)]`.

RuchitJagodara:
Sorry, because of my mid-term exams, I was not able to make these changes earlier. But I think the test that I have added is not dependent on ordering of the group, and is not trivial, also.

RuchitJagodara:
> That is essentially the same test. The point is to have a more interesting group than something (trivially) isomorphic to the symmetric group. So something like the group generated by `[(1,2)(2,3), (4,5)]`.

When I try to create a permutation group using the generators [(1,2)(2,3), (4,5)], it gives me a ValueError, saying 'invalid data to initialize a permutation.' I am a bit confused as to why that error is occurring.

tscrim:
> > That is essentially the same test. The point is to have a more interesting group than something (trivially) isomorphic to the symmetric group. So something like the group generated by `[(1,2)(2,3), (4,5)]`.
> 
> When I try to create a permutation group using the generators [(1,2)(2,3), (4,5)], it gives me a ValueError, saying 'invalid data to initialize a permutation.' I am a bit confused as to why that error is occurring.

Sorry, there was a typo on my part. I meant
```
sage: G = PermutationGroup([[(1,2),(3,4)], [(4,5)]])
sage: G.isomorphism_to(DihedralGroup(6))
Permutation group morphism:
  From: Permutation Group with generators [(4,5), (1,2)(3,4)]
  To:   Dihedral group of order 12 as a permutation group
  Defn: [(4,5), (1,2)(3,4)] -> [(1,2)(3,6)(4,5), (1,3)(4,6)]
```
Although you could interpret it as
```
sage: G = PermutationGroup([[(1,2,3)], [(4,5)]])  # = Z3 x Z2
```

src/sage/combinat/permutation.py
```diff
@@ -7092,7 +7092,17 @@ def _element_constructor_(self, x, check=True):
...
             sage: Permutations(6)(x)            # known bug
             [1, 4, 5, 2, 3, 6]
+
+        Ensure that :issue:`37284` is fixed::
+
+            sage: PG = PermutationGroup([[(1,2,3),(5,6)],[(7,8)]])
+            sage: P5 = Permutations(8)
+            sage: p = PG.an_element()
+            sage: P5(p).parent()
+            Standard permutations of 8
```
tscrim:
```suggestion
            sage: P8 = Permutations(8)
            sage: p = PG.an_element()
            sage: P8(p).parent()
            Standard permutations of 8
```
Since the group is larger now. It would be good to display the output as well.

tscrim (changes requested):
Anyways, the test is better now. Thank you. Just a few details left.

tscrim (approved):
Thank you. LGTM now.

PR #37287, opened 2024-02-10
sagemath-standard: include sage_setup in sdist

tornaria:
This makes it so that `sage_setup` is included in `sagemath_standard`
for convenience and ease of installation.

Having `sage_setup` as a separate distribution leads to chicken-and-egg problem. For example, this is what happens when I try to build `sagemath-standard` from #37270:
```
cd pkgs/sagemath-standard
$ python -m build -s 
* Creating venv isolated environment...
* Installing packages in isolated environment... (cypari2 >=2.1.1, cysignals >=1.10.2, cython >=3.0, != 3.0.3, <4.0, gmpy2 ~=2.1.b999, jinja2 >=3.0, jupyter_core >=4.6.3, memory_allocator, numpy >=1.19, pkgconfig, pplpy >=0.8.6, sage-conf ~= 10.3b7, sage-setup ~= 10.3b7, sage_setup[autogen], setuptools >= 68.1.1, wheel >=0.36.2)

[... snip ...]

ERROR: Could not find a version that satisfies the requirement sage-setup~=10.3b7 (from versions: 9.4rc1)
ERROR: No matching distribution found for sage-setup~=10.3b7

[...]
```
This is because the version of sage-setup which is available on pypi requires python <= 3.11. Even if the branch I'm trying to build allows python 3.12, I cannot use it.

The main idea of this change is that `sagemath-standard` sdists are self-contained and include the exact matching version of sage-setup. The self-contained part is good for the pypi sdist, but the matching version allows for greater flexibility when testing, given that the implementations of `sagemath-standard/setup.py` and `sage_setup` are so intrincately connected and really need to match.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.

PR #37290, opened 2024-02-11
sagemath-standard: don't require Cython to create sdist

tornaria:
The implementation of sage.misc.package_dir.read_distributions uses the function `Cython.Utils.open_source_file()` to open a file instead of `io.open()`. This function was introduced in #29701 (b582789).

The only difference between `open_source_file()` and `io.open()` is that the former will open the file as binary to check the encoding. But all our files should be utf-8 so we don't need this trick.

In case I’m mistaken and there is a reason this trick is really necessary (@mkoeppe?) it seems better to include a version of open_source_file(), for two main reasons: (a) we don’t pull the whole of Cython just for a trivial function; (b) changes in Cython don’t affect sagemath (this is not a documented function of Cython, afaict)

### :memo: Checklist

### DIFF
--- a/src/sage/misc/package_dir.py
+++ b/src/sage/misc/package_dir.py
@@ -118,8 +118,7 @@ def read_distribution(src_file):
         sage: read_distribution(os.path.join(SAGE_SRC, 'sage', 'graphs', 'graph_decompositions', 'modular_decomposition.py'))
         ''
     """
-    from Cython.Utils import open_source_file
-    with open_source_file(src_file, error_handling='ignore') as fh:
+    with open(src_file, encoding='utf-8', errors='ignore') as fh:
         for line in fh:
             # Adapted from Cython's Build/Dependencies.py
             line = line.lstrip()


pkgs/sagemath-standard/sage_setup
```diff
@@ -0,0 +1 @@
+../../src/sage_setup
```
mkoeppe: please remove this change

mkoeppe:
The change to just use `open` is fine with me.

tornaria:
With this PR, I can build a sdist with just setuptools as here:
```
./bootstrap sagelib
cd pkgs/sagemath-standard
python -m venv venv_for_sdist
venv_for_sdist/bin/pip install build
venv_for_sdist/bin/pip install setuptools
venv_for_sdist/bin/python -m build -snx
```

Note that I have to ignore (-x) because the build-dependencies have not been changed (they should, because the build-system requires are supposed to be just what is required to run the build-system; for build wheel requirements see https://peps.python.org/pep-0517/#get-requires-for-build-wheel).

mkoeppe:
AFAIK setuptools does not support specifying requires-for-build-{sdist,wheel,editable}.

We can of course put our own PEP-517/660 build_meta in there if we really want to.

pkgs/sagemath-standard/sage_setup
tornaria: I didn't get a chance to explain. The symlink is there so building the sdist actually works just with setuptools.

Without this symlink, there's no way to build the sdist:
```
$ VENV_FOR_SDIST/bin/python -m build -snx
...
ModuleNotFoundError: No module named 'sage_setup'
```
or worse:
```
$ VENV_FOR_SDIST/bin/python -m build -sx
* Creating venv isolated environment...
* Installing packages in isolated environment... (cypari2 >=2.1.1, cysignals >=1.10.2, cython >=3.0, != 3.0.3, <4.0, gmpy2 ~=2.1.b999, jinja2 >=3.0, jupyter_core >=4.6.3, memory_allocator, numpy >=1.19, pkgconfig, pplpy >=0.8.6, sage-conf ~= 10.3b7, sage-setup ~= 10.3b7, sage_setup[autogen], setuptools >= 68.1.1, wheel >=0.36.2)
...
ERROR: Could not find a version that satisfies the requirement sage-setup~=10.3b7 (from versions: 9.4rc1)
ERROR: No matching distribution found for sage-setup~=10.3b7
```

mkoeppe: Obviously you need to make the distribution sage-setup in the correct version available.

tornaria:
> AFAIK setuptools does not support specifying requires-for-build-{sdist,wheel,editable}.
> 
> We can of course put our own PEP-517/660 build_meta in there if we really want to.

See: https://peps.python.org/pep-0517/#get-requires-for-build-wheel and https://peps.python.org/pep-0517/#get-requires-for-build-sdist.

I assume we are aiming for pep-517 build and running setup.py directly is not supported.

mkoeppe:
And PEP-660 for modern editable.

tornaria:
Can you suggest how to build a sdist without the symlink or the `PYTHONPATH` hack?

mkoeppe:
I'll prepare a branch.

pkgs/sagemath-standard/sage_setup
tornaria: ok, I removed the symlink.

mkoeppe:
That's:
- #37292

tornaria:
> That's:
> 
>     * [pkgs/sagemath-standard: Remove self-reference at build time #37292](https://github.com/sagemath/sage/pull/37292)

What is this supposed to fix?

mkoeppe:
It fixes the "circular reference" that you complained about in https://github.com/sagemath/sage/pull/37287#issuecomment-1937167697

tornaria:
> Can you suggest how to build a sdist without the symlink or the `PYTHONPATH` hack?

mkoeppe:
We seem to be miscommunicating. You build the sdist by making the build requirements declared per PEP-518 available. For example by building wheels of sage-setup and then sagemath-environment in the required versions and making them available for `python -m build -s`

tornaria:
> It fixes the "circular reference" that you complained about in [#37287 (comment)](https://github.com/sagemath/sage/pull/37287#issuecomment-1937167697)

Nothing to do with this PR.

You were about to tell me how to build a sdist of sagemath-standard from the github repo, but instead you've just turned one obstacle (sage-setup) into two (sage-setup + sagemath-environment).

mkoeppe:
They are not obstacles, they are build dependencies. Where's the difficulty in building them?

tornaria:
> They are not obstacles, they are build dependencies. Where's the difficulty in building them?

They are NOT independent. The idea is that build-system dependencies should be readily available so I can, you know, use them to build the package.

Here I cannot modify and build from source if I need to change sage-setup in sync with sagemath-standard because when I try to build sagemath-standard the build frontend will try to build sage-setup from pypi instead of my version.

mkoeppe:
> The idea is that build-system dependencies should be readily available

Whose idea would that be?

mkoeppe:
> Here I cannot modify and build from source if I need to change sage-setup in sync with sagemath-standard because when I try to build sagemath-standard the build frontend will try to build sage-setup from pypi instead of my version.

You already know how to do this. If you don't want to use the venv that is automatically provisioned by `build`, you make your own venv with the versions that you want to use. Then you run `python -m build -nx`.

tornaria:
> > Here I cannot modify and build from source if I need to change sage-setup in sync with sagemath-standard because when I try to build sagemath-standard the build frontend will try to build sage-setup from pypi instead of my version.
> 
> You already know how to do this. If you don't want to use the venv that is automatically provisioned by `build`, you make your own venv with the versions that you want to use. Then you run `python -m build -nx`.

Have you tried it? As I explained already it doesnt work:
```
$ python -m build
* Creating venv isolated environment...
* Installing packages in isolated environment... (cypari2 >=2.1.1, cysignals >=1.10.2, cython >=3.0, != 3.0.3, <4.0, gmpy2 ~=2.1.b999, jinja2 >=3.0, jupyter_core >=4.6.3, memory_allocator, numpy >=1.19, pkgconfig, pplpy >=0.8.6, sage-conf ~= 10.3b7, sage-setup ~= 10.3b7, sage_setup[autogen], sagemath-environment ~= 10.3b7, setuptools >= 68.1.1, wheel >=0.36.2)
...
ERROR: Could not find a version that satisfies the requirement sagemath-environment~=10.3b7 (from versions: none)
ERROR: No matching distribution found for sagemath-environment~=10.3b7
```

mkoeppe:
Sounds like you need to learn about `PIP_FIND_LINKS`?

tornaria:
Also: you asked me to remove the symlink,, and I did. I asked politely for an alternative and you just hijacked the issue to push forward completely unrelated stuff.

tornaria:
> > The idea is that build-system dependencies should be readily available
> 
> Whose idea would that be?

PEP-518: "The purpose of this PEP though, is to specify the minimal set of requirements for the build system to simply begin execution."

mkoeppe:
If you set this PR to "needs review", I'll be happy to set it to "positive review" immediately. It's a simple enough change.

mkoeppe:
There you go.

tornaria:
> If you set this PR to "needs review", I'll be happy to set it to "positive review" immediately. It's a simple enough change.

I "asked for review" in the github interface, but I forgot that even that is NIH.

mkoeppe:
Yes, we were clearly miscommunicating.

Start from scratch? Feel free to pick a place.

tornaria:
> Sounds like you need to learn about `PIP_FIND_LINKS`?

Can't find it anywhere. If it's a pip thing, let me remind you that we agreed the aim is to pep-517 build, not "pip" build.

tornaria:
> Yes, we were clearly miscommunicating.
> 
> Start from scratch? Feel free to pick a place.

github issues are definitely not a good place to build a community.

I wish #sagemath would be used, but alas it's not.

mkoeppe:
> I wish #sagemath would be used, but alas it's not.

I can be on Zulip if necessary.

mkoeppe:
> > Sounds like you need to learn about `PIP_FIND_LINKS`?
> 
> Can't find it anywhere. If it's a pip thing, let me remind you that we agreed the aim is to pep-517 build, not "pip" build.

PEP-518 does not specify how exactly the declared requirements are to be satisfied. 
It is the implementation of the frontend that does.
pypa/build uses pip to provision the declared requirements.
So the environment variables equivalent to pip command-line options can be used to control stuff.

mkoeppe:
You can see an example of this in https://github.com/sagemath/sage/blob/develop/build/pkgs/sagemath_objects/spkg-install.in

PR #37291, opened 2024-02-11
Support optional debugpy

orlitzky:
Gentoo is removing it, but since only ipykernel has it as a dependency, we can skip debugpy entirely when ipykernel from the system is installed. This lets the system ipykernel package decide if it wants to pull in debugpy.

### DIFF
--- a/build/pkgs/debugpy/spkg-configure.m4
+++ b/build/pkgs/debugpy/spkg-configure.m4
@@ -1,1 +1,14 @@
-SAGE_SPKG_CONFIGURE([debugpy], [SAGE_PYTHON_PACKAGE_CHECK([debugpy])])
+SAGE_SPKG_CONFIGURE([debugpy], [
+  SAGE_PYTHON_PACKAGE_CHECK([debugpy])
+], [
+  dnl REQUIRED-CHECK
+  dnl
+  dnl Skip debugpy if ipykernel from the system will be used.
+  dnl This allows downstream packagers to treat debugpy (a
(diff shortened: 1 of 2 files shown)


build/pkgs/debugpy/spkg-configure.m4
```diff
@@ -1 +1,14 @@
-SAGE_SPKG_CONFIGURE([debugpy], [SAGE_PYTHON_PACKAGE_CHECK([debugpy])])
+SAGE_SPKG_CONFIGURE([debugpy], [
+  SAGE_PYTHON_PACKAGE_CHECK([debugpy])
+], [
+  dnl REQUIRED-CHECK
+  dnl
+  dnl Skip debugpy if ipykernel from the system will be used.
+  dnl This allows downstream packagers to treat debugpy (a
+  dnl somewhat problematic package) as optional.
```
mkoeppe, line 8: include a link to a discussion what makes it problematic?

orlitzky: There isn't much of a discussion, only a comment in `package.mask` that says "Very fragile, keeps breaking." But IIRC it's from all the vendored junk under https://github.com/microsoft/debugpy/tree/main/src/debugpy/_vendored/pydevd

dimpase (approved):
lgtm

dimpase:
update the branch please

orlitzky:
@dimpase OK to set to positive review?

dimpase:
done.

PR #37261, opened 2024-02-08
`README.md`: Improve PyPI install instructions

mkoeppe:
Based on https://groups.google.com/g/sage-devel/c/1viBzw-ZaoQ/m/ZBbjHXGEAAAJ, we improve the section on installation from PyPI.

We also improve **sage_conf** so that `python3 -m sage_conf` is equivalent to `sage-config`. 
After the next stable release is out, we can use this to simplify these installation instructions.

Blocker because we need this to go in the 10.3 stable release.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37292, opened 2024-02-11
pkgs/sagemath-standard: Remove self-reference at build time

mkoeppe:
As explained in https://github.com/sagemath/sage/pull/37287#issuecomment-1937204900, we make **sagemath-environment** a build requirement of  `pkgs/sagemath-standard`.

The monolithic build system in `SAGE_ROOT/src` is left unchanged.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37261
- Depends on #37432
- Depends on #37434
- Depends on doing something about this kernelspec install thing
- Depends on #38200


tobiasdiez:
With these changes, how do you pip install the monolithic sagelib (ie the one independent of any modularized distributions)?

mkoeppe:
> With these changes, how do you pip install the monolithic sagelib

There is no change regarding this.

tobiasdiez:
> > With these changes, how do you pip install the monolithic sagelib
> 
> There is no change regarding this.

It seems that only pkgs/sagemath-standard is published on pypi, not the one in src. And since you break the smylink it now matters.

mkoeppe:
> pkgs/sagemath-standard is published on pypi

That's right.

> it now matters.

*What* matters?

tobiasdiez:
It matters for the following. Right now it is possible to install the monolithic sagelib distribution via pip, with your changes this is no longer possible.

> > With these changes, how do you pip install the monolithic sagelib
> 
> There is no change regarding this.

mkoeppe:
> Right now it is possible to install the monolithic sagelib distribution via pip, with your changes this is no longer possible.

Nonsense.

mkoeppe:
@kiwifb Please review

kiwifb (commented):
Looks good to me. It will not change anything from the sdist point of view as far as I can tell. If we are going to split things, I certainly want us to eat our own dog food and see stuff like pyproject-monolithic.toml.m4 going out as quickly as possible.

mkoeppe:
> It will not change anything from the sdist point of view as far as I can tell.

That's right, avoiding the self-ref is just a cleanup.

tobiasdiez:
> > Right now it is possible to install the monolithic sagelib distribution via pip, with your changes this is no longer possible.
> 
> Nonsense.

You then might want to explain how a *user* would install it then from pypi (that you can still checkout the code and use the pyproject.toml from src is obvious).

mkoeppe:
#37261 has updated the PyPI install instructions in our README.md, waiting for review.

I'll mark it as a dependency

orlitzky:
This is now obsolete as we are not building any of these distributions.

PR #37299, opened 2024-02-11
use parent in quaternions

fchapoton:
replace the auld-style class by `Parent` and categories for quaternion algebras.

### :memo: Checklist


src/sage/categories/algebras.py
```diff
@@ -144,6 +144,84 @@ def Supercommutative(self):
...
+                sage: L.has_standard_involution()
+                Traceback (most recent call last):
+                ...
+                NotImplementedError: has_standard_involution is not implemented for this algebra
+                """
+            field = self.base_ring()
+            try:
+                basis = self.basis()
+            except AttributeError:
+                raise AttributeError("Basis is not yet implemented for this algebra.")
```
tscrim: A while-we-are-at-it thing:
```suggestion
                raise AttributeError("basis is not yet implemented for this algebra")
```

src/sage/categories/algebras.py
```diff
@@ -144,6 +144,84 @@ def Supercommutative(self):
...
+                sage: # needs sage.modules
+                sage: ZZ.characteristic()
+                0
+                sage: A = GF(7^3, 'a')                                                      # needs sage.rings.finite_rings
+                sage: A.characteristic()                                                    # needs sage.rings.finite_rings
+                7
+            """
+            return self.base_ring().characteristic()
+
+        def has_standard_involution(self):
```
tscrim, line 164: Should we move this method to `AlgebrasWithBasis` instead? Then we can assume it has a `basis()` method.

tscrim (changes requested):
Just some little things. Otherwise LGTM.

src/sage/categories/algebras.py
fchapoton: not here and now. This would require making quaternion algebras to be `WithBasis` (they are not) and this takes some work

tscrim: Understood, no problem.

tscrim (approved):

tscrim:
LGTM.

PR #37303, opened 2024-02-12
use Parent in Weyl algebra

fchapoton:
replace the auld-class `Algebra` by `Parent` in Weyl algebras

this requires a tweak in the `_first_n_gens` mechanism to handle families

### :memo: Checklist

### DIFF
--- a/src/sage/algebras/weyl_algebra.py
+++ b/src/sage/algebras/weyl_algebra.py
@@ -654,7 +654,7 @@ class DifferentialWeylAlgebraElement(Element):
         return self.parent().diff_action(self, p)
 
 
-class DifferentialWeylAlgebra(Parent, UniqueRepresentation):
+class DifferentialWeylAlgebra(UniqueRepresentation, Parent):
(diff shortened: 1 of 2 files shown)


src/sage/algebras/weyl_algebra.py
```diff
@@ -655,7 +654,7 @@ def diff(self, p):
         return self.parent().diff_action(self, p)
 
 
-class DifferentialWeylAlgebra(Algebra, UniqueRepresentation):
+class DifferentialWeylAlgebra(Parent, UniqueRepresentation):
```
tscrim:
```suggestion
class DifferentialWeylAlgebra(UniqueRepresentation, Parent):
```
A while-we-are-at-it thing: better (well, mostly safer) for UR to be first in the MRO.

tscrim (approved):
Mostly a triviality. Whether you change it or not, you can set a positive review once decided.

tscrim:
Thank you.

PR #37311, opened 2024-02-13
Add tips for choosing an exception in developer guide

kwankyu:
as discussed in https://groups.google.com/g/sage-devel/c/xzp0sxinCW8/m/RLZ4vGdPAAAJ

### :memo: Checklist


src/doc/en/developer/coding_in_python.rst
```diff
@@ -473,6 +473,41 @@ example:
 Note that the syntax in ``except`` is to list all the exceptions that
 are caught as a tuple, followed by an error message.
 
+If you are writing new code to Sage, you soon face the problem of choosing an
+exception most suitable for an errorneous situation. This is often a tricky
+business. The following aims to guide you in choosing from the most relevant
+exceptions to Sage developers. A method (or a function) accepts input described
+in the ``INPUT`` block of :ref:`the docstring <section-docstring-function>` and
+returns output. You may raise
```
mantepse: (Thank you!!!)

I'd suggest to shorten this significantly and slightly rearrange, for example:

* remove the first two sentences (in particular "errorneous situation" is probably misleading)
* move the sentence "A method ... output" to the beginning to the paragraph.  After that, you could continue with "If the input can not be handled by the code, it may raise an exception.", and then "The following...".

src/doc/en/developer/coding_in_python.rst
```diff
@@ -473,6 +473,41 @@ example:
...
+
+- `TypeError <https://docs.python.org/3/library/exceptions.html#TypeError>`_:
+  if the input belongs to a class of objects that are not supported by the
+  method.  For example, a method works only with monic polynomials over a
```
mantepse, line 485: maybe drop "monic" here?

src/doc/en/developer/coding_in_python.rst
kwankyu: Good idea. Done.

src/doc/en/developer/coding_in_python.rst
```diff
@@ -473,6 +473,40 @@ example:
...
+  if the input has a value not supported by the method. For example, the above
+  method was given a non-monic polynomial.
+
+- `ArithmeticError
+  <https://docs.python.org/3/library/exceptions.html#ArithmeticError>`_: if the
+  method performs an arithmetic operation (sum, product, quotient, and the
+  like) but the input is not appropriate.
+
+- `ZeroDivisionError
+  <https://docs.python.org/3/library/exceptions.html#TypeError>`_: if the
```
grhkm21:
```suggestion
  <https://docs.python.org/3/library/exceptions.html#ZeroDivisionError>`_: if the
```

wrong link

grhkm21:
Otherwise looks good to me - @mantepse do you agree? 😂

src/doc/en/developer/coding_in_python.rst
```diff
@@ -473,6 +473,40 @@ example:
 Note that the syntax in ``except`` is to list all the exceptions that
 are caught as a tuple, followed by an error message.
 
+A method or a function accepts input described in the ``INPUT`` block of
+:ref:`the docstring <section-docstring-function>` and returns output. If the
```
mantepse: I'd suggest to remove "and returns output", sorry I did not notice earlier.

mantepse:
In principle I agree, but I think it would be good to post the snippet on sage-devel again, both for advertising and for consensus.

kwankyu:
Instead of carrying around the code, I posted on sage-devel an invitation to come and review this PR.

mezzarobba:
Re `ArithmeticError` vs `ZeroDivisionError` for non-invertible elements of rings: a possible argument for raising a `ZeroElementError` would be that, pretty often, if you don't test upfront that the element is invertible and just apply whatever inversion algorithm you have instead, you will indeed end up trying to divide by the zero element of a field. Yet to *catch* an exception due to something not being invertible I guess it is more robust to test for an `ArithmeticError`, and then since `ZeroDivisionError` derives from `ArithmeticError` it doesn't really matter which one is thrown...

kwankyu:
> ... Yet to _catch_ an exception due to something not being invertible I guess it is more robust to test for an `ArithmeticError`, and then since `ZeroDivisionError` derives from `ArithmeticError` it doesn't really matter which one is thrown...

Ah, thanks! That is a good point. Perhaps we may mention that in the guide.

src/doc/en/developer/coding_in_python.rst
```diff
@@ -473,6 +473,42 @@ example:
 Note that the syntax in ``except`` is to list all the exceptions that
 are caught as a tuple, followed by an error message.
 
+A method or a function accepts input described in the ``INPUT`` block of
+:ref:`the docstring <section-docstring-function>`. If the input cannot be
+handled by the code, then it may raise an exception.  The following aims to
+guide you in choosing from the most relevant exceptions to Sage. Raise
+
+- `TypeError <https://docs.python.org/3/library/exceptions.html#TypeError>`_:
```
mkoeppe: You can just use `` :class:`TypeError` `` to refer to these Exception classes in the standard library

mkoeppe (approved):
That's an exceptionally nice addition~

kwankyu:
Thanks for all the comments that improved this PR!

PR #37316, opened 2024-02-13
use CommutativeRing in ring_extension

fchapoton:
Removing one of the last uses of `CommutativeAlgebra` class, towards its deprecation.

Also a few pep8 fixes in the modified file.

### :memo: Checklist


mantepse:
Does this actually mean that inheriting from `CommutativeAlgebra` doesn't do anything beyond making
```
sage: isinstance(K, CommutativeAlgebra)
True
```
?

More precisely, I tried:
```
sage: F = GF(5^2)
sage: k = GF(5^4)
sage: z4 = k.gen()
sage: K.<a> = k.over(F, gen=1-z4); K
Field in a with defining polynomial x^2 + z2*x + 4 over its base
sage: K in Algebras(K.base_ring())
False
sage: from sage.rings.ring import CommutativeAlgebra
sage: isinstance(K, CommutativeAlgebra)
True
```
and with the new branch, the last line is `False` (which looks completely OK, of course, I am just surprised).

(I admit that I'm not completely sure what `ring_extension.pyx` actually does)

fchapoton:
yes, probably your are right. The usage of `isinstance` should be avoided  and categories should be used instead.

My aim for now is to keep only the bad class `CommutativeRing` and deprecate several others auld-parent classes like `Algebra`, `CommutativeAlgebra`, etc. Then and only then, I will try to disentangle what happens for all commutative rings (and this will be  a huge mess)

The change here is a step towards deprecating `CommutativeAlgebra`, which is almost no longer used already.

mantepse (approved):
Thank you!

PR #37318, opened 2024-02-13
Allowing HeckeAlgebraSymmetricGroupT to accept elements of SymmetricGroup(n) as input

AndrewMathas:
Currently, the following fails in sage:
```
sage: S4 = SymmetricGroup(4)
sage: H4 = HeckeAlgebraSymmetricGroupT(ZZ, 4)
sage: H4( Permutations(4).an_element() )
T[4, 1, 2, 3]
sage: H4( S4.an_element() )
...
TypeError: do not know how to make x (= (2,3,4)) an element of self (=Hecke algebra of the symmetric group of order 4 on the T basis over Univariate Polynomial Ring in q over Integer Ring)
```
I view this as a bug because the elements of this algebra are indexed by elements of the symmetric group and `HeckeAlgebraSymmetricGroupT` does accept elements of `Permutations(n)`.

This PR makes works work as well as accepting elements of `Permutations(N)`, for `N>n, that fix `n+1,...,N`.

Previously, `HeckeAlgebraSymmetricGroupT` contain a `_coerce_start` method. I have renamed this as `_element_constructor_`, which is what I think it should be, and made some minor changes there. Probably, I should deprecate `_coerce_start` but this method does not seem to exist anywhere else in sage, so I suspect that this belongs to some code that was deprecated years ago.

### :memo: Checklist


src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -3065,24 +3065,34 @@ def q(self):
...
         """
         EXAMPLES::
 
             sage: H3 = HeckeAlgebraSymmetricGroupT(QQ, 3)
-            sage: H3._coerce_start([2,1])
+            sage: H3([2,1])  # indirect doc test
+            T[2, 1, 3]
+            sage: H3( SymmetricGroup(3).an_element() )
+            [1, 3, 2]
+            sage: H3( [2, 1] )
```
GiacomoPope: The test for `H3([2,1])` appears twice

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -3065,24 +3065,34 @@ def q(self):
...
+            T[2, 1, 3]
+            sage: H3( [2, 1, 3, 4, 5, 6] )
             T[2, 1, 3]
         """
```
GiacomoPope, line 3088: Maybe add examples for both `SymmetricGroup(3).an_element()` and `Permutations(3).an_element()` to match the description in this PR?

AndrewMathas: Permutations are not new and are, in effect, tested on line 3077, but I agree so I have added 
```
H3( Permutation(3).an_element() )
```
and I have removed the duplicate test for `H3([2,1])`.

GiacomoPope (approved):
Looks good to me

PR #37321, opened 2024-02-13
Allow factoring `Factorization` object

grhkm21:
Currently, `factor(factor(12))` fails. This is annoying since some functions allow the user to provide a pre-factored object, resulting in code like this:

```python
def func(x):
    if not isinstance(x, Factorization):
        x = x.factor()
    ...
```

This PR fixes these.


grhkm21:
I'm pretty sure the "merging 'develop' to this branch" action caused the linter to fail... this is very weird. I will fix it.

Pietro-D:
The operation `factor(factor(12))` fails because the factor() function accepts only the following inputs 


    -  ``n`` -- a nonzero integer

    -  ``proof`` -- bool or ``None`` (default: ``None``)

    -  ``int_`` -- bool (default: ``False``) whether to return
       answers as Python ints

    -  ``algorithm`` -- string

       - ``'pari'`` -- (default) use the PARI c library

       - ``'kash'`` -- use KASH computer algebra system (requires that
         kash be installed)

       - ``'magma'`` -- use Magma (requires magma be installed)

    -  ``verbose`` -- integer (default: 0); PARI's debug
       variable is set to this; e.g., set to 4 or 8 to see lots of output
       during factorization.`
To be honest I don't see how this operation can yield any additional information or meaningful result.  To remove this error maybe we could  just manage this case inside `arith/misc.py/factor`  by adding the following check:
`if isinstance(n, Factorization):
        return  n`

grhkm21:
Not sure what you are talking about.

Pietro-D:
We can handle the case where the input is a Factorization object within the function body
`arith/misc.py/factor`. I believe this makes the code cleaner.

src/sage/groups/generic.py
mezzarobba, line 1: It would be nice to have a test for the case where `n` is an `int` here. And I would have called `py_scalar_to_element` directly instead of changing the `.factor()` method call to `factor(...)`, but I suppose it's a matter of taste. Otherwise lgtm!

grhkm21: Thanks for the review, please kindly take a look again :)

grhkm21:
> We can handle the case where the input is a Factorization object within the function body `arith/misc.py/factor`. I believe this makes the code cleaner.

I decided not to do this since `a.factor()` is a common way to write it in Sage, and your suggested fix will still fail when `a` is a `Factorization`

yyyyx4:
It is possible to construct things like `Factorization([(9,1),(12,1)])`. I'd argue that the semantics of calling `.factor()` on this object should be to return a full factorization (i.e., `2^2 * 3^3`).

mezzarobba:
On Wednesday, 21 February 2024 03:00:48 CET Lorenz Panny wrote:
> It is possible to construct things like
> `Factorization([(9,1),(12,1)])`. I'd argue that the semantics of
> calling `.factor()` on this object should be to return a full
> factorization (i.e., `2^2 * 3^3`).

Ah, good point! One should be able for instance to convert a squarefree 
factorization into a full factorization this way.

grhkm21:
@mezzarobba Do you mind reviewing? It's a simple extension to the previous commits.

mezzarobba:
My first reaction was that `simplify` should be `True` by default, so that one can pass a known partial factorization to a function that calls `factor()` (and hence presumably requires a full factorization) and have it work as expected. However:
* this means paying the cost of checking that the factors are irreducible every time,
* sometimes you have a partial factorization but you want to pretend that the factors are irreducible.

These issues could be solved by passing additional parameters to functions that manipulate factorizations or to the constructor of `Factorization`, but this looks like over-engineering. So I'm not sure what to say. @yyyyx4, what do you think?

grhkm21:
I think he is busy, and you mentioned two of my main points too. A lot of times having a partial factorisation is good enough. Also, this method is so that people can pass a pre-factored object into a method that requires factorisation, so they probably know what they're doing.

mezzarobba:
Hmm, I am still not completely comfortable with having a `factor()` method that does not factor by default. Maybe having a constructor parameter (say `frozen=False`, for lack of a better name) instead of the `simplify` parameter in `factor()` isn't that bad after all. This way users who know what they are doing can create their factorization objects with `frozen=True` and get the behavior you want, but the default behavior would still be the least surprising one. What do you think?

src/sage/groups/generic.py
```diff
@@ -1481,12 +1483,18 @@ def has_order(P, n, operation='+'):
...
     """
+    from sage.arith.misc import factor
+
     if isinstance(n, sage.rings.integer.Integer):
         if n <= 0:
             return False
         n = n.factor()
+    elif isinstance(n, int):
+        if n <= 0:
+            return False
```
vincentmacri, line 1494: Add a test covering this case.

src/sage/structure/factorization.py
```diff
@@ -755,6 +755,51 @@ def unit(self):
         """
         return self.__unit
 
+    def factor(self, simplify=False):
```
vincentmacri, line 758: I agree with having `simplify` default to `True`. While I can see there are some cases where you might be fine with a partial factorization, giving a full factorization seems to be the most reasonable default behaviour, and code that is fine with a partial factorization can specify `simplify=False`.

PR #37324, opened 2024-02-13
Basis completion for matrices over univariate polynomials

vneiger:
This PR (the fruit of discussions with @xcaruso during SageDays 125) introduces a new functionality for matrices over univariate polynomials: completing bases. A description of what this means is below: this is a direct copy-paste of the doc strip written for the function.

Two algorithms are offered: the one which one might call "naive", based on Smith form computation (requiring also one of the unimodular transformations); the other one is based on an algorithm by Wei Zhou and George Labahn ([Proceedings of ISSAC, 2014, Unimodular completion of polynomial matrices](https://dl.acm.org/doi/10.1145/2608628.2608640)).

The default has been put to the latter, which is faster in most circumstances, sometimes significantly (*). In fact, it does not exploit fast polynomial multiplication (it should, but does not in the current state of Sage), so it is the case that the Smith form variant is faster in a restricted range of matrices of small dimensions and medium-large degrees. If there is interest in efficiency of this computation, one could work out some thresholds to switch between the two algorithms; but a much more impactful move would be to incorporate fast polynomial arithmetic in approximant bases and kernel bases, which would dramatically speed up Zhou and Labahn's algorithm.

(*) Measured, for example: speed-up x20 for completing a 20x40 matrix of degree 5, 10, 20, or 40; speed-up about x200 for completing a 40x80 matrix of degree 5 or 10.

Here [basis_completion.zip](https://github.com/sagemath/sage/files/14272879/basis_completion.zip) is an archive with one file with sage code to help if one wishes to run somewhat extensive tests as well as benchmarks, and two files showing the results of some benchmarks.

```
def basis_completion(self, row_wise=True, algorithm="approximant"):
        Return a Smith form-preserving nonsingular completion of a basis of
        this matrix: row-wise completing a row basis if ``row_wise`` is True;
        column-wise completing a column basis if it is False.

        For a more detailed description, consider the row-wise case (the
        column-wise case is the same up to matrix transposition). Let `A` be
        the input matrix, `m \times n` over univariate polynomials
        `\Bold{K}[x]`, for some field `\Bold{K}`, and let `r` be the rank of
        `A`, which is unknown a priori. This computes a matrix `C` of
        dimensions `(n-r) \times n` such that stacking both matrices one above
        the other, say `[[A],[C]]`, gives a matrix of maximal rank `n` and with
        the same nontrivial Smith factors as `A`. In particular, `C` has full
        row rank, and the rank of the input matrix may be recovered from the
        number of rows of `C`.

        As a consequence, if `B` is a basis of the module generated by the rows
        of `A` (for example `B = A` if `A` has full row rank), then `[[B],[C]]`
        is nonsingular, and its determinant is the product of the nonzero Smith
        factors of `A` up to multiplication by a nonzero element of `\Bold{K}`.

        In particular, for `A` with full row rank: if the rows `A` can be
        completed into a basis of `\Bold{K}[x]^{n}` (or equivalently, `A` has
        unimodular column bases, or also, if the rows of `A` generate all
        polynomial vectors in the rational row space of `A`), then `C` provides
        such a completion. In this case, `[[A],[C]]` is unimodular: it is
        invertible over `\Bold{K}[x]`, and `det([[A],[C]])` is a nonzero
        element of the base field `\Bold{K}`.
```


src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+
+    def _basis_completion_via_reversed_approx(self):
+        r"""
+        Return a Smith form-preserving nonsingular completion of a row basis of
+        this matrix. For a more detailed description, see
+        :meth:`basis_completion`, in the row-wise case.
+
+        EXAMPLES::
+
+            sage: pR.<x> = GF(7)[]
```
xcaruso: Something is missing here...

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+            else :
+                C = matrix.identity(ring, m)[:,rk:]
+                UU = U.inverse_of_unit()
+                return UU * C
+
+        else :
+            raise ValueError("algorithm must be one of \"approximant\" or \"smith\".")
+
+
+    def is_basis_completion(self, mat, row_wise=True):
```
xcaruso: OK for keeping this method but I cannot really see a use case.
Could you please explain a bit why it is important?

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+            sage: pR.<x> = GF(7)[]
+        """
+        from sage.matrix.constructor import matrix # for identity
+
+        ring,var = self.base_ring().objgen()
+        m = self.nrows()
+        n = self.ncols()
+
+        # corner cases: after this, m>0 and n>0
+        if m == 0 :
```
xcaruso: I think you should remove the space before `:` here (and everywhere else in your ticket).

xcaruso (commented):
Great job! Thanks, Vincent!

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+        zcols = []
+        nonzcols = []
+        for j in range(n):
+            if cdeg[j] < 0:
+                zcols.append(j)
+            else :
+                nonzcols.append(j)
+
+        if len(nonzcols) == 0 :
+            return matrix.identity(ring, n).matrix_from_rows(zcols)
```
grhkm21:
```suggestion
        if len(nonzcols) == 0 :
            return matrix.identity(ring, n)
```

If `nonzcols` is empty, then `zcols == list(range(n))`, so the `.matrix_from_rows` part is nop.

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+        element of the base field `\Bold{K}`.
+
+        INPUT:
+
+        - ``row_wise`` -- (optional, default: ``True``) boolean, if ``True`` then
+          compute a row-wise completion, else compute a column-wise completion.
+
+        - ``algorithm`` -- (optional, default: ``approximant``) selects the
+          approach for computing the completion; currently supported:
+          ``approximant`` and ``smith``.
```
grhkm21:
```suggestion
        - ``algorithm`` -- (optional, default: ``"approximant"``) selects the
          approach for computing the completion; currently supported:
          ``"approximant"`` and ``"smith"``.
```

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+          which a completion is readily obtained.
+
+        EXAMPLES:
+
+        Three polynomials whose GCD is `1` can be completed into a unimodular
+        matrix::
+
+            sage: ring.<x> = GF(7)[]
+            sage: mat = matrix([[x*(x-1)*(x-2), (x-2)*(x-3)*(x-4), (x-4)*(x-5)*(x-6)]])
+            sage: print(mat)
```
grhkm21:
```suggestion
            sage: mat
```
You don't need to `print` in Sage REPL. This applies to the entire docstring.

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+        The following matrix has rank `2` and trivial Smith form. It can be
+        completed row-wise into a `3 \times 3` unimodular matrix (column-wise,
+        there is nothing to complete)::
+
+            sage: mat = matrix(ring, 2, 3, \
+                    [[x^2 + 5*x + 5,   3*x^2 + x + 3, 4*x^2 + 5*x + 4], \
+                     [5*x^2 + 4*x,   3*x^2 + 4*x + 5, 5*x^2 + 5*x + 3]])
+            sage: rcomp = mat.basis_completion(); print(rcomp)
+            [  2*x^2 + 1 4*x^2 + 3*x 2*x^2 + 3*x]
+            sage: mat.stack(rcomp).determinant()
```
grhkm21, line 4332: Perhaps I misunderstood something, but shouldn't a unimodular matrix have determinant $\pm 1$? Here $3 \not \equiv \pm 1 \pmod{7}$.

Also, it is a good idea to have an example over $\mathbb{Q}$ or a number field (IMO).

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,462 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+                UU = U.inverse_of_unit()
+                return UU * C
+
+        else :
+            raise ValueError("algorithm must be one of \"approximant\" or \"smith\".")
+
+
+    def is_basis_completion(self, mat, row_wise=True):
+        r"""
+        Return true if and only if this matrix is a basis completion for
```
grhkm21:
```suggestion
        Return whether this matrix is a basis completion for
```
(please line wrap for me)

src/sage/matrix/matrix_polynomial_dense.pyx
grhkm21: (To just add on that, if method is useful internally, name it `_is_basis_completion`.

grhkm21 (commented):
General feedbacks. I did not look too closely at the algorithm since I am not familiar with it.

Also, is it possible to simplify the code for column-wise computations by taking transposes? As in completing the basis completion column-wise = transpose (completing the transpose(basis) row-wise).

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: Actually, I wrote this function for testing purposes while writing the ``basis_completion`` method. I agree it should be ``_is_basis_completion`` (I have just done this and pushed), or even removed (to be discussed). I admit that I have no clear use case in mind at the moment, but it is unclear to me that it is useless either: there may be contexts where one may end up one way or another with a candidate completion, and would like to know if it is indeed a completion or not; since completions are not unique, it is not trivial to use ``basis_completion`` to check this. Now that it is there, and tested, what would be reasons for discarding it?

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: Oops... had forgotten the examples on this one, thanks. I have added them, taking from the main method ``basis_completion``. This made me realise also that in the latter, I should (and will) add more tests related to the Smith algorithm (not necessarily in EXAMPLES, but at least in TESTS).

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: Done, thanks!

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: Absolutely correct, thank you!

vneiger:
> General feedbacks. I did not look too closely at the algorithm since I am not familiar with it.

Thanks a lot for your useful feedback!
 
> Also, is it possible to simplify the code for column-wise computations by taking transposes? As in completing the basis completion column-wise = transpose (completing the transpose(basis) row-wise).

This is indeed what is done for the ``approximant`` variant, in the first lines of code for ``basis_completion``. For performance (time and memory), this is not always a good strategy, but here it is acceptable since the approximant and kernel subroutines are far from optimized anyway. For the Smith-based variant, since handling both cases explicitly added only a handful of lines, I preferred to do so (without resorting to transposes).

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: I might be mistaken but I think the general definition is that a (square) matrix over some ring is said to be unimodular if it is invertible (over that ring), or equivalently that its determinant is a unit (in the ring). For univariate polynomials, units are all nonzero constants; the set of units being +-1 would be e.g. for relative integers.

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,525 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+            []
+            sage: matrix(ring, 0, 2)._basis_completion_via_reversed_approx()
+            [1 0]
+            [0 1]
+            sage: matrix(ring, 2, 0)._basis_completion_via_reversed_approx()
+            []
+        """
+        from sage.matrix.constructor import matrix # for identity
+
+        ring,var = self.base_ring().objgen()
```
xcaruso:
```suggestion
        ring, var = self.base_ring().objgen()
```

xcaruso (commented):
a small stuff

vneiger:
> I'm happy with your code and ready to give a positive review when tests pass.

Nice! I have a couple of examples to add (over rationals, and using Smith-based algorithm), hopefully this will be here and tested by the end of tomorrow.

src/sage/matrix/matrix_polynomial_dense.pyx
grhkm21: Ahh okay, you are correct, wikipedia is wrong (as always). I think I had this exact same question and discussion at least 5 times in the past year...Ahh okay, that makes sense, wikipedia is wrong (as always). I think I had this exact same question and discussion at least 5 times in the past year...

xcaruso:
Sure, I'm waiting :-)

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: examples added!

vneiger:
Examples added, over the rationals and over a number field, and using mostly the ``algorithm="smith"`` variant.

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,564 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
                     column_indices.append(j)
             return P[:,column_indices]
+
+    def _basis_completion_via_reversed_approx(self):
+        r"""
+        Return a Smith form-preserving nonsingular completion of a row basis of
+        this matrix. For a more detailed description, see
+        :meth:`basis_completion`, in the row-wise case.
+
+        EXAMPLES::
```
grhkm21:
```suggestion
        EXAMPLES:
```

This should be single colon since there are no tests *immediately* after this indicator. The double colon below (... unimodular matrix::) is correct.

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,564 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+
+            sage: matrix(ring, 0, 0)._basis_completion_via_reversed_approx()
+            []
+            sage: matrix(ring, 0, 2)._basis_completion_via_reversed_approx()
+            [1 0]
+            [0 1]
+            sage: matrix(ring, 2, 0)._basis_completion_via_reversed_approx()
+            []
+        """
+        from sage.matrix.constructor import matrix # for identity
```
grhkm21:
```suggestion
        from sage.matrix.constructor import matrix  # for identity
```
Two spaces for inline comments ([PEP8](https://peps.python.org/pep-0008/#inline-comments) or `pycodestyle` E225)

grhkm21 (changes requested):
Sorry for not spotting these earlier. Just two small changes!

I thought `rst` is supposed to detect the first one, but I guess not.

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: thanks!

src/sage/matrix/matrix_polynomial_dense.pyx
vneiger: Oh, I didn't know about this, thanks

grhkm21: To be fair, this is not required by the Sage lint CI, which only requires the lint checks specified by `pycodestyle-minimal` [here](https://github.com/sagemath/sage/blob/develop/src/tox.ini#L138). But when possible, new code should follow PEP8 - at least the ones I find nice :)

src/sage/matrix/matrix_polynomial_dense.pyx
```diff
@@ -4079,3 +4084,564 @@ cdef class Matrix_polynomial_dense(Matrix_generic_dense):
...
+            [1/2*x - 1/2]
+            [  1/2*x - 1]
+            [1/2*x - 3/2]
+            sage: ccomp.augment(mat).smith_form(transformation=False)
+            [    1     0     0]
+            [    0     1     0]
+            [    0     0 1/2*x]
+
+            sage: ring.<y> = NumberField(x**2 - 2, "a")[]
+            sage: a = ring.base_ring().gen()
```
xcaruso:
```suggestion
            sage: K.<a> = NumberField(x^2 - 2)
            sage: S.<y> = K[]
```
I find this more readable.

vneiger: yes, thanks for the suggestion, I've done this (up to renaming K and S to variable names more consistent with the things around).

xcaruso:
Difficult to know if checks have passed or not, but I nevertheless give a positive review.
Please remove it if you disagree.

vneiger:
> Difficult to know if checks have passed or not, but I nevertheless give a positive review. Please remove it if you disagree.

Thanks! I don't disagree. There is a conda build&test failure, but the error message seems to indicate it does not come from this PR. Other than that, there is a "Build & Test / build (pull_request) Successful in 154m". Another confidence helper is that the tests did go through on my machine: I only tested the matrices directory, but since the only changes are in new functionalities which are not yet used in other functions, there should not be impacts on other tests.

grhkm21:
I have never seen that Conda 3.9 test pass in the past few months(?)

PR #37325, opened 2024-02-13
Remove emptyset from empty SimplicialComplex

gmou3:
Allow a `SimplicialComplex` to be faceless, rather than always forcing it to include the `emptyset`.

```
sage: SimplicialComplex([])  # no faces
Simplicial complex with vertex set () and facets {}
sage: SimplicialComplex([()])  # one face
Simplicial complex with vertex set () and facets {()}
```
Previously the result of these commands was the same (with `facets {()}`). This change is advantageous in cases where a computation should result in the empty complex instead of the one with one face; an example of this is the broken circuit complex of a matroid with a loop.


src/sage/topology/cell_complex.py
```diff
@@ -165,7 +165,7 @@ def dimension(self):
             4
         """
         try:
-            return max([x.dimension() for x in self._facets])
+            return max([-2] + [x.dimension() for x in self._facets])
         except AttributeError:
```
fchapoton, line 169: this could use `max(iterator, default=-2)`

gmou3: Thanks; that's better.

tscrim:
@jhpalmieri You might have a comment about this. I am not sure if this is desirable from the (co)homology perspective.

jhpalmieri:
Another option would be to allow two different objects: the simplicial complex containing only the empty simplex, and also the simplicial complex containing no simplices. Currently we only have the first, and with this change we would only have the second. Why not allow both?

jhpalmieri:
It seems to me that in the setting with no simplices, the `chain_complex` method should probably not allow `augmented=True` — there is no empty simplex to create the augmentation. Similarly, `_homology_` should not allow `reduced=True`. I don't know if there are other things that will have to be special-cased. I'm speaking as an algebraic topologist, though, and I have no experience with this very empty simplicial complex, so maybe I'm misunderstanding something.

gmou3:
I think there is no need to adjust these functions as long as they still work fine for the non-(very empty) simplicial complexes (the tests look fine). I only wish to allow this very empty complex for the (somewhat degenerate) case where a computation yields an empty collection of sets instead of a nonempty simplicial complex.

In any case, no problem if you are ambivalent about this; we can always cancel the change and I'll just add a few ifs elsewhere to handle this.

tscrim:
Thinking more about this, I do not believe the default should be the "very" empty simplicial complex. All of our non-empty simplicial complexes also come with the empty set/simplex as a $-1$ dimensional face:
```
sage: Y = SimplicialComplex([[1,2]])
sage: Y.faces()
{-1: {()}, 0: {(1,), (2,)}, 1: {(1, 2)}}
```
This is very natural from the theory (speaking as a combinatorialist here, we want the faces under inclusion to be a nice poset).

We also need to be careful about the output of the methods with their possible arguments, not just that they work and doctests pass. For example, the `augmented` for the `chain_complex()` has a subtlety as @jhpalmieri mentioned. The augmentation is there to represent the empty set, and if the empty set is not in the simplicial complex, it doesn't make sense to have anything other than the trivial chain complex.

I understand your point about having the special case it in other code, but I think handling that will be easier than going though the simplicial complex code and having extra checks for this very empty simplicial complex. It might be useful to have an `is_empty()` method and/or have `__bool__` return `False` for the empty simplicial complex (this can be checked by using `S.dimension() == -1`, but that requires a bit more understanding of the theory that such other methods).

Side remark: the `ascii_art` and `unicode_art` of a trivial chain complex results in an error.

tscrim:
Sorry, I know this makes the stuff your doing with matroids more annoying in these corner cases.

PR #37328, opened 2024-02-14
remove "known bug" marker for #6413 from doctest

yyyyx4:
The long-standing issue #6413 was resolved in #37096, but there is still one `# known bug` marker for it in `src/sage/schemes/elliptic_curves/hom_scalar.py`. Let's remove it.

(Incidentally, the claimed output for the invocation that used to not work was slightly wrong, too.)


grhkm21:
Looks good!

PR #37329, opened 2024-02-14
add optional order= argument to .log() method for PARI finite-field elements

yyyyx4:
When the order of the base element of a finite-field logarithm is already known, factoring the unit-group order can be skipped. PARI's `fflog()` accepts the order of the base element as an optional input, but that argument is currently not exposed in Sage. This patch takes care of it. An example of the speedup is given in the new doctests.


GiacomoPope (commented):
Thank you for adding this into Sage. It's been on my TODO list for a while and should have done it myself. This change looks good to me.

Is it worth doing some kind of naive check that the order supplied is the order of both `self` and the `base`? Maybe with a `check=True` arg if nothing else? Just to catch when users accidentally send the wrong order as a param.

```py
        else:
            base_order = _new_GEN_from_mpz_t((<Integer>order).value)
            if check:
                if not (base**base_order).is_one() and (self**base_order).is_one():
                    raise ValueError
```

yyyyx4:
Thanks! I added a `check=` parameter to toggle checking the provided order of the base element. Checking that `self` lies in the group generated by `base` should not be disabled by `check=False` since it's documented that the method will throw an error on such inputs.

GiacomoPope (approved):
Looks good to me, should check be false by default? If we make it true the function will be a little slower but those who read the documentation can set it to be fast for speedier results?

PR #37331, opened 2024-02-14
 sets/recursively_enumerated_set: fix LaTeX/ Sphinx rendering in forest-structure docs

### LINKED ISSUE #37084: fix documentation of `sets/recursively_enumerated_set.py`
### Steps To Reproduce

The html, as can be seen here, https://doc.sagemath.org/html/en/reference/sets/sage/sets/recursively_enumerated_set.html, does not like the LaTeX environment `picture`.  This seems to be the only occurrence of the `picture` environment in the whole source tree.

There is also an `EXAMPLES:` section at the very beginning, which looks odd because a heading follows.

Also, maybe check for typos, such as here:

> One solution is to take elements of one by one an insert them at every positions. 

Apart from documentation, I am also missing the functionality to find the distance of an element to the seeds.  It is quite possible that this functionality is actually there, but I couldn't find it - but then an example on how to do it would be good.

### Expected Behavior

The html should display nicely.

### Actual Behavior

The html does not display nicely.

### Additional Information

_No response_

### Environment

```markdown
irrelevant.
```

### Checklist

janmenjayap:
Modified the doc in `sets/recursively_enumerated_set.py ` (in the section `Example: Forest structure`) to display the HTML nicely; also corrected some typos.

Sphinx `.. MATH::` directive is unable to render the `picture` environment in LaTeX, as a result of which the image is not begin displayed nicely in [sets/recursively_enumerated_set](https://doc.sagemath.org/html/en/reference/sets/sage/sets/recursively_enumerated_set.html). I generated the pdf and png of the required image by compiling the LaTeX code and attached the image. The inspiration is from `Figure: The five complete binary trees with four leaves` in [combinat/tutorial](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/tutorial.html).

Fixes #37084.

The link of the original documentation -- [sets/recursively_enumerated_set](https://doc.sagemath.org/html/en/reference/sets/sage/sets/recursively_enumerated_set.html).

### :memo: Checklist

- [x]  The title is concise, informative, and self-explanatory.
- [x]  The description explains in detail what this PR is about.
- [x]  I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes. (Not required it seems.)
- [x]  I have updated the documentation accordingly.

### :hourglass: Dependencies

None.


grhkm21:
I'm not sure creating a new media is a good idea, since it's hard to change. Instead, try following how [combinat/bijectionist](https://github.com/sagemath/sage/blob/30b3d78fac8d4a015ad2641cad44c3d530f4c10c/src/sage/combinat/bijectionist.py#L457) does it. You can see the result [here](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/bijectionist.html)

Omegaconstant:
Hi,
I tried recreating the same stuff using arrays in a local latex environment as follows and it worked.

  ```
% LaTeX
  \begin{equation*}
  \begin{array}{rcl}
      % Root and Legend
      & \text{``\ "} \\
      \text{2) A function of an element enumerating}\quad \quad& \scalebox{1.5}{$\bullet$} & \scalebox{2}{$\leftarrow$}\quad\text{1) An initial element} \\
      \text{its children (if any) \hspace{3.75cm}} & & \\
  
      % Arrows
      \scalebox{2}{$\searrow$}\hspace{2cm} \scalebox{2}{$\swarrow$}  & \scalebox{2}{$\downarrow$} & \scalebox{2}{$\searrow$} \\[5pt]
  
      % First Children
      \text{``a"} \hspace{4.5em}& \text{``b"} & \hspace{4.5em}\text{``c"} \\
      \scalebox{1.5}{$\bullet$}\hspace{5em} & \scalebox{1.5}{$\bullet$} & \hspace{5em}\scalebox{1.5}{$\bullet$}\\
  
      % Second Children
      \begin{array}{rcl}
          \scalebox{2}{$\swarrow$}  & \scalebox{2}{$\downarrow$} & \scalebox{2}{$\searrow$} \\[5pt]
          \text{``aa"} \quad& \text{``ab"} & \quad\text{``ac"} \\
          \scalebox{1.5}{$\bullet$}\hspace{1.65em} & \scalebox{1.5}{$\bullet$} & \hspace{1.65em}\scalebox{1.5}{$\bullet$}\\
      \end{array} & 
      \begin{array}{rcl}
          \scalebox{2}{$\swarrow$}  & \scalebox{2}{$\downarrow$} & \scalebox{2}{$\searrow$} \\[5pt]
          \text{``ba"} \quad& \text{``bb"} & \quad\text{``bc"} \\
          \scalebox{1.5}{$\bullet$}\hspace{1.65em} & \scalebox{1.5}{$\bullet$} & \hspace{1.65em}\scalebox{1.5}{$\bullet$}\\
      \end{array} & 
      \begin{array}{rcl}
          \scalebox{2}{$\swarrow$}  & \scalebox{2}{$\downarrow$} & \scalebox{2}{$\searrow$} \\[5pt]
          \text{``ca"} \quad& \text{``cb"} & \quad\text{``cc"} \\
          \scalebox{1.5}{$\bullet$}\hspace{1.65em} & \scalebox{1.5}{$\bullet$} & \hspace{1.65em}\scalebox{1.5}{$\bullet$}\\
      \end{array} \\
  \end{array}
  \end{equation*}
```

Here's the preview for that:
image-array

But, it was kind of difficult to manage the alignments, font size, inverted commas etc in the pyx file itself. I will try to implement it and commit. Thank you.

File: src/sage/sets/recursively_enumerated_set.pyx [hunk 1]
```diff
@@ -9,25 +9,23 @@ sets that are described by some ``seeds`` and a successor function
...
 
-EXAMPLES:
-
 No hypothesis on the structure
 ------------------------------
 
 What we mean by "no hypothesis" is that the set is not known
 to be a forest, symmetric, or graded. However, it may have other
-structure, like not containing an oriented cycle, that does not
+structures, such as not containing an oriented cycle, that do not
```
mantepse: No, it should be "structure, such as ... does" (indeed, I like "such as" better than "like")

mantepse:
Cool!  Could you try to remove the bullet points alltogether?  

Also, the arrow labelled 2) must point to an arrow, not to a vertex of the tree.

In fact, either the two arrows labelled 1) and 2) (which "explain" the diagram) should look very different from the other arrows, or should be removed entirely (together with the text "an initial element" and "a function ...").

I would actually vote for removing them: the concept is not all that difficult to understand to warrant such a complicated figure.

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -406,7 +369,7 @@ def RecursivelyEnumeratedSet(seeds, successors, structure=None,
 
     .. WARNING::
 
-        If you do not set the good structure, you might obtain bad results,
+        If you do not set a good structure, you might obtain bad results,
```
mantepse, line 383: "good" should "valid"

(good is french english :-))

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -1920,8 +1883,8 @@ class RecursivelyEnumeratedSet_forest(Parent):
     def _elements_of_depth_iterator_rec(self, depth=0):
         r"""
         Return an iterator over the elements of ``self`` of given depth.
-        An element of depth `n` can be obtained applying `n` times the
-        children function from a root. This function is not affected
+        An element of depth `n` can be obtained by applying the
+        children function `n` times from the root. This function is not affected
```
mantepse: it is "a root" here, because it is a forest

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -1951,8 +1914,8 @@ class RecursivelyEnumeratedSet_forest(Parent):
     def elements_of_depth_iterator(self, depth=0):
         r"""
         Return an iterator over the elements of ``self`` of given depth.
-        An element of depth `n` can be obtained applying `n` times the
-        children function from a root.
+        An element of depth `n` can be obtained by applying the
+        children function `n` times from the root.
```
mantepse: it is "a root", because it is a forest

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -1574,7 +1537,7 @@ def search_forest_iterator(roots, children, algorithm='depth'):
          [0, 1, 2], [0, 2, 1], [1, 0, 2], [1, 2, 0], [2, 0, 1], [2, 1, 0]]
     """
     # Little trick: the same implementation handles both depth and
-    # breadth first search. Setting position to -1 makes a depth search
+    # breadth first search. Setting the position to -1 makes a depth search
```
mantepse: "position" refers to the variable here, therefore no "the".

However, it would be good to replace "makes" with "initiates" (here and two lines below).

mantepse:
Great job!  Thank you for looking so carefully through the whole text!

Did you use a tool or did you do it by hand?

Omegaconstant:
Thank you for your comments :-)

The initial part was done by me; later I used Grammarly to cross-check things. It suggested some modifications like replacing 'depth first search' with 'depth-first search', 'non negative' with 'non-negative' and so on, which I did not appreciate by going through some other documentation of Sage.

Also, I wanted to say that maybe some Sphinx directive like `.. raw:: latex` (which may at least allow to directly integrate TikZ codes in .pyx/ .py files), `.. raw:: html` and so, should be enabled in Sage, which may reduce the use of media and allow more flexibility :-)

mantepse:
It would be cool if you could finish this, so it makes it into the upcoming release!

Omegaconstant:
Hello,
Apologies for the delay and thank you for your patience.

I have made a new commit with the changes suggested. Just a few observations:
* The final image generated looks as follows: 
recursively_enumerated_set
* Apparently, `\text{``a''}` does not act as expected and generates the quotes as follows: 
a
* Also, the orientation and length of `\swarrow ` and `\searrow` in the first set of arrows do not seem quite correct; however, `\rotatebox[]()()` does not work here as it does in the usual LaTeX math environment.

mantepse:
Thank you!

One question: it might be even better to simplify to single quotes, as one might be used to from working with sage.  I.e.,
```
''
'a'
'ab'
```
(and without the \hspace)

What do you think?

Omegaconstant:
Consider the following:
```
  \begin{array}{rcl}
      & {^{`}} {^{'}} \\
      \swarrow & \downarrow & \searrow \\[5pt]
      \text{'a'} & \text{`b'} & \text{``c''} \\
      \begin{array}{rcl}
          \swarrow & \downarrow & \searrow \\[5pt]
          ^{``}aa^{"} & ^{`}ab^{'} & `ac' \\
      \end{array} & 
      \begin{array}{rcl}
          \swarrow & \downarrow & \searrow \\[5pt]
          {\scriptstyle `}ba' & ^{\scriptstyle `}bb^{'} & ^{`}bc^{'} \\
      \end{array} & 
      \begin{array}{rcl}
          \swarrow & \downarrow & \searrow \\[5pt]
          ^{`}ca^{'} & ^{`}cb^{'} & ^{`}cc^{'} \\
      \end{array}
  \end{array}
```

The output is as follows:
modification

Some observations:
* Without `\hspace` it is kind of difficult to customize spacing. I need to check with the different possibilities with `\quad`,  `\:`, `\;` and `\,`.
* For single quotes with `` `' ``, the left one tends to get a bit larger than the right one (see `` ^{`}ab^{'} ``) and for backtick, it is required to use `` ^{`} `` instead of `` ` `` (see `` `ac'  ``). Also the use of `\scriptstyle` does not work in `` ^{`} `` (see `` ^{\scriptstyle `}bb^{'} ``).
* With double quotes ``` {^{``}} {^{"}} ```, the sizes of the left and right quotes seem similar (see ``` ^{``}aa^{"} ```), but the same does not seem true for single quotes (see `` ^{`}ab^{'} ``).
* Use of `\text` generates desirable results neither for single quotes nor for double quotes (see `` \text{'a'} ``, `` \text{`b'} `` and ``` \text{``c''} ``` ).

mantepse:
I'd go with
```
\begin{array}{ccc}
      & ``\," \\
      \hfil\swarrow & \downarrow & \searrow\hfil\\
      ``a" & ``b" & ``c" \\
      \begin{array}{ccc}
          \swarrow & \downarrow & \searrow \\
          ``aa" & ``ab" & ``ac" \\
      \end{array} & 
      \begin{array}{ccc}
          \swarrow & \downarrow & \searrow \\
          ``ba" & ``bb" & ``bc" \\
      \end{array} & 
      \begin{array}{ccc}
          \swarrow & \downarrow & \searrow \\
          ``ca" & ``cb" & ``cc" \\
      \end{array}
  \end{array}
```
It is not all that nice, but it's readable and very simple.

Omegaconstant:
Cool.
I have modified accordingly.

mantepse (approved):
looks great!  I would like to wait for the linter, to make sure that there is no trailing whitespace.  Unfortunately, I do not have the rights to approve the workflow, it seems!

mantepse:
there is indeed a whitespace problem. i'm sorry.

Omegaconstant:
Oh,
May I ask what does it mean and how it can be resolved?

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -9,25 +9,23 @@ sets that are described by some ``seeds`` and a successor function
...
-See documentation of :func:`RecursivelyEnumeratedSet` below for the
+See the documentation of :func:`RecursivelyEnumeratedSet` below for the
 description of the inputs.
 
 AUTHORS:
 
 - Sébastien Labbé, April 2014, at Sage Days 57, Cernay-la-ville
 
-EXAMPLES:
-
```
tscrim: This should not be removed. What follows are examples.

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -406,7 +386,7 @@ def RecursivelyEnumeratedSet(seeds, successors, structure=None,
 
     .. WARNING::
 
-        If you do not set the good structure, you might obtain bad results,
+        If you do not set a valid structure, you might obtain bad results,
```
tscrim:
```suggestion
        If you do not set a good structure, you might obtain bad results,
```
Valid is not quite strong enough here I think. In either case, it is better to be a bit more vague.

src/sage/sets/recursively_enumerated_set.pyx
```diff
@@ -1574,9 +1554,9 @@ def search_forest_iterator(roots, children, algorithm='depth'):
          [0, 1, 2], [0, 2, 1], [1, 0, 2], [1, 2, 0], [2, 0, 1], [2, 1, 0]]
     """
     # Little trick: the same implementation handles both depth and
-    # breadth first search. Setting position to -1 makes a depth search
+    # breadth first search. Setting position to -1 initiates a depth search
     # (you ask the children for the last node you met). Setting
-    # position on 0 makes a breadth search (enumerate all the
+    # position on 0 initiates a breadth search (enumerate all the
```
tscrim: This change is not quite correct as it doesn't actually start the computation (that is done after this is set). "makes" is correct here.

tscrim (changes requested):

src/sage/sets/recursively_enumerated_set.pyx
mantepse: If you look at the html, it looks terrible - and quite confusing.
image

src/sage/sets/recursively_enumerated_set.pyx
mantepse: OK, I am surprised to learn that "good" is *not* frenglish!

More importantly, I don't understand the sentence, it seems contradictory to me:
* valid is *not* strong enough
* it is *better* to be more vague

src/sage/sets/recursively_enumerated_set.pyx
mantepse: OK, I thought it was obvious that the computation is started later, and "makes" simply sounds wrong to my (non-native speaker) ears.

Would "initiates the ... search" be more correct?

src/sage/sets/recursively_enumerated_set.pyx
tscrim: Honestly I don't think it is all that strange, but you could considering it as wrong type of heading is used for the sub portions.

src/sage/sets/recursively_enumerated_set.pyx
tscrim: The term "good" is not well-defined (more vague), but it implies that there could be something more needed than what is specified as valid (stronger conditions are needed).

src/sage/sets/recursively_enumerated_set.pyx
tscrim: It is not inconceivable that setting a variable initializes a computation. It is an unlikely situation that someone misinterprets, but it is technically wrong.

"make" is being used somewhere in between "creating" and "forcing" here. Perhaps "initializes"? Although I wouldn't change it.

src/sage/sets/recursively_enumerated_set.pyx
mantepse: Well, when I saw it I thought that for some reason (eg., bad markup) the examples were not typeset.

Would it be OK for you if we add instead `EXAMPLES::` right before the examples in each subsection?

I thought that the `EXAMPLES::` markup is only used in docstrings of classes, methods and functions, not at module level, and I could not find file where we have examples at module level interoduced with `EXAMPLES::`.

But in any case, I agree that the sizing of the headings is not good.  Do you have a best-practice example with a similar structure (i.e., with a guided tour at module level)?  I only know of `bijectionist.py`, `findstat.py`, `growth.py`, but they have all been written by myself.

src/sage/sets/recursively_enumerated_set.pyx
mantepse: OK, then "initializes the subsequent ... search"?

mantepse: OK, then "setting position to -1 results in a depth search"?

tscrim: That's good with me.

src/sage/sets/recursively_enumerated_set.pyx
tscrim: The whole list are the examples broken down into different sections. We do have other examples scattered around, but they are less frequent as the class-level doc is usually a better place for `?` access (e.g., `libgap_group.py` has some module-level doc examples, but likely not the best place). So I would prefer to keep it unchanged.

We don't really have a best practice, but perhaps `.. RUBRIC::`? H4 level might be fine too.

src/sage/sets/recursively_enumerated_set.pyx
janmenjayap: Done.

src/sage/sets/recursively_enumerated_set.pyx
janmenjayap: Does the following sound good?
Instead of the statement:
`If you do not set a valid/ good structure, you might obtain bad results, like elements generated twice:`
            
we may write:

`A poorly defined recursion might lead to an unexpected result, such as some of the elements being generated multiple times:`

tscrim: No, that isn't quite right and becomes too vague as it doesn't reference the structure of the poset/recursion. It would be best just to leave it as "good" with changing "the" to "a".

src/sage/sets/recursively_enumerated_set.pyx
janmenjayap: Okay,
I have added `EXAMPLES::` just before each of the examples. Please kindly have a look at it and suggest any changes if required.

tscrim: @Omegaconstant Please revert this. As I said, each block is part of the examples, grouped accordingly.

mantepse: I disagree here. We have several sections, each with its own examples. Why are you against this?

Please take note that the original had me thinking that there would be a mistake.

tscrim: Because it is not describing the class/structure in each section. It is dividing the examples by section/structure. By putting `EXAMPLES::` for each section, it no longer makes sense because there is nothing within each section but examples. In fact, it even has three `Forest Structure` examples (with the latter two having a redundant "Example:" included). The original version is just displayed larger than it should, which might not have been as such on previous versions too.

janmenjayap: Oops,
I just reverted. :sweat:
Apologies for the confusion. Please let me know what is to be done here.

Actually, in the original doc, the `EXAMPLES::` tag exists just before each of the examples in almost all of the places, except for the initial few paragraphs/ sections.

mantepse: OK, I don't like it at all, but I don't care enough.

For me, "EXAMPLES:" looks like a heading.  I would say that, ordinarily, a heading can only be followed by a sub-heading - however, "EXAMPLES:" is always just in capital letters, which would be the smallest possible heading.

janmenjayap: In my opinion,
The initial sections including `No hypothesis on the structure`, `Symmetric structure`, `Graded structure` and `Forest structure` -- each of these has a brief description regarding the structure followed by examples (just before which I added the `EXAMPLES::` tag). In the Forest structure section, there are three consecutive examples, I removed the large section heading of `Example: Forest structure` and `Example: Forest structure 2` and added a single tag `EXAMPLES::` before the start of the examples just after the forest structure.

The inclusion of the `EXAMPLES::` tag just before the example makes more sense than adding a single tag at the beginning of the entire document just after the name of the author and just before the `No hypothesis on the structure`.

I hope this is fine.

tscrim: > In my opinion, The initial sections including `No hypothesis on the structure`, `Symmetric structure`, `Graded structure` and `Forest structure` -- each of these has a brief description regarding the structure followed by examples (just before which I added the `EXAMPLES::` tag).

That's not generally true. I only see this for the "no hypothesis" section.

> In the Forest structure section, there are three consecutive examples, I removed the large section heading of `Example: Forest structure` and `Example: Forest structure 2` and added a single tag `EXAMPLES::` before the start of the examples just after the forest structure.
> 
> The inclusion of the `EXAMPLES::` tag just before the example makes more sense than adding a single tag at the beginning of the entire document just after the name of the author and just before the `No hypothesis on the structure`.

No, it doesn't. All of those that follow in the original `EXAMPLES:` are examples of the module that are explained in detail. They don't talk about the assumptions/structure needed to be satisfies (e.g., a formal definition) or have any generality.

As further evidence, your added `EXAMPLES::` is following after each block already described the example. This is clearly out of place (plus the English sentence syntax is wrong with the previous lines ending with the colon).

Either this needs to be reverted or the doc needs to be completely rewritten in a different format. We can address @mantepse's issue by replacing the header part with a smaller block; e.g,:
```diff
-Example: Forest structure
--------------------------
+.. RUBRIC:: Forest structure [Ex num]
```
That should make it look much better in the compiled doc.

janmenjayap: Cool.
I have understood your point and I have made the changes accordingly (as of now).
Actually, I did not look at the initial sections of the doc from this point of view (that -- the initial sections are the examples for some instances of the recursively enumerated set under some assumptions) and yes, they lack the proper formal definition. We should discuss about it.

Also, I am thankful to everyone for this discussion :-)

Omegaconstant:
It seems that there have been too many commits. 
Shall I squash them to a single commit (just for cleanup)?

mantepse:
No, I don't think so.

tscrim (approved):
No, I don't think so either. It has useful information of things that have been tried.

mantepse:
@tscrim, did you look at the generated html?  I think it looks terrible :-(

I am unhappy about the `EXAMPLES:` being followed by a heading, and I am unhappy about the example LaTeX - for some reason, the LaTeX - html converter completely messes up the quotes.  Maybe we should simply remove them, and use `\emptyset` for the empty word?  (I must say, the example is not all that brilliant, but never mind.)

Omegaconstant:
Hello,

I want to make two points.

1. Regarding the examples, as far as the initial section is concerned, the current doc looks like:
```
Recursively enumerated set

## Definitions and properties of RES

AUTHORS:

Sébastien Labbé, April 2014, at Sage Days 57, Cernay-la-ville

EXAMPLES:
No hypothesis on the structure

## Definition of No hypothesis on the structure
## Examples of No hypothesis on the structure

Symmetric structure

## Examples of Symmetric structure

Graded structure
## Examples of Graded structure

Forest structure
Forest structure [Ex 1]
Forest structure [Ex 2]
```

This is a temporary solution. I believe, as suggested by @tscrim and as mentioned by @mantepse, we should include the definitions of `Symmetric structure`, `Graded structure` and `Forest structure` and finally it may look like the following:
```
Recursively enumerated set

## Definitions and properties of RES

AUTHORS:

Sébastien Labbé, April 2014, at Sage Days 57, Cernay-la-ville

No hypothesis on the structure

## Definition of No hypothesis on the structure
EXAMPLES:
## Examples of No hypothesis on the structure

Symmetric structure

## Definition of Symmetric structure
EXAMPLES:
## Examples of Symmetric structure

Graded structure

## Definition of Graded structure
EXAMPLES:
## Examples of Graded structure

Forest structure

## Definition of Forest structure
EXAMPLES:
## Examples of Forest structure
Forest structure [Ex 1]
Forest structure [Ex 2]
```

This seems like a more robust and better solution concerning the examples than the existing one. This will address both the issues raised by @mantepse and @tscrim and give a better overview of the doc.

2. Regarding the LaTeX, actually that quotation thing I observed earlier while I generated HTML in my local system and I suggested `` ^{`}ca^{'} `` instead of ``` ``ca'' ``. Yep, either we can try out with the former one (aka `` ^{`}ca^{'} `` ) or we may remove the quotations. But I think the first one is better, as it represents the string structure.

Omegaconstant:
Also, the Lint test is failing. I suppose, once again there is some trailing whitespace. 
Will fix it (if any).

tscrim:
@mantepse I don’t think it looks that bad, but okay. Let me again suggest keeping how the original version was but changing all of the headings to be `.. RUBRIC::`. Looking a little more carefully, I don’t think it makes much sense to subhead the examples for the forest structure. Just start and label them one-by-one as `.. RUBRIC:: Forest Structure (Example 1)` as they are examples.

I don’t think it is useful to duplicate documentation with the (precise) definitions. They are examples. If someone wants the definitions, they can look later in the doc.

I agree with @mantepse that the quotes in the latex looks pretty bad in the html. I would remove them all and replace the empty with $\emptyset$ (`\emptyset`).

The lint test is very reliable. If it is failing, then it needs to be fixed.

tscrim:
@mantepse What do you think about the doc now?

mantepse:
Looks OK!  Possibly the module heading should be `Recursively Enumerated Sets`, to make it more similar to the rest?  (https://deploy-preview-37331--sagemath.netlify.app/html/en/reference/sets/)

mantepse:
> Is it required to change the file name from `recursively_enumerated_set.pyx` to `recursively_enumerated_sets.pyx` and any other such changes that are analogous and possible, concerning recursively enumerated sets?

NO!

tscrim:
@mantepse Any additional comments? I am ready to set this to a positive review.

mantepse:
Perfect!

PR #37336, opened 2024-02-14
fixed bug in `.map_coefficients`, `.map_support`, `.map_item` for matrix operation

### LINKED ISSUE #37320: `map_*` fails for `Matrix` object
### Steps To Reproduce

To reproduce, run:

```sage
sage: M = Matrix(RR, 3)
sage: M.map_item(lambda x: x + 1)
```

It should apply the function to every value in the Matrix. In other words, it should return (a Matrix with the same values as) `Matrix.ones(RR, 3)`. But it doesn't.

### Additional Information

The method isn't a "real method" of Matrix, but rather inherited (incorrectly) from `ModulesWithBasis` I believe. I think we can keep the same name and just implement the method correctly.

### Checklist

amanmoon:
Fixed bug in `.map_coefficients`, `.map_support`, `.map_item` methods for Marrix.

`.map_coefficients`: This maps the given function to all the coefficients of equations in the matrix.
`.map_support`: This maps the given function to all the variables of equations in matrix.
`.map_item`: This maps the given function to all the elements in the matrix.

Fixes: #37320

### :memo: Checklist

### DIFF
--- a/src/sage/categories/modules_with_basis.py
+++ b/src/sage/categories/modules_with_basis.py
@@ -2065,6 +2065,16 @@ class ModulesWithBasis(CategoryWithAxiom_over_base_ring):
                 sage: a = s([2,1]) + 2*s([3,2])                                         # needs sage.combinat sage.modules
                 sage: a.map_coefficients(lambda x: x * 2)                               # needs sage.combinat sage.modules
                 2*s[2, 1] + 4*s[3, 2]
+
+            TESTS:
+
+            map_coefficients works on Matrix::
+
+                sage: M = Matrix([[1, 0], [0, 1]])
+                sage: f = lambda x: x + 1
+                sage: M.map_coefficients(f)             # This actually failed at some point!!! See #37320
+                [2 0]
+                [0 2]
             """
             return self.parent().sum_of_terms( (m, f(c)) for m,c in self.monomial_coefficients(copy=False).items() )
 
@@ -2175,6 +2185,14 @@ class ModulesWithBasis(CategoryWithAxiom_over_base_ring):
                 sage: a = s([2,1]) + s([1,1,1])                                         # needs sage.combinat sage.modules
                 sage: a.map_item(f)                                                     # needs sage.combinat sage.modules
                 2*s[2, 1] + 2*s[3]
+
+            To apply the function on all entries including 0 coefficient::
+
+                sage: M = Matrix([[1, 0], [0, 1]])
+                sage: f = lambda x: x + 1
+                sage: M.parent()(map(f, M.list()))
+                [2 1]
+                [1 2]
             """
             return self.parent().sum_of_terms( f(m,c) for m,c in self.monomial_coefficients(copy=False).items() )
 
--- a/src/sage/categories/modules_with_basis.py
+++ b/src/sage/categories/modules_with_basis.py
@@ -2066,7 +2066,7 @@ class ModulesWithBasis(CategoryWithAxiom_over_base_ring):
                 sage: a.map_coefficients(lambda x: x * 2)                               # needs sage.combinat sage.modules
                 2*s[2, 1] + 4*s[3, 2]
             """
-            return self.parent().sum_of_terms( (m, f(c)) for m,c in self )
+            return self.parent().sum_of_terms( (m, f(c)) for m,c in self.monomial_coefficients(copy=False).items() )
 
         def map_support(self, f):
             """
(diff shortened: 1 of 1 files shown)


grhkm21:
Should this also be implemented for other `.map_*` methods? `map_support`, `map_support_skip_none` and `map_coefficients`.

amanmoon:
@grhkm21

tscrim:
Note that matrices already have a `map_coefficients` function from the category. It doesn't work because the generic case (in `sage/src/sage/categories/modules_with_basis.py`) should be
```python
return self.parent().sum_of_terms( (m, f(c)) for m,c in self.monomial_coefficients(copy=False).items() )
```
(it was moved from a specific implementation). Similarly for `map_support`.

I don't understand why you need, much less should, special case `Expression`. This will lead to subtle (and likely very annoying) differences between working with `SR` and polynomial rings. IMO, you shouldn't differentiate and just fix the generic implementations.

amanmoon:
@tscrim, in the `map_item` method, if any element in the matrix has a value of `0`, map_item will `not` apply the function to that element. Should I mention this with an example, or should I modify the map_item method to handle elements with the value 0?
writing an example seems good to me

tscrim:
`map_item` is for the pairs `(supp, coeff)`, where `supp` is a basis element that is supported by the element (i.e., the coefficient is non-zero). It definitely should **not** act on zero coefficients. This includes matrices. If a user wants to act on all entries, they can easily write such a method. Special casing one specific class (even more so if it is a specific implementation of a concept) is bad practice.

You should also add some example showing these now work.

grhkm21:
I disagree that changing the method to have different behaviour here is bad practice, but I don't really mind. In the docstring, please add an example showing how to achieve "applying the function on all entries" by something like

```sage
sage: M = Matrix([[1, 0], [0, 1]])
sage: f = lambda x: x + 1
sage: M.parent()(map(f, M.list()))
[2 1]
[1 2]
```

tscrim:
@grhkm21 No, it is bad practice. Matrices are modules with a fixed basis. As such, all methods specified by the category should behave the same way for all subclasses. That is a fundamental OOP principle (inheriting from the category is effectively subclassing).

src/sage/categories/modules_with_basis.py
```diff
@@ -2065,8 +2065,18 @@ def map_coefficients(self, f):
                 sage: a = s([2,1]) + 2*s([3,2])                                         # needs sage.combinat sage.modules
                 sage: a.map_coefficients(lambda x: x * 2)                               # needs sage.combinat sage.modules
                 2*s[2, 1] + 4*s[3, 2]
+
+            TESTS:
+
+            map_coefficients works on Matrix::
```
tscrim, line 2071:
```suggestion
            This works for matrices (:issue:`37320`)::
```

src/sage/categories/modules_with_basis.py
```diff
@@ -2065,8 +2065,18 @@ def map_coefficients(self, f):
...
+
+                sage: M = Matrix([[1, 0], [0, 1]])
+                sage: f = lambda x: x + 1
+                sage: M.map_coefficients(f)             # This actually failed at some point!!! See #37320
```
tscrim, line 2075:
```suggestion
                sage: M.map_coefficients(f)
```
You already said this basically. No need to sensationalize either.

src/sage/categories/modules_with_basis.py
```diff
@@ -2142,7 +2152,7 @@ def map_support_skip_none(self, f):
                 sage: y.parent() is B                                                   # needs sage.modules
                 True
             """
-            return self.parent().sum_of_terms( (fm,c) for (fm,c) in ((f(m), c) for m,c in self) if fm is not None)
+            return self.parent().sum_of_terms( (fm,c) for (fm,c) in ((f(m), c) for m,c in self.monomial_coefficients(copy=False).items()) if fm is not None)
```
tscrim, line 2155:
```suggestion
            mc = self.monomial_coefficients(copy=False)
            return self.parent().sum_of_terms((fm, c) for m, c in mc.items() if (fm := f(m)) is not None)
```
This can be simplified a bit.

tscrim (changes requested):
You should add doctests showing each method works for matrices, not just `map_coefficients()`.

I disagree with adding a standalone test applying something to every coefficient. I actually think we should not add any tests as it doesn't relate at all to the method. Yet, if you absolutely insist, then it should be in the context of explaining the difference with using the method.

Note that the test also is in `map_item`, but you are simply doing something coefficient by coefficient. (Note that this doesn't even make sense for a general free module (with finite support), even with a distinguished basis.) If you are going to do this, it should be using general functionality, not using a specialized `list()` method (e.g., iterating over all of the basis keys).

PR #37338, opened 2024-02-14
Speedup database_matroids.py tests

gmou3:
I strategically recast a few matroids in `database_matroids.py` to `CircuitsMatroid`s (given #37148).
On my old laptop, this results in a `x5` speedup in the testing of the file, and in a `x2` speedup in the testing of the whole of the `matroids` module (saving `~1m`).


tscrim (approved):
LGTM.

PR #37339, opened 2024-02-14
Removed whitespaces in `ncsym.py`

S17A05:
Removed the whitespaces in `ncsym.py`, which currently cause all PR tests to fail at Lint.


fchapoton (approved):
thanks

PR #37340, opened 2024-02-14
Add class FlatsMatroid

gmou3:
This matroid subclass allows the definition and internal representation of a matroid based on its flats (closed sets). This representation can be advantageous for some algorithms. This completes the list of the most standard matroid definitions (bases, circuits, flats, rank function).

This PR is similar to #37148.


src/sage/matroids/flats_matroid.pyx
```diff
@@ -0,0 +1,534 @@
...
+    cpdef _rank(self, X) noexcept:
+        """
+        Return the rank of a set ``X``.
+
+        This method does no checking on ``X``, and ``X`` may be assumed to have
+        the same interface as ``frozenset``.
+
+        INPUT:
+
+        - ``X`` -- an object with Python's ``frozenset`` interface
```
mkoeppe:
```suggestion
        - ``X`` -- an object with Python's :class:`frozenset` interface
```

src/sage/matroids/flats_matroid.pyx
```diff
@@ -0,0 +1,534 @@
...
+
+    cpdef flats(self, k) noexcept:
+        r"""
+        Return the flats of the matroid of specified rank.
+
+        INPUT:
+
+        - ``k`` -- an integer
+
+        OUTPUT: a SetSystem
```
mkoeppe:
```suggestion
        OUTPUT: a :class:`SetSystem`
```

src/sage/matroids/unpickling.pyx
```diff
@@ -173,6 +175,50 @@ def unpickle_circuit_closures_matroid(version, data):
     return M
 
 
+#############################################################################
+# FlatsMatroid
+#############################################################################
+
+def unpickle_flats_matroid(version, data):
+    """
+    Unpickle a FlatsMatroid.
```
mkoeppe:
```suggestion
    Unpickle a :class:`FlatsMatroid`.
```

src/sage/matroids/constructor.py
```diff
@@ -180,6 +181,7 @@ def Matroid(groundset=None, data=None, **kwds):
     - ``circuits`` -- The list of circuits of the matroid.
     - ``nonspanning_circuits`` -- The list of nonspanning_circuits of the
       matroid.
+    - ``flats`` -- The dictionary of flats indexed by their rank.
```
mkoeppe, line 184: Add an example below?

src/sage/matroids/flats_matroid.pxd
```diff
@@ -0,0 +1,20 @@
+from sage.matroids.matroid cimport Matroid
+
+cdef class FlatsMatroid(Matroid):
+    cdef frozenset _groundset  # _E
+    cdef int _matroid_rank  # _R
+    cdef dict _F  # flats
+    cpdef groundset(self) noexcept
```
mkoeppe: note that over in #37667, we are getting rid of many `noexcept` directives that had been placed incorrectly. Basically anything with `self` or involving any other Python objects shouldn't be `noexcept`

src/sage/matroids/flats_matroid.pyx
```diff
@@ -0,0 +1,534 @@
...
+            return rich_to_bool(op, 1)
+        if lt.full_rank() != rt.full_rank():
+            return rich_to_bool(op, 1)
+        return richcmp(lt._F, rt._F, op)
+
+    # copying, loading, saving
+
+    def __copy__(self):
+        """
+        Create a shallow copy.
```
mkoeppe: I know that you are keeping this parallel to the existing classes, but for immutable objects, both `__copy__` and `__deepcopy__` really should just return `self`. 
See #13811

src/sage/matroids/flats_matroid.pxd
gmou3: Great, happy to be getting rid of noisy stuff.

src/sage/matroids/flats_matroid.pyx
gmou3: Done. I think I will have to open a separate PR to do this for other classes too. In fact, it would suffice to only do it in the main abstract matroid class. This will save many (redundant) lines of code.

mkoeppe: Yes, that's a good idea.

mkoeppe (approved):
LGTM

PR #37341, opened 2024-02-14
Avoid algebra in polynomials

fchapoton:
One avoids using `Algebra` and `CommutativeAlgebra` in the polynomial classes.

And the general method `is_commutative` is delegated to the categories of Rings and CommutativeRings.

### :memo: Checklist

### DIFF
--- a/src/sage/categories/commutative_rings.py
+++ b/src/sage/categories/commutative_rings.py
@@ -47,6 +47,24 @@ class CommutativeRings(CategoryWithAxiom):
 
     """
     class ParentMethods:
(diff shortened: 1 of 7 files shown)


fchapoton:
just because I am not cleaning away all calls to `is_commutative` everywhere right now.. One could aim to remove entirely the method `is_commutative` for rings, but this will take much more work. I take the temporary solution to put that in the category of rings for the moment. My main aim here is to get rid of one `Algebra` and one `CommutativeAlgebra`

mantepse (approved):
♥️

Issue #37308, opened 2024-02-13
Running gap with incorrect GAP_ROOT_PATHS results in awful error messages


tornaria:
- For gap expect interface:
```
$ GAP_ROOT_PATHS=/bad sage -c 'gap(1)'
** Gap crashed or quit executing 'SaveWorkspace("/home/tornaria/.sage/gap/tmp1xkouu8e");' **
Restarting Gap and trying again
** Gap crashed or quit executing '\$sage1:=1;;' **
Restarting Gap and trying again
** Gap crashed or quit executing 'SaveWorkspace("/home/tornaria/.sage/gap/tmpq435w0q8");' **
Restarting Gap and trying again
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/sage/interfaces/expect.py", line 545, in _start
    self._expect.expect(self._prompt)
  File "/usr/lib/python3.12/site-packages/pexpect/spawnbase.py", line 354, in expect
    return self.expect_list(compiled_pattern_list,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pexpect/spawnbase.py", line 383, in expect_list
    return exp.expect_loop(timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pexpect/expect.py", line 179, in expect_loop
    return self.eof(e)
           ^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pexpect/expect.py", line 122, in eof
    raise exc
pexpect.exceptions.EOF: End Of File (EOF). Exception style platform.
Gap with PID 7218 running /usr/bin/gap -A -l /bad -b -p -T -E -m 64m /usr/lib/python3.12/site-packages/sage/ext_data/gap/sage.g
command: /usr/bin/gap
args: ['/usr/bin/gap', '-A', '-l', '/bad', '-b', '-p', '-T', '-E', '-m', '64m', '/usr/lib/python3.12/site-packages/sage/ext_data/gap/sage.g']
buffer (last 100 chars): b''
before (last 100 chars): b"'-l <gaproot>'?@J@n If you ran the GAP binary directly, try running the 'gap.sh' or 'gap.bat' scr\\@J"
after: <class 'pexpect.exceptions.EOF'>
match: None
match_index: None
exitstatus: None
flag_eof: True
pid: 7218
child_fd: 8
closed: False
[...]

Issue #33446, opened 2022-03-02
To determine `GAP_SO` sage.env looks for `libgap.so` but it should look for `libgap.so*`


tornaria:
<div id="comment:0"></div>

My system (void linux) only has

```
$ ls -l /usr/lib/libgap.so*
lrwxrwxrwx 1 root root      15 Jan 14 11:14 /usr/lib/libgap.so.0 -> libgap.so.0.0.0
-rwxr-xr-x 1 root root 2080920 Jan 14 11:14 /usr/lib/libgap.so.0.0.0
```
installed. Indeed, the symlink `/usr/lib/libgap.so -> libgap.so.0.0.0` is shipped in `gap-devel` which is not needed to run sagemath at all.

This is because the implementation of the function `sage.env._get_shared_lib_path()` will look for `libgap.so` instead of `libgap.so*`. The fix is trivial:

```
--- a/src/sage/env.py
+++ b/src/sage/env.py
@@ -293,7 +293,7 @@ def _get_shared_lib_path(*libnames: str) -> Optional[str]:
             if sys.platform == 'darwin':
                 ext = 'dylib'
             else:
-                ext = 'so'
+                ext = 'so*'
 
             if SAGE_LOCAL:
                 search_directories.append(Path(SAGE_LOCAL) / 'lib')
```

Note that this function is only used once in `sage.env` to set `GAP_SO`.

Without the fix above my `GAP_SO` is set to `None` and I get failures like

```
sage: libgap.eval("2+2")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [1], in <module>
----> 1 libgap.eval("2+2")
...
TypeError: expected str, NoneType found
```

Depends on #34391

CC:  @dimpase @orlitzky @williamstein

Component: **packages: standard**

Author: **Gonzalo Tornaría**

Branch/commit - see the corr. PR

Reviewer: **Dima Pasechnik**

_Issue created by migration from https://trac.sagemath.org/ticket/33446_

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
<div id="comment:47"></div>

Branch pushed to git repo; I updated commit sha1. This was a forced push. New commits:
[...]

PR #37344, opened 2024-02-14
GAP: make GAP_ROOT_PATHS configuration unnecessary

tornaria:
As shown in #37308, running with `GAP_ROOT_PATHS` misconfigured results in awful breakage, and there are several situations where it's difficult to get this rigth (#37263, https://github.com/sagemath/sage/issues/37308#issuecomment-1939941434).

The current PR makes minimal changes in `gap` and `libgap` to make this configuration unnecessary. Instead, we obtain the necessary information from the `gap` command.

Note that the user can still change the gap command with `SAGE_GAP_COMMAND`. This includes the possibility to change the paths to a custom with something like `SAGE_GAP_COMMAND='gap -l "path1;path2;path3"'`, etc.

### Details:

- interfaces/gap: remove __make_workspace, not used since 6be205b03c9
- interfaces/gap: cleanup gap_command()
- interfaces/gap: replace global WORKSPACE by Gap() attribute

The first three commits are cleanup to replace global WORKSPACE by an attribute. This necessary to prevent circular references, and it's also good form and simplifies a number of tests which want running a gap instance with a temporary workspace file.

- interfaces/gap: don't use GAP_ROOT_PATHS

This one is trivial since the `gap` command knows the paths.

- gap: fast implementation of KERNEL_INFO()

Implements a fast version of KERNEL_INFO() which obtains the configuration we need by running the `gap` command but avoiding the full initialization of gap which is costly.

See https://github.com/sagemath/sage/issues/33446#comment:48 and https://github.com/sagemath/sage/issues/33446#comment:40 for this idea, which I discarded previously because it was pretty slow. Compare:
```
sage: from sage.interfaces.gap import KERNEL_INFO
sage: KERNEL_INFO._KERNEL_INFO__instance = None # force reload
sage: %time KERNEL_INFO()
CPU times: user 3.53 ms, sys: 1.04 ms, total: 4.57 ms
Wall time: 28.6 ms
{'KERNEL_VERSION': '4.12.2', 'GAP_ARCHITECTURE': 'x86_64-unknown-linux-gnu-default64-kv8', 'GAP_ROOT_PATHS': '/usr/lib64/gap/;/usr/share/gap/'}
```
with
```
sage: import subprocess
sage: %time subprocess.getoutput('gap -q --nointeract --bare -r -c \'Display(JoinStringsWithSeparator(KERNEL_INFO().GAP_ROOT_PATHS,";"));\'')
CPU times: user 0 ns, sys: 1.7 ms, total: 1.7 ms
Wall time: 790 ms
'/usr/lib64/gap/;/usr/share/gap/'
```

The implementation returns only a few selected variables from `KERNEL_INFO()` but it would be easy to add more variables as needed in `src/sage/interfaces/gap_kernel_info.g`.

**TODO: add documentation and tests for `KERNEL_INFO()`.**

- interfaces/gap_workspace: use KERNEL_INFO() to create filename
- libs/gap/saved_workspace: use KERNEL_INFO() to compute gap timestamp
- libgap: get GAP_ROOT_PATHS from KERNEL_INFO()

It is now very easy to rid of all usages of `GAP_ROOT_PATHS`, replacing it by use of `KERNEL_INFO()`.

This even makes some code easier, e.g., `gap_workspace_file()` previously had to find and parse `sysinfo.gap` to read version and architecture, now it can obtain all the required information directly from `KERNEL_INFO()`.

### :memo: Checklist

EDIT: formatting

### DIFF
--- a/src/sage/env.py
+++ b/src/sage/env.py
@@ -236,6 +236,7 @@ NTL_INCDIR = var("NTL_INCDIR")
 NTL_LIBDIR = var("NTL_LIBDIR")
 LIE_INFO_DIR = var("LIE_INFO_DIR", join(SAGE_LOCAL, "lib", "LiE"))
 SINGULAR_BIN = var("SINGULAR_BIN") or "Singular"
+GAP_BIN = var("GAP_BIN", "gap")
 
 # OpenMP
 OPENMP_CFLAGS = var("OPENMP_CFLAGS", "")
--- a/src/sage/interfaces/gap.py
+++ b/src/sage/interfaces/gap.py
@@ -232,17 +232,6 @@ if platform.processor() == 'ia64' and os.path.exists('/usr/bin/prctl'):
     gap_cmd = 'prctl --unaligned=silent ' + gap_cmd
 
 
-def gap_command(use_workspace_cache=True, local=True):
-    if use_workspace_cache:
-        if local:
-            return "%s -L %s" % (gap_cmd, WORKSPACE), False
-        else:
-            # TO DO: Use remote workspace
-            return gap_cmd, False
-    else:
-        return gap_cmd, True
-
-
 # ########### Classes with methods for both the GAP3 and GAP4 interface
 
 class Gap_generic(ExtraTabCompletion, Expect):
@@ -1073,7 +1062,10 @@ class Gap(Gap_generic):
             True
         """
         self.__use_workspace_cache = use_workspace_cache
-        cmd, _ = gap_command(use_workspace_cache, server is None)
+        cmd = gap_cmd
+        # -L: restore a saved workspace (TO DO: Use remote workspace)
+        if use_workspace_cache and server is None:
+            cmd += f" -L {WORKSPACE}"
         # -b: suppress banner
         # -p: enable "package output mode"; this confusingly named option
         #     causes GAP to output special control characters that are normally
@@ -1805,8 +1797,8 @@ def gap_console():
     TESTS::
 
         sage: import subprocess as sp
-        sage: from sage.interfaces.gap import gap_command
-        sage: cmd = 'echo "quit;" | ' + gap_command(use_workspace_cache=False)[0]
+        sage: from sage.interfaces.gap import gap_cmd
+        sage: cmd = 'echo "quit;" | ' + gap_cmd
         sage: gap_startup = sp.check_output(cmd, shell=True,
         ....:                               stderr=sp.STDOUT,
         ....:                               encoding='latin1')
@@ -1820,6 +1812,6 @@ def gap_console():
     from sage.repl.rich_output.display_manager import get_display_manager
     if not get_display_manager().is_in_terminal():
         raise RuntimeError('Can use the console only in the terminal. Try %%gap magics instead.')
-    cmd, _ = gap_command(use_workspace_cache=False)
+    cmd = gap_cmd
     cmd += ' ' + os.path.join(SAGE_EXTCODE, 'gap', 'console.g')
     os.system(cmd)
--- a/src/sage/interfaces/gap.py
+++ b/src/sage/interfaces/gap.py
@@ -1073,7 +1073,7 @@ class Gap(Gap_generic):
             True
         """
         self.__use_workspace_cache = use_workspace_cache
-        cmd, self.__make_workspace = gap_command(use_workspace_cache, server is None)
+        cmd, _ = gap_command(use_workspace_cache, server is None)
         # -b: suppress banner
         # -p: enable "package output mode"; this confusingly named option
         #     causes GAP to output special control characters that are normally
@@ -1197,12 +1197,9 @@ class Gap(Gap_generic):
                 gap_reset_workspace(verbose=False)
                 Expect._start(self, "Failed to start GAP.")
                 self._session_number = n
-                self.__make_workspace = False
             else:
                 raise
 
-        if self.__use_workspace_cache and self.__make_workspace:
-            self.save_workspace()
         # Now, as self._expect exists, we can compile some useful pattern:
         self._compiled_full_pattern = self._expect.compile_pattern_list([
             r'@p\d+\.', '@@', '@[A-Z]', r'@[123456!"#$%&][^+]*\+',
--- a/src/sage/interfaces/gap.py
+++ b/src/sage/interfaces/gap.py
@@ -229,6 +229,45 @@ if platform.processor() == 'ia64' and os.path.exists('/usr/bin/prctl'):
     gap_cmd = 'prctl --unaligned=silent ' + gap_cmd
 
 
+class KERNEL_INFO(dict):
+    """
+    doc
+    """
+
+    # singleton pattern
+    __instance = None
+
+    def __new__(cls):
+        if cls.__instance is None:
+            cls.__instance = super().__new__(cls)
+
+        return cls.__instance
+
+    def __init__(self):
+
+        if self.__dict__ is self:
+            # already initialized
+            return
+
+        from sage.misc.timing import cputime, walltime
+        t = cputime()
+        w = walltime()
+
+        # access keys as object attributes
+        self.__dict__ = self
+
+        from importlib.resources import files
+        import subprocess
+
+        gap_kernel_info = files(__package__).joinpath('gap_kernel_info.g')
+
+        for line in subprocess.getoutput(
+                f"{gap_cmd} --systemfile {gap_kernel_info}"
+                ).split('\n'):
+            var, value = line.split("=")
+            self[var] = value
+
+
 # ########### Classes with methods for both the GAP3 and GAP4 interface
 
 class Gap_generic(ExtraTabCompletion, Expect):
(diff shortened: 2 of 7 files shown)


tornaria:
> I think the assumption that `gap` in PATH at runtime is the correct binary from which to extract this info is quite problematic.

We already assume that `gap` in PATH at runtime is the correct binary and it can be modified at runtime with `SAGE_GAP_COMMAND` as documented. With this PR changing `SAGE_GAP_COMMAND` will result in the correct `GAP_ROOT_PATHS`.

> This should really go through an `Executable` feature.

Orthogonal to this PR. Please open a separate issue to avoid cluttering this one.

This PR is focused on making `GAP_ROOT_PATHS` unnecessary to avoid some types of misconfiguration, with as minimal change in code and behavior as possible.

Fixing other issues, unless caused by this PR, is not a goal.

mkoeppe:
It's not orthogonal because it's the present PR would be a regression for library use of sage.libs.libgap, which currently works with the GAP as configured at compile time

tornaria:
> It's not orthogonal because it's the present PR would be a regression for library use of sage.libs.libgap, which currently works with the GAP as configured at compile time

I'd appreciate a more detailed report.

AFAICT the current version defaults `GAP_ROOT_PATHS` to `$SAGE_LOCAL/lib/gap;$SAGE_LOCAL/share/gap` which is wrong when gap is taken from system prefix (e.g. /usr) which is different from SAGE_LOCAL (e.g. when running sagemath-standard in a venv).

OTOH if using `gap` in PATH is a problem for `sage.libs.libgap`, it's also a problem for `sage.interfaces.gap`. The problem is then that `SAGE_GAP_COMMAND` is not being set correctly (not a regression since I didn't change anything about `SAGE_GAP_COMMAND` here).

Note that unlike `gap` command, the library *does not* know the paths [0]. The paths are taken from the `gap` command in PATH at configure time in the hope that they are the correct ones for the libgap that comes with `-lgap`.  In particular this means that if `SAGE_GAP_COMMAND` is set to the `gap` command in PATH at configure time, this PR would not change the status quo. Does that sound ok?

There's still an issue that `SAGE_GAP_COMMAND` defaults to something non-trivial like `gap -r -A <blah>` maybe we really want to have just `GAP_BIN = var("GAP_BIN", "gap")` and configure `GAP_BIN`? While still giving the user the choice to override the gap invocation with `SAGE_GAP_COMMAND`.


[0] at best we could take the location of the library, say `<PATH>/libgap.so`, then look if `<PATH>/gap/sysinfo.gap` is available which should define a variable `GAP` which should point to the "right" binary, then use my code to get the info we want. Unfortunately, `sysinfo.gap` doesn't contain `GAP_ROOT_PATHS` either.

tornaria:
@mkoeppe I added one more commit for `GAP_BIN`, does it look better this way?

If this is good for you, the missing bits are:
 1. document `KERNEL_INFO()`
 2. remove the `AC_SUBST` for `GAP_ROOT_PATHS` no longer used
 3. add `AC_SUBST` for `GAP_BIN` to fix your issue

Otherwise, a wise man said: "describe the actual system configuration (not an imagined one), describe what happens". Maybe that would make it easier to work out. I already described explicitly two circumstances under which the status quo is broken and needs to be fixed.

orlitzky:
Same comment as on https://github.com/sagemath/sage/issues/37308:

We append the `SAGE_LOCAL` path to the system one to support installing `gap_packages` as an SPKG with the system copy of GAP. Otherwise, you can ignore the `SAGE_LOCAL` path entirely outside of the sage-distribution. The (one) correct system path is already detected at build-time by `./configure`. Tobias's meson branch moves that detection into sagelib itself.

mkoeppe:
What makes this a "blocker"?

tobiasdiez:
The conda workflow is failing here and these failures propagate to every other PR via the blocker label. Hence degrading the priority.

mkoeppe:
@tobiasdiez Good call. I've seen these failures on PRs but didn't have a chance to check where they were coming from.

tornaria:
> The conda workflow is failing here and these failures propagate to every other PR via the blocker label. Hence degrading the priority.

Thanks for noticing it. I don't understand what's going on here:
```
...
  File "/home/runner/work/sage/sage/src/sage/interfaces/gap.py", line 262, in __init__
    gap_kernel_info = files(__package__).joinpath('gap_kernel_info.g')
  File "/usr/share/miniconda3/envs/sage/lib/python3.9/importlib/resources.py", line 147, in files
    return _common.from_package(_get_package(package))
  File "/usr/share/miniconda3/envs/sage/lib/python3.9/importlib/_common.py", line 14, in from_package
    return fallback_resources(package.__spec__)
  File "/usr/share/miniconda3/envs/sage/lib/python3.9/importlib/_common.py", line 18, in fallback_resources
    package_directory = pathlib.Path(spec.origin).parent
  File "/usr/share/miniconda3/envs/sage/lib/python3.9/pathlib.py", line 1082, in __new__
    self = cls._from_parts(args, init=False)
  File "/usr/share/miniconda3/envs/sage/lib/python3.9/pathlib.py", line 707, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/usr/share/miniconda3/envs/sage/lib/python3.9/pathlib.py", line 691, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
Error: Process completed with exit code 1.
```

In effect we are doing something like:
```
from sage.interfaces.gap import __package__
gap_kernel_info = files(__package__).joinpath('gap_kernel_info.g')
```

Is there a bug in that code? What do you get doing that in conda? 

What would be the way to access package data files? I could place the file in `sage/ext_data/gap/` but I thought the goal is to eventually move everything out of there.

I guess an alternative is to do `os.path.join(os.path.dirname(__file__), 'gap_kernel_info.g')`.

tornaria:
I think the issue is that `importlib.resources` does not support namespace packages in python 3.9 :facepalm:

tobiasdiez:
> I think the issue is that `importlib.resources` does not support namespace packages in python 3.9 🤦

I agree. The runs for the other Python versions seem to work. Another reason to opt-out from the namespace packages for the moment https://github.com/sagemath/sage/pull/36753.

mkoeppe:
People, that's why we carry the backport package `importlib_resources`.

tornaria:
> People, that's why we carry the backport package `importlib_resources`.

importlib.resources is supported since python 3.7; importlib_resources is not used at all in sagelib, and can (should) be removed.

mkoeppe:
> > People, that's why we carry the backport package `importlib_resources`.
> 
> importlib.resources is supported since python 3.7; importlib_resources is not used at all in sagelib, and can (should) be removed.

Yes, the module `importlib.resources` exists since python 3.7.

The backport package `importlib_resources` is maintained so that the features of newer Python versions can be used. https://pypi.org/project/importlib-resources/ offers a friendly version table.

More questions?

mkoeppe:
> importlib_resources is not used at all in sagelib

I know that -- I added it in https://github.com/sagemath/sage/pull/35203 because I knew that it will be needed later. And you just found the first use.

tornaria:
> The backport package `importlib_resources` is maintained so that the features of newer Python versions can be used. https://pypi.org/project/importlib-resources/ offers a friendly version table.

Not useful. This is unnecessary dependency; some distros don't ship importlib_resources anymore so we should stick to what's available in python. Besides, `importlib.resources` is in the stdlib which takes precedence over platlib so there's no way `importlib_resources` can override `importlib.resources` afaict.

> More questions?

I did not ask a question, but since you knew python 3.9 did not support namespace packages, a mention _before_ I had to debug this would have been nice.

mkoeppe:
> so there's no way `importlib_resources` can override `importlib.resources`

It doesn't try to. You use `try: from importlib_resources import ....  except: from importlib.reosurces import ....`.

mkoeppe:
Yes, that's how the official backport packages work.

mkoeppe:
> ince you knew python 3.9 did not support namespace packages, a mention _before_ I had to debug this would have been nice

Sorry I didn't notice you are using importlib.resources in this branch! 
But I haven't even really looked at your branch here yet, other than for checking my concern about "gap in PATH at runtime" before posting my comment https://github.com/sagemath/sage/pull/37344#issuecomment-1944588360

mkoeppe:
> some distros don't ship importlib_resources anymore

Could it be that these are distros that only ship Python packages for the latest version of Python?

mkoeppe:
@tornaria Now that you know to use `importlib_resources`, how about correcting your comment https://github.com/sagemath/sage/pull/36753#issuecomment-1952569267?

vbraun:
LGTM

orlitzky:
> LGTM

I don't think this is in a mergeable state, e.g.

```python
class KERNEL_INFO(dict):
    """
    doc
    """
```

And while I would also love it if we didn't need `sage_conf` for `GAP_ROOT_PATHS`, I still hate to see a static piece of configuration become something that sage has to slowly infer whenever it starts.

vbraun:
Its only instantiated on use, and not on Sage startup I presume?

Lets fix then docstring, then.

orlitzky:
> Its only instantiated on use, and not on Sage startup I presume?

I can't say for sure because the sage build fails for me right now and I don't have time to figure out why. Regardless, a huge portion of sage uses GAP, so it's not like you're going to avoid the penalty.

orlitzky:
I'm going to cherry-pick the one remaining cleanup commit; otherwise, this has been addressed.

PR #35075, opened 2023-02-11
Acyclic orientations of graphs

GermainPoullot:
Added a class to enumerate all acyclic orientations of a graph.

Added class AcyclicOrientations and a method .acyclic_orientations in graphs.graph.
The method just call the class.

The algorithm is non trivial: backtracks through the orientations of the graph without go through all of them.

Solve ticket #22468, I think (I'm not use to trac, feel free to explain me how it works)?

### 📝 Checklist

- [x] I have made sure that the title is self-explanatory and the description concisely explains the PR.
- [ ] I have linked an issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

PR #37345, opened 2024-02-14
implemented function for acyclic orientations

### LINKED ISSUE #37253: Implement a function to generate the acyclic orientations of an undirected graph
### Problem Description

In graph theory, an acyclic orientation of an undirected graph is an assignment of a direction to each edge (an orientation) that does not form any directed cycle and therefore makes it into a directed acyclic graph. Every graph has an acyclic orientation. Currently, SageMath lacks a built-in functionality to find acyclic orientations of undirected graphs. The absence of this feature limits users who require such computations for graph-related research and applications.

### Proposed Solution

We will be referring to a research paper by Matthew B. Squire [https://www.sciencedirect.com/science/article/abs/pii/S0196677497908919](url)

The proposed solution involves introducing a new function to address the current lack of a dedicated feature for computing acyclic orientations in SageMath. 
The function would take an undirected graph as input and implement a multi-step process: 
- first, it would reorder the vertices using a greedy approach based on degree centrality, ensuring efficient computation. 
- Then, the edges would be labeled according to the reordering of vertices. 
- Following this, the function would utilize a recursive algorithm to explore subsets of the graph's edges and generate acyclic orientations by considering a partially ordered set. 

The proposed solution encapsulates the essential logic demonstrated in the provided Python code, offering a seamless and integrated way for users to obtain acyclic orientations within the SageMath environment.

### Alternatives Considered

**Manual Implementation:** Users can manually implement the logic provided in the example code, but a built-in function would simplify the process and promote code reusability.

### Additional Information

_No response_

saatvikraoIITGN:
**Description**
This PR introduces an efficient algorithm for generating all acyclic orientations of an undirected graph based on the work by Mathew B. Squire. The algorithm utilizes a recursive approach along with concepts from posets and subsets to improve the efficiency of acyclic orientation generation significantly.

**Changes Made**
1. Implemented the reorder_vertices function to efficiently reorder vertices based on a specific criterion, improving subsequent acyclic orientation generation.
2. Created the order_edges function to assign labels to edges based on the reordered vertices, simplifying the acyclic orientation process.
3. Introduced the generate_orientations function to efficiently generate acyclic orientations by considering subsets of edges and checking for upsets in a corresponding poset.
4. Implemented the recursive helper function to generate acyclic orientations for smaller graphs, building up to larger graphs.
5. Modified the main function to use SageMath for graph manipulation and incorporated the new algorithm.

Fixes #37253 

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -114,7 +114,7 @@ def acyclic_orientations(G):
     Graph with one edge::
 
         sage: list(Graph([(0, 1)]).acyclic_orientations())
-        [Graph on 2 vertices, Graph on 2 vertices]
+        [DiGraph on 2 vertices, DiGraph on 2 vertices]
 
     Graph with two edges::
 
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -86,7 +86,7 @@ def acyclic_orientations(G):
         sage: import timeit
         sage: G_C100, start_time = graphs.CycleGraph(100), timeit.default_timer()
         sage: it_C100, first_10_orientations = G_C100.acyclic_orientations(), list(islice(G_C100.acyclic_orientations(), 10))
-        sage: print(timeit.default_timer() - start_time < 0.05)
+        sage: print(timeit.default_timer() - start_time < 0.5)
         True
 
     TESTS:
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -106,6 +106,8 @@ def acyclic_orientations(G):
         42
     """
     from sage.rings.infinity import Infinity
+    from sage.graphs.graph import Graph
+    from sage.graphs.graph import DiGraph
     from sage.combinat.subset import Subsets
 
     def reorder_vertices(G):
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -257,14 +257,13 @@ def acyclic_orientations(G):
     # Create a mapping between original and new vertex labels
     reverse_vertex_labels = {label: vertex for vertex, label in vertex_labels.items()}
 
-    # Iterate over acyclic orientations and relabel the vertices
+    # Iterate over acyclic orientations and create relabeled graphs
+    from sage.graphs.graph import Graph
     for orientation in orientations:
-        relabeled_orientation = {}
-        for (u, v), label in orientation.items():
-            relabeled_orientation[(reverse_vertex_labels[u], reverse_vertex_labels[v])] = label
+        relabeled_graph = Graph([(reverse_vertex_labels[u], reverse_vertex_labels[v], label) for (u, v), label in orientation.items()])
+        yield relabeled_graph
 
-        yield relabeled_orientation
-
+    # output the orientations
     return orientations
 
 
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -80,15 +80,6 @@ def acyclic_orientations(G):
         sage: len(list(it))
         54
 
-    For C100 graph, we get first 10 orientations pretty quickly::
-
-        sage: from itertools import islice
-        sage: import timeit
-        sage: G_C100, start_time = graphs.CycleGraph(100), timeit.default_timer()
-        sage: it_C100, first_10_orientations = G_C100.acyclic_orientations(), list(islice(G_C100.acyclic_orientations(), 10))
-        sage: print(timeit.default_timer() - start_time < 0.5)
-        True
-
     TESTS:
 
     Acyclic orientations of a complete graph::
@@ -248,11 +239,8 @@ def acyclic_orientations(G):
     # Reorder vertices based on the logic in reorder_vertices function
     vertex_labels = reorder_vertices(G)
 
-    # Create a new graph with updated vertex labels using SageMath
-    from sage.graphs.graph import Graph
-    new_G = Graph()
-    for u, v in G.edges(labels=False):  # Assuming the graph edges are unlabelled
-        new_G.add_edge(vertex_labels[u], vertex_labels[v])
+    # Create a new graph with updated vertex labels using SageMath, Assuming the graph edges are unlabelled
+    new_G = G.relabel(perm=vertex_labels, inplace=False)
 
     G = new_G
 
@@ -265,6 +253,18 @@ def acyclic_orientations(G):
     m = len(edge_labels)
     k = len(vertex_labels)
     orientations = helper(G, globO, m, k)
+
+    # Create a mapping between original and new vertex labels
+    reverse_vertex_labels = {label: vertex for vertex, label in vertex_labels.items()}
+
+    # Iterate over acyclic orientations and relabel the vertices
+    for orientation in orientations:
+        relabeled_orientation = {}
+        for (u, v), label in orientation.items():
+            relabeled_orientation[(reverse_vertex_labels[u], reverse_vertex_labels[v])] = label
+
+        yield relabeled_orientation
+
     return orientations
 
 
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -263,9 +263,6 @@ def acyclic_orientations(G):
         relabeled_graph = Graph([(reverse_vertex_labels[u], reverse_vertex_labels[v], label) for (u, v), label in orientation.items()])
         yield relabeled_graph
 
-    # output the orientations
-    return orientations
-
 
 def strong_orientations_iterator(G):
     r"""
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -92,17 +92,17 @@ def acyclic_orientations(G):
     To count the number of acyclic orientations for a graph with 0 vertices::
 
         sage: list(Graph().acyclic_orientations())
-        [Graph on 0 vertices]
+        []
 
     To count the number of acyclic orientations for a graph with 1 vertex::
 
         sage: list(Graph(1).acyclic_orientations())
-        [Graph on 0 vertices]
+        []
 
     To count the number of acyclic orientations for a graph with 2 vertices::
 
         sage: list(Graph(2).acyclic_orientations())
-        [Graph on 0 vertices]
+        []
 
     Acyclic orientations of a complete graph::
 
@@ -127,6 +127,10 @@ def acyclic_orientations(G):
         6
 
     """
+    if not G.size():
+        # A graph without edge cannot be oriented
+        return
+
     from sage.rings.infinity import Infinity
     from sage.combinat.subset import Subsets
 
--- a/src/sage/graphs/orientations.py
+++ b/src/sage/graphs/orientations.py
@@ -47,10 +47,11 @@ def acyclic_orientations(G):
 
     ALGORITHM:
 
-    The algorithm is based on a paper by Mathew B. Squire, presenting an efficient
-    algorithm for listing the acyclic orientations of a graph. The algorithm is
-    shown to require O(n) time per acyclic orientation generated, making it the
-    most efficient known algorithm for generating acyclic orientations.
+    The algorithm is based on [Sq1998]_.
+    It presents an efficient algorithm for listing the acyclic orientations of a
+    graph. The algorithm is shown to require O(n) time per acyclic orientation
+    generated, making it the most efficient known algorithm for generating acyclic
+    orientations.
 
     The function uses a recursive approach to generate acyclic orientations of the
     graph. It reorders the vertices and edges of the graph, creating a new graph
@@ -246,7 +247,7 @@ def acyclic_orientations(G):
 
     # Reorder vertices based on the logic in reorder_vertices function
     vertex_labels = reorder_vertices(G)
-    
+
     # Create a new graph with updated vertex labels using SageMath
     new_G = Graph()
     for u, v in G.edges(labels=False):  # Assuming the graph edges are unlabelled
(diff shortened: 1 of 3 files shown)


src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
 from sage.graphs.digraph import DiGraph
 
 
+def acyclic_orientations(G):
+    r"""
+    Return an iterator over all acyclic orientations of an undirected graph `G`.
+
+    ALGORITHM:
+
+    The algorithm is based on a paper by Mathew B. Squire, presenting an efficient
```
grhkm21: Can you add a citation to the paper? (See the REFERENCES part of the [developer guide](https://doc.sagemath.org/html/en/developer/coding_basics.html#headings-of-sage-library-code-files)).

grhkm21:
Is this related to #35075? Also, lint CI is failing.

src/sage/graphs/orientations.py
saatvikraoIITGN: Sure!

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+        # Create a new graph to determine reachable vertices
+        new_G = DiGraph()
+        for i in range(0, starting_of_Ek):
+            u, v = keys[i]
+            if globO[(u, v)] == 1:
+                new_G.add_edge(v, u)
+            else:
+                new_G.add_edge(u, v)
+
+        for i in range(starting_of_Ek, m - 1):
```
dcoudert: To avoid this for loop, it suffices to add all vertices of `G` to `new_G`, for instance using `new_G = DiGraph([G, []], format='vertices_and_edges')`.

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+            new_G.add_edge(k, k - 1)
+        else:
+            new_G.add_edge(k-1, k)
+
+        for i in range(starting_of_Ek, m - 1):
+            for j in range(starting_of_Ek, m - 1):
+                u, v = keys[i]
+                w, x = keys[j]
+                # w should be reachable from u and v should be reachable from x
+                if (new_G.shortest_path_length(u, w) < Infinity and new_G.shortest_path_length(x, v) < Infinity):
```
dcoudert: an alternative could be `if w in new_G.depth_first_search(u) and v in new_G.depth_first_search(v)`. I'm not sure which one is the fastest.

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+    G = new_G
+
+    # Order the edges based on the logic in order_edges function
+    edge_labels = order_edges(G, vertex_labels)
+
+    # Create globO array
+    m = len(edge_labels)
+    globO = {}
+    for (u, v) in edge_labels:
+        globO[(u, v)] = 0
```
dcoudert: you can use `globO[u, v]` instead of `globO[(u, v)]`

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+        new_G.add_edge(vertex_labels[u], vertex_labels[v])
+
+    G = new_G
+
+    # Order the edges based on the logic in order_edges function
+    edge_labels = order_edges(G, vertex_labels)
+
+    # Create globO array
+    m = len(edge_labels)
+    globO = {}
```
dcoudert: you can simply do `globO = {uv: 0 for uv in edge_labels}`

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+
+        sage: G = graphs.CompleteGraph(5)
+        sage: it = g.acyclic_orientations(G)
+        sage: len(list(it))
+        120
+    """
+    from sage.rings.infinity import Infinity
+    from sage.combinat.subset import Subsets
+
+    def reorder_vertices(G):
```
dcoudert, line 137: if the vertices of `G` are not labeled in range `0..n-1`, `vertex_labels` should be a mapping vertex to int.

dcoudert (commented):
Thank you for this first draft. I see that there is room for improvements. A first issue is that you implicitly assume that the vertices of the input graph are labeled `0..n-1` which is not always the case. To cope with this issue, you should use method `relabel` to get both a copy of the input graph with suitable vertex labels  and the mapping.

Below are a few comments.

src/sage/graphs/orientations.py
saatvikraoIITGN: We are assuming that the edges are unlabelled 
image

dcoudert: Yes, edges are unlabelled, but the vertex of a graph can be an integer, a string or any hashable object (frozenset, immutable graph, etc.). The current implementation is not working if a vertex is named `'abc'`.

saatvikraoIITGN: Got it! I have made the changes and added test cases for the same.

src/sage/graphs/orientations.py
saatvikraoIITGN: Thanks for the suggestion!

src/sage/graphs/orientations.py
saatvikraoIITGN: Yes, this would be simpler.

dcoudert:
To fix the failing doctests, you must expose the method in `Graph`. See the bottom of file `src/sage/graphs/graph.py` to understand how to do that.

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,232 @@
...
+
+            # Orienting H_k as 0
+            globO[(k-1, k)] = 0
+            yield from generate_orientations(globO, starting_of_Ek, m, k, keys)
+
+    # Reorder vertices based on the logic in reorder_vertices function
+    vertex_labels = reorder_vertices(G)
+
+    # Create a new graph with updated vertex labels using SageMath
+    new_G = Graph()
```
dcoudert: As can be seen in the failing doctests, you need to add `from sage.graphs.graph import Graph` before.

dcoudert (changes requested):

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+    EXAMPLES:
+
+    To count number acyclic orientations for a graph::
+
+        sage: g = Graph([(0, 3), (0, 4), (3, 4), (1, 3), (1, 2), (2, 3), (2, 4)])
+        sage: it = g.acyclic_orientations()
+        sage: len(list(it))
+        54
+
+    For C100 graph, we get first 10 orientations pretty quickly::
```
dcoudert: I don't like this test as it depends on the machine on which it is executed, and so main generate a doctest error. You never know what the system can do...

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+
+            # Orienting H_k as 0
+            globO[(k-1, k)] = 0
+            yield from generate_orientations(globO, starting_of_Ek, m, k, keys)
+
+    # Reorder vertices based on the logic in reorder_vertices function
+    vertex_labels = reorder_vertices(G)
+
+    # Create a new graph with updated vertex labels using SageMath
+    from sage.graphs.graph import Graph
```
dcoudert: The simpler form is
```diff
-    from sage.graphs.graph import Graph
-    new_G = Graph()
-    for u, v in G.edges(labels=False):  # Assuming the graph edges are unlabelled
-        new_G.add_edge(vertex_labels[u], vertex_labels[v])
+   new_G = G.relabel(perm=vertex_labels, inplace=False)
```

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+    new_G = Graph()
+    for u, v in G.edges(labels=False):  # Assuming the graph edges are unlabelled
+        new_G.add_edge(vertex_labels[u], vertex_labels[v])
+
+    G = new_G
```
dcoudert, line 280: The drawback of this instruction is that if the the vertices of the original graphs are `'a', 'b', 'c'`, the graphs returned by the method will have vertex labels `0, 1, 2`.

At least, each graph should be properly relabeled with original vertex labels before it is yielded.

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+
+    # Order the edges based on the logic in order_edges function
+    edge_labels = order_edges(G, vertex_labels)
+
+    # Create globO array
+    globO = {uv: 0 for uv in edge_labels}
```
dcoudert, line 286: Are you sure that `globO` should be a dictionary and not simply a list ?

dcoudert (changes requested):
We still have serious issues. We also have some minor issues regarding the coding style (too many parenthesis in unnecessary places, iteration of `G.vertices()` when over `G` is enough, etc.), but this can be fixed in a separate PR.

src/sage/graphs/orientations.py
saatvikraoIITGN: Oh right! It does fail sometimes. I guess we can remove it.

src/sage/graphs/orientations.py
saatvikraoIITGN: Doesn't a dictionary provide a direct way to associate edge labels with their orientations? I might be wrong.

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,233 @@
...
+
+    m = len(edge_labels)
+    k = len(vertex_labels)
+    orientations = helper(G, globO, m, k)
+
+    # Create a mapping between original and new vertex labels
+    reverse_vertex_labels = {label: vertex for vertex, label in vertex_labels.items()}
+
+    # Iterate over acyclic orientations and relabel the vertices
+    for orientation in orientations:
```
dcoudert, line 296: This loop yields the relabeling, not the relabeled graphs. You should do something like
```py
for orientation in helper(G, globO, m, k):
    orientation.relabel(perm=reverse_vertex_labels, inplace=True)
    yield orientation
```

dcoudert (changes requested):

src/sage/graphs/orientations.py
dcoudert: I don't understand why you don't use the `relabel` method.

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,232 @@
...
+    reverse_vertex_labels = {label: vertex for vertex, label in vertex_labels.items()}
+
+    # Iterate over acyclic orientations and create relabeled graphs
+    from sage.graphs.graph import Graph
+    for orientation in orientations:
+        relabeled_graph = Graph([(reverse_vertex_labels[u], reverse_vertex_labels[v], label) for (u, v), label in orientation.items()])
+        yield relabeled_graph
+
+    # output the orientations
+    return orientations
```
dcoudert: This is no longer necessary

dcoudert (changes requested):

src/sage/graphs/orientations.py
saatvikraoIITGN: Tried but we can't use it for a dictionary

dcoudert: You are right.

dcoudert:
Can you add doctests for small cases and check that the output is consistent. I don't think it is so far. 
```py
sage: list(Graph().acyclic_orientations())
[Graph on 0 vertices]
sage: list(Graph(1).acyclic_orientations())
[Graph on 0 vertices]
sage: list(Graph(2).acyclic_orientations())
[Graph on 0 vertices]
```

dcoudert:
A graph without edge has no orientation, right ? so the method should not return an empty graph is such case.

saatvikraoIITGN:
> A graph without edge has no orientation, right ? so the method should not return an empty graph is such case.

The reason why the code is giving [Graph on 0 vertices] as output when the input graph has no edges is because the function is constructed to always yield a graph, regardless of whether it has edges or not. 
We can modify the function to explicitly handle this case and return an empty list [].

dcoudert:
You could add at the beginning of the method:
```py
if not G.size():
    # A graph without edge cannot be oriented
    return
```

src/sage/graphs/orientations.py
```diff
@@ -41,6 +41,264 @@
...
+    k = len(vertex_labels)
+    orientations = helper(G, globO, m, k)
+
+    # Create a mapping between original and new vertex labels
+    reverse_vertex_labels = {label: vertex for vertex, label in vertex_labels.items()}
+
+    # Iterate over acyclic orientations and create relabeled graphs
+    from sage.graphs.graph import Graph
+    for orientation in orientations:
+        relabeled_graph = Graph([(reverse_vertex_labels[u], reverse_vertex_labels[v], label) for (u, v), label in orientation.items()])
```
dcoudert: oups, this should be a `DiGraph`, not a `Graph`.

We are almost done.

dcoudert (changes requested):

dcoudert:
Don't forget that the output of the doctests are now `DiGraph(...)` and not `Graph(...)`.

src/sage/graphs/orientations.py
```diff
@@ -114,7 +114,7 @@ def acyclic_orientations(G):
     Graph with one edge::
 
         sage: list(Graph([(0, 1)]).acyclic_orientations())
-        [Graph on 2 vertices, Graph on 2 vertices]
+        [DiGraph on 2 vertices, DiGraph on 2 vertices]
```
dcoudert: Actually, the output is `[Digraph on 2 vertices, Digraph on 2 vertices]`. Sorry for the confusion.

dcoudert (changes requested):

dcoudert (approved):
LGTM.

We can certainly improve the code to get a faster method, but in a future PR. I will think about it.

PR #37357, opened 2024-02-15
allow the warning about root isolation failures in CBF[x] to be turned off

mezzarobba:


mezzarobba:
Thanks for the review, I had completely forgotten about this PR!

On Thursday, 26 September 2024 08:20:41 CEST Frédéric Chapoton wrote:
> please add the new keyword to the doc

I'm not sure where to put it. I already added an example with the new 
keyword to the docs of the main roots() method; I feel that adding 
something in the INPUT block there for a keyword that only applies to a 
specific coefficient ring would be more confusing than anything. And the 
docstring of the interval `_roots_univariate_polynomial` method does not 
look like a good place either. What would you suggest?

fchapoton:
ok, then let's see if the tests pass and I will set to positive

fchapoton (approved):
ok

PR #37358, opened 2024-02-15
Implement gcd and lcm on lists of `Factorization`s

grhkm21:
Fixes #33781.

None of the changes is implemented by me. They belong to the authors at #33781.

Depends on #37321.


tscrim (changes requested):
This feels hacky and brittle. It would be better to explicitly check the type of the input rather than assuming that the first element in the list is what would be desired.

grhkm21:
There's some not-so-related commits from #37321 that shows up in the commit log, which is weird since I thought `git merge` should hide them... maybe I misunderstand it. Anyways, this should be ready for review. I rewrote `gcd` so it's cleaner

src/sage/arith/misc.py
```diff
@@ -1790,6 +1790,30 @@ def gcd(a, b=None, **kwargs):
         sage: type(gcd([]))
         <class 'sage.rings.integer.Integer'>
 
+    It is possible to compute gcd of :class:`Factorization`::
```
tscrim:
```suggestion
    It is possible to compute the gcd of a :class:`Factorization`::
```

src/sage/arith/misc.py
```diff
@@ -1842,76 +1865,29 @@ def gcd(a, b=None, **kwargs):
...
-    -  ``v`` - A sequence (possibly empty)
-
-
-    OUTPUT: The gcd of the elements of the sequence as an element of
-    the sequence's universe, or the integer 0 if the sequence is
-    empty.
-
-    EXAMPLES::
+    if U is ZZ:
+        return GCD_list(seq)
```
tscrim, line 1880: This should be tested before `Objects()` as it is more likely to be called (if that test remains).

src/sage/arith/misc.py
```diff
@@ -1842,76 +1865,29 @@ def gcd(a, b=None, **kwargs):
...
-        return __GCD_sequence(seq, **kwargs)
-
+    U = seq.universe()
 
-GCD = gcd
+    if len(seq) == 0:
+        return U(Integer(0))
 
+    if U is Objects():
+        raise ValueError("gcd arguments must share a common parent")
```
tscrim:
```suggestion
```
I don't think this should be tested. Let the function fail in the way it would. The elements don't need to belong to a common parent either; just implement a `gcd` (that works with the given inputs).

src/sage/arith/misc.py
```diff
@@ -1842,76 +1865,29 @@ def gcd(a, b=None, **kwargs):
...
-        sage: X=polygen(QQ)
-        sage: __GCD_sequence(Sequence((2*X+4,2*X^2,2)))
-        1
-        sage: X=polygen(ZZ)
-        sage: __GCD_sequence(Sequence((2*X+4,2*X^2,2)))
-        2
-        sage: __GCD_sequence(Sequence((1/1,1/2)))
-        1/2
+    from functools import reduce
+    return reduce(gcd, map(U, seq))
```
tscrim:
```suggestion
    from functools import reduce
    return reduce(gcd, map(U, seq))
```
This should be more expensive than the previous implementation as it goes through all of the checks with additional redirections. Please check the timings as this is an important function. (Although use `ZZ.zero()` instead of `ZZ(0)` since the former is cached.)

src/sage/arith/misc.py
```diff
@@ -1842,76 +1865,29 @@ def gcd(a, b=None, **kwargs):
...
-    if seq.universe() is ZZ:
-        return GCD_list(seq)
-    else:
-        return __GCD_sequence(seq, **kwargs)
-
+    U = seq.universe()
 
-GCD = gcd
+    if len(seq) == 0:
+        return U(Integer(0))
```
tscrim:
```suggestion
    if not seq:
        return U(ZZ.zero())
```
Both of these are (marginally) faster tests.

src/sage/groups/generic.py
```diff
@@ -1481,12 +1483,18 @@ def has_order(P, n, operation='+'):
...
         if n <= 0:
             return False
         n = n.factor()
+    elif isinstance(n, int):
+        if n <= 0:
+            return False
+        from sage.structure.coerce import py_scalar_to_element
+        n = py_scalar_to_element(n).factor()
 
-    G = P.parent()
```
tscrim, line 1489: This cannot be deleted as it is called in the `return _rec(P, n)`.

src/sage/structure/sequence.py
```diff
@@ -228,8 +233,8 @@ def Sequence(x, universe=None, check=True, immutable=False, cr=False, cr_str=Non
                     x = x.gens()
 
     if universe is None:
-        orig_x = x
-        x = list(x)  # make a copy even if x is a list, we're going to change it
+        orig_x = list(x)
```
tscrim:
```suggestion
        if not isinstance(x, (list, tuple)):
            orig_x = list(x)
        else:
            orig_x = x
```
No need to make a second copy if it already is a list. The `isinstance` check should be faster in nearly all cases than creating the shallow copy.

src/sage/structure/sequence.py
```diff
@@ -212,6 +212,11 @@ def Sequence(x, universe=None, check=True, immutable=False, cr=False, cr_str=Non
         Traceback (most recent call last):
         ...
         TypeError: unable to convert a to an element of Integer Ring
+
+    This used to construct ``[]`` since the iterator is consumed twice::
```
tscrim:
```suggestion
    The iterator used to be consumed twice during the construction when
    there is no coercion (see :issue:`37358`)::
```
I think this helps clarify what the doctest is protecting against, but there likely is a better way to explain...

src/sage/structure/factorization.py
```diff
@@ -755,6 +755,51 @@ def unit(self):
         """
         return self.__unit
 
+    def factor(self, simplify=False):
+        r"""
+        Return the :class:`Factorization` of this object, i.e. itself.
+
+        If ``simplify`` is set, this factorization is fully factorized.
```
tscrim, line 762: I don't understand the point of the `simplify` here. It also seems to be describing something different than what is in the `Factorization` object and nothing else other than `Factorization` seems to support it. I this it should just be removed. If you have some need for it, then it should instead simply call `self.simplify()` before returning `self`.

src/sage/groups/generic.py
```diff
@@ -1481,12 +1483,18 @@ def has_order(P, n, operation='+'):
         In some cases, order *testing* can be much faster than
         *computing* the order using :func:`order_from_multiple`.
     """
+    from sage.arith.misc import factor
+
```
tscrim:
```suggestion
```
This doesn't seem to be used here.

tscrim (changes requested):

src/sage/structure/factorization.py
grhkm21: #37321

src/sage/groups/generic.py
grhkm21: What do you mean?
```python
def _rec(Q, fn):
    if not fn:
        return isid(Q)
                                       
    if len(fn) == 1:
        p, k = fn[0]
        for _ in range(k):
            if isid(Q):
                return False
            Q = mult(Q, p)
        return isid(Q)
                                       
    fl = fn[::2]
    fr = fn[1::2]
    l = prod(p**k for p,k in fl)
    r = prod(p**k for p,k in fr)
    L, R = mult(Q, r), mult(Q, l)
    return _rec(L, fl) and _rec(R, fr)
```
doesn't call `G` anywhere, and my linter says it's unused as well.

src/sage/arith/misc.py
grhkm21: Oh wow, I didn't realise how much slower this is. Comparing `reduce(gcd, arr)` with the explicit for loop on `x = randrange(2**32); arr = [ZZ(x * randrange(2**32)) for _ in range(10**6)]` shows reduce is 40% to 50% slower. Sorry for that

grhkm21:
`rst` lint fails due to #37470

src/sage/groups/generic.py
tscrim: Sorry, I misread the `P` as `G` down below (where `_rec` is called).

maxale:
What's going on with this PR?

PR #37360, opened 2024-02-15
Fix doctest for `multiplication_by_m`

S17A05:
The current doctest can create a singular curve, for example with `random-seed=129979402993014347526756638208394445987`.


grhkm21:
Oh, this is introduced by me isn't it. Thanks. LGTM.

grhkm21 (approved):

PR #37361, opened 2024-02-15
Add iterators, Whitney numbers, characteristic polynomial, etc

gmou3:
Changes (mostly in `matroid.pyx`):
- Add `_iterator` functions for `bases`, `circuits`, etc, and use them internally (this avoids memory consumption/blowups)
- Add Whitney numbers functions (for the first and second kind)
- Correct `f_vector` definition (the definition up till now referred to the Whitney numbers of second kind)
- Add characteristic polynomial
- Correct `is_valid` function to be `False` on a negative rank value
- Optimize `circuits` function (e.g. `sage: %time C = matroids.CompleteGraphic(8).circuits()` drops from `4m` to `1m`)
- Optimize `broken_circuit_complex` function
- Improve docstrings (1-line outputs and other changes)

### ⌛ Dependencies
* Depends on #37667


src/sage/matroids/basis_exchange_matroid.pyx
```diff
@@ -1957,8 +1956,8 @@ cdef class BasisExchangeMatroid(Matroid):
             else:
                 k = min(self.full_rank() - 1, 2)
                 fie, f_vec = self._flat_element_inv(k)
-                self._weak_invariant_var = hash(tuple([tuple([(f, len(fie[f])) for f in sorted(fie)]), f_vec]))
-                self._weak_partition_var = SetSystem(self._E, [fie[f] for f in sorted(fie)])
+                self._weak_invariant_var = hash(tuple([tuple([(f, len(fie[f])) for f in sorted(fie, key=str)]), f_vec]))
+                self._weak_partition_var = SetSystem(self._E, [fie[f] for f in sorted(fie, key=str)])
```
tscrim: Why not let the sorting use the natural way? This makes `10` appear before `2`. If you can a case where this fails, a doctest should be added. In that case, I think a `try`-`except` block falling back to `str` sorting is better (assuming that uniquely determines things).

src/sage/matroids/basis_exchange_matroid.pyx
```diff
@@ -1232,27 +1232,26 @@ cdef class BasisExchangeMatroid(Matroid):
...
-    cpdef f_vector(self) noexcept:
+    cpdef whitney_numbers2(self) noexcept:
         r"""
-        Return the `f`-vector of the matroid.
+        Return the Whitney numbers of the second kind of the matroid.
 
-        The `f`-*vector* is a vector `(f_0, ..., f_r)`, where `f_i` is the
-        number of flats of rank `i`, and `r` is the rank of the matroid.
+        The Whitney numbers of the second kind are here encoded as a vector
+        `(W_0, ..., W_r)`, where `W_i` is the number of flats of rank `i`, and
```
tscrim:
```suggestion
        `(W_0, \ldots, W_r)`, where `W_i` is the number of flats of rank `i`, and
```
It makes the compiled doc slightly better.

src/sage/matroids/matroid.pyx
```diff
@@ -174,9 +178,9 @@ A subclass should always override the underscored method, if available, and as
 a rule leave the regular method alone.
 
 These underscored methods are not documented in the reference manual. To see
-them, within Sage you can create a matroid ``M`` and type ``M._<tab>``. Then
-``M._rank?`` followed by ``<tab>`` will bring up the documentation string of
-the ``_rank()`` method.
+them, within Sage you can create a matroid ``M`` and type ``M._`` followed by
+:kbd:`Tab`. Then ``M._rank?`` followed by :kbd:`Tab` will bring up the
```
tscrim:
```suggestion
:kbd:`Tab`. Then ``M._rank?`` followed by :kbd:`Enter` will bring up the
```
Only noticed because the line changed.

src/sage/matroids/matroid.pyx
```diff
@@ -1107,19 +1073,18 @@ cdef class Matroid(SageObject):
...
+        frozensets ``X`` and ``Y`` so that ``N`` is isomorphic to
+        ``self.minor(X, Y)``.
 
         INPUT:
 
         - ``N`` -- An instance of a ``Matroid`` object,
-        - ``certificate`` -- Boolean (Default: ``False``) If ``True``, returns
+        - ``certificate`` -- boolean (Default: ``False``) If ``True``, returns
           ``True, (X, Y, dic) where ``N`` is isomorphic to ``self.minor(X, Y)``,
           and ``dic`` is an isomorphism between ``N`` and ``self.minor(X, Y)``.
```
tscrim:
```suggestion
        - ``certificate`` -- boolean (default: ``False``); if ``True``, returns
          ``True, (X, Y, dic)`` where ``N`` is isomorphic to ``self.minor(X, Y)``,
          and ``dic`` is an isomorphism between ``N`` and ``self.minor(X, Y)``
```
Fixing some things here.

src/sage/matroids/matroid.pyx
```diff
@@ -1107,19 +1073,18 @@ cdef class Matroid(SageObject):
...
         """
-        Test if matroid has the specified minor,
-        and optionally return frozensets ``X`` and ``Y`` so that ``N`` is isomorphic to ``self.minor(X, Y)``.
+        Test if matroid has the specified minor, and optionally return
+        frozensets ``X`` and ``Y`` so that ``N`` is isomorphic to
+        ``self.minor(X, Y)``.
 
         INPUT:
 
         - ``N`` -- An instance of a ``Matroid`` object,
```
tscrim:
```suggestion
        - ``N`` -- an instance of a :class:`Matroid` object
```
A while-we-are-at-it thing.

src/sage/matroids/matroid.pyx
```diff
@@ -2383,21 +2290,52 @@ cdef class Matroid(SageObject):
...
+            C.add(self._circuit(S))
+        return SetSystem(list(self.groundset()), C)
+
+    def circuits_iterator(self, k=None):
+        """
+        Return an iterator over the circuits of the matroid.
+
+        .. SEEALSO::
+
+            :meth:`M.circuit() <sage.matroids.matroid.Matroid.circuit>`
```
tscrim:
```suggestion
            :meth:`~sage.matroids.matroid.Matroid.circuit`
```

src/sage/matroids/matroid.pyx
```diff
@@ -2572,6 +2527,33 @@ cdef class Matroid(SageObject):
...
+
+            sage: M = matroids.Uniform(2, 4)
+            sage: list(M.nonbases_iterator())
+            []
+            sage: [sorted(X) for X in matroids.catalog.P6().nonbases_iterator()]
+            [['a', 'b', 'c']]
+
+        ALGORITHM:
+
+        Test all subsets of the groundset of cardinality ``self.full_rank()``
```
tscrim:
```suggestion
        Test all subsets of the groundset of cardinality ``self.full_rank()``.
```
Might also be better before the `EXAMPLES::`.

src/sage/matroids/matroid.pyx
```diff
@@ -2605,15 +2583,40 @@ cdef class Matroid(SageObject):
                 res.append(X)
         return res
 
+    def dependent_r_sets_iterator(self, long r):
+        r"""
+        Return an iterator over the dependent subsets of fixed size.
+
+        INPUT:
+
+        - ``r`` -- a nonnegative integer.
```
tscrim:
```suggestion
        - ``r`` -- a nonnegative integer
```

src/sage/matroids/matroid.pyx
```diff
@@ -2605,15 +2583,40 @@ cdef class Matroid(SageObject):
...
+            sage: list(M.dependent_r_sets_iterator(3))
+            []
+            sage: sorted([sorted(X) for X in
+            ....: matroids.catalog.Vamos().dependent_r_sets_iterator(4)])
+            [['a', 'b', 'c', 'd'], ['a', 'b', 'e', 'f'], ['a', 'b', 'g', 'h'],
+            ['c', 'd', 'e', 'f'], ['e', 'f', 'g', 'h']]
+
+        ALGORITHM:
+
+        Test all subsets of the groundset of cardinality ``r``
```
tscrim:
```suggestion
        Test all subsets of the groundset of cardinality ``r``.
```
Same comment as the other one.

src/sage/matroids/matroid.pyx
```diff
@@ -2636,18 +2639,77 @@ cdef class Matroid(SageObject):
...
+            sage: sorted([sorted(X) for X in M.bases_iterator()])
+            [[0, 1], [0, 2], [0, 3], [1, 2], [1, 3], [2, 3]]
+
+        ALGORITHM:
+
+        Test all subsets of the groundset of cardinality ``self.full_rank()``
+
+        .. SEEALSO::
+
+            :meth:`M.independent_r_sets() <sage.matroids.matroid.Matroid.independent_r_sets>`
```
tscrim:
```suggestion
        ALGORITHM:

        Test all subsets of the groundset of cardinality ``self.full_rank()``.

        .. SEEALSO::

            :meth:`M.independent_r_sets() <sage.matroids.matroid.Matroid.independent_r_sets>`
```
Same comment.

src/sage/matroids/matroid.pyx
```diff
@@ -2719,6 +2775,40 @@ cdef class Matroid(SageObject):
...
         return res
 
+    def independent_r_sets_iterator(self, r):
+        r"""
+        Return an iterator over the size-``r`` independent subsets of the
+        matroid.
+
+        INPUT:
+
+        - ``r`` -- a nonnegative integer.
```
tscrim:
```suggestion
        - ``r`` -- a nonnegative integer
```

src/sage/matroids/matroid.pyx
```diff
@@ -2719,6 +2775,40 @@ cdef class Matroid(SageObject):
...
+            []
+            sage: S = list(M.independent_r_sets_iterator(3))
+            sage: len(S)
+            75
+            sage: frozenset({'a', 'c', 'e'}) in S
+            True
+
+        ALGORITHM:
+
+        Test all subsets of the groundset of cardinality ``r``
```
tscrim, line 2804:
```suggestion
        Test all subsets of the groundset of cardinality ``r``.
```
Same comment (and for the `.. SEEALSO::`).

src/sage/matroids/matroid.pyx
```diff
@@ -2990,10 +3131,12 @@ cdef class Matroid(SageObject):
             minimal-removal convention, while the implementation is not
             modified from the published algorithm.
         """
-        cdef list rev_order
+        if len(self.loops()) > 0:
+            return []
```
tscrim:
```suggestion
        if self.loops():
            return []
```
Mirco-optimization.

src/sage/matroids/matroid.pyx
```diff
@@ -2990,10 +3131,12 @@ cdef class Matroid(SageObject):
...
 
+        cdef list rev_order
         if ordering is None:
-            rev_order = sorted(self.groundset(), reverse=True)
+            rev_order = sorted(self.groundset(), key=str, reverse=True)
```
tscrim: Same comment about sorting.

src/sage/matroids/matroid.pyx
```diff
@@ -3032,20 +3174,82 @@ cdef class Matroid(SageObject):
...
+        ::
+
+            sage: M = Matroid(circuits=[[1,2,3], [1,4,5], [2,3,4,5]])
+            sage: SimplicialComplex(list(M.no_broken_circuits_sets_iterator([5,4,3,2,1])))
+            Simplicial complex with vertex set (1, 2, 3, 4, 5)
+             and facets {(1, 3, 5), (2, 3, 5), (2, 4, 5), (3, 4, 5)}
+        """
+        if len(self.loops()) == 0:
+
+            if ordering is None:
```
tscrim:
```suggestion
        if not self.loops():
            if ordering is None:
```

src/sage/matroids/matroid.pyx
```diff
@@ -3032,20 +3174,82 @@ cdef class Matroid(SageObject):
...
         - ``R`` -- the base ring
-        - ``ordering`` -- (optional) an ordering of the ground set
-        - optional parameter ``invariant`` -- (optional, default: None) either
-          a semigroup ``G`` whose ``__call__`` acts on the groundset, or pair
-          ``(G, action)`` where ``G`` is a semigroup and ``action``
-          is a function ``action(g,e)`` which takes a pair of a group
-          element and a grounset element and returns the groundset
-          element which is the result of ``e`` acted upon by ``g``
+        - ``ordering`` -- (optional) an ordering of the groundset
+        - ``invariant`` -- (optional, default: None) either a semigroup ``G``
```
tscrim:
```suggestion
        - ``invariant`` -- (optional) either a semigroup ``G``
```

src/sage/matroids/matroid.pyx
```diff
@@ -3133,11 +3337,15 @@ cdef class Matroid(SageObject):
...
         vector_e = FreeModule(ZZ, n).basis()
         convert = {ind: i for i, ind in enumerate(self.groundset())}
-        vertices = [sum(vector_e[convert[i]] for i in B)
-                    for B in self.bases()]
+        vertices = []
+        for B in self.bases_iterator():
+            sum = 0
+            for i in B:
+                sum += vector_e[convert[i]]
+            vertices += [sum]
```
tscrim:
```suggestion
            total = 0
            for i in B:
                total += vector_e[convert[i]]
            vertices += [total]
```
Best to avoid special functions.

src/sage/matroids/matroid.pyx
```diff
@@ -3177,8 +3385,12 @@ cdef class Matroid(SageObject):
         ambient = FreeModule(ZZ, n)
         vector_e = ambient.basis()
         convert = {ind: i for i, ind in enumerate(self.groundset())}
-        vertices = [ambient.sum(vector_e[convert[i]] for i in IS)
-                    for IS in self.independent_sets()]
+        vertices = []
```
tscrim:
```suggestion
        cdef list lst, vertices = []
```

src/sage/matroids/matroid.pyx
```diff
@@ -3177,8 +3385,12 @@ cdef class Matroid(SageObject):
...
+        for IS in self.independent_sets_iterator():
+            lst = []
+            for i in IS:
+                lst += [vector_e[convert[i]]]
+            vertices += [ambient.sum(lst)]
```
tscrim:
```suggestion
                lst.append(vector_e[convert[i]])
            vertices.append(ambient.sum(lst))
```
Not sure if this is faster, but it could be. IMO, it is easier to understand the intent of adding a single element to the lists.

src/sage/matroids/matroid.pyx
```diff
@@ -3193,13 +3405,11 @@ cdef class Matroid(SageObject):
 
         INPUT:
 
-        - ``other`` -- A matroid,
-        - optional parameter ``certificate`` -- Boolean.
-
-        OUTPUT:
+        - ``other`` -- a matroid,
```
tscrim:
```suggestion
        - ``other`` -- a matroid
```

src/sage/matroids/matroid.pyx
```diff
@@ -3236,12 +3446,10 @@ cdef class Matroid(SageObject):
         INPUT:
 
         - ``other`` -- A matroid,
-        - optional parameter ``certificate`` -- Boolean.
-
-        OUTPUT:
+        - ``certificate`` -- boolean (default: ``False``).
```
tscrim:
```suggestion
        - ``certificate`` -- boolean (default: ``False``)
```

src/sage/matroids/matroid.pyx
```diff
@@ -6184,9 +6313,9 @@ cdef class Matroid(SageObject):
 
         if basis is None:
             basis = self.basis()
-        basis = sorted(basis)
+        basis = sorted(basis, key=str)
```
tscrim: Same comment about sorting.

src/sage/matroids/matroid.pyx
```diff
@@ -7293,8 +7398,9 @@ cdef class Matroid(SageObject):
 
         INPUT:
 
-        - ``other`` -- a matroid with the same ground set as ``self``.
-        - ``Y`` -- an common independent set of ``self`` and ``other`` of size `k`.
+        - ``other`` -- a matroid with the same groundset as ``self``
+        - ``Y`` -- an common independent set of ``self`` and ``other`` of size
+          `k`
```
tscrim:
```suggestion
        - ``Y`` -- an common independent set of ``self`` and ``other`` of size `k`
```
Even if this is slightly over the 80 char/line rule, it looks so much better in the source code and worth the violation.

src/sage/matroids/matroid.pyx
```diff
@@ -7622,15 +7722,63 @@ cdef class Matroid(SageObject):
...
         x, y = R._first_ngens(2)
         T = R(0)
-        for B in self.bases():
+        for B in self.bases_iterator():
             T += x ** len(self._internal(B)) * y ** len(self._external(B))
         if a is not None and b is not None:
             T = T(a, b)
         return T
 
+    cpdef characteristic_polynomial(self, l=None) noexcept:
```
tscrim:
```suggestion
    cpdef characteristic_polynomial(self, la=None) noexcept:
```
IMO a bit better variable name (and more clearly matches the doc).

src/sage/matroids/matroid.pyx
```diff
@@ -7622,15 +7722,63 @@ cdef class Matroid(SageObject):
...
+            sage: M.characteristic_polynomial()
+            l^4 - 10*l^3 + 35*l^2 - 50*l + 24
+            sage: M.characteristic_polynomial().factor()
+            (l - 4) * (l - 3) * (l - 2) * (l - 1)
+            sage: M.characteristic_polynomial(5)
+            24
+
+        .. SEEALSO::
+
+            :meth:`whitney_numbers() <sage.matroids.matroid.Matroid.whitney_numbers>`
```
tscrim:
```suggestion
            :meth:`~sage.matroids.matroid.Matroid.whitney_numbers`
```

src/sage/matroids/matroid.pyx
```diff
@@ -7622,15 +7722,63 @@ cdef class Matroid(SageObject):
...
+        R = ZZ['l']
+        l = R._first_ngens(1)[0]
+        chi = R(0)
+        cdef int r = self.rank()
+        cdef list w = self.whitney_numbers()
+        for i in range(len(w)):
+            chi += w[i] * l**(r-i)
+        if val is not None:
+            return chi(val)
+        return chi
```
tscrim:
```suggestion
        chi = self._generic_char_poly()
        if la is not None:
            return chi(la)
        return chi

    @cached_method
    def _generic_char_poly(self):
        R = ZZ['l']
        cdef list w = self.whitney_numbers()
        w.reverse()
        return R(w)
```
I would cache the generic case and then call as appropriate. Also, let the poly ring handle the list itself; it likely can construct it faster.

src/sage/matroids/matroid.pyx
```diff
@@ -7765,7 +7919,7 @@ cdef class Matroid(SageObject):
         # Create the ambient polynomial ring
         from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
         try:
-            names = ['A{}'.format(''.join(str(x) for x in sorted(F))) for F in flats]
+            names = ['A{}'.format(''.join(str(x) for x in sorted(F, key=str))) for F in flats]
```
tscrim: Same comment about sorting.

src/sage/matroids/matroid.pyx
```diff
@@ -7957,7 +8115,12 @@ cdef class Matroid(SageObject):
              and facets {(1, 3, 5), (1, 4, 5), (2, 3, 5), (2, 4, 5)}
         """
         from sage.topology.simplicial_complex import SimplicialComplex
-        return SimplicialComplex(self.no_broken_circuits_sets(ordering))
+        cdef int r = self.rank()
+        cdef list facets = []
+        for S in self.no_broken_circuits_sets_iterator(ordering):
+            if len(S) == r:
+                facets += [S]
```
tscrim:
```suggestion
                facets.append(S)
```

tscrim (changes requested):
A lot of comments are the same, but I wanted to make a note of them (specifically the sorting). This could potentially generate a lot of merge conflicts, but hopefully that won't be a problem...

src/sage/matroids/basis_exchange_matroid.pyx
gmou3: Oops, I didn't think of that. I want to allow the groundset to have elements of mixed type (integers and strings). I have fixed it now.

src/sage/matroids/matroid.pyx
gmou3: This wouldn't compile so I did something in between.

src/sage/matroids/matroid.pyx
```diff
@@ -531,9 +531,7 @@ cdef class Matroid(SageObject):
         - ``X`` -- An object with Python's ``frozenset`` interface containing
           a subset of ``self.groundset()``.
 
-        OUTPUT:
-
-        ``frozenset`` instance containing a subset of the groundset.
+        OUTPUT: ``frozenset`` instance containing a subset of the groundset
```
mkoeppe:
```suggestion
        OUTPUT: :class:`frozenset` instance containing a subset of the groundset
```
(throughout)

or perhaps: ```a subset of the groundset as a :class:`frozenset` ```

src/sage/matroids/matroid.pyx
```diff
@@ -2383,21 +2291,52 @@ cdef class Matroid(SageObject):
...
-        Return the list of nonspanning circuits of the matroid.
+        Return the nonspanning circuits of the matroid.
 
         A *nonspanning circuit* is a circuit whose rank is strictly smaller
         than the rank of the matroid.
 
-        OUTPUT:
-
-        An iterable containing all nonspanning circuits.
+        OUTPUT: a SetSystem
```
mkoeppe:
```suggestion
        OUTPUT: a :class:`SetSystem`
```

src/sage/matroids/matroid.pyx
```diff
@@ -3032,20 +3175,82 @@ cdef class Matroid(SageObject):
...
-          a semigroup ``G`` whose ``__call__`` acts on the groundset, or pair
-          ``(G, action)`` where ``G`` is a semigroup and ``action``
-          is a function ``action(g,e)`` which takes a pair of a group
-          element and a grounset element and returns the groundset
-          element which is the result of ``e`` acted upon by ``g``
+        - ``ordering`` -- (optional) an ordering of the groundset
+        - ``invariant`` -- (optional) either a semigroup ``G`` whose
+          ``__call__`` acts on the groundset, or pair ``(G, action)`` where
+          ``G`` is a semigroup and ``action`` is a function ``action(g,e)``
+          which takes a pair of a group element and a grounset element and
```
mkoeppe:
```suggestion
          which takes a pair of a group element and a groundset element and
```

src/sage/matroids/matroid.pyx
```diff
@@ -5275,9 +5422,7 @@ cdef class Matroid(SageObject):
             to calling ``is_kconnected(4,certificate)``.
```
mkoeppe:
```suggestion
            to calling ``is_kconnected(4, certificate)``.
```

src/sage/matroids/matroid.pyx
```diff
@@ -5643,8 +5782,8 @@ cdef class Matroid(SageObject):
 
         OUTPUT:
 
-        - `False, None`  -- if there is no ``m``-separator.
-        - `True, E` -- if there exist a ``m``-separator ``E``.
+        - `False, None`  -- if there is no ``m``-separator
+        - `True, E` -- if there exist an ``m``-separator ``E``
```
mkoeppe:
```suggestion
        - ``False, None``  -- if there is no ``m``-separator
        - ``True, E`` -- if there exist an ``m``-separator ``E``
```

src/sage/matroids/matroid.pyx
```diff
@@ -6056,14 +6191,12 @@ cdef class Matroid(SageObject):
...
-        - ``verify`` -- (default: ``True``), a Boolean; if ``True``,
+        - ``verify`` -- boolean (default: ``True``); if ``True``,
           any output will be a binary matroid representing ``self``; if
           ``False``, any output will represent ``self`` if and only if the
           matroid is binary
 
-        OUTPUT:
-
-        Either a BinaryMatroid, or ``None``
+        OUTPUT: either a ``BinaryMatroid``, or ``None``
```
mkoeppe:
```suggestion
        OUTPUT: either a :class:`BinaryMatroid`, or ``None``
```

src/sage/matroids/matroid.pyx
```diff
@@ -7622,15 +7723,72 @@ cdef class Matroid(SageObject):
...
+            l^4 - 10*l^3 + 35*l^2 - 50*l + 24
+            sage: M.characteristic_polynomial().factor()
+            (l - 4) * (l - 3) * (l - 2) * (l - 1)
+            sage: M.characteristic_polynomial(5)
+            24
+
+        .. SEEALSO::
+
+            :meth:`~sage.matroids.matroid.Matroid.whitney_numbers`
+
```
mkoeppe, line 7778:
```suggestion
```

src/sage/matroids/matroid.pyx
```diff
@@ -7622,15 +7723,72 @@ cdef class Matroid(SageObject):
...
+        TESTS::
+
+            sage: M = Matroid(groundset=[0,1,2], circuits=[[0]])
+            sage: M.characteristic_polynomial()
+            0
+            sage: l = -1
+            sage: for M in matroids.AllMatroids(6):  # optional - matroid_database
+            ....:     r = M.rank()
+            ....:     assert M.characteristic_polynomial(l) == (-1)**r * M.tutte_polynomial(1-l, 0)
```
mkoeppe:
```suggestion
            ....:     assert M.characteristic_polynomial(l) == (-1)**r * M.tutte_polynomial(1 - l, 0)
```

mkoeppe:
Overall LGTM

src/sage/matroids/matroid.pyx
```diff
@@ -1149,9 +1115,9 @@ cdef class Matroid(SageObject):
             if certificate:
                 return False, None
             return False
-        YY = self.dual().independent_r_sets(cd)
-        for X in self.independent_r_sets(rd):
-            for Y in YY:
+        D = self.dual()
+        for X in self.independent_r_sets_iterator(rd):
+            for Y in D.independent_r_sets_iterator(cd):
```
tscrim: I am not sure this is the best change as it has to regenerate the independent sets for each iteration. I would continue to use the list version for `D`.

src/sage/matroids/matroid.pyx
```diff
@@ -2383,21 +2291,52 @@ cdef class Matroid(SageObject):
...
+            for X in combinations(self.groundset(), k):
+                X = frozenset(X)
+                if self._is_circuit(X):
+                    yield X
+        elif not k:
+            for k in range(self.rank() + 2):
+                for X in combinations(self.groundset(), k):
+                    X = frozenset(X)
+                    if self._is_circuit(X):
+                        yield X
```
tscrim:
```suggestion
        rk = self.rank()
        if k is None:
            start = 0
            stop = rk + 2
        else:
            if k >= rk + 2:
                return
            start = k
            stop = k + 1
        for j in range(start, stop):
            for X in combinations(self.groundset(), j):
                X = frozenset(X)
                if self._is_circuit(X):
                    yield X
```
I think this simplifies the logic. I don't expect the extra little overhead will be consequential how this is used.

src/sage/matroids/matroid.pyx
```diff
@@ -2703,6 +2764,38 @@ cdef class Matroid(SageObject):
             sage: frozenset({'a', 'c', 'e'}) in S
             True
 
+        .. SEEALSO::
+
+            :meth:`M.independent_sets() <sage.matroids.matroid.Matroid.independent_sets>`
+            :meth:`M.bases() <sage.matroids.matroid.Matroid.bases>`
+        """
+        res = []
```
tscrim:
```suggestion
        cdef list res = []
```

src/sage/matroids/matroid.pyx
```diff
@@ -2882,32 +2966,90 @@ cdef class Matroid(SageObject):
         r"""
         Return the `f`-vector of the matroid.
 
-        The `f`-*vector* is a vector `(f_0, ..., f_r)`, where `f_i` is the
-        number of flats of rank `i`, and `r` is the rank of the matroid.
+        The `f`-*vector* is a vector `(f_0, \dots, f_r)`, where `f_i` is the
```
tscrim:
```suggestion
        The `f`-*vector* is a vector `(f_0, \ldots, f_r)`, where `f_i` is the
```
`\dots` are evil; it's good to be specific.

src/sage/matroids/matroid.pyx
```diff
@@ -2882,32 +2966,90 @@ cdef class Matroid(SageObject):
...
+            f.append(sum)
+        return f
 
-        List of integers.
+    cpdef whitney_numbers(self) noexcept:
+        r"""
+        Return the Whitney numbers of the first kind of the matroid.
+
+        The Whitney numbers of the first kind -- here encoded as a vector
+        `(w_0=1, ..., w_r)` -- are numbers of alternating sign, where `w_i` is
```
tscrim: Why not change to `\ldots` here (and below)?

src/sage/matroids/matroid.pyx
```diff
@@ -3032,20 +3175,82 @@ cdef class Matroid(SageObject):
...
+
+        ::
+
+            sage: M = Matroid(circuits=[[1,2,3], [1,4,5], [2,3,4,5]])
+            sage: SimplicialComplex(list(M.no_broken_circuits_sets_iterator([5,4,3,2,1])))
+            Simplicial complex with vertex set (1, 2, 3, 4, 5)
+             and facets {(1, 3, 5), (2, 3, 5), (2, 4, 5), (3, 4, 5)}
+        """
+        if not self.loops():
+
```
tscrim, line 3218:
```suggestion
```

tscrim (changes requested):
I have a few more changes to request; minor stuff basically.

gmou3:
Thanks to both for the suggestions.

src/sage/matroids/matroid.pyx
```diff
@@ -3193,13 +3408,11 @@ cdef class Matroid(SageObject):
...
-        - ``other`` -- A matroid,
-        - optional parameter ``certificate`` -- Boolean.
-
-        OUTPUT:
+        - ``other`` -- a matroid
+        - ``certificate`` -- boolean (default: ``False``)
 
-        Boolean,
-        and, if certificate = True, a dictionary or None
+        OUTPUT: boolean, and, if ``certificate = True``, a dictionary or
```
mkoeppe:
```suggestion
        OUTPUT: boolean, and, if ``certificate=True``, a dictionary or
```

src/sage/matroids/matroid.pyx
```diff
@@ -8159,23 +8337,21 @@ cdef class Matroid(SageObject):
...
 
-    def direct_sum(self, matroids):
+    cpdef direct_sum(self, matroids) noexcept:
         r"""
         Return the matroid direct sum with another matroid or list of
         matroids.
 
         Let `(M_1, M_2, \ldots, M_k)` be a list of matroids where each `M_i`
-        has ground set `E_i`. The matroid sum `(E_1,I_1),\ldots,(E_n,I_n)`
+        has groundset `E_i`. The matroid sum `(E_1,I_1),\ldots,(E_n,I_n)`
```
mkoeppe: should this not say the "matroid sum of ... is"?

Or use an operator instead of commas?

src/sage/matroids/matroid.pyx
```diff
@@ -5932,19 +6069,21 @@ cdef class Matroid(SageObject):
...
             True
+            sage: M = matroids.Theta(4)
+            sage: M.is_paving()
+            False
         """
-        for C in self.circuits():
-            if len(C) < self.rank():
-                return False
+        if self.rank() >= 2:
+            for X in combinations(self.groundset(), self.rank()-1):
```
mkoeppe:
```suggestion
            for X in combinations(self.groundset(), self.rank() - 1):
```

src/sage/matroids/matroid.pyx
```diff
@@ -3149,7 +3360,7 @@ cdef class Matroid(SageObject):
 
         over all independent sets `I` of the matroid. Here `e_i` are
         the standard basis vectors of `\RR^n`. An arbitrary labelling
-        of the groundset by `{0,\dots,n-1}` is chosen.
+        of the groundset by `{0,\ldots,n-1}` is chosen.
```
mkoeppe:
```suggestion
        of the groundset by `\{0,\ldots,n-1\}` is chosen.
```

src/sage/matroids/matroid.pyx
```diff
@@ -3106,7 +3313,7 @@ cdef class Matroid(SageObject):
 
         over all bases `B` of the matroid. Here `e_i` are the standard
         basis vectors of `\RR^n`. An arbitrary labelling of the
-        groundset by `{0,\dots,n-1}` is chosen.
+        groundset by `{0,\ldots,n-1}` is chosen.
```
mkoeppe:
```suggestion
        groundset by `\{0,\ldots,n-1\}` is chosen.
```

mkoeppe (approved):

src/sage/matroids/matroid.pyx
```diff
@@ -2882,32 +2969,90 @@ cdef class Matroid(SageObject):
...
+            sage: for M in matroids.AllMatroids(5):                                     # optional - matroid_database
+            ....:     assert M.f_vector() == SimplicialComplex(M.bases()).f_vector()
+        """
+        cdef list f = []
+        cdef int i, sum
+        for i in range(self.full_rank() + 1):
+            sum = 0
+            for _ in self.independent_r_sets_iterator(i):
+                sum += 1
+            f.append(sum)
```
tscrim:
```suggestion
        cdef int i, s
        for i in range(self.full_rank() + 1):
            s = 0
            for _ in self.independent_r_sets_iterator(i):
                sum += 1
            f.append(ZZ(sum))
```
Two little things here: let's not use Python special function names and we should try to make sure `int`'s don't leak out.

src/sage/matroids/matroid.pyx
```diff
@@ -2882,32 +2969,90 @@ cdef class Matroid(SageObject):
...
+        TESTS::
+
+            sage: M = Matroid(groundset=[0,1,2], circuits=[[0]])
+            sage: M.whitney_numbers()
+            []
+        """
+        cdef list abs_w = [0] * (self.rank()+1)
+        for S in self.no_broken_circuits_sets_iterator():
+            abs_w[len(S)] += 1
+        return [(-1)**i * abs_w[i] for i in range(len(abs_w)) if abs_w[i] != 0]
```
tscrim:
```suggestion
        return [ZZ((-1)**i * val) for i, val in enumerate(abs_w) if val != 0]
```
`enumerate` is slightly faster and we don't want `int`s to leak out.

src/sage/matroids/matroid.pyx
```diff
@@ -2882,32 +2969,90 @@ cdef class Matroid(SageObject):
...
+        EXAMPLES::
+
+            sage: M = matroids.catalog.BetsyRoss()
+            sage: M.whitney_numbers2()
             [1, 11, 20, 1]
         """
         loops = self._closure(set())
         flags = [[loops, set(), self.groundset() - loops]]
-        f_vec = [1]
+        W = [1]
```
tscrim:
```suggestion
        W = [ZZ.one()]
```

src/sage/matroids/matroid.pyx
```diff
@@ -2882,32 +2969,90 @@ cdef class Matroid(SageObject):
...
         for r in range(self.full_rank()):
             flags = self._extend_flags(flags)
-            f_vec.append(len(flags))
-        return f_vec
+            W.append(len(flags))
```
tscrim:
```suggestion
            W.append(ZZ(len(flags))
```
`int`s again.

src/sage/matroids/matroid.pyx
```diff
@@ -3032,20 +3178,81 @@ cdef class Matroid(SageObject):
...
+             and facets {(1, 3, 5), (1, 4, 5), (2, 3, 5), (2, 4, 5)}
+
+        ::
+
+            sage: M = Matroid(circuits=[[1,2,3], [1,4,5], [2,3,4,5]])
+            sage: SimplicialComplex(list(M.no_broken_circuits_sets_iterator([5,4,3,2,1])))
+            Simplicial complex with vertex set (1, 2, 3, 4, 5)
+             and facets {(1, 3, 5), (2, 3, 5), (2, 4, 5), (3, 4, 5)}
+        """
+        if not self.loops():
```
tscrim: This is a slight change from what we current have for a matroid with loops., where it returns the empty set. I am not sure offhand what behavior is correct. In either case, we should document it. I would also do this (possibly with a `yield emptyset()`) to avoid having a big indentation level too:
```suggestion
        if self.loops():
            return
        [rest of method]
```

src/sage/matroids/matroid.pyx
```diff
@@ -3177,8 +3388,12 @@ cdef class Matroid(SageObject):
...
         vector_e = ambient.basis()
         convert = {ind: i for i, ind in enumerate(self.groundset())}
-        vertices = [ambient.sum(vector_e[convert[i]] for i in IS)
-                    for IS in self.independent_sets()]
+        cdef list lst, vertices = []
+        for IS in self.independent_sets_iterator():
+            lst = []
+            for i in IS:
+                lst.append(vector_e[convert[i]])
+            vertices.append(ambient.sum(lst))
```
tscrim:
```suggestion
            lst = [None] * len(IS)
            for ind, i in enumerate(IS):
                lst[ind] = vector_e[convert[i]]
            vertices.append(ambient.sum(lst))
```
It is faster to initially populate the list and update the elements IIRC.

tscrim (changes requested):
Sorry, just a few more little things.

src/sage/matroids/matroid.pyx
gmou3: For a matroid with loops the correct behavior is no sets at all. I am planning to do some more work on the broken circuit methods in another PR.

tscrim (approved):
Okay, thank you. Good to go.

mkoeppe:
@gmou You may want to merge #37667 and update the changed pxd files

PR #37366, opened 2024-02-16
use Parent in skew polynomials

fchapoton:
instead of the auld-class `Ring`

### :memo: Checklist


tscrim (approved):
LGTM.

PR #37367, opened 2024-02-16
improve random sampling of quotient-ring elements

yyyyx4:
Currently, random sampling in generic quotient rings is restricted to a very small (and special) subset of the elements:
```sage
sage: R.<x,y> = QQ[]
sage: S = R.quotient([x^3, y^2])
sage: {S.random_element() for _ in range(999)}
{-2, -1, 0, 1, 2}
```

In this patch we add an implementation of `.random_element()` which simply calls the `.random_element()` method of the cover ring and maps the result to the quotient. This is still far from perfect for many kinds of quotient rings, but it's definitely an improvement compared to the current behavior.

### DIFF
--- a/src/sage/rings/quotient_ring.py
+++ b/src/sage/rings/quotient_ring.py
@@ -1354,6 +1354,30 @@ class QuotientRing_nc(ring.Ring, sage.structure.parent_gens.ParentWithGens):
         """
         return self.__R.term_order()
 
+    def random_element(self):
+        r"""
+        Return a random element of this quotient ring obtained by
+        sampling a random element of the cover ring and reducing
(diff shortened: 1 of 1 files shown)


GiacomoPope (approved):
Looks good to me, and even if there's a smarter method this could be done in the future. I think this is good to go.

tscrim:
It would be good to also pass options along.

yyyyx4:
Good point, but it looks like @vbraun is already merging this — is it okay to still change the branch now?

tscrim:
Hmmm….possibly no. Let’s just make another PR, and cc me for the review.

PR #37372, opened 2024-02-16
Strength 2 Covering Array constructions

### LINKED ISSUE #37371: Adding strength 2 covering array methods
### Problem Description

Now that the basics of covering arrays are included in SageMath, some methods to generate covering arrays can be added. 

This is a follow up to #34279

### Proposed Solution

The number of known covering arrays with optimal size is relatively small, and all have strength t=2, so a good place to start is to implement all of them.

### Alternatives Considered

Some strength 3 methods would be relatively simple to add as well.

### Additional Information

_No response_

aadwyer:
This patch aims to add functionality to the covering arrays module that was recently added to SageMath. It adds methods that create many covering arrays of strength 2 that can later be used for recursive constructions as well.

Specifically: added methods to generate covering arrays and a main CA function to call them based on inputted parameters. 
Added methods include direct implementation of optimal arrays in the database of small combinatorial designs, methods that create a CA(N,t=2,k,v=2), for either k or N inputted and a recursive method that removes columns from a larger array. Added required documentation and reference for new methods.

This fixes #37371


src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+        sage: C = Kleitman_Spencer_Katona(2)
+        Traceback (most recent call last):
+        ...
+        ValueError: N must be greater than 3
+        sage: C = Kleitman_Spencer_Katona(5)
+        sage: is_covering_array(C,parameters=True)
+        (True, (5, 2, 4, 2))
+
+    """
+    from math import ceil
```
mkoeppe: Consider using `sage.arith.misc.integer_ceil`

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+def covering_array(strength, number_columns, levels):
+    r"""
+    Build a `CA(N; t, k, v)` using direct constructions, where `N` is the
+    smallest size known.
+
+    INPUT:
+
+    - ``strength`` (integer) -- the parameter `t` of the covering array,
+      such that in any selection of `t` columns of the array, every `t`
+      -tuple appears at least once.
```
mkoeppe:
```suggestion
      such that in any selection of `t` columns of the array, every 
      `t`-tuple appears at least once.
```

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+        for i in p:
+            S[i] = 1
+        result.append(S)
+    return(list(map(list, zip(*result))))
+
+
+def column_Kleitman_Spencer_Katona(k):
+    r"""
+    Return a covering array with `k` columns using the Kleitman Spencer Katona
+    method.
```
mkoeppe:
```suggestion
    Return a covering array with `k` columns using the Kleitman-Spencer-Katona
    method.
```

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+        return column_Kleitman_Spencer_Katona(number_columns)
+
+    elif 3 <= levels <= 6:
+        import sage.combinat.designs.database as DB
+        for i in DB.CA_constructions[(strength, levels)]:
+            if number_columns <= i[1]:
+                CA = "ca_{}_{}_{}_{}".format(i[0], strength, i[1], levels)
+                f = getattr(DB, CA)
+                return truncate_columns(f(), number_columns)
+
```
mkoeppe, line 261: can this case fall through here?

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+        sage: C5 = covering_array(2, 50, 7)
+        No direct construction known and/or implemented for a CA(N; 2, 50, 7)
+
+    """
+    from sage.combinat.designs.orthogonal_arrays import orthogonal_array
+
+    if strength != 2:
+        raise ValueError("CAs and OAs only implemented for strength 2")
+
+    elif levels == 2:
```
mkoeppe:
```suggestion
    if levels == 2:
```
because the `if` branch always raises

src/sage/combinat/designs/database.py
```diff
@@ -5033,6 +5035,462 @@ def BIBD_56_11_2():
                         k, ', '.join('`{}`'.format(q) for q in sorted(EDS[k]) if EDS[k][q] is not False))
                         for k in sorted(EDS))
 
+
+def ca_11_2_5_3():
+    r"""
+    Return a CA with the given parameters, This CA is proven to be optimal.
```
mkoeppe:
```suggestion
    Return a CA with the given parameters. This CA is proven to be optimal.
```

(same several times below)

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+    else:
+        result = []
+        for row in array:
+            result.append(row[:k])
+        return result
+
+
+def Kleitman_Spencer_Katona(N):
+    r"""
+    Return a `CA(N; 2, k, 2)` where `k = \binom {N-1}{\lceil \frac{N}{2} \rceil}`.
```
mkoeppe:
```suggestion
    Return a `CA(N; 2, k, 2)` where `k = \binom {N-1}{\lceil N/2 \rceil}`.
```
maybe better to use inline division?

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+        sage: column_Kleitman_Spencer_Katona(25000)
+        Traceback (most recent call last):
+        ...
+        AssertionError: not implemented for k > 24310
+
+    """
+    assert k <= 24310, "not implemented for k > 24310"
+
+    kdict = {3:4,4:5,10:6,15:7,35:8,56:9,126:10,210:11,462:12,792:13,
+            1716:14,3003:15,6435:16,11440:17,24310:18}
```
mkoeppe:
```suggestion
    kdict = {3: 4, 4: 5, 10: 6, 15: 7, 35: 8, 56: 9, 126: 10, 210: 11, 462: 12, 792: 13,
            1716: 14, 3003: 15, 6435: 16, 11440: 17, 24310: 18}
```

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+
+    elif oldk < k:
+        raise ValueError("array only has {} columns".format(oldk))
+
+    else:
+        result = []
+        for row in array:
+            result.append(row[:k])
+        return result
+
```
sraaphorst: I'm not sure about the Sage guideline, but in the interests of efficiency, the `else` part is probably better done:
```
return [row[:k] for row in array]
```
Then the looping happens behind the hood in C instead of in Python. For objects of these sizes, it probably doesn't matter, though.
for efficiency. It's best to

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+    if N < 4:
+        raise ValueError("N must be greater than 3")
+
+    result = []
+    for p in combinations(range(N-1), ceil(N/2)):
+        S = [0]*N
+        for i in p:
+            S[i] = 1
+        result.append(S)
+    return(list(map(list, zip(*result))))
```
sraaphorst: The logic here is correct, but the last line is confusing. You might want to build the actual matrix instead of its transpose, or at least add a line before the last line saying that you are taking the transpose. `map` is generally not used much in Python 3+ since it is typically considerably slower than just using list comprehensions.

sraaphorst: Also, I don't know about SageMath conventions, but in Python, the convention is to not use caps in function names. Just thought I'd point it out.

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+    assert k <= 24310, "not implemented for k > 24310"
+
+    kdict = {3:4,4:5,10:6,15:7,35:8,56:9,126:10,210:11,462:12,792:13,
+            1716:14,3003:15,6435:16,11440:17,24310:18}
+
+    for ki in kdict:
+        if k <= ki:
+            N = kdict[ki]
+            break
+    return truncate_columns(Kleitman_Spencer_Katona(N), k)
```
sraaphorst: The `kdict` approach may not work: Python does not guarantee that dictionaries maintain their order. If it places the 24310 at the beginning, for example, you will always get 18.

I would implement this as a list of tuple:
```
kdict = [(3, 4), (4, 5), (10, 6), (15, 7), (35, 8), (56, 9), (126, 10), (210, 11), (462, 12), (792, 13), (1716, 14), (3003, 15), (6435, 16), (11440, 17), (24310, 18)]
```

I also wouldn't hard-copy the maximum value of `k` in the code like you do above. If you can ever extend `kdict`, then you'll have to change it in two places. I would do (after the above definition):

```
assert k <= kdict[-1][0]
```

Then the rest of the logic turns into:

```
for ki, N in kdict:
    if k <= ki:
        return truncate_columns(Kleitman_Spencer_Katona(N), k)
```

Depending on the version of Python you're using, all of this can probably be replaced by the `bisect` library, using, I believe, the `bisect_left` method.

src/sage/combinat/designs/covering_array.py
sraaphorst: Right. No point in having an `elif` here.

I'm aslo not sure why `assert` is being used in the preceding function, but exceptions are being used in this function. One or the other should be picked.

I would probably change the assert in the previous function to an `if ...: raise ValueError(...)`.

sraaphorst (changes requested):
This looks good in terms of logic. Some changes should be made for standardization and correctness.

src/sage/combinat/designs/covering_array.py
sraaphorst: When you say "fall through," what do you mean? I'm trying to understand the comment.

aadwyer: I am not sure if I understand either, hopefully the changes I made in the recent patch resolve this. If not please let me know what you mean.

src/sage/combinat/designs/covering_array.py
aadwyer: I am not sure, the style guide does say the same thing, but other functions in sage capitalize names I believe.

src/sage/combinat/designs/covering_array.py
aadwyer: The first suggestion is much appreciated. Also bisect_left was very interesting to learn about but I am not sure what is faster when the lists are so small like this, even if it is extended I do not imagine it will be extended very much.

aadwyer:
The Lint check is failing because of a 'check == True' statement I made. I used it because the function 'orthogonal_array' may output 'Unknown', in which case I would like to ignore it. What is the recommended way that I handle this?

src/sage/combinat/designs/covering_array.py
mkoeppe: > Python does not guarantee that dictionaries maintain their order.

Not true any more since Python 3.7. Dictionary maintain insertion order

src/sage/combinat/designs/covering_array.py
```diff
@@ -50,3 +54,190 @@
...
+        sage: C = column_Kleitman_Spencer_Katona(20)
+        sage: is_covering_array(C,parameters=True)
+        (True, (8, 2, 20, 2))
+        sage: column_Kleitman_Spencer_Katona(25000)
+        Traceback (most recent call last):
+        ...
+        AssertionError: not implemented for k > 24310
+
+    """
+    assert k <= 24310, "not implemented for k > 24310"
```
mkoeppe: This should be an `if... raise NotImplementedError(...)`

src/sage/combinat/designs/database.py
```diff
@@ -5033,6 +5035,462 @@ def BIBD_56_11_2():
...
+
+# Make dictionary with keys (t, v) and values (N, k) which are the
+# smallest N and largest k such that a CA(N; t, k, v) can be made using
+# the database.
+CA_constructions = {
+    (2,3) : ((11,5), (12,7), (13,9), (14,10), (15,20)),
+    (2,4) : ((19,6), (21,7)),
+    (2,5) : ((29,7),),
+    (2,6) : ((37,4), (39,5), (41,6))
+}
```
mkoeppe:
```suggestion
CA_constructions = {
    (2,3): ((11,5), (12,7), (13,9), (14,10), (15,20)),
    (2,4): ((19,6), (21,7)),
    (2,5): ((29,7),),
    (2,6): ((37,4), (39,5), (41,6))
}
```
coding style

src/sage/combinat/designs/covering_array.py
mkoeppe: "Fall through" == a case in which the body of the `if` is not terminated by `return`.

src/sage/combinat/designs/covering_array.py
```diff
@@ -49,3 +54,216 @@
...
+        sage: is_covering_array(E,parameters=True)
+        (True, (11, 2, 4, 3))
+
+    """
+    oldk = len(array[0])
+
+    if oldk == k:
+        return array
+
+    elif oldk < k:
```
mkoeppe: In an `if ... return`, best to continue with just `if`, not `elif`

src/sage/combinat/designs/covering_array.py
```diff
@@ -49,3 +54,216 @@
...
+    from sage.combinat.designs.orthogonal_arrays import orthogonal_array
+
+    if levels == 2 and strength == 2:
+        return column_Kleitman_Spencer_Katona(number_columns)
+
+    in_database = database_check(number_columns, strength, levels)
+    if in_database:
+        return in_database
+
+    if orthogonal_array(number_columns, levels, strength, existence=True) == True:
```
mkoeppe: see message from Ruff

aadwyer: https://github.com/sagemath/sage/pull/37372#issuecomment-2174348903

I noticed this but I am not sure how to proceed to make it happy while also keeping the functionality we need.

mkoeppe:
```suggestion
    if orthogonal_array(number_columns, levels, strength, existence=True) is True:
```

src/sage/combinat/designs/covering_array.py
sraaphorst: I agree with this comment. This should probably be written as:

```python
oldk = len(array[0])

if oldk == k:
    return array

if oldk < k:
    raise ValueError("...")

return [row[:k] for row in array]
```

Handle the exceptional cases first that return trivial results and handle bad input, and then present the default case.

src/sage/combinat/designs/covering_array.py
sraaphorst: Yes, you are correct, @aadwyer: maps are now ordered.

src/sage/combinat/designs/covering_array.py
sraaphorst: Ah. I see what you mean now. Yes, if the `for` terminates and no `return` is triggered, the code doesn't reach a `return` at all.

aadwyer: Thank you, there definitely was a problem with the code at that point but the current version does not fall through when 3<v<6.

src/sage/combinat/designs/covering_array.py
aadwyer: Because there are only a small number of entries, the NotImplementedError is raised fast and it will often have to read each entry in either case, I will keep this as a list despite the fact dictionaries could work now.

mkoeppe (approved):
LGTM

PR #37375, opened 2024-02-16
Speed up square matrix times vector over GF(2)

kedlaya:
When doing a matrix-times-vector multiplication over F_2, there is some inefficiency caused by calling `VectorSpace` to create a parent for the result. This becomes noticeable when repeatedly multiplying small matrices.

When the dimensions are all the same, this can be improved by using the parent of the vector also as the parent for the result. Before:
```
sage: A = random_matrix(GF(2),10,10)
sage: v0 = vector(random_matrix(GF(2),10,1))
sage: %timeit A*v0
2.26 µs ± 4.57 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
After:
```
sage: A = random_matrix(GF(2),10,10)
sage: v0 = vector(random_matrix(GF(2),10,1))
sage: %timeit A*v0
981 ns ± 9.57 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```


yyyyx4 (approved):
Looks good to me.

grhkm21:
Just a note, your timing doesn't show anything, since the dimensions are different (one's with $n = 100$, one is with $n = 10$)

yyyyx4:
Oh! Good catch. But there is still a significant speedup: The reduction in runtime is about $50\\,\\%$ for $10\times10$ matrices and about $20\\,\\%$ for $100\times100$ matrices.

kedlaya:
I fixed the examples to both use 10 by 10 matrices, as indeed this problem is more acute with small matrices. 

There seems to be a fixed amount of overhead with matrix/vector creation which is rampant throughout Sage; a broader fix for that would obviate the need for such trickery.

PR #37377, opened 2024-02-17
fix bad Frac(sparse polynomial ring over finite field)

mezzarobba:
In passing, add support for fraction fields of polynomial rings over prime fields based on NTL.

Fixes #37374

PR #35949, opened 2023-07-13
Compute composite degree (separable) isogenies of EllipticCurves

grhkm21:
- Implement `.isogenies_degree` for EllipticCurves

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

PR #37388, opened 2024-02-18
Add `.is_cyclic()` method for elliptic curves isogenies, and `cyclic_after` flag for `.isogenies_prime_degree()`

hellman:
This PR provides method to check if an isogeny between elliptic curves is cyclic (i.e., separable with cyclic degree).

This includes a generic method in `sage.schemes.elliptic_curves.hom.EllipticCurveHom.is_cyclic` which checks for separability and being non-zero first. Then, for each divisor $\ell^2$ of the isogeny degree, $\ell$ prime not equal to the characteristic, it checks that the division polynomial $\psi_{\ell}$ on the domain curve does not divide the kernel polynomial of the isogeny.

Then, it also includes methods for specific isogeny subclasses (scalar multiplication, frobenius, composite isogeny).

For composite isogenies, the algorithm is more involved. Again, we first check separability and check all suitable $\ell$ as follows. We skip isogenies with degrees non-divisible by $\ell$ and also skip the longest prefix of $\ell$-isogenies for which by checking j-invariants we see that they are non-backtracking. Then, we are left with one contiguous sub-list of isogenies. We start with the $\ell$-th division polynomial $m=\psi_{\ell}$ on the first curve of the sublist, and propagate it through the isogenies in the sublist, reducing $m$ by its gcd with each kernel polynomial. If it's fully destroyed ($m=1$), the isogeny is non-cyclic. There is also an optimization which replaces $m$ by its image on the second curve, which is essentiall the kernel polynomial of the dual of the first isogeny. This is useful because $m$ then has degree $\ell-1$ instead of $\ell^2-1-\ell$. (This happens because on the first curve only the subgroup is killed, but not its cosets; so we rather look at the image).

Small benchmark for the optimizations: https://gist.github.com/hellman/409ba128c67d6887c9a130c19b66f01a

This method is also very useful to construct isogeny chains (for example, for isogeny search, meet-in-the-middle, #35949 ). For this purpose, I also added an option `cyclic_after` for `E.isogenies_prime_degree` which allows to consider only non-backtracking (in the sense of being non-dual) isogenies.

Thanks to @yyyyx4 @rtb123 @remyoudompheng @grhkm21 @GiacomoPope  for helpful discussions! I hope I didn't mess it up with my understanding.

### :memo: Checklist

### :hourglass: Dependencies

- #37377 : arithmetic error causes two tests to fail.

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1364,7 +1364,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
           This is only used if ``l`` is None.
 
         - ``cyclic_after`` -- (default: ``None``) an isogeny with codomain equal to
-          `self`. If set, then only isogenies that are cyclic continuations of
+          ``self``. If set, then only isogenies that are cyclic continuations of
           the given isogeny are returned.
 
         OUTPUT:
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1363,7 +1363,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
         - ``max_l`` -- (default: 31) a bound on the primes to be tested.
           This is only used if ``l`` is None.
 
-        - ``cyclic_after`` -- (default: None) an isogeny with codomain equal to
+        - ``cyclic_after`` -- (default: ``None``) an isogeny with codomain equal to
           `self`. If set, then only isogenies that are cyclic continuations of
           the given isogeny are returned.
 
@@ -1533,8 +1533,8 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             sage: E.isogenies_prime_degree(5)
             []
 
-        The optional argument `cyclic_after` can be used to easily construct cyclic
-        isogeny chains (i.e., avoid hitting dual isogenies):
+        The optional argument ``cyclic_after`` can be used to easily construct cyclic
+        isogeny chains (i.e., avoid hitting dual isogenies)::
 
             sage: E = EllipticCurve(GF(13^6), [2,8])
             sage: phi1 = choice(E.isogenies_prime_degree(3))
@@ -1627,7 +1627,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             ...
             ValueError: 4 is not prime.
 
-        Test for bug in PR#37388::
+        Test for bug in :issue:`37388`::
 
             sage: E0 = EllipticCurve(GF(19), [4, 0])
             sage: phi1 = E0.isogenies_prime_degree(5)[0]
@@ -1662,9 +1662,9 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
         else:
             dom = cyclic_after.domain()
             if cyclic_after.codomain() != self:
-                raise ValueError("The codomain of `cyclic_after` must be the same curve")
+                raise ValueError("the codomain of `cyclic_after` must be the same curve")
             if not cyclic_after.is_separable():
-                raise ValueError("The `cyclic_after` isogeny must be separable")
+                raise ValueError("the `cyclic_after` isogeny must be separable")
             ret = []
             for d in L:
                 for phi in isogenies_prime_degree(self, d):
@@ -1672,7 +1672,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
                     # but also do some overhead (construct composte map, check degrees
                     # are primes, etc.)
                     # So we check j-invariants here to make most cases pass fast
-                    if (cyclic_after.degree().is_prime() and phi.codomain().j_invariant() != dom.j_invariant()):
+                    if cyclic_after.degree().is_prime() and phi.codomain().j_invariant() != dom.j_invariant():
                         ret.append(phi)
                     elif (phi * cyclic_after).is_cyclic():
                         ret.append(phi)
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1658,7 +1658,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
 
         from .isogeny_small_degree import isogenies_prime_degree
         if cyclic_after is None:
-            return sum([isogenies_prime_degree(self, d) for d in L], [])
+            return [phi for d in L for phi in isogenies_prime_degree(self, d)]
 
         dom = cyclic_after.domain()
         if cyclic_after.codomain() != self:
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1659,24 +1659,28 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
         from .isogeny_small_degree import isogenies_prime_degree
         if cyclic_after is None:
             return sum([isogenies_prime_degree(self, d) for d in L], [])
-        else:
-            dom = cyclic_after.domain()
-            if cyclic_after.codomain() != self:
-                raise ValueError("the codomain of `cyclic_after` must be the same curve")
-            if not cyclic_after.is_separable():
-                raise ValueError("the `cyclic_after` isogeny must be separable")
-            ret = []
-            for d in L:
-                for phi in isogenies_prime_degree(self, d):
-                    # .is_cyclic() would only check j-invariants in most cases
-                    # but also do some overhead (construct composte map, check degrees
-                    # are primes, etc.)
-                    # So we check j-invariants here to make most cases pass fast
-                    if cyclic_after.degree().is_prime() and phi.codomain().j_invariant() != dom.j_invariant():
-                        ret.append(phi)
-                    elif (phi * cyclic_after).is_cyclic():
-                        ret.append(phi)
-            return ret
+
+        dom = cyclic_after.domain()
+        if cyclic_after.codomain() != self:
+            raise ValueError("the codomain of `cyclic_after` must be the same curve")
+        if not cyclic_after.is_separable():
+            raise ValueError("the `cyclic_after` isogeny must be separable")
+
+        prev_deg = cyclic_after.degree()
+        ret = []
+        for d in L:
+            for phi in isogenies_prime_degree(self, d):
+                # .is_cyclic() would only check j-invariants in most cases
+                # but also do some overhead (construct composte map, check degrees
+                # are primes, etc.)
+                # So we check j-invariants here to make most cases pass fast
+                # (also no need to check if degrees are coprime)
+                if prev_deg.is_prime() and \
+                   (prev_deg % d or phi.codomain().j_invariant() != dom.j_invariant()):
+                    ret.append(phi)
+                elif (phi * cyclic_after).is_cyclic():
+                    ret.append(phi)
+        return ret
 
     def is_isogenous(self, other, field=None):
         """
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1626,6 +1626,16 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             Traceback (most recent call last):
             ...
             ValueError: 4 is not prime.
+
+        Test for bug in PR#37388::
+
+            sage: E0 = EllipticCurve(GF(19), [4, 0])
+            sage: phi1 = E0.isogenies_prime_degree(5)[0]
+            sage: phi2 = phi1.codomain().isogenies_prime_degree(2)[0]
+            sage: phi12 = phi2 * phi1
+            sage: phi3_list = phi12.codomain().isogenies_prime_degree(2, cyclic_after=phi12)
+            sage: [(phi3 * phi12).is_cyclic() for phi3 in phi3_list]
+            [True, True]
         """
         F = self.base_ring()
         if isinstance(F, sage.rings.abc.RealField):
@@ -1653,6 +1663,8 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             dom = cyclic_after.domain()
             if cyclic_after.codomain() != self:
                 raise ValueError("The codomain of `cyclic_after` must be the same curve")
+            if not cyclic_after.is_separable():
+                raise ValueError("The `cyclic_after` isogeny must be separable")
             ret = []
             for d in L:
                 for phi in isogenies_prime_degree(self, d):
@@ -1660,7 +1672,9 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
                     # but also do some overhead (construct composte map, check degrees
                     # are primes, etc.)
                     # So we check j-invariants here to make most cases pass fast
-                    if phi.codomain().j_invariant() != dom.j_invariant() or (phi * cyclic_after).is_cyclic():
+                    if (cyclic_after.degree().is_prime() and phi.codomain().j_invariant() != dom.j_invariant()):
+                        ret.append(phi)
+                    elif (phi * cyclic_after).is_cyclic():
                         ret.append(phi)
             return ret
 
--- a/src/sage/schemes/elliptic_curves/ell_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_field.py
@@ -1350,7 +1350,7 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             E._fetch_cached_order(self)
         return E
 
-    def isogenies_prime_degree(self, l=None, max_l=31):
+    def isogenies_prime_degree(self, l=None, max_l=31, cyclic_after=None):
         """
         Return a list of all separable isogenies of given prime degree(s)
         with domain equal to ``self``, which are defined over the base
@@ -1362,6 +1362,10 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
 
         - ``max_l`` -- (default: 31) a bound on the primes to be tested.
           This is only used if ``l`` is None.
+
+        - ``cyclic_after`` -- (default: None) an isogeny with codomain equal to
+          `self`. If set, then only isogenies that are cyclic continuations of
+          the given isogeny are returned.
 
         OUTPUT:
 
@@ -1529,6 +1533,15 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
             sage: E.isogenies_prime_degree(5)
             []
 
+        The optional argument `cyclic_after` can be used to easily construct cyclic
+        isogeny chains (i.e., avoid hitting dual isogenies):
+
+            sage: E = EllipticCurve(GF(13^6), [2,8])
+            sage: phi1 = choice(E.isogenies_prime_degree(3))
+            sage: phi2_list = phi1.codomain().isogenies_prime_degree(3, cyclic_after=phi1)
+            sage: [(phi2 * phi1).is_cyclic() for phi2 in phi2_list]
+            [True, True, True]
+
         An example over a rational function field::
 
             sage: R.<t> = GF(5)[]
@@ -1634,7 +1647,22 @@ class EllipticCurve_field(ell_generic.EllipticCurve_generic, ProjectivePlaneCurv
                 L = [ZZ(d) for d in l]
 
         from .isogeny_small_degree import isogenies_prime_degree
-        return sum([isogenies_prime_degree(self, d) for d in L], [])
+        if cyclic_after is None:
+            return sum([isogenies_prime_degree(self, d) for d in L], [])
+        else:
+            dom = cyclic_after.domain()
+            if cyclic_after.codomain() != self:
+                raise ValueError("The codomain of `cyclic_after` must be the same curve")
+            ret = []
+            for d in L:
+                for phi in isogenies_prime_degree(self, d):
+                    # .is_cyclic() would only check j-invariants in most cases
+                    # but also do some overhead (construct composte map, check degrees
+                    # are primes, etc.)
+                    # So we check j-invariants here to make most cases pass fast
+                    if phi.codomain().j_invariant() != dom.j_invariant() or (phi * cyclic_after).is_cyclic():
+                        ret.append(phi)
+            return ret
 
     def is_isogenous(self, other, field=None):
         """
--- a/src/sage/schemes/elliptic_curves/hom.py
+++ b/src/sage/schemes/elliptic_curves/hom.py
@@ -924,6 +924,75 @@ class EllipticCurveHom(Morphism):
         """
         return not self.degree()
 
+    def is_cyclic(self):
+        r"""
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
+
+        This is a general method, it requires factorization of the degree
+        and computation on the kernel polynomial (if not squarefree).
+        For special isogeny implementations (Frobenius, scalar multiplication,
+        composite) more efficient methods are used.
+
+        ALGORITHM:
+
+        Separable isogenies with squarefree degrees are cyclic.
+        For each $\ell \ne p$ such that its square divides the degree of the isogeny,
+        test if $E[\ell]$ is mapped to 0 by reducing the kernel polynomial
+        modulo the division polynomial. If for all $\ell$ it does not happen,
+        then the result is cyclic.
+
+        EXAMPLES:
+
+        Here we include also special cases which do not use the general algorithm. ::
+
+            sage: E = EllipticCurve(GF(79), [7,7])
+            sage: phi = E.scalar_multiplication(5)
+            sage: phi.is_cyclic()
+            False
+            sage: phi = E.scalar_multiplication(1)
+            sage: phi.is_cyclic()
+            True
+            sage: phi = E.isogenies_prime_degree(2)[0]
+            sage: phi.is_cyclic()
+            True
+            sage: phi.dual().is_cyclic()
+            True
+            sage: (phi.dual() * phi).is_cyclic()
+            False
+
+            sage: phi = E.isogenies_prime_degree(79)[0]
+            sage: phi.is_cyclic()
+            True
+            sage: (phi.dual() * phi).is_cyclic()
+            False
+
+        .. SEEALSO::
+
+        - :meth:`sage.schemes.elliptic_curves.hom_composie.EllipticCurveHom_composite.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_frobenius.EllipticCurveHom_frobenius.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_scalar.EllipticCurveHom_scalar.is_cyclic`
+        """
+        if self.is_zero() or not self.is_separable():
+            return False
+
+        deg_fac = ZZ(self.degree()).factor()
+        # squarefree => cyclic
+        if not deg_fac or max(e for ell, e in deg_fac) <= 1:
+            return True
+
+        E0 = self.domain()
+        ker_poly = self.kernel_polynomial()
+        for ell, e in deg_fac:
+            # check only prime factors with e >= 2 and not the characteristic
+            # since E[p] is always cyclic, no need to check
+            # (inseparability is already checked)
+            if e <= 1 or ell == self.base_ring().characteristic():
+                continue
+            div_poly = E0.division_polynomial(ell)
+            if ker_poly % div_poly == 0:
+                return False
+        return True
+
     def __neg__(self):
         r"""
         Return the negative of this elliptic-curve morphism. In other
--- a/src/sage/schemes/elliptic_curves/hom.py
+++ b/src/sage/schemes/elliptic_curves/hom.py
@@ -936,14 +936,12 @@ class EllipticCurveHom(Morphism):
         ALGORITHM:
 
         Separable isogenies with squarefree degrees are cyclic.
-        For each $\ell \ne p$ such that its square divides the degree of the isogeny,
-        test if $E[\ell]$ is mapped to 0 by reducing the kernel polynomial
-        modulo the division polynomial. If for all $\ell$ it does not happen,
+        For each `\ell \ne p` such that its square divides the degree of the isogeny,
+        test if `E[\ell]` is mapped to `0` by reducing the kernel polynomial
+        modulo the division polynomial. If for all `\ell` it does not happen,
         then the result is cyclic.
 
-        EXAMPLES:
-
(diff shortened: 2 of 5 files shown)


src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1363,6 +1363,10 @@ def isogenies_prime_degree(self, l=None, max_l=31):
         - ``max_l`` -- (default: 31) a bound on the primes to be tested.
           This is only used if ``l`` is None.
 
+        - ``cyclic_after`` -- (default: None) an isogeny with codomain equal to
```
grhkm21:
```suggestion
        - ``cyclic_after`` -- (default: ``None``) an isogeny with codomain equal to
```
I am not sure if `cyclic_after` is a good name, but I can't think of a better one.

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1529,6 +1533,15 @@ def isogenies_prime_degree(self, l=None, max_l=31):
             sage: E.isogenies_prime_degree(5)
             []
 
+        The optional argument `cyclic_after` can be used to easily construct cyclic
```
grhkm21:
```suggestion
        The optional argument ``cyclic_after`` can be used to easily construct cyclic
```
In docstrings, single backticks means $\LaTeX$ code, double backticks means code.

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1529,6 +1533,15 @@ def isogenies_prime_degree(self, l=None, max_l=31):
...
+        isogeny chains (i.e., avoid hitting dual isogenies):
```
grhkm21:
```suggestion
        isogeny chains (i.e., avoid hitting dual isogenies)::
```

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1613,6 +1626,16 @@ def isogenies_prime_degree(self, l=None, max_l=31):
             Traceback (most recent call last):
             ...
             ValueError: 4 is not prime.
+
+        Test for bug in PR#37388::
```
grhkm21:
```suggestion
        Test for bug in :issue:`37388`::
```

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1634,7 +1657,26 @@ def isogenies_prime_degree(self, l=None, max_l=31):
...
         from .isogeny_small_degree import isogenies_prime_degree
-        return sum([isogenies_prime_degree(self, d) for d in L], [])
+        if cyclic_after is None:
+            return sum([isogenies_prime_degree(self, d) for d in L], [])
+        else:
+            dom = cyclic_after.domain()
+            if cyclic_after.codomain() != self:
+                raise ValueError("The codomain of `cyclic_after` must be the same curve")
+            if not cyclic_after.is_separable():
+                raise ValueError("The `cyclic_after` isogeny must be separable")
```
grhkm21:
```suggestion
                raise ValueError("the codomain of `cyclic_after` must be the same curve")
            if not cyclic_after.is_separable():
                raise ValueError("the `cyclic_after` isogeny must be separable")
```
Error messages should start with lowercase

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1634,7 +1657,26 @@ def isogenies_prime_degree(self, l=None, max_l=31):
...
+            ret = []
+            for d in L:
+                for phi in isogenies_prime_degree(self, d):
```
grhkm21:
```suggestion
                for phi in self.isogenies_prime_degree(d):
```

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1634,7 +1657,26 @@ def isogenies_prime_degree(self, l=None, max_l=31):
...
+                    # .is_cyclic() would only check j-invariants in most cases
+                    # but also do some overhead (construct composte map, check degrees
+                    # are primes, etc.)
+                    # So we check j-invariants here to make most cases pass fast
+                    if (cyclic_after.degree().is_prime() and phi.codomain().j_invariant() != dom.j_invariant()):
```
grhkm21:
```suggestion
                    if cyclic_after.degree().is_prime() and phi.codomain().j_invariant() != dom.j_invariant():
```

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,102 @@ def is_zero(self):
...
+        and computation on the kernel polynomial (if not squarefree).
+        For special isogeny implementations (Frobenius, scalar multiplication,
+        composite) more efficient methods are used.
+
+        ALGORITHM:
+
+        Separable isogenies with squarefree degrees are cyclic.
+        For each $\ell \ne p$ such that its square divides the degree of the isogeny,
+        test if $E[\ell]$ is mapped to 0 by reducing the kernel polynomial
+        modulo the division polynomial. If for all $\ell$ it does not happen,
```
grhkm21:
```suggestion
        For each `\ell \ne p` such that its square divides the degree of the isogeny,
        test if `E[\ell]` is mapped to `0` by reducing the kernel polynomial
        modulo the division polynomial. If for all `\ell` it does not happen,
```

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,102 @@ def is_zero(self):
...
+        then the result is cyclic.
+
+        EXAMPLES:
+
+        Here we include also special cases which do not use the general algorithm. ::
```
grhkm21:
```suggestion
        EXAMPLES::
```
I think just showing the examples is fine

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,102 @@ def is_zero(self):
...
+        - :meth:`sage.schemes.elliptic_curves.hom_composie.EllipticCurveHom_composite.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_frobenius.EllipticCurveHom_frobenius.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_scalar.EllipticCurveHom_scalar.is_cyclic`
+        """
+        if self.is_zero() or not self.is_separable():
+            return False
+
+        deg_fac = ZZ(self.degree()).factor()
+        # squarefree => cyclic
+        if not deg_fac or max(e for ell, e in deg_fac) <= 1:
```
grhkm21:
```suggestion
        if not deg_fac or all(e <= 1 for ell, e in deg_fac):
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
         return prod(phi.inseparable_degree() for phi in self._phis)
+
+    def is_cyclic(self, _check_j=True, _reduce_kernel=True):
+        r"""
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
+
+        ALGORITHM:
+
+        For each prime $\ell$ with $\ell^2$ dividing the isogeny degree,
+        test that $\ell$-torsion is not fully nullified. This is done by
```
grhkm21:
```suggestion
        For each prime `\ell` with `\ell^2` dividing the isogeny degree,
        test that `\ell`-torsion is not fully nullified. This is done by
```
There are a few more in this docstring, I won't comment on them.

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+
+        If `_check_j` is set, then the segment is further shrinked by eliminating
+        prefix/suffix chains of $\ell$-isogenies via j-invariant checks. In many
+        applications, the isogeny degrees would be ordered, so that only
+        j-invariant checks would be performed (unless there is a "backtracking"
+        j-invariant step, which is not necessarily the dual).
+
+        If `_reduce_kernel` is set, then, after the first step in the segment,
+        the modulus is replaced by the kernel polynomial of the dual isogeny,
+        which has degree (l-1)/2 instead of (l^2-1)/2.
```
grhkm21:
```suggestion
        INPUT:

        - ``_check_j`` -- (bool, default: ``True``); if set, the segment is further shrinked by eliminating
          prefix/suffix chains of `\ell`-isogenies via j-invariant checks. In many
          applications, the isogeny degrees would be ordered, so that only
          j-invariant checks would be performed (unless there is a "backtracking"
          j-invariant step, which is not necessarily the dual). For internal use only.

        - ``_reduce_kernel`` -- (bool, default: ``True``); if set, after the first step in the segment,
          the modulus is replaced by the kernel polynomial of the dual isogeny,
          which has degree `(\ell-1)/2` instead of `(\ell^2-1)/2`. For internal use only.
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+            sage: phi1 = E0.isogenies_prime_degree(2)[0]
+            sage: (phi1.dual() * phi1).is_cyclic()
+            False
+
+            sage: phi2_list = phi1.codomain().isogenies_prime_degree(2)
+            sage: sorted((phi2 * phi1).is_cyclic() for phi2 in phi2_list)
+            [False, True, True]
+
+        Works even if the cyclicity is broken across more than 2 isogenies:
+        (e.g., when commuting isogenies are swapped: degrees 2-2-5 <-> 2-5-2)
```
grhkm21:
```suggestion
        The method works even if the cyclicity is broken across more than 2 isogenies
        such as when the order of commuting isogenies are swapped from 2-2-5 to 2-5-2::
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+
+            sage: phi2 = phi1.codomain().isogenies_prime_degree(5)[0]
+            sage: phi3_list = phi2.codomain().isogenies_prime_degree(2)
+            sage: sorted((phi3 * phi2 * phi1).is_cyclic() for phi3 in phi3_list)
+            [False, True, True]
+
+        "Recursive" composite isogenies are not flattened, but should still
+        return correct results:
```
grhkm21:
```suggestion
        Endomorphisms as composite isogenies are not flattened but still returns correct results::
```
I think this is what you mean.

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+            sage: EllipticCurveHom_composite.from_factors([E0.scalar_multiplication(37)]).is_cyclic()
+            False
+            sage: EllipticCurveHom_composite.from_factors([EllipticCurveHom_frobenius(E0, 0)]).is_cyclic()
+            True
+            sage: EllipticCurveHom_composite.from_factors([EllipticCurveHom_frobenius(E0, 1)]).is_cyclic()
+            False
+
+        This test involves multiple "backtracking" edges which are not duals. We have only 2 supersingular
+        j-invariants, and so all supersingular isogenies have to be between them. In particular,
+        there are $7^2$-isogenies on j-invariants 7-7-7 which is cyclic.
```
grhkm21:
```suggestion
        there are `7^2`-isogenies on j-invariants 7-7-7 which is cyclic::
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+    Sieve $\ell$-torsion through isogenies.
+
+    Helper for :meth:`EllipticCurveHom_composite.is_cyclic`.
+    The torsion is represented by the polynomial with
+    x-coordinates as roots (a divisor of the division polynomial
+    of the starting curve).
+
+    INPUT:
+
+    - phis (list of isogenies) -- list of isogenies to propagate the torsion through
```
grhkm21:
```suggestion
    - ``phis`` -- a list of isogenies to propagate the torsion through.
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+
+    - ell -- an integer defining the torsion $E[\ell]$
```
grhkm21:
```suggestion
    - ``ell`` -- an integer defining the torsion `E[\ell]`.
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+
+    - use_second_curve (optional, default: ``True``): whether to base
```
grhkm21:
```suggestion
    - ``use_second_curve`` (optional, default: ``True``): whether to base
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+        x^2 + 4
+        sage: _sieve_torsion_through_isogenies([phi], 3).degree()
+        4
+
+    .. SEEALSO::
+
+        - :meth:`sage.schemes.elliptic_curves.hom_composit.EllipticCurveHom_composite.is_cyclic`
+    """
+    if not phis:
+        raise ValueError("List of isogenies must be nonempty")
```
grhkm21:
```suggestion
        raise ValueError("list of isogenies must be nonempty")
```
(Change the corresponding doctests)

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,291 @@ def inseparable_degree(self):
...
+    if any(phi.is_zero() or not phi.is_separable() for phi in phis):
+        return phis[0].domain().base_ring()['x'](1)
```
grhkm21:
```suggestion
        return phis[0].domain().base_ring()['x'].one()
```
(If I understand correctly) this is faster

src/sage/schemes/elliptic_curves/hom_frobenius.py
```diff
@@ -519,3 +519,30 @@ def inseparable_degree(self):
             True
         """
         return self._degree
+
+    def is_cyclic(self):
+        """
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
+
+        Since this class implements only purely inseparable isogenies,
+        they are only cyclic in the trivial case (degree = p^0 = 1).
```
grhkm21:
```suggestion
        they are only cyclic in the trivial case i.e. when degree is `1`.
```

src/sage/schemes/elliptic_curves/hom_scalar.py
```diff
@@ -514,3 +514,29 @@ def __neg__(self):
             w = negation_morphism(self._domain).rational_maps()
             result._rational_maps = tuple(f(*w) if f is not None else None for f in self._rational_maps)
         return result
+
+    def is_cyclic(self):
+        """
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
+
+        Since this class implements only scalar multiplication isogenies,
+        they are never cyclic, except the trivial [1].
```
grhkm21:
```suggestion
        they are never cyclic except for the trivial `[1]`.
```

grhkm21 (changes requested):
Thank you so much for the contribution! I'm surprised this is your first.

I gave some reviews on the docstrings. I have **not** looked at any of the code or algorithms, so another reviewer would be appreciated.

The main changes are that:

- Single backticks means $\LaTeX$ code, and double backticks means `code`. 
- Error message should start with lower case (and not end with full stop, which you did correctly)
- The comment of each EXAMPLE/TEST block ends with `::` to indicate a test block
- The INPUT segment looks like this
```
INPUT:

- ``a`` -- (integer) mysterious number
```

src/sage/schemes/elliptic_curves/hom.py
hellman: Then the `not deg_fac` can be removed I think

src/sage/schemes/elliptic_curves/hom_composite.py
hellman: Hmm, no, I meant EllipticCurveHom_composite where one or more internal isogenies are also a EllipticCurveHom_composite instances.

src/sage/schemes/elliptic_curves/ell_field.py
hellman: This would call the current method recursively, I don't see a reason for this. Calling the function `isogenies_prime_degree` directly just mimicks the code from L1661.

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1363,6 +1363,10 @@ def isogenies_prime_degree(self, l=None, max_l=31):
         - ``max_l`` -- (default: 31) a bound on the primes to be tested.
           This is only used if ``l`` is None.
 
+        - ``cyclic_after`` -- (default: ``None``) an isogeny with codomain equal to
+          `self`. If set, then only isogenies that are cyclic continuations of
+          the given isogeny are returned.
+
         OUTPUT:
 
         (list) All separable `l`-isogenies for the given `l` with domain self.
```
GiacomoPope, line 1372: Tiny comment, you could do `\ell` rather than `l` here -- this means using raw strings (r""" instead of """ for the comment at the top), but I can see how using `l` matches with the argument `l` :D

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1634,7 +1657,26 @@ def isogenies_prime_degree(self, l=None, max_l=31):
                 L = [ZZ(d) for d in l]
 
         from .isogeny_small_degree import isogenies_prime_degree
-        return sum([isogenies_prime_degree(self, d) for d in L], [])
+        if cyclic_after is None:
+            return sum([isogenies_prime_degree(self, d) for d in L], [])
+        else:
```
GiacomoPope: I guess you could remove the else here and just remove one level of indentation as the if above ends with a return so can be isolated.

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,100 @@ def is_zero(self):
...
+        .. SEEALSO::
+
+        - :meth:`sage.schemes.elliptic_curves.hom_composie.EllipticCurveHom_composite.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_frobenius.EllipticCurveHom_frobenius.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_scalar.EllipticCurveHom_scalar.is_cyclic`
+        """
+        if self.is_zero() or not self.is_separable():
+            return False
+
+        deg_fac = ZZ(self.degree()).factor()
```
GiacomoPope: I have a feeling `self.degree()` returns something of type `Integer` anyway?

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +896,290 @@ def inseparable_degree(self):
...
+            return False
+
+        # this code is 40x slower.. even on degrees [2, 2, 2, 2, 2]
+        # even without separate prime test
+        # deg_fac = Factorization(())
+        # for phi in self._phis:
+        #     deg_fac *= ZZ(phi.degree()).factor()
+        deg_fac = {}
+        for phi in self._phis:
+            phi_deg = ZZ(phi.degree())
```
GiacomoPope: I guess same comment here, the degree I think will already be an integer.

GiacomoPope (commented):
Just a few comments, will look more closely at maths tomorrow.

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,100 @@ def is_zero(self):
...
+            sage: E0.scalar_multiplication(37).is_cyclic()
+            False
+            sage: EllipticCurveHom_frobenius(E0, 0).is_cyclic()
+            True
+            sage: EllipticCurveHom_frobenius(E0, 1).is_cyclic()
+            False
+
+        .. SEEALSO::
+
+        - :meth:`sage.schemes.elliptic_curves.hom_composie.EllipticCurveHom_composite.is_cyclic`
```
yyyyx4, line 996:
```suggestion
        - :meth:`sage.schemes.elliptic_curves.hom_composite.EllipticCurveHom_composite.is_cyclic`
```

src/sage/schemes/elliptic_curves/ell_field.py
grhkm21: Oh wait, I see, sorry for that. I understand the logic now

src/sage/schemes/elliptic_curves/hom.py
grhkm21: Think so

src/sage/schemes/elliptic_curves/hom_composite.py
grhkm21: Ahh okay. I think “recursive” is still not the correct word, or it should at least be defined somewhere somewhat formally. Also the double colon at the end :D

yyyyx4: How about "nested"?

src/sage/schemes/elliptic_curves/ell_field.py
hellman: Hmm, it's not a part of this PR and other methods also use `l`, I guess to match the argument indeed.

src/sage/schemes/elliptic_curves/ell_field.py
hellman: Done!

src/sage/schemes/elliptic_curves/ell_field.py
```diff
@@ -1634,7 +1657,30 @@ def isogenies_prime_degree(self, l=None, max_l=31):
                 L = [ZZ(d) for d in l]
 
         from .isogeny_small_degree import isogenies_prime_degree
-        return sum([isogenies_prime_degree(self, d) for d in L], [])
+        if cyclic_after is None:
+            return sum([isogenies_prime_degree(self, d) for d in L], [])
```
yyyyx4:
```suggestion
            return [phi for d in L for phi in isogenies_prime_degree(self, d)]
```

mkoeppe:
needs merge/rebase

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,100 @@ def is_zero(self):
...
+        - :meth:`sage.schemes.elliptic_curves.hom_frobenius.EllipticCurveHom_frobenius.is_cyclic`
+        - :meth:`sage.schemes.elliptic_curves.hom_scalar.EllipticCurveHom_scalar.is_cyclic`
```
yyyyx4, line 998:
```suggestion
        .. SEEALSO::

            - :meth:`sage.schemes.elliptic_curves.hom_composie.EllipticCurveHom_composite.is_cyclic`
            - :meth:`sage.schemes.elliptic_curves.hom_frobenius.EllipticCurveHom_frobenius.is_cyclic`
            - :meth:`sage.schemes.elliptic_curves.hom_scalar.EllipticCurveHom_scalar.is_cyclic`
```

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,100 @@ def is_zero(self):
...
+        """
+        if self.is_zero() or not self.is_separable():
+            return False
```
yyyyx4, line 1001: Not sure if this should be `ValueError` instead: The definition of cyclicity is a bit ambiguous for morphisms that aren't separable isogenies. Hence, if we get called with inseparable isogenies or zero, it's fairly likely to be a mistake of some sort.

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,100 @@ def is_zero(self):
...
+
+        deg_fac = self.degree().factor()
+        # squarefree => cyclic
+        if all(e <= 1 for ell, e in deg_fac) <= 1:
```
yyyyx4, line 1005:
```suggestion
        if all(e <= 1 for ell, e in deg_fac):
```

src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -924,6 +924,100 @@ def is_zero(self):
...
+        E0 = self.domain()
+        ker_poly = self.kernel_polynomial()
+        for ell, e in deg_fac:
+            # check only prime factors with e >= 2 and not the characteristic
+            # since E[p] is always cyclic, no need to check
+            # (inseparability is already checked)
+            if e <= 1 or ell == self.base_ring().characteristic():
+                continue
+            div_poly = E0.division_polynomial(ell)
+            if ker_poly % div_poly == 0:
```
yyyyx4, line 1017:
```suggestion
            if div_poly.divides(ker_poly):
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
             1331
         """
         return prod(phi.inseparable_degree() for phi in self._phis)
+
+    def is_cyclic(self, _check_j=True, _reduce_kernel=True):
```
yyyyx4, line 898:
```suggestion
    def is_cyclic(self, *, _check_j=True, _reduce_kernel=True):
```
(To avoid people accidentally using these internal arguments.)

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
+
+        ALGORITHM:
+
+        For each prime `\ell` with `\ell^2` dividing the isogeny degree,
+        test that `\ell`-torsion is not fully nullified. This is done by
+        evaluating the isogeny modulo the division polynomial of the domain curve,
+        reducing the modulus on the way when parts of the torsion get nullified.
+
+        For each $\ell$, only the smallest segment of isogenies is considered,
```
yyyyx4, line 909:
```suggestion
        For each `\ell`, only the smallest segment of isogenies is considered,
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+            sage: from sage.schemes.elliptic_curves.hom_composite import EllipticCurveHom_composite
+            sage: EllipticCurveHom_composite.from_factors([phi1.dual() * phi1]).is_cyclic()
+            False
+            sage: phi2 = choice([phi2 for phi2 in phi2_list if (phi2 * phi1).is_cyclic()])
+            sage: EllipticCurveHom_composite.from_factors([phi2 * phi1]).is_cyclic()
+            True
+
+        .. SEEALSO::
+
+        - :meth:`sage.schemes.elliptic_curves.hom.EllipticCurveHom.is_cyclic`
```
yyyyx4, line 988:
```suggestion
            - :meth:`sage.schemes.elliptic_curves.hom.EllipticCurveHom.is_cyclic`
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+        """
+        if self.is_zero() or not self.is_separable():
+            return False
```
yyyyx4, line 991: Same here: I'd probably go for `ValueError` in this case.

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+                continue
+
+            # we only need to check compositions of isogenies
+            # with degree divisible by ell
+            # so we restrict ourselves to the smallest segment
+            # containing all occurrences of ell in the degree
+            first = None
+            last = None
+            for i, phi in enumerate(self._phis):
+                if phi.degree() % ell == 0:
```
yyyyx4, line 1025:
```suggestion
                if ell.divides(phi.degree()):
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+                    self._phis[first:last+1], ell, use_second_curve=_reduce_kernel)
+                if survived.degree() <= 0:
+                    # fully destroyed, non-cyclic
+                    return False
+        return True
+
+
+def _sieve_torsion_through_isogenies(phis, ell, use_second_curve=True):
+    r"""
+    Sieve $\ell$-torsion through isogenies.
```
yyyyx4, line 1076:
```suggestion
    Sieve `\ell`-torsion through isogenies.
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+        sage: _sieve_torsion_through_isogenies([phi], 2, use_second_curve=False).monic()
+        x^2 + 4
+        sage: div = E0.division_polynomial(2); (div // phi.kernel_polynomial()).monic()
+        x^2 + 4
+        sage: _sieve_torsion_through_isogenies([phi], 3).degree()
+        4
+
+    .. SEEALSO::
+
+        - :meth:`sage.schemes.elliptic_curves.hom_composit.EllipticCurveHom_composite.is_cyclic`
```
yyyyx4, line 1127:
```suggestion
        - :meth:`sage.schemes.elliptic_curves.hom_composite.EllipticCurveHom_composite.is_cyclic`
```

src/sage/schemes/elliptic_curves/hom_composite.py
```diff
@@ -893,3 +894,290 @@ def inseparable_degree(self):
...
+        x_rational_map = phi.x_rational_map()
+
+        den = x_rational_map.denominator()(y)
+        g = den.lift().gcd(mod)
+        if g != 1:
+            # a new part of torsion is nullified
+            # so we need to update the modulus and the quotient ring
+            mod //= g
+            assert mod
+            if mod.degree() <= 0:  # just 1? or -1?
```
yyyyx4, line 1163:
```suggestion
            if mod.is_constant():
```
(Not sure what the comment means.)

src/sage/schemes/elliptic_curves/hom_frobenius.py
```diff
@@ -519,3 +519,30 @@ def inseparable_degree(self):
             True
         """
         return self._degree
+
+    def is_cyclic(self):
+        """
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
```
yyyyx4, line 525: (This should be consistent with the definition elsewhere: In my world "cyclic" is either only defined for separable isogenies to begin with, or it depends only on the separable part of an isogeny, but that's up for discussion.)

src/sage/schemes/elliptic_curves/hom_scalar.py
```diff
@@ -514,3 +514,29 @@ def __neg__(self):
             w = negation_morphism(self._domain).rational_maps()
             result._rational_maps = tuple(f(*w) if f is not None else None for f in self._rational_maps)
         return result
+
+    def is_cyclic(self):
+        """
+        Determine whether the isogeny is cyclic (separable with cyclic kernel).
```
yyyyx4, line 520: Same here.

yyyyx4 (changes requested):
Needs rebase, and some smaller details need fixing, but this looks pretty good overall.

grhkm21:
bump!

yyyyx4:
> bump!

`+1` :slightly_smiling_face:

PR #37389, opened 2024-02-18
Implement a more general method for computing dual isogenies of elliptic curves

hellman:
Upgrade the `.dual()` method of separable elliptic curve isogenies for a more generic one (the current one is very restricted).

Current implementation of isogeny duals is based on computing some kind of normalized domain/codomain curves (isomorphic to the original codomain/domain curves respectively), and then computing an isogeny between these curves generically:

```py
phi_hat = EllipticCurveIsogeny(E1, None, E2, d)
```

This method has limitations as its not implemented for many cases (in particular, currently it requires 4d+4 < p or something like that). Also, I am not sure that, even if implemented, the normalization performed ensures a unique isogeny. In general there can be multiple isogenies between E1 and E2 but only one of them is the dual. I don't know how this can be correct.

Here is an alternative and a more direct, algebraic approach. Compute the rational x-coordinate isogeny map modulo the division polynomial on the source curve divided by its gcd with the map's denominator (i.e., excluding the isogeny kernel from the full torsion encoded by the division polynomial). This reduces the modulus degree from $\ell^2-1$ to $\ell^2-1-\ell$: we still have cosets of the remaining torsion subgroup, that's why it is still large. But now we can check the *image* of this set, which should be exactly the kernel of the dual isogeny. For this, we simply compute the minimal polynomial of the x-coordinate-map in the aforementioned quotient ring.

In fact, almost exactly the same method is already used for the inseparable part of the dual, in the same `dual()` method.

I am not sure if there are edge cases for this method and I do not fully understand the normalization process, but it seems to work well (passes all existing tests + new ones that did not work before), and is at least worth to explore.

### :memo: Checklist

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_curve_isogeny.py
+++ b/src/sage/schemes/elliptic_curves/ell_curve_isogeny.py
@@ -3050,37 +3050,6 @@ class EllipticCurveIsogeny(EllipticCurveHom):
 
         self.__set_post_isomorphism(codomain, isom)
 
-    def decompose_inseparable(self):
-        """
-        Decompose the isogeny into purely inseparable part and a separable part.
-
-        OUTPUT: a pair of a purely inseparable (or identity) isogeny and a separable isogeny,
-        which equal to the current one when composed as (separable * purely inseparable).
-        """
-        F = self.__base_field
-        p = F.characteristic()
-        frob_k = self._degree.valuation(p)
-        deg = self._degree // p**frob_k
-        # degree = deg * p**k
-
-        from sage.schemes.elliptic_curves.hom_frobenius import EllipticCurveHom_frobenius
-        frob = EllipticCurveHom_frobenius(self._codomain, frob_k)
-
-        f = self.kernel_polynomial()
-
-        psi = self._domain.division_polynomial(p)
-        mu_num = self._domain._multiple_x_numerator(p)
-        mu_den = self._domain._multiple_x_denominator(p)
-
-        for _ in range(frob_k):
-            f //= f.gcd(psi)
-            S = f.parent().quotient_ring(f)
-            mu = S(mu_num) / S(mu_den)
-            f = mu.minpoly()
-
-        sep = self._domain.isogeny(f, codomain=frob.codomain())
-        return frob, sep
-
     def dual(self):
         r"""
         Return the isogeny dual to this isogeny.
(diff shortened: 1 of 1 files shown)


hellman:
Leaving as draft for now, comments or tests would be welcome!

hellman:
Rewrote scaling factor, now passes this exhaustive test for small p / small ell:

```py
for p in primes(5, 51):
    E = EllipticCurve(GF(p**2), [4, 0])
    for ell in primes(2, 23):
        if ell == p:
            continue
        print(p, ell, E.is_supersingular())
        lst = E.isogenies_prime_degree(ell)
        if not lst:
            continue
        for phi in lst:
            assert phi.dual().dual() == phi
            assert phi.dual() * phi == phi.domain().scalar_multiplication(ell), (p, ell)
            assert phi * phi.dual() == phi.codomain().scalar_multiplication(ell), (p, ell)
    print()
```

hellman:
Rewrote scaling factor, now passes this exhaustive test for small p / small ell:

```py
for p in primes(5, 51):
    E = EllipticCurve(GF(p**2), [4, 0])
    for ell in primes(2, 23):
        if ell == p:
            continue
        print(p, ell, E.is_supersingular())
        lst = E.isogenies_prime_degree(ell)
        if not lst:
            continue
        for phi in lst:
            assert phi.dual().dual() == phi
            assert phi.dual() * phi == phi.domain().scalar_multiplication(ell), (p, ell)
            assert phi * phi.dual() == phi.codomain().scalar_multiplication(ell), (p, ell)
    print()
```

hellman:
On the other hand, this seems to be slower than previous code.

hellman:
After profiling, the majority of time (~60%) is taken in the *constructor* of the quotient ring:

```py
S = psi.parent().quotient_ring(psi)
```

which apparently checks for irreducibility of the polynomial..

And ~20% is minpoly which makes sense.

hellman:
I guess we should keep both approaches and choose the one to use dynamically (depending on the characteristic and the isogeny degree). Also the current one fails for even degrees which are non-powers of 2:

```py
E = EllipticCurve(GF( (2**64+13)**2 ), [4, 0])
o = E.order()
t = 2 * 17
while True:
    P = E.random_element() * (o // 4 // t)
    if P.order() == t:
        break
E.isogeny(P).dual()
```

```
NotImplementedError                       Traceback (most recent call last)
...
NotImplementedError: Kohel's algorithm currently only supports cyclic isogenies (except for [2])
```

PR #37393, opened 2024-02-19
Tensor, exterior, and symmetric product of semigroup representations

tscrim:
This allows us to take tensor,. exterior, and symmetric products of semigroup representations.
We also provide the following additional changes/improvements to help build examples and fix bugs:

- Reflection representation of Coxeter groups.
- Latex output of the identity in Artin groups and permutation groups.
- ascii and unicode art of CFMs that do not have a `one_basis()`.

### :memo: Checklist

### DIFF
--- a/src/sage/algebras/clifford_algebra.py
+++ b/src/sage/algebras/clifford_algebra.py
@@ -51,7 +51,7 @@ class CliffordAlgebraIndices(UniqueRepresentation, Parent):
     A facade parent for the indices of Clifford algebra.
     Users should not create instances of this class directly.
     """
-    def __init__(self, Qdim):
+    def __init__(self, Qdim, degree=None):
         r"""
         Initialize ``self``.
 
@@ -67,9 +67,22 @@ class CliffordAlgebraIndices(UniqueRepresentation, Parent):
             1111
             sage: type(i)
             <class 'sage.data_structures.bitset.FrozenBitset'>
+
+            sage: idx = CliffordAlgebraIndices(7, 3)
+            sage: idx._nbits
+            7
+            sage: idx._degree
+            3
+            sage: idx._cardinality
+            35
         """
         self._nbits = Qdim
-        self._cardinality = 2 ** Qdim
+        if degree is None:
+            self._cardinality = 2 ** Qdim
+        else:
+            from sage.functions.other import binomial
+            self._cardinality = binomial(Qdim, degree)
+        self._degree = degree
         # the if statement here is in case Qdim is 0.
         category = FiniteEnumeratedSets().Facade()
         Parent.__init__(self, category=category, facade=True)
@@ -129,6 +142,12 @@ class CliffordAlgebraIndices(UniqueRepresentation, Parent):
             True
             sage: len(idx) == 2^7
             True
+
+            sage: idx = CliffordAlgebraIndices(7, 3)
+            sage: idx.cardinality() == binomial(7, 3)
+            True
+            sage: len(idx) == binomial(7, 3)
+            True
         """
         return self._cardinality
 
@@ -149,14 +168,20 @@ class CliffordAlgebraIndices(UniqueRepresentation, Parent):
             Subsets of {0}
             sage: CliffordAlgebraIndices(2)
             Subsets of {0,1}
+            sage: CliffordAlgebraIndices(5, 3)
+            Subsets of {0,1,...,4} of size 3
         """
+        if self._degree is not None:
+            extra = f" of size {self._degree}"
+        else:
+            extra = ""
         if self._nbits == 0:
-            return "Subsets of {}"
+            return "Subsets of {}" + extra
         if self._nbits == 1:
-            return "Subsets of {0}"
+            return "Subsets of {0}" + extra
         if self._nbits == 2:
-            return "Subsets of {0,1}"
-        return f"Subsets of {{0,1,...,{self._nbits-1}}}"
+            return "Subsets of {0,1}" + extra
+        return f"Subsets of {{0,1,...,{self._nbits-1}}}" + extra
 
     def _latex_(self):
         r"""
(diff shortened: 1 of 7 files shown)


tscrim:
This will likely depend on #36748, which uses `Representation_abstract` and is likely to be merged before this.

src/sage/modules/with_basis/representation.py
```diff
@@ -750,12 +750,27 @@ def __classcall_private__(cls, reps, **options):
...
+            sage: type(T)
+            <class 'sage.combinat.free_module.CombinatorialFreeModule_Tensor_with_category'>
+            sage: T = tensor([R, C])
+            sage: type(T)
+            <class 'sage.combinat.free_module.CombinatorialFreeModule_Tensor_with_category'>
         """
-        assert (len(reps) > 0)
-        R = reps[0].base_ring()
+        assert len(reps) > 0
+        assert isinstance(reps[0], Representation_abstract)
```
trevorkarn, line 772: Is there a reason to an `assert` here instead of raising a `ValueError`?

src/sage/algebras/clifford_algebra.py
```diff
@@ -67,9 +67,22 @@ def __init__(self, Qdim):
...
             sage: type(i)
             <class 'sage.data_structures.bitset.FrozenBitset'>
+
+            sage: idx = CliffordAlgebraIndices(7, 3)
+            sage: idx._nbits
+            7
+            sage: idx._degree
+            3
+            sage: idx._cardinality
+            35
```
trevorkarn, line 77: Does `degree=0` work as expected?

src/sage/modules/with_basis/representation.py
tscrim: Yes, this should never occur. This class is constructed from `tensor([A,B,C])` -> `A.tensor(B,C)`, so there must be at least one such object. It's possible a user directly calls this, but no matter what, some information about what this is a representation for must be specified, which has to be given through the input modules.

src/sage/algebras/clifford_algebra.py
tscrim: There are other tests that do this indirectly, but I can add an explicit test for it.

tscrim: This was a good idea as not all of the methods were tested and there were places where this corner case needed to be handled.

src/sage/modules/with_basis/representation.py
```diff
@@ -475,132 +713,709 @@ def product_by_coercion(self, left, right):
...
+        return self.linear_combination(((self._on_basis(g, m), c)
+                                       for m, c in vec._monomial_coefficients.items()), not vec_on_left)
 
-    class Element(CombinatorialFreeModule.Element):
-        def _acted_upon_(self, scalar, self_on_left=False):
-            """
-            Return the action of ``scalar`` on ``self``.
 
-            EXAMPLES::
+class Representation_Tensor(CombinatorialFreeModule_Tensor, Representation_abstract):
```
mkoeppe, line 740: I think the naming of `CombinatorialFreeModule_Tensor` is unfortunate, and perhaps it's not too late to change it to conform to our conventions elsewhere, where it would be `CombinatorialFreeModule_tensor` or `CombinatorialFreeModule_tensor_product`.

src/sage/modules/with_basis/representation.py
```diff
@@ -201,9 +217,234 @@ def twisted_invariant_module(self, chi, G=None, **kwargs):
 
         return super().twisted_invariant_module(G, chi, side=side, **kwargs)
 
+    def representation_matrix(self, g, side=None, sparse=False):
```
mkoeppe, line 220: consider allowing other matrix constructor keywords, in particular `implementation`

src/sage/modules/with_basis/representation.py
```diff
@@ -475,132 +713,709 @@ def product_by_coercion(self, left, right):
...
+        R = rep.base_ring()
+        dim = rep.dimension()
+        if degree not in ZZ or degree < 0:
+            raise ValueError(f"the degree must be a nonnegative integer")
+        self._symalg = PolynomialRing(R, 'e', dim)
+        self._basis_order = list(rep.basis().keys())
+        G = self._symalg.gens()
+        self._inv_map = {b: G[i] for i, b in enumerate(self._basis_order)}
+        ind = IntegerVectors(degree, dim)
+        Representation_abstract.__init__(self, rep.semigroup(), rep.base_ring(), rep.side(),
```
mkoeppe, line 1198: Could this be set up as a subobject of `Representation_Tensor`?
And the exterior power as a quotient?

src/sage/modules/with_basis/representation.py
tscrim: I would prefer for now to follow the convention there.

src/sage/modules/with_basis/representation.py
tscrim: Yes, but it would make the code *significantly* more complicated and slower and we would only gain a certain amount of mathematical "purity." This really should stay as it is.

src/sage/modules/with_basis/representation.py
tscrim: We can't do this because `matrix.zero` does not support this.

mkoeppe: Why use that method anyway, not the `matrix` constructor?
I'd also suggest to return an immutable matrix in this method.

tscrim: > Why use that method anyway, not the `matrix` constructor?

`matrix` also does not support that. When I tested this, it seemed faster to manipulate the entries of the matrix rather than feed the data off (more transient objects as well), even more so when the matrices have less support (which can often happen).

> I'd also suggest to return an immutable matrix in this method.

Why? It isn't cached, and it is easy to make matrices immutable if needed (but less so the other way around).

mkoeppe: > `matrix` also does not support that.

You are right, sorry.

It does support the `space` keyword through which one can prescribe an `implementation`.

mkoeppe: Anyway, just a wishlist item.

src/sage/algebras/clifford_algebra.py
```diff
@@ -67,9 +67,30 @@ def __init__(self, Qdim):
...
+            0
+            sage: idx._cardinality
+            1
         """
         self._nbits = Qdim
-        self._cardinality = 2 ** Qdim
+        if degree is None:
+            self._cardinality = 2 ** Qdim
+        else:
+            from sage.functions.other import binomial
```
fchapoton: maybe prefer the fast version `from sage.arith.misc import binomial`

src/sage/modules/with_basis/representation.py
mkoeppe: OK

src/sage/modules/with_basis/representation.py
mkoeppe: Sure

fchapoton (approved):
ok, thanks

vbraun:
```
sage -t --long --warn-long 38.4 --random-seed=123 src/sage/combinat/specht_module.py
File "src/sage/combinat/specht_module.py", line 51, in sage.combinat.specht_module.SymmetricGroupRepresentation.__init__
Failed example:
    TestSuite(SM).run()
Expected nothing
Got:
    Failure in _test_not_implemented_methods:
      File "sage/structure/sage_object.pyx", line 644, in sage.structure.sage_object.SageObject._test_not_implemented_methods
        getattr(self, name)
      File "/home/release/Sage/src/sage/misc/abstract_method.py", line 220, in __get__
        raise NotImplementedError(repr(self))
    NotImplementedError: <abstract method _semigroup_action at 0x7fd9dc95dc60>
    <BLANKLINE>
    <BLANKLINE>
      File "/home/release/Sage/src/sage/misc/sage_unittest.py", line 297, in run
        test_method(tester=tester)
      File "sage/structure/sage_object.pyx", line 647, in sage.structure.sage_object.SageObject._test_not_implemented_methods
        tester.fail("Not implemented method: %s" % name)
      File "/home/release/Sage/local/var/lib/sage/venv-python3.11.1/lib/python3.11/unittest/case.py", line 703, in fail
        raise self.failureException(msg)
    AssertionError: Not implemented method: _semigroup_action
    ------------------------------------------------------------
    The following tests failed: _test_not_implemented_methods
```

tscrim:
Trivial fix from a mistake in changing how the generic representation refactoring. It should allow for the `_acted_upon_()` to be overwritten, so the `_semigroup_action()` should be an optional abstract method. I made this change. @fchapoton Can you quickly verify and then remove the needs work?

fchapoton (approved):
ok, looks good to me

tscrim:
Thank you!

PR #37397, opened 2024-02-19
fix some ruff suggestions in cluster_algebra_quiver

fchapoton:
this was scripted using `ruff` for the codes `UP` and `C4` in the cluster folder.

Plus a few hand-made changes about pep8

### :memo: Checklist


tscrim (approved):
LGTM.

I will say one comment: I don't really care for the change `set(...)` to `{...}`. The latter is very close to a `dict` construction in Python, leading to a very easy to make error of `{}` meant to be the empty set.

PR #36676, opened 2023-11-08
Restructure `sage.*.all` for modularization, replace relative by absolute imports

mkoeppe:
We restructure the `all.py` files for modularization.

Guided by the technical constraints outlined in https://groups.google.com/g/sage-devel/c/ozh-7ZZ848s, #35095 defines distribution packages (pip-installable packages) **sagemath-{brial, combinat, eclib, flint, gap, giac, glpk, graphs, groups, homfly, lcalc, libbraiding, libecm, linbox, modules, mpmath, ntl, pari, plot, polyhedra, schemes, singular, standard-no-symbolics, symbolics}**.

When a namespace package such as `sage.misc` is filled by modules from several distribution packages, we create modules named:
- `src/sage/misc/all__sagemath_environment.py` 
- `src/sage/misc/all__sagemath_objects.py`
- `src/sage/misc/all__sagemath_repl.py`

Import statements are moved from `src/sage/misc/all.py` to these files as appropriate, and `src/sage/misc/all.py` imports `*` from there.

Also some imports are replaced by lazy imports.

The new files provide the top level namespaces for the modularized distribution packages, thus [enabling modularized testing](https://doc.sagemath.org/html/en/developer/packaging_sage_library.html#testing-distribution-packages).

This design was introduced in #29865 (merged in Jan 2022, early in the Sage 9.6 development cycle).

- Copied from #35095.
- Part of #29705

Moreover, applied a one-line command to replace relative by absolute imports, thus complementing #36666, which does not touch `.all*` files.

The changes to other files in `sage.modules` etc. come from the PRs #36597, #36572, #36588, #36589 merged here and do not need review.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36597 (merged here to resolve merge conflict)
- Depends on https://github.com/sagemath/sage/pull/36572 (merged here to resolve merge conflict)
- Depends on https://github.com/sagemath/sage/pull/36588 (merged here to resolve merge conflict)
- Depends on https://github.com/sagemath/sage/pull/36589 (merged here to resolve merge conflict)
- Depends on #36449 (merged here to resolve merge conflict)
- Depends on #36390 (merged here to resolve merge conflict)

PR #37399, opened 2024-02-19
Developer Guide: Document practices for data files

mkoeppe:
Topics here:
- ext_data is legacy location only, see #33037
- separate repos for large data files
- importlib_resources 

Follow-ups:
- use of Features

### :memo: Checklist

### DIFF
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -184,18 +184,13 @@ included in one of the following places:
   :func:`importlib.resources.as_file`. It should be imported in the
   same way as shown above.
 
-  .. NOTE::
+- Older code in the Sage library accesses
+  the package data in more direct ways. For example,
+  ``SAGE_ROOT/src/sage/interfaces/maxima.py`` uses the file
+  ``SAGE_ROOT/src/sage/interfaces/maxima.lisp`` at runtime, so it
+  refers to it as::
 
-     You may notice that some older code in the Sage library accesses
-     the package data in more direct ways. For example,
-     ``SAGE_ROOT/src/sage/interfaces/maxima.py`` uses the file
-     ``SAGE_ROOT/src/sage/interfaces/maxima.lisp`` at runtime, so it
-     refers to it as::
-
-       os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')
-
-     This is no longer recommended, and PRs that update such uses
-     are welcome.
+    os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')
 
 - In an appropriate subdirectory of ``SAGE_ROOT/src/sage/ext_data/``.
   (At runtime, it is then available in the directory indicated by
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -154,12 +154,23 @@ included in one of the following places:
 
 - In the directory of the Python code that uses that file.  When the
   Sage library is installed, the file will be installed in the same
-  location as the Python code. For example,
-  ``SAGE_ROOT/src/sage/interfaces/maxima.py`` needs to use the file
-  ``SAGE_ROOT/src/sage/interfaces/maxima.lisp`` at runtime, so it refers
-  to it as ::
+  location as the Python code. This is referred to as "package data".
 
-    os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')
+  The preferred way to access the data from Python is using the
+  function :func:`importlib.resources.files`.
+
+  .. NOTE::
+
+     You may notice that some older code in the Sage library accesses
+     the package data in more direct ways. For example,
+     ``SAGE_ROOT/src/sage/interfaces/maxima.py`` uses the file
+     ``SAGE_ROOT/src/sage/interfaces/maxima.lisp`` at runtime, so it
+     refers to it as::
+
+       os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')
+
+     This is no longer recommended, and PRs that update such uses
+     are welcome.
 
 - In an appropriate subdirectory of ``SAGE_ROOT/src/sage/ext_data/``.
   (At runtime, it is then available in the directory indicated by
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -157,7 +157,13 @@ included in one of the following places:
   location as the Python code. This is referred to as "package data".
 
   The preferred way to access the data from Python is using the
-  function :func:`importlib.resources.files`.
+  function :func:`importlib.resources.files`. It should be imported
+  as follows (see :ref:`section-python-language-standard`)::
+
+    try:
+        from importlib_resources import files
+    except ImportError:
+        from importlib.resources import files
 
   .. NOTE::
 
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -157,7 +157,8 @@ included in one of the following places:
   location as the Python code. This is referred to as "package data".
 
   The preferred way to access the data from Python is using the
-  `importlib.resources API <https://importlib-resources.readthedocs.io/en/latest/using.html>`,
+  `importlib.resources API
+  <https://importlib-resources.readthedocs.io/en/latest/using.html>`_,
   in particular the function :func:`importlib.resources.files`.
   Using it, you can:
 
@@ -203,9 +204,11 @@ is proposed to do the following:
 
 - add metadata to the repository that make it a pip-installable
   package (distribution package), as explained for example in the
-  `Python Packaging User Guide <https://packaging.python.org/en/latest/tutorials/packaging-projects/>`_,
+  `Python Packaging User Guide
+  <https://packaging.python.org/en/latest/tutorials/packaging-projects/>`_,
 
-- `upload it to PyPI <https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives>`_,
+- `upload it to PyPI
+  <https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives>`_,
 
 - create metadata in ``SAGE_ROOT/build/pkgs`` that make your new
   pip-installable package known to Sage; see :ref:`chapter-packaging`.
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -168,6 +168,15 @@ included in one of the following places:
     except ImportError:
         from importlib.resources import files
 
+  After this import, you can:
+
+  - open a resource for text reading: `fd = files(package).joinpath(resource).open('rt')`
+  - open a resource for binary reading: `fd = files(package).joinpath(resource).open('rb')`
+  - read a resource as text: `text = files(package).joinpath(resource).read_text()`
+  - read a resource as bytes: `bytes = files(package).joinpath(resource).read_bytes()`
+  - open a xz resource for text reading: `fd = lzma.open(files(package).joinpath(resource), 'rt')`
+  - open a xz resource for binary reading: `fd = lzma.open(files(package).joinpath(resource), 'rb')`
+
   If the file needs to be used outside of Python, then the
   preferred way is using the context manager
   :func:`importlib.resources.as_file`. It should be imported as
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -159,25 +159,14 @@ included in one of the following places:
   The preferred way to access the data from Python is using the
   `importlib.resources API <https://importlib-resources.readthedocs.io/en/latest/using.html>`,
   in particular the function :func:`importlib.resources.files`.
-  It should be imported as follows
-  (see :ref:`section-python-language-standard`)::
+  Using it, you can:
 
-    import sys
-
-    if sys.version_info < (3, 11):
-        # Use backport package providing Python 3.11 features
-        from importlib_resources import files
-    else:
-        from importlib.resources import files
-
-  After this import, you can:
-
-  - open a resource for text reading: `fd = files(package).joinpath(resource).open('rt')`
-  - open a resource for binary reading: `fd = files(package).joinpath(resource).open('rb')`
-  - read a resource as text: `text = files(package).joinpath(resource).read_text()`
-  - read a resource as bytes: `bytes = files(package).joinpath(resource).read_bytes()`
-  - open a xz resource for text reading: `fd = lzma.open(files(package).joinpath(resource).open('rb'), 'rt')`
-  - open a xz resource for binary reading: `fd = lzma.open(files(package).joinpath(resource).open('rb'), 'rb')`
+  - open a resource for text reading: ``fd = files(package).joinpath(resource).open('rt')``
+  - open a resource for binary reading: ``fd = files(package).joinpath(resource).open('rb')``
+  - read a resource as text: ``text = files(package).joinpath(resource).read_text()``
+  - read a resource as bytes: ``bytes = files(package).joinpath(resource).read_bytes()``
+  - open an xz-compressed resource for text reading: ``fd = lzma.open(files(package).joinpath(resource).open('rb'), 'rt')``
+  - open an xz-compressed resource for binary reading: ``fd = lzma.open(files(package).joinpath(resource).open('rb'), 'rb')``
 
   If the file needs to be used outside of Python, then the
   preferred way is using the context manager
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -200,7 +200,9 @@ included in one of the following places:
     from sage.env import SAGE_EXTCODE
     file = os.path.join(SAGE_EXTCODE, 'directory', 'file')
 
-In both cases, the files must be listed (explicitly or via wildcards) in
+  This practice is deprecated, see :issue:`33037`.
+
+In all cases, the files must be listed (explicitly or via wildcards) in
 the section ``options.package_data`` of the file
 ``SAGE_ROOT/pkgs/sagemath-standard/setup.cfg.m4`` (or the corresponding
 file of another distribution).
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -157,13 +157,27 @@ included in one of the following places:
   location as the Python code. This is referred to as "package data".
 
   The preferred way to access the data from Python is using the
-  function :func:`importlib.resources.files`. It should be imported
-  as follows (see :ref:`section-python-language-standard`)::
+  `importlib.resources API <https://importlib-resources.readthedocs.io/en/latest/using.html>`,
+  in particular the function :func:`importlib.resources.files`.
+  It should be imported as follows
+  (see :ref:`section-python-language-standard`)::
 
     try:
+        # Use backport package providing Python 3.11 features
         from importlib_resources import files
     except ImportError:
         from importlib.resources import files
+
+  If the file needs to be used outside of Python, then the
+  preferred way is using the context manager
+  :func:`importlib.resources.as_file`. It should be imported as
+  follows::
+
+    try:
+        # Use backport package providing Python 3.11 features
+        from importlib_resources import as_file
+    except ImportError:
+        from importlib.resources import as_file
 
   .. NOTE::
 
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -104,7 +104,7 @@ of the directory containing the Sage sources:
             setup.py
             ...
             sage/            # Sage library
-                ext_data/    # extra Sage resources (formerly src/ext)
+                ext_data/    # extra Sage resources (legacy)
             bin/             # the scripts in local/bin that are tracked
         upstream/            # tarballs of upstream sources
         local/               # installed binaries
@@ -160,6 +160,16 @@ of the following places:
   to it as ::
 
     os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')
+
+- Large data files should not be added to the Sage source tree.
+  Instead:
+
+  - Create a separate git repository for them
+  - Add metadata in your repository that make it a pip-installable
+    package (distribution package)
+  - Upload it to PyPI
+  - Create metadata in ``SAGE_ROOT/build/pkgs`` that make your new
+    pip-installable package known to Sage
 
 - In an appropriate subdirectory of ``SAGE_ROOT/src/sage/ext_data/``.
   (At runtime, it is then available in the directory indicated by
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -180,12 +180,13 @@ is proposed to do the following:
 - create a separate git repository and upload them there [2]_,
 
 - add metadata to the repository that make it a pip-installable
-  package (distribution package),
+  package (distribution package), as explained for example in the
+  `Python Packaging User Guide <https://packaging.python.org/en/latest/tutorials/packaging-projects/>`_,
 
-- upload it to PyPI,
+- `upload it to PyPI <https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives>`_,
 
 - create metadata in ``SAGE_ROOT/build/pkgs`` that make your new
-  pip-installable package known to Sage.
+  pip-installable package known to Sage; see :ref:`chapter-packaging`.
 
 For guiding examples of external repositories that host large data
 files, see https://github.com/sagemath/conway-polynomials, and
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -89,9 +89,9 @@ In particular,
 Files and directory structure
 =============================
 
-Roughly, the Sage directory tree is layout like this. Note that we use
-``SAGE_ROOT`` in the following as a shortcut for the (arbitrary) name
-of the directory containing the Sage sources:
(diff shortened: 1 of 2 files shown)


mkoeppe:
@tornaria @gmou3 Let's collaborate here to bring our documentation up to date on these topics.

src/doc/en/developer/coding_basics.rst
```diff
@@ -165,6 +165,16 @@ of the following places:
...
+- Large data files should not be added to the Sage source tree.
+  Instead:
+
+  - Create a separate git repository for them
+  - Add metadata in your repository that make it a pip-installable
+    package (distribution package)
+  - Upload it to PyPI
+  - Create metadata in ``SAGE_ROOT/build/pkgs`` that make your new
+    pip-installable package known to Sage
+
```
tornaria: Nice. Some ideas:
 - give https://github.com/sagemath/conway-polynomials and https://github.com/gmou3/matroid-database as examples
 - suggest compressing large data files with xz, read transparently as in https://github.com/sagemath/conway-polynomials/blob/master/src/conway_polynomials/__init__.py#L116
 - suggest trying `xz -T 0 -e` to compress, sometimes gives big gains for highly repetitive text files (without affecting the cost for decompression).

mkoeppe: Thanks! Yes, linking to these projects as examples of best practices is a good idea. Will do (unless someone sends me a PR first... hint, hint)

tornaria:
For accessing resources in packages, link to https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files. Note everything else is deprecated.

Basic examples:

1. open a resource for text reading: `fd = files(package).joinpath(resource).open('r')`
2. open a resource for binary reading: `fd = files(package).joinpath(resource).open('rb')`
3. read a resource as text: `text = files(package).joinpath(resource).read_text()`
4. read a resource as bytes: `bytes = files(package).joinpath(resource).read_bytes()`
5. open a xz resource for text reading: `fd = lzma.open(files(package).joinpath(resource), 'r')`
6. open a xz resource for binary reading: `fd = lzma.open(files(package).joinpath(resource), 'rb')`

Note: `as_files()` is only necessary if one needs an actual file by path, e.g. to pass to an external command. AFAICT, it's not needed to use as in the examples above.

tornaria:
Notes: 
1. I think in both conway-polynomials and matroid-database, the construct `TextIOWrapper(lzma.open(FILE))` is equivalent to `lzma.open(FILE, 'r')`
2. Is the `as_files(dbpath)` really necessary in https://github.com/sagemath/conway-polynomials/blob/master/src/conway_polynomials/__init__.py#L117 ?
3. Is the `.open("rb")` really necessary in https://github.com/gmou3/matroid-database/blob/main/src/matroid_database/__init__.py#L89 ?

gmou3:
This code looks nice to me:
```
import lzma

path = files(module).joinpath(name).with_suffix(".txt.xz")
try:
    return lzma.open(path, 'rt')
except FileNotFoundError:
```
It seems like we want the `'rt'` option: <https://docs.python.org/3/library/lzma.html>.

tornaria:
> It seems like we want the `'rt'` option: https://docs.python.org/3/library/lzma.html.

You are right; a different default from `io.open()` and also a different meaning of `'r'` :shrug:

src/doc/en/developer/coding_basics.rst
```diff
@@ -153,17 +153,28 @@ Adding new top-level packages below :mod:`sage` should be done
...
+     You may notice that some older code in the Sage library accesses
+     the package data in more direct ways. For example,
+     ``SAGE_ROOT/src/sage/interfaces/maxima.py`` uses the file
+     ``SAGE_ROOT/src/sage/interfaces/maxima.lisp`` at runtime, so it
+     refers to it as::
+
+       os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')
+
+     This is no longer recommended, and PRs that update such uses
+     are welcome.
```
tornaria: We should not un-recommend this until `sage.interfaces` is a package or we remove support for python 3.9. The alternative is _ugly_, and is adding code that will have to be replaced again soon.

mkoeppe: > adding code that will have to be replaced again soon

why is this?

mkoeppe:
@orlitzky @soehms You might be interested in this PR as well

src/doc/en/developer/coding_basics.rst
tornaria: importlib_resources is only needed for namespace packages for python 3.9 so your `try` hack is unecessary with python 3.10.

mkoeppe: Please. It's not a "hack". These are the best practices.

And by documenting this method to import it, we avoid the "too much information" problem where we tell developers every little detail of what is supported on what version etc.

tornaria: > Please. It's not a "hack". These are the best practices.

According to whom?

orlitzky:
> Notes:
> 
>     1. I think in both conway-polynomials and matroid-database, the construct `TextIOWrapper(lzma.open(FILE))` is equivalent to `lzma.open(FILE, 'r')`
> 
>     2. Is the `as_files(dbpath)` really necessary in https://github.com/sagemath/conway-polynomials/blob/master/src/conway_polynomials/__init__.py#L117 ?

They make the type-checker (mypy) happy.

src/doc/en/developer/coding_basics.rst
```diff
@@ -153,17 +153,57 @@ Adding new top-level packages below :mod:`sage` should be done
...
+  `importlib.resources API <https://importlib-resources.readthedocs.io/en/latest/using.html>`,
+  in particular the function :func:`importlib.resources.files`.
+  It should be imported as follows
+  (see :ref:`section-python-language-standard`)::
+
+    try:
+        # Use backport package providing Python 3.11 features
+        from importlib_resources import files
+    except ImportError:
+        from importlib.resources import files
```
tornaria:
```suggestion
    from importlib.resources import files
```

maybe with a note about not using resources in namespace packages until 3.9 is removed support, i.e. apr 2024?. Really not worth doing this for 2 months and using the version of importlib.resources from python is better to avoid bleeding edge and stick to long term stable API.

Your code allows one, in principle, to use features from new `importlib_resources` which are not available in the version of python used. But if those features are indeed used, this results in an incompatibility with a system where python is new enough to have `importlib.resources` but where `importlib_resources` is not installed (or not up to date). From that point of view is much better to stick to the stable API, meaning features that are common available in all the supported versions of python.

The only newer nicety I see (available on 3.12) is that one can use `files()` to stand for `files(__package__)`

mkoeppe: The case of "`importlib_resources` not up to date" is a real point. I'll make a change.

I don't agree with the rest what you're saying.

src/doc/en/developer/coding_basics.rst
mkoeppe: @tornaria are these points taken care of by the current version of the text already?

src/doc/en/developer/coding_basics.rst
mkoeppe: OK, maybe second-best practices

src/doc/en/developer/coding_basics.rst
```diff
@@ -153,17 +153,57 @@ Adding new top-level packages below :mod:`sage` should be done
...
+
+  After this import, you can:
+
+  - open a resource for text reading: `fd = files(package).joinpath(resource).open('rt')`
+  - open a resource for binary reading: `fd = files(package).joinpath(resource).open('rb')`
+  - read a resource as text: `text = files(package).joinpath(resource).read_text()`
+  - read a resource as bytes: `bytes = files(package).joinpath(resource).read_bytes()`
+  - open a xz resource for text reading: `fd = lzma.open(files(package).joinpath(resource), 'rt')`
+  - open a xz resource for binary reading: `fd = lzma.open(files(package).joinpath(resource), 'rb')`
```
tornaria:
```suggestion
  - open a xz resource for text reading: `fd = lzma.open(files(package).joinpath(resource).open('rb'), 'rt')`
  - open a xz resource for binary reading: `fd = lzma.open(files(package).joinpath(resource).open('rb'), 'rb')`
```

It seems `lzma.open()` accepts either a filename or an open stream, so the previous suggestion _will not work_ e.g. if the module is zipped.

The two ways to work around this are:
1. use `open('rb')` in the resource (as here, as in matroid-database)
2. use `as_file(...)` to create an actual file to pass the filename to `lzma.open(...)` (as done in conway-polynomials)

I think the first one is simpler and possibly cheaper for the case of module-within-a-zip-file since it doesn't need to extract the xz file from the zip file to the filesystem before decompressing the xz file.

I guess both should satisfy the "type-checker" but what can you expect if you add a trunk to a duck...

The documentation of `lzma.open()` claims that passing `'rt'` results in a `TextIOWrapper` and this matches my experiment.

src/doc/en/developer/coding_basics.rst
mkoeppe: *Now* (with abf3ae6), at best practices.

src/doc/en/developer/coding_basics.rst
mkoeppe: In abf3ae6, I've changed it to using an explicit version check (for 3.11!) instead of `try... except`.

I'm using 3.11, not 3.9. 
Note the section on "Language Standard" in the Developer's Guide: The idea is that we develop the Sage library to one language standard, and that's currently 3.9 + the backports for 3.11.

mkoeppe: > until 3.9 is removed support, i.e. apr 2024?

We don't have a set schedule for this yet.

mkoeppe: > The only newer nicety I see (available on 3.12) is that one can use files() to stand for files(__package__)

That's quite nice, so let's please merge
- #37402

src/doc/en/developer/coding_basics.rst
gmou3: I removed the `.open('rb')` yesterday: https://github.com/gmou3/matroid-database/blob/39b05995006d4a9ce5d07b15d3ff3e66a62a0175/src/matroid_database/__init__.py#L88

How can I replicate the issue you are mentioning?

tornaria: To reproduce create a zip file with your package (e.g. build the wheel) and run with `PYTHONPATH=path/to/zipfile`. For instance:

```
$ PYTHONPATH=src python
Python 3.12.2 (main, Feb 15 2024, 06:47:30) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from matroid_database import *
>>> for m in all_matroids_revlex(2, 1): print(m)
... 
**
0*
```
works, but
```
$ PYTHONPATH=dist/matroid_database-0.2-py3-none-any.whl python
Python 3.12.2 (main, Feb 15 2024, 06:47:30) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from matroid_database import *
>>> for m in all_matroids_revlex(2, 1): print(m)
... 
  File "<stdin>", line 1, in <module>
  File "/home/tornaria/src/sage/matroid-database/dist/matroid_database-0.2-py3-none-any.whl/matroid_database/__init__.py", line 103, in all_matroids_revlex
ImportError: cannot import name '_all' from 'matroid_database' (/home/tornaria/src/sage/matroid-database/dist/matroid_database-0.2-py3-none-any.whl/matroid_database/__init__.py)
```
doesn't.

However, the error above is not because of `lzma.open(...)`: in a surprising turn of events, it seems zip files don't support namespace packages either so `from . import _all` fails :rofl:

Also `.with_suffix()` doesn't work on resources from zip files. With the following change:

```diff
--- a/src/matroid_database/__init__.py
+++ b/src/matroid_database/__init__.py
@@ -83,7 +83,7 @@ def _open_data(module, name):
     import lzma
     from importlib.resources import files
 
-    path = files(module).joinpath(name).with_suffix('.txt.xz')
+    path = files(module).joinpath(name + '.txt.xz')
     try:
         return lzma.open(path, 'rt')
     except FileNotFoundError:
@@ -100,8 +100,7 @@ def all_matroids_revlex(n, r):
     Return an iterator over the revlex encodings of all matroids of given
     number of elements and rank.
     """
-    from . import _all
-    with _open_data(_all, f"allr{r}n{n:02d}") as f:
+    with _open_data(__package__, f"_all/allr{r}n{n:02d}") as f:
         while s := f.readline():
             yield s.strip()
 
```

Now we get the issue with `lzma.open(..)` not supporting resources in zip files:
```
$ PYTHONPATH=dist/matroid_database-0.2-py3-none-any.whl python -c $'from matroid_database import *\nfor m in all_matroids_revlex(2, 1): print(m)'
  File "<string>", line 2, in <module>
  File "/home/tornaria/src/sage/matroid-database/dist/matroid_database-0.2-py3-none-any.whl/matroid_database/__init__.py", line 103, in all_matroids_revlex
  File "/home/tornaria/src/sage/matroid-database/dist/matroid_database-0.2-py3-none-any.whl/matroid_database/__init__.py", line 88, in _open_data
  File "/usr/lib/python3.12/lzma.py", line 309, in open
    binary_file = LZMAFile(filename, lz_mode, format=format, check=check,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/lzma.py", line 127, in __init__
    raise TypeError("filename must be a str, bytes, file or PathLike object")
TypeError: filename must be a str, bytes, file or PathLike object
```

Now apply:
```diff
--- a/src/matroid_database/__init__.py
+++ b/src/matroid_database/__init__.py
@@ -85,7 +85,7 @@ def _open_data(module, name):
 
     path = files(module).joinpath(name + '.txt.xz')
     try:
-        return lzma.open(path, 'rt')
+        return lzma.open(path.open('rb'), 'rt')
     except FileNotFoundError:
         raise ValueError(
             "unable to open %s" % path +
```

And voila:
```
$ PYTHONPATH=dist/matroid_database-0.2-py3-none-any.whl python -c $'from matroid_database import *\nfor m in all_matroids_revlex(2, 1): print(m)'
**
0*
```


I'm not sure what is the value of supporting running from zip files, but the whole push against `ext_data` and in favor of `importlib.resources` is to support it...

tornaria: BTW, to build or rebuild the wheel I just do
```
$ python -m build -nxw
...
Successfully built matroid_database-0.2-py3-none-any.whl
```
Redo this step each time you change the source.

src/doc/en/developer/coding_basics.rst
tornaria: > > until 3.9 is removed support, i.e. apr 2024?
> 
> We don't have a set schedule for this yet.

42 months as per NEP 29 is apr 2024. They call this best practices.

src/doc/en/developer/coding_basics.rst
```diff
@@ -153,17 +153,53 @@ Adding new top-level packages below :mod:`sage` should be done
...
+  It should be imported as follows
+  (see :ref:`section-python-language-standard`)::
+
+    import sys
+
+    if sys.version_info < (3, 11):
+        # Use backport package providing Python 3.11 features
+        from importlib_resources import files
+    else:
+        from importlib.resources import files
```
tornaria:
```suggestion
    from importlib.resources import files
```
I already gave rationale.

src/doc/en/developer/coding_basics.rst
tornaria: Please give an external reference (i.e. not yourself)

src/doc/en/developer/coding_basics.rst
mkoeppe: You already know that I disagree, and I have explained it.

tornaria: If you don't agree with the suggestion, feel free to label the PR as disputed.

src/doc/en/developer/coding_basics.rst
mkoeppe: > I'm not sure what is the value of supporting running from zip files, 

I don't have a use case for it either, except for using it to have `pip` self-install from its own wheel.

> but the whole push against `ext_data` and in favor of `importlib.resources` is to support it...

No, it's not.
It's to support modularization via distributions shipping namespace packages.

src/doc/en/developer/coding_basics.rst
mkoeppe: That's not how it works.

src/doc/en/developer/coding_basics.rst
mkoeppe: Please @tornaria. Our project has not adopted NEP 29. If you want to know about that, read
 https://github.com/sagemath/sage/wiki/NEP-29:-Python-version-strategy

src/doc/en/developer/coding_basics.rst
gmou3: Ok, thanks. I'll do it like this since it is more robust. One last aesthetic thing: is there a reason to call the data folders `_all` and `_unorientable` instead of `all` and `unorientable`?

src/doc/en/developer/coding_basics.rst
mkoeppe: Hard to tell what you are looking for. https://importlib-resources.readthedocs.io/en/latest/using.html# leaves no doubt how to import from it. We have decided to conditionalize the dependency on the backport package in #36776 @orlitzky, hence here we conditionalize the import.

In case you are looking for this: You can see how it is used in projects in the wild in this code search:
https://github.com/search?q=%22from+importlib_resources+import%22&type=code&p=1

mkoeppe: > We should not un-recommend this until ...

@tornaria I've changed it in 4d38922eb81, please take a look.

mkoeppe:
Let's get this in please.

orlitzky:
I agree with @tornaria that it would be better to use `importlib.resources` unconditionally, and simply add an `__init__.py` to the database packages in the meantime (I already did this with conway-polynomials). We would be better off without the three backport packages.

mkoeppe:
But this document does not describe the use of importlib.resources in the standalone packages.
It describes its use in the Sage library.

orlitzky:
> But this document does not describe the use of importlib.resources in the standalone packages. It describes its use in the Sage library.

Ok, but same answer for all the same reasons.

mkoeppe:
Well, no. The modularization of the Sage library and its uses of namespace packages does not disappear just like that.

orlitzky:
The database should only be present in one distribution right? And if not, I think there's a decent consensus that we should chill for a while on the init file removal until namespace packages are well-supported upstream.

mkoeppe:
> I think there's a decent consensus that we should chill for a while on the init file removal until namespace packages are well-supported upstream.

No, Michael, there is no such consensus, and certainly not a "decent" one. That the Sage library uses namespace packages for the purpose of the modularization is documented policy in the developer's guide.

mkoeppe:
> The database should only be present in one distribution right?

For the separate distributions (conway-polynomials, matroid-database), using an ordinary package (with `__init__.py`) is of course the right choice, as they control their own namespaces.

mkoeppe:
What we are trying to document here is to cover cases such as `SAGE_ROOT/src/sage/interfaces/maxima.lisp`.
`sage.interfaces` is a namespace package, and in fact one that is designed to be split over 6 distribution packages:
https://github.com/sagemath/sage/pull/36676/files#diff-2af172ee945eebb48bbe8a3cdd6b523221c1a995b2e4f8ac23fa7c3e1181e6ef

orlitzky:
> No, Michael, there is no such consensus, and certainly not a "decent" one. That the Sage library uses namespace packages for the purpose of the modularization is documented policy in the developer's guide.

It's in the developer's guide because you put it there.


> What we are trying to document here is to cover cases such as `SAGE_ROOT/src/sage/interfaces/maxima.lisp`.
`sage.interfaces` is a namespace package, and in fact one that is designed to be split over 6 distribution packages:

Handled by doing nothing and waiting for namespace package support upstream. Nobody needs this by tomorrow, and the maintenance to keep it working is a headache.

mkoeppe:
None of this sounds like the beginning of an interesting discussion; rather it sounds like the end of a discussion.

src/doc/en/developer/coding_basics.rst
mkoeppe: I have taken this part out now to regain velocity

tornaria: Thanks.

src/doc/en/developer/coding_in_python.rst
```diff
@@ -36,11 +37,24 @@ using one of two mechanisms:
...
+        if sys.version_info < (3, 11):
+            # Use backport package providing Python 3.11 features
+            from importlib_resources import files
+        else:
+            from importlib.resources import files
+
   - `typing_extensions <../reference/spkg/typing_extensions>`_
-    (to be used in place of ``typing``).
+    (to be used in place of ``typing`` when any of the features
+    introduced after Python 3.9.0 are used).
```
tornaria: You forgot about this, other that this the PR looks good to me.

I would remove almost all of this section, starting from "Some key language and library functions..." to the end, but I understand if you don't want to do it here.

mkoeppe: I didn't forget about it. I actually changed the language so it does not unconditionally asks for it to be imported like that.

tornaria: ~~Please, let's not play games.~~

Two reviewers showed concern for this. You can have a different opinion, but if you accept to take it out "to regain velocity", then do it ~~without tricks~~ fully. I already stated my positive review conditional on this.

Edit: toned it down. I didn't intend to offend @mkoeppe at all. The underlying message is a constructive review to get the 95% of your PR that we all agree is a very nice improvement.

mkoeppe: Gonzalo, this is uncalled for.

mkoeppe: What is your concern about this improvement of the documentation in the "Language Standard" section?

tornaria: > What is your concern about this improvement

This is your opinion, reviewers don't agree, please respect our point of view.

As it was mentioned before, some distributions don't ship `importlib_resources` anymore for good reasons so we should not use it (we don't) nor instruct people to use it.

AFAICT there is no use of any of the backport packages in sagemath. The fact that importlib.resources doesn't support namespace packages is minor because:
 a. it will be supported in python 3.9 which is not very far away
 b. for the time being we can either place package data in normal packages, or just use the old method as in `os.path.join(os.path.dirname(__file__), 'sage-maxima.lisp')`, this one actually makes it for a good example because it's in a namespace package.

NOTE: the main difference of using `importlib.resources` instead of the old way is that the former supports using package data from a zip file (i.e. directly from a wheel). However, running from zip files doesn't support extensions since dynamic libraries can't be `dlopen`ed from a zip file, so sagemath will never be able to run directly from wheels.

mkoeppe: Gonzalo, even the toned down version ("if you accept to take it out 'to regain velocity', then do it fully.") is not acceptable.

As I asked you privately on Zulip, it is necessary that you revise your communication style so that you do not operate with accusations and similar rhetorics of blame and intimidation. Such tactics are not welcome here.

src/doc/en/developer/coding_basics.rst
```diff
@@ -153,15 +153,35 @@ Adding new top-level packages below :mod:`sage` should be done
...
 - In the directory of the Python code that uses that file.  When the
   Sage library is installed, the file will be installed in the same
-  location as the Python code. For example,
-  ``SAGE_ROOT/src/sage/interfaces/maxima.py`` needs to use the file
-  ``SAGE_ROOT/src/sage/interfaces/maxima.lisp`` at runtime, so it refers
-  to it as ::
+  location as the Python code. This is referred to as "package data".
+
+  The preferred way to access the data from Python is using the
+  `importlib.resources API <https://importlib-resources.readthedocs.io/en/latest/using.html>`,
```
soehms: This line doesn't look rendered correctly when I follow the documentation link.

src/doc/en/developer/coding_in_python.rst
soehms: I'm not really familiar with the functionality discussed here. The only thing I can do is moderate the conflict (if my understanding of it is good enough). Could this be a compromise: we leave the example code-block as is and add a note that *some distributions no longer ship `importlib_resources` for good reasons*. Ideally we should add some `URL` pointers that address these *good reasons*.

tornaria: > I'm not really familiar with the functionality discussed here. The only thing I can do is moderate the conflict (if my understanding of it is good enough). Could this be a compromise: we leave the example code-block as is and add a note that _some distributions no longer ship `importlib_resources` for good reasons_. Ideally we should add some `URL` pointers that address these _good reasons_.

Sorry I missed this before.

It seems we don't agree whether to recommend or not using `importlib_resources`. In this
context, I think merging this nice PR without this small snippet is a good way to go, and I thought that's what @mkoeppe was proposing. That's why I gave it positive review conditional on this.

mkoeppe: I'll split out the improvement of this section of the manual (orthogonal to the goal of this PR) out to a separate PR and will take @soehms's suggestion into account there.

src/doc/en/developer/coding_basics.rst
mkoeppe: Fixed in 6045e5c7a36, I hope

src/doc/en/developer/coding_in_python.rst
mkoeppe: Done now in the rebased branch.

mkoeppe: ... and in #37654.

mkoeppe:
Other follow-ups:
- https://github.com/sagemath/sage/pull/37649
- https://github.com/sagemath/sage/pull/37650

mkoeppe:
Let's get this in please.

mkoeppe:
Thanks.

PR #37400, opened 2024-02-19
minor ruff fixes in quadratic forms (C4, UP, PERF)

fchapoton:
just a minor cleanup, after `ruff` suggestions, in `quadratic_forms` folder

### :memo: Checklist

### DIFF
--- a/src/sage/quadratic_forms/binary_qf.py
+++ b/src/sage/quadratic_forms/binary_qf.py
@@ -177,7 +177,7 @@ class BinaryQF(SageObject):
             sage: type(gp(f))
             <class 'sage.interfaces.gp.GpElement'>
         """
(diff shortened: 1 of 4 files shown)


tscrim:
Note that for a matrix space `MS`, the result of `MS(0)` and `MS.zero()` are different. The former is mutable, but the latter is immutable.

grhkm21 (approved):
LGTM. I assume it's only required for me to verify correctness, not look for additional improvements.

Just a question, what does the C4, UP, PERF mean?

fchapoton:
Thanks a lot. These letter combinations are some of the codes that the linter `ruff` can check and fix, see https://docs.astral.sh/ruff/rules/

PR #37402, opened 2024-02-20
`exceptiongroup`, `importlib_metadata`, `importlib_resources`, `typing_extensions`: Require 3.12 features

mkoeppe:
### :memo: Checklist

We tighten the lower bounds of these backport packages in install-requires so that the features of Python 3.12 can be used.

Also some updates to latest.

### :hourglass: Dependencies

- Depends on #38226


orlitzky:
Since Gentoo doesn't have an importlib_metadata package, I really don't want to have to install importlib_metadata "just in case."

mkoeppe:
Could you explain how this PR changes anything regarding this?

orlitzky:
Before I needed python-3.11 to skip the package that does nothing, but now I need python-3.12 to skip it.

mkoeppe:
```
$ cat build/pkgs/importlib_metadata/distros/gentoo.txt
dev-python/importlib_metadata
```

Is this not correct?

orlitzky:
> ```
> $ cat build/pkgs/importlib_metadata/distros/gentoo.txt
> dev-python/importlib_metadata
> ```
> 
> Is this not correct?

Oh, yeah. It's the other one. We don't have importlib-resources any more since https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=ef0ca954c

mkoeppe:
```
$ cat build/pkgs/importlib_resources/distros/gentoo.txt
dev-python/importlib_resources
```

mkoeppe:
> We don't have importlib-resources any more since https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=ef0ca954c

Well, this removal may have been a mistake...

mkoeppe:
Which two places?

orlitzky:
gentoo.git and sage.git

mkoeppe:
Oh, OK. Well, as you know, we have this information system for users and to provision the testing environment, and both is done in a way that is robust to inaccuracies.

mkoeppe:
> Before I needed python-3.11 to skip the package that does nothing, but now I need python-3.12 to skip it.

Is this a packaging concern, or about what you have on your machine?

orlitzky:
> > Before I needed python-3.11 to skip the package that does nothing, but now I need python-3.12 to skip it.
> 
> Is this a packaging concern, or about what you have on your machine?

Little of both. When doing sage development, I shouldn't have to install these as SPKGs -- that's one part of it.

For distribution packaging they're either superfluous or problematic. Right now they're superfluous. But if someone actually uses a python-3.12 feature from importlib-resources, then not having it would be problematic.

tornaria:
The spam bot just pinged me on this!

I fail to understand your motivation. What feature of python 3.12 is so important that you need to do this?

It seems much more reasonable to drop the packages altogether and avoid features not available in the range of supported python (3.9 - 3.12).

Can you make a list of the features you would like to have available that are not available on 3.9? Note that 3.9 is already outside the NEP range (42 months) so it wouldn't be such a shame to drop it if there is a strong reason.

mkoeppe:
> It seems much more reasonable

Why would you say that?

tornaria:
> > It seems much more reasonable
> 
> Why would you say that?

Can you make a list of the features you would like to have available that are not available on 3.9?

tornaria:
Can you make a list of the features you would like to have available that are not available on python 3.9?

PR #37404, opened 2024-02-20
add a polytope method to hypergeometric motives

fchapoton:
This is adding one new method to the class of hypergeometric motives, namely the construction of a lattice polytope.

### :memo: Checklist


src/sage/modular/hypergeometric_motive.py
```diff
@@ -1131,6 +1137,44 @@ def canonical_scheme(self, t=None):
         ideal = ring.ideal([eq0, eq1, self.M_value() * eq2_neg - t * eq2_pos])
         return Spec(ring.quotient(ideal))
 
+    def polytope(self):
```
mkoeppe: I would suggest to call it `lattice_polytope`
and provide a method `polyhedron` that returns a Polyhedron

fchapoton: I prefer to have just one method.

mkoeppe: Makes sense, I don't question your motives ;)

fchapoton:
good to go ?

mkoeppe (approved):
LGTM

fchapoton:
thanks

PR #37415, opened 2024-02-21
make EllipticCurvePoint_field inherit from EllipticCurvePoint

pjbruin:
The method `EllipticCurve_generic.__contains__()` checks for inheritance from `EllipticCurvePoint`; however, the other classes for points on elliptic curves do not inherit from this class.  This pull request fixes this. Note that there is some overlap with #33228.

This change makes `P in E` much faster:
- Over **Q**:
```python
E = EllipticCurve('37a1')
P = E((0, -1))
%timeit P in E
```
Before: `35.3 µs ± 186 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)`
After: `324 ns ± 0.535 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)`

- Over a finite field:
```python
F.<a> = FiniteField((23, 2), modulus='random')
E = EllipticCurve_from_j(F.random_element())
P = E.random_point()
%timeit P in E
```
Before: `36.4 µs ± 542 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)`
After: `333 ns ± 5.03 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)`


yyyyx4:
I agree that points over extension fields should be `in` a curve in principle, but I'm a bit worried that tons of code has been written (by people such as myself) under the assumption that this check only returns `True` when the base field is the same...

pjbruin:
I looked at how you dealt with this in #33228; I think this is a neat solution and I'm happy to simplify `__contains__()` in the same way you did it there.

yyyyx4:
Eventually we should still make the machinery for points over extension fields work. But I believe this will need some larger adjustments, in particular within some of the isogeny-related code...

pjbruin:
Indeed, there is definitely some work to be done.  I ran into a bug with point homsets over extension fields: #37427

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -153,14 +151,44 @@
...
+            sage: P.scheme()                                                            # needs sage.rings.number_field
+            Elliptic Curve defined by y^2 = x^3 + x + 1 over Number Field in a with defining polynomial x^2 - 3
+        """
+        return self.scheme()
+
+
+# While the class EllipticCurvePoint itself currently provides little
+# functionality, inheritance from EllipticCurvePoint is checked in the
+# __contains__() and __call__() methods of EllipticCurve_generic.
+class EllipticCurvePoint_field(EllipticCurvePoint,
```
kwankyu, line 187: I guess that the comment in lines 187-190 should go to the class definition ` EllipticCurvePoint`. Or you may just delete it since it has little content.

pjbruin: The comment was meant to clarify the new inheritance, but I agree it can be removed.

kwankyu: Thanks.

kwankyu (approved):
LGTM.

PR #37419, opened 2024-02-22
`src/setup.py`, `sage_setup`: Use `logging` instead of `distutils.log` and `print`

mkoeppe:
This takes care of most of:
- #33065.

By replacing `print` with logging commands, some messages are no longer printed out of order.
Also setting a logging format that shows the elapsed time. 
Together this makes it easier to see where time is spent.

### :memo: Checklist

### DIFF
--- a/pkgs/sagemath-bliss/setup.py
+++ b/pkgs/sagemath-bliss/setup.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 
-from distutils import log
 from setuptools import setup
 
 # Work around a Cython problem in Python 3.8.x on macOS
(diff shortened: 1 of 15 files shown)


grhkm21 (approved):
Looks good. Having #33065 is nice too.

By the way, I am not confident to review anything build system-related (yet) apart from trivial ones like this, so for bigger reviews definitely ping someone else instead.

mkoeppe:
Thanks!

vbraun:
Please don't set to positive review when the CI fails, there is an obvious broken doctest

vbraun:
Test still fail, as so often can already be seen on the CI so why set it to positive review?

orlitzky:
This is obsolete:

* setup.py is gone
* We don't maintain, ship, or test the sagemath-* distributions (don't need `sage_build_ext` etc.)
* Since we stopped using setup.py, AFAIK there are no ordering issues with using `print()` for the interpreter-generation message

PR #37420, opened 2024-02-22
Remove long/int relic

nbruin:
in py2, long and int were different types. In py3, only int remains, with long a synonym in cython.

### :memo: Checklist


videlec:
Wouldn't it make sense to get rid of `sage.rings.integer.int_to_Z`? (actually I would get rid of it and then rename `long_to_Z -> int_to_Z`)

nbruin:
Yes, it's worse. `int_to_Z` still exists but is buggy for what `int` is in Py3. Removed the definition to see what happens. Will probably rename `long_to_Z` to `int_to_Z` in the next step (with that change hopefully we don't have to change any code elsewhere)

nbruin:
OK, testing seems successful. Ready to be merged, I think. Review?

grhkm21 (approved):
Looks good to me, cleaning up code is always nice.

grhkm21:
And I agree that keeping (alias) `long_to_Z` is not necessary. Looking through the 244 usages of `long_to_Z` on [GitHub](https://github.com/search?q=long_to_Z+language%3ACython&type=code&p=5&l=Cython) I see all of them are just forks on Sage, so it's okay.

Also there's a function `PyLong_to_ZZ` still in Sage, but that's correct to keep since `PyLong` is a CPython thing(?)

PR #37432, opened 2024-02-23
`setup.py`: Remove `lazy_import_cache` use

mkoeppe:
### :memo: Checklist

Both versions of the sagelib `setup.py` have code that tries to remove the "lazy import cache".

The file name of the lazy import cache in `~/.sage` is keyed to `SAGE_LIB`, so any file name determined at build time is meaningless at runtime when installation goes through wheel building. We remove this code.

As `find ~/.sage -name "*-lazy_import*" | xargs ls -lat` shows, it has not been in use since 2021, as the Sage library no longer uses any `lazy_import` of `*`. But just in case some downstream / user code uses it, we reinstate the cache removal code in `build/pkgs/sagelib/spkg-install.in`

We deprecate the feature in #37433.

### DIFF
--- a/build/pkgs/sagelib/spkg-install.in
+++ b/build/pkgs/sagelib/spkg-install.in
@@ -76,6 +76,11 @@ else
     fi
 fi
 
+# Remove (potentially invalid) star import caches.
+# lazy star imports are not used by the Sage library, but could still be used by
+# downstream code.
+python3 -c 'import pathlib; from sage.misc.lazy_import_cache import get_cache_file; pathlib.Path(get_cache_file()).unlink(missing_ok=True)'
+
 # Issue #33103: The temp.* directories are large after a full build.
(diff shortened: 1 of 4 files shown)


kiwifb:
Moving something that is only useful for sage the distro into spkg-install.in and removing it from setup.py? Sounds like a no-brainer, I do not want that stuff in my packages.

mkoeppe:
> Moving something that is only useful for sage the distro into spkg-install.in and removing it from setup.py? Sounds like a no-brainer, I do not want that stuff in my packages.

Most likely it's not useful for anything, distro or not. Just keeping it out of excessive caution.
Per separation of phases in modern installs, it makes no sense in setup.py any more.

kiwifb (approved):

mkoeppe:
Thank you!

vbraun:
merge conflict

mkoeppe:
Trivial rebase.

PR #37440, opened 2024-02-23
Strengthen inequality in rigorous_line_integral

### LINKED ISSUE #37438: Bug in integration on a Riemann surface<title>
### Steps To Reproduce

Inside the Sage terminal, run 

```
R.<z,x> = QQ[]
f = x*(1^5+z^5) + (x*1*z)^2 - x^4*1*z - 2*1^3*z^3
S = Curve(f).riemann_surface()
S.riemann_matrix()
```

### Expected Behavior

You would expect this to successfully compute the Riemann matrix of the corresponding Riemann surface. 

### Actual Behavior

It gives a `ValueError` inside `rigorous_line_integral`. 

### Additional Information

This error doesn't occur if instead you either use `R.<x,z> = QQ[]` or `S = Curve(f).riemann_surface(integration_method="heuristic")`, which suggests it is an error in how the path-splitting method is being applied in certain configurations. This has been discussed [on sage-devel](https://groups.google.com/g/sage-devel/c/wcznXH3WLSM/m/EVP_1k-FAQAJ). 

### Environment

```markdown
- **OS**: both Red Hat Enterprise 8.9 and Ubuntu 18.04.6
- **Sage Version**: both 9.8, 10.3.beta0, and 10.3.beta4
```

### Checklist

DisneyHogg:
Simple edit to an inequality in `rigorous_line_integral`. This fixes #37438, which was an error where a division by 0 was occurring in the computation. This error happened because while initially we have `rho_t > rt`, floating point operations occur which meant that when we eventually divided by (effectively) `rho_t - rt` this floating point number was registering as 0. By strengthening the inequality such that we end up with `rho_t > rt + E_global`, we have effectively added some guard bits to ensure the distance away from zero remains nonzero after the additions and multiplications. 

Because this is only a very minor edit, I don't think it warrants a change to the test suite or the documentation. 

### :memo: Checklist

### DIFF
--- a/src/sage/schemes/riemann_surfaces/riemann_surface.py
+++ b/src/sage/schemes/riemann_surfaces/riemann_surface.py
@@ -2152,7 +2152,10 @@ class RiemannSurface():
         z1 = zwt(1)[0]
 
         # list of (centre, radius) pairs that still need to be processed
-        ball_stack = [(self._RR(1 / 2), self._RR(1 / 2), 0)]
+        # None is a sentinel value to indicate that the minimum number of
+        # nodes required to integrate on the corresponding segment within
+        # the required error tolerance is not yet known.
+        ball_stack = [(self._RR(1 / 2), self._RR(1 / 2), None)]
         alpha = self._RR(912 / 1000)
         # alpha set manually for scaling purposes. Basic benchmarking shows
         # that ~0.9 is a sensible value.
@@ -2236,20 +2239,15 @@ class RiemannSurface():
             ncts = [ct - rt / 2, ct + rt / 2]
             nrt = rt / 2
 
-            # lN == 0 is a placeholder value used to indicate that a value of
-            # N corresponding to a segment has not yet been computed.
-            # Because the output of local_N is always >= 3, we have no worries
-            # about 0 being the output misleadingly.
-            # As pointed out, 0 Should perhaps be replaced as a sentinel value
-            if lN == 0:
+            if lN is None:
                 cz = (1 - ct) * z0 + ct * z1
                 distances = [(cz - b).abs() for b in self.branch_locus]
                 rho_z = min(distances)
                 rho_t = rho_z / (z1_minus_z0).abs()
 
                 if rho_t <= rt:
-                    ball_stack.append((ncts[0], nrt, 0))
-                    ball_stack.append((ncts[1], nrt, 0))
+                    ball_stack.append((ncts[0], nrt, None))
+                    ball_stack.append((ncts[1], nrt, None))
                     continue
 
                 lN = local_N(ct, rt)
--- a/src/sage/schemes/riemann_surfaces/riemann_surface.py
+++ b/src/sage/schemes/riemann_surfaces/riemann_surface.py
@@ -2224,7 +2224,12 @@ class RiemannSurface():
             ncts = [ct - rt / 2, ct + rt / 2]
             nrt = rt / 2
 
-            if not lN:
+            # lN == 0 is a placeholder value used to indicate that a value of 
+            # N corresponding to a segment has not yet been computed. 
+            # Because the output of local_N is always >= 3, we have no worries
+            # about 0 being the output misleadingly. 
+            # As pointed out, 0 Should perhaps be replaced as a sentinel value 
+            if lN == 0:
                 cz = (1 - ct) * z0 + ct * z1
                 distances = [(cz - b).abs() for b in self.branch_locus]
                 rho_z = min(distances)
--- a/src/sage/schemes/riemann_surfaces/riemann_surface.py
+++ b/src/sage/schemes/riemann_surfaces/riemann_surface.py
@@ -2230,7 +2230,7 @@ class RiemannSurface():
                 rho_z = min(distances)
                 rho_t = rho_z / (z1_minus_z0).abs()
 
-                if rho_t <= rt:
+                if rho_t <= rt + E_global:
                     ball_stack.append((ncts[0], nrt, 0))
                     ball_stack.append((ncts[1], nrt, 0))
                     continue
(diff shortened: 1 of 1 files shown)


nbruin:
I always get nervous if float precision problems are avoided using an absolute error correction. Do you have some guarantee that rt is sufficiently close to 1 (or 0) that rt+E_global actually ends up being different from rt? Is it more appropriate to check that rho_t > rt*(1+E_global)? (or some other suitable value)

Another way of dealing with the problem would be in
https://github.com/sagemath/sage/blob/5c09be7c5a18139ec09591e0fab3227c6a589886/src/sage/schemes/riemann_surfaces/riemann_surface.py#L2215-L2218 to check for the denominator being 0 and then just setting Nrequired to (practically) infinity.

On a related note, in code close by:
https://github.com/sagemath/sage/blob/5c09be7c5a18139ec09591e0fab3227c6a589886/src/sage/schemes/riemann_surfaces/riemann_surface.py#L2234-L2235
can you document there that storing lN=0 means:no value computed? For that purpose I also think that `lN == 0` is the more appropriate check than `not lN`.

By the way, I can imagine that with the bounds we have on the function, for a small enough interval, we could actually find that 0 evaluation points would allow us to estimate the integral within precision (by 0). I don't know if the code will even produce that bound, but I'm not sure 0 is actually a safe sentinel value. You could use -1 instead?

DisneyHogg:
> I always get nervous if float precision problems are avoided using an absolute error correction. Do you have some guarantee that rt is sufficiently close to 1 (or 0) that rt+E_global actually ends up being different from rt? Is it more appropriate to check that rho_t > rt*(1+E_global)? (or some other suitable value)
> 
> Another way of dealing with the problem would be in
> 
> https://github.com/sagemath/sage/blob/5c09be7c5a18139ec09591e0fab3227c6a589886/src/sage/schemes/riemann_surfaces/riemann_surface.py#L2215-L2218
> 
> to check for the denominator being 0 and then just setting Nrequired to (practically) infinity.

See the latest commits, I have done something along the lines of the latter idea. In particular, I know check if the `N_required` produced is infinity (as was happening before, and which is actually ok), and if it is, set the output to be `2**self._prec`, which is essentially infinite if `2**(-self._prec)` is essentially zero. 


> On a related note, in code close by:
> 
> https://github.com/sagemath/sage/blob/5c09be7c5a18139ec09591e0fab3227c6a589886/src/sage/schemes/riemann_surfaces/riemann_surface.py#L2234-L2235
> 
> 
> can you document there that storing lN=0 means:no value computed? For that purpose I also think that `lN == 0` is the more appropriate check than `not lN`.
> By the way, I can imagine that with the bounds we have on the function, for a small enough interval, we could actually find that 0 evaluation points would allow us to estimate the integral within precision (by 0). I don't know if the code will even produce that bound, but I'm not sure 0 is actually a safe sentinel value. You could use -1 instead?

We never do produce 0 as an output of `local_N` so we don't need to worry about this sentinel value, but I have added a comment about this. You are right it may make sense in a separate pull request to just make the sentinel value something different, I would even suggest `None` which is more in keeping with what it's meant to represent in this case.

nbruin:
OK, while we're here: self._prec could be relatively low. I think it's possible to do these computations in `ComplexField(10)`, so `2**self._prec` is not a safe infinity stand-in. So you'd have to go a bit higher than that (you'd at least want a value for N that won't finish in reasonable time if it were used -- although that is a bit mean to the user who runs into a bug around this. You could use actual "Infinity", but that might cause problems elsewhere.

I would accept a value that is at least in the 2**31 range. Perhaps 2**60? you'll definitely run through a loop of that length.

Also for comments: please make them self-contained or put in a clear reference that won't go stale if you really have to. Comments like "It has been pointed out ..." will be puzzling and not helpful to future people (including you!) looking at the code. In this case that can just go. If you really want to highlight the possibility for change:

Since N=0 is technically a valid outcome if we know the integral can be approximated within the specified tolerance by 0, perhaps we should use a different sentinel value anyway?

src/sage/schemes/riemann_surfaces/riemann_surface.py
```diff
@@ -2208,31 +2224,30 @@ def local_N(ct, rt):
...
-                M = Delta
+                M = delta_z * cdgdz.abs() + pf2 * M_tilde
                 N_required = (
                     (M * (self._RR.pi() + 64 / (15 * (expr**2 - 1))) / E_global).log()
                     / (2 * expr.log())
-                ).ceil()
-                Ni = max(Ni, N_required)
+                )
+                if N_required == Infinity:
+                    return 2**max(60, self._prec)
```
nbruin, line 2233: Well, if we're going to use None as a sentinel value we might as well use it here. Also, I think `N_required.is_positive_infinity()` is the appropriate conditional here.

src/sage/schemes/riemann_surfaces/riemann_surface.py
```diff
@@ -2208,31 +2224,30 @@ def local_N(ct, rt):
...
                 rho_t = rho_z / (z1_minus_z0).abs()
 
                 if rho_t <= rt:
-                    ball_stack.append((ncts[0], nrt, 0))
-                    ball_stack.append((ncts[1], nrt, 0))
+                    ball_stack.append((ncts[0], nrt, None))
+                    ball_stack.append((ncts[1], nrt, None))
                     continue
 
                 lN = local_N(ct, rt)
```
nbruin, line 2253: I guess this line could now return `None`, so the line
```
            if sum(nNs) < lN:
```
below should now read
```
            if lN is None or sum(nNs) < lN:
```
(I guess `nNs` could contain None values too, but it is hard to imagine how that could happen without lN being None as well)

src/sage/schemes/riemann_surfaces/riemann_surface.py
DisneyHogg: I have made the change around check if `N_required` is infinite, but I don't think we need to change the output here. In this case it is not that `N_required` hasn't (initialisation) or can't (because circles won't fit which don't contain branch points) be computed yet, it's that it has been computed but it's just very large (in such a way that floating point errors can distinguish between its denominator and 0). The `nNs` condition thus doesn't need changing later.

nbruin (approved):
OK, I'm not sure I'm super happy with "large number" meaning "infinity", but I don't see a way in which it could make a difference in practice. The main issue here is that we're returning an *integer* and those don't naturally come with an "infinity". Sage has extensions that do allow for it, but it seems overkill to invoke those. So let's role with this.

Thank you for fixing!

PR #37441, opened 2024-02-23
Bind to FLINT/NTL API for polynomial modular exponentiation (new version)

GiacomoPope:
This is a rejuvenation of the pull request by @remyoudompheng https://github.com/sagemath/sage/pull/35320.

All I have done is rebase with develop and modify a few lines to fit with the current structuring of SageMath. I have also added a few additional docstrings.

As an example of the performance gain:

### Flint polynomials over prime fields

Old timings:

About 5x faster for flint using `nmod_poly`

```py
sage: R.<x> = GF(5)[]
sage: %time pow(x+1, 5**100, x^5 + 4*x + 3)
CPU times: user 551 µs, sys: 1e+03 ns, total: 552 µs
Wall time: 555 µs
x + 1
```

New timings:

```py
sage: R.<x> = GF(5)[]
sage:  %time pow(x+1, 5**100, x^5 + 4*x + 3)
CPU times: user 175 µs, sys: 1e+03 ns, total: 176 µs
Wall time: 179 µs
x + 1
```

### NTL polynomials over prime fields

About 1000x faster for NTL $\textrm{GF}(p)$
Old timings:

```py
sage: set_random_seed(0)
....: p = 2**255 - 19
....: R.<x> = F[]
....: f = R.random_element(degree=12)
....: g = R.random_element(degree=12)
....: %timeit pow(f, 2**33, g)
92.3 ms ± 947 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
sage: %timeit pow(f, 2**64, g)
1.29 s ± 35.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

New timings:

```py
sage: set_random_seed(0)
....: p = 2**255 - 19
....: R.<x> = F[]
....: f = R.random_element(degree=12)
....: g = R.random_element(degree=12)
....: %timeit pow(f, 2**33, g)
651 µs ± 28.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
sage:  %timeit pow(f, 2**64, g)
1.17 ms ± 10.7 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### NTL polynomials over extension fields

About 10x faster for NTL polynomial rings over $\textrm{GF}(p^k)$
Old timings:

```py
sage: set_random_seed(0)
....: 
....: p = 2**255 - 19
....: F.<a> = GF(p^3)
....: q = F.order()
....: R.<x> = F[]
....: 
[...]

PR #37443, opened 2024-02-23
Fix bugs and regression in performance of `any_root()`

### LINKED ISSUE #37359: Test for `elliptic_curves/hom.py` times out
### Steps To Reproduce

`sage -t --random-seed=212256703328484530299262546120222329602 src/sage/schemes/elliptic_curves/hom.py`

### Expected Behavior

The test should complete successfully.

### Actual Behavior

The test times out.

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Windows 11 (WSL Ubuntu 22.04)
- **Sage Version**: 10.3beta8
```

### Checklist

### LINKED ISSUE #37442: `any_root(ring=F_ext)` is underperforming
### Steps To Reproduce

In my refactoring of `any_root()` https://github.com/sagemath/sage/pull/37170 I accidentally introduced a performance regression when finding roots over extension fields.

Sage 10.2

```py
sage: %time _ = f.any_root()
CPU times: user 6.89 ms, sys: 321 µs, total: 7.21 ms
Wall time: 6.91 ms
sage: %time _ = f.any_root(L)
CPU times: user 7.34 ms, sys: 316 µs, total: 7.66 ms
Wall time: 7.42 ms
sage: %time _ = f.change_ring(L).any_root()
CPU times: user 3.13 s, sys: 9.92 ms, total: 3.14 s
Wall time: 3.15 s
```

SageMath version 10.3.beta8

```py
sage: %time _ = f.any_root()
CPU times: user 6.68 ms, sys: 116 µs, total: 6.8 ms
Wall time: 6.75 ms
sage: %time _ = f.any_root(L)
CPU times: user 844 ms, sys: 4.69 ms, total: 849 ms
Wall time: 849 ms
sage: %time _ = f.change_ring(L).any_root()
CPU times: user 819 ms, sys: 3.62 ms, total: 823 ms
Wall time: 823 ms
```

### Expected Behavior

When I refactor code, it should not have worse performance

### Actual Behavior

I have refactored code and introduced poor performance

### Additional Information

The fix will be to go back to `any_root()` and complicate the implementation to look for a root in the smallest extension before casting the root to `F_ext`. Currently, the implementation finds a root in `f.change_ring(F_ext).any_root()` which is very slow for very large extensions.

The new version should look for some `degree` degree distinct degree factorisation over various extensions working with the smallest one possible.

### Environment

```markdown
N/A
```

### Checklist

### LINKED ISSUE #37445: Bug in `any_root` introduced in #37170
### Steps To Reproduce

The trace computation in the helper method `_cantor_zassenhaus_split_to_irreducible` introduced by #37170 is implemented incorrectly. More specifically, these lines at https://github.com/sagemath/sage/blob/develop/src/sage/rings/polynomial/polynomial_element.pyx#L2176-L2177
```python
            if q % 2:
                h = self.gcd(pow(T, (q-1)//2, self) - 1)
```
should be
```python
            if q % 2:
                h = self.gcd(pow(T, (q**degree-1)//2, self) - 1)
```

Proof:
```sage
sage: R.<x> = GF(13)[]
....: def irr(d, R): return f.monic() if (f := R.random_element(d)).is_irreducible() else irr(d, R)
....: f = prod(irr(6, R) for _ in range(10))
....: f.factor()[0][0].degree()
....: f.any_root(degree=6) # errors
```

### Expected Behavior

don't error

### Actual Behavior

error

### Additional Information

We have a fix already

### Environment

```markdown
Latest `develop` branch
```

### Checklist

### LINKED ISSUE #37471: Incorrect assumptions about polynomial in `is_HCP()`
### Steps To Reproduce

The current code has the following snippet:

```py
        p = next_prime(p)
        n += 1
        fp = f.change_ring(GF(p))
        # Compute X^p-X mod fp
        z = fp.parent().gen()
        r = pow(z, p, fp) - z
        d = r.gcd(fp).degree()  # number of roots mod p
        if d == 0:
            continue
        if not fp.is_squarefree():
            continue
        if d < h and d not in h2list:
            return zero
        jp = fp.any_root(degree=1, assume_squarefree=True, assume_distinct_deg=True)
```

The problem is, `fp.any_root()` is called with `assume_distinct_deg` `True`. In this case, the polynomial is expected to be the product of irreducible polynomials of degree `degree` (in this case linear polynomials).

However, this is a typo. The polynomial is not of this form. The correct polynomial to use is the polynomial

```py
r = pow(z, p, fp) - z
r = r.gcd(fp)
```

As a result of this, tests such as:

```py
from sage.schemes.elliptic_curves.cm import is_HCP
set_random_seed(297388353221545796156853787333338705098)
is_HCP(hilbert_class_polynomial(-55))
```

Currently error.

This was introduced in the PR https://github.com/sagemath/sage/pull/37170 because in the refactoring `assume_distinct_deg` worked as intended where as the older function seemed to simply always find roots the slow way regardless of how the function was called.

### Expected Behavior

The function should work

### Actual Behavior

The function does not work

### Additional Information

This has been fixed in https://github.com/sagemath/sage/pull/37443. The only thing needed to do was to use the polynomial `r`, as intended, instead of the polynomial `fp`. This oversight lived in the code until recently purely because any_root() behaved weirdly before.

### Environment

```markdown
- **Sage Version**: SageMath version 10.3.beta8, Release Date: 2024-02-13
```

### Checklist

GiacomoPope:
This pull request aims to fix some issues which came with the refactoring of `any_root()` in https://github.com/sagemath/sage/pull/37170.

I am afraid that apart from a little more cleaning up and verbose comments, this "fix" really relies on `f.roots()` rather than `f.any_root()` when roots over extension fields are computed.

The core problem currently is that the proper workflow should be the following:

- Given a polynomial $f$, find an irreducible factor of smallest degree. When `ring` is `None` and `degree` is `None`, only linear factors are searched for. This code is unchanged and works fast.
- When `degree` is `None` and `ring` is not `None`, the old code used to perform `f.change_ring(ring).any_root()` and look for a linear factor in `ring`. The issue is when `ring` is a field with a large degree, arithmetic in this ring is very slow and root finding is very slow.
- When `degree` is not `None` then an irreducible of degree `degree` is searched for, $f$, and then a root is found in an extension ring. This is either the user supplied `ring`, or it is in the extension `self.base_ring().extension(degree)`. Again, as this extension could be large, this can be slow.

Now, the reason that there's a regression in speed, as explained in https://github.com/sagemath/sage/issues/37359, is that the method before refactoring simply called `f.roots(ring)[0]` when working for these extensions. As the root finding is always performed by a C library this is much faster than the python code which we have for `any_root()` and so the more "pure" method I wrote simply is slower.

The real issue holding this method back is that if we are in some field $GF(p^k)$ and `ring` is some extension $GF(p^{nk})$ then we are not guaranteed a coercion between these rings with the current coercion system. Furthermore, if we extend the base to some $GF(p^{dk})$ with $dk$ dividing $nk$ we also have no certainty of a coercion from this minimal extension to the user supplied ring.

As a result, a "good" fix is delayed until we figure out finite field coercion. **Issue to come**.

Because of all of this, this PR goes back to the old behaviour, while keeping the refactored and easier to read code. I hope one day in the future to fix this.

Fixes https://github.com/sagemath/sage/issues/37359
Fixes https://github.com/sagemath/sage/issues/37442
Fixes https://github.com/sagemath/sage/issues/37445
Fixes https://github.com/sagemath/sage/issues/37471

**EDIT**

I have labeled this as critical as the slow down in the most extreme cases causes code to complete hang when looking for a root. See for example: https://github.com/sagemath/sage/issues/37442. I do not want https://github.com/sagemath/sage/pull/37170 to be merged into 10.3 without this patch. 

Additionally, a typo meant that a bug was introduced in the CZ splitting. This has also been fixed.

Also, also in the refactoring the function behaved as intended (rather than the strange behaviour from before) which exposed an issue https://github.com/sagemath/sage/issues/37471 which I have fixed.


GiacomoPope:
Ultimately I believe for large extension fields it's simply faster to call f.roots(). The only way around this would be overhauling the coercion between finite fields and even then I'm not convinced it's faster. 

The question should then be, at what cut offs should "any_root" simply return the roots after factorisation handled by the underlying library (eg NTL or flint)?

GiacomoPope:
Comparison of new code to 10.2

```py
sage: set_random_seed(0)
....: K.<x> = GF((29, 3), names="a")[]
....: f = K.random_element(15).monic()
....: L.<a> = f.splitting_field()
....: %time _ = f.roots(L)
....: %time _ = f.any_root(L)
CPU times: user 90.3 ms, sys: 399 µs, total: 90.7 ms
Wall time: 90.7 ms
CPU times: user 11.1 ms, sys: 615 µs, total: 11.7 ms
Wall time: 12.8 ms
```


```py
│ SageMath version 10.3.beta8, Release Date: 2024-02-13              │
sage: set_random_seed(0)
....: K.<x> = GF((29, 3), names="a")[]
....: f = K.random_element(15).monic()
....: L.<a> = f.splitting_field()
....: %time _ = f.roots(L)
....: %time _ = f.any_root(L)
CPU times: user 81 ms, sys: 347 µs, total: 81.3 ms
Wall time: 81.4 ms
CPU times: user 7.63 ms, sys: 407 µs, total: 8.04 ms
Wall time: 8.92 ms
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2210,10 +2222,13 @@ cdef class Polynomial(CommutativePolynomial):
         of self and then finds a factor with Cantor-Zassenhaus
         splitting.
 
-        If degree is not None, then only irreducible factors of degree
-        `degree` are searched for, otherwise the smallest degree factor
+        If degree is not ``None``, then only irreducible factors of degree
```
grhkm21:
```suggestion
        If ``degree`` is not ``None``, then only irreducible factors of degree
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2210,10 +2222,13 @@ cdef class Polynomial(CommutativePolynomial):
...
+        ``degree`` are searched for, otherwise the smallest degree factor
         is found.
 
+        If ext_degree is not ``None``, then only irreducible factors whose
```
grhkm21:
```suggestion
        If ``ext_degree`` is not ``None``, then only irreducible factors whose
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2151,6 +2151,18 @@ cdef class Polynomial(CommutativePolynomial):
             True
             sage: f % factor == 0
             True
+
+        TESTS:
+
+        Ensure that :issue:`37445` is fixed::
+
+            sage: R.<x> = GF(13)[]
+            sage: def irr(d, R): return f.monic() if (f := R.random_element(d)).is_irreducible() else irr(d, R)
```
grhkm21, line 2159: Beautiful code! Great use of recursion! Nice syntax!

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2286,6 +2304,11 @@ cdef class Polynomial(CommutativePolynomial):
           this polynomial is assumed to be the product of irreducible
           polynomials of degree equal to ``degree``.
 
+        - ``ext_degree`` (None or positive integer) -- (default: ``None``).
+          Used for polynomials over finite fields. If not ``None`` only returns
+          irreducible factors of ``self`` whose degree divides ``ext_degree``.
+          Assumes that ``degree`` is ``None``.
```
grhkm21:
```suggestion
        - ``ext_degree`` -- positive integer or ``None`` (default);
          used for polynomials over finite fields. If not ``None`` only returns
          irreducible factors of ``self`` whose degree divides ``ext_degree``.
```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2560,22 +2583,69 @@ cdef class Polynomial(CommutativePolynomial):
...
+                try:
+                    f = self.change_ring(ring)
+                except ValueError:
+                    raise(f"cannot coerce polynomial {self} to the new ring: {ring}")
+                return f.any_root()
+
+            # If the ring is a finite field, ensure it's an extension of the base ring
+            if self.base_ring().characteristic() != ring.characteristic():
+                raise ValueError("ring must have the same characteristic as the base ring")
+            if not self.base_ring().degree().divides(ring.degree()):
```
grhkm21: This is not sufficient to ensure there's a coercion. Try `if *.coerce_map_from(*) is None`.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2617,13 +2687,17 @@ cdef class Polynomial(CommutativePolynomial):
...
+
+        # TODO: The proper thing to do here would be to call
+        #       return f.change_ring(ring).any_root()
+        #       but as we cannot work in the minimal extension (see above) working
+        #       in the extension for f.change_ring(ring).any_root() is almost always
+        #       much much much slower than using the call for roots() which uses
+        #       C library bindings for all finite fields.
+        #       Until the coercion system for finite fields works better,
+        #       this will be the most performant
+        return f.roots(ring, multiplicities=False)[0]
```
grhkm21, line 2706: Should this be randomised instead? Some users might be using `.any_root()` to get a(ny) random root.

grhkm21 (changes requested):
Small changes

src/sage/rings/polynomial/polynomial_element.pyx
GiacomoPope: lol

src/sage/rings/polynomial/polynomial_element.pyx
GiacomoPope: Done

src/sage/rings/polynomial/polynomial_element.pyx
GiacomoPope: Done

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2200,20 +2212,23 @@ cdef class Polynomial(CommutativePolynomial):
         # If you are reaching this error, chances are there's a bug in the code.
         raise AssertionError(f"no splitting of degree {degree} found for {self}")
 
-    def _any_irreducible_factor_squarefree(self, degree=None):
+    def _any_irreducible_factor_squarefree(self, degree=None, ext_degree=None):
         """
-        Helper function for any_irreducible_factor which computes
+        Helper function for ``any_irreducible_factor()`` which computes
```
grhkm21:
```suggestion
        Helper function for :meth:`any_irreducible_factor` which computes
```

grhkm21 (changes requested):
Sorry, I missed this

grhkm21 (approved):
Thanks for the quick response!

grhkm21 (changes requested):
Sorry, I had an oversight... I believe there's still problems with the method. The following code throws an `AssertionError`:

```sage
sage: R.<x> = GF(13)[]
....: f = 11*x^3 + 6*x^2 + 4*x + 12
....: f.any_root(GF(13^2))
```

An Error is *correct*, because `f` doesn't have a root over GF(13^2), but it should not be an `AssertionError`.

Another related problem is that the error takes very long to get to, because you run 1000 loops of the code above. It is probably related in the code logic and should be fixed?

GiacomoPope:
Hmm I'm worried about this. The loop over 1000 tries is the CZ splitting -- this shouldn't ever try 1000 times it should find a result in 1 or 2 tries so maybe there's an issue here when ring is not None?

GiacomoPope:
I think I don't have the best error messages for when ext_degree is not None. I'll handle this later.

grhkm21:
> Hmm I'm worried about this. The loop over 1000 tries is the CZ splitting -- this shouldn't ever try 1000 times it should find a result in 1 or 2 tries so maybe there's an issue here when ring is not None?

Yes I agree, I think it's a bug in implementation / code logic rather than "1000 times being slow"

GiacomoPope:
OK I believe I have better handled errors when `ext_degree` is not `None`.

GiacomoPope:
There are some strange failing doctests though... I will have to look into these.

GiacomoPope:
New commit makes `any_root()` deterministic again, because to allow it to be randomised would mean breaking almost every example of `any_root()` being used for number fields which breaks a huge amount of docstrings for elliptic curve stuff.

It's always been this way so leaving it this way I think is fine.

GiacomoPope:
Also to be explicit, rather then erroring with an `AssertionError` for the sample above we now have the more sensible:

```
sage: f.any_root(GF(13^2))
...
ValueError: polynomial 11*x^3 + 6*x^2 + 4*x + 12 has no irreducible factor of degree dividing 2
```

grhkm21:
I'll test it first :D

GiacomoPope:
Great! This is because I now error early if ext_degree is set and an irreducible polynomial of degree larger than ext_degree is found.

grhkm21 (approved):
LGTM, thanks for fixing it

grhkm21:
@GiacomoPope New bug just dropped :( Found in #37441 ([here](https://github.com/sagemath/sage/actions/runs/8033768660/job/21944594203)), not patched by this PR.

To reproduce:
```sage
./sage -t --random-seed=216579537934655877233640320811570547917 src/sage/schemes/elliptic_curves/cm.py
```

Sorry! 🥲

grhkm21:
I took a look at the code. I believe simply removing `assume_distinct_deg=True` from the call at https://github.com/sagemath/sage/blob/db3ecc42dd07df0b54761885ae10760df4974852/src/sage/schemes/elliptic_curves/cm.py#L274 is sufficient.

I also strongly suggest to rename the argument. It doesn't make sense to me that `assume_distinct_deg=True` means $f = \prod_i g_i$ with $\deg(g_i) = \deg(g_j)$. To me it seems to be the opposite... Naming it `assume_equal_deg=True` seems more appropriate.

Sorry for not spoting these problems during the original PR. In my defense, I was not as familiar with the algorithm 3 days ago compared to now.

yyyyx4:
> I also strongly suggest to rename the argument. It doesn't make sense to me that `assume_distinct_deg=True` means f=∏igi with deg⁡(gi)=deg⁡(gj). To me it seems to be the opposite... Naming it `assume_equal_deg=True` seems more appropriate.

I guess `assume_distinct_deg=` was supposed to mean that the input to this method is the result of having applied distinct-degree factorization first... But I agree that it doesn't really make sense.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2554,28 +2581,71 @@ cdef class Polynomial(CommutativePolynomial):
...
             try:
-                f = self.any_irreducible_factor(degree=1, assume_squarefree=assume_squarefree)
+                f = self.any_irreducible_factor(assume_squarefree=assume_squarefree, ext_degree=allowed_extension_degree)
             except ValueError:
-                raise ValueError(f"no root of polynomial {self} can be computed")
-
-            return - f[0] / f[1]
+                raise ValueError(f"no root of polynomial {self} can be computed over the ring {ring}")
+            # When d != 1 we then find the smallest extension
+            # TODO: This is really annoying. What we should do here is compute some minimal
```
tscrim:
```suggestion
            # TODO: What we should do here is compute some minimal
```
Let's be professional and not editorialize. `:)`

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2554,28 +2581,71 @@ cdef class Polynomial(CommutativePolynomial):
...
+            # TODO: The proper thing to do here would be to call
+            #       return f.change_ring(ring).any_root()
+            #       but as we cannot work in the minimal extension (see above) working
+            #       in the extension for f.change_ring(ring).any_root() is almost always
+            #       much much much slower than using the call for roots() which uses
+            #       C library bindings for all finite fields.
+            #       Until the coercion system for finite fields works better,
+            #       this will be the most performant
+            roots = f.roots(ring, multiplicities=False)
+            return choice(roots)
```
tscrim:
```suggestion
            return f.roots(ring, multiplicities=False)[0]
```
There is no need to make this random.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2617,13 +2687,18 @@ cdef class Polynomial(CommutativePolynomial):
...
+        # TODO: The proper thing to do here would be to call
+        #       return f.change_ring(ring).any_root()
+        #       but as we cannot work in the minimal extension (see above) working
+        #       in the extension for f.change_ring(ring).any_root() is almost always
+        #       much much much slower than using the call for roots() which uses
+        #       C library bindings for all finite fields.
+        #       Until the coercion system for finite fields works better,
+        #       this will be the most performant
+        roots = f.roots(ring, multiplicities=False)
+        return choice(roots)
```
tscrim:
```suggestion
        return f.roots(ring, multiplicities=False)[0]

```

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2554,28 +2581,71 @@ cdef class Polynomial(CommutativePolynomial):
...
-                return rs[0]
-            raise ValueError(f"polynomial {self} has no roots")
+        if self.base_ring() not in FiniteFields():
+            if ring not in FiniteFields():
+                rs = self.roots(ring=ring, multiplicities=False)
+                if rs:
+                    # TODO: this has been deterministic and changing this to be random
+                    # breaks many doctests. For now we leave it deterministic but you
+                    # could change it to choice(rs). This mainly breaks examples over
+                    # of elliptic curve stuff over number fields
```
tscrim:
```suggestion
```
Deterministic is correct and good.

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -133,7 +133,7 @@ from sage.categories.morphism cimport Morphism
 
 from sage.misc.superseded import deprecation_cython as deprecation, deprecated_function_alias
 from sage.misc.cachefunc import cached_method
-
+from sage.misc.prandom import choice
```
tscrim:
```suggestion
```

tscrim (changes requested):
`any_root()` should be deterministic. Random output for things makes debugging much harder and could have very subtle differences that lead to major performance changes (e.g., from using cached output). If it was supposed to be a random root, then it would be called `random_root()`. So please remove all random outputs here.

yyyyx4:
> `any_root()` should be deterministic. Random output for things makes debugging much harder and could have very subtle differences that lead to major performance changes (e.g., from using cached output). If it was supposed to be a random root, then it would be called `random_root()`. So please remove all random outputs here.

No, `.random_root()` would be stronger: The name suggests that the returned root is sampled from some well-behaved distribution (ideally, uniformly among all roots). Historically `.any_root()` never made any promises about the distribution of the output as far as I can tell, and nothing about its name or documentation suggested deterministic behaviour. The fact that it was previously deterministic and that this assumption was baked into tons of other library code is essentially an accident. All three variants (any root, deterministic root, random root) have their respective use cases.

GiacomoPope:
I have no time today to work on this as I'm with family. 

1. `any_root()` uses Cantor-Zassenhaus for the splitting which IS random. For cases where there exists more than one linear factor the returned root was ALWAYS random. My PR didn't change this.

The issue was the original implementation was lazy for the case of needing extension degrees which fell back to picking the first root by fully factoring (rather than finding a single root). By using choice(rts[0]) at least the extension case is random just like the case for linear roots. 

If the extension degree coercion wasn't so broken, `any_root()` would never need to call `roots()` for performance and the whole algorithm would naturally be random. This awful compromise at the moment is simply because of Sage limitations. 


2. The assume_distinct_degree can be renamed. This comes from the previously HORRIBLE design choice that if the degree parameter was set to be negative then the algorithm assumed that the input polynomial was of the form Prod g_i(x) where all polynomials had degree `degree`. I see that there's a new bug and I'll look into it on Monday

tscrim:
You're right that forcing this to be deterministic is not desirable as it is an independent this. However, I don't agree with introducing randomness just for the sake of matching some other case. Deterministic algorithms and/or outputs are better than random ones IMO (under equal time/memory requirements).

@GiacomoPope Please be kinder to other's work (just for the record, I am not involved with that implementation). You can think it is a bad design choice and point that out, but the all caps is a bit over the top. ;)

However, no problem and no rush. This is the weekend.

GiacomoPope:
OK, sorry -- I don't mean to be unkind, I have just spent a lot of time debugging the original function and I'm tired / frustrated. I should not have used the language above.

I want to include randomness as I currently do, as previously many doctests relied on the determinism of this function which was undocumented. If I can fix the coercion system then the whole thing will use `any_root()` and naturally be random and this will stop needing to fix doctests in the future.

Does this make sense?

tscrim:
> OK, sorry -- I don't mean to be unkind, I have just spent a lot of time debugging the original function and I'm tired / frustrated. I should not have used the language above.

No problem. I know you didn't and I certainly can understand the frustration. It was fine, just I felt the all caps was just a little bit too much. 

> I want to include randomness as I currently do, as previously many doctests relied on the determinism of this function which was undocumented. If I can fix the coercion system then the whole thing will use `any_root()` and naturally be random and this will stop needing to fix doctests in the future.
> 
> Does this make sense?

I don't see that as a problem; in fact, I see that as a positive as it makes debugging when the do fail much easier. If it does end up being random, then it is better to have those doctest changes at that time as it makes it easier to figure out if a change introduced a problem. Also would you have a different viewpoint if someone just documented which cases were deterministic and random?

Perhaps as a contrasting case, consider `an_element()`. Would you advocate this to return an element at random in general? (This isn't exactly an equal comparison as `an_element()` is used by the coercion framework as a tester element, so it needs to be as fast as possible.) How about `some_elements()`?

Feel free to take the rest of today and spend time with your family. We are used to not having immediate responses with Sage development.

GiacomoPope:
RE: the failing doctest, it seems the real problem is this assumption on the factorisation of the polynomial.

I made this dummy function:

```py
def is_HCP_debug(f):
    zero = ZZ(0)
    # optional check that input is monic and irreducible
    if check_monic_irreducible:
        try:
            if not (all(c in ZZ for c in f) and f.is_monic()):
                return zero
            f = f.change_ring(ZZ)
        except AttributeError:
            return zero

    from sage.rings.real_mpfr import RR
    from sage.rings.finite_rings.finite_field_constructor import GF

    h = f.degree()
    h2list = [d for d in h.divisors()
              if (d-h) % 2 == 0 and d.prime_to_m_part(2) == 1]
    pmin = 33 * (h**2 * (RR(h+2).log().log()+2)**2).ceil()
    # Guarantees 4*p > |D| for fundamental D under GRH
    p = pmin-1
    n = 0
    from sage.arith.misc import next_prime
    from sage.schemes.elliptic_curves.constructor import EllipticCurve

    while True:
        p = next_prime(p)
        n += 1
        fp = f.change_ring(GF(p))
        # Compute X^p-X mod fp
        z = fp.parent().gen()
        r = pow(z, p, fp) - z
        d = r.gcd(fp).degree()  # number of roots mod p
        if d == 0:
            continue
        if not fp.is_squarefree():
            continue
        if d < h and d not in h2list:
            return zero
        return fp
```

Which allows me to look at the polynomial for which we want a root.

```py
sage: set_random_seed(297388353221545796156853787333338705098)
sage: f = is_HCP(hilbert_class_polynomial(-55))
sage: f.factor()
(x + 27) * (x + 582) * (x^2 + 952*x + 129)
sage: 
sage: set_random_seed(0)
sage: f = is_HCP_debug(hilbert_class_polynomial(-55))
sage: f.factor()
(x + 27) * (x + 582) * (x^2 + 952*x + 129)
sage: 
sage: set_random_seed(297388353221545796156853787333338705098)
sage: f = is_HCP_debug(hilbert_class_polynomial(-55))
sage: f.factor()
(x + 27) * (x + 582) * (x^2 + 952*x + 129)
```

So we see that this doesnt have only linear factors and so `assume_distinct_deg=False` should be set.

The question then is why does this work for 10.2??

The original code was:

```py
jp = fp.any_root(degree=-1, assume_squarefree=True)
```

looking at the old documentation:


```py
   * "degree" (None or nonzero integer) -- Used for polynomials over
     finite fields.  Return a root of degree "abs(degree)" over the
     ground field.  If negative, also assumes that all factors of this
     polynomial are of degree "abs(degree)".  If None, returns a root
     of minimal degree contained within the given ring.
```

As we see, they use `degree = -1` so they assume everything has linear factors yet the old code still works. 

I cannot determine why this code works even when it shouldn't??? But I would say this is not a bug in `any_root()` now, but rather an old bug in `any_root()` meant this worked even when it shouldn't??

@grhkm21 what do you think?

GiacomoPope:
In response to Travis, OK I will fix the determinism and include in the docstring a .. NOTE:: about the deterministic aspects of the implementation. Then if this ever changes people can deal with it accordingly. This is fine.

My motivation previously was mathematical (CZ is random) but I can see that while we use the factorisation workaround for performance we may as well do determinism.

tscrim:
@GiacomoPope Thank you. I appreciate it.

I can't comment on the other aspects of the PR and will leave this to others to finish the review. If nobody finishes it by next week, I can likely finalize it.

GiacomoPope:
Oh! Looking at the HCP code is it just buggy? It looks like they should take the roots of r instead of the roots of fp?

GiacomoPope:
```
sage -t --random-seed=216579537934655877233640320811570547917 src/sage/schemes/elliptic_curves/cm.py
    [89 tests, 3.55 s]
All tests passed!
Total time for all tests: 3.7 seconds
    cpu time: 2.8 seconds
    cumulative wall time: 3.5 seconds
Features detected for doctesting: sage.libs.flint,sage.libs.pari,sage.rings.number_field
pytest is not installed in the venv, skip checking tests that rely on it
```

I have a fix locally. I will make an issue explaning the bug and do all the work after my daughter is asleep tonight :)

src/sage/rings/polynomial/polynomial_element.pyx
grhkm21: Yeah I disagree with this, as Giacomo pointed out the CZ method is inherently randomised anyways (IIRC there's no competitively-fast deterministic root finding method).

grhkm21:
Hi it seems you found the bug, root cause of the bug, congrats :D Thanks for the incredibly hard work

src/sage/rings/polynomial/polynomial_element.pyx
yyyyx4: Well, the fact that a randomized algorithm is best for some cases of `.any_root()` does not imply that we should actively randomize the cases that are currently taking other code paths, too — and I agree that it's a bit arbitrary (or *random*) to do so here. As I argued in another comment, in my book it's fine for `.any_root()` to be randomized or deterministic or a little bit of both depending on the input. Eventually, forcing deterministic behaviour should probably be supported using a keyword argument, while the default would be to run the fastest algorithm that finds *any root* as the name suggests, but that's for later.

src/sage/schemes/elliptic_curves/cm.py
```diff
@@ -233,6 +233,13 @@ def is_HCP(f, check_monic_irreducible=True):
         sage: all(not is_HCP(hilbert_class_polynomial(D) + 1)
         ....:     for D in srange(-4,-100,-1) if D.is_discriminant())
         True
+
+    Ensure that :issue:`37471` is fixed::
+
+        from sage.schemes.elliptic_curves.cm import is_HCP
+        set_random_seed(297388353221545796156853787333338705098)
+        is_HCP(hilbert_class_polynomial(-55))
+        -55
```
grhkm21:
```suggestion
        sage: from sage.schemes.elliptic_curves.cm import is_HCP
        sage: set_random_seed(297388353221545796156853787333338705098)
        sage: is_HCP(hilbert_class_polynomial(-55))
        -55
```

grhkm21 (changes requested):

src/sage/schemes/elliptic_curves/cm.py
grhkm21: (Also the commit message links to the wrong issue, but it doesn't matter)

src/sage/rings/polynomial/polynomial_element.pyx
```diff
@@ -2286,6 +2305,18 @@ cdef class Polynomial(CommutativePolynomial):
...
           polynomials of degree equal to ``degree``.
 
+        - ``ext_degree`` -- positive integer or ``None`` (default);
+          used for polynomials over finite fields. If not ``None`` only returns
+          irreducible factors of ``self`` whose degree divides ``ext_degree``.
+
+        .. NOTE::
+
+            Any root is non-deterministic when finding linear roots of a polynomial
+            over the base ring. However, if ``degree`` is greater than one, or `ring`
```
grhkm21:
```suggestion
            over the base ring. However, if ``degree`` is greater than one, or ``ring``
```
Also it should be wrapped to 80 characters.

GiacomoPope:
OK -- I hopefully have totally fixed the bug which lived inside `cm.py` as well as addressed all the determinism / small comments from Travis. Fingers crossed this has no more bugs lmao.

grhkm21 (changes requested):
final one then it should be ok unless Travis has more comments

(I accidentally submitted the previous review as a "single comment")

grhkm21 (approved):
Wait for CI then ok

GiacomoPope:
Forgot about the suggestion to rename the optional argument, so I did that just now. @grhkm21

GiacomoPope:
@grhkm21 the weird timeout error happened again, but this time in a new branch?? 

https://github.com/sagemath/sage/actions/runs/8041059303/job/21977243528?pr=37443

GiacomoPope:
Failure for 3.9 is unrelated to this PR, identified in https://github.com/sagemath/sage/issues/37488 and fixed in https://github.com/sagemath/sage/pull/37489

grhkm21:
Yeah idk I can't reproduce it anywhere either.

tscrim (commented):
It is okay with me, although I think your `.. NOTE::` about the determinism is only applicable for finite fields and/or with some additional arguments. It would be nice to clarify this, but not required.

GiacomoPope:
Thanks for taking a look. I have updated the NOTE to be more explicit about the case of non-finite fields.

PR #37446, opened 2024-02-24
Add dependency groups to `pyproject.toml`

tobiasdiez:
Following [PEP 735](https://peps.python.org/pep-0735/#canonical-names-of-dependencies-and-dev-el-split-packages), add a few groups of dependencies that one usually needs while developing sage (e.g. for tests or docs).

This info will then be used to generate the conda lock files and once support for PEP 735 is more widespread other tools will be able to use the metadata as well. Eg, once https://github.com/pypa/pip/issues/12963 is implemented one can install all test dependencies with `pip install --groups test`.

### :memo: Checklist

- [ ] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37447, opened 2024-02-24
Replace bootstrap-conda by grayskull and update conda lock files

### LINKED ISSUE #34626: Minimize conda environment
<div id="comment:0"></div>

The dev-conda environment contains quite a lot of packages that are not directly needed to compile or run sage, but only to compile one of sage's dependencies. Since conda already takes care of building the dependency tree (and only including runtime dependencies in it), there is no need to include the deep-dependencies in the environment file. Doing so just unnecessary bloats up the install time and size.

In fact, `du -hs /opt/conda/envs/**` shows

```
6.5G    /opt/conda/envs/sage-dev
1.4G    /opt/conda/envs/sage-dev-min
```
that the sage-dev environment is very large and there is a lot of room for improvement. The second env in the above list corresponds to the following env:

```
name: sage-dev-min
channels:
  - conda-forge
  - nodefaults
dependencies:
  - tar
  - cypari2 # sagelib/dep
  - singular # sagelib/dep
  - gap-defaults # sagelib/dep
# - html5lib # not referenced in src/
  - libgd # sagelib/dep
  - sagemath-db-polytopes
  - gengetopt
  - symmetrica # sagelib/dep
  - palp
# - sphinxcontrib-websupport # not referenced in src/
  - pkgconfig # sagelib/dep
  - git # Additional dev tool
# - pybind11 # not referenced in src/
  - cliquer # sagelib/dep
  - libatomic_ops
  - pari-elldata
  - zn_poly # sagelib/dep
  - pycodestyle # Additional dev tool
  - pari-galpol
  - gfan # features/gfan
  - memory-allocator # sagelib/dep
  - prompt_toolkit
  - eclib # sagelib/dep
  - arb # sagelib/dep
  - tachyon # sagemath_doc_html/dep
  - bc
  - threejs-sage
  - lcalc # sagelib/dep
  - sagemath-db-graphs
# - ninja # not referenced in src/
  - backports.zoneinfo
# - cppy # not referenced in src/
  - networkx
  - jupyter-jsmol
  - pcre
  - sympy
# - appnope # not referenced in src/
  - flintqs # interfaces/qsieve but not declared as dep
  - r
  - matplotlib
  - r-essentials
  - cvxopt # sagemath-standard/req
  - libhomfly # sagelib/dep
  - pari-seadata
# - poetry-core # not referenced in src/
  - linbox # sagelib/dep
# - importlib_metadata # mentioned in the docs but not used?
  - scipy # sagemath-standard/req
  - patch
# - qhull # not referenced in src/
  - jupyter-packaging
  - jupyter_sphinx # sagemath_doc_html/dependencies
  - compilers
  - esbonio # Additional dev tool
  - openssh # Additional dev tool
  - giac # sagelib/dep
  - fpylll # sagemath-standard/req
  - pplpy # sagelib/dep
# - hatchling # not referenced in src/
  - sagemath-db-elliptic-curves
  - jmol # sagemath_doc_html/dependencies
  - tox # additional dev tool but not declared as such
# - sagetex # not actively referenced in src/
  - primecountpy # sagelib/dep
  - furo # sagemath_doc_html/dependencies
  - rpy2 # sagemath-standard/req
# - pythran # not referenced in src/
  - planarity # sagelib/dep
  - python-lrcalc # sagelib/dep
# - vcversioner # not referenced in src/
  - rw # sagelib/dep
  - brial # sagelib/dep
  - maxima
  - sympow # sage/lfunctions/sympow.py but not declared as dep
  - automake # Package needed for ./bootstrap
  - sagemath-db-combinatorial-designs
# - importlib-resources # not referenced in src/
# - cmake # only mentioned in docs as not often used
# - setuptools_scm # not referenced in src/
  - boost-cpp # sagelib/dep
  - blas # sagelib/dep
  - libbraiding # sagelib/dep
  - mathjax # sagemath_doc_html/dependencies
  - sagemath-db-conway-polynomials
```
This env has been created using https://github.com/conda-incubator/conda-tree minimizing function (which only takes the leaves of the dependency tree) and then manually going through the remaining dependencies to see which ones are not needed by sage (see comments). Sage compiles within this env, but I've not yet run the doctests. I propose to include this env file directly, replacing the existing automatically generated `sage-dev`.

Comments about whether the packages without any comments are used/needed are welcome.

---

To see which packages are still rather big, use

```
grep '"size":' ${CONDA_PREFIX}/conda-meta/*.json | sort -k3rn | sed 's/.*conda-meta\///g' | column -t
```

CC:  @isuruf @dimpase @saraedum @mkoeppe @slel

Component: **build**

Author: **Tobias Diez**

_Issue created by migration from https://trac.sagemath.org/ticket/34626_

tobiasdiez:
Using the metadata added in https://github.com/sagemath/sage/pull/37446, we automatically generate the conda environment files. This no longer makes any reference to the `conda.txt` files contained in sage-the-distribution. Thus sage-on-top-of-conda is now completely independent of sage-the-distribution (only relying on information specified by sage-the-library). In particular, after this PR is merged the `conda.txt` files could be deleted from sage-the-distribution.

In particular, we no longer need to maintain the mapping of pypi packages to conda packages but instead can rely on the offical mappings maintained by the conda team (https://github.com/regro/cf-graph-countyfair/tree/master/mappings/pypi).

To test:
```
pip install grayskull conda-lock
python tools/update-conda.py
```

The updated conda files are committed here as well.

Moreover, the `environment-dev` files have been deleted as there was only little difference between these files and some people rightfully complained that having too many files in the root folder is distracting.

As a byproduct, this fixes #34626.

### :memo: Checklist

### :hourglass: Dependencies

- https://github.com/sagemath/sage/pull/37446: specifies additional requirements in the pyproject.toml which are used to generate the conda env files
- https://github.com/sagemath/sage/pull/38983: for the update of numpy
- #38982: to fix meson build


dimpase:
I've removed `disputed` as it was added without a word of explanation.

dimpase:
do you mean `pip install git+https://github.com/conda/grayskull` ?

Anyhow, after I do the latter, I get
```
$ python tools/update-conda.py .
Conda environment file written to src/environment-3.9.yml
Updating lock file for src/environment-3.9.yml at src/environment-3.9-linux
  File "/mnt/opt/Sage/sage-clang/tools/update-conda.py", line 150, in <module>
    update_conda(Path(options.sourcedir))
  File "/mnt/opt/Sage/sage-clang/tools/update-conda.py", line 75, in update_conda
    subprocess.run(
  File "/usr/lib/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1950, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'conda-lock'
```

What do I do wrong? Does it need to be run in a conda env?

tobiasdiez:
> What do I do wrong? Does it need to be run in a conda env?

Sorry, forgot that the script uses conda-lock as well. Please try again after `pip install conda-lock`. Later today, I'll add a bit of documentation to the script.

dimpase:
Why do I see on an amd64 box 
```
Locking dependencies for ['linux-aarch64']...
```
while running `python tools/update-conda.py .` ? Looks like an overkill, no?

dimpase:
I was getting errors about unresolved packages,so I applied
```diff
diff --git a/tools/update-conda.py b/tools/update-conda.py
index f123c7e4151..176b016d183 100644
--- a/tools/update-conda.py
+++ b/tools/update-conda.py
@@ -20,10 +20,6 @@ options = parser.parse_args()
 
 platforms = {
     "linux-64": "linux",
-    "linux-aarch64": "linux-aarch64",
-    "osx-64": "macos",
-    "osx-arm64": "macos-arm64",
-    "win-64": "win",
 }
 pythons = ["3.9", "3.10", "3.11"]
 tags = ["", "-dev"]
```
and this worked. Obv. `win-64` won't fly, many Sage packages don't exist there.

dimpase (changes requested):
please remove win-*-stuff

tobiasdiez:
> Why do I see on an amd64 box
> 
> ```
> Locking dependencies for ['linux-aarch64']...
> ```
> 
> while running `python tools/update-conda.py .` ? Looks like an overkill, no?

Note that the `update-conda` script is not run at bootstrap time (there is no point in doing this as the env files are already in the src folder under version control) but only for maintenance, i.e to update the lock files for all systems. My plan is to automate the procedure and have a GH scheduled workflow that runs the script and opens a PR every two weeks or so. This is btw exactly the same setup as https://github.com/scikit-learn/scikit-learn/blob/main/build_tools/update_environments_and_lock_files.py.

> please remove win-*-stuff

Good catch. Thanks! I've played around with the windows packages for a bit, but then forgot to uncomment it again. Done now.

saraedum:
Please clarify what about this PR is *disputed*. (@tobiasdiez is it possible that @mkoeppe is lacking the permission to comment here?)

saraedum:
I took the liberty to add the "needs work" label since it's unclear what's disputed here. (Also, you could maybe check or remove the checklist at the top of the PR to make it clear that there's nothing from that checklist missing here.)

dimpase:
@saraedum what's disputed here by Mkoeppe (who does block me on GH) is that he wants to do everything sage-distro, and Sage i  general-related His Way (TM) and doesn't want to even try to get any consensus if people have different views.

mkoeppe:
@saraedum @roed314 @jhpalmieri I set "disputed" here because when the PR was opened, I was not able to comment because Tobias had blocked me without explanation or giving notice to the sage-abuse committee. It appears that now I can comment, so I am removing "disputed".

There has not been any meaningful review yet here, so conducting a vote would seem premature.

mkoeppe:
It is correct that on my own PRs, @dimpase is blocked because of his persistent and aggressive code of conduct violations, including the type of denunciations that he has just written above.

mkoeppe:
In any case, this PR here is based on https://github.com/sagemath/sage/pull/37446, which is in disputed status because of https://github.com/sagemath/sage/pull/37446#issuecomment-1996711954

dimpase:
@mkoeppe - how come you comment here, despite being blocked by Tobias? Are you already installed as (nonB)DFL?

dimpase (approved):
I am giving it a positive review;  I tried it out, I pointed a problem, the problem got fixed - despite a rather abusive "There has not been any meaningful review yet here, so conducting a vote would seem premature" comment.

tools/update-conda.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python3
```
saraedum, line 1: I think it would be good to add an explanation at the top that explains what this script does. Or maybe add a `--help` to the argument parser that prints such documentation.

tools/update-conda.py
```diff
@@ -0,0 +1,154 @@
...
+    all_requirements.append("blas=2.*=openblas")
+    all_requirements.remove("{{ compiler('c') }}")
+    all_requirements.append("c-compiler")
+    all_requirements.remove("{{ compiler('cxx') }}")
+    all_requirements.append("cxx-compiler")
+
+    # Correct pypi name for some packages
+    python_requirements = pyproject_metadata.get("install_requires", [])
+    # Specify concrete packages for some packages not yet in grayskull
+    # TODO: It seems to be a bug that these external.dependencies are added to install_requires by grayskull
```
saraedum: Could we reference an upstream issue if we believe that's a bug?

tools/update-conda.py
```diff
@@ -0,0 +1,154 @@
...
+                            "--lockfile",
+                            str(lock_file),
+                            "--filename-template",
+                            str(lock_file),
+                        ]
+                    )
+
+
+def get_dependencies(pyproject_toml: Path) -> list[str]:
+    graystull_config = Configuration("sagemath")
```
saraedum: Is this intentional or a typo?

tools/update-conda.py
```diff
@@ -0,0 +1,154 @@
...
+def get_dev_dependencies(pyproject_toml: Path) -> list[str]:
+    pyproject = tomllib.load(pyproject_toml)
+    dependency_groups = pyproject["dependency-groups"]
+    dev_dependencies = dependency_groups.get("test", []) + dependency_groups.get(
+        "docs", []
+    )
+    return dev_dependencies
+
+
+update_conda(Path(options.sourcedir))
```
saraedum: Not super important but if we put all the global bits of this script in a `def main()` and do the usual `if __name__ == "__main__"` or similar, then this could be imported by others to use bits of the machinery for other purposes.

saraedum (commented):
I understand that there are reservations about having the `pyproject.toml` in the root directory. Otherwise, this PR does seem like a reasonable idea to me (replace a shell script with Python and relay some of the heavy lifting to an external project.)

Is there a better place for the pyproject.toml that would make this work for everybody?

mkoeppe:
Creating a new top-level `tools/` directory should probably go through discussion in sage-devel.

And actually, there's no need for a new top-level directory: We already have a place for scripts of exactly this kind -- it is `SAGE_ROOT/build/bin/`.
In other words, the `tools/update-conda.py` script should just become part of the sage-bootstrap distribution package.

saraedum:
> Creating a new top-level `tools/` directory should probably go through discussion in sage-devel.

I agree, that script should probably just go elsewhere. But otherwise, if we somehow move the pyproject.toml file, I don't see a problem here.

mkoeppe:
> otherwise, if we somehow move the pyproject.toml file, I don't see a problem here.

Same here. This could just be done on top of #37482.

tools/update-conda.py
tobiasdiez: No of course not ;-) Thanks for catching it!

tools/update-conda.py
tobiasdiez: I forgot why I thought this would be a bug. I've removed the comment now.

dimpase:
What's happening here?

tobiasdiez:
> What's happening here?

It still needs a bit of refinement, but is mostly working and I've used it to create https://github.com/sagemath/sage/pull/38878. I'll continue here after https://github.com/sagemath/sage/pull/38728 is merged.

In the meantime, it would be helpful if you (or someone else) could review https://github.com/sagemath/sage/pull/37446

tools/update-conda.py
tobiasdiez: I've now added a bit of documentation in the `README` file in the same directory, and a very short comment on the usage in this file. Do you think this is sufficient?

tobiasdiez:
Applying the label "ci-fix" to fix the issues reported at https://groups.google.com/g/sage-release/c/n0MODfiPcWE (for some reason I cannot reply there)

This is now also ready for review.

tobiasdiez:
If you don't mind adding grayskull and its dependencies to your main sage env, you can just use this one. Alternatively, create a new venv and run it there.

It should also be `pip install grayskull conda-lock`, will update the description above.

dimpase:
I am not sure I follow - I normally don't use conda.

dimpase:
> It should also be `pip install grayskull conda-lock`, will update the description above.

for  some reason I also had to add `toml`, not only conda-lock.

What I don't understand is why running `python tools/update-conda.py` leads to all these checked-in files `env*.yml` getting modified. Are they meant to be checked in afterwards? Or, rather, I'd prefer them not being tracked by git at all - what's the point? If you create a package to be distributed then you can add them there, otherwise it's better to keep them out, no?

tobiasdiez:
> I am not sure I follow - I normally don't use conda.

The script can be run from any python venv, not necessarily from a conda venv. It updates the lock files for conda, which one can use to create a conda env.

> for some reason I also had to add `toml`, not only conda-lock.

Thanks, I've added it to tools/readme.

> What I don't understand is why running `python tools/update-conda.py` leads to all these checked-in files `env*.yml` getting modified. Are they meant to be checked in afterwards? Or, rather, I'd prefer them not being tracked by git at all - what's the point? If you create a package to be distributed then you can add them there, otherwise it's better to keep them out, no?

These lock files should indeed be checked in. They serve as the ground-truth of which dependency versions are working correctly. Otherwise the conda ci workflows would constantly fail due to some random package update in the conda universe. From time to time, I'll run the `update-conda` script of this PR to update these environments.

tobiasdiez:
Updating the macos packages results in gd no longer being found. I have no idea why, so let's postpone the update of the macos py 11 (the one used in the ci) for now. The ci is failing, but that was already the case also before this PR. Would be awesome if someone on Macos has the time to look into this issue

dimpase:
> results in gd no longer being found

Do you mean a conda package gd ? A Sage package gd? I'm not sure. If you can provide details of what to check on macOS, I can do this.

dimpase:
> `pip install grayskull conda-lock`
please replace this, in the PR description, with 
```
pip install grayskull conda-lock toml
```

dimpase:
I tried `python tools/update-conda.py` in a "normal" Python venv created with Homebrew Python 3.12, on aarch64/arm64 macOS, after `git clean -fdx`, and it finished without errors. Specifically for `libgd` I get
```
% git grep libgd envir*
environment-3.10-linux-aarch64.yml:  - libgd=2.3.3=h6818b27_10
environment-3.10-linux.yml:  - libgd=2.3.3=hd3e95f3_10
environment-3.10-macos-x86_64.yml:  - libgd=2.3.3=h2e77e4f_10
environment-3.10-macos.yml:  - libgd=2.3.3=hac1b3a8_10
environment-3.11-linux-aarch64.yml:  - libgd=2.3.3=h6818b27_10
environment-3.11-linux.yml:  - libgd=2.3.3=hd3e95f3_10
environment-3.11-macos-x86_64.yml:  - libgd=2.3.3=h2e77e4f_10
environment-3.11-macos.yml:  - libgd=2.3.3=hac1b3a8_10
environment-3.9-linux-aarch64.yml:  - libgd=2.3.3=h6818b27_10
environment-3.9-linux.yml:  - libgd=2.3.3=hd3e95f3_10
environment-3.9-macos-x86_64.yml:  - libgd=2.3.3=h2e77e4f_10
environment-3.9-macos.yml:  - libgd=2.3.3=hac1b3a8_10
```

tobiasdiez:
> > results in gd no longer being found
> 
> Do you mean a conda package gd ? A Sage package gd? I'm not sure. If you can provide details of what to check on macOS, I can do this.

The conda package is installed, but when trying to install sage it fails to detect libgd (at least on ci):
```
Building wheels for collected packages: sagemath-standard
  Created temporary directory: /private/var/folders/5q/zcc7tkt11f13mhczs1yfd4qc0000gn/T/pip-wheel-j43hu9a5
  Destination directory: /private/var/folders/5q/zcc7tkt11f13mhczs1yfd4qc0000gn/T/pip-wheel-j43hu9a5
  Building editable for sagemath-standard (pyproject.toml): started
  Running command Building editable for sagemath-standard (pyproject.toml)
  Generating interpreters for fast_callable in /Users/runner/work/sage/sage/src/sage/ext/interpreters
  -> First build of interpreters
  ************************************************************************
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
      main()
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 273, in build_editable
      return hook(wheel_directory, config_settings, metadata_directory)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/setuptools/build_meta.py", line 476, in build_editable
      return self._build_with_temp_dir(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/setuptools/build_meta.py", line 407, in _build_with_temp_dir
      self.run_setup()
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/setuptools/build_meta.py", line 320, in run_setup
      exec(code, locals())
    File "<string>", line 97, in <module>
    File "/Users/runner/work/sage/sage/src/sage/env.py", line 448, in cython_aliases
      aliases[var + "CFLAGS"] = pkgconfig.cflags(lib).split()
                                ^^^^^^^^^^^^^^^^^^^^^
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/pkgconfig/pkgconfig.py", line 144, in cflags
      _raise_if_not_exists(package)
    File "/Users/runner/miniconda3/envs/sage/lib/python3.11/site-packages/pkgconfig/pkgconfig.py", line 103, in _raise_if_not_exists
      raise PackageNotFoundError(package)
  pkgconfig.pkgconfig.PackageNotFoundError: gdlib not found
```
https://github.com/sagemath/sage/actions/runs/12024995797/job/33521404350

So it would be great if you could indeed test this locally as follows:
- Checkout this PR
- Create a conda env based on the existing, functional lock file: `mamba env create --file src/environment-3.11-macos.yml --name sage-dev-old`
- Update the conda lock files as you already did by running `python tools/update-conda.py` (only the macos py 3.11 should have major changes at this point)
- Run the following:
```bash
mamba env create --file src/environment-3.11-macos.yml --name sage-dev
conda activate sage-dev
./bootstrap
pip install --no-build-isolation --no-deps --config-settings editable_mode=compat -v -v -e ./src
```

Does this work, or result in the same error as above? If not, try to find out what's the difference between `sage-dev-old` and `sage-dev` that make it fail. I look at the files used to install `gd`, and they don't seem to have any differences in the `pkg-config` file. Judging by https://github.com/sagemath/sage/actions/runs/11880394940/job/33103497734?pr=38879, meson doesn't have any issues finding `gd` in this conda env (but it's also not using pkg-config). You can see where the conda env files are installed by running `conda env list`.

dimpase:
> These lock files should indeed be checked in

as this is run as a part of `./bootstrap`, it would be natural for the lock files to be a part  of configure spkg -- which is used to keep autotools-generated `./configure` etc -- rather than these to be checked in the repo in directly

dimpase:
> * `mamba env create --file src/environment-...`
do we expect to `environment-...` to be in `src/`, rather than in the root directory of the Sage install?

dimpase:
I was able to reproduce the macOS error. Digging further showed that `pkg-config` was broken, to fix it I had to run `mamba install expat`. After the latter, the build did not show any `gdlib` error any more. Building further now.

dimpase:
PS. With the above addition of `expat`, the build succeeds, and the resulting `./sage` starts. 

However, things are missing, e.g. there is no `gap`/`libgap`. This is remedied by `mamba install gap`.  But after this,
re-running  `pip install --no-build-isolation --no-deps --config-settings editable_mode=compat -v -v -e ./src` fails with a cryptic message about `setuptools`. Aren't we already supposed to be using `meson` instead?

tobiasdiez:
> I was able to reproduce the macOS error. Digging further showed that `pkg-config` was broken, to fix it I had to run `mamba install expat`. After the latter, the build did not show any `gdlib` error any more. Building further now.
> 
> This looks like a conda bug to me, in fact. @saraedum ?

Awesome! Thanks a lot, after adding `expat` (and fortran) the tests are finally green! Yeah.

> However, things are missing, e.g. there is no `gap`/`libgap`. This is remedied by `mamba install gap`. 

Not sure why this doesn't work for you locally. On ci, it finds gap. The correct command would also be `mamba install gap-defaults` I think.

dimpase:
> Not sure why this doesn't work for you locally

never mind, it was just a bit broken venv, mixing in things from Homebrew and things from conda; this broke C++ toolchain.

dimpase (approved):

tobiasdiez:
Thanks Dima!

vbraun:
```
$ ./bootstrap -s
[...]
./bootstrap: installing /home/release/Sage/build/pkgs/sagemath_repl/src/requirements-editable.txt
./bootstrap: installing /home/release/Sage/build/pkgs/sagemath_repl/src/requirements.txt
./bootstrap: installing /home/release/Sage/build/pkgs/sagemath_sirocco/src/pyproject.toml
./bootstrap: installing /home/release/Sage/build/pkgs/sagemath_sirocco/src/requirements.txt
./bootstrap: installing /home/release/Sage/build/pkgs/sagemath_tdlib/src/pyproject.toml
./bootstrap: installing /home/release/Sage/build/pkgs/sagemath_tdlib/src/requirements.txt
configure.ac:64: installing 'config/compile'
configure.ac:64: installing 'config/config.guess'
configure.ac:64: installing 'config/config.sub'
configure.ac:50: installing 'config/install-sh'
configure.ac:50: installing 'config/missing'
Creating upstream/configure-33ed2859206b71e3ed381b33585d0139499ac34d.tar.gz...
find: warning: you have specified the global option -maxdepth after the argument -name, but global options are not positional, i.e., -maxdepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.
tar: environment-3.[89].yml: Cannot stat: No such file or directory
tar: environment-3.1[0-9].yml: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

dimpase:
How about something like this? (it does fix `./bootstrap -s`)

```diff
--- a/bootstrap
+++ b/bootstrap
@@ -225,14 +225,14 @@ save () {
 	build/make/Makefile-auto.in \
 	src/doc/en/installation/*.txt \
 	$(find src/doc/en/reference/spkg -name index.rst -prune -o -maxdepth 1 -name "*.rst" -print) \
-	environment-3.[89].yml environment-3.1[0-9].yml \
-	src/Pipfile \
+	environment-3.[9]-*.yml environment-3.1[0-9]-*.yml \
 	build/pkgs/cypari/version_requirements.txt \
 	build/pkgs/cysignals/version_requirements.txt \
 	build/pkgs/cython/version_requirements.txt \
 	build/pkgs/gmpy2/version_requirements.txt \
 	build/pkgs/jupyter_core/version_requirements.txt \
 	build/pkgs/memory_allocator/version_requirements.txt \
+	build/pkgs/meson/version_requirements.txt \
 	build/pkgs/numpy/version_requirements.txt \
 	build/pkgs/pkgconfig/version_requirements.txt \
 	build/pkgs/pplpy/version_requirements.txt \
```
Also, what's the point of `src/Pipfile.m4` - in particular as it's just a template, not filled in anymore?
And is `src/Pipfile` used anywhere?

tobiasdiez:
> How about something like this? (it does fix `./bootstrap -s`)
> 
> ```diff
> --- a/bootstrap
> +++ b/bootstrap
> @@ -225,14 +225,14 @@ save () {
>  	build/make/Makefile-auto.in \
>  	src/doc/en/installation/*.txt \
>  	$(find src/doc/en/reference/spkg -name index.rst -prune -o -maxdepth 1 -name "*.rst" -print) \
> -	environment-3.[89].yml environment-3.1[0-9].yml \
> -	src/Pipfile \
> +	environment-3.[9]-*.yml environment-3.1[0-9]-*.yml \
>  	build/pkgs/cypari/version_requirements.txt \
>  	build/pkgs/cysignals/version_requirements.txt \
>  	build/pkgs/cython/version_requirements.txt \
>  	build/pkgs/gmpy2/version_requirements.txt \
>  	build/pkgs/jupyter_core/version_requirements.txt \
>  	build/pkgs/memory_allocator/version_requirements.txt \
> +	build/pkgs/meson/version_requirements.txt \
>  	build/pkgs/numpy/version_requirements.txt \
>  	build/pkgs/pkgconfig/version_requirements.txt \
>  	build/pkgs/pplpy/version_requirements.txt \
> ```
> 

Thanks!

> Also, what's the point of `src/Pipfile.m4` - in particular as it's just a template, not filled in anymore? And is `src/Pipfile` used anywhere?

These files have been deleted now in #39031.

Issue #37454, opened 2024-02-24
Method `.random_element` not impl'ed for`LaurentPolynomialRing`


grhkm21:
#### New Method

```sage
sage: LaurentPolynomialRing(QQ, "x").random_element() # errors
```

#### Original Method

Run `sage.rings.tests.test_random_arith` with seed `97302818739086349733003447647548552705`:

```sh
$ sage -c "from sage.rings.tests import test_random_arith as t; t(trials=1000, seed=97302818739086349733003447647548552705)" | tail -n 7
Univariate Laurent Polynomial Ring in x over Number Field in aa with defining polynomial x^2 - 3*x - 99 over its base field
Random testing has revealed a problem in test_random_arith
Please report this bug!  You may be the first
person in the world to have seen this problem.
Please include this random seed in your bug report:
Random seed: 97302818739086349733003447647548552705
NotImplementedError()
```

### Additional Information

Discovered in https://github.com/sagemath/sage/actions/runs/8031658497/job/21940219082?pr=37441 of #37441

Edit: Note that the cause of the workflow failing is NOT this method (as the failure is marked as a "timeout"). However, this is still logged as "Please report this bug".

### Environment

```markdown
Sage Version: both 10.2 & `develop`
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

grhkm21:
It seems that
```sage
sage: LaurentPolynomialRing(QQ, "x").random_element() # errors
```
is not implemented. I guess I can try to implement a quick one real quick...

grhkm21:
Also I am very confused. I get this bug almost *every* run with `test_random_arith(trials=1000)`. How is this test only caught now?

tscrim:
That seems to occur not too often and it simply should not have been included, or at least protected against, since it is not implemented (although the implementation is quite easy by using a random polynomial). Anyways, we should *independently* implement the random Laurent polynomial.

GiacomoPope:
Would the implementation of this differ all that much from the usual impl of random elements in a polynomial ring?

grhkm21:
No, pretty much you just sample a shifted version then divide by a power of x

tscrim:
Indeed, it just needs an additional `min_val` to specify the minimal valuation, which would correspond to the appropriate shift. (Note this also affects the polynomial sampling degree.)

GiacomoPope:
[...]

PR #37457, opened 2024-02-24
fix wrong doctests that don't throw errors properly under `@random_testing`

grhkm21:
depends on #37490


grhkm21:
I temporarily replaced `LaurentPolynomialRing.random_element` to return $0$ for now. It is an open issue #37454.

grhkm21:
Actually hmm, I am not sure that this is the correct implementation. Looking at the existing docstring, the decorator is used as follows:

> Setting `print_seed` to true is useless in doctests, because the random seed printed will never match the expected doctest result (and using `# random` means the doctest framework will never report an error even if one happens). However, it is useful if you have a random test that sometimes segfaults. The normal print-the-random-seed-on-exceptions won’t work then, so you can run:
> > `while True: test_foo(print_seed=True)`
>
> and look at the last seed that was printed before it crashed.

However, my patch currently doesn't allow this behaviour anymore, as I redirect stdout into the `StringIO` object. Any ideas on how to implement this correctly would be appreciated!

tscrim (changes requested):
Catching an redirecting `stdout` is very brittle (in addition to the issue you mentioned); as such, I think it is a bad approach to addressing this. Also a `...` will still eat error outputs.

I actually don't quite understand the issue that is being addressed. What is wrong with printing the output instead of raising it? This is meant for doctests, so the output is really what we care about.

Also, the random Laurent polynomial should be handled separately. Returning `0` is just hiding more issues.

grhkm21:
> Catching an redirecting stdout is very brittle (in addition to the issue you mentioned); as such, I think it is a bad approach to addressing this.

What's a better approach?

> Also a ... will still eat error outputs.
> I actually don't quite understand the issue that is being addressed. What is wrong with printing the output instead of raising it? This is meant for doctests, so the output is really what we care about.

No not really. Consider the following doctest

```python
@random_testing
def f():
    """
    a

    EXAMPLES::

        sage: f()
        1
        ...
    """
    for i in range(1, 6):
        print(i)
        raise AssertionError()
```

Before this patch, the doctest passes, but now a `RuntimeError` from the decorator will be raised, so the first line of the output is Traceback ..., which causes the doctest to fail.

> Also, the random Laurent polynomial should be handled separately. Returning 0 is just hiding more issues.

Okay, expect this PR to be stalled for a while then.

GiacomoPope:
This could have a dependency on https://github.com/sagemath/sage/pull/37490 which would solve this issue?

tscrim:
> > Catching an redirecting stdout is very brittle (in addition to the issue you mentioned); as such, I think it is a bad approach to addressing this.
> 
> What's a better approach?

Why not just reraise the error message after printing the data (possibly editing its message)?

> > Also a ... will still eat error outputs.
> > I actually don't quite understand the issue that is being addressed. What is wrong with printing the output instead of raising it? This is meant for doctests, so the output is really what we care about.
> 
> No not really. Consider the following doctest
> 
> ```python
> @random_testing
> def f():
>     """
>     a
> 
>     EXAMPLES::
> 
>         sage: f()
>         1
>         ...
>     """
>     for i in range(1, 6):
>         print(i)
>         raise AssertionError()
> ```
> 
> Before this patch, the doctest passes, but now a `RuntimeError` from the decorator will be raised, so the first line of the output is Traceback ..., which causes the doctest to fail.

This has more of a feeling of being a bad doctest. It seems like even with (re)raising the error it should pass (in particular, the print should happen before the error is raised, which is useful for debugging).

grhkm21:
> It seems like even with (re)raising the error it should pass

Yeah but it doesn't, that's the point.

tscrim:
@grhkm21 You mean with this PR, right?

Print messages should happen *before* the error because they occurred *before* the error is raised. This is a fundamental programming principle. Strictly speaking, the print statement is not an output either.

Bad doctests are bad doctests. Changing code in complicated ways to make a doctest "work" has a very strong code smell to me. (For instance, in your example I could simply change the doctest to do the `...` first and make it pass with the PR.)

Furthermore, I still don't really understand why the current implementation is fundamentally wrong (as opposed to changing doctests to not catch everything). I could see re-raising the error instead, but again, since this is meant for doctests, this isn't so important. Also it is not unreasonable for it to not error out to continue to run the random tests.

GiacomoPope:
I'm not totally following the exchange above, but this all came from the fact that there was randomised testing for rings, which was "passing" in the CLI for doctests but was printing that it failed. I think the proposed change is that random testing should fail in a discoverable way.

Either the above is done, or these older functions which relied on this decorator need reformatting? The current situation, where a random test "passes" and prints to the testing CLI output that it failed is obviously not intended?

GiacomoPope:
Explicitly:

```py
Doctesting 1 file.
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/tests.py
    [62 tests, 16.00 s]
All tests passed!
Total time for all tests: 17.5 seconds
    cpu time: 14.7 seconds
    cumulative wall time: 16.0 seconds
Features detected for doctesting: sage.modules,sage.rings.finite_rings,sage.rings.number_field,sage.rings.padics,sage.symbolic
pytest is not installed in the venv, skip checking tests that rely on it
```

But if we actually run the functions from this test:

```
sage -c "from sage.rings.tests import test_random_arith as t; t(trials=1000, seed=286735480429121101562228604801325644303)"

...

survived 26 tests
Univariate Laurent Polynomial Ring in x over Finite Field of size 59539
Random testing has revealed a problem in test_random_arith
Please report this bug!  You may be the first
person in the world to have seen this problem.
Please include this random seed in your bug report:
Random seed: 286735480429121101562228604801325644303
NotImplementedError()
```

grhkm21:
depends on #37490

grhkm21:
I think this should work now, maybe have a look. (Waiting for CI's of course)

PR #37465, opened 2024-02-25
use Parent for the Pari pseudo-ring

fchapoton:
switch the Pari pseudoring to use `Parent` instead of the auld `Ring` class

### :memo: Checklist


mkoeppe (approved):
LGTM

PR #37469, opened 2024-02-25
various details in permutation.py (pep8 and ruff PERF)

fchapoton:
various small enhancements in permutation.py

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/permutation.py
+++ b/src/sage/combinat/permutation.py
@@ -451,20 +451,17 @@ class Permutation(CombinatorialElement):
             return l
         elif isinstance(l, PermutationGroupElement):
             l = l.domain()
-        #if l is a string, then assume it is in cycle notation
+        # if l is a string, then assume it is in cycle notation
         elif isinstance(l, str):
             if l == "()" or l == "":
                 return from_cycles(0, [])
             cycles = l.split(")(")
             cycles[0] = cycles[0][1:]
             cycles[-1] = cycles[-1][:-1]
-            cycle_list = []
-            for c in cycles:
-                cycle_list.append([int(_) for _ in c.split(",")])
-
+            cycle_list = [[int(k) for k in c.split(",")] for c in cycles]
             return from_cycles(max(max(c) for c in cycle_list), cycle_list)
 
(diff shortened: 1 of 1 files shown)


grhkm21 (approved):
Looks good. Just wondering, do you do these manually?

fchapoton:
yes, I do

src/sage/combinat/permutation.py
```diff
@@ -1070,12 +1067,12 @@ def _to_cycles_set(self, singletons=True):
         cycles = []
 
         if not singletons:
-            #remove the fixed points
-            L = set(i+1 for i,pi in enumerate(p) if pi != i+1)
+            # remove the fixed points
+            L = {i+1 for i,pi in enumerate(p) if pi != i+1}
```
mantepse, line 1071: I think there is a whitespace after the comma missing

mantepse: Also, here and below, enumerate(p, 1) would be better.

grhkm21: [PEP8](https://peps.python.org/pep-0008/#other-recommendations) says it's not required

> If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator:
```python
# Correct:
i = i + 1
submitted += 1
x = x*2 - 1
hypot2 = x*x + y*y
c = (a+b) * (a-b)
```

That's why I didn't point it out

fchapoton: thanks, Martin, but let us keep that for later, please. This is already positive reviewed.

mantepse: Comma, not operator.

grhkm21: Sorry for rushing the review

mantepse: @fchapoton: in case you open a new ticket, please let me know.  (I admit I do not care enough, but I'm willing to browse through the changes.)

PR #37473, opened 2024-02-25
Fix hanging computation when calling `_isomorphic_ring()` with large characteristic

### LINKED ISSUE #37472: `_isomorphic_ring()` hangs for large characteristic
### Steps To Reproduce

```py
def extend_to_finite_field(field, modulus):
    if not modulus.is_irreducible():
        raise ValueError

    if field != modulus.base_ring():
        raise ValueError

    R = modulus.parent()
    quo_ring = R.quotient_ring(modulus, "xbar")
    _, _, ff = quo_ring._isomorphic_ring()
    return ff

F.<a> = GF((2**255 - 19, 2))
f = PolynomialRing(F, names="x").irreducible_element(2)
%time _ = F.extension(f)
CPU times: user 3.38 ms, sys: 22 µs, total: 3.4 ms
Wall time: 3.4 ms
%time _ = extend_to_finite_field(F, f)
# Takes forever
```

### Expected Behavior

It should finish in reasonable time

### Actual Behavior

It doesnt finish

### Additional Information

This is due to a know problem for constructing finite fields with large characteristic without the name variable. Adding a name variable `GF(N, names="a")` fixes this.

However, this fix breaks much more than it helps. At the moment when the `names` value is `None` then the Conway polynomial is used which aids with the coercion maps. Adding the names value breaks this and also effects how isomorphisms can even be ensured.

I started a PR for this, but closed it when I realised how much more complex the correct solution is.

### Environment

```markdown
- **OS**: Mac OS
- **Sage Version**: SageMath version 10.3.rc0, Release Date: 2024-02-25
```

### Checklist

GiacomoPope:
When constructing a finite field of non-prime order, if no name is given then a conway polynomial is computed which is computationally very expensive for large characteristic.

This PR adds a name argument to the `GF()` construction in `_isomorphic_ring()` to stop this.

## Before PR

```py
def extend_to_finite_field(field, modulus):
    if not modulus.is_irreducible():
        raise ValueError

    if field != modulus.base_ring():
        raise ValueError

    R = modulus.parent()
    quo_ring = R.quotient_ring(modulus, "xbar")
    _, _, ff = quo_ring._isomorphic_ring()
    return ff

F.<a> = GF((2**255 - 19, 2))
f = PolynomialRing(F, names="x").irreducible_element(2)
%time F.extension(f)
CPU times: user 5.56 ms, sys: 4 µs, total: 5.57 ms
Wall time: 5.57 ms
%time _ = extend_to_finite_field(F, f)
# Doesn't finish
```

## After PR

```py
F.<a> = GF((2**255 - 19, 2))
f = PolynomialRing(F, names="x").irreducible_element(2)
%time _ = F.extension(f)
CPU times: user 3.38 ms, sys: 22 µs, total: 3.4 ms
Wall time: 3.4 ms
%time _ = extend_to_finite_field(F, f)
CPU times: user 270 ms, sys: 1.57 ms, total: 272 ms
Wall time: 271 ms
```

Fixes https://github.com/sagemath/sage/issues/37472

### DIFF
--- a/src/sage/rings/polynomial/polynomial_quotient_ring.py
+++ b/src/sage/rings/polynomial/polynomial_quotient_ring.py
@@ -218,7 +218,7 @@ class PolynomialQuotientRingFactory(UniqueFactory):
             raise TypeError("ring must be a polynomial ring")
         if not isinstance(polynomial, polynomial_element.Polynomial):
             raise TypeError("must be a polynomial")
-        if not polynomial.parent() is ring:
+        if polynomial.parent() is not ring:
             raise TypeError("polynomial must be in ring")
 
         c = polynomial.leading_coefficient()
--- a/src/sage/rings/polynomial/polynomial_quotient_ring.py
+++ b/src/sage/rings/polynomial/polynomial_quotient_ring.py
@@ -2009,7 +2009,7 @@ class PolynomialQuotientRing_generic(QuotientRing_generic):
             Univariate Quotient Polynomial Ring in b
              over Finite Field in a of size 2^2 with modulus b^2 + b + a
             sage: from_M, to_M, M = L._isomorphic_ring(); M
-            Finite Field in z4 of size 2^4
+            Finite Field in zn of size 2^4
             sage: R.<c> = L[]
             sage: M.<c> = L.extension(c^2 + b*c + b); M
             Univariate Quotient Polynomial Ring in c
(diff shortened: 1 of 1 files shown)


GiacomoPope:
This seems to have broken a lot of weird things internally so I'll mark it as draft but for something so minor maybe it should be closed and ignored.

GiacomoPope:
I'm just going to close this. By allowing `names` in the field, the conway polynomial not being used messes up the coercion system and the fix just further complicates an already complicated function.

I will think of a better fix.

PR #37477, opened 2024-02-26
Add sheaf cohomology for projective schemes

kwankyu:
so we add a sheaf cohomology machinery of coherent sheaves on projective schemes. Using the machinery, we implement some invariants for projective subschemes. In particular,

- add Euler characterstic of coherent sheaves
- add arithmetic genus of projective schemes

### :memo: Checklist


kwankyu:
> I can review this when its ready.

I made this PR somewhat prematurely, to test the new PR template.

I will soon update and make it ready. Thanks!

src/sage/schemes/projective/projective_subscheme.py
```diff
@@ -1515,3 +1515,77 @@ def local_height_arch(self, i, prec=None):
...
+            sage: X.arithmetic_genus()
+            3
+            sage: X.geometric_genus()
+            1
+            sage: X.is_singular()
+            True
+        """
+        p = self.structure_sheaf().euler_characteristic() - 1
+        if self.dimension() % 2:
+            p = -p
```
tscrim, line 1570: Could you leave a code comment about this?

I am guessing the even dimensional case has a structure sheaf whose Euler characteristic is 0. Wouldn't it be faster to first check the dimension and just return `ZZ.one()` in that case?

src/sage/schemes/projective/projective_subscheme.py
```diff
@@ -1515,3 +1515,77 @@ def local_height_arch(self, i, prec=None):
...
+        Return the sheaf defined by the graded ``module``.
+
+        If ``twist`` is a non-zero integer `n`, the sheaf twisted by
+        `\OO_{\PP^r}(n)` is returned.
+
+        INPUT:
+
+        - ``module`` -- a free module or a quotient module
+
+        - ``twist`` -- (default: `0`) an integer
```
tscrim:
```suggestion
        - ``twist`` -- (default: ``0``) an integer
```
since it is code input :)

src/sage/schemes/projective/projective_space.py
```diff
@@ -2266,6 +2267,71 @@ def line_through(self, p, q):
...
+            Coherent sheaf on Projective Space of dimension 3 over Rational Field
+        """
+        M = FreeModule(self.coordinate_ring(), rank=1)
+        return self.coherent_sheaf(M, twist=twist)
+
+    def arithmetic_genus(self):
+        r"""
+        Return the arithmetic genus of this projective space.
+
+        This is known to be zero. This method simply checks the fact.
```
tscrim: I think it would be better to simply return `0` here. Then you can add doctests that verifies it by the computation.

src/sage/schemes/projective/coherent_sheaf.py
```diff
@@ -0,0 +1,323 @@
...
+        except AttributeError:
+            pass
+
+        assert module.cover() == module.free_cover()
+
+        self._base_scheme = scheme
+        self._module = module
+        self._twist = twist
+
+    @cached_property
```
tscrim:
```suggestion
@abstract_method(optional=True)
```
Since the default is to raise an error, there is no need to cache it and this will be overwritten by subclasses. This would be better for the developer IMO (and you wouldn't need to raise the error).

src/sage/schemes/projective/coherent_sheaf.py
```diff
@@ -0,0 +1,323 @@
...
+
+        EXAMPLES::
+
+            sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
+            sage: X = P2.subscheme(x^4 + y^4 + z^4)
+            sage: sh = X.structure_sheaf()
+            sage: sh.twist(1).euler_characteristic()
+            2
+            sage: sh.twist(2).euler_characteristic()
+            6
```
tscrim, line 183: Please add a doctest showing the twist of a twisted shift adds.

src/sage/schemes/projective/coherent_sheaf.py
```diff
@@ -0,0 +1,323 @@
...
+    r"""
+    Coherent sheaf on a projective space.
+
+    EXAMPLES::
+
+        sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
+        sage: P2.structure_sheaf()
+        Coherent sheaf on Projective Space of dimension 2 over Rational Field
+    """
+    @cached_property
```
tscrim, line 229: What `@cached_property` instead of `@lazy_attribute`? Also, wouldn't it make more sense for `MaruyamaComplex` to construct itself lazily?

src/sage/schemes/projective/coherent_sheaf.py
```diff
@@ -0,0 +1,323 @@
...
+
+    EXAMPLES::
+
+        sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
+        sage: X = P2.subscheme(x^4 + y^4 + z^4)
+        sage: X.structure_sheaf()
+        Coherent sheaf on Closed subscheme of Projective Space of dimension 2
+         over Rational Field defined by: x^4 + y^4 + z^4
+    """
+    @cached_property
```
tscrim, line 256: Same comment as above. Although because of the `image_to_ambient_space`, this would make more sense to do lazily here.

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,500 @@
...
+    &\quad -\rank H^0(f_1)-\rank H^r(f_r)
+    \end{split}
+
+in which the complex of (bottom) cohomology groups
+
+.. MATH::
+
+    \bigoplus_{j=1}^{t_{i+1}}H^0(\OO_{\PP^r}(-m^{(i+1)}_j))\overset{H^0(f_{i+1})}{\longrightarrow}
+    \bigoplus_{j=1}^{t_i}H^0(\OO_{\PP^r}(-m^{(i)}_j))\overset{H^0(f_{i})}{\longrightarrow}
+    \bigoplus_{j=1}^{t_{i-1}}H^0(\OO_{\PP^r}(-m^{(i-1)}_j))
```
tscrim:
```suggestion
    \bigoplus_{j=1}^{t_{i-1}}H^0(\OO_{\PP^r}(-m^{(i-1)}_j)),
```

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,500 @@
...
+            sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
+            sage: X = P2.subscheme([x^4 + y^4 + z^4])
+            sage: c = X.structure_sheaf()._cohomology
+            sage: c.cohomology_group_bottom(0)
+            Bottom Cohomology Group of dimension 1
+        """
+        return f'Bottom Cohomology Group of dimension {self.rank}'
+
+
+class CohomologyGroupTop:
```
tscrim: This is just a copy of `CohomologyGroupBottom` except for the string representation and the integer vectors used. I believe. It would be better to combine them with an input of `is_top=boolean` that splits between the two cases. Or refactoring things to a common ABC.

Also, why aren't these subclasses of `Parent`, or at least `UniqueRepresentation`?

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,500 @@
...
+        elif 1 <= q and q < r:
+            return self.differential_top(r - q).kernel().quotient(self.differential_top(r - q + 1).image())
+        elif q == 0:
+            # 0th cohomology group of (global sections) is not realized in
+            # terms of cohomology groups and differential maps of twisted
+            # structure sheaves of the projective space.
+            return VectorSpace(self.coefficient_field, self.h(0))
+        elif q > r:
+            return VectorSpace(self.coefficient_field, 0)
+        raise IndexError('index must be non-negative')
```
tscrim, line 478:
```suggestion
        return VectorSpace(self.coefficient_field, 0)
```
I think extending it by 0-dimensional spaces in both directions is the most natural.

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,500 @@
...
+        elif 1 <= q and q < r:
+            return self.differential_top(r - q).kernel().dimension() - self.differential_top(r - q + 1).rank()
+        elif q == 0:
+            a = (self.cohomology_group_bottom(0).rank - self.cohomology_group_top(r + 1).rank
+                 + self.cohomology_group_top(r).rank)
+            b = self.differential_bottom(1).rank() + self.differential_top(r).rank()
+            return a - b
+        elif q > r:
+            return 0
+        raise IndexError('index must be non-negative')
```
tscrim:
```suggestion
        return ZZ.zero()
```
Same comment. Also, let's try to make sure Python `int`'s don't get returned.

tscrim (changes requested):

src/sage/schemes/projective/projective_subscheme.py
kwankyu: The formula, actually the definition of the arithmetic genus $p_a$ is 

$p_a=(-1)^r(\chi - 1)$

where $\chi$ is the Euler characteristic and $r$ is the dimension. The code just follows the formula. Hence I think there is no need of special casing.

It is also hard to write a code comment. So I will add the formula itself to the docstring.

src/sage/schemes/projective/projective_subscheme.py
kwankyu: OK.

src/sage/schemes/projective/projective_space.py
kwankyu: OK. Done.

src/sage/schemes/projective/coherent_sheaf.py
kwankyu: Right. Thanks for catching this.

Removed `@cached_property`.

src/sage/schemes/projective/coherent_sheaf.py
kwankyu: Good idea. Done.

src/sage/schemes/projective/coherent_sheaf.py
kwankyu: `@cached_property` is standard python while `@lazy_attribute` is a sage thing. As far as I know, there is no difference in functionality. Is there?

I prefer `@cached_property` because I am aware that using certain sage-specific decorators is presently an obstacle in removing `src/sage/sage_docbuild/ext/sage_autodoc.py`, which is preventing us to use vanilla Sphinx and is a major friction in upgrading Sphinx. I hope someday to replace all `@lazy_attribute` with `@cached_property` :-)

`MaruyamaComplex` is an internal class. I don't see a need to make it fancy.

src/sage/schemes/projective/coherent_sheaf.py
kwankyu: I don't understand. `MaruyamaComplex` is constructed lazily, when requested via `cohomology()`, and then cached.

src/sage/schemes/projective/cohomology.py
kwankyu: Thanks.

src/sage/schemes/projective/cohomology.py
kwankyu: They are two simple classes (in code length and the number of methods) each of which represent different mathematical objects. They are better to be separate. They are internal as the docstring of the module says. There is zero possibility that new method will be added to them in future. No need to make them technically complicated.

Please let them be left simple.

src/sage/schemes/projective/cohomology.py
kwankyu: No. Cohomology group with negative index is not even defined. Cohomology groups do not form a complex.

src/sage/schemes/projective/cohomology.py
kwankyu: Right. I will use `Integer(0)`.

src/sage/schemes/projective/cohomology.py
kwankyu: Ah, you may be misled because the class is called `MaruyamaComplex`. There are complexes behind, but the class itself is a structure to compute the cohomology groups from the complexes behind. Sorry for the confusing name.

kwankyu: If needed, I may rename it perhaps simply to `Maruyama`.

src/sage/schemes/projective/cohomology.py
tscrim: The high level of duplication makes keeping them separate a fair amount of technical debt. I don't think they are simple enough to warrant this. They are not so distinct though. It's accessible through the API (as shown through the doctests), so it is not internal either. I think they should at least inherit from `UniqueRepresentation` and `SageObject`. I also think that they will never be extended is a bad assumption.

src/sage/schemes/projective/coherent_sheaf.py
tscrim: It is better design to have `MaruyamaComplex` be a `UniqueRepresentation`. This will handle the appropriate caching, it could then be pickled (so used in multiprocessing, used across sessions, etc.), and it's not internal since you give the user access to it (which I think is the right thing to do).

re: `@cached_property`, I am not sure about the differences between the two. From some basics checks, they agree, but I didn't check carefully or in Cython. I would be *extremely* careful about such mass replacements at this point.

src/sage/schemes/projective/coherent_sheaf.py
tscrim: It has to do with the call to `image_to_ambient_space`, which would be an argument for keeping the lazily construction in this class (rather than in `MaruyamaComplex`).

src/sage/schemes/projective/cohomology.py
tscrim: Okay, I kind of understand. I don't know the mathematics here and am taking guesses based on the PR. ~It would be probably be good to add a more precise description of what is being done (mathematically).~ (Edit: I forgot this is in the module-level doc.)

Technically the cohomology groups are still defined, just like for those extended for "large" indices, but they are just all trivial.

`Maruyama` by itself is not descriptive enough. Perhaps `MaruyamaComputation`?

src/sage/schemes/projective/projective_subscheme.py
tscrim: Thanks.

src/sage/schemes/projective/cohomology.py
tscrim: You claim that this is a complex, which is necessary to do the computation. Well, it's more of two complexes melded together from the description.

tscrim:
Thinking a bit more, I am wondering if your implementation is the right way to do it. Mathematically, it seems like you should have two separate complexes, which then `_cohomology_group` uses to do the main computations.

For anything you want to make internal, you should start with the leading `_` and hide the methods/attributes. In a similar way, I don't think it is a good ideal to use hidden methods in public doctests.

Can you also fix the linter issue too?

kwankyu:
> Thinking a bit more, I am wondering if your implementation is the right way to do it. Mathematically, it seems like you should have two separate complexes, which then `_cohomology_group` uses to do the main computations.

There are other methods to compute the cohomology (the dimension of the cohomology group) of coherent sheaves. Actually Maruyama's method is not the usual one implemented. Eisenbud-Floystad-Schreyer method is used in Macaulay2, and perhaps in Magma.

In the future, someone may implement other methods. Perhaps I won't.  Then the main public API `cohomology()` (and the hidden method `_cohomology_group()` would get `algorithm` argument for a user to choose the method.

Hence the module `cohomology` actually computing the cohomology (and the cohomology group) is better to be separate from the module `coherent_sheaf` implementing coherent sheaves and providing `cohomology()` for a user. There is no need for the user to directly interface with the internal computation engines (Maruyama method or Eisenbud-Floystad-Schreyer method).


Now there remains the question of how much "hidden" my implementation of Maruyama's method should be. I thought that declaring the module `cohomology` "internal" in the module docstring is enough - The reader is warned but if he wants he can read further to see what are in the module in the documentation. This is very common indeed. "internal" modules appear in our documentation. For instance, https://doc.sagemath.org/html/en/reference/groups/sage/groups/libgap_mixin.html. I guess you do not mean that all "internal" modules contain only classes whose name starts with `_`.

> In a similar way, I don't think it is a good idea to use hidden methods in public doctests.
> 
> Can you also fix the linter issue too?

OK. I will do these later.

tscrim:
It’s okay to have the class names be public (I would tend to say this is a good thing for most cases, this one included). However, you can directly access it by public methods (here, `maruyama_complex()`), so it is no longer internal.

src/sage/schemes/projective/cohomology.py
kwankyu: As it is confusing, I renamed the class with `MaruyamaMethod`. It keeps the two complexes behind for computation purposes, but is not a mathematical object in itself.

src/sage/schemes/projective/cohomology.py
kwankyu: I made them inherit from `SageObject`.

I want to keep the two classes separate as they represent different  mathematical objects, and there is nothing to make up a common superclass.

kwankyu:
> However, you can directly access it by public methods (here, `maruyama_complex()`), so it is no longer internal.

There is no such method.

src/sage/schemes/projective/coherent_sheaf.py
kwankyu: > It is better design to have `MaruyamaComplex` be a `UniqueRepresentation`. This will handle the appropriate caching, it could then be pickled (so used in multiprocessing, used across sessions, etc.), 

Appropriate caching is handled by  `@cached_property`, and there is no problem in pickling.

```sage
sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
sage: X = P2.subscheme([x^4 + y^4 + z^4])
sage: sh = X.structure_sheaf()
sage: sh
Coherent sheaf on Closed subscheme of Projective Space of dimension 2 over Rational Field defined by:
  x^4 + y^4 + z^4
sage: loads(dumps(sh))
Coherent sheaf on Closed subscheme of Projective Space of dimension 2 over Rational Field defined by:
  x^4 + y^4 + z^4
sage: loads(dumps(sh._cohomology))
Maruyama Method using S(0) <-- S(-4) <-- 0
```

I don't see a benefit of using `UniqueRepresentation`...

> and it's not internal since you give the user access to it (which I think is the right thing to do).

There is user access via `sh._cohomology`. It is still internal. What do you think is not internal?

kwankyu: I removed `_cohomology_group()` method of sheaf, by the way.

kwankyu:
> Can you also fix the linter issue too?

Thanks. Done.

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,508 @@
...
+    sage: sh._cohomology.H(1)
+    Vector space quotient V/W of dimension 3 over Rational Field where
+    V: Vector space of degree 3 and dimension 3 over Rational Field
+    Basis matrix:
+    [1 0 0]
+    [0 1 0]
+    [0 0 1]
+    W: Vector space of degree 3 and dimension 0 over Rational Field
+    Basis matrix:
+    []
```
mkoeppe, line 91:
```suggestion
     V: Vector space of degree 3 and dimension 3 over Rational Field
        Basis matrix:
        [1 0 0]
        [0 1 0]
        [0 0 1]
     W: Vector space of degree 3 and dimension 0 over Rational Field
        Basis matrix:
        []
```

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,508 @@
...
+    sage: sh._cohomology.H(1).dimension() == sh.cohomology(1)
+    True
+    sage: sh.cohomology(2)
+    0
+    sage: sh._cohomology.H(2)
+    Vector space quotient V/W of dimension 0 over Rational Field where
+    V: Vector space of dimension 0 over Rational Field
+    W: Vector space of degree 0 and dimension 0 over Rational Field
+    Basis matrix:
+    []
```
mkoeppe, line 101:
```suggestion
    Vector space quotient V/W of dimension 0 over Rational Field where
     V: Vector space of dimension 0 over Rational Field
     W: Vector space of degree 0 and dimension 0 over Rational Field
        Basis matrix:
        []
```

src/sage/schemes/projective/cohomology.py
```diff
@@ -0,0 +1,508 @@
...
+            sage: c.H(1)
+            Vector space quotient V/W of dimension 3 over Rational Field where
+            V: Vector space of degree 3 and dimension 3 over Rational Field
+            Basis matrix:
+            [1 0 0]
+            [0 1 0]
+            [0 0 1]
+            W: Vector space of degree 3 and dimension 0 over Rational Field
+            Basis matrix:
+            []
```
mkoeppe, line 464:
```suggestion
             V: Vector space of degree 3 and dimension 3 over Rational Field
                Basis matrix:
                [1 0 0]
                [0 1 0]
                [0 0 1]
             W: Vector space of degree 3 and dimension 0 over Rational Field
                Basis matrix:
                []
```

kwankyu: Thanks. Done.

src/sage/schemes/projective/coherent_sheaf.py
kwankyu: The attribute `_cohomology` of `CoherentSheaf_xxx` class has `@cached_property`. 

So the object of class `MaruyamaMethod` is constructed when requested (lazily), and it is kept both in  the attribute `_cohomology` of `CoherentSheaf_on_projective_subscheme` and the attribute `_cohomology` of `CoherentSheaf_on_projective_space`. I see no problem here.

> It is better design to have `MaruyamaMethod` be a `UniqueRepresentation`.

Sorry. I still don't get it. 

`UniqueRepresentation` is used when returning the same object for the repeated requests with the same input is desirable.  I see no situation like that here.

src/sage/schemes/projective/cohomology.py
tscrim: That’s like saying we should have two classes for the even and odd integers. They are different mathematical objects, but as far as their uses, functionality, and implementation are concerned, they are essentially the same.

The code is very clear that there should be at least a common base class (you’re not disputing that there is a high level of duplication, right?), if not just having the output differ by an input (e.g., a boolean argument for top versus bottom).

src/sage/schemes/projective/cohomology.py
tscrim: But it is a mathematical object; as you just said, it is a pair of complexes. However, I don’t see why the two complexes shouldn’t be separated.

tscrim:
Sorry, I think I misread a few things while going over the code quickly. Yes, `MaruyamaMethod` is not public, but I felt like the right thing that it should be part of the API. You’re (baselessly) assuming that nobody will want to look at this data, but somebody might want to check it (even if for nothing else but pedagogical reasons or to check against hand computations).

However, I am fairly convinced the entire design should be reconsidered as the two separate complexes for the top and bottom complexes, both of which are public facing that could be done by a single class (differing by a single input parameter). So we would need the appropriate comparisons (where `UniqueRepresentation` becomes very useful), hashing, etc. Note that this is *data* that is associated with the scheme that is then used by the algorithm to do the computation. It isn’t a class that does the computation and needs to have many subroutines, internal parameters, etc. Moreover, I feel like this design matches the mathematics better.

If you don’t want to change the design, would you be open to me rewriting it with this approach as a replacement PR?

src/sage/schemes/projective/cohomology.py
kwankyu: There are only two methods `__init__` and `__repr__`, and the two methods are different in the two classes. What high level of duplication is there? 

I guess that your hypothetical `EvenIntegers` and `OddIntegers` have over 10 (or say 3) methods and all of them are identical except `__init__` and `__repr__`. In that case, I would agree that there is a high level of duplication. 

Here `__init__` and `__repr__` are all, and they are not identical. I see no high level of duplication. 

Having the separate two classes, I could add an appropriate docstring to each class. Combining the two classes to one by using "a boolean argument for top versus bottom" is sacrificing readability for negligible code economy. I don't want to do that.

______

Fearing that you may be overestimating `CohomologyGroupTop` and `CohomologyGroupBottom` classes, I remark that the objects of the classes are not user-accessible at all. They are intermediate objects constructed and used during the computation and disappear.

src/sage/schemes/projective/cohomology.py
kwankyu: > But it is a mathematical object; as you just said, it is a pair of complexes.

A object of the class computes cohomology using the data about the pair of complexes. The cohomology, which is an integer, is not a property of the pair of complexes, it is a property of the sheaf.  It does not make a sense that the object "is" a pair of complexes. 

I wrote the class, and that is how I see it. Please respect my view.

> However, I don’t see why the two complexes shouldn’t be separated.

I don't understand. One complex is used to compute H(0), and the other is used to compute H(r) for r > 0.

kwankyu:
I know that you have a good intention. Thanks for that.

But I don't understand your proposal. It just sounds like overkill, like coating a break pedal with platinum (I just made up this analogy.  I suspect that you are overzealous :-) I do appreciate your attention and time in reviewing this PR. Please don't get me wrong. 

But let's solve the situation constructively. How about taking one of these two actions:

(A) You accept this PR as it is. Then you author a subsequent PR that implements your design. I may review the PR, but cannot be sure now that I would be positive to it. 

(B) Or you leave this PR, and I will wait for a new reviewer.

?

kwankyu:
I assume that you chose (B). Thanks for the review!

src/sage/schemes/projective/coherent_sheaf.py
```diff
@@ -0,0 +1,298 @@
...
+            Maruyama Method using S(0) <-- S(-4) <-- 0
+        """
+        return self.image_to_ambient_space()._cohomology
+
+    def image_to_ambient_space(self):
+        """
+        Return the direct image of this sheaf to the ambient space.
+
+        The image is with respect to the inclusion morphism from the base
+        scheme into the projective Space.
```
mkoeppe:
```suggestion
        scheme into the projective space.
```

src/sage/misc/latex_macros.py
```diff
@@ -175,7 +175,12 @@ def convert_latex_macro_to_mathjax(macro):
 latex_macros = [r"\newcommand{\SL}{\mathrm{SL}}",
                 r"\newcommand{\PSL}{\mathrm{PSL}}",
                 r"\newcommand{\lcm}{\mathop{\operatorname{lcm}}}",
-                r"\newcommand{\dist}{\mathrm{dist}}"]
+                r"\newcommand{\dist}{\mathrm{dist}}",
+                r"\newcommand{\im}{\mathop{\operatorname{im}}}",
+                r"\newcommand{\rank}{\mathop{\operatorname{rank}}}",
+                r"\newcommand{\PP}{\mathbf{P}}",
+                r"\newcommand{\OO}{\mathcal{O}}",
```
mkoeppe, line 178: are we sure that we want to create the macro `\OO`?

kwankyu: I also have qualm about it.  I need it since I use it extensively for docstrings in `src/sage/schemes/projective/cohomology.py`. On the other hand,  I feel a bit guilty for increasing the size of the list `latex_macros` since this will slow down documentation build...

kwankyu: But I have no idea how to circumvent this dilemma...

kwankyu: Sheaves are written in calligraphic font, if this is what you asked. I am following the convention.

See https://en.wikipedia.org/wiki/Sheaf_(mathematics)#:~:text=together%20with%20a-,structure%20sheaf,-%F0%9D%91%82 for example.

kwankyu: Macaulay2's convention seen in 

https://macaulay2.com/doc/Macaulay2-1.21/share/doc/Macaulay2/Macaulay2Doc/html/_sheaf_lp__Ring_rp.html

also affected me.

src/sage/misc/latex.py
```diff
@@ -583,6 +583,10 @@ def latex_extra_preamble():
         \newcommand{\PSL}{\mathrm{PSL}}
         \newcommand{\lcm}{\mathop{\operatorname{lcm}}}
         \newcommand{\dist}{\mathrm{dist}}
+        \newcommand{\im}{\mathop{\operatorname{im}}}
+        \newcommand{\rank}{\mathop{\operatorname{rank}}}
+        \newcommand{\PP}{\mathbf{P}}
```
mkoeppe: Should this be using the `\Bold` macro, like `\ZZ` does?

src/sage/misc/latex_macros.py
mkoeppe: Apologies for not asking a more specific question. Mainly I asked because the double-letter command suggests boldface/double-stroke to me, and this is how all of our existing double-letter commands expand. For the calligraphic O, I would expect to see many different conventions for latex macros including `\calO` or similar. (Note in IPython, one types `\scrO` + TAB to get this symbol.) The M2 reference may be convincing enough though.

kwankyu: Yes, `\calO` may be alternative, but I prefer `\OO` to `\calO` because of the M2 reference.
 
For the general calligraphic O, I think`\calO` is just slightly better than the vanilla `\mathcal{O}`.

src/sage/misc/latex.py
kwankyu: Yes. Thanks.

src/sage/misc/latex_macros.py
kwankyu: On the other hand, `\OO` is a latex macro only useful for algebraic geometry while the list `latex_macros` is supposed to keep macros in general use. I think in the long run, we should devise some mechanism to define latex macros for local use within the sage library... Right now I have no idea what the mechanism should be like...

kwankyu: > Mainly I asked because the double-letter command suggests boldface/double-stroke to me, and this is how all of our existing double-letter commands expand. 

Right. Fortunately, there is no prominent mathematical object denoted by **O**, as far as I know.

kwankyu: However, when someone claims `\OO` for boldfaced O, then we should reconsider this issue seriously :-)

mkoeppe: OK. To keep the good taste, I'll not mention the terrible things I've seen that were done using \mathbb{O}.

kwankyu: I cannot assert that I never saw boldfaced O in papers...

kwankyu:
Regarding the code duplication that Travis pointed out, I think I may combine methods of class `MaruyamaMethod`:

`cohomology_group_bottom()` and `cohomology_group_top()` to `cohomology_group(top=True)`;

`differential_top()` and `differential_bottom()` to `differential(top=True)`

This looks a bit ugly, but really removes a nontrivial code duplication. Let me know if this is desirable.

kwankyu:
Better not to do the refactoring suggested in https://github.com/sagemath/sage/pull/37477#issuecomment-2109609672, since Travis authored a subsequent PR #38003 that does more extensive refactoring based on the present PR.

kwankyu:
OK. I regard #38003 as a sequel to this PR. Please copy and paste your remarks above into the description of your PR.

kwankyu:
rebased.

mkoeppe:
I'm not sure how to review this, in light of the alternative implementation in #38003.

If this PR is to be merged first, would it make sense to mark some of the implementation details that are intended to be done differently in #38003 as `@experimental`?

tscrim:
It isn't a sequel though, it is an alternative proposal. I have reservations about including the code as-is. You are asking me to include your code (despite my objections) without any guarantees that my code will not be opposed.

kwankyu:
> It isn't a sequel though, it is an alternative proposal. I have reservations about including the code as-is. You are asking me to include your code (despite my objections) ...

That was my interpretation of your saying "The bulk of the code for #38003 is based off this branch". I misunderstood and I am sorry. I now see your intention.

> without any guarantees that my code will not be opposed.

I don't oppose a PR as a reviewer. I never did. If I don't like a PR, I leave it to others. I suggest my ideas to improve the code, but it is all up to the author whether to accept them or not.

Actually I did not see your PR yet. I don't know whether I will be positive or just leave.

kwankyu:
As you left this comment here:

> * ... For example, if they want to implement a nice basis for the cohomology (which can be done by Maruyama's method in most cases as I understand it; I'm not sure how much the current code does so) or implement group actions on the cohomology groups.

For the purpose,  I exposed `sheaf._cohomology.H(i)`. His paper defines an isomorphism between this vector space to the cohomology group for i > 0. Morphism between cohomology groups may be transferred to a map between these vector spaces, and then one may construct a basis nice to understand the map. But I didn't pursue the details...

> * The cohomology class could be easily expanded to include the cup product or Poincaré duality.

Note that the paper does not construct a vector space isomorphic to the 0-th cohomology group.

PR #37481, opened 2024-02-26
Implement `minimum_generating_set` function for group theory

### LINKED ISSUE #37362: Implement a function for finding minimum generating set of a group
### Problem Description

Currently, SageMath (or GAP) does not have any function to find the minimum generating set of a group. If a user wants to find the minimum generating set of a group, they have to go through all possible combinations of the elements of the group. As a result, the time complexity of this algorithm becomes exponential in terms of the cardinality of the group, which takes a lot of time for large groups.

### Proposed Solution

Implement a polynomial time (in terms of cardinality of a group) algorithm to find the minimum generating set of a group using the logic provided in the [research paper](https://www.sciencedirect.com/science/article/pii/S0021869323005720?via%3Dihub).

### Alternatives Considered

Users can manually implement the logic provided in the [research paper](https://www.sciencedirect.com/science/article/pii/S0021869323005720?via%3Dihub), but a built-in function would simplify the process and promote code reusability.

### Additional Information

_No response_

RuchitJagodara:
This patch is about implementing a minimum generating set function for group theory in Sage. The function implemented here operates in polynomial time, which makes it more interesting. Previously, we only knew of the trivial method, which was exponential in time. However, based on [research paper](https://www.sciencedirect.com/science/article/pii/S0021869323005720?via%3Dihub), we can implement  a 
 polynomial time algorithm for the minimum generating set function, and that's what this PR is about.

fixes #37362

### :memo: Checklist

### :hourglass: Dependencies

NA


RuchitJagodara:
@tscrim, can you please review this PR?

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
             (False, False, False)
         """
         return self.gap().IsomorphismGroups(H.gap()) != libgap.fail
+
+
+@cached_method
```
tscrim: This isn't a method. However, I don't think this should be cached. If something should be cached, it should be the method attached to the group.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+def minimum_generating_set(group, gap_based=False) -> set:
+    """
+    Return a list containing minimum generating set of the ``group``.
```
tscrim:
```suggestion
    Return a list of the minimum generating set of ``group``.
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+
+    INPUT:
+
+    - ``group`` -- a group object.
+    - ``gap_based`` -- boolean (default: False). If True, elements of the resulting list are GAP objects.
```
tscrim:
```suggestion
    - ``group`` -- a group
    - ``gap_based`` -- boolean (default: ``False``); if ``True``, elements
      of the resulting list are GAP objects
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+
+    OUTPUT:
+
+    A list of elements that generate the group.
+
+    REFERENCE:
+
+    For more information about the algorithm, see https://arxiv.org/abs/2306.07633 or
+    https://www.sciencedirect.com/science/article/pii/S0021869323005720?via%3Dihub.
```
tscrim:
```suggestion
    ALGORITHM:

    We follow :doi:`10.1016/j.jalgebra.2023.11.012` (:arxiv:`2306.07633`).
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+
+    EXAMPLES::
+
+        sage: G = GL(2,GF(3))
+        sage: s = G.minimum_generating_set();s
```
tscrim:
```suggestion
        sage: G.minimum_generating_set()
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+    A list of elements that generate the group.
+
+    REFERENCE:
+
+    For more information about the algorithm, see https://arxiv.org/abs/2306.07633 or
+    https://www.sciencedirect.com/science/article/pii/S0021869323005720?via%3Dihub.
+
+    EXAMPLES::
+
+        sage: G = GL(2,GF(3))
```
tscrim: These examples are best in the method (since that is what is called here) as that is the doc/examples most people will look at first. You should have one simple test here that direct imports the function and calls it.

Also, add a `.. SEEALSO::` pointing to the method for groups.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+        [1 0 0]  [1 1 0]
+        [0 1 1]  [0 1 0]
+        [0 0 1], [0 0 1]
+        ]
+
+    TESTS::
+
+        Check if the elements of the final list are able to generate the group or not::
+
+            sage: p = libgap.eval("DirectProduct(AlternatingGroup(5),AlternatingGroup(5))")
```
tscrim: These are over-indented.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +964,166 @@ def is_isomorphic(self, H):
...
+    from sage.libs.gap.element import GapElement
+
+    if not isinstance(group, GapElement):
+        converter = group._element_constructor_
+        group = group._libgap_()
+    elif not gap_based:
+        return minimum_generating_set(group, gap_based=True)
+
+    if not group.IsFinite().sage():
+        raise NotImplementedError("Implemented for finite group only.")
```
tscrim:
```suggestion
        raise NotImplementedError("only implemented for finite groups")
```
This should also have a doctest.

tscrim (changes requested):
I don't understand the `gap_based` argument. At the end, everything you are doing only works for groups implemented in GAP. So I think you need to handle the case when a group does not have a working `_libgap_` method more cleanly (e.g., `Permutations(5)`). I also don't see the utility in returning GAP objects; I think it should always return elements of the corresponding Sage group. This would also simplify the code (this could also be done by the method as a sort of "post-processing", with the function returning GAP objects in case someone does find a use for it).

Also, check for PEP8 spacing.

tscrim:
> > So I think you need to handle the case when a group does not have a working _libgap_ method more cleanly (e.g., Permutations(5))
> 
> But Permutations(5) is not a group, right? And this method is implemented for groups only.

It most certainly is a group. Moreover, it is a very special group, it is the symmetric group (on 5 letters).

> > I also don't see the utility in returning GAP objects; I think it should always return elements of the corresponding Sage group.
> 
> I was considering the case when a user is working with a GAP based group (a GAP Object) and wants to use this function, so I used the `gap_based` argument but I will remove it and update the function so that it will give the answer in whatever form user has given the group, like a Sage object or a GAP object. And here the `gap_based` argument was being used to reduce the number of conversions happening, if I do not use `gap_based` argument then function will convert elements of the resultant list into sage and in the next call, it will again be converted to a GAP object but I have found a way to tackle this problem and will implement that.

I doubt the (un)wrapping is such a major part of the time, but okay. In that case, I would *always* return a GAP object. Let the public user-facing method handle the conversion back to Sage objects. It is what knows exactly the type of objects it should return (specifically, elements of `self`).

RuchitJagodara:
> It most certainly is a group. Moreover, it is a very special group, it is the symmetric group (on 5 letters).

Yes, you are right; it satisfies the properties of a group. Initially, I implemented this method using sage functions only, but at that time, I was having trouble finding the quotient group of a group because sage has a working quotient group method but it does not give the exact quotient group; instead, it returns some group isomorphic to the quotient group, and I also noticed that most of the groups were using wrapper functions from gap so I thought it would be good if this implementation is done using GAP functions only, and also I was not able to find some functions that the current implementation is using.

So sticking with this implementation is good, and for groups like these (i.e. Permutations(5)) should I return a NotImplemented error? What are your thoughts?

RuchitJagodara:
> I doubt the (un)wrapping is such a major part of the time, but okay. In that case, I would always return a GAP object. Let the public user-facing method handle the conversion back to Sage objects. It is what knows exactly the type of objects it should return (specifically, elements of self).

Or we can just make an inner function which is going under recursion, and this inner function will give us the GAP object which we can convert to sage outside the inner function depending on the group given by the user.

tscrim:
I would raise a `NotImplementedError` when it doesn’t have a working `gap()` method (but, of course, one should try this method rather than checking types). Note that it would make sense to put the method in the category of groups (rather than to a specific implementation).

Your “inner” function is the function you’ve defined. It isn’t what the user is meant to call (it is the method attached to the group). Just change the function so that it always outputs GAP objects (and document this).

RuchitJagodara:
I have made all the necessary changes.

src/sage/groups/group.pyx
```diff
@@ -266,6 +267,52 @@ cdef class Group(Parent):
...
+
+        TESTS:
+
+            Test that function gives an error for infinite groups::
+
+                sage: G = GL(2, ZZ)
+                sage: G.minimum_generating_set()
+                Traceback (most recent call last):
+                ...
+                NotImplementedError: only implemented for finite groups
```
tscrim: This block is over-indented.

src/sage/groups/group.pyx
```diff
@@ -266,6 +267,52 @@ cdef class Group(Parent):
...
+        """
+        from sage.groups.libgap_mixin import minimum_generating_set
+        return [self._element_constructor(x) for x in minimum_generating_set(self)]
```
tscrim:
```suggestion
        return [self(x) for x in minimum_generating_set(self)]
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
             (False, False, False)
         """
         return self.gap().IsomorphismGroups(H.gap()) != libgap.fail
+
+
+def minimum_generating_set(group) -> list:
+    """
+    Return a list of the minimum generating set of ``group``. Elements of the list are
+    objects of GAP.
```
tscrim:
```suggestion
    Return a list of the minimum generating set of ``group``.
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+
+    INPUT:
+
+    - ``group`` -- a group
+
+    OUTPUT:
+
+    A list of elements that generate the group. Elements of the list are objects of GAP.
```
tscrim:
```suggestion
    A list of GAP objects that generate the group.
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+
+        :meth:`sage.groups.group.Group.minimum_generating_set`
+
+    ALGORITHM:
+
+    We follow :doi:`10.1016/j.jalgebra.2023.11.012` (:arxiv:`2306.07633`).
+
+    TESTS:
+
+        Test that the resultant list is able to generate the original group::
```
tscrim: All of these doctests are over-indented.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+            sage: s = minimum_generating_set(G); s
+            [(2,3), (1,3,2)(4,5)]
+            sage: s[0].parent()
+            C library interface to GAP
+    """
+    if not isinstance(group, GapElement):
+        try:
+            group = group._libgap_()
+        except AttributeError:
+            raise NotImplementedError("only implemented for groups that contain a working __libgap__ mehtod")
```
tscrim:
```suggestion
        try:
            group = group.GAP()
        except (AttributeError, ValueError, TypeError):
            raise NotImplementedError("only implemented for groups that can construct a gap group")
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+
+    if not group.IsFinite().sage():
+        raise NotImplementedError("only implemented for finite groups")
+
+    # A function to check if the group is generated by the given generators or not.
+    def is_groupbygens(group, gens):
+        if not isinstance(group, GapElement):
+            group = group._libgap_()
```
tscrim:
```suggestion
```
This shouldn’t be needed now since everything manipulated in here should be a GAP object.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+        try:
+            group = group._libgap_()
+        except AttributeError:
+            raise NotImplementedError("only implemented for groups that contain a working __libgap__ mehtod")
+
+    if not group.IsFinite().sage():
+        raise NotImplementedError("only implemented for finite groups")
+
+    # A function to check if the group is generated by the given generators or not.
+    def is_groupbygens(group, gens):
```
tscrim:
```suggestion
    def is_group_by_gens(group, gens):
```
Better to follow standard Python conventions.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+    l = len(g)
+
+    if N.IsAbelian().sage():
+        if is_groupbygens(group, g):
+            return g
+
+        for i in range(l):
+            for j in range(len(n)):
+                temp = g[i]
+                g[i] = g[i]*n[j]
```
tscrim:
```suggestion
                g[i] *= n[j]
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+    # of the group according to the algorithm.
+    def gen_combinations(g, N, t):
+        if t > len(g):
+            t = len(g)
+        if t <= 0:
+            yield g
+        for go in gen_combinations(g,N,t-1):
+            for j in range(len(N)):
+                temp = go[t-1]
+                go[t-1] = go[t-1]*N[j]
```
tscrim:
```suggestion
        for go in gen_combinations(g, N, t-1):
            for elt inN:
                temp = go[t-1]
                go[t-1] *= elt
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+                yield go
+                go[t-1] = temp
+
+    # Here, the value of t is different from the one in the research paper mentioned
+    # in the Algorithm section becuase in the paper the value of t was calculated using
+    # cheif series of the group but currently sage does not have cheif series implemented.
+    t = ceil((13/5 + log(group.Size().sage(), 2) / log(N.Size().sage(), 2)))
```
tscrim: You can access the chief series from GAP.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+    N_list = N.AsList()
+
+    for nl in N_list:
+        for raw_gens in gen_combinations(g, N_list, t):
+            if nl == N_list[0]:   # considering N_list[0] is the identity element
+                gens = raw_gens
+            else:
+                gens = raw_gens+[nl]
```
tscrim:
```suggestion
                gens = raw_gens + [nl]
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +965,128 @@ def is_isomorphic(self, H):
...
+
+            if is_groupbygens(group, gens):
+                return gens
```
tscrim: What happens if it happens to reach the end of the function? I think you should have an `assert` for that case (that should never happen, right?).

tscrim (changes requested):
You should also move the `minimal_generating_set` method into the category of `Groups` (as an `ParentMethod`) instead of as a specific method of `Group`. Tests should also be added for things without a gap group.

Note that `gap()` will (or at least should) return the corresponding libgap object.

src/sage/groups/libgap_mixin.py
RuchitJagodara: Yes, that will never happen according to the algorithm.

src/sage/groups/libgap_mixin.py
RuchitJagodara: Can I do that in another PR? Because calculating t is a big task and will take time. I am also busy making GSOC proposals and working on my coursework.

tscrim: No, it should be done here. Why is it such a big task? It's there in the paper, right? Otherwise you would need to *prove* that your bound is sufficient. It doesn't seem to be in the paper, but I haven't read it super carefully.

Also, it would be good if there was a bit more specific detail about where in the paper you are referencing, both here and in the `ALGORITHM:` section Actually, given the writing in the paper, it probably would be better to sketch out the algorithm in the `ALGORITHM:` section.

src/sage/groups/libgap_mixin.py
tscrim: You should add a message error saying that it should not fall into here.

fingolfin:
Very nice to see this new paper!

Just to say, on a purely mathematical note, there are two different notions of "polynomial time" when it comes to algorithmic group theory: polynomial in the order of the group, or polynomial in the order of a generating set. The algorithm here and in the referenced paper is polynomial in the former sense, but exponential in the latter. 

Anyway, note that GAP already implements efficient minimal generating set functions for finite solvable groups and various other special cases. All the examples highlighted here are covered by that, and usually handled in submilliseconds. The plots indeed seem to plot the logarithm of time spent, and only for really small inputs / times, making it not very representative, IMHO. It would be interesting to see comparison for inputs that are not trivial, and compare your algorithm to what is in GAP in the cases were GAP can deal with it -- and perhaps some examples in cases GAP can't deal with it currently (most non-solvable groups if it can't trivially find a generating set meeting some "obvious" lower bounds -- e.g. a generating set of size 2 of a non-solvable group is automatically minimal)

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +966,128 @@ def is_isomorphic(self, H):
...
+
+        for go in gen_combinations(g, N, t-1):
+            for elt in N:
+                temp = go[t-1]
+                go[t-1] *= elt
+                yield go
+                go[t-1] = temp
+
+    # Here, the value of t is different from the one in the research paper mentioned
+    # in the Algorithm section becuase in the paper the value of t was calculated using
```
fingolfin:
```suggestion
    # in the Algorithm section because in the paper the value of t was calculated using
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +966,128 @@ def is_isomorphic(self, H):
...
+    # cheif series of the group but currently sage does not have cheif series implemented.
```
fingolfin:
```suggestion
    # chief series of the group but currently sage does not have chief series implemented.
```

RuchitJagodara:
> Anyway, note that GAP already implements efficient minimal generating set functions for finite solvable groups and various other special cases. 

The function implemented here is minimum_generating_set, not "minimal". Also, This algorithm works for any finite group (both solvable and non-solvable). And sorry for the late response!

src/sage/groups/libgap_mixin.py
RuchitJagodara: I have changed the method, and now it does not require computing t; while revisiting the paper for the proof, I noticed that the paper is just proving that there exists some t which limits the number of checks which we have to do. But if we check the combinations in some specific order then we actually do not have to compute t and that is what I have done in this new commit and this is efficient also because it eliminates steps to calculate t.

RuchitJagodara: > Also, it would be good if there was a bit more specific detail about where in the paper you are referencing, both here and in the ALGORITHM: section Actually, given the writing in the paper, it probably would be better to sketch out the algorithm in the ALGORITHM: section.

I will complete this by tomorrow.

fingolfin:
The GAP function I referenced computes a generating set with size equal to the rank of the group. Isn't that what this function here does, too?

My comment already acknowledges that the GAP function does not deal with arbitrary groups (deliberately!), I asked for a comparison in cases where both apply.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,115 @@ def is_isomorphic(self, H):
...
+                return [ele]
+
+    if G.IsSimple().sage():
+        n = len(group_elements)
+        for i in range(n):
+            for j in range(i+1, n):
+                if is_group_by_gens(G, [group_elements[i], group_elements[j]]):
+                    return [group_elements[i], group_elements[j]]
+
+    N = G.MinimalNormalSubgroups()[0]
```
fingolfin: This is rather expensive... To give a simple benchmark, consider the group `DirectProduct(AlternatingGroup(50),AlternatingGroup(50))`.

The GAP function `MinimalGeneratingSet` (which, in your terminology, computes a "minimum" generating set) takes 2.7 seconds on my laptop to yield a generating set of size 2 (since this group is non-abelian, that proves it is indeed minimal).

In contrast, computing the minimal normal subgroups hasn't finished running after a few minutes.

Of course you don't need all minimal normal subgroups, just one. I haven't read your paper in detail but a quick glance reveals that you actually work with a chief series. Computing that is much cheaper; `ChiefSeries(G);` terminates in 2.5 seconds for that example group. So you may wish to tune your algorithm a bit.

RuchitJagodara:
> The GAP function I referenced computes a generating set with size equal to the rank of the group. Isn't that what this function here does, too?

Now, I have changed the method to try computing the minimum generating set with the MinimalGeneratingSet method of GAP. If it gives any error, it will follow this algorithm to cover all cases, including non-solvable groups. I have also added the algorithm in the Algorithm section according to the comment of @tscrim.

src/sage/groups/libgap_mixin.py
RuchitJagodara: I tried the algorithm with the below implementation of the minimal normal subgroup, which finds only one minimal normal subgroup, but this also takes almost the same time as before. 
```python
    # Find minimal normal subgroup N of G
    L = list(libgap.Elements(G))
    if len(L)==1:
        return None
    x = L[1]
    o = int(libgap.Order(x))
    p = 2
    q = None
    while p**2 <= o:
        if o%p == 0:
            q = o//p
            break
        p += 1
    if q is None:
        q = 1
        p = o
    x = x**q
    N = G.Subgroup([x])
    N = libgap.NormalClosure(G,N)
```



> So you may wish to tune your algorithm a bit.

This implementation is the most straightforward polytime implementation proposed by the paper. We can consider future optimisations, but can we go with this implementation for now because it works for almost every group? If I change the implementation, I am not sure what will be the time taken by the new implementation (the one using the chief series).

src/sage/categories/groups.py
```diff
@@ -451,6 +451,54 @@ def conjugacy_class(self, g):
             from sage.groups.conjugacy_classes import ConjugacyClass
             return ConjugacyClass(self, g)
 
+        def minimum_generating_set(self):
+            """
+            Return a list of the minimum generating set of this group.
```
tscrim, line 456:
```suggestion
            Return a list of a minimum generating set of this group.
```
Since it is not necessarily unique.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,133 @@ def is_isomorphic(self, H):
...
+
+    A list of GAP objects that generate the group.
+
+    .. SEEALSO::
+
+        :meth:`sage.categories.groups.Groups.ParentMethods.minimum_generating_set`
+
+    ALGORITHM:
+
+    We follow the algorithm described in the research paper "A New Algorithm for Finding the Minimum Generating Set of a Group" by John Doe (:doi:`10.1016/j.jalgebra.2023.11.012`).
```
tscrim: Please use a proper reference `[Doe2023]_`, which your temp author should be replaced accordingly.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,133 @@ def is_isomorphic(self, H):
...
+    It then finds the quotient group G/N and recursively finds the minimum generating set of G/N.
+    Let S be the minimum generating set of G/N. The algorithm finds representatives g of S in G.
+
+    If N is abelian, the algorithm checks if any case of the form g_1, g_2, ..., g_i*s_j, g_i+1, ..., g_lg
+    (where lg is the length of g) is able to generate G. If not, it uses the set g_1, g_2, ..., g_lg, s_j
+    as the minimum generating set of G.
+
+    If N is non-abelian, the algorithm checks if any case of the form g_1*n_1, g_2*n_2, ..., g_lg*n_lg
+    is able to generate G, where n_1, n_2, ..., n_lg can be any elements of N. If not, it checks if any
+    case of the form g_1*n_1, g_2*n_2, ..., g_lg*n_lg, n_lg+1 is able to generate G.
```
tscrim: Please use latex formatting for all of this, e.g.
```
`g_1, g_2, \ldots, g_n`
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,133 @@ def is_isomorphic(self, H):
...
+                g[i] = temp
+
+        return g + [S[0]]
+
+    # A function to generate some combinations of the generators
+    # of the group according to the algorithm. Here it is considered that
+    # the first element of the group N is the identity element.
+    def gen_combinations(g, N, lg):
+        for iter in product(N, repeat=lg):
+            yield [g[i] * iter[lg-i-1] for i in range(lg)]
```
tscrim:
```suggestion
        for it in product(N, repeat=lg):
            yield [g[i] * it[lg-i-1] for i in range(lg)]
```
Let's not use Python builtin functions as variable names.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,133 @@ def is_isomorphic(self, H):
...
+        for ele in group_elements:
+            if is_group_by_gens(G, [ele]):
+                return [ele]
+
+    if G.IsSimple().sage():
+        n = len(group_elements)
+        for i in range(n):
+            for j in range(i+1, n):
+                if is_group_by_gens(G, [group_elements[i], group_elements[j]]):
+                    return [group_elements[i], group_elements[j]]
```
tscrim:
```suggestion
        from itertools import combinations
        for a, b in combinations(group_elements, 2):
            if is_group_by_gens(G, [a, b]):
                return [a, b]
```
This should be a bit faster of a check. At the very least, it is a bit shorter code.

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,133 @@ def is_isomorphic(self, H):
...
+        try:
+            G = G.gap()
+        except (AttributeError, ValueError, TypeError):
+            raise NotImplementedError("only implemented for groups that can construct a gap group")
+
+    if not G.IsFinite().sage():
+        raise NotImplementedError("only implemented for finite groups")
+
+    def is_group_by_gens(group, gens):
+        return set(group.AsList()) == set(libgap.GroupByGenerators(gens).AsList())
```
tscrim: This seems awfully slow. At the very least, we should cache the set of elements in the group. It might also be good to check some easier to compute things first, such as the cardinalities, before checking all elements. Some timings should be done.

src/sage/groups/libgap_mixin.py
tscrim: @RuchitJagodara, you should follow @fingolfin's advice. You should also profile your code to see where time is being spent. Right now, GAP's implementation is apparently completely beating yours, so it should be the preferred one unless you have data suggesting otherwise.

tscrim (changes requested):

RuchitJagodara:
@fingolfin, can you please review the PR now? I have updated the algorithm, which now uses the chief series approach. The changes have made the algorithm very fast now, so thank you for your advice, and I also apologise for the delay.

Also, I am very thankful to @pranav-joshi-iitgn, who helped me with this PR.

RuchitJagodara:
As, @tscrim told we have used latex format for documenting the algorithm but it is now showing some kind of linting error and also failing the build documentation test, so can you please help us with this?

src/sage/groups/libgap_mixin.py
```diff
@@ -968,175 +968,171 @@ def is_isomorphic(self, H):
         """
         return self.gap().IsomorphismGroups(H.gap()) != libgap.fail
 
-def minimum_generating_set(G: GapElement) -> list:
-    r"""
```
pranav-joshi-iitgn: Probably not a good idea to make it a normal string, rather than a raw string considering all the LaTeX in the docstring. It leads to linting errors like these : 
```
./sage/groups/libgap_mixin.py:983:39: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:983:64: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:983:67: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:986:64: W291 trailing whitespace
./sage/groups/libgap_mixin.py:1002:12: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1002:21: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1002:30: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:1002:41: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1002:58: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1002:72: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:1006:13: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1006:22: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1006:31: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:1006:56: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1006:71: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1006:84: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:1016:12: W605 invalid escape sequence '\l'
./sage/groups/libgap_mixin.py:1016:18: W605 invalid escape sequence '\l'
./sage/groups/libgap_mixin.py:1016:35: W605 invalid escape sequence '\i'
./sage/groups/libgap_mixin.py:1016:57: W291 trailing whitespace
./sage/groups/libgap_mixin.py:1017:13: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1017:23: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1017:54: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1017:59: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:1022:21: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1022:35: W605 invalid escape sequence '\i'
./sage/groups/libgap_mixin.py:1023:12: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1023:39: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1023:63: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1023:72: W605 invalid escape sequence '\}'
./sage/groups/libgap_mixin.py:1027:21: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1027:47: W605 invalid escape sequence '\i'
./sage/groups/libgap_mixin.py:1028:12: W605 invalid escape sequence '\{'
./sage/groups/libgap_mixin.py:1028:39: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1028:63: W605 invalid escape sequence '\d'
./sage/groups/libgap_mixin.py:1028:72: W605 invalid escape sequence '\}'
```

pranav-joshi-iitgn (commented):
There can be a bit less of LaTeX in the ALGORITHM section. Something like this maybe ? : 
```
    ALGORITHM:

    First we cover the cases when we can use the ``MinimalGeneratingSet``
    method of GAP directly. This happens when `G` is either Simple,
    or Solvable, or Nilpotent. Then, if ``libgap.MinimalGeneratingSet``
    fails, we proceed assuming that `G` is not a simple group.
    So, we are guaranteed to find a chief series of length at least 2.

    `S := ChiefSeries(G) = [G,G_1,G_2 ... G_l]` where `G_l = \{ e \}`

    Let `g` be the set of representatives of the minimum generating set
    of `G/G_1`. This can be found easily (since `G/G_1` is simple group)
    as ``g = libgap.MinimalGeneratingSet(GbyG1)``

    for k = 2 to `l` , compute `G/G_k` , `G_{k-1}/G_k` and set
    `g := lift( g, (G_{k-1}/G_k) , (G/G_k) )`

    return `g`

    `lift` function details:

    It computes the minimum generating set (as representative elements) of
    a quotient group `G/G_i` in a chief series, given the
    minimum generating set (as representatives) of `G/G_{i-1}`, namely `g`
    (what we are calling ``G_by_Gim1_mingen_reps`` in the code).
    the factor group `G_{i-1}/G_i` and the quotient group `G/G_i` itself.
    The function does these steps:

    First, we compute some essential quantities:

    `n :=\{ n_1,n_2 ... n_k \}`
    where `\{ n_1 G_i,n_2 G_i ... n_k G_i \}` is any generating set of
    `G_{i-1}/G_i` , i.e. it's the representative elements of any prefferably
    small, but not necessarily minimal generating set of `G_{i-1}/G_i`

    `N := \{ N_1,N_2 ... N_m \}` where
    `G_{i-1}/G_i = \{N_1 G_i,N_2 G_2 ... N_m G_m \}`.
    This is simply a list of representative elements of `G_{i-1}/G_i`.

    We wish to find the representatives of a minimum generating set of `G/G_i`.
    Here, we have two cases to consider.

    First, if `G_{i-1}/G_i` is abelian :

    if `<gG_i> = G/G_i`, return `g`

    for `0 < p < s+1`  and `n_j` in `n`, we calculate
    `g^* := \{ g_1,g_2 ... g_{p-1} ,g_p n_j,g_{p_1}, ... \}`.
    If `<g^* G_i> = G/G_i` , return `g^*`

    Second, if `G_{i-1}` is not abelian:

    First, for all combinations of (not necessarily distinct) elements
    `N_{i_1},N_{i_2}... N_{i_t}` in `N`, compute
    `g^* = \{g_1N_{i_1},g_{i_2}N_{i_3}... g_{i_t}N_t,g_{t+1}... g_s\}`
    (This is done using the ``gen_combinations`` generator).
    If `\{x G_i | x \in g^* \}` generates `G/G_i`, return `g^*`

    Then, for all combinations of (not necessarily distinct) elements
    `N_{i_1},N_{i_2}... N_{i_t} N_{i_{t+1}}` in `N`, compute
    `g^* = \{g_1N_{i_1},g_{i_2}N_{i_3}... g_{i_t}N_t,g_{t+1}... g_s\}`
    (This is done using the ``gen_combinations`` generator).
    If `\{x G_i | x \in g^* \}` generates `G/G_i`, return `g^*`

    By now, we must have exhausted our search.
```

src/sage/groups/libgap_mixin.py
RuchitJagodara: I have converted the docstring from normal string to raw string; thank you for your suggestion!

RuchitJagodara: But still doctests are failing, any idea why ?

src/sage/categories/groups.py
fingolfin: What does "a list of a set" mean? Perhaps instead like this?
```suggestion
            Return a minimum generating set of this group.
```
If you want to emphasize that it returns a list object and not a set object, I'd be explicit about that, e.g. "a min. gen. set, represent by a list of group elements" (though IMHO that's unnecessary)

src/sage/groups/group.pyx
```diff
@@ -266,6 +266,7 @@ cdef class Group(Parent):
         """
         raise NotImplementedError
 
+
```
fingolfin, line 269:
```suggestion
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+
+        :meth:`sage.categories.groups.Groups.ParentMethods.minimum_generating_set`
+
+    ALGORITHM:
+
+    We follow the algorithm described in the research paper "Algorithms for
+    the minimum generating set problem" by Bireswar Das and Dhara Thakkar (:doi:`10.48550/arXiv.2305.08405`).
+
+    When group ``G`` is a simple, solvable or nilpotent then we directly use
+    the MinimalGeneratingSet funtion from GAP which gives us the minimum generating set
```
fingolfin:
```suggestion
    the MinimalGeneratingSet function from GAP which gives us a minimum generating set
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+    of that group.
+
+    If the MinimalGeneratingSet function of GAP gives any error, then it is guaranteed
+    that the group ``G`` is not a simple group and it will have a cheif series of length 2.
```
fingolfin, line 997:
```suggestion
    that the group ``G`` is not a simple group and it will have a chief series of length at least 2.
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+
+    `S := ChiefSeries(G) = [G,G_1,G_2 \dots G_l]` where `G_l = \{e\}`
+
+    Let `g` be the set of representatives of the minimum generating set of `G/G_1`.
```
fingolfin: "the" versus "a" everywhere, I won't mark it all, please go through the text yourself looking for that

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+    .. SEEALSO::
+
+        :meth:`sage.categories.groups.Groups.ParentMethods.minimum_generating_set`
+
+    ALGORITHM:
+
+    We follow the algorithm described in the research paper "Algorithms for
+    the minimum generating set problem" by Bireswar Das and Dhara Thakkar (:doi:`10.48550/arXiv.2305.08405`).
+
+    When group ``G`` is a simple, solvable or nilpotent then we directly use
```
fingolfin: Note that nilpotent implies solvable so listing it in this sentence is redundant.
```suggestion
    When the group ``G`` is simple or solvable then we directly use
```

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+
+    It computes the minimum generating set (as representative elements) of a quotient group `G/G_i`
+    in a chief series, given the minimum generating set (as representatives) of `G/G_{i-1}`, namely `g`
+    (what we are calling ``G_by_Gim1_mingen_reps`` in the code). the factor group `G_{i-1}/G_i`
+    and the quotient group `G/G_i` itself. The function does these steps:
+
+    First, we compute some essential quantities:
+
+    `{n} :=\{n_1,n_2\dots n_k\}` where `\{n_1 G_i,n_2G_i \dots n_kG_{i}\}` is any generating set of
+    `G_{i-1}/G_i , i.e. it's the representative elements of any prefferably small, but not
```
fingolfin:
```suggestion
    `G_{i-1}/G_i , i.e. it is the representative elements of any preferably small, but not
```
Just a typo fix and a style suggestion, but even with that this sentence reads weird to me, and I don't like the expression "representative elements" (I don't know what really is meant by it)

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+    necessarily minimal generating set of `G_{i-1}/G_i`
+
+    `{N} := \{N_1,N_2\dots N_m\}` where `G_{i-1}/G_i = \{N_1G_i,N_2G_2\dots N_m G_m\}`.
+    This is simply a list of representative elements of `G_{i-1}/G_i`.
```
fingolfin: What are "representative elements" of a quotient `G/N` ??

src/sage/groups/libgap_mixin.py
```diff
@@ -963,3 +967,180 @@ def is_isomorphic(self, H):
...
+        2
+
+    Test that elements of resultant list are GAP objects::
+
+        sage: from sage.groups.libgap_mixin import minimum_generating_set
+        sage: G = PermutationGroup([(1,2,3), (2,3), (4,5)])
+        sage: s = minimum_generating_set(G); s
+        [(2,3), (1,3,2)(4,5)]
+        sage: s[0].parent()
+        C library interface to GAP
```
fingolfin, line 1067: Why does it say that here?

fingolfin (commented):
Note that I am not a member of the SageMath team, and normally don't do reviews here (nor do my reviews here carry any weight), but since you asked, here is some feedback. Overall I found the explanation of the algorithm difficult to follow (and gave up trying to do so) as there were too many phrases I found confusing and unclear. Once you clarified you abundant use of "representative objects" and fixed the "a" vs "the" issues it hopefully will be better.

RuchitJagodara:
Thank you for your feedback, I will try to make it better.

tscrim:
> Note that I am not a member of the SageMath team, and normally don't do reviews here (nor do my reviews here carry any weight), but since you asked, here is some feedback. Overall I found the explanation of the algorithm difficult to follow (and gave up trying to do so) as there were too many phrases I found confusing and unclear. Once you clarified you abundant use of "representative objects" and fixed the "a" vs "the" issues it hopefully will be better.

Thank you for taking the time to look at this all the same. I am not an expert in computational group theory, so it is good to have someone who understands what parts to look at and where complexity can arise. Additionally, since we have a long standing policy of "anyone can review", your review carries the same weight as one from me.

If I had to guess, "representative objects" means coset representatives.

@RuchitJagodara You should compile the doc (unfortunately our CI is currently broken, so it cannot be done through here like before and must be done locally) and look at that. There are lots of missing commas and ill-formatted portions from what I can see. Also, I strong advise people to avoid `\dots`; it can be wrong and it is always better to be specific (e.g., `\ldots`, `\cdots`).

src/sage/categories/groups.py
```diff
@@ -451,6 +451,51 @@ def conjugacy_class(self, g):
...
+                Symmetric group of order 3! as a permutation group
+
+                sage: A5 = AlternatingGroup(5)
+                sage: A5.minimum_generating_set()
+                [(1,2,3,4,5), (3,4,5)]
+
+                sage: H = groups.matrix.Heisenberg(1,3); H
+                Heisenberg group of degree 1 over Ring of integers modulo 3
+                sage: H.minimum_generating_set()
+                [
```
pranav-joshi-iitgn, line 480: I got this as the output instead :
```
[
[1 0 0]  [1 1 0]
[0 1 1]  [0 1 0]
[0 0 1], [0 0 1]
]
```
Although it's the same thing, I think it's good to update it.

RuchitJagodara:
@tscrim, can you please help me with the failing build and test check, I am not able to understand what is the exact problem here.

tscrim:
> @tscrim, can you please help me with the failing build and test check, I am not able to understand what is the exact problem here.

The CI testing is currently broken.

RuchitJagodara:
Oh, right! Can you please review the PR and tell me if it needs any further modifications?

pranav-joshi-iitgn:
For anyone interested, I have [implemented the same algorithm in GAP](https://github.com/gap-system/gap/pull/5716#pullrequestreview-2062648739), which I felt is a better place to put it since it "completes" `MinimalGeneratingSet` by making it usable for non solvable (finite) groups.

RuchitJagodara:
@pranav-joshi-iitgn have you opened a pr or contributed your gap code into gap, if yes then I think it would be better to close this pr and if not then I request @tscrim to review this pr.

pranav-joshi-iitgn:
> @pranav-joshi-iitgn have you opened a pr or contributed your gap code into gap, if yes then I think it would be better to close this pr and if not then I request @tscrim to review this pr.

The code is still in the process of being reviewed. The version of GAP that sage uses needs to be updated for the method `MinimalGeneratingSet` to be usable for any finite group, not just solvable or simple ones.

PR #37490, opened 2024-02-27
Include `random_element()` method to `LaurentPolynomialRing`

### LINKED ISSUE #37454: Method `.random_element` not impl'ed for`LaurentPolynomialRing`

GiacomoPope:
This PR is an attempt to fix issue https://github.com/sagemath/sage/issues/37454 to add a method for random sampling for the `LaurentPolynomialRing` class which is causing failure in some random testings.

This is not an object I have much familiarity with, but essentially I have followed the usual random sampling for multivariate polynomial rings and tried to follow the existing function name and arguments.

```py
sage: R = LaurentPolynomialRing(QQ, 2, 'x')
sage: R.random_element()
-2*x0*x1^-1 - 2*x0^-2*x1^2 + 5/2*x0^-1 + x0^-2*x1
sage: R.random_element(-5, 0)
7/61*x0^-3*x1^-4 + 1/3*x0^-4*x1^-4 - 1/2*x0^-5*x1^-4
sage: R.random_element(-5, 10)
1/31*x0^-2*x1^7 - 7*x0^-5*x1^10 + 4*x0^2*x1^-2 + 1/3*x0^-2*x1 - 3/44*x0*x1^-5
```

Included in the doctests is some random testing which tests the properties I have tried to include in this function:

```py
        TESTS::
            sage: rings = [QQ, ZZ, GF(13), GF(7^3)]
            sage: for ring in rings:
            ....:     d = randint(1, 5)
            ....:     R = LaurentPolynomialRing(ring, d, 'x')
            ....:     for _ in range(100):
            ....:         n, m = randint(0, 10), randint(0, 10)
            ....:         f = R.random_element(-n, m)
            ....:         for x in R.gens():
            ....:             assert f.degree(x) <= m
            ....:             assert f.degree(x) >= -n
```

Fixes https://github.com/sagemath/sage/issues/37454


src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,99 @@ def krull_dimension(self):
...
+          The maximal valuation of monomial in the polynomial
+
+        - ``terms`` -- number of terms requested (default: 5). If more
+          terms are requested than exist, then this parameter is
+          silently reduced to the maximum number of available terms.
+
+        - ``choose_degree`` -- choose degrees of monomials randomly first
+          rather than monomials uniformly random.
+
+        - ``**kwargs`` -- passed to the random element generator of the base
```
grhkm21:
```suggestion
        - ``*args, **kwds`` -- passed to the random element generator of the base
```

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,99 @@ def krull_dimension(self):
         """
         raise NotImplementedError
 
-    def random_element(self, low_degree=-2, high_degree=2, terms=5, choose_degree=False,*args, **kwds):
+    def random_element(self, low_degree=-2, high_degree=2, terms=5, choose_degree=False, *args, **kwds):
         """
+        Return a random polynomial of degree at most ``high_degree`` with
+        lowest degree at most ``low_degree``.
```
grhkm21:
```suggestion
        Return a random polynomial of degree at most ``high_degree`` and
        lowest degree at least ``low_degree``.
```

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,99 @@ def krull_dimension(self):
...
+        - ``high_degree`` -- non-negative integer (default: 2).
+          The maximal valuation of monomial in the polynomial
+
+        - ``terms`` -- number of terms requested (default: 5). If more
+          terms are requested than exist, then this parameter is
+          silently reduced to the maximum number of available terms.
+
+        - ``choose_degree`` -- choose degrees of monomials randomly first
+          rather than monomials uniformly random.
+
```
grhkm21:
```suggestion
```
I think it's better to put these into `*args, **kwds` instead, since in this method you don't use them anyways, you just pass them to `self._R.random_element` directly.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,99 @@ def krull_dimension(self):
...
+            ....:             assert f.degree(x) <= m
+            ....:             assert f.degree(x) >= -n
+        """
+        # Ensure the low_degree is non-positive
+        if low_degree > 0:
+            raise ValueError("`low_degree` is expected to be a non-positive integer")
+
+        # Ensure the high_degree is non-negative
+        if high_degree < 0:
+            raise ValueError("`low_degree` is expected to be a non-negative integer")
```
grhkm21: Why are these checks done? I imagine someone might still want `LR.random_element(low_degree=3, high_degree=5)`.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,99 @@ def krull_dimension(self):
...
+
+        # First sample a polynomial from the associated polynomial
+        # ring of `self` of degree `(high_degree + abs(low_degree))`
+        abs_deg = (high_degree + abs(low_degree))
```
grhkm21:
```suggestion
        abs_deg = (high_degree - low_degree)
```
(related to the change above)

grhkm21 (changes requested):
small changes

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
GiacomoPope: ahh ok, yeah I can do this.

GiacomoPope:
I have kept the additional argument which were suggested to be pruned as these seem to be used by people who like multivariate polynomial rings so maybe they are interesting here.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,109 @@ def krull_dimension(self):
...
+            ....:         for x in R.gens():
+            ....:             assert f.degree(x) <= m
+            ....:             assert f.degree(x) >= n
+        """
+        # Ensure the degree parameters are sensible
+        if high_degree < low_degree:
+            raise ValueError("`high_degree` must be greater than or equal to `low_degree`")
+
+        # First sample a polynomial from the associated polynomial
+        # ring of `self` of degree `(high_degree + abs(low_degree))`
```
grhkm21:
```suggestion
        # ring of `self` of degree `(high_degree - low_degree)`
```

grhkm21 (approved):
I still think the arguments should be removed, but I don't have strong opinions.

GiacomoPope:
I think if someone wants to use this function but is unaware of the underlying randomness function then the extra arguments help the user. But I can also see the reason to remove them.

grhkm21:
> I think if someone wants to use this function but is unaware of the underlying randomness function then the extra arguments help the user. But I can also see the reason to remove them.

Yeah but the same can be said for e.g. `QQ["x"].random_element(den_bound=100)` or `ZZ["x"].random_element(x=-10,y=10)`, you just have to know they exist. But yeah I don't mind keeping it, and there's doctests

GiacomoPope:
I suppose the difference there is you can construct these rings with many different base rings / fields, but the underlying polynomial ring is *always* a multivariate polynomial ring (even for the case with one generator haha)

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,109 @@ def krull_dimension(self):
...
+        - ``high_degree`` -- integer (default: ``2``).
+          The maximal valuation of monomial in the polynomial
+
+        - ``terms`` -- number of terms requested (default: ``5``).
+          If more terms are requested than exist, then this parameter is
+          silently reduced to the maximum number of available terms.
+
+        - ``choose_degree`` -- bool (default: ``False``)
+          choose degrees of monomials randomly first rather than monomials
+          uniformly random.
```
tscrim:
```suggestion
        - ``terms`` -- (default: ``5``) number of terms requested;
          if more terms are requested than exist (from specifying
          the valuation and degree), then this parameter is silently
          reduced to the maximum number of available terms

        - ``choose_degree`` -- bool (default: ``False``) choose
          degrees of monomials randomly first rather than monomials
          uniformly random
```

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,109 @@ def krull_dimension(self):
...
+        :meth:`sage.rings.polynomial.multi_polynomial_ring_base.MPolynomialRing_base.random_element`
+        then shifts this polynomial down to compute the correct degrees.
+
+        INPUT:
+
+        - ``low_degree`` -- integer (default: ``-2``).
+          The smallest valuation of monomial in the polynomial
+
+        - ``high_degree`` -- integer (default: ``2``).
+          The maximal valuation of monomial in the polynomial
```
tscrim:
```suggestion
        - ``min_valuation`` -- integer (default: ``-2``); the
          minimal allowed valuation of the polynomial

        - ``max_degree`` -- integer (default: ``2``); the
          maximal allowed degree of the polynomial
```

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +489,109 @@ def krull_dimension(self):
...
 
-    def random_element(self, low_degree=-2, high_degree=2, terms=5, choose_degree=False,*args, **kwds):
+    def random_element(self, low_degree=-2, high_degree=2, terms=5, choose_degree=False, *args, **kwds):
         """
+        Return a random polynomial of degree at most ``high_degree`` and
+        lowest degree at least ``low_degree``.
+
+        Internally uses the random sampling from
+        :meth:`sage.rings.polynomial.multi_polynomial_ring_base.MPolynomialRing_base.random_element`
+        then shifts this polynomial down to compute the correct degrees.
```
tscrim:
```suggestion
        Internally uses the random sampling from the corresponding
        polynomial ring then shifts this polynomial down to compute
        the correct degrees.
```
It may not use that method depending on the underlying polynomial ring.

tscrim (changes requested):
It is very confusing to the have `degree` referring to the valuation. Essentially degree and valuation are opposite ends of the allowable support (in the standard monomial basis). So we should change the argument names.

You're wrong about the multivariate polynomial rings:
```
sage: L.<x> = LaurentPolynomialRing(QQ)
sage: L.polynomial_ring()
Univariate Polynomial Ring in x over Rational Field
```
This comes from the construction of polynomial rings (there is a cautionary note about this in that doc IIRC). You should add some examples using this.

The only reason I can think of for keeping the arguments is that this method then sets a default value. However, it does force all underlying implementations to handle these arguments. So I would be in favor of absorbing them into `*args` and `**kwds`.

You should also add some examples over more "exotic" base rings to catch more implementations. For example
```
sage: L.<x> = LaurentPolynomialRing(s)
sage: type(L.polynomial_ring())
<class 'sage.rings.polynomial.polynomial_ring.PolynomialRing_integral_domain_with_category'>
sage: L.<x,y,z> = LaurentPolynomialRing(s)
sage: type(L.polynomial_ring())
<class 'sage.rings.polynomial.multi_polynomial_ring.MPolynomialRing_polydict_with_category'>
```

Lastly, the code comment change still is needed.

grhkm21:
I won't give positive reviews anymore :)

tscrim:
Oh the melodrama. *faints* :P

But more seriously, I do appreciate the work you (both!) are putting into helping improve Sage. I'm just picky and overly pedantic about some things. >.> <.<

grhkm21:
I will clarify that what I meant is I don't have enough experience with reviewing to make sure it's good for Sage, so I won't give them out so easily :)

> Oh the melodrama. *faints* :P

Not sure what that means.

GiacomoPope:
Yeah I thought the argument names were a little odd but I'm not an expert in this area and I figured whoever wrote the original function had a good idea of what to do. 

Thanks for spotting the issue with the univariate case! I'll ensure I handle this better in the sampling as I assume this will break my current code. 

Thanks for looking over the PR. Off all my recent contributions this one really is outside of the tools I regularly use!

GiacomoPope:
Thanks for the reviews and comments. I have tried to implement everything but I cannot do the exotic cases as they do not have their own `.random_element()` methods.

For example:

```py
sage: s
Symmetric Functions over Rational Field in the Schur basis
sage: s.random_element()
NotImplementedError                       Traceback (most recent call last)
Cell In[3], line 1
----> 1 s.random_element()

File ~/sage/sage/src/sage/categories/modules_with_basis.py:1412, in ModulesWithBasis.ParentMethods.random_element(self, n)
   1410 if not indices.is_empty():
   1411     for i in range(n):
-> 1412         a += self.term(indices.random_element(),
   1413                        self.base_ring().random_element())
   1414 return a

File ~/sage/sage/src/sage/categories/infinite_enumerated_sets.py:63, in InfiniteEnumeratedSets.ParentMethods.random_element(self)
     49 def random_element(self):
     50     """
     51     Raise an error because ``self`` is an infinite enumerated set.
     52 
   (...)
     61     TODO: should this be an optional abstract_method instead?
     62     """
---> 63     raise NotImplementedError("infinite set")

NotImplementedError: infinite set
```

GiacomoPope:
Open question:

For the univariate case we have the `valuation()` function for the type `sage.rings.polynomial.laurent_polynomial.LaurentPolynomial_univariate`

```py
    def valuation(self, p=None):
        """
        Return the valuation of ``self``.

        The valuation of a Laurent polynomial `t^n u` is `n` plus the
        valuation of `u`.

        EXAMPLES::

            sage: R.<x> = LaurentPolynomialRing(ZZ)
            sage: f = 1/x + x^2 + 3*x^4
            sage: g = 1 - x + x^2 - x^4
            sage: f.valuation()
            -1
            sage: g.valuation()
            0
        """
        return self.__n + self.__u.valuation(p)
```

But for the multivariate case we get something of type `sage.rings.polynomial.laurent_polynomial_mpair.LaurentPolynomial_mpair` which has a `degree()` method but nothing for `valuation()`.

I could add this in for this PR if people think this is helpful?

GiacomoPope:
Another comment:

I'm not totally happy with how the sampling is done... At the moment the underlying polynomial returns a sum of monomials for which each term has degree at most `degree`. However, I *always* shift by a term of `prod(gens)**min_valuation` which has degree `gens * min_valuation`. As a result, when there are many gens in the ring we almost always get every term with negative valuation:

```py
sage: R = LaurentPolynomialRing(GF(2), 4, 'x')
sage: RR = R._R
sage: RR.random_element(degree=10)
x0*x1*x2^5*x3
sage: RR.random_element(degree=10)
x0^2*x1^8 + x0*x2^3*x3^5
sage: RR.random_element(degree=10)
x0*x1^5*x3^3 + x0*x2^5*x3 + x0^4*x2^2
sage: 
sage: R.random_element(-10, 10)
x0^-9*x1^8*x2^-10*x3^-10 + x0^-2*x1^-6*x2^-9*x3^-4
sage: R.random_element(-10, 10)
x0^-8*x1^-9*x2^-2*x3^-6
sage: R.random_element(-10, 10)
x0^-7*x1^-10*x2^-1*x3^-8 + x0^-1*x1^-9*x2^-10*x3^-9
sage: R.random_element(-10, 10)
x0^-6*x1^-4*x2^-10*x3^-7
```

I'm not sure this works how someone would expect it to (even if technically the bounds are being treated properly)

GiacomoPope:
I have pushed an additional function for the valuation of multivariate polynomial elements of the `LaurentPolynomialRing` -- i have left some TODO as I'm not sure I'm being as efficient as possible with the cython stuff. Review is appreciated.

The benefit with `valuation()` is we can now have better doctests for this `random_element()`.

My other comment above has not been addressed as I don't quite know what to do about getting a better distribution of positive and negative expo. in the random elements.

GiacomoPope:
Here is an example which is maybe more closely aligned with expectation but it uses less code reuse...

```py
def new_idea(R, min_val=-2, max_deg=2, terms=5):
    n = R.ngens()
    res = R.zero()
    for _ in range(terms):
        s = R.base_ring().random_element()
        ele = [randint(min_val,max_deg) for _ in range(n)]
        res += s * R.monomial(*ele)
    return res
```

GiacomoPope:
Pushed the new idea for the above, expanded doctests and the output seems much nicer:

```py
sage: R = LaurentPolynomialRing(GF(2), 4, 'x')
sage: R.random_element()
x1^2*x2*x3^-1 + x0^-2*x1^2*x2 + x0^-2*x2*x3^-2 + x0^-2*x1^-2*x2^2*x3^-1 + x0^-1*x1^-2*x2^-2*x3^-2
sage: R.random_element()
x0^2*x1^-1*x2*x3^2 + x0^-1*x1*x2^2 + x0^-2*x2^-2
sage: R.random_element()
0
sage: R.random_element()
x0^-2*x2^2*x3^2 + x0^-1*x1^-1*x2^2 + x0^2*x1^-1*x2^-1*x3^-2
sage: R.random_element()
x0^2*x1^2*x2^-2*x3^-2 + x0^-2*x1^-1*x3^2 + x0^-1*x1^-1*x3^-1
```

tscrim:
`QQbar` is fine (the other thought I had was the universal cyclotomic field). Another example would be over a non-domain:
```
sage: I = R.ideal([x^2+x*y+y^2,x+y,x^3,y^3])
sage: Q = R.quotient(I)
sage: Q.is_integral_domain()
False
sage: S.<a,b,c> = PolynomialRing(Q)
```

Now with regards to the valuation. For the univariate case, there is nothing special that needs to be done. However, as you have noticed you cannot multiply all of the variables together and raise that to `min_valuation`. That gives a valuation of `n * min_valuation`, where `n` is the number of variables. A simple solution for the multivariate case would be to sample a random (nonnegative) integer vector of length `n` of sum `min_valuation` and scale by the corresponding monomial. Then everything else can be redirected to the polynomial.

I am not sure how much sense it makes to take the valuation with respect to a single variable, but I guess we can leave it in.

Rather that `sum(e)`, it might be faster to do `e.unweighted_degree()`, but the result should be cast to `ZZ` (it would be an `int` otherwise). You also should not check `g[i] is x` but `g[i] == x`. The rest of the Cython stuff looks good.

Lastly, error messages should not end with a period/full-stop. Plus it is good to have tests/examples for the error messages too.

I likely won't be able to respond for a little bit (traveling from Japan to Europe).

GiacomoPope:
> Another example would be over a non-domain:

I believe that the Laurent Polynomial Ring needs to have an integral domain as input. Using the example above I get a value error:

```py
sage: I = R.ideal([x^2+x*y+y^2,x+y,x^3,y^3])
sage: Q = R.quotient(I)
sage: Q.is_integral_domain()
False
sage: S.<a,b,c> = PolynomialRing(Q)
sage: LaurentPolynomialRing(S, 2, 'x')
...
ValueError: base ring must be an integral domain
```

Otherwise, thank you for all the comments. I have correct the errors and added doctests. I have kept the `min_val` and `max_deg` referring to the degree/valuation of the generators rather than the polynomial, if this really upsets someone they can change it but this seems totally usable for randomised testing etc.

> I likely won't be able to respond for a little bit (traveling from Japan to Europe).

Enjoy your travels!

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +488,199 @@ def krull_dimension(self):
...
+        if self._n == 1:
+            # First sample a polynomial from the associated polynomial
+            # ring of `self` of degree `(max_degree - min_valuation)`
+            abs_deg = (max_degree - min_valuation)
+            f_rand = self._R.random_element(degree=abs_deg, **kwds)
+
+            # Case the polynomial base to self and scale to ensure
+            # that min_valuation is satisfied
+            s = self.gen() ** min_valuation
+            return self(f_rand) * s
```
tscrim:
```suggestion
            f_rand = self(f_rand)
            try:
                return f_rand.shift(min_valuation)
            except AttributeError:
                return self.gen() ** min_valuation * f_rand
```
This should be implemented for all univariate cases (admittedly, this would be hard to do for something allowed to be multivariate...) and be faster. However, we do need the fallback.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +488,199 @@ def krull_dimension(self):
...
+        # We now sample `terms`` terms with exponents picked randomly
+        # with degree at most `max_degree` and valuation greater or
+        # equal to min_valuation scaled by an element of the base ring
+        k = self.base_ring()
+        n = self._n
+        res = self.zero()
+        for _ in range(terms):
+            s = k.random_element(**kwds)
+            ele = [randint(min_valuation, max_degree) for _ in range(n)]
+            res += s * self.monomial(*ele)
```
tscrim: This could lead to something having a lower valuation than `min_valuation` if you're unlucky. (Although the probability of such an event is likely quite close to 0 since you would have to chose the `min_valuation` or close to it many times in succession.) Note that for random multivariate polynomials, the `degree` is computing the *total* degree. Likewise, we should have the valuation done in the same way. So you just need to use
```
sage: IntegerVectors(5, 3).random_element()
[1, 4, 0]
```
to get the monomial to shift the random polynomial by.

tscrim:
Ah, it probably needs the integral domain because it is constructed as a localization. (Now this can be extended to that case, but one needs to be a lot more careful and Sage doesn't have that implemented (yet).)

tscrim:
Do you have an example of two (Laurent) polynomial that fail one of the conditions in https://en.wikipedia.org/wiki/Discrete_valuation? I can't think of any right now (but that might be my jet lag). Or are you saying that it is failing the conditions for a DVR and the method name is misleading?

I think anything with `degree` is likely to cause confusion since that is (basically) always "maximal".

Side note: I vaguely remember talking about these kind of things with Martin when doing the lazy Laurent series (although we didn't do multivariate there because it doesn't even form a ring unless additional conditions are imposed).

mantepse:
> Side note: I vaguely remember talking about these kind of things with Martin when doing the lazy Laurent series (although we didn't do multivariate there because it doesn't even form a ring unless additional conditions are imposed).

However, we *did* do multivariate power series, and define the valuation as minimal total degree.  We *did not* put them into `Category of discrete valuation rings`, unless they are univariate, though.

nbruin:
> Do you have an example of two (Laurent) polynomial that fail one of the conditions in https://en.wikipedia.org/wiki/Discrete_valuation? I can't think of any right now (but that might be my jet lag). Or are you saying that it is failing the conditions for a DVR and the method name is misleading?

Yes:
```
sage: R.<x,y>=LaurentPolynomialRing(QQ)
sage: def mindeg(a):
....:     return min([min(v) for v in a.exponents()])
....: 
sage: a=x/y
sage: b=y/x
sage: mindeg(a)
-1
sage: mindeg(b)
-1
sage: mindeg(a*b)
0
sage: mindeg(a*b) == mindeg(a) + mindeg(b)
False
```

as you can see, mindeg fails to be multiplicative. And that's fundamentally because the maximal ideal is not principal.

> I think anything with `degree` is likely to cause confusion since that is (basically) always "maximal".

with local monomial orderings, the leading term would be of minimal total degree (for most choices, I think), so perhaps leading_degree or degree_of_leading_term if you don't like minimal_degree ?

GiacomoPope:
Just to note: I define the valuation here as the valuation of the term with the smallest valuation in the generator (when the generator is None I look for the term with the lowest valuation)

This seems to obey all the properties linked by Travis so it is *some* kind of valuation?

GiacomoPope:
Oh! Thanks for this example. It's very obvious after you see it but this is definitely helpful for me.

tscrim:
@nbruin Thank you for the example. That's very helpful.

hmmm... `leading_degree` would be acceptable.

I guess that brings up another question: Do we want a `valuation` method for multivariate Laurent polynomials? I think we do as it also gets used for a univariate case. @nbruin Do you think it should it error out when trying to get a multivariate valuation?

mantepse:
I'm a bit confused now: in which cases do you think that a method name `valuation` is acceptable?  Is minimal total degree OK?

In the lazy code, we need that concept, even though it is not always a valuation, if I'm not mistaken.  I cannot quite remember in which cases it is.

Would you prefer `order` for the concept of minimal total degree?

GiacomoPope:
Could we instead have the valuation of the polynomial as `return min([sum(v) for v in a.exponents()])` which would be the mirror of degree?

I also think f.valuation(x) for a generator x is well defined? In the above example we would get -1, 1 then 0. 

I cannot "work" on this for a while today but it occurred to me that rather than min min the valuation should be min sum?

tscrim:
@GiacomoPope `min(sum(e) …)` is analogous to (total) degree for polynomials. Actually, with that definition, @nbruin’s example is still multiplicative (everything is degree and valuation `0 = 0 + 0`). Following up further, aren’t (x) and (x,y) both the unit ideal since they contain $1 = x^{-1} x = y^{-1} y$? I am a bit confused now too. Perhaps there is some miscommunication here. Am I missing something basic here?

Additionally, shouldn’t it be `trailing_degree` instead of `leading_degree` since the leading monomial is the largest degree typically?

GiacomoPope:
Ok I have an idea for refactoring-- I'll implement it this evening once my daughter is asleep 😴

GiacomoPope:
Hello!

I have found some times while I have a sleeping house. 

1. I have updated the `valuation()` for the multivariate case to be `min(sum(e) for e in self.exponents()` and I hope this solves some of the above discussion?
2. I have rewritten how I sample the random terms for the multivariate case.

I now sample a term such that every generator has exponent between `min_valuation <= e <= max_degree` and I have also ensured that `sum(self.exponents())` for each term also obeys the limits. 

Maybe an example is clearer:

```py
            sage: L = LaurentPolynomialRing(GF(13), 3, 'x')
            sage: f = L.random_element(-5, 10)
            sage: f.degree() <= 10
            True
            sage: f.valuation() >= -5
            True
            sage: tuple(f.degree(x) <= 10 for x in L.gens())
            (True, True, True)
            sage: tuple(f.valuation(x) >= -5 for x in L.gens())
            (True, True, True)
```

We now have the bounds respected for all generators `x` as well as the polynomial itself, for which the `degree()` and `valuation()` are the `max/min` of the sums of exponents for each term

GiacomoPope:
oh god something terrible has happened with the commit history here.

nbruin:
OK, it took a while but I now agree that min_total_degree is indeed a valuation; and in fact one that is relatively canonically connected to `K[[x,y]]`. The function field `k(x,y)` is the function field of the plane and `K[[x,y]]` is the completion of the local ring at `(x,y)=(0,0)`. And most of the time, we think of K((x,y)) as the field of fractions of that. Problem: the (prime, maximal) ideal `(x,y)=(0,0)` is not a length one ideal (its locus is not codimension 1) so the local ring does not come with a discrete valuation.

However, we can *blow up* at (0,0) and then we do have a divisor above (0,0). Blow-up is a birational operation, so the resulting variety has a function field together with an isomorphism to the original function field.

In this case, when we blow up the point (0,0), we get affine charts (x,y,t) with xt=y and (x,y,u) with yu=x. So we get function fields k(x,t) isomorphic to k(x,y=xt) and k(y,u) isomorphic to k(x=yu,y). In the (x,t)-plane the curve x=0 is the one that maps to (x,y)=(0,0) and in the (y,u)-plane it's y=0. We have that k(t)[[y]] is a complete DVR and k(u)[[x]] as well. They induce the same valuation on k(x,y).

So, min_total_deg induces the valuation that you get from k((y/x))[[y]] .

This does generalize to higher-dimensional cases as well.

On a Laurent series ring this is I guess a rather distinguished valuation. On a laurent polynomial ring, it's only distinguished if you think of the laurent polynomial ring embedded in the laurent series ring.

GiacomoPope:
At the moment when the bounds are "bad" I allow exponents for certain generators to be zero even when this is outside the bounds to satisfy the total bounds. 

However another option is to add https://github.com/sagemath/sage/pull/37513 as a dependency and when the bounds cannot be satisfied for all exponents (likely when both bounds are positive or negative) we could simply return the zero polynomial which would have degree -Inf and valuation +Inf and all bounds would be properly respected.

mantepse:
Hi Nils!

> OK, it took a while but I now agree that min_total_degree is indeed a valuation; and in fact one that is relatively canonically connected to `K[[x,y]]`. The function field `k(x,y)` is the function field of the plane and `K[[x,y]]` is the completion of the local ring at `(x,y)=(0,0)`. And most of the time, we think of K((x,y)) as the field of fractions of that. Problem: the (prime, maximal) ideal `(x,y)=(0,0)` is not a length one ideal (its locus is not codimension 1) so the local ring does not come with a discrete valuation.
> 
> However, we can _blow up_ at (0,0) and then we do have a divisor above (0,0). Blow-up is a birational operation, so the resulting variety has a function field together with an isomorphism to the original function field.

I think that roughly here my mathematical understanding comes to an end :-)

> In this case, when we blow up the point (0,0), we get affine charts (x,y,t) with xt=y and (x,y,u) with yu=x. So we get function fields k(x,t) isomorphic to k(x,y=xt) and k(y,u) isomorphic to k(x=yu,y). In the (x,t)-plane the curve x=0 is the one that maps to (x,y)=(0,0) and in the (y,u)-plane it's y=0. We have that k(t)[[y]] is a complete DVR and k(u)[[x]] as well. They induce the same valuation on k(x,y).

I'll re-enter the conversation roughly here :-)
 
> So, min_total_deg induces the valuation that you get from k((y/x))[[y]] .

> This does generalize to higher-dimensional cases as well.

I'm not completely sure what you mean with *higher-dimensional cases*.  More variables?

> On a Laurent series ring this is I guess a rather distinguished valuation. On a laurent polynomial ring, it's only distinguished if you think of the laurent polynomial ring embedded in the laurent series ring.

As far as I understand, multivariate Laurent series don't make much sense, at least if you are interested in formal series.  Does this have any consequence for the discussion about valuations?

nbruin:
> As far as I understand, multivariate Laurent series don't make much sense, at least if you are interested in formal series. Does this have any consequence for the discussion about valuations?

Yes, apologies. I was thinking about multivariate power series rings and their fields of fractions. The elements there are not necessarily "multivariate" Laurent series. That's actually a rather direct consequence of k[[x_1,..,x_n]] not being a DVR for n >=2.

However, the discussion above suggests that k[[x_1,...,x_n]] does have a relatively natural valuation on it that coincides with what you'd get from considering the ring as a subring of
k[[x_1/x_n,...,x_{n-1}/x_n]][[x_n]]

This undoubtedly is know, described and used somewhere in the literature already

As you can see, given a field of fractions K of k[[x_1/x_n,...,x_{n-1}/x_n]], we can now get the field of fractions as K((x_n)). So it would give a more symmetric way of representing a field of fractions of a multivariate power series ring in terms of iterated Laurent series. Given that it actually has a reasonable theoretical interpretation, it may actually be useful for computational work too.

tscrim:
@GiacomoPope I think this is way more complicated than it needs to be. You should be able to do
```python
def random_element(self, min_val=-2, max_deg=2, *args, **kwds):
    tot_deg = max_deg - min_val
    poly = self._R.random_element(tot_deg, *args, **kwds)
    poly = self(poly)
    if min_val == 0:
        return poly
    mon = self.monomial(*IntegerVectors(abs(min_val), self.ngens()).random_element())
    if min_val < 0:
        return ~mon * poly
    return mon * poly
```
and get a reasonable sampling. The other advantage it has is that it avoids duplicating stuff of the random polynomials. (Special casing $n = 1$ gens might still be useful here.)

Although perhaps both of these approaches are not so good. The dimension of the degree $k$ space is always infinite for any fixed $k$ provided there are at least $n = 2$ generators. However, we both only will sample from a finite dimensional subspace of each degree $k$ portion. Perhaps we should also include into the exponent of `mon` (in the above snippet) an extra random element in $\mathbb{Z}^n$ that sums to 0 by sampling an element in $\mathbb{Z}^{n-1}$ and then adding the correction at the end?

GiacomoPope:
Yeah this maybe is fine. It will allow for the polynomial to have degree and valuation outside of the bounds for a particular generator though. Maybe that's fine?

I think I need to have the opinion of what someone really wants from a method like this? What I have already seems fine and not overly complicated-- your implementation reuses parts of sage more which is a benefit. 

I really implemented this to help out from the bug we found in the CI and have no big opinions on what it should be.

tscrim:
It's fine to go outside for a single generator because the degree/valuation is not defined for a single generator. It is defined for the total sum of the exponents. We don't have to go outside of it, but it needs to be mentioned in the doc since it will not sample every element. (Since we are sampling from something infinite (dimensional) rather than finite, things are more delicate and restricting to finite pieces is useful.)

GiacomoPope:
The degree and valuation are defined for a single generator. The degree was already and I matched this in my implementation of valuation. 

This behaviour of degree is exactly why I attempted to bound both the exponent and their sum.

tscrim:
Be aware of differences between $R[x][y]$ and $R[x,y]$. If you define degree on a single generator, say $y$, then you are in the former case. If it is the total degree, then it is the later. This is natural because you want to view things as graded $R$-algebras (or at least, that is how I have typically seen them). IMO, Sage allows a user to restrict to a particular generator as a convenience. (Note the default/doc of `degree()` for a multivariate polynomial.)

The same holds for valuations.

GiacomoPope:
Yes I'm aware of the difference but what I mean to say is it's not obvious for me from the current documentation whether f.degree() or f.degree(x) is the main usage. Note that the documentation (although x is by default None) seems to point towards the latter. 

As I am not an expert in this area, my PR aims to appease both of these bounds as best as I could. 

If you think only the total degree and valuation should be respected then OK, but it's not obvious at all to me about what is "correct".

mantepse:
> However, the discussion above suggests that k[[x_1,...,x_n]] does have a relatively natural valuation on it that coincides with what you'd get from considering the ring as a subring of k[[x_1/x_n,...,x_{n-1}/x_n]][[x_n]]

Isn't this "relatively natural valuation" simply the minimal total degree?

> As you can see, given a field of fractions K of k[[x_1/x_n,...,x_{n-1}/x_n]], we can now get the field of fractions as K((x_n)). So it would give a more symmetric way of representing a field of fractions of a multivariate power series ring in terms of iterated Laurent series. Given that it actually has a reasonable theoretical interpretation, it may actually be useful for computational work too.

sorry, that's a bit too abstract for me - at least right now.  (is it really "a field of fractions"? - i.e., could there be several different fields of fractions?)

GiacomoPope:
@tscrim I have removed my complication and now return something very similar to what you suggest above. Thank you for the pointer to `IntegerVectors` which is essentially what my other function was needing.

I have adjusted the tests and they all pass.

In my opinion this is "fine" and if someone uses this function in the future and needs more out of it, they can adjust it?

src/sage/rings/polynomial/laurent_polynomial_ring.py
```diff
@@ -47,6 +47,7 @@
 from sage.rings.polynomial.laurent_polynomial_ring_base import LaurentPolynomialRing_generic
 from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
 from sage.structure.element import parent
+from sage.misc.prandom import randint, shuffle
```
tscrim:
```suggestion
```
This can be removed.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1183,6 +1183,61 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
             raise TypeError("x must be a generator of parent")
         return self._poly.degree(self._parent._R.gens()[i]) + self._mon[i]
 
+    def valuation(self, x=None):
+        """
+        Return the valuation of ``x`` in ``self``.
+
+        If ``x`` is ``None``, return the minimal valuation of ``self``.
```
tscrim:
```suggestion
        If ``x`` is ``None``, return the valuation of ``self`` using the total degree.
```
Although I am more of the opinion this should be removed and replaced with something like
```
INPUT:

- ``x`` -- (optional) a generator; if given, return the valuation with respect to this generator
```
Using the total degree (= sum of exponents) is likely going to be the most common input.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1183,6 +1183,61 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+        # Get the index of the gen
+        cdef tuple g = <tuple > self._parent.gens()
+        cdef Py_ssize_t i
+        cdef bint no_generator_found = True
+        for i in range(len(g)):
+            if g[i] == x:
+                no_generator_found = False
+                break
+        if no_generator_found:
+            raise TypeError("x must be a generator of parent")
```
tscrim:
```suggestion
        try:
            I = g.index(x)
        except ValueError:  # not in the tuple
            raise TypeError(f"{x} is not a generator")
```
This can be simplified (and is likely faster).

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1183,6 +1183,61 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+        """
+        # Valuation of zero polynomial is defined to be +Infinity
+        if self.is_zero():
+            return Infinity
+
+        # When x is None find the minimal valuation by finding the minimal
+        # valuation of the sum of exponents
+        if x is None:
+            val = min(sum(e) for e in self.exponents())
+            return val
```
tscrim:
```suggestion
            return ZZ(min(sum(e) for e in self.exponents()))
```
We should make sure to cast the result to `ZZ` (don’t want Python `int`’s to escape out).

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +488,145 @@ def krull_dimension(self):
...
+        abs_deg = (max_degree - min_valuation)
+        f_rand = self._R.random_element(degree=abs_deg, *args, **kwds)
+
+        # Case this polynomial back the `self``
+        f = self(f_rand)
+
+        # For the univariate case we simply shift by x**min_valuation
+        if self._n == 1:
+            s = self.gen() ** min_valuation
+            return f * s
```
tscrim: I think it would be better to first try if there is a `shift` method since that will be faster then fallback to multiplication.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +488,145 @@ def krull_dimension(self):
...
+
+        # For the multivariate case, we sample a single monomial of degree
+        # exactly min_valuation and then use this to shift the polynomial
+        # into the correct bounds
+        s = self.monomial(*IntegerVectors(abs(min_valuation), self._n).random_element())
+        s = self(s)
```
tscrim:
```suggestion
```
This cast is unnecessary since `self.monomial` will return an element of `self`.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +488,145 @@ def krull_dimension(self):
...
+        """
+        # Ensure the degree parameters are sensible
+        if max_degree < min_valuation:
+            raise ValueError("`max_degree` must be greater than or equal to `min_valuation`")
+
+        # Sample a polynomial in the base ring of degree `max_degree - min_valuation`
+        abs_deg = (max_degree - min_valuation)
+        f_rand = self._R.random_element(degree=abs_deg, *args, **kwds)
+
+        # Case this polynomial back the `self``
```
tscrim:
```suggestion
        # Cast this polynomial to ``self``
```

tscrim (commented):
@mantepse Yes, that’s correct. @nbruin is giving an algebro-geometric description. I think we are all okay now with calling the method `valuation` and the argument name for the random element as such.

@GiacomoPope It would be good to also have a test/example where the min valuation and max degree are both positive and both negative. It is covered by the random choices, but it is better to make sure it is tested every time. Otherwise I think this is good for now.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
GiacomoPope: I'm confused by this review as my current code has the following:

```py
        # For the multivariate case, we sample a single monomial of degree
        # exactly min_valuation and then use this to shift the polynomial
        # into the correct bounds
        s = self.monomial(*IntegerVectors(abs(min_valuation), self._n).random_element())
        if min_valuation < 0:
            s = ~s
        return f * s
```

Which was taken pretty much verbatim from your above comment.

GiacomoPope: Ahh i see the comment below so I'll just work from that

GiacomoPope:
Thank you for another review, I really appreciate it. I should have captured everything you mentioned and also hardcoded some of the bounds in doctests to allow both positive/negative.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1172,17 +1172,68 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+        cdef tuple g = <tuple > self._parent.gens()
+        cdef Py_ssize_t i
+        try:
+            i = g.index(x)
+        except ValueError:  # not in the tuple
+            raise TypeError(f"{x} is not a generator of parent")
+
+        # Find the minimal valuation of x by checking each term
+        return min(e[i] for e in self.exponents())
+
```
tscrim:
```suggestion
```
Only 1 blankline between methods.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1172,17 +1172,68 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+        # Get the index of the generator or error
+        cdef tuple g = <tuple > self._parent.gens()
+        cdef Py_ssize_t i
+        try:
+            i = g.index(x)
+        except ValueError:  # not in the tuple
+            raise TypeError(f"{x} is not a generator of parent")
+
+        # Find the minimal valuation of x by checking each term
+        return min(e[i] for e in self.exponents())
```
tscrim:
```suggestion
        return Integer(min(e[i] for e in self.exponents()))
```
Just some extra insurance.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
```diff
@@ -488,16 +488,150 @@ def krull_dimension(self):
...
 
-            sage: LaurentPolynomialRing(QQ, 2, 'x').random_element()
+            sage: L.<x> = LaurentPolynomialRing(QQ)
+            sage: f = L.random_element()
+            sage: f.degree() <= 2
+            True
+            sage: f.valuation() >= -2
+            True
+            sage: f.parent() is L
+            True
```
tscrim, line 519: Can you add the same test but with `L.<x> = LaurentPolynomialRing(QQ, 1)` (this won't have the `shift` method for the elements)?

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1172,17 +1172,68 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+            raise TypeError(f"{x} is not a generator of parent")
         return self._poly.degree(self._parent._R.gens()[i]) + self._mon[i]
 
+    def valuation(self, x=None):
+        """
+        Return the valuation of ``self``.
+
+        INPUT:
+
+        - ``x`` -- (optional) a generator; if given, return the valuation with respect to this generator
```
tscrim:
```suggestion
        - ``x`` -- (optional) a generator; if given, return the valuation
          with respect to this generator
```
80 char / line (PEP8).

tscrim (changes requested):
Thank you. Just a few last details.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
GiacomoPope: Yes, but I moved it to TEST rather than examples as I don't think it's good to encourage the construction of this weird "multivariate" polynomial in one generator. Is this OK? Also, this case was previously handled with my randomised testing.

nbruin:
@mantepse: Apologies for this relatively off-topic tangent. There are some elements here that may be good to get addressed in the documentation concerning the "valuation" terminology, though, so I think there is relevance the larger sagemath context of this ticket. Plus, you're asking :-)

> Isn't this "relatively natural valuation" simply the minimal total degree?

Yes. The issue I had is that multivariate power series rings are not discrete valuation rings. That doesn't say they don't admit valuations, but it makes one suspicious if there is a `valuation` method that doesn't take an argument to specify which valuation is to be taken. Indeed, as we see, the field of fractions of k[[x_1,...,x_n]] does have a valuation on it that corresponds to minimal total degree, but the ring of elements of non-negative valuation is considerably bigger than k[[x_1,...,x_n]]. As I described, there is an algebraic-geometric interpretation of where valuations on function fields are coming from and I did identify what that means for the minimal total degree. That one is in hindsight clear to me, but it was certainly not when I started thinking about it. So a hint onthe justification of the name "valuation" may be appropriate (for discoverability as well ... I wonder if it is the best name if its justification is not so obvious)

The terminology on Laurent polynomial rings is clearly all directly derived from what one gets for power series rings. That's why I'm looking there for justification. Given that Laurent polynomial rings are just subrings of rational function fields, the source for valuations is quite different (richer actually), so it's even stranger that Laurent polynomial rings would have a distinguished valuation on them. Hence, I think the above considerations apply to Laurent polynomial rings as well.

> sorry, that's a bit too abstract for me - at least right now. (is it really "a field of fractions"? - i.e., could there be several different fields of fractions?)

No, that's fine. There is a canonical way to construct a field of fractions for an integral domain. Giventhat a multivariate power series ring can just be constructed in an iterative way, by taking k[[x_1]][[x_2]]...[[x_n]], a natural construction is
k((x_1)((x_2))...((x_n)) [which would be quite horrible to compute in]. Computationally there are big differences between the various constructions, particularly because one would have to compute with truncations much of the time. But mathematically, they represent the same object.

Note that this last construction does have a fairly obvious valuation on it (valuation wrt. x_n), but that is not the one that multivariate laurent polynomial rings use!

GiacomoPope:
If you would like some of the above included in the docstrings in some (condensed) way, I'm happy to make the edit but I need guidance on what would be useful in this context. My work was really only meant to generalise in some way what we had in the univariate case for testing the random element but it seems as thought there's much more maths here than I appreciated.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1172,17 +1172,68 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+        try:
+            i = g.index(x)
+        except ValueError:  # not in the tuple
+            raise TypeError(f"{x} is not a generator of parent")
         return self._poly.degree(self._parent._R.gens()[i]) + self._mon[i]
 
+    def valuation(self, x=None):
+        """
+        Return the valuation of ``self``.
+
```
nbruin, line 1187: I think "valuation" deserves a description here, since it is not clear at all what the valuation of a laurent polynomial should be. Perhaps something like:
"""
If no argument is given, or None is given, the returned valuation is the minimal total degree of the monomials occurring in self. Geometrically, this is the order of vanishing of self at the generic point of the blow-up of the point (0,0,...,0).

If an argument x is given then it  must be a generator. In that case, the minimum degree of that generator occurring in the polynomial is returned. Geometrically, this is the order of vanishing of self at the generic point of the curve $x=0$.
"""

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1172,17 +1172,68 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+        if self.is_zero():
+            return Infinity
+
+        # When x is None find the minimal valuation by finding the minimal
+        # valuation of the sum of exponents
+        if x is None:
+            return Integer(min(sum(e) for e in self.exponents()))
+
+        # Get the index of the generator or error
+        cdef tuple g = <tuple > self._parent.gens()
```
nbruin, line 1237: this makes me think that an index of a generator would also be valid input. It's much more direct.

Argument against that would be that in principle, any irreducible polynomial could be valid input, since it would describe a prime divisor at which a valuation can be computed. Whether that needs to be implemented is another matter.

GiacomoPope: Hmm this is a good point. I think if we do this for `valuation()` then we must also do this for `degree()` which has an almost identical form.

I'm happy to do the change which allows `x` to be either a generator or index. Does anyone else have an opinion?

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
GiacomoPope: Thanks so much for the help here. The latest commit adds this extra text to the docstring.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
tscrim: Given how `degree()` is implemented, I would vote (at least for now) to leave it as it is currently implemented.

src/sage/rings/polynomial/laurent_polynomial_ring_base.py
tscrim: Deterministic tested > random testing. It is annoying to figure out why a test fails only some of the time (when it is not based on the element).

tscrim:
@nbruin Thank you for the detailed explanations.

@GiacomoPope It was clearly good for us to have this discussion as we clarified some of the mathematics and improved the documentation. Unfortunately it wasn't quite as simple of a fix as expected, but it's better for it. :)

tscrim (approved):
I don't think so. If @nbruin @grhkm21 @mantepse or someone else has any other changes they want, feel free to revert the positive review.

tscrim:
@GiacomoPope Sorry, one extra thing (for a followup PR that would depend on this). We should also take into account the grading for the valuation with a `std_grading` flag...

GiacomoPope:
Oh yeah, this should indeed follow the way degree behaves... I don't think I'll have time today but if you wanted to do the PR I could maybe review it tonight.

To make matters worse the degree for this case is also broken:

```py
sage: R.<x, y> = LaurentPolynomialRing(ZZ, order=TermOrder("wdegrevlex", [1, 3]))
sage: R(x).degree()
1
sage: R(y).degree()
3
sage: R(1/x).degree()
-1
sage: R(1/y).degree()
-1
```

It just doesnt deal with negative exponents properly so in fact the weighting for `LaurentPolynomialRing` is just broken.

PR #37492, opened 2024-02-27
make sagelib work with singular>=4.3.2.p15 (future 4.4)

antonio-rojas:
Two major breaking changes:

- The `rp` ordering is renamed to `ip`
- `hilb` now returns a `bigintvec` object (which isn't even exposed in the C++ API, seems to be an instance of `bigintmat`)

This branch builds and works fine with 4.3.2.p15, but I have no idea how to make it work with both old and new singular, suggestions welcome.

### DIFF
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -2025,6 +2025,10 @@ class SingularElement(ExtraTabCompletion, ExpectElement, sage.interfaces.abc.Sin
             sage: type(singular(int(5)).sage())
             <class 'sage.rings.integer.Integer'>
 
+        Test that bigintvec can be coerced::
+
+            sage: singular('hilb((ideal(x)), 1)').sage()
+            (1, -1, 0, 0, -1, 1, 0)
         """
         typ = self.type()
         if typ == 'poly':
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -1405,8 +1405,9 @@ class SingularElement(ExtraTabCompletion, ExpectElement, sage.interfaces.abc.Sin
         # compatibility for singular 4.3.2p10 and before
         if s.startswith("polynomial ring,"):
             from sage.rings.polynomial.term_order import singular_name_mapping
+            from sage.repl.rich_output import get_display_manager
             # this is our cue that singular uses `rp` instead of `ip`
-            if singular_name_mapping['invlex'] == 'rp':
+            if singular_name_mapping['invlex'] == 'rp' and 'doctest' in str(get_display_manager()):
                 s = re.sub('^(// .*block.* : ordering )rp$', '\\1ip',
                            s, 0, re.MULTILINE);
         return s
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -1402,6 +1402,13 @@ class SingularElement(ExtraTabCompletion, ExpectElement, sage.interfaces.abc.Sin
         if self._name in s:
             if self.get_custom_name() is None and self.type() == 'matrix':
                 s = self.parent().eval('pmat(%s,20)' % (self.name()))
+        # compatibility for singular 4.3.2p10 and before
+        if s.startswith("polynomial ring,"):
+            from sage.rings.polynomial.term_order import singular_name_mapping
+            # this is our cue that singular uses `rp` instead of `ip`
+            if singular_name_mapping['invlex'] == 'rp':
+                s = re.sub('^(// .*block.* : ordering )rp$', '\\1ip',
+                           s, 0, re.MULTILINE);
         return s
 
     def __copy__(self):
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -1206,18 +1206,15 @@ class Singular(ExtraTabCompletion, Expect):
 
         EXAMPLES::
 
-            sage: import re
             sage: r = PolynomialRing(GF(127),3,'xyz', order='invlex')
-            sage: rsing = r._singular_()
-            sage: print(re.sub('ordering rp', 'ordering ip', rsing._repr_()))
+            sage: r._singular_()
             polynomial ring, over a field, global ordering
             //   coefficients: ZZ/127
             //   number of vars : 3
             //        block   1 : ordering ip
             //                  : names    x y z
             //        block   2 : ordering C
-            sage: curring = singular.current_ring()
-            sage: print(re.sub('ordering rp', 'ordering ip', curring._repr_()))
+            sage: singular.current_ring()
             polynomial ring, over a field, global ordering
             //   coefficients: ZZ/127
             //   number of vars : 3
@@ -2470,20 +2467,6 @@ def singular_version():
         "Singular ... version 4...
     """
     return singular.eval('system("--version");')
-
-
-def singular_version_number():
-    """
-    Return the version number of Singular being used as a string.
-
-    EXAMPLES::
-
-        sage: singular.version_number()
-        '4...'
-    """
-    import re
-    r = re.compile(r"\((\d+),")
-    return r.findall(singular_version())[0]
 
 
 class SingularGBLogPrettyPrinter:
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -1206,15 +1206,18 @@ class Singular(ExtraTabCompletion, Expect):
 
         EXAMPLES::
 
+            sage: import re
             sage: r = PolynomialRing(GF(127),3,'xyz', order='invlex')
-            sage: r._singular_()
+            sage: rsing = r._singular_()
+            sage: print(re.sub('ordering rp', 'ordering ip', rsing._repr_()))
             polynomial ring, over a field, global ordering
             //   coefficients: ZZ/127
             //   number of vars : 3
             //        block   1 : ordering ip
             //                  : names    x y z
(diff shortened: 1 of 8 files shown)


antonio-rojas:
cc @tornaria @kiwifb

tornaria:
This is working fine for me and I'm releasing sagemath 10.3 on void linux with this patch and singular 4.3.2.p16 (with flint 3.1.2)

mkoeppe:
best to remove "WIP:" from the title if it's ready

antonio-rojas:
> best to remove "WIP:" from the title if it's ready

I don't think this is ready unless we want to break support for older singular.

mkoeppe:
Then probably "positive review" should be replaced by "needs work"

src/sage/libs/singular/singular.pyx
```diff
@@ -1699,6 +1699,25 @@ cdef object si2sa_intvec(intvec *v) noexcept:
         l.append(v.get(r))
     return tuple(l)
 
+cdef object si2sa_bigintvec(bigintmat *v) noexcept:
```
tornaria:
```suggestion
cdef object si2sa_bigintvec(bigintmat *v):
```

src/sage/libs/singular/singular.pxd
```diff
@@ -29,6 +29,7 @@ cdef object   si2sa_ZZmod(number *n, ring *_ring, object base) noexcept
 cdef object   si2sa_NF(number *n, ring *_ring, object base) noexcept
 
 cdef object si2sa_intvec(intvec *v) noexcept
+cdef object si2sa_bigintvec(bigintmat *v) noexcept
```
tornaria:
```suggestion
cdef object si2sa_bigintvec(bigintmat *v)
```

Also, after #37667 you'll need to adjust the context by removing the `noexcept` in lines 29, 31, 35.

The rule is: functions returning python objects cannot be `noexcept`. Cython 3.0 ignores it and starting with 3.0.9 it emits a warning; it could be an error in cython 3.1.

antonio-rojas:
First attempt at making this work with older singular

src/sage/interfaces/singular.py
```diff
@@ -1206,19 +1206,22 @@ def current_ring(self):
...
+            sage: import re
             sage: r = PolynomialRing(GF(127),3,'xyz', order='invlex')
-            sage: r._singular_()
+            sage: rsing = r._singular_()
+            sage: print(re.sub('ordering rp', 'ordering ip', rsing._repr_()))
             polynomial ring, over a field, global ordering
             //   coefficients: ZZ/127
             //   number of vars : 3
-            //        block   1 : ordering rp
+            //        block   1 : ordering ip
```
tornaria, line 1214:
```suggestion
            //        block   1 : ordering ...
```
wouldn't this make it simpler?

src/sage/libs/singular/decl.pxd
```diff
@@ -244,6 +244,7 @@ cdef extern from "singular/Singular/libsingular.h":
         ringorder_lp
         ringorder_dp
         ringorder_rp
+        ringorder_ip
```
tornaria, line 253: Can we just keep using `ringorder_rp` for the time being since they seem to have a compat `#define ringorder_rp ringorder_ip` ?

src/sage/rings/polynomial/term_order.py
```diff
@@ -386,9 +386,15 @@
...
+from sage.interfaces.singular import singular_version_number
+if int(singular_version_number()[0:3]) < 432 or (int(singular_version_number()[0:3]) == 432 and int(singular_version_number()[3:]) < 15):
+    invlex_singular_name = 'rp'
+else:
+    invlex_singular_name = 'ip'
+
 singular_name_mapping = {
     'lex'           : 'lp',
-    'invlex'        : 'rp',
+    'invlex'        : invlex_singular_name,
```
tornaria: Seems to be the only place we need a different code for different versions of singular.

src/sage/rings/polynomial/multi_polynomial_libsingular.pyx
```diff
@@ -1280,30 +1280,36 @@ cdef class MPolynomialRing_libsingular(MPolynomialRing_base):
...
+            sage: import re
             sage: R = PolynomialRing(GF(2**8,'a'),10,'x', order='invlex')               # needs sage.rings.finite_rings
-            sage: singular(R)                                                           # needs sage.rings.finite_rings
+            sage: out = singular(R)                                                     # needs sage.rings.finite_rings
+            sage: print(re.sub('ordering rp', 'ordering ip', out._repr_()))             # needs sage.rings.finite_rings
             polynomial ring, over a field, global ordering
             //   coefficients: ZZ/2[a]/(a^8+a^4+a^3+a^2+1)
             //   number of vars : 10
-            //        block   1 : ordering rp
+            //        block   1 : ordering ip
```
tornaria, line 1288:
```suggestion
            //        block   1 : ordering ...
```
etc.

src/sage/libs/singular/ring.pyx
```diff
@@ -71,6 +66,11 @@ order_dict = {
     "a":  ringorder_a,
 }
 
+from sage.interfaces.singular import singular_version_number
+if int(singular_version_number()[0:3]) < 432 or (int(singular_version_number()[0:3]) == 432 and int(singular_version_number()[3:]) < 15):
+    order_dict["rp"] = ringorder_rp
+else:
+    order_dict["ip"] = ringorder_ip
```
tornaria: seems unnecessary if we keep using `ringorder_rp` for the time being.

EDIT: wrong.. still need "rp" or "ip" :thinking: 

EDIT 2: Could just add both as in
```
"rp" : ringorder_rp,  # this is deprecated
"ip" : ringorder_ip,
```
As a matter of fact, `ringorder_rp` is an alias for `ringorder_ip` in old and in new singular. The only thing changed now is the name "rp" was changed to "ip".

src/sage/interfaces/singular.py
antonio-rojas: I considered this, but do we really want to skip testing that the correct ordering is used here?

tornaria: We really need a mechanism to support backward compatibility. Sprinkling examples with regular expressions to fix stuff like this is quite ugly in terms of documentation.

Similar problems we have right now with gap 4.13 (#37616, see also https://github.com/sagemath/sage/pull/37590#issuecomment-2013350669) and all over the place with results that have to be sorted to compare, etc.

For numerical noise there's the `# tol` tag which affects how the output of a doctest is compared. For order, I think a tag `# random_order` may be preferable to having to sort in the example. What could we do for this case?

An alternative is to add code in `SingularElement._repr_()` so the output of old singular is changed into the output of new singular (i.e. replace `ordering rp` with `ordering ip`). But this seems a bit hacky...

tornaria:
@antonio-rojas 

https://github.com/antonio-rojas/sage/compare/singular-4.4...tornaria:sage:singular

This is my attempt to support singular 4.3.2p10 and before. This is built on top of your first commit here, but I dropped the others.

1. add a doctest for `bigintvec` coercion (you implemented this but there's no test that will use this code)
2. fix the "arity" renumbering by printing the symbol instead of a number (seems nicer)
3. fix the compatibility for 4.3.2p10 by adding some `#define` and (only for "old" singular) patching the mapping tables so we use `rp` on singular side and patch output so it prints `ordering ip` even when using old singular (this solves your concern about testing the ordering).

I've tested with 4.3.2p10 (old) and 4.3.2p16 (new).

If you like the approach, feel free to take my commits into your PR, rebase at will. I don't make a PR to your PR since I am dropping some of your commits.

Note that I left out the stuff about `noexcept`; since #37667 will be merged in beta2, it seems easier to wait and fix it when rebasing on top of beta2.

mkoeppe:
I've tested this with:
- #37570

The tests include a mixture of platforms where Singular is built from the SPKG or taken from the system.
All works well. 
Thanks for preparing this!

mkoeppe (approved):

src/sage/interfaces/singular.py
```diff
@@ -1402,6 +1402,13 @@ def _repr_(self):
         if self._name in s:
             if self.get_custom_name() is None and self.type() == 'matrix':
                 s = self.parent().eval('pmat(%s,20)' % (self.name()))
+        # compatibility for singular 4.3.2p10 and before
+        if s.startswith("polynomial ring,"):
+            from sage.rings.polynomial.term_order import singular_name_mapping
+            # this is our cue that singular uses `rp` instead of `ip`
+            if singular_name_mapping['invlex'] == 'rp':
+                s = re.sub('^(// .*block.* : ordering )rp$', '\\1ip',
+                           s, 0, re.MULTILINE);
```
tornaria, line 1412:
```suggestion
                           s, 0, re.MULTILINE)
```

antonio-rojas:
I modified the ordering string replacement to happen only in doctests. Otherwis it can be confusing to get a different output for Singular code when run from inside Sage.

mkoeppe:
Tested again with #37570 (and some other upgrades) with the full CI Linux in https://github.com/mkoeppe/sage/actions/runs/8761823524, looking good.

mkoeppe:
Is there anything else to do here? From my side, this would be a positive review.

antonio-rojas:
> Is there anything else to do here? From my side, this would be a positive review.

Not that I'm aware of

vbraun:
merge conflict

antonio-rojas:
Rebased over beta 4

PR #37497, opened 2024-02-28
remove zombie code

mantepse:
We remove zombie code for solving linear systems over a finite field, which was (likely accidentally) created in #23214, but leads to errors when solving sparse linear equations over finite fields.

This does not quite fix the issue, because we now fall back to generic code.

Fixes #28586


videlec:
This code did test the `linbox` interface. I am not sure it is a good idea to remove it completely.

@ClementPernet what do you think?

mantepse:
> This code did test the `linbox` interface.

Could you expand on this?  If I understand the code correctly, whenever `_solve_vector_linbox` or `_solve_matrix_linbox` was called it would raise an exception.

videlec:
> > This code did test the `linbox` interface.
> 
> Could you expand on this? If I understand the code correctly, whenever `_solve_vector_linbox` or `_solve_matrix_linbox` was called it would raise an exception.

What about this doctest
```
            sage: algos = ["default", "dense_elimination", "sparse_elimination",
            ....:          "blackbox", "wiedemann"]
            sage: for i in range(20):
            ....:     dim = randint(1, 30)
            ....:     M = MatrixSpace(ZZ, dim, sparse=True)
            ....:     density = min(1, 4/dim)
            ....:     m = M.random_element(density=density)
            ....:     while m.rank() != dim:
            ....:         m = M.random_element(density=density)
            ....:     U = m.column_space().dense_module()
            ....:     for algo in algos:
            ....:         u, d = m._solve_vector_linbox(U.zero(), algorithm=algo)
            ....:         assert u.is_zero()
            ....:         b = U.random_element()
            ....:         x, d = m._solve_vector_linbox(b, algorithm=algo)
            ....:         assert m * x == d * b
```

mantepse:
This should be calling `Matrix_integer_sparse._solve_vector_linbox`, I hope.

videlec:
> This should be calling `Matrix_integer_sparse._solve_vector_linbox`, I hope.

Then at least, do not remove the doctest (unless it is duplicated somewhere else).

videlec:
I now understand my mistake. Sorry for being long. The diff
```
$ git diff c8fa7bb27f4724a7263d76bec437ebcf7f87dec3 matrix_modn_sparse.pyx
```
seems to explain how the call to linbox is supposed to be in the modn case. Instead of removing it as you propose would be better to try to have that code back again.

videlec:
The linbox interface for modn matrices seems to have been implemented by Martin Albrecht a while ago fdca585168aeb8b56afc5f794db9057440d7be5d

mantepse:
> The linbox interface for modn matrices seems to have been implemented by Martin Albrecht a while ago [fdca585](https://github.com/sagemath/sage/commit/fdca585168aeb8b56afc5f794db9057440d7be5d)

I don't know enough about linbox to be able to resurrect this code.  I had a brief look, it seems that a lot of things have changed since then (Sep 15, 2007).

mantepse:
ping?

mantepse:
Could we please have this?    Of course it would be nicer to have specialized code, but it seems that nobody has the skills to do this right now.

mkoeppe (approved):
LGTM.

If anyone wants to work on correct specialized versions, the removed code is always available in the commit history.

PR #37502, opened 2024-02-28
fixes incorrect large scale y-labels, cf #34233

dimpase:
This fixes #34233

Fix is provided in a comment by @niranjankm (Niranjana K.M.)

### :memo: Checklist


kcrisman:
> @niranjankm - hopefully your fix will make it into Sage

Yes, hopefully soon!  I would just want to test how this works with "unusual" formatters like the log and pi ones - I haven't looked at this code for some time, and just want to check there aren't funny interactions.

dimpase:
also fixes #7964 (I've made sure there is no `1e`, but $10^5$ etc there.

niranjankm:
Thanks Dima,
    I am so happy for my first fix getting into sagemath.
    Also in documentation,
https://deploy-preview-37502--sagemath.netlify.app/html/en/reference/plotting/sage/plot/plot

Please change in this example

plot(x^2, (x,300,500))  # scientific notation on y-axis
Graphics object consisting of 1 graphics primitive

Plot function into x^3 or x^4

Currenty x^2 is not producing numbers large enough to give scientific notation.


With regards

Niranjana K M

dimpase:
@niranjankm I'd also want to add Sphinx plots to manual examples showing `CustomScalarFormatting` in action, but the example
```python
from sage.plot.plot import CustomScalarFormatter
import matplotlib.pyplot as plt
import numpy as np
z = np.linspace(0, 5000, 100)
fig, ax = plt.subplots()
xmajorformatter = CustomScalarFormatter(replace_values=([2000,0],['$x_0$','']))
ymajorformatter = CustomScalarFormatter(replace_values=([1E7,0],['$y_0$','']))
ax.xaxis.set_major_formatter(xmajorformatter)
ax.yaxis.set_major_formatter(ymajorformatter)
ax.plot(z,z**2)
plt.show()
```
if I put it in `.. PLOT::` directive, messes up other plots (Sphinx is quite weird...).

So the question is how to explicitly reset `CustomScalarFormatting` etc back to the "normal" state. Do I also need to reset `matplotlib.pyplot` somehow?

niranjankm:
@dimpase
Thanks a lot Dima for making it happen.

src/sage/plot/plot.py
```diff
@@ -4138,3 +4138,94 @@ def generate_plot_points(f, xrange, plot_points=5, adaptive_tolerance=0.01,
...
+    Note that the first label entries are cumpulsorily int or float values. No strings.
+    No such restrictions for the second list. Because they will be anyway converted to strings in the end.
+
+    Authored by Niranjana K. M.
+
+    EXAMPLES:
+
+    check that :issue:`34233` is fixed::
+
+        sage: plot(90000*p/(100-p), (p,0,100), ymin=0, ymax=1000000)
```
kcrisman: Obviously we'll need to define p here, as below.

src/sage/plot/plot.py
```diff
@@ -4138,3 +4138,94 @@ def generate_plot_points(f, xrange, plot_points=5, adaptive_tolerance=0.01,
...
+        sage: from sage.plot.plot import CustomScalarFormatter
+        sage: import matplotlib.pyplot as plt
+        sage: import numpy as np
+        sage: z = np.linspace(0, 5000, 100)
+        sage: fig, ax = plt.subplots()
+        sage: xmajorformatter = CustomScalarFormatter(replace_values=([2000,0],['$x_0$','']))
+        sage: ymajorformatter = CustomScalarFormatter(replace_values=([1E7,0],['$y_0$','']))
+        sage: ax.xaxis.set_major_formatter(xmajorformatter)
+        sage: ax.yaxis.set_major_formatter(ymajorformatter)
+        sage: ax.plot(z,z**2)
```
kcrisman: Apparently this returns an object now.

kcrisman:
There's something funny about the new scientific notation.  Probably still better than the current situation, but ... it doesn't seem to put the $\times 10^{10}$ or whatever where it belongs.  For instance, try this:

    sage: plot(x^4, (x,-500,5))

tmp_bla_2067

kcrisman:
That would probably be acceptable, though inconsistent.  I'm not sure about this one:

    sage: plot_semilogy(x^4,(x,5,5e10))

tmp_hs0zx_l9

I'm uncertain about having different styles of labeling for both.  But maybe that's okay?  (And again, better than before.)  To be clear, the "better" log scale labeling has nothing to do with this PR.

kcrisman:
Finally, needs work, because it breaks custom formatting (which is a very nice selling point):

    sage: plot(x^2, (x,0,3), ticks=[[1,2.5],pi/2], tick_formatter=[["$x_1$","$x_2$"],pi])

tmp_2s5zc85q

I imagine one can get around this, but it would have to be only invoked as a custom formatter when there is no other custom formatter.  Studying why the log scales don't use this new one probably will be instructive.

kcrisman:
In addition, adding the 

    list_plot([2**n for n in range(100)])

example once this is fixed (from #30983 ) would be good once this is all sorted out.

dimpase:
> That would probably be acceptable, though inconsistent.  I'm not sure about this one:
> 
>     sage: plot_semilogy(x^4,(x,5,5e10))
> 
> tmp_hs0zx_l9
> 
> I'm uncertain about having different styles of labeling for both.  But maybe that's okay?  (And again, better than before.)  To be clear, the "better" log scale labeling has nothing to do with this PR.

are you saying that log/semilog plots got changed by this PR?

As far as I can see, the only affected case should be where both x and y-axes are linear.

kcrisman:
>
>
> I'm uncertain about having different styles of labeling for both. But
> maybe that's okay? (And again, better than before.) To be clear, the
> "better" log scale labeling has nothing to do with this PR.
>
> are you saying that log/semilog plots got changed by this PR?
>
I'm saying that before they were inconsistent in a different way than they
are now for semilog plots.  (loglog is fine.)


> As far as I can see, the only affected case should be where both x and
> y-axes are linear.Message ID: ***@***.***
> .com>
>
As far as my testing reveals, that is correct.  Unfortunately the other
problems are problems, particularly the custom tick formatting going
bye-bye.   But I'm hopeful that there will be an easy way to fix that - ?

niranjankm:
I am unable to get the both,
1. Scientific notation
2. Custom formatting
together at a time with derived formatters here viz `CustomScalarFormatter` and `SelectiveFormatter`.

Let me get back to the basics, the following lines in place of earlier formatter lines in the `if` statement under discussion in `graphics.py` gives success in resolving the issue.
```
            if (xmiddle and ymiddle and xscale == 'linear' == yscale):
                ## get the ticks, which are the numeric location of the ticks, and eliminate 0 from them
                if type(ticks[0]) is not list:
                    subplot.set_xticks([i for i in subplot.get_xticks() if i!=0][1:-1])
                if type(ticks[1]) is not list:
                    subplot.set_yticks([i for i in subplot.get_yticks() if i!=0][1:-1])
```

This attached zip file contains a notebook with examples used for testing. Ignore the printing of type of variable `ticks`. It was used for knowing the data types which make up the variable `ticks`.
[plot_scaling_bug-2.zip](https://github.com/sagemath/sage/files/14465198/plot_scaling_bug-2.zip)

niranjankm:
Except the ever taunting question that why the derived formatter, here `SelectiveFormatter`, after passing the remaining data to the default `ScalarFormatter` lost scientific notation, i hope the issue is resolved.

Also a warning note should be made in the `sage.plot.plot.SelectiveFormatter(formatter, skip_values)` section of the documentation that it misses the scientific notation. For example,
```
from sage.plot.plot import SelectiveFormatter
import matplotlib.pyplot as plt
import numpy
z = np.linspace(0, 500, 100)
fig, ax = plt.subplots()
ax.plot(z,z**4)
formatter = SelectiveFormatter(ax.yaxis.get_major_formatter(), skip_values=[0])
ax.yaxis.set_major_formatter(formatter)
plt.show()
```

kcrisman:
Thanks, @niranjankm Just to clarify, are you saying the code in [this comment](https://github.com/sagemath/sage/pull/37502#issuecomment-1973804436) plots all previously good plots correctly, *and* prints scientific notation?  Or that we don't have scientific notation at all?  Or some other combination?  My apologies for not quite understanding what you are claiming here.   If this works (which seems reasonable), we can probably just remove the other formatter completely.

I won't be able to actually test this immediately, but we definitely appreciate your engagement on this, as a fix would resolve quite a few annoying things about the plotting.  Thanks!

niranjankm:
Yes it will resolve the issues without using any customized matplotlib
formatter. The attached notebook illustrates it.

Don't know, SelectiveFormatter may be used elsewhere also. We can still
keep all these formatters as they serve some purpose atleast. They will
also serve as templates to further develop new tick formatters.

kcrisman:
> Yes it will resolve the issues without using any customized matplotlib formatter. The attached notebook illustrates it. Don't know, SelectiveFormatter may be used elsewhere also. We can still keep all these formatters as they serve some purpose atleast. They will also serve as templates to further develop new tick formatters.

Thanks, that what I thought you meant.  I'll check this out - we'll need an updated PR, I want to make sure you get credit though.

dimpase:
actually, to create or update  a PR, one does not need  a local install of sage. Everything could be done on github directly, in the browser.

I presume that we would like to do this new, and rather small, change, and add an example showing it works (the one we have in the PR already)

kcrisman:
> actually, to create or update a PR, one does not need a local install of sage. Everything could be done on github directly, in the browser.

For a small one, yes; I found it cumbersome to do a more sophisticated one (though I haven't used codespaces), and have gotten chided for doing so on another project.

> I presume that we would like to do this new, and rather small, change, and add an example showing it works (the one we have in the PR already)

Yeah, if this is the only change that would probably be best.  Let me test it a bit.

kcrisman:
> I presume that we would like to do this new, and rather small, change, and add an example showing it works (the one we have in the PR already)

This is great.  Thank you, @niranjankm !  What a much easier solution than the custom formatter, and easier to understand exactly why it works too.

For positive review of this particular ticket, I'd say we need:
* The minimal change in https://github.com/sagemath/sage/pull/37502#issuecomment-1973804436
* Modifying the examples that already exist to do `x^4`/`x**4`, as in the original PR
* The example in https://github.com/sagemath/sage/pull/37502#issuecomment-1973828320 if we don't want to immediately remove the formatter (I'm agnostic on that if it's the only place it shows up)

In addition, I think that this addition which @niranjankm added is more "math-y", and maybe we should use it too, see [matplotlib doc](https://matplotlib.org/stable/gallery/ticks/scalarformatter.html):
* `rcParams['axes.formatter.use_mathtext'] = True`

The documentation fixes for these should be easy, and maybe now that we have issues/PRs rather than Trac tickets, could be done here too:
* #7964 
* #30983 
* #31447 

My only remaining gripe/comment is where the labels show up in crossing the axis sometimes (see `plot(x,(x,-5000000,0))`) but given that we have ridiculous plots right now like `plot(x,(x,-5000000,5000000))` looking like it's from -5 to 5, as @williamstein pointed out, this is a very distinct improvement.

niranjankm:
Thanks @kcrisman , for appreciations and reviewing.
Thanks @dimpase , for getting the fix into Sagemath.

kcrisman:
I like this, so positive review modulo one issue which isn't really about this ticket, but needs an answer.

That is that apparently now test blocks are omitted from visible documentation (sad to me, though I vaguely recall something about this in a time I was not in a place to pay close attention to such things).  But the new plot for this example still shows up.  So ... I'm in favor of that, of course, but should a version of the plot (or just itself) be moved above the test block with some minimal verbiage pointing out that we have enabled `axes.formatter.use_mathtext` and expect scientific notation when the ranges are pretty big?

Otherwise thanks to both @dimpase  and @niranjankm this certainly needed fixing, and has done for some time.

dimpase:
how about moving it to Examples - then  both the code and the graph will be shown

kcrisman:
>
>
> how about moving it to Examples - then both the code and the graph will be
> shown
>
> I thought of that, but then we'd want to say what it's there for, maybe
instead of "check that :issue:`34233` is fixed::" it could be "Scientific
notation appears with large enough values" and then the code could be
repeated (but not the sphinx graph) in testing with the current wording
about 34233.

kcrisman:
looks good, assuming the auto-build has the plot in the right place.  Thank you!

dimpase:
OK,my last commit only removed extra whitespace in a doctest, so this does not need reviewing again.

dimpase:
I got confused by `..PLOT:` code being plain Python, and not a doctest, vs Sage code, doctested... Should be OK now

kcrisman:
> I got confused by `..PLOT:` code being plain Python, and not a doctest, vs Sage code, doctested... Should be OK now

I think the next line will also return a plot object.  I

dimpase:
that's why I put `...` there

kcrisman:
> that's why I put `...` there

My apologies, the code browser must not have shown that line for me, I only saw one changed line.  LGTM and presumably to the bot as well!

dimpase:
duh, I forgot that `var()` call is noisy, no matter what you do. Fixed now

kcrisman:
> duh, I forgot that `var()` call is noisy, no matter what you do. Fixed now

Yeah, the only thing you can do is `p = var('p')`, if I recall correctly.  

Anyway, this has positive review, for the record.

vbraun:
Please don't set to positive review when there are obvious CI failures

dimpase:
I see
```
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/plot/multigraphics.py
Error: Failed example:: Got: (0.025689729876243465,
 0.03415489151312756,
 0.44787900242824213,
 0.9345951084868724)
                                                                                                                                                                                                                                                                                                                                                                                              
    G.position(0)  # tol 5.0e-3
Expected:
    (0.025045451349937315,
     0.03415488992713045,
     0.4489880779745068,
     0.9345951100728696)
Got:
    (0.025689729876243465,
     0.03415489151312756,
     0.44787900242824213,
     0.9345951084868724)
Tolerance exceeded in 1 of 4:
    0.025045451349937315 vs 0.025689729876243465, tolerance 3e-2 > 5e-3
```

kcrisman:
>
>
> sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/plot/multigraphics.pyMessage ID: ***@***.***>
>
> But that's not this ticket, is it?

Maybe what happened is that something changed its location because of a
label slightly with the slight change in plotting from here.   If somebody
can compare https://sagecell.sagemath.org/?q=bjiapf with what happens in
this one (I don't have access to a build with this ticket this weekend) one
could see if it's actually relevant or looks different in any way.

dimpase:
@kcrisman - I looked at the graphs before and after the change, I don't see any difference (after all, it's a very small difference). So I just updated the output in the last commit

dimpase:
oops, last commit was wrong

kcrisman:
Wait, this was never merged?  Or was it?   I'm sorry I didn't put it back to positive review, I guess I thought that might implicitly happen, but of course that is wishful thinking.  Now it will have to be rebased yet again, I suppose; I apologize.

kcrisman:
Looks like it's failing some checks that Volker usually wants passing first.   At the least
```
sage -t --warn-long 5.0 --random-seed=10895229620020990782941135544346796037 src/sage/plot/multigraphics.py
File "src/sage/plot/multigraphics.py", line 1273, in sage.plot.multigraphics.GraphicsArray.position
Failed example:
    G.position(0)  # tol 5.0e-3
Expected:
    (0.025045451349937315,
     0.03415488992713045,
     0.4489880779745068,
     0.9345951100728696)
Got:
    (0.025689729876243465,
     0.034236451130373324,
     0.44787900242824213,
     0.9345135488696267)
Tolerance exceeded in 1 of 4:
    0.025045451349937315 vs 0.025689729876243465, tolerance 3e-2 > 5e-3
1 item had failures:
   1 of   8 in sage.plot.multigraphics.GraphicsArray.position
    [196 tests, 1 failure, 18.83s wall]
```
I suppose it's possible the new labeling slightly moves things.

dimpase:
ok, I'll fix these

dimpase:
as matplotlib is upgraded, this should be looked at again

PR #37319, opened 2024-02-13
`build/pkgs/meson`: Update to 1.3.1, require >= 1.2.3 for Python 3.12 support, reject meson installed in inaccessible user schemes

mkoeppe:
As reported in https://groups.google.com/g/sage-devel/c/9ETciH8efTE/m/UfUO7xmsAQAJ, our `configure` script currently accepts "system" meson that will not be functional in our venv.
```
$ pip install --user meson
Collecting meson
  Using cached meson-1.3.1-py3-none-any.whl.metadata (1.8 kB)
Using cached meson-1.3.1-py3-none-any.whl (976 kB)
Installing collected packages: meson
Successfully installed meson-1.3.1
$ python3 -c 'import sysconfig; print(sysconfig.get_path("scripts", "posix_user"))'
/Users/mkoeppe/Library/Python/3.11/bin
$ /Users/mkoeppe/Library/Python/3.11/bin/meson --version
1.3.1
$ ./sage -sh -c '/Users/mkoeppe/Library/Python/3.11/bin/meson --version'
Traceback (most recent call last):
  File "/Users/mkoeppe/Library/Python/3.11/bin/meson", line 5, in <module>
    from mesonbuild.mesonmain import main
ModuleNotFoundError: No module named 'mesonbuild'
```

Also tightening the version requirement. Only needed for Python 3.12 support (https://github.com/mesonbuild/meson-python/blob/main/pyproject.toml#L36), but we do this unconditionally for simplicity.

Also updating meson to latest and changing to a wheel package.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37504, opened 2024-02-28
`cmake`, `ninja_build`: Reject installations in inaccessible user schemes

mkoeppe:
These two are installable via binary wheels on PyPI.
When they are installed with `pip install --user`, as appears to be becoming more popular, then our `configure` script will see them, but they will not work within the Sage environment because we set PYTHONUSERBASE.

Fix as done previously for `meson` in #37319.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37319 (merged here)


mkoeppe:
@comboproblem Please try out this branch (in a new worktree). `./bootstrap && ./configure` should say that there is no cmake.

PR #37508, opened 2024-03-01
Developer guide: Recommend raw strings for all docstrings

### LINKED ISSUE #25998: Recommend r"""docstring""" over """docstring"""
<div id="comment:0"></div>

The general conventions webpage https://doc.sagemath.org/html/en/developer/coding_basics.html#latex-typesetting says that both docstring syntaxes (`r"""docstring"""` and `"""docstring"""`) are valid (which is true), but fails to recommend one over the other.

Since it is a *conventions* webpage, I think it is appropriate to recommend `r"""docstring"""` over the latter here.  The `r` convention is nicer because you don't need double backslashes in latex code, which is shorter and makes it easier to copy latex code from docstrings to a latex document and vice-versa.  Moreover, the webpage says that we follow the PEP257 conventions (https://www.python.org/dev/peps/pep-0257/) which themselves say 'Use r"""raw triple double quotes""" if you use any backslashes in your docstrings.'

Happy to make the changes myself once everyone is on board.

Component: **documentation**

Keywords: **convention, raw**

Author: **Matthew Lancellotti**

_Issue created by migration from https://trac.sagemath.org/ticket/25998_

mkoeppe:
- Resolves #25998

### :memo: Checklist

### DIFF
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -667,6 +667,9 @@ You are strongly encouraged to:
 
 - Use LaTeX typesetting (see :ref:`section-latex-typeset`).
 
+- Use raw strings (``r"""..."""``), regardless of whether the docstring
+  currently contains any backslashes or not.
+
 - Liberally describe what the examples do.
 
   .. NOTE::
@@ -778,9 +781,9 @@ In Sage's documentation LaTeX code is allowed and is marked with **backticks**:
 
     ```x^2 + y^2 = 1``` yields `x^2 + y^2 = 1`.
 
-**Backslashes:** For LaTeX commands containing backslashes, either use double
-backslashes or begin the docstring with a ``r"""`` instead of ``"""``. Both of
-the following are valid::
+**Backslashes:** For LaTeX commands containing backslashes, it is equivalent to
+use either double backslashes or begin the docstring with a ``r"""`` instead
+of ``"""``, but we strongly suggest to use the latter::
 
     def cos(x):
         """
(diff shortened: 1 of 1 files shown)


src/doc/en/developer/coding_basics.rst
```diff
@@ -778,9 +781,9 @@ In Sage's documentation LaTeX code is allowed and is marked with **backticks**:
 
     ```x^2 + y^2 = 1``` yields `x^2 + y^2 = 1`.
 
-**Backslashes:** For LaTeX commands containing backslashes, either use double
-backslashes or begin the docstring with a ``r"""`` instead of ``"""``. Both of
-the following are valid::
+**Backslashes:** For LaTeX commands containing backslashes, it is equivalent to
+use either double backslashes or begin the docstring with a ``r"""`` instead
+of ``"""``, but we strongly suggest to use the latter::
```
kwankyu, line 786: I agree. But I suggest to write
```diff
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -781,9 +781,8 @@ In Sage's documentation LaTeX code is allowed and is marked with **backticks**:
 
     ```x^2 + y^2 = 1``` yields `x^2 + y^2 = 1`.
 
-**Backslashes:** For LaTeX commands containing backslashes, it is equivalent to
-use either double backslashes or begin the docstring with a ``r"""`` instead
-of ``"""``, but we strongly suggest to use the latter::
+**Backslashes:** For LaTeX commands containing backslashes, either use double
+backslashes or begin the docstring with a ``r"""`` instead of ``"""``::
 
     def cos(x):
         """
@@ -795,6 +794,8 @@ of ``"""``, but we strongly suggest to use the latter::
         Return `\sin(x)`.
         """
 
+but we strongly suggest to use the latter.
+
 **MATH block:** This is similar to the LaTeX syntax ``\[<math expression>\]``
 (or ``$$<math expression>$$``). For instance:
 ```

kwankyu (approved):
Thanks. LGTM.

mkoeppe:
Thanks!

PR #37510, opened 2024-03-01
src/doc/*/tutorial/programming.rst: Update 'standalone script'

### LINKED ISSUE #8478: Programming tutorial page seems incorrect
<div id="comment:0"></div>

The section Standalone !Python/Sage Scripts at http://www.sagemath.org/doc/tutorial/programming.html#section-loadattach seems to have two problems (profuse apologies if these are actually user errors).

1. __`#!/usr/bin/env sage -python`__ doesn't work in linux since a shebang line can only handle a single argument ('sage' in this case). One workaround is to use __`#!/path/to/sage -python`__ -- I have since verified, however, that simply using __`#!/usr/bin/env sage`__ **does** work with sage 5.8.
2. The script itself seems broken.  When testing it using sage 4.3.3 (and 5.8), running __`./factor 2006`__ works as expected, but __`./factor "32*x^5-1"`__ gives:

```
  File "./factor", line 11, in <module>
    print factor(sage_eval(sys.argv[1]))
  File "/opt/sage-4.3.3-linux-32bit-ubuntu_9.10-i686-Linux/local/lib/python2.6/site-packages/sage/misc/sage_eval.py", line 199, in sage_eval
    return eval(source, sage.all.__dict__, locals)
  File "<string>", line 1, in <module>
NameError: name 'x' is not defined
```

To correct this, see the comment attached to this ticket which provides a new version of the 'factor' script.

Component: **documentation**

_Issue created by migration from https://trac.sagemath.org/ticket/8478_

mkoeppe:
This tutorial and its translations used to use `#!/usr/bin/env sage -python` in this example but that was changed to `#!/usr/bin/env sage`, perhaps to address #8478, except in the Japanese version.
Here we switch that version over and also remove an unnecessary import.

Fixes #8478.

### :memo: Checklist

### DIFF
--- a/src/doc/de/tutorial/programming.rst
+++ b/src/doc/de/tutorial/programming.rst
@@ -177,7 +177,6 @@ Polynome, usw.:
     #!/usr/bin/env sage
 
     import sys
-    from sage.all import *
 
     if len(sys.argv) != 2:
         print("Usage: %s <n>" % sys.argv[0])
--- a/src/doc/en/tutorial/programming.rst
+++ b/src/doc/en/tutorial/programming.rst
@@ -166,7 +166,6 @@ etc:
     #!/usr/bin/env sage
 
     import sys
-    from sage.all import *
 
     if len(sys.argv) != 2:
         print("Usage: %s <n>" % sys.argv[0])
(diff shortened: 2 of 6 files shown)


kwankyu (approved):
LGTM now. But we may have to put standalone Python script examples too later.

mkoeppe:
Thanks!

PR #37513, opened 2024-03-01
Define the zero polynomial to have degree `-Infinity` for `LaurentPolynomialRing`

### LINKED ISSUE #37491: `LaurentPolynomialRing` define the zero polynomial to have degree `-1` which is ambiguous
### Steps To Reproduce

```py
sage: R = LaurentPolynomialRing(QQ, 3, 'x')
sage: x0, x1, x2 = R.gens()
sage: R.zero().degree()
-1
sage: R.zero().degree(x0)
-1
sage: R.zero().degree(x1)
-1
sage: R.zero().degree(x2)
-1
sage: (1/x0).degree(x0)
-1
sage: (1/x0).degree()
-1
```

### Expected Behavior

The degree of `0` should be distinct from polynomials we can construct

### Actual Behavior

The degree of `0` is `-1`, the same as `1/x0` or other similar polynomials.

### Additional Information

For standard univariate and multivariate polynomial rings, the degree being `-1` is essentially a nice trick in place of $-\infty$, but for Laurent polynomial rings where negative degree makes total sense, this ambiguity with zero is confusing.

One idea would be to instead have it return `- infinity`.

### Environment

```markdown
- **Sage Version**: SageMath version 10.3.rc0, Release Date: 2024-02-25
```

### Checklist

GiacomoPope:
At the moment `LaurentPolynomialRing` computes the degree of the zero polynomial element to be `-1`, which is ambiguous as `R(1/x)` also has degree `-1`.

This PR changes the behaviour for both univariate and multivariate `LaurentPolynomialRings` so the degree of the zero polynomial is `-Infinity`. This is a common definition for the zero polynomial.

This PR has been made following the discussion on `sage-devel`: https://groups.google.com/g/sage-devel/c/XVz-4kqWuuo about exactly when the degree of the zero polynomial should be `-1`, `-Infinity` or undefined.

```py
sage: R.<x> = LaurentPolynomialRing(ZZ)
sage: R(1/x).degree()
-1
sage: R.zero().degree()
-Infinity
sage: 
sage: R.<x, y> = LaurentPolynomialRing(ZZ)
sage: R(1/x).degree()
-1
sage: R.zero().degree()
-Infinity
```

Fixes https://github.com/sagemath/sage/issues/37491


src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1168,7 +1169,17 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+
+            sage: R.<x> = LaurentPolynomialRing(ZZ)
+            sage: R.zero().degree()
+            -Infinity
         """
+        # The zero polynomial is defined to have degree -Infinity
+        if self.is_zero():
+            return minus_infinity
+
         if not x:
```
mantepse: This should actually be `x is None`, and documented.

GiacomoPope: I'm sorry I don't understand. `x` here is an optional argument which looks for the degree of the generator `x` in the polynomial in `f(x, y, z, ...)`.

Are you saying we should document that both `R.zero().degree() = R.zero().degree(x) = -Infinity`?

GiacomoPope: Oh! I see the docstring only says

> Return the degree of `x` in `self`.

So yes, this needs a more verbose docstring to handle when `x` is `None`.

**EDIT**:

I have tried to address this in the most recent commit.

mantepse: Thank you! Actually, I didn't realise that it would return the degree of the highest degree term, which seems extremely odd to me.  I would have thought that it would return total degree in this case.  Did you check whether this is used anywhere?

GiacomoPope: Maybe I am misunderstanding and it returns the total degree and my docstring example is confusing?

What is the total degree of a Laurent Polynomial Ring? Sorry for my naivety-- this isn't an area I work in much.

mantepse: Well, simply the sum of the degrees in each variable.  But I have no idea whether either is more useful than the other.  For LazyPowerSeries, we return the minimal total degree.  For example,
```
sage: L.<x,y> = LazyPowerSeriesRing(QQ)
sage: (x*y^4 + x^2*y).valuation()
3
```
(There is no such thing as a multivariate lazy Laurent series.)

GiacomoPope: I think my understanding of the current behaviour is correct, for both the univariate and multivariate cases the degree is the degree of the term with the highest degree.

```py
sage: R.<x> = LaurentPolynomialRing(ZZ)
sage: (x^3 + x).degree()
3
sage: (x^-3 + x).degree()
1
sage: 
sage: R.<x, y> = LaurentPolynomialRing(ZZ)
sage: (x + y^3).degree()
3
sage: (x*y^6 + y^3).degree()
7
```

Following this logic, in PR https://github.com/sagemath/sage/pull/37490 I added valuation for the multivariate case and this behaves similarly to degree:

```py
    def valuation(self, x=None):
         """
         Return the valuation of ``x`` in ``self``.
         If ``x`` is ``None``, return the minimal valuation of ``self``.
         EXAMPLES::
             sage: R.<x,y> = LaurentPolynomialRing(ZZ)
             sage: f = 2*x^2*y^-3 - 13*x^-1*y^-3 + 2*x^2*y^-5 - 2*x^-3*y^2
             sage: f.valuation()
             -4
             sage: f.valuation(x)
             -3
             sage: f.valuation(y)
             -5
             sage: R.zero().valuation()
             +Infinity
```

and matches the behaviour of the univariate case

```py
sage: R.<x> = LaurentPolynomialRing(ZZ)
sage: (x^-3 + x).valuation()
-3
```

GiacomoPope: Oh! but a comment from @nbruin in this PR says that valuation is a bad term for this, so maybe I will change this anyway!

I'm not sure I'm the one to be dealing with all these PR!

mantepse: I did not doubt that your interpretation of the code was correct, I was just surprised.  I must admit, surprised negatively, because this conflicts with the definition used in `LazyPowerSeriesRing`.

I think we first have to sort out the mathematics involved.

GiacomoPope: I actually wonder if there's a bug here as the underlying call is poly._total_degree() which is using singular. 

This area of mathematics is not my speciality so there's many things I may misunderstand but I'm happy to try and help. 

Maybe worth first digging into what .degree() is trying to compute (maybe the multivariate total_degree() is weirdly behaved?

mantepse:
I really think we should switch to total degree, but that needs discussion on sage-devel - on the same thread as the infinity change.

GiacomoPope:
I'm sorry it was always returning the total degree. What I meant before is that it returns `max(sum(e) for e in self.exponents()` -- which I saw as returning the degree of the term in the polynomial with the highest degree. From conversations elsewhere I now see this is exactly the total degree.

Sorry for my lack of language / understanding in this area. I have adjusted the docstring for degree.

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1155,21 +1156,48 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
         return [a.eadd(self._mon) for a in self._poly.exponents()]
 
     def degree(self, x=None):
-        """
-        Return the degree of ``x`` in ``self``.
+        r"""
+        If ``x`` is ``None``, return the total degreee of ``self``.
+        Otherwise, if ``x`` is generator of the ring, returns the
+        degree of ``x`` in ``self``.
```
mantepse: I think it is better to use a similiar docstring as in `multi_polynomial_libsingular.pyx`:
```
        Return the degree of this polynomial.

        INPUT:

        - ``x`` -- (default: ``None``) a generator of the parent ring

        OUTPUT:

        If ``x`` is not given, return the maximum degree of the monomials of
        the polynomial.  If ``x`` is given, it is (or coercible to) a generator of the
        parent ring and the output is the maximum degree in ``x``.
```

GiacomoPope: I have implemented this change, but removed the `(or coercible to)` as there is no coercion in this function.

GiacomoPope:
@mantepse do you think this needs further review? I just want to organise the labels

src/sage/rings/polynomial/laurent_polynomial_mpair.pyx
```diff
@@ -1155,21 +1156,56 @@ cdef class LaurentPolynomial_mpair(LaurentPolynomial):
...
+
+        INPUT:
+
+        - ``x`` -- (default: ``None``) a generator of the parent ring
+
+        OUTPUT:
+
+        If ``x`` is ``None``, return the maximum degree of the monomials of
+        ``self``.  If ``x`` is given and it is a generator of the parent
+        ring, the output is the maximum degree in ``x`` of ``self``.
```
tscrim:
```suggestion
        If ``x`` is ``None``, return the total degree of ``self``.
        If ``x`` is a given generator of the parent ring,
        the output is the maximum degree of ``x`` in ``self``.
```
I think this is a better way to write this. Total degree is a very commonly defined concept IMO, so there will be no ambiguity. Maximum degree of the monomials is vague to the wrong definition (depending on the interpretation).

tscrim (changes requested):
Edit: Looking at @mantepse's comment, I actually disagree with it. I would say the doc for `multi_polynomial_libsingular.pyx` should be changed (although this doesn't have to be done here, but I agree with his general principal that they should match).

tscrim:
@mantepse I don't think we need to have a sage-devel discussion to change the doc to total degree. I am going to cc @nbruin to see if he has any thoughts on the doc change to using total degree in both places.

GiacomoPope:
So we change the `OUTPUT` of the `degree()` function in `multi_polynomial_libsingular.pyx` in the following way?

```py
If ``x`` is not given, return the maximum degree of the monomials of        
the polynomial. Note that the degree of a monomial is affected by the
gradings given to the generators of the parent ring. If ``x`` is given,
it is (or coercible to) a generator of the parent ring and the output
is the maximum degree in ``x``. This is not affected by the gradings of
ithe generators.
```

```py
If ``x`` is ``None``, return the total degree of ``self``. Note that this
result is affected by the gradings given to the generators of the parent
ring. Otherwise, if ``x`` is (or is coercible to) a generator of the parent
ring, the output is the maximum degree of ``x`` in ``self``. This is not
affected by the gradings of the generators.
```

mantepse:
> @mantepse I don't think we need to have a sage-devel discussion to change the doc to total degree.

No, we certainly don't - I thought, by mistake, that it wouldn't compute the total degree but rather the maximum of the degrees of the generators.

nbruin:
I find the description on multi_polynomial_singular confusing. If the degree is dependent on the term ordering, does that mean it returns some kind of degree for the leading coefficient? I don't think it does:
```
sage: R.<x,y>=QQ[] # I think this has lex order
sage: (x^2*y+x*y^3).degree()
4
sage: (x^3*y+x*y^2).degree()
4
sage: (x^3*y).degree()
4
sage: (x*y^2).degree()
3
```
Or does it mean that variables may have a weight and that the weighted total degree (shortened to weighted degree if the weighting is not the standard all 1's) ? Degrees generally arise from putting some grading on your ring (not necessarily into finite-dimensional pieces!) and then the degree of a polynomial is the highest graded piece your polynomial has a component in. From what I see above, I don't think `degree` without a parameter uses a degree that is fully derived from the term ordering.

So I think it's worthwhile looking at what degree actually does there; use the term "total degree" for what it actually means and explain "weighted degree" when necessary. "maximum degree" derived from the term ordering seems to me "(weighted) degree of leading monomial", which probably isn't what we use in the case of lex ordering.

GiacomoPope:
I don't really understand how grading is defined here, but for all standard orderings the degree is unchanged (as one would expect from the total degree) but for weighting this indeed has a different result:

````py
sage: R.<x, y> = PolynomialRing(QQ)
sage: (x^2*y + x*y^3).degree()
4
sage: (x^3*y + x*y^2).degree()
4
sage: (x^3*y).degree()
4
sage: (x*y^2).degree()
3
sage: R.<x, y> = PolynomialRing(QQ, order=TermOrder("wdegrevlex", [1, 3]))
sage: (x^2*y + x*y^3).degree()
10
sage: (x^3*y + x*y^2).degree()
7
sage: (x^3*y).degree()
6
sage: (x*y^2).degree()
7
```

nbruin:
@GiacomoPope +1. It's cleaner if this documentation changes to multi_polynomials goes on a separate ticket, though.

GiacomoPope:
OK I have made a new PR for this change. Does anything else need work / review on this PR?

tscrim:
I think I am okay. @mantepse any other comments?

mantepse:
fine with me!

tscrim (approved):
Then let's get this in. We will need to have the `degree()` function for Laurent polynomials work with the grading first before changing the doc here to match the regular polynomials. That should be on its own PR as well.

GiacomoPope:
Yeah I think there needs to be a PR for a check of weighted polynomials and this class. Maybe loads is broken.

tscrim:
Laurent polynomials have not received as much care as regular polynomials (this is unsurprising to me though). There are likely other similar inconsistencies. I probably would be more reserved with suggesting there might be large amounts of broken code, but that's me.

tscrim:
To require less reading between the lines: I very much appreciate the work you're doing in finding and fixing these issues with Laurent polynomials.

GiacomoPope:
Oh! I didn't mean so much that the Laurent code is broken but rather the use of weighted polynomials seems to be permitted in many places but not expected often. 

I wouldn't be surprised if many functions simply didn't anticipate weighted polynomials and so will just behave weirdly. 

As far as I can tell the weighted polynomials and their uses in sage aren't fully mature. An example I came across recently is really weighted polynomials are natural for hyperelliptic curves but currently aren't used and moving to them from normal ones reveals some bugs.

tscrim:
It's no problem. I know what you meant and it wasn't meant as an admonishment in any way. You had good qualifiers in your statement too.

I just try to be careful with these types of statements with potential misinterpretation. I spend some time going around trying to convince people to use and develop for Sage (via [SageDays](https://wiki.sagemath.org/Workshops) (Please consider coming to one or more! Typically, they are only announced on that page.), research visits, math talks), and having some precise statements about where there are problems helps. There have been a few cases recently where people go "everything in area X is horribly broken" (paraphrased here) because of some isolated bugs, which might have made me too sensitive on this too...

GiacomoPope:
Yeah my recent sage burst of sage energy came off of sage days 123. In fact aside from a few rabbit holes most of these PR I've been looking at have been to try and support (hyper)elliptic and isogeny code. 

Nice to do clean up PR like these ones too. I owe a lot of my maths understanding to sage and I really want it to be more robust

tscrim:
> Yeah my recent sage burst of sage energy came off of sage days 123. In fact aside from a few rabbit holes most of these PR I've been looking at have been to try and support (hyper)elliptic and isogeny code.

That's great. Sorry I couldn't be there for that SageDays (I was still teaching and am in Europe for all of this month of March).

> Nice to do clean up PR like these ones too. I owe a lot of my maths understanding to sage and I really want it to be more robust

That is exactly how I learn a number of (or verify my understanding of) math things. Programming doesn't allow for shortcuts. Although I also take this as part of my (broader community) service work too...

nbruin:
> Then let's get this in. We will need to have the `degree()` function for Laurent polynomials work with the grading first before changing the doc here to match the regular polynomials. That should be on its own PR as well.

If we have different weights on the generators, I think this might affect what one would consider the "valuation" as well. I suspect there is some toric geometry that can make sense of such a weighting. The valuation would not correspond to a straight blow-up of the origin, but I suspect there is some interpretation that makes sense.

tscrim:
> If we have different weights on the generators, I think this might affect what one would consider the "valuation" as well. I suspect there is some toric geometry that can make sense of such a weighting. The valuation would not correspond to a straight blow-up of the origin, but I suspect there is some interpretation that makes sense.

I don't know enough (algebraic) geometry to try and construct this, or even to understand the details your previous description. I imagine someone has already thought of this before since it is such a natural (or at least naive) construction considering it algebraically.

vbraun:
merge conflict

GiacomoPope:
@vbraun i merged develop into this branch. There were no merge conflicts but I suppose after the CI runs we can make sure everything is ok

GiacomoPope:
This should now be all good @vbraun -- thank you for your help with this.

GiacomoPope:
@vbraun I think I have handled all the failures and conflicts. There is a package failure in the CI but I think this is unrelated? Pinging because of the needs work tab

GiacomoPope:
Sorry for the ping, @vbraun but wanted to follow up on the needs work tag.

tscrim:
@GiacomoPope Same comment as the other ticket (@dimpase has already removed the tag).

PR #35801, opened 2023-06-21
Add `cmr` (Combinatorial Matrix Recognition library) and Cython interface

mkoeppe:
### :books: Description

- [x] Adding an SPKG for the Combinatorial Matrix Recognition Library (successor of TUtest) by Matthias Walter
https://discopt.github.io/cmr/, and dependencies
- [x] Cython interface in `sage.libs.cmr`
- [x] New matrix backend class `Matrix_cmr_chr_sparse`
- [ ] Classes for decomposition nodes (`CMR_DEC_...`)
- [x] New methods for [`Matrix_integer_dense`](https://doc.sagemath.org/html/en/reference/matrices/sage/matrix/matrix_integer_dense.html#module-sage.matrix.matrix_integer_dense) along the lines of the existing methods `smith_form` etc.; delegate to `Matrix_cmr_chr_sparse` methods
- [ ] Definitions of all terms
- [ ] Cite literature
- [ ] Examples and tests for all methods/classes
- [ ] Link to CMR documentation - https://discopt.github.io/cmr/ (maybe use #37577)
- [ ] Methods for matroids
- [ ] Subclasses of `Matroid` corresponding node types of the Seymour decomposition
- [ ] Connect to [modular decomposition of graphs](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph_decompositions/modular_decomposition.html)
- [ ] Connect to `sage.graphs.connectivity.spqr_tree`?
- [ ] Matrix class that is backed by a `DecompositionNode`
- [ ] Graph backend class
- [ ] Wrap remaining functions of CMR
- [ ] pip-installable package **sagemath-cmr** (depending on sagemath-modules, #35095)

Follow-ups:
- #38139 

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Upstream: https://github.com/discopt/cmr/issues/32
- Depends on #37512
- Depends on #37514 
- Depends on #37516 
- Depends on #37606 
- Depends on #37692
- Depends on #37998 (merged here)

@xuluze @jsantillan3 @discopt

PR #37514, opened 2024-03-01
`MatrixSpace`: Support constructing `Hom(CombinatorialFreeModule)`

mkoeppe:
The `FreeModule` constructor was recently generalized so that it can create `CombinatorialFreeModule` objects. In particular,
```
sage: ZZ^3
Ambient free module of rank 3 over the principal ideal domain Integer Ring
sage: ZZ^['a','b','c']
Free module generated by {'a', 'b', 'c'} over Integer Ring
```

Here as a follow-up, we extend `MatrixSpace` in a similar way: When lists or enumerated sets of row indices and column indices are given, it constructs `CombinatorialFreeModule`s whose bases are indexed by these index sets, and then returns the homset. In particular,
```
sage: ZZ^(2, 3)
Full MatrixSpace of 2 by 3 dense matrices over Integer Ring
sage: ZZ^(['x','y'], ['a', 'b', 'c'])
Set of Morphisms from Free module generated by {'a', 'b', 'c'} over Integer Ring to Free module generated by {'x', 'y'} over Integer Ring in Category of finite dimensional modules with basis over Integer Ring
```

Follow-up PRs:
- extend `matrix` likewise; guiding application: adjacency / incidence "matrices" of graphs, where rows/columns are indexed by nodes and edges 
- see also #35801

### :memo: Checklist


src/sage/matrix/matrix_space.py
```diff
@@ -434,9 +434,75 @@ def get_matrix_class(R, nrows, ncols, sparse, implementation):
...
+    INPUT:
+
+    - ``base_ring`` -- a ring
+
+    - ``nrows`` or ``row_keys`` -- (nonnegative integer) the number of rows, or
+      a finite or enumerated family of arbitrary objects that index the rows
+      of the matrix
+
+    - ``ncols`` or ``column_keys`` -- (nonnegative integer, default ``nrows``)
+      the number of columns, or a finite or enumerated family of arbitrary objects
```
fchapoton: I would prefer to remove "or enumerated" for safety.

mkoeppe: OK, that's finite with me, done in 6885c2d

mkoeppe:
Ready to go?

gmou3 (approved):

tscrim (commented):
What happens if you mix the dimension and keys? For example, row keys with number of columns. (Of course, there is more evil input of row keys and row dimension, which might be good to check against with doctests.)

mkoeppe:
It's checked for consistency, and errors are raised if there's a mismatch

mkoeppe:
Yes, I can add more doctests for that. If there are specific ones that you'd like to see, let me know

tscrim:
Just doctests testing the compatibility, such as
```
MatrixSpace(QQ, ['a','b'], 4)
MatrixSpace(QQ, 4, ['a','b'])
MatrixSpace(QQ, ['a','b'], ['x','y'], nrows=2)  # I think should pass
MatrixSpace(QQ, ['a','b'], ['x','y'], nrows=4)
MatrixSpace(QQ, ['a','b'], ['x','y'], ncols=2)  # I think should pass
MatrixSpace(QQ, ['a','b'], ['x','y'], ncols=4)
MatrixSpace(QQ, ['a','b'], ['x','y'], nrows=2, ncols=2)  # I think should pass
MatrixSpace(QQ, ['a','b'], ['x','y'], nrows=2, ncols=4)
MatrixSpace(QQ, ['a','b'], ['x','y'], nrows=4, ncols=4)
MatrixSpace(QQ, 4, ['a','b'], nrows=4, ncols=2)  # I think should pass if the second above passes
```

Also, your `INPUT:` block and docstring is not correct. There is `nrows_or_row_keys` (which I think should be changed to `row_keys`) and similarly for columns.

mkoeppe:
> There is `nrows_or_row_keys` (which I think should be changed to `row_keys`)

It's intended as the name of a positional-only argument. 

> MatrixSpace(QQ, 4, ['a','b'], nrows=4, ncols=2)  # I think should pass if the second above passes

No, I disagree, it should be rejected as a case of "duplicate argument" nrows.

tscrim:
> > There is `nrows_or_row_keys` (which I think should be changed to `row_keys`)
> 
> It's intended as the name of a positional-only argument.

Ah, sorry I missed it in the long list of arguments. Now there are more compatibility tests to be done as you have to mix with the `row_keys` and `nrows`.

> > MatrixSpace(QQ, 4, ['a','b'], nrows=4, ncols=2)  # I think should pass if the second above passes
> 
> No, I disagree, it should be rejected as a case of "duplicate argument" nrows.

It isn't a duplication per se. I could see it being used as a sanity check by a user. I would be a bit confused by this resulting in an error because of course it has 4 rows and 2 columns and the 4 is a shorthand for the index set [0, 1, 2, 3] in this setting.

Also, I believe this is true but I would like confirmation, the values that can be converted to `int` will always be less than `sys.maxsize`, correct?

mkoeppe:
> the 4 is a shorthand for the index set [0, 1, 2, 3] in this setting.

Well, there's a clear distinction between passing 4 and passing range(4). 
One creates a FreeModule and the other creates a CombinatorialFreeModule.

mkoeppe:
I have added the doctests (and input checking code), please take a look

mkoeppe:
> the values that can be converted to `int` will always be less than `sys.maxsize`, correct?

No, `int(x)` converts any integer to an `int`. So we get, as expected, 
```
sage: MatrixSpace(ZZ, sys.maxsize+1)
OverflowError: number of rows and columns may be at most 9223372036854775807
```

tscrim:
> > the values that can be converted to `int` will always be less than `sys.maxsize`, correct?
> 
> No, `int(x)` converts any integer to an `int`. So we get, as expected,
> 
> ```
> sage: MatrixSpace(ZZ, sys.maxsize+1)
> OverflowError: number of rows and columns may be at most 9223372036854775807
> ```

That didn't exactly answer my question. I don't know what limits Python has imposed on the `int` right now, and while it might work for our systems, it should be universal.

mkoeppe:
> > > the 4 is a shorthand for the index set [0, 1, 2, 3] in this setting.
> > 
> > 
> > Well, there's a clear distinction between passing 4 and passing range(4). One creates a FreeModule and the other creates a CombinatorialFreeModule.
> 
> If _both_ of the arguments were integers, then I would agree. However, because one of the inputs is a list, passing `4` gets implicitly converted into a list before being passed along. 

No, there's no such mechanism. It constructs a morphism between a FreeModule and a CombinatorialFreeModule in this case.
```
      sage: MatrixSpace(QQ, ['a','b'], 4)
        Set of Morphisms (Linear Transformations)
         from Vector space of dimension 4 over Rational Field
           to Free module generated by {'a', 'b'} over Rational Field
```

mkoeppe:
> > > the values that can be converted to `int` will always be less than `sys.maxsize`, correct?
> > 
> > 
> > No, `int(x)` converts any integer to an `int`. So we get, as expected,
> > ```
> > sage: MatrixSpace(ZZ, sys.maxsize+1)
> > OverflowError: number of rows and columns may be at most 9223372036854775807
> > ```
> 
> That didn't exactly answer my question. I don't know what limits Python has imposed on the `int` right now, and while it might work for our systems, it should be universal.

I don't understand what you mean. What limits are you talking about?

tscrim:
Okay, I've checked, in Python3, `int` is bounded by the size of the memory, which should always be larger than `sys.maxsize` (unless your machine is completely crazy).

tscrim:
> > > > the 4 is a shorthand for the index set [0, 1, 2, 3] in this setting.
> > > 
> > > 
> > > Well, there's a clear distinction between passing 4 and passing range(4). One creates a FreeModule and the other creates a CombinatorialFreeModule.
> > 
> > 
> > If _both_ of the arguments were integers, then I would agree. However, because one of the inputs is a list, passing `4` gets implicitly converted into a list before being passed along.
> 
> No, there's no such mechanism. It constructs a morphism between a `FreeModule` and a `CombinatorialFreeModule` in this case.
> 
> ```
>       sage: MatrixSpace(QQ, ['a','b'], 4)
>         Set of Morphisms (Linear Transformations)
>          from Vector space of dimension 4 over Rational Field
>            to Free module generated by {'a', 'b'} over Rational Field
> ```
> 
> Edit: I'll add that many of these homsets are not fully implemented, but that's outside of the scope of this PR. I have added some `# known bug` now to document this. For completing the implementation, see for example:
> 
>     * [Fix scaling of vector space morphisms on the left #37709](https://github.com/sagemath/sage/pull/37709)
> 
>     * [Make homsets of FiniteDimensionalModulesWithBasis finite-dimensional #37516](https://github.com/sagemath/sage/pull/37516)

Ah I see. This is acceptable then.

Although now I notice another issue. I don't quite like how, e.g., setting `ncols` overrides the CFM to a standard free module:
````
sage: MatrixSpace(QQ, ['a','b'], ['x','y'], ncols=2)
Set of Morphisms (Linear Transformations)
 from Vector space of dimension 2 over Rational Field
  to Free module generated by {'a', 'b'} over Rational Field
```
I am treating the `ncols` as verification here of the dimension.

mkoeppe:
> Although now I notice another issue. I don't quite like how, e.g., setting `ncols` overrides the CFM to a standard free module:

That's a bug.

mkoeppe:
Fixed now

mkoeppe:
Any more comments or can we merge it?

tscrim (approved):
Yes, this is acceptable to me.

mkoeppe:
Thanks!

PR #37521, opened 2024-03-02
fix debian.txt for qhull

dimpase:
added missing Debian package for qhull

### DIFF
--- a/build/pkgs/qhull/distros/debian.txt
+++ b/build/pkgs/qhull/distros/debian.txt
@@ -1,1 +1,2 @@
+qhull-bin
 libqhull-dev


kcrisman:
Did you add the wrong reviewer by mistake?  I know almost nothing about Debian and how it works, nor the syntax for this.  I didn't see that the Debian files for other packages were this different than (say) Arch or Homebrew, but maybe qhull is unusual that way:
* https://packages.debian.org/buster/qhull-bin
* https://packages.debian.org/buster/libqhull-dev

So does this just say which packages to pull in from Debian?  I notice their last time stamp is 2015, hope that's okay.

dimpase:
buster is an old Debian. current stable is bookworm

it's really is a trivial fix. qhull is used both as a library and an an executable, and in bookworm they are in separate packages

dimpase:
yes, these records are just listing OS packages to be installed

kcrisman:
> buster is an old Debian. current stable is bookworm

Haha, I didn't even look at that piece of the URL.

Okay, LGTM.

PR #37535, opened 2024-03-03
Upgrade to jupyterlab 4.1.3 and notebook 7.1.1

### LINKED ISSUE #37533: Sage should upgrade to jupyterlab 4.1.3 and notebook 7.1.1 for 10.3
### Problem Description

The versions of jupyterlab and notebook used by the 10.3rc1 release (4.0.6 and 7.0.6) have a problem which is specifically relevant to Sage and which is fixed in the slightly newer versions suggested here.  The problem is that the Sage documentation links do not appear in the Help menu in the notebook.  I opened a ticket for this with the notebook project at the time of the 10.2 release.  They fixed it, but the the fix did not take effect immediately because it depended on a change to Jupyterlab which had not been released at that time.  The fix is now in place with jupyterlab 4.1.2 and notebook 7.1.1.  I have tested these (they are included in the Sage_macOS 10.3 prerelease) and the links do appear correctly in the Help menu, and they work.

### Proposed Solution

Make these small bumps (4.0.6 -> 4.1.2 and 7.0.6 -> 7.1.1) in the package versions used in these two wheel spkgs.

### Alternatives Considered

Continuing to use a redundant second venv in the macOS binary.

### Additional Information

Currently the Sage_macOS binary includes a second, mostly redundant, venv in which these packages and their dependencies are installed.  This ends up being fairly large, because there are lots of dependencies.  I think the redundant venv could be removed if Sage made this minor upgrade.  This would be a significant reduction (176 MB) in the size of the binary package.

mkoeppe:
- https://github.com/jupyter/notebook/releases/tag/v7.1.0
- https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#v4-1

Fixes #37533

### :memo: Checklist


mkoeppe:
```
[jupyterlab-4.1.2] ERROR: Could not find a version that satisfies the requirement httpx>=0.25.0 (from jupyterlab) (from versions: none)
[jupyterlab-4.1.2] ERROR: No matching distribution found for httpx>=0.25.0
```

culler (commented):
I think this is fine, as far as it goes.  But I think that Sage should really update as many of the dependencies of the  jupyterlab and notebook packages as possible.  Please see PR #37581, where I attempted to do this.

mkoeppe:
Unlikely that we can get the bigger upgrade into 10.3.
So let's set this one here to "positive review" and "blocker" (so it will be considered for merging)?

culler:
OK. Fine with me.

mkoeppe:
Thanks!

PR #37538, opened 2024-03-04
Improve genus method of function fields and curves

kwankyu:
- make genus method of function fields over number fields work. 
```sage
sage: R.<T> = QQ[]
sage: N.<a> = NumberField(T^2 + 1)
sage: K.<x> = FunctionField(N)
sage: K.genus()
0
sage: S.<t> = PolynomialRing(K)
sage: L.<y> = K.extension(t^2 - x^3 + x)
sage: L.genus()  # not working without this PR
1
```

- allow integral curves to be defined over number fields
- make geometric genus method of integral curves defined over number fields work
```sage
sage: R.<x> = QQ[]
sage: N.<a> = NumberField(x^2 + 1)
sage: P2.<x,y,z> = ProjectiveSpace(N, 2)
sage: C = Curve(y^2*z - x^3 + x*z^2, P2)
sage: C.geometric_genus() # not working without this PR
1
```

- allow integral curves to be defined from one-dimensional ambient space
```sage
sage: R.<x> = QQ[]
sage: N.<a> = NumberField(x^2 + 1)
sage: P1.<x,y> = ProjectiveSpace(N, 1)
sage: C = Curve(P1)
sage: C.geometric_genus()  # not working without this PR
0
```

- add comments warning for nonsensical result of geometric genus method applied to geometrically-reducible curves
```sage
sage: P2.<x,y,z> = ProjectiveSpace(QQ, 2)
sage: C = Curve(x^2 + y^2, P2)
sage: C.geometric_genus()  # negative geometric genus!
-1
```

- remove unnecessary condition (irreducibility) on arithmetic genus of curves
```sage
sage: P.<x,y,z> = ProjectiveSpace(QQ, 2)
sage: C = Curve(x^2 - y^2, P)
sage: C.is_irreducible()
False
sage: C.arithmetic_genus()  # not working without this PR
0
```

### :memo: Checklist

### DIFF
--- a/src/sage/rings/function_field/function_field_polymod.py
+++ b/src/sage/rings/function_field/function_field_polymod.py
@@ -34,11 +34,15 @@ from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
 from sage.rings.integer import Integer
 from sage.categories.homset import Hom
 from sage.categories.function_fields import FunctionFields
+from sage.categories.number_fields import NumberFields
 
 from .element import FunctionFieldElement
 from .element_polymod import FunctionFieldElement_polymod
 from .function_field import FunctionField
 from .function_field_rational import RationalFunctionField
+
+_FunctionFields = FunctionFields()
+_NumberFields = NumberFields()
 
 
 class FunctionField_polymod(FunctionField):
@@ -160,7 +164,7 @@ class FunctionField_polymod(FunctionField):
         self._polynomial = polynomial
 
         FunctionField.__init__(self, base_field, names=names,
-                               category=FunctionFields().or_subcategory(category))
+                               category=_FunctionFields.or_subcategory(category))
 
         from .place_polymod import FunctionFieldPlace_polymod
         self._place_class = FunctionFieldPlace_polymod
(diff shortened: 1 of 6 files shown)


src/sage/rings/function_field/function_field_polymod.py
```diff
@@ -34,12 +34,16 @@
...
 from sage.categories.function_fields import FunctionFields
+from sage.categories.number_fields import NumberFields
 
 from .element import FunctionFieldElement
 from .element_polymod import FunctionFieldElement_polymod
 from .function_field import FunctionField
 from .function_field_rational import RationalFunctionField
 
+_FunctionFields = FunctionFields()
+_NumberFields = NumberFields()
```
tscrim: Why do we pin these in memory here? Is the `FunctionField_polymod` called so often in tight loops that the extra construction of the category has a noticeable impact?

`NumberFields` should not be because the genus shouldn't be called that often (we might even be okay just doing the import there as well, but that isn't so important).

src/sage/schemes/curves/constructor.py
```diff
@@ -76,8 +72,8 @@
                            IntegralAffinePlaneCurve,
                            IntegralAffinePlaneCurve_finite_field)
 
-
-from sage.schemes.plane_conics.constructor import Conic
+_Fields = Fields()
+_NumberFields = NumberFields()
```
tscrim: Same comment. I don't see the advantages of having yet another place where these are nailed in memory.

src/sage/schemes/curves/curve.py
```diff
@@ -246,12 +238,22 @@ def geometric_genus(self):
             sage: C.geometric_genus()
             3
 
+        Note that the geometric genus is only defined for `geometrically
+        irreducible curve <https://stacks.math.columbia.edu/tag/0BYE>`_. This
+        method does not check the condition. Be warned that you may get a
+        nonsensical result if the curve is not geometrically irreducible.
```
tscrim: Perhaps better as a `.. NOTE::` or `.. WARNING::`??

tscrim (changes requested):

src/sage/rings/function_field/function_field_polymod.py
kwankyu: I wanted to save some micro seconds. But you are right. It is not likely to be called in tight loops. 

OK. Done.

src/sage/schemes/curves/constructor.py
kwankyu: Done.

src/sage/schemes/curves/curve.py
kwankyu: OK. I put it to `.. WARNING::`.

tscrim (approved):
Thank you. LGTM.

I think these are already nailed in memory in a few places (in one of the categories; if you want to import one from there, please go ahead). I think it would be better to have just one place where such things are constructed if we decide to change this in some fashion. It also can save loading time by a few milliseconds.

kwankyu:
Thanks!

PR #37545, opened 2024-03-05
various list-comprehension in combinat (ruff PERF)

fchapoton:
following some `ruff --select=PERF src/sage/combinat` suggestions

i.e. using more often list comprehension where possible

(also turned one method to an iterator)

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/cluster_algebra_quiver/mutation_type.py
+++ b/src/sage/combinat/cluster_algebra_quiver/mutation_type.py
@@ -149,7 +149,7 @@ def _triangles(dg):
     return trians
 
 
-def _all_induced_cycles_iter( dg ):
+def _all_induced_cycles_iter(dg):
     """
     Return an iterator for all induced oriented cycles of length
     greater than or equal to 4 in the digraph ``dg``.
@@ -356,9 +356,9 @@ def _connected_mutation_type(dg):
     for edge in edges:
         label = edge[2]
         if label not in [(1,-1),(2,-2),(1,-2),(2,-1),(4,-1),(1,-4)]:
-    # _false_return(i) is a simple function that simply returns 'unknown'.  For debugging purposes, it
-    # can also output 'DEBUG: error i' if desired.
-    # this command is used many times in this code, something times without the argument i.
+            # _false_return(i) is a simple function that simply returns 'unknown'.  For debugging purposes, it
+            # can also output 'DEBUG: error i' if desired.
+            # this command is used many times in this code, something times without the argument i.
             return _false_return(2)
         elif label == (2,-2):
             dg.set_edge_label( edge[0], edge[1], 1 )
@@ -1032,7 +1032,7 @@ def _connected_mutation_type_AAtildeD(dg, ret_conn_vert=False):
             test_triangles = [[tuple(trian) for trian in oriented_trians
                                if edge in trian]
                               for edge in multiple_trian_edges]
-            unique_triangle = set(test_triangles[0]).intersection( *test_triangles[1:] )
+            unique_triangle = set.intersection(*map(set, test_triangles))
             if len(unique_triangle) != 1:
                 return _false_return(19)
             else:
@@ -1299,7 +1299,7 @@ def load_data(n, user=True):
     return data
 
 
-def _mutation_type_from_data( n, dig6, compute_if_necessary=True ):
+def _mutation_type_from_data(n, dig6, compute_if_necessary=True):
     r"""
     Return the mutation type from the given dig6 data by looking into
     the precomputed mutation types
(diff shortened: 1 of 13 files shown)


src/sage/combinat/set_partition.py
```diff
@@ -1873,8 +1873,7 @@ def arcs(self):
         arcs = []
         for p in self:
             p = sorted(p)
-            for i in range(len(p) - 1):
-                arcs.append((p[i], p[i + 1]))
+            arcs.extend((p[i], p[i + 1]) for i in range(len(p) - 1))
         return arcs
```
mantepse, line 1877: Alternatively:
```
return [(p[i], p[i + 1]) for b in self if (p := sorted(b)) for i in range(len(p) - 1)]
```
but I'm not sure whether I like it better.

src/sage/combinat/tableau.py
```diff
@@ -2613,16 +2613,12 @@ def slide_multiply(self, other):
...
-        else:
-            l = len(self[0])
 
-        for row in other:
-            st.append((None,)*l + row)
-        for row in self:
-            st.append(row)
+        l = len(self[0])
+        st = [(None,) * l + row for row in other]
+        st.extend(row for row in self)
```
mantepse: That should be the same as `st.extend(self)`.

src/sage/combinat/plane_partition.py
```diff
@@ -324,12 +324,10 @@ def cells(self) -> list[list[int]]:
             sage: PP.cells()
             [[0, 0, 0], [0, 0, 1], [0, 0, 2], [0, 1, 0], [1, 0, 0], [1, 0, 1]]
         """
-        L = []
-        for r in range(len(self)):
-            for c in range(len(self[r])):
-                for h in range(self[r][c]):
-                    L.append([r, c, h])
-        return L
+        return [[r, c, h]
```
mantepse: I think we should be returning tuples here, really.  We do so for partitions and tableaux, too.

src/sage/combinat/cluster_algebra_quiver/mutation_type.py
```diff
@@ -1023,17 +1023,17 @@ def _connected_mutation_type_AAtildeD(dg, ret_conn_vert=False):
...
-            for edge in multiple_trian_edges:
-                test_triangles.append([ tuple(trian) for trian in oriented_trians if edge in trian ])
+            test_triangles = [[tuple(trian) for trian in oriented_trians
+                               if edge in trian]
+                              for edge in multiple_trian_edges]
             unique_triangle = set(test_triangles[0]).intersection( *test_triangles[1:] )
-            if len( unique_triangle ) != 1:
+            if len(unique_triangle) != 1:
                 return _false_return(19)
             else:
```
mantepse, line 1038: `else:` can be removed.

src/sage/combinat/cluster_algebra_quiver/mutation_type.py
```diff
@@ -1023,17 +1023,17 @@ def _connected_mutation_type_AAtildeD(dg, ret_conn_vert=False):
         multiple_trian_edges = list(set(multiple_trian_edges))
 
         # test that there at most three edges appearing in exactly two oriented triangles
-        count = len( multiple_trian_edges )
+        count = len(multiple_trian_edges)
         if count >= 4:
             return _false_return(321)
         # if two edges appearing in exactly two oriented triangles, test that the two edges together
         # determine a unique triangle
         elif count > 1:
```
mantepse, line 1031: `elif` can be shortened to `if`.

src/sage/combinat/cluster_algebra_quiver/mutation_type.py
```diff
@@ -1023,17 +1023,17 @@ def _connected_mutation_type_AAtildeD(dg, ret_conn_vert=False):
...
-            test_triangles = []
-            for edge in multiple_trian_edges:
-                test_triangles.append([ tuple(trian) for trian in oriented_trians if edge in trian ])
+            test_triangles = [[tuple(trian) for trian in oriented_trians
+                               if edge in trian]
+                              for edge in multiple_trian_edges]
             unique_triangle = set(test_triangles[0]).intersection( *test_triangles[1:] )
```
mantepse: this can be simplified to
```
unique_triangle = set.intersection(*map(set, test_triangles))
```

src/sage/combinat/set_partition.py
fchapoton: I would rather keep it like I made it

mantepse:
see #35715 for the doctest error in rings, which is therefore unrelated.

mantepse (approved):
Thank you!

PR #37563, opened 2024-03-07
trying a change in unique_representation

fchapoton:
namely break the loop over `mro` once found what we want

I am not sure at all that this keeps the expected properties, but this is more efficient..

### :memo: Checklist

### DIFF
--- a/src/sage/structure/unique_representation.py
+++ b/src/sage/structure/unique_representation.py
@@ -1008,7 +1008,7 @@ class CachedRepresentation(metaclass=ClasscallMetaclass):
             True
         """
         instance = typecall(cls, *args, **options)
-        assert isinstance( instance, cls )
+        assert isinstance(instance, cls)
         if instance.__class__.__reduce__ == CachedRepresentation.__reduce__:
             instance._reduction = (cls, args, options)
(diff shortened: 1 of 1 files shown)


tscrim:
There is a subtle but important change that is made here. In the previous version, it found the last thing in the MRO that has a `__classcall__.cache`, whereas yours finds the first. I don't know if there is supposed to be a unique case for this or what impacts this might have, but I am worried about it because of that.

However, using the list comprehension is definitely faster and better.

fchapoton:
Yes, I am also worried. One can see that this breaks no doctests. May this would be ok by adding a detailed comment in the code around the break staement ?

fchapoton:
adding a check triggers only a few failures, notably in `sage.tensor.modules.finite_rank_free_module.FiniteRankFreeModule`

fchapoton:
Ok, I have kept only the non-risky changes. so it should be good to go now.

tscrim (approved):
Okay, thanks. Let it be so.

@nthiery cc-ing you in case you have any insights about the `_clear_cache_` getting the last class in the MRO instead of the first.

PR #37564, opened 2024-03-07
use Parent in infinity

fchapoton:
Let us use `Parent` for the ugly Unsigned Infinity Ring.

The case of the Infinity Ring seems to be even more problematic, and is not touched upon here.

### :memo: Checklist


tscrim (approved):
LGTM.

PR #37401, opened 2024-02-19
Developer Guide: Improvements to explanation of version files

mkoeppe:
### :memo: Checklist

Improvements prompted by the discussions in
- https://groups.google.com/g/sage-devel/c/MIU-xo9b7pc
- https://groups.google.com/g/sage-devel/c/5kmxaw105lg

All about `install-requires.txt`, `package-version.txt`, `requirements.txt`, ...

- [x] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37581, opened 2024-03-10
Update Sage spkgs to match, as closely as possible, the dependencies of current versions of notebook and jupyterlab.

culler:
Sage's notebook and jupyterlab spkgs have many dependencies, but are essentially independent from all other parts of Sage.  The versions of the dependencies in Sage are mostly far behind those of current versions of those two packages as available on PyPI.

The starting point for this PR was to create a clean venv and install jupyter_server 2.12.5 (since 2.13.0 broke the menu item for shutting down the server) as well as current versions of notebook and jupyterlab.  I tested the jupyter-notebook and jupyterlab executables in the venv and found them satisfactory.  I then generated a requirements.txt file from that venv.

My intent was to update Sage's spkgs to the versions in that requirements.txt file.  That proved to be impossible within the confines of Sage's policies (mainly not allowing platform specific binary wheels) together with my understanding of the current consensus of the Sage developers (mainly, that they do not want to add a Rust toolchain to Sage).  So I adjusted the versions in order to find the balance point where the packages are as new as possible, subject to those constraints, and not too old to meet the requirements of the current notebook and jupyterlab packages. 

This package brings Sage's notebooks and their dependencies up to date as much as possible.

This PR extends and would supersede the PR [#37535](https://github.com/sagemath/sage/pull/37535) .

### :memo: Checklist


build/pkgs/anyio/SPKG.rst
```diff
@@ -4,7 +4,11 @@ anyio: High level compatibility layer for multiple asynchronous event loop imple
 Description
 -----------
 
-High level compatibility layer for multiple asynchronous event loop implementations
+This spkg is installed from a wheel downloaded from
+PyPI.
+
+For a detailed description of the package, please visit its
+`PyPI page <https://pypi.org/project/anyio/>`_.
```
mkoeppe: These changes to SPKG.rst are not good. We build our manual from these! e.g. https://doc.sagemath.org/html/en/reference/spkg/appnope.html#spkg-appnope

build/pkgs/anyio/install-requires.txt
```diff
@@ -1 +1 @@
-anyio
+anyio ==4.3.0
```
mkoeppe: Our `install-requires.txt` files should not pin to specific versions. We do the pinning using the package-version.txt files.
See https://github.com/sagemath/sage/pull/37401

build/pkgs/anyio/SPKG.rst
culler: Of course the SPKG.rst files could be reverted.  I almost did that.  But, after looking at many of them,  I decided that a link to the PyPI README file would be just as useful and in many cases more useful.  Most of those descriptions are written by copy-pasting from the PyPI README anyway.  These auto-generated SPKG.rst files provide the link.  A user reading the documentation who wanted to know the details about the package could simply click it.

So ultimately I decided that using the link was an improvement, and also that it clearly reduces the demands on the time of Sage developers.

build/pkgs/anyio/install-requires.txt
culler: I see.  I will update the template in my pypi_spkg package.

Do you have any objections to me posting that package on pypi?

build/pkgs/anyio/SPKG.rst
mkoeppe: No, I disagree. It's intended that these pages in the documentation are readable rather than just being a collection of links. 
Actually I'd like to expand some of these descriptions with a description of *how* the package is being used in the Sage project. (That's part of a larger effort to improve the relation of Sage with upstream projects.)

Our `sage -package create` script does take the description from the PyPI README.
But some packages don't have a good description in their README, and we have some improvements. (Some of those should probably be upstreamed.)

By the way, we already have the PyPI link in the "Upstream Contact" section (if we don't, we should add it).

build/pkgs/anyio/install-requires.txt
mkoeppe: Of course, no objections here!

But, I hope it's needless to say, you are of course also very welcome to participate in directly contributing to the `sage_bootstrap` package, where our `sage --package create` script lives.

build/pkgs/anyio/SPKG.rst
culler: Certainly a well-written description that explains how Sage uses the package would be much better than a link to PyPI.  It looks like a lot of work for one person to write those, but with a lot of people pitching in it might be feasible.  The link was the best I could do with the time I had available.  Feel free to revert or rewrite any of the descriptions!

PR #37591, opened 2024-03-11
ruff PERF fixes in algebras/

fchapoton:
a few changes about using more list comprehensions in `algebras`

plus some pep8, and code details

Note: the `classical_lie_algebras` file has very slow doctest for E7.

### :memo: Checklist


src/sage/algebras/lie_algebras/free_lie_algebra.py
```diff
@@ -387,7 +389,7 @@ def _repr_(self):
             sage: LieAlgebra(QQ, 3, 'x')
             Free Lie algebra generated by (x0, x1, x2) over Rational Field
         """
-        n = tuple(map(LieGenerator, self._names, range(len(self._names)))) # To remove those stupid quote marks
+        n = tuple(map(LieGenerator, self._names, range(len(self._names))))  # To remove those stupid quote marks
```
GiacomoPope, line 392: looks like this TODO could be removed?

src/sage/algebras/lie_algebras/free_lie_algebra.py
```diff
@@ -266,7 +267,7 @@ def graded_dimension(self, k):
             return 0
         from sage.arith.misc import moebius
         s = len(self.lie_algebra_generators())
-        k = ZZ(k) # Make sure we have something that is in ZZ
+        k = ZZ(k)  # Make sure we have something that is in ZZ
```
GiacomoPope, line 270: Can this fail? The comment sounds like there's coercion here -- do we need error handling?

fchapoton: Well, somebody could input a python integer for `k`. That `k` is a sage integer is only required to have the method `divisors`. So, everything is in order here, I think.

src/sage/algebras/lie_algebras/free_lie_algebra.py
fchapoton: There is no TODO here, just a remark that this is a complicated but necessary way to get the correct string.

GiacomoPope: Sorry, was reading too fast. Sorry for wasting your time.

src/sage/algebras/lie_algebras/free_lie_algebra.py
tscrim: It also can frequently happen that Python ints find their way as inputs through Sage library code too.

src/sage/algebras/schur_algebra.py
```diff
@@ -675,13 +666,10 @@ def GL_irreducible_character(n, mu, KK):
...
     phi = mbasis.zero()
-    for aa in range(len(contents)):
-        mat = []
-        for kk in range(len(JJ[aa])):
-            temp = []
-            for j in range(length):
-                temp.append(graded_basis[aa][kk].inner_product(carter_lusztig[j]))
-            mat.append(temp)
+    for aa, c_aa in enumerate(contents):
+        mat = [[graded_basis[aa][kk].inner_product(carter_lusztig[j])
```
tscrim: This would probably benefit from defining len(`JJ[aa])` and `graded_basis[aa]` as variables.

src/sage/algebras/steenrod/steenrod_algebra_bases.py
```diff
@@ -348,36 +350,37 @@ def steenrod_algebra_basis(n, basis='milnor', p=2, **kwds):
...
+    # Arnon 'C' basis
     elif not generic and basis == 'arnonc':
         return arnonC_basis(n)
     else:
         raise ValueError("Unknown basis: %s at the prime %s" % (basis, p))
 
+
 # helper functions for producing bases
 
 def restricted_partitions(n, l, no_repeats=False):
```
tscrim, line 381: Could we add a `.. TODO::` here to implement this non-recursively. It is easy enough to do, but that would really expand the scope of the PR here...

tscrim (commented):
Thanks for cleaning up some of my old code.

Unfortunately that $E_7$ test is the shortest test for that type of implementation, so I feel it should be run as part of the "fast" tests. I plan to make that faster, but I haven't found the time to do so yet...

tscrim (approved):
Thanks. LGTM.

Issue #37593, opened 2024-03-11
Elliptic curve addition performance has regressed


GiacomoPope:
### Steps To Reproduce

#### Prime fields

Small characteristic 
```py
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 10.3.rc3, Release Date: 2024-03-10                │
│ Using Python 3.11.4. Type "help()" for help.                       │
└────────────────────────────────────────────────────────────────────┘
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Warning: this is a prerelease version, and it may be unstable.     ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
sage: p = 11
....: E = EllipticCurve(GF(p), [1,1])
....: set_random_seed(0)
....: P = E.random_point()
....: Q = E.random_point()
....: %timeit P + P
....: %timeit P + Q
18.8 µs ± 1.14 µs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
16.7 µs ± 365 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```py
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 10.2, Release Date: 2023-12-03                    │
│ Using Python 3.11.4. Type "help()" for help.                       │
└────────────────────────────────────────────────────────────────────┘
sage: p = 11
....: E = EllipticCurve(GF(p), [1,1])
....: set_random_seed(0)
....: P = E.random_point()
....: Q = E.random_point()
....: %timeit P + P
....: %timeit P + Q
11.1 µs ± 41.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
9.75 µs ± 99.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Large characteristic

```py
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 10.3.rc3, Release Date: 2024-03-10                │
│ Using Python 3.11.4. Type "help()" for help.                       │
└────────────────────────────────────────────────────────────────────┘
[...]

PR #37596, opened 2024-03-11
fix performance regression in elliptic-curve point constructor

yyyyx4:
See [here](https://github.com/sagemath/sage/issues/37593#issuecomment-1989251306).

### DIFF
--- a/src/sage/dynamics/arithmetic_dynamics/wehlerK3.py
+++ b/src/sage/dynamics/arithmetic_dynamics/wehlerK3.py
@@ -1668,7 +1668,7 @@ class WehlerK3Surface_ring(AlgebraicScheme_subscheme_product_projective):
             l = L.index(max(L))
             newP = copy(PK)
             #normalize PK
-            newP.scale_by([1/PK[0][i],1])
+            newP.scale_by([~PK[0][i], ZZ.one()])
 
             #Find B and A, helper values for the local height
             if PK[1][j].abs() <= PK[1][l].abs():
@@ -1679,7 +1679,7 @@ class WehlerK3Surface_ring(AlgebraicScheme_subscheme_product_projective):
 
             #Normalize Q
             newQ = copy(Q)
-            newQ.scale_by([1,1/Q[1][l]])
+            newQ.scale_by([ZZ.one(), ~Q[1][l]])
 
             if PK[0][i].abs() <= PK[0][k].abs():
                 A = Ry(W.Gpoly(0, k))(tuple(newQ[1]))*PK[0][i]/PK[0][k]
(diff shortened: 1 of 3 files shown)


GiacomoPope:
If you have this built locally, can you run a test or two from the examples in https://github.com/sagemath/sage/issues/37593 to show the new relative performance?

Nice work also cleaning up all the R(1) in the code.

yyyyx4:
$q=11$, Sage 10.2:
```
6.05 µs ± 11.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
5.22 µs ± 41.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
$q=11$, this branch:
```
6.61 µs ± 21.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
5.74 µs ± 27.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

$q=(2^{127}-1)^3$, Sage 10.2:
```
28.3 µs ± 299 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
19.3 µs ± 105 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
$q=(2^{127}-1)^3$, this branch:
```
28.3 µs ± 61.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
19.2 µs ± 66.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

GiacomoPope (approved):
LGTM

Thanks for the quick fix!

Could you add a "Fixes https://github.com/sagemath/sage/issues/37593" to close this issue on merge?

yyyyx4:
Thanks for noticing the problem and for the review, and sorry for causing the slowdown in the first place!

> Could you add a "Fixes #37593" to close this issue on merge?

This is in one of the commit messages, which should work as well, according to [the GitHub documentation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

GiacomoPope:
> which should work as well

I agree but I didn't see the linked issue in the GitHub UI on the PR tracker... Maybe this is a silly UI issue though and what you've done is fine.

As an aside, have you ever looked into `P._add_` and adding 2 methods which deal separately for short and long weierstrass curves? This would save a handful of additions and multiplications by zero for short weierstrass curves.

GiacomoPope:
Doctest failure due to `if t.is_zero():  #what if R(t) == 0 ?` where `t` is of type `int`

GiacomoPope (approved):
LGTM -- thank you!

GiacomoPope:
@vbraun although this is a small fix, it would be great to get this into 10.3 so that we don't ship degraded performance on something so widely used.

PR #37575, opened 2024-03-09
Intersphinx for cvxopt, cvxpy, cypari2, cysignals, flint, fpylll, gmpy2, ipywidgets, matplotlib, mpmath, networkx, numpy, rpy2, scipy, sympy

mkoeppe:
[Intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) allows us to refer to functions and classes in other projects using the standard Sphinx roles `:func:`,  `:class:`, ...

Examples in the documentation preview:
- [References to FLINT functions in sage.libs.flint.arith_sage](https://deploy-preview-37575--sagemath.netlify.app/html/en/reference/libs/sage/libs/flint/arith_sage#sage.libs.flint.arith_sage.bell_number)
- [References to SciPy functions in sage.manifolds.differentiable.integrated_curve](https://deploy-preview-37575--sagemath.netlify.app/html/en/reference/manifolds/sage/manifolds/differentiable/integrated_curve#sage.manifolds.differentiable.integrated_curve.IntegratedCurve.solve)
- [References to NetworkX functions in sage.graphs.generic_graph](https://deploy-preview-37575--sagemath.netlify.app/html/en/reference/graphs/sage/graphs/generic_graph#sage.graphs.generic_graph.GenericGraph.export_to_file)
- [Examples in the Developer Guide](https://deploy-preview-37575--sagemath.netlify.app/html/en/developer/sage_manuals#hyperlinks)

Based on a rebased version of #29231 by @mwageringel. In addition to the `scipy` intersphinx done in #29231, here we add a number of relevant Python libraries, as well as `flint`, whose docs are built with Sphinx as well.

To address concerns in the discussion there about the docbuild contacting remote servers for the inventory files, here we instead vendor the inventory files:
```
$ ls -l src/doc/common/_vendor
-rw-r--r--  1 mkoeppe  staff    1942 Mar  9 21:20 cvxopt.inv
-rw-r--r--  1 mkoeppe  staff   13251 Mar  9 21:20 cvxpy.inv
-rw-r--r--  1 mkoeppe  staff   10728 Mar  9 21:20 cypari2.inv
-rw-r--r--  1 mkoeppe  staff     775 Mar  9 21:20 cysignals.inv
-rw-r--r--  1 mkoeppe  staff  246266 Mar  9 21:20 flint.inv
-rw-r--r--  1 mkoeppe  staff    1603 Mar  9 21:20 fpylll.inv
[...]

PR #37598, opened 2024-03-12
Sphinx extlinks for ecl, gap, giac, maxima, meson, ppl, qepcad, scip, singular, soplex

mkoeppe:
We create a standard way to refer to the online manuals of these packages, which are generated by tools other than Sphinx and therefore cannot be linked to using Intersphinx.

This extends what has been already done with Pari.

Examples:
- [Examples in the Developer Guide](https://deploy-preview-37598--sagemath.netlify.app/html/en/developer/sage_manuals#hyperlinks)

Fixes #27495.

Outside of the scope of this PR:
- adding such extlinks everywhere
- working with upstream projects to upgrade to better documentation tools
- projects that can be linked to using Intersphinx; that's #37575
- any consideration how to link instead to local copies of those manuals, whether installed from SPKG or system packages.

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/designs/incidence_structures.py
+++ b/src/sage/combinat/designs/incidence_structures.py
@@ -1810,8 +1810,7 @@ class IncidenceStructure:
 
         REFERENCE:
 
-        - Soicher, Leonard, Design package manual, available at
-          https://www.gap-system.org/Manuals/pkg/design/htm/CHAP003.htm
+        - Soicher, Leonard, :gap_package:`Design package manual <design/htm/CHAP003.htm>`
         """
         if algorithm == "gap":
             libgap.load_package("design")
--- a/src/sage/combinat/designs/incidence_structures.py
+++ b/src/sage/combinat/designs/incidence_structures.py
@@ -1810,7 +1810,7 @@ class IncidenceStructure:
 
         REFERENCE:
 
-        - Soicher, Leonard, :gap_package:`Design package manual <design/htm/CHAP003.htm>`
+        - Leonard Soicher, :gap_package:`Design package manual <design/htm/CHAP003.htm>`
         """
         if algorithm == "gap":
             libgap.load_package("design")
--- a/src/sage/misc/sagedoc.py
+++ b/src/sage/misc/sagedoc.py
@@ -517,7 +517,7 @@ extlinks = {
     'common_lisp': ('https://www.lispworks.com/documentation/lw50/CLHS/Body/%s.htm', 'Common Lisp: %s'),
     'ecl': ('https://ecl.common-lisp.dev/static/manual/%s.html', 'ECL: %s'),
     'gap': ('https://docs.gap-system.org/doc/ref/%s_mj.html', 'GAP: %s'),
-    'gap_package': ('https://www.gap-system.org/Packages/%s.html', 'GAP package %s'),
+    'gap_package': ('https://docs.gap-system.org/pkg/%s', 'GAP package %s'),
     'giac_cascmd': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac/doc/en/cascmd_en/%s.html', 'Giac: %s'),
     'giac_us': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac_us.html#%s', 'Giac API: %s'),
     'maxima': ('https://maxima.sourceforge.io/docs/manual/maxima_singlepage.html#%s', 'Maxima: %s'),
--- a/src/sage/misc/sagedoc.py
+++ b/src/sage/misc/sagedoc.py
@@ -516,7 +516,7 @@ extlinks = {
     'mathscinet': ('https://www.ams.org/mathscinet-getitem?mr=%s', 'MathSciNet %s'),
     'common_lisp': ('https://www.lispworks.com/documentation/lw50/CLHS/Body/%s.htm', 'Common Lisp: %s'),
     'ecl': ('https://ecl.common-lisp.dev/static/manual/%s.html', 'ECL: %s'),
-    'gap': ('https://gap-system.org/Manuals/doc/ref/%s_mj.html', 'GAP: %s'),
+    'gap': ('https://docs.gap-system.org/doc/ref/%s_mj.html', 'GAP: %s'),
     'gap_package': ('https://www.gap-system.org/Packages/%s.html', 'GAP package %s'),
     'giac_cascmd': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac/doc/en/cascmd_en/%s.html', 'Giac: %s'),
     'giac_us': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac_us.html#%s', 'Giac API: %s'),
--- a/src/sage/misc/sagedoc.py
+++ b/src/sage/misc/sagedoc.py
@@ -514,12 +514,15 @@ extlinks = {
     'doi': ('https://doi.org/%s', 'doi:%s'),
     'pari': ('https://pari.math.u-bordeaux.fr/dochtml/help/%s', 'pari:%s'),
     'mathscinet': ('https://www.ams.org/mathscinet-getitem?mr=%s', 'MathSciNet %s'),
+    'common_lisp': ('https://www.lispworks.com/documentation/lw50/CLHS/Body/%s.htm', 'Common Lisp: %s'),
+    'ecl': ('https://ecl.common-lisp.dev/static/manual/%s.html', 'ECL: %s'),
     'gap': ('https://gap-system.org/Manuals/doc/ref/%s_mj.html', 'GAP: %s'),
     'gap_package': ('https://www.gap-system.org/Packages/%s.html', 'GAP package %s'),
     'giac_cascmd': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac/doc/en/cascmd_en/%s.html', 'Giac: %s'),
     'giac_us': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac_us.html#%s', 'Giac API: %s'),
     'maxima': ('https://maxima.sourceforge.io/docs/manual/maxima_singlepage.html#%s', 'Maxima: '),
     'ppl': ('https://www.bugseng.com/products/ppl/documentation/user/ppl-user-1.2-html/%s.html', 'PPL: %s'),
+    'qepcad': ('https://www.usna.edu/CS/qepcadweb/B/%s.html', 'QEPCAD: %s'),
     'singular': ('https://www.singular.uni-kl.de/Manual/4-3-2/%s.htm', 'Singular: %s'),
 
 }
--- a/src/sage/misc/sagedoc.py
+++ b/src/sage/misc/sagedoc.py
@@ -520,11 +520,14 @@ extlinks = {
     'gap_package': ('https://www.gap-system.org/Packages/%s.html', 'GAP package %s'),
     'giac_cascmd': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac/doc/en/cascmd_en/%s.html', 'Giac: %s'),
     'giac_us': ('https://www-fourier.ujf-grenoble.fr/~parisse/giac_us.html#%s', 'Giac API: %s'),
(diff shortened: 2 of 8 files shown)


src/sage/misc/sagedoc.py
```diff
@@ -513,7 +513,21 @@ def process_dollars(s):
     'oeis': ('https://oeis.org/%s', 'OEIS sequence %s'),
     'doi': ('https://doi.org/%s', 'doi:%s'),
     'pari': ('https://pari.math.u-bordeaux.fr/dochtml/help/%s', 'pari:%s'),
-    'mathscinet': ('https://www.ams.org/mathscinet-getitem?mr=%s', 'MathSciNet %s')
+    'mathscinet': ('https://www.ams.org/mathscinet-getitem?mr=%s', 'MathSciNet %s'),
+    'common_lisp': ('https://www.lispworks.com/documentation/lw50/CLHS/Body/%s.htm', 'Common Lisp: %s'),
+    'ecl': ('https://ecl.common-lisp.dev/static/manual/%s.html', 'ECL: %s'),
+    'gap': ('https://gap-system.org/Manuals/doc/ref/%s_mj.html', 'GAP: %s'),
```
fingolfin: That's the old address, it will redirect, but perhaps just use the new address directly:
```suggestion
    'gap': ('https://docs.gap-system.org/doc/ref/%s_mj.html', 'GAP: %s'),
```

src/sage/misc/sagedoc.py
```diff
@@ -513,7 +513,21 @@ def process_dollars(s):
...
+    'gap_package': ('https://www.gap-system.org/Packages/%s.html', 'GAP package %s'),
```
fingolfin: Just as heads-up, those URLs are probably not going to exist for much longer. Not sure what the goal here is, is it to link to the homepages of these packages? Or their documentation? Or what?

mkoeppe: Anything really, I don't need anything very specific here for now. Package documentation would be fine.

src/sage/misc/sagedoc.py
mkoeppe: Thanks @fingolfin!

src/sage/misc/sagedoc.py
mkoeppe: @fingolfin What URL would you recommend to use for the package documentation?

mkoeppe: @fingolfin And more generally (not for this PR) -- what would be the best way to link to specific functions or other objects in the GAP documentation in a stable way? 
(For packages that use Sphinx to build their documentation, we use the cross-referencing tool Intersphinx for this purpose.)

fingolfin: While we still mirror the docs of distributed GAP packages on www.gap-system.org, the URLs there are not quite uniform, as the entry pages have different names / paths for each package.

This information is contained in the metadata of each package. That metadata is available in a JSON file (e.g. <https://github.com/gap-system/gap/releases/download/v4.13.0/package-infos.json.gz> for the packages distributed with GAP 4.13.0). I can provide further details if desired.

On the website we only link to the manual for the latest GAP release, and the manuals of the packages within that. So these are not really "stable".

I hope that we can eventually enhance this by adding version pickers. In preparation for this future (which may never come *sigh* but I am hoping), I made sure that from a certain point on, we had separate URLs for each GAP version -- these are not currently used anywhere! But as I wrote, I hope this will change.

The "regular" URLs for our mirrors of package manuals look like these (note the different paths after the package name)
- <https://docs.gap-system.org/pkg/smallsemi/doc/chap0.html>
- <https://docs.gap-system.org/pkg/transgrp/htm/chapters.htm>

The "versioned" ones look like this:
- <https://docs.gap-system.org/v4.13.0/pkg/smallsemi/doc/chap0.html>
- <https://docs.gap-system.org/v4.13.0/pkg/transgrp/htm/chapters.htm>

(So they are versioned by GAP version, not by package version. We could debate this at length, but let's not).

Anyway, so my suggestion would be to use these URLs.

But the annoying bit remains that you can't know the full URLs (`.../doc/chap0.html` or `.../doc/chap0_mj.html` or `..../htm/chapters.htm` or maybe something else) without consulting that JSON file.

But perhaps we could just add a step to [the script which updates the manuals on the GAP website for a new release](https://github.com/gap-system/GapWWW/blob/master/etc/extract_manuals.py), and add redirects ...
- from <https://docs.gap-system.org/v4.13.0/pkg/smallsemi/> to <https://docs.gap-system.org/v4.13.0/pkg/smallsemi/doc/chap0.html>,
- from <https://docs.gap-system.org/v4.13.0/pkg/transgrp/> to <https://docs.gap-system.org/v4.13.0/pkg/transgrp/htm/chapters.htm>,
- and so on.

Since [that script](https://github.com/gap-system/GapWWW/blob/master/etc/extract_manuals.py) already is parsing the JSON data and more, it should be fairly easy to do. We use Apache, so the script could just produce a `.htaccess` file with suitable rules... (Actually it might be even easier to write a separate script which consumes a `package-infos.json` and spits out the `.htaccess` file; then it'd be very easy to apply it retroactively to prior GAP releases).

All that said, I am very busy with other stuff right now so I don't want to promise anything. If someone would like to help out, I can provide further details.

mkoeppe: Thanks a lot for these details!

> they are versioned by GAP version, not by package version.

That's good enough for us here, I think, thanks.

> add redirects ... from https://docs.gap-system.org/v4.13.0/pkg/smallsemi/ to https://docs.gap-system.org/v4.13.0/pkg/smallsemi/doc/chap0.html,

That could be nice.

More important for use would be a way to link to specific functions in a stable (or computable) way though. For example, https://docs.gap-system.org/doc/ref/chap16.html#X821888E77EB43F67 leads to Permanent, but can I assume that this is a permalink?

There's no urgency for either of these things. Here on this PR I'm really just making a first approximation with the goal of normalizing/encouraging creating richer documentation links to our upstream projects. We can revisit this when you have a bit more time!

mkoeppe: @fingolfin For example, might there be interest on your side in extending GAPDoc so it can output RST files and then go through Sphinx for the HTML build?

fingolfin: Those anchors are hashes and are essentially permanent. For our website we basically keep an index file (see https://github.com/gap-system/GapWWW/blob/master/_data/help.json which is exported from GAP via [this script](https://github.com/gap-system/gap/blob/master/dev/releases/HelpLinks-to-JSON.g).) 

Regarding "output RST files and then go through Sphinx for the HTML build": I am not really familiar with RST files and Sphinx. I am pretty sure I'd have a hard time convincing most GAP package maintainers to learn about those and installing relevant parts. But I guess it might still be potentially useful as an optional backend with added features (search, the ability to keep multiple versions and switch between them come to mind) that perhaps then is only used by the GAP website (and possibly via GH Action jobs also on package websites, if we can provide package authors with a "drop in and forget" solution). 

As such, I would find that potentially interesting. To be fully useful, also the GAP help system would have to be taught to deal with the resulting HTML (GAP can be set up so that the REPL help opens the HTML docs at the correct position). I guess this is less important if it only is about producing online versions of those docs.

I have zero interest in working on that, but if someone else did, I'd be happy to try it out and give feedback.

I have no idea if there is any chance to get such a patch into GAPDoc, though (the maintainer sometimes has... strong oppinions). But it might not even be necessary, GAPDoc already has three "backends" ("plain text with escape sequences" for the REPL help; HTML; PDF-via-latex), and I think this is done in a way that makes it easy to add extra backends. So such a thing could perhaps be done via an extension package.

fingolfin: From the GAPDoc:
```
  # - add hash strings of the latter for links in HTML and PDF version
  #   (such that links remain stable as long as the (sub)section exists
  #   and stays in the same chapter)
```

mkoeppe: > I guess it might still be potentially useful as an optional backend with added features (search, the ability to keep multiple versions and switch between them come to mind) that perhaps then is only used by the GAP website (and possibly via GH Action jobs also on package websites, if we can provide package authors with a "drop in and forget" solution).

Yes, that's what I would imagine. No need to bring local building by GAP users/developers into the equation.

> I have zero interest in working on that

Likewise, but that's a kind of project that would pretty easy to hire someone for. No math required!

fingolfin: Ah, I wish I could just hire someone for that kind of work, but in general this is very difficult with German university funding and general regulations on hiring people :/. Essentially I can hire local students (but good luck finding one with the skills). *sigh*

kwankyu:
looks good to me except that `:qepcad:`QEPCAD: Entering formulas <user/EnterForm>` fails for me. The website is down or gone?

mkoeppe:
The link https://www.usna.edu/CS/qepcadweb/B/user/EnterForm.html works just fine here.

kwankyu:
Even `https://www.usna.edu` does not work for me. 

OK then.

kwankyu (approved):
LGTM.

mkoeppe:
Thank you!

src/sage/combinat/designs/incidence_structures.py
```diff
@@ -1810,8 +1810,7 @@ def dual(self, algorithm=None):
 
         REFERENCE:
 
-        - Soicher, Leonard, Design package manual, available at
-          https://www.gap-system.org/Manuals/pkg/design/htm/CHAP003.htm
+        - Soicher, Leonard, :gap_package:`Design package manual <design/htm/CHAP003.htm>`
         """
```
kwankyu, line 1814: I got a request from someone to replace "Soicher, Leonard" with "Leonard Soicher". Please.

mkoeppe: OK, let me press Cmd-G and make this change

src/sage/combinat/designs/incidence_structures.py
```diff
@@ -1810,8 +1810,7 @@ def dual(self, algorithm=None):
 
         REFERENCE:
 
-        - Soicher, Leonard, Design package manual, available at
-          https://www.gap-system.org/Manuals/pkg/design/htm/CHAP003.htm
+        - Soicher, Leonard, :gap_package:`Design package manual <design/htm/CHAP003.htm>`
```
mkoeppe:
```suggestion
        - Leonard Soicher, :gap_package:`Design package manual <design/htm/CHAP003.htm>`
```

src/sage/combinat/designs/incidence_structures.py
kwankyu: Thanks. What do you mean by "pressing Cmd-G"??

kwankyu: Is it an emacs command?

kwankyu: Perhaps searching next item?

kwankyu: Ah, I got it: "interrupting current commands" :-)

mkoeppe: Here on GitHub, if you press Cmd-G in a comment, it creates a "suggestion"

kwankyu: Ah I see. Thanks.

GiacomoPope:
This is making several RST lint checks fail

```
Jack: sage % ./sage -tox -e rst -- src/sage/algebras/quantum_groups/quantum_group_gap.py
rst: commands[0] /Users/Jack/sage/sage/src> flake8 --select=RST sage/algebras/quantum_groups/quantum_group_gap.py
sage/algebras/quantum_groups/quantum_group_gap.py:8:1: RST304 Unknown interpreted text role "gap_package".
rst: exit 1 (0.77 seconds) /Users/Jack/sage/sage/src> flake8 --select=RST sage/algebras/quantum_groups/quantum_group_gap.py pid=5868
  rst: FAIL code 1 (0.85=setup[0.07]+cmd[0.77] seconds)
  evaluation failed :( (0.99 seconds)
```

GiacomoPope:
I made an issue: https://github.com/sagemath/sage/issues/37711

fchapoton:
this requires to add these also in `src/tox.ini`

PR #37599, opened 2024-03-12
extend the definition of q-binomial coefficients to (k > 0, n < 0)

fchapoton:
This is extending the domain of the q-binomial coefficients.

This is done by polynomiality wrt n, for k >= 0.

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/q_analogues.py
+++ b/src/sage/combinat/q_analogues.py
@@ -16,6 +16,8 @@ from sage.combinat.partition import _Partitions
 from sage.misc.cachefunc import cached_function
 from sage.misc.misc_c import prod
 from sage.rings.integer_ring import ZZ
+from sage.rings.polynomial.laurent_polynomial_ring import LaurentPolynomialRing
+from sage.rings.polynomial.polynomial_ring import polygen
 from sage.structure.element import parent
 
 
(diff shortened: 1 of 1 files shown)


src/sage/combinat/q_analogues.py
```diff
@@ -328,6 +330,11 @@ def q_binomial(n, k, q=None, algorithm='auto'):
         from sage.rings.polynomial.polynomial_element import Polynomial
         is_polynomial = isinstance(q, Polynomial)
 
+    if n >= 0:
+        k = min(n - k, k)  # Pick the smallest k
+    else:
+        return (-1)**k * q**(k * n - (k * k - k) // 2) * q_binomial(-n + k - 1, k, q=q)
```
tscrim:
```suggestion
    if n < 0:
        return (-1)**k * q**(k * n - (k * k - k) // 2) * q_binomial(-n + k - 1, k, q=q)

    k = min(n - k, k)  # Pick the smallest k        
```
More of a stylistic thing, but I find this easier to follow the logic.

tscrim (changes requested):
LGTM overall. I have one minor code stylistic change, and one question. Should the default ring for the $n < 0$ case be $\mathbb{Z}[q,q^{-1}]$ instead of the fraction field of $\mathbb{Z}[q]$? This has properties closer to the polynomial ring and arithmetic is faster. What do you think?

fchapoton:
yes, right, this is natural. Done, also for `q_int`.

tscrim (approved):
Thanks. LGTM.

(Personally, I am not a fan of `polygen` as it is less explicit, but that's just my preference.)

PR #37604, opened 2024-03-13
Remove duplicated functions from genus two hyperelliptic curves

GiacomoPope:
At some point, four functions seem to have been duplicated in a class. This PR deletes the duplicates.


fchapoton (approved):
ok, thanks. I have been seing this one for so long...

PR #37605, opened 2024-03-13
Ensure returned degree of multivariate polynomial is type `Integer` for `MPolynomialRing_libsingular` class

GiacomoPope:
Previously the degree of multivariate polynomials was returned as a python `int` instead of a SageMath `Integer` resulting in ugly `ZZ(f.degree())` calls in various places.

This PR only focuses on the case of `MPolynomialRing_libsingular` to keep the noise of the PR as low as possible.

**Future work**: This same issue is in `MPolynomial_polydict`, I am happy to also do the work here, or make a new PR -- really I don't like sage returning a `int` when we dont have to.

**Edit**

This follow up work has been done in PR: https://github.com/sagemath/sage/pull/37611


mkoeppe (approved):
LGTM.

PR #37606, opened 2024-03-13
Border matrix

mkoeppe:
Plain TeX users may remember `\bordermatrix`. Here we implement this as options of the `Matrix` class's `str` method.
```
            sage: M = matrix([[1,2,3], [4,5,6], [7,8,9]])
            sage: M.subdivide(None, 2)
            sage: print(M.str(unicode=True,
            ....:             top_border=['ab', 'cde', 'f'],
            ....:             bottom_border=['*', '', ''],
            ....:             left_border=[1, 10, 100],
            ....:             right_border=['', ' <', '']))
                        ab cde   f
                     1⎛  1   2│  3⎞
                    10⎜  4   5│  6⎟ <
                   100⎝  7   8│  9⎠
                         *
```

Follow-up PR: As the guiding application for this feature, we equip finite-dimensional modules with basis with methods `_repr_matrix`, `_ascii_art_matrix`, `_unicode_art_matrix` that can be used as in this example:
```
                sage: M = matrix(ZZ, [[1, 0, 0], [0, 1, 0]],
                ....:            column_keys=['a', 'b', 'c'],
                ....:            row_keys=['v', 'w']); M
                Generic morphism:
                From: Free module generated by {'a', 'b', 'c'} over Integer Ring
                To:   Free module generated by {'v', 'w'} over Integer Ring
                sage: M._unicode_art_ = M._unicode_art_matrix
                sage: unicode_art(M)                            # indirect doctest
                    a b c
                  v⎛1 0 0⎞
                  w⎝0 1 0⎠
```

### :memo: Checklist

### DIFF
--- a/src/sage/matrix/matrix0.pyx
+++ b/src/sage/matrix/matrix0.pyx
@@ -1795,7 +1795,9 @@ cdef class Matrix(sage.structure.element.Matrix):
         return self.str()
 
     def str(self, rep_mapping=None, zero=None, plus_one=None, minus_one=None,
-            *, unicode=False, shape=None, character_art=False):
+            *, unicode=False, shape=None, character_art=False,
+            left_border=None, right_border=None,
+            top_border=None, bottom_border=None):
         r"""
         Return a nice string representation of the matrix.
 
@@ -1919,6 +1921,21 @@ cdef class Matrix(sage.structure.element.Matrix):
             sage: A
             [ 0.333333333333333   66.6666666666667]
             [ -3.00000000000000 1.00000000000000e6]
+
+        Matrices with borders::
+
+            sage: M = matrix([[1,2,3], [4,5,6], [7,8,9]])
+            sage: M.subdivide(None, 2)
+            sage: print(M.str(unicode=True,
+            ....:             top_border=['ab', 'cde', 'f'],
+            ....:             bottom_border=['*', '', ''],
+            ....:             left_border=[1, 10, 100],
+            ....:             right_border=['', ' <', '']))
+                   ab cde   f
+               ab⎛  1   2│  3⎞
+              cde⎜  4   5│  6⎟ <
+                f⎝  7   8│  9⎠
+                    *
 
         TESTS:
 
@@ -2062,6 +2079,12 @@ cdef class Matrix(sage.structure.element.Matrix):
 
         # compute column widths
         S = []
+        if top_border is not None:
+            for x in top_border:
+                S.append(str(x))
+            top_count = 1
+        else:
+            top_count = 0
         for x in entries:
             # Override the usual representations with those specified
             if callable(rep_mapping):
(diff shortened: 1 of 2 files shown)


src/sage/matrix/matrix0.pyx
```diff
@@ -1920,6 +1922,21 @@ cdef class Matrix(sage.structure.element.Matrix):
...
+            sage: M = matrix([[1,2,3], [4,5,6], [7,8,9]])
+            sage: M.subdivide(None, 2)
+            sage: print(M.str(unicode=True,
+            ....:             top_border=['ab', 'cde', 'f'],
+            ....:             bottom_border=['*', '', ''],
+            ....:             left_border=[1, 10, 100],
+            ....:             right_border=['', ' <', '']))
+                        ab cde   f
+                     1⎛  1   2│  3⎞
+                    10⎜  4   5│  6⎟ <
```
dcoudert: Why not adding a space between the left border and the left parenthesis ?

src/sage/matrix/matrix0.pyx
```diff
@@ -2074,39 +2097,83 @@ cdef class Matrix(sage.structure.element.Matrix):
...
-
-        last_row = len(rows) - 1
-        if last_row == 0:
-            rows[0] = slb + rows[0] + srb
+            if bottom_border is not None and 0 <= r < nr:
+                right.append(str(right_border[r]))
+            else:
+                right.append("")
+        else:
+            if nr == nr + bottom_count:
```
dcoudert, line 2188: Do you really want to check if `bottom_count == 0` ?

dcoudert (changes requested):
You could also check corner cases: empty matrix, small dimensions

src/sage/matrix/matrix0.pyx
mkoeppe: yes, that's the same. i've written it in this peculiar way to match the `for` loop.

src/sage/matrix/matrix0.pyx
mkoeppe: wasn't sure about this one. i thought it would be too far away from the parenthesis

mkoeppe: i could make this another parameter, of course...

mkoeppe:
> You could also check corner cases: empty matrix, small dimensions

Done now -- which caught a bug (fixed now)

src/sage/matrix/matrix0.pyx
dcoudert: It's ok like that.

However, it took me some time to understand the number of left spaces before the values of the left border. These spaces have nothing to do with the call to `rjust(left_width)` but are choices of the author in the presentation of the doctests. I suggest to remove the extra spaces to be consistent with what a user obtain in a terminal. It will also ease the work of future maintainers.

dcoudert (changes requested):

src/sage/matrix/matrix0.pyx
mkoeppe: Thanks, done in d45422e64ea

dcoudert (approved):
LGTM.

PR #37608, opened 2024-03-14
src/sage/modules: Doctest cosmetics

mkoeppe:
Standard reformatting of doctests and their outputs

Split out from https://github.com/sagemath/sage/pull/35095

### :memo: Checklist


fchapoton (approved):
really not useful at all, but ok

mkoeppe:
Thanks.

I have to disagree though: The code snippets in our documentation are the most visible part of our source code. It is important that we use a coding style there that follows standard practices, so that it is recognizable to viewers as contemporary Python. And the reformatting of output to fit in the margins and to expose the structure rather than obscuring structure addresses a major usability issue.

PR #37611, opened 2024-03-14
Ensure degree and total degree return `Integer` type for `MPolynomial_polydict` class

### LINKED ISSUE #37603: Multivariate polynomial ring returns degree as an `int`
### Steps To Reproduce

```py
sage: R.<x, y> = PolynomialRing(GF(5))
sage: f = R.random_element(5)
sage: type(f)
<class 'sage.rings.polynomial.multi_polynomial_libsingular.MPolynomial_libsingular'>
sage: type(f.degree())
<class 'int'>
```

```py
sage: R.<x,y,z> = PolynomialRing(QQbar)
sage: f = x + y + z
sage: type(f)
<class 'sage.rings.polynomial.multi_polynomial_element.MPolynomial_polydict'>
sage: type(f.degree())
<class 'int'>
```

```py
sage: R.<x> = PolynomialRing(GF(5))
sage: f = R.random_element(5)
sage: type(f)
<class 'sage.rings.polynomial.polynomial_zmod_flint.Polynomial_zmod_flint'>
sage: type(f.degree())
<class 'sage.rings.integer.Integer'>
```

```py
sage: R.<x> = PolynomialRing(QQbar)
sage: f = 1 + x
sage: type(f)
<class 'sage.rings.polynomial.polynomial_ring.PolynomialRing_field_with_category.element_class'>
sage: type(f.degree())
<class 'sage.rings.integer.Integer'>
```

### Expected Behavior

The degree should be a sage type `Integer`

### Actual Behavior

The degree is be a python type `int` for `MPolynomial_polydict` and `MPolynomial_libsingular`

### Additional Information

_No response_

### Environment

```markdown
- **Sage Version**: SageMath version 10.3.rc3, Release Date: 2024-03-10
```

### Checklist

GiacomoPope:
This is a follow up PR to https://github.com/sagemath/sage/pull/37605 which ensures that `degree()` and `total_degree()` return a sage `Integer` type instead of a python `int` for the `MPolynomial_polydict` class.

However, `f.degrees()` returns something of type `sage.rings.polynomial.polydict.ETuple` which has elements as `int`. I am loathed to change this to return `Integer` in case of performance regression, so the following status of `MPolynomial_polydict` is the following:

```py
sage: R.<x, y> = PolynomialRing(QQbar)
sage: f = 1 + x + y^2
sage: type(f.degree())
<class 'sage.rings.integer.Integer'>
sage: type(f.degree(x))
<class 'sage.rings.integer.Integer'>
sage: type(f.degree(y))
<class 'sage.rings.integer.Integer'>
sage: type(f.total_degree())
<class 'sage.rings.integer.Integer'>
sage: type(f.degrees())
<class 'sage.rings.polynomial.polydict.ETuple'>
sage: type(f.degrees()[0])
<class 'int'>
```

I would like advice on how to proceed with the `degrees()` function. Is leaving it in it's current state OK?

Fixes https://github.com/sagemath/sage/issues/37603


src/sage/rings/polynomial/multi_polynomial_element.py
```diff
@@ -668,11 +668,22 @@ def degree(self, x=None, std_grading=False):
...
+            <class 'sage.rings.integer.Integer'>
+            sage: type(f.degree(x)) == type(f.degree(y)) == type(f.degree(z))
+            True
         """
         if x is None:
             if std_grading or not self.parent().term_order().is_weighted_degree_order():
-                return self.element().degree(None)
-            return self.weighted_degree(self.parent().term_order().weights())
+                return Integer(self.element().degree(None))
+            return Integer(self.weighted_degree(self.parent().term_order().weights()))
```
grhkm21: This line is not necessary since [weighted_degree](https://github.com/sagemath/sage/blob/develop/src/sage/rings/polynomial/multi_polynomial.pyx#L2079) is wrapped.

grhkm21 (changes requested):
Just a small change.

For future reference, I considered moving the wrapper on L685 to the underlying PolyDict, but that is not desirable since `polydict` is a "general" dictionary-like class and hence should be fast.

GiacomoPope:
> but that is not desirable since polydict is a "general" dictionary-like class and hence should be fast.

This is why I didnt make a patch for `degrees()` as well. I don't know what the right thing to do here is.

src/sage/rings/polynomial/multi_polynomial_element.py
GiacomoPope: Thank you

grhkm21 (approved):

GiacomoPope:
Thanks!

PR #37613, opened 2024-03-14
Fix hyperelliptic curve dynamic class construction to allow proper method inheritance

### LINKED ISSUE #37612: Hyperelliptic curve constructor has wrong inheritance
### Steps To Reproduce

```py
sage: R.<x> = QQ[]
sage: f = x^5 - x^4 + 3
sage: HyperellipticCurve(f).jacobian()
Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - x^4 + 3
sage: HyperellipticCurve(f).jacobian
<bound method HyperellipticCurve_generic.jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - x^4 + 3>
sage: type(HyperellipticCurve(f).jacobian())
<class 'sage.schemes.hyperelliptic_curves.jacobian_generic.HyperellipticJacobian_generic_with_category'>
```

### Expected Behavior

For genus two curves, the base class should be `hyperelliptic_g2` which has the function

```py
    def jacobian(self):
        """
        Return the Jacobian of the hyperelliptic curve.

        EXAMPLES::

            sage: R.<x> = QQ[]
            sage: f = x^5 - x^4 + 3
            sage: HyperellipticCurve(f).jacobian()
            Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - x^4 + 3
        """
        return jacobian_g2.HyperellipticJacobian_g2(self)
```

So for a genus two curve, we expect a Jacobian of type `HyperellipticJacobian_generic.HyperellipticJacobian_g2_generic` but this does not happen as only the `jacobian()` function of `HyperellipticCurve_generic` is in scope.

### Actual Behavior

Class construction should set the right inheritance so that the genus two functions work, or the genus two special cases should be handled within the generic class instead.

### Additional Information

_No response_

### Environment

```markdown
- **Sage Version**: 10.2
```

### Checklist

GiacomoPope:
While working on some hyperelliptic code, I realised the dynamic class construction had an issue with inheritance for the genus two classes and certain methods which should have been available were not.

This is discussed in more detail in the issue https://github.com/sagemath/sage/issues/37612

I do not know a *good* fix for this, but I have found a fix which at least allows the functions supposed to be accessed to be accessed.

I have added a small doctest to ensure that the method called for genus two curves is bound to the correct class.

## Overview of the fix

The original class was constructed as:

```py
    cls = dynamic_class(class_name,
                        tuple(superclass),
                        cls=HyperellipticCurve_generic,
                        doccls=HyperellipticCurve)
```

Where `superclass` is either an empty list, a list with a single child from:

- `HyperellipticCurve_finite_field`
- `HyperellipticCurve_rational_field`
- `HyperellipticCurve_padic_field`
- `HyperellipticCurve_g2`

Notice that all four of these classes are children of `HyperellipticCurve_generic`.

Or, in the case of the base field being one of the above AND the curve being genus two this list is of the form:

```
[HyperellipticCurve_XXX_field, HyperellipticCurve_g2]
```

In this case, I found that the dynamic class insertion into `cls` meant that the methods shared by one of the "`superclass`" (probably should be called base classes?) and `cls` itself would call the method from `cls` rather than the superclass and so all specialised functions were unavailable, making the genus two optimisations redundant.

In my new fix, if `superclass` has a length of zero, I set `cls` to be `HyperellipticCurve_generic`, otherwise `cls` is `None`.

This seems to work, but I don't know if this is good practice.

Fixes https://github.com/sagemath/sage/issues/37612


GiacomoPope:
After thinking about it a little more, I will write what I think is happening.

When we call

```py
dynamic_class(name_of_class: str, base_classes: tuple[Class], cls: Class, ...)
```

I believe we take the class supplied in `cls` and try and add all methods from the `base_classes` into `cls` and give this class a new name `name_of_class`

For the current code, this then means we take

```
HyperellipticCurve_generic.HyperellipticCurve_g2
```

As the super class and make something which looks like...

```
new_class_name = HyperellipticCurve_generic <- HyperellipticCurve_g2
```

So new_class_name has all the methods of `HyperellipticCurve_g2` inserted first into `HyperellipticCurve_generic ` which is then renamed into the new class. However, if there is a method in both `HyperellipticCurve_generic` and `HyperellipticCurve_g2`, then the method on `HyperellipticCurve_generic` is preferred which is why we lose the `jacobian()` method from `HyperellipticCurve_g2` in the above example.

So the question now is what's the right thing to do? As `HyperellipticCurve_generic` is the parent child of all others, we essentially want to make a dynamic class from all the classes found in construction. We then have

```
new_class_name(*superclass)
```

But this will fail when no special classes have been found.

I therefore think

```py
     if not superclass:
         superclass.append(HyperellipticCurve_generic)

     class_name = "_".join(cls_name)
     cls = dynamic_class(class_name,
                         tuple(superclass),
                         doccls=HyperellipticCurve)
```

Is a reasonable fix. We will either have a tuple of classes whose parent is `HyperellipticCurve_generic` and when this tuple is empty we simply use the class `HyperellipticCurve_generic` itself.

GiacomoPope:
Hi @kwankyu, I hope it's OK to ask for a review on this. I saw you had previously done great work with class refactoring and that you might be a good person to check whether what I did was satisfactory for the long term usage of these constructors.

src/sage/schemes/hyperelliptic_curves/constructor.py
```diff
@@ -265,19 +279,22 @@ def is_pAdicField(x):
...
+    if not bases:
+        bases = [HyperellipticCurve_generic]
+
+    # Dynamically build a class from multiple inheritance. Note that
+    # all classes we slect from are children of HyperellipticCurve_generic
     class_name = "_".join(cls_name)
-    cls = dynamic_class(class_name, tuple(superclass),
-                        HyperellipticCurve_generic, doccls=HyperellipticCurve)
+    cls = dynamic_class(class_name, tuple(bases), doccls=HyperellipticCurve)
     return cls(PP, f, h, names=names, genus=g)
```
kwankyu, line 299: You removed `HyperellipticCurve_generic` here. `HyperellipticCurve_generic` is supposed to provide the "generic" methods (from itself and from the base classes) to the fabricated class constructed by `dynamic_class`.  By your change, the "generic" methods are still accessible but placed farther in the method resolution order (mro). This does not look good.

Is it your objective to construct `your_jacobian_g2_class` object for the `HyperellipticCurve_g2` object? Then there is a standard way to solve the problem:

You add a class variable `_jacobian_class` to `HyperellipticCurve_generic` and make its `jacobian` method to construct an object of the class `_jacobian_class`. Then the subclass `HyperellipticCurve_g2` set the class variable `_jacobian_class = your_jacobian_g2`. You don't need to provide `jacobian` method to `HyperellipticCurve_g2`. 

See `sage.rings.function_field.function_field.FunctionField`, which has class variable `_differentials_space` for similar purpose.

GiacomoPope: I'm sorry. I think I have not clearly identified the issue above. The issue is about inheritance with the current construction. Say we have some methods on a hyperelliptic curve. Many of these can be generic ones suitable for the parent `HyperellipticCurve_generic`.

However, sometimes for the case when we know we are also on a genus two curve we may have methods which are more efficient and so should overload the generic method. We also may have methods which are only possible in genus two. The same could for example happen for certain fields.

Consider the following dummy parent class and children classes:

```py
class Generic():
    def a(self):
        print("generic a")

    def b(self):
        print("generic b")

    def c(self):
        print("generic c")

class GenusTwo(Generic):
    def b(self):
        print("genus two optimised b")

    def d(self):
        print("genus two d")

class FiniteField(Generic):
    def c(self):
        print("finite field optimised c")

    def e(self):
        print("finite field e")
```

This is a minimal example of the above. We now want to try and construct a class which inherits all the generic methods, but uses the optimised implementations when available if we are on a genus two curve over a finite field.

The current implementation constructs the following class:

```py
cls = dynamic_class("new_class_name", (GenusTwo, FiniteField), cls=Generic)
BrokenExample = cls()
```

Let us now call methods on this class:

```py
sage: cls = dynamic_class("new_class_name", (GenusTwo, FiniteField), cls=Generic)
sage: BrokenExample = cls()
sage: BrokenExample.a()
generic a
sage: BrokenExample.b()
generic b
sage: BrokenExample.c()
generic c
sage: BrokenExample.d()
genus two d
sage: BrokenExample.e()
finite field e
```

You see that we indeed get these extra methods `d()` and `e()`, but the optimised implementations for `b()` and `c()` are discarded because of the use of the dynamic class creation.

To make this a little more clear, imagine we have some other class `Dead` which is not a child of `Generic`.

```py
sage: class Dead:
....:     def a(self):
....:         print("dead a")
....: 
....:     def b(self):
....:         print("dead b")
....: 
....:     def c(self):
....:         print("dead c")
....:         
sage: cls = dynamic_class("new_class_name", (GenusTwo, FiniteField), cls=Dead)
sage: DeadExample = cls()
sage: DeadExample.a()
dead a
sage: DeadExample.b()
dead b
sage: DeadExample.c()
dead c
sage: DeadExample.d()
genus two d
sage: DeadExample.e()
finite field e
```

Maybe now it is easy to see what happens. The methods from the tuple of classes in `(GenusTwo, FiniteField)` are inserted into the class `cls`, but `cls` takes priority for inheritance and so kills all methods below it.

Now. My proposed fix is to stop using `cls` as an argument here, as I think it is being used incorrectly.

Consider for example the new case:

```py
sage: cls = dynamic_class("new_class_name", (GenusTwo, FiniteField))
sage: FixedExample = cls()
sage: FixedExample.a()
generic a
sage: FixedExample.b()
genus two optimised b
sage: FixedExample.c()
finite field optimised c
sage: FixedExample.d()
genus two d
sage: FixedExample.e()
finite field e
```

As both `GenusTwo` and `FiniteField` are children of `Generic` we inherit `a` in our constructed class, but we also get the optimised methods for `b()` and `c()` together with the new methods `d()` and `e()`.

The only case we must be careful of is when in construction the tuple `()` is empty in the dynamic class. Here, we must then simply use the class `Generic` to maintain the core methods.

This is why I believe my fix works with the current code made for hyperelliptic curves.

kwankyu: I understand how your fix works. 

What I suggested above only works if what you want is only `jacobian()` method.

It seems that you want to override more methods of `HyperellipticCurve_generic` in your subclass. Then the problem is normally solved by python's native inheritance mechanism of classes. Unfortunately `HyperellipticCurve` is not using the normal mechanism, but is using `dynamic_class`. 

If you want to stick to `dynamic_class` mechanism, perhaps what you did in this PR seems the only way to solve the problem.

If you want to switch to inheritance of classes, it means more work to you. You need to define more subclasses of `HyperellipticCurve_generic`. I believe this is doable. You may check `Curve` classes: https://github.com/sagemath/sage/blob/develop/src/sage/schemes/curves/projective_curve.py

I think `dynamic_class` was invented to solve the problem of combinatorial explosion of classes in the categories framework. It was not intended for the current case that defining a few classes in the right way solves the problem.

GiacomoPope: OK -- well I can do the extra work if this is the only way to get this PR merged. Otherwise I believe this PR is a way to fix the current code which does not work as intended.

I have just written something like:

```py
    # -----------------------
    # Class selection
    # -----------------------

    # Special class for finite fields
    if base_ring in FiniteFields():
        if genus == 2:
            cls = HyperellipticCurveSmoothModel_g2_finite_field
        else:
            cls = HyperellipticCurveSmoothModel_finite_field
    # Special class for padic fields
    elif isinstance(base_ring, pAdicField):
        if genus == 2:
            cls = HyperellipticCurveSmoothModel_g2_padic_field
        else:
            cls = HyperellipticCurveSmoothModel_padic_field
    # Special class for rational fields
    elif is_RationalField(base_ring):
        if genus == 2:
            cls = HyperellipticCurveSmoothModel_g2_rational_field
        else:
            cls = HyperellipticCurveSmoothModel_rational_field
    # Default class for all other fields
    else:
        if genus == 2:
            cls = HyperellipticCurveSmoothModel_g2
        else:
            cls = HyperellipticCurveSmoothModel_generic     
```

For the case of the new implementation I am working on: https://github.com/GiacomoPope/HyperellipticCurves/

GiacomoPope: I suppose here the dynamic class creation may end up being helpful if there are optimisations for many different genus -- but I can't see this happening

kwankyu: > OK -- well I can do the extra work if this is the only way to get this PR merged. 

There seems to be misunderstanding. I didn't request any extra work for me to approve this PR. 

If you would want to abandon using `dynamic_class` and adopt inheritance mechanism to achieve what you want, then you would need an extra work (to define needed subclasses). That's all.

If you are okay with tweaking the `dynamic_class` mechanism already set in place  (and I guess so), then I am okay with it too.

GiacomoPope: Sorry for the misunderstanding. 

I don't really mind what the fix is, I just want to do the correct fix. I am happy to either leave it as it is or make these additional classes.

Looking through the current usage of the dynamic class there are things like...

```
class = base_class
if condition:
    class = dynamic_class("name", (class, additional_class))
```
Which we could also do here but i dont think it's any cleaner than what we currently have

kwankyu: Let's minimize changes for now.

Later you may attempt to refactor subclasses of `HyperellipticCurve_generic` to get rid of `dynamic_class` if you want for whatever reason.

I think the use of `dynamic_class` should be minimized as this mechanism is specific to sage. So perhaps I myself may attempt to do it someday :-)

src/sage/schemes/hyperelliptic_curves/constructor.py
```diff
@@ -203,59 +203,73 @@ def HyperellipticCurve(f, h=0, names=None, PP=None, check_squarefree=True):
...
+                    "check_squarefree=False to skip this check"
+                )
         if not smooth:
-            raise ValueError("Not a hyperelliptic curve: "
-                             "singularity in the provided affine patch.")
+            raise ValueError(
+                "not a hyperelliptic curve: "
+                "singularity in the provided affine patch"
+            )
     R = P.base_ring()
```
kwankyu, line 255: Combine the two lines or keep the original.

src/sage/schemes/hyperelliptic_curves/constructor.py
```diff
@@ -203,59 +203,73 @@ def HyperellipticCurve(f, h=0, names=None, PP=None, check_squarefree=True):
...
-    superclass = []
+    bases = []
     cls_name = ["HyperellipticCurve"]
 
+    # For certain genus we specialise to child classes with
+    # optimised methods
     genus_classes = {2: HyperellipticCurve_g2}
+    if g in genus_classes:
+        bases.append(genus_classes[g])
+        cls_name.append("g%s" % g)
```
kwankyu, line 269: Use f-string `f"g{g}"`?

src/sage/schemes/hyperelliptic_curves/constructor.py
```diff
@@ -203,59 +203,73 @@ def HyperellipticCurve(f, h=0, names=None, PP=None, check_squarefree=True):
...
 
+    # For certain base fields, we specialise to child classes
+    # with special case methods
     def is_FiniteField(x):
```
kwankyu, line 272: I think "subclass" is the usual python term rather than "child class".

src/sage/schemes/hyperelliptic_curves/hyperelliptic_g2.py
```diff
@@ -45,6 +45,15 @@ def jacobian(self):
...
+
+        TESTS:
+
+        Ensure that :issue:`37612` is fixed::
+
+            sage: R.<x> = QQ[]
+            sage: f = x^5 - x^4 + 3
+            sage: HyperellipticCurve(f).jacobian
+            <bound method HyperellipticCurve_g2.jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - x^4 + 3>
         """
```
kwankyu, line 57: Instead you may test the type of the object returned by `HyperellipticCurve(f).jacobian()`. I think this is more to the point of the test.

GiacomoPope:
Thank you for the comments.

src/sage/schemes/hyperelliptic_curves/constructor.py
```diff
@@ -265,19 +278,22 @@ def is_pAdicField(x):
...
             break
 
+    # If no specialised child class was found, we simply use the
+    # generic class in the class construction
+    if not bases:
+        bases = [HyperellipticCurve_generic]
+
+    # Dynamically build a class from multiple inheritance. Note that
+    # all classes we slect from are children of HyperellipticCurve_generic
     class_name = "_".join(cls_name)
```
kwankyu, line 297: `all classes we select from are subclasses of HyperellipticCurve_generic`

src/sage/schemes/hyperelliptic_curves/constructor.py
```diff
@@ -265,19 +278,22 @@ def is_pAdicField(x):
...
     for name, test, cls in fields:
         if test(R):
-            superclass.append(cls)
+            bases.append(cls)
             cls_name.append(name)
             break
 
+    # If no specialised child class was found, we simply use the
+    # generic class in the class construction
+    if not bases:
```
kwankyu, line 292: `no specializ(or s)ed subclass was found, ...`

kwankyu:
Would you rebase to the latest develop?

GiacomoPope:
Updated and merged.

kwankyu (approved):
Thanks. LGTM.

PR #37615, opened 2024-03-15
Removed `BUG` method and include a docstring for `matrix_of_frobenius()`

GiacomoPope:
When reviewing code for hyperelliptic curves I found a method labeled bug with a typo in the name which was unnecessary. I have removed this method and included doctests which were previously absent.

### :memo: Checklist


GiacomoPope:
@kwankyu maybe you'll have time to review this? One question have is why is this method defined for curves over the rationals if we explicitly change the ring anyway? Feels like maybe I should move it to generic?

grhkm21:
Because only rational elements are coercable to p-adic numbers, probably.

```sage
sage: K.<a> = NumberField(x^3 + 2)
sage: E = EllipticCurve(K, [2 * a + 3, 1])
sage: monsky_washnitzer.matrix_of_frobenius_hyperelliptic(E.change_ring(pAdicField(3, 53)))
# TypeError: Unable to coerce 2*a + 3 to a rational
```

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
```diff
@@ -20,21 +20,48 @@ class HyperellipticCurve_rational_field(hyperelliptic_generic.HyperellipticCurve
...
+        INPUT:
 
-        # BUG: should get this method from HyperellipticCurve_generic
-        def my_chage_ring(self, R):
-            from .constructor import HyperellipticCurve
-            f, h = self._hyperelliptic_polynomials
-            y = self._printing_ring.gen()
-            x = self._printing_ring.base_ring().gen()
-            return HyperellipticCurve(f.change_ring(R), h, "%s,%s" % (x,y))
+        - ``p`` (prime integer or pAdic ring / field ) -- if ``p`` is an integer,
```
grhkm21:
```suggestion
        - ``p`` (prime integer or pAdic ring / field) -- if ``p`` is an integer,
```

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
```diff
@@ -20,21 +20,48 @@ class HyperellipticCurve_rational_field(hyperelliptic_generic.HyperellipticCurve
...
+            sage: H.matrix_of_frobenius(3, prec=2)
+            [        O(3^2)     3 + O(3^2)         O(3^2)         O(3^2)]
+            [    3 + O(3^2)         O(3^2)         O(3^2) 2 + 3 + O(3^2)]
+            [  2*3 + O(3^2)         O(3^2)         O(3^2)    3^-1 + O(3)]
+            [        O(3^2)         O(3^2)     3 + O(3^2)         O(3^2)]
+        """
         import sage.schemes.hyperelliptic_curves.monsky_washnitzer as monsky_washnitzer
         if isinstance(p, (sage.rings.abc.pAdicField, sage.rings.abc.pAdicRing)):
             K = p
         else:
```
grhkm21, line 62: We probably want to check that `p` is a prime integer

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
```diff
@@ -20,21 +20,48 @@ class HyperellipticCurve_rational_field(hyperelliptic_generic.HyperellipticCurve
...
-            f, h = self._hyperelliptic_polynomials
-            y = self._printing_ring.gen()
-            x = self._printing_ring.base_ring().gen()
-            return HyperellipticCurve(f.change_ring(R), h, "%s,%s" % (x,y))
+        - ``p`` (prime integer or pAdic ring / field ) -- if ``p`` is an integer,
+          constructs a ``pAdicField`` with ``p`` to compute the matrix of
+          Frobenius, otherwise uses the supplied pAdic ring or field.
 
+        - ``prec`` (optional) -- if ``p`` is an prime integer, the `p`-adic
+          precision of the coefficient ring constructed
```
grhkm21:
```suggestion
          precision of the coefficient ring constructed
```

grhkm21 (changes requested):
General comments.

Fun fact: The bug is from [this line](https://github.com/sagemath/sage/commit/08a4b9d36b3abbff0a5f211e47cea5f880e8f05c#diff-0ef5c5afc8dd043ccc00974a531d5673310ec7f1e51f870d04cffcbcbb4c83d1R58). No further comment.

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
GiacomoPope: There's no change in this suggestion?

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
GiacomoPope: I was just relying on the underlying construction to return an error in this case rather than handling it in this call. I can add the extra check though?

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
grhkm21: Oh oops. Add a full stop.

src/sage/schemes/hyperelliptic_curves/hyperelliptic_rational_field.py
grhkm21: Either way is fine, my personal preference is the check to be done when needed

GiacomoPope: I think I'll leave it and allow the field constructor to catch the bad p or prec inputs as that check is done anyway

grhkm21:
If CI passes then it looks good to me

GiacomoPope:
CI failure from some other unrelated bug i think

mkoeppe:
> CI failure from some other unrelated bug i think

Yes, #37734 is unrelated and fixed in #37750

GiacomoPope:
Thanks for helping with the labels and confirming about the failure.

PR #37620, opened 2024-03-16
`pkgs/sage-conf_pypi`: Skip configure/make when `SAGE_CONF_FILE` or `SAGE_CONF_ENV_FILE` are set

mkoeppe:
The distribution package sage-conf as found on PyPI is the version from pkgs/sage-conf_pypi, which creates a SAGE_ROOT and then runs `configure` and `make`.

Here we make the same distribution package usable in a different way: Set `SAGE_CONF_FILE` or/and `SAGE_CONF_ENV_FILE` to whatever static file has your configuration, or just `/dev/null`, and that will be taken.

This may be a useful simplification for downstream packagers.

We also add another environment variable: `SAGE_CONF_CONFIGURE_ARGS` and add some documentation.

### :memo: Checklist


mkoeppe:
@kiwifb Is this roughly what would enable you to use sage-conf from PyPI?

kiwifb:
I'll have to try it, but on the description, that would avoid most of the patching I do. That could motivate me, to fix all the issues that come with trying to ship the `sage-config` script.

pkgs/sage-conf_pypi/setup.py
```diff
@@ -26,57 +26,82 @@
...
+                try:
+                    shutil.copyfile(SAGE_CONF_ENV_FILE or '/dev/null',
+                                    os.path.join(HERE, 'bin', 'sage-env-config'))
+                except OSError:
+                    raise SetupError(f"environment variable SAGE_CONF_ENV_FILE is set to {SAGE_CONF_ENV_FILE}, "
+                                     f"but the file cannot be copied")
+            else:
+                # build_scripts skips empty files, so create something nonempty
+                with open(os.path.join(HERE, 'bin', 'sage-env-config'), 'w') as f:
+                    print("# -*- shell-script -*-", file=f)
```
kiwifb: I am perfectly happy with it being skipped. I am currently not shipping this file in any way shape or form and everything is fine. The main consumer at runtime is the sage script and it can cope with it not being present https://github.com/sagemath/sage/blob/e417e2205be84d6d567b8897527fa6945ad09bdb/src/bin/sage#L176
So, I'd rather skip it rather than having a token file.

mkoeppe: The problem is that after skipping the installation, setuptools gives an error because the file is named in the metadata but is not present...

kiwifb: Probably why I currently remove it from setup.cfg. I must have seen that error some time ago. I guess, that acceptably would avoid the patching then.

mkoeppe: > I'd rather skip it rather than having a token file.

Done in d85432a569d

mkoeppe: Let's get this in please.

orlitzky:
sage-conf was removed in c8a83f50c20f3d1e787ac950faa2149f303650ea

PR #37628, opened 2024-03-18
use the polynomial ring also for non-commutative baserings as _internal_poly_ring

### LINKED ISSUE #37625: Squares not working in completion of free algebra
### Steps To Reproduce

```
sage: R=algebras.Free(QQ,('A','B'),degrees=(1,2))
sage: Rc=R.completion()
sage: A,B=R.gens()
sage: Rc(A)+1
1 + A
sage: Rc(A)**2
...
TypeError: unable to convert (A,) to a rational
```

### Expected Behavior

This should return `A^2`

### Actual Behavior

raise TypeError

### Additional Information

_No response_

### Environment

```markdown
- **OS**:ubuntu 22.04
- **Sage Version**:10.3.rc4
```

### Checklist

mantepse:
The `LazyCompletionGradedAlgebra` used a univariate `FreeAlgebra` as `_internal_poly_ring`, which doesn't work for 'exact' lazy series, i.e., when the degree is finite and all coefficients are known.

This was done, because it was assumed that there are issues with univariate polynomials over a non-commutative base ring.  However, it seems that these issues are not visible anymore.

Fixes #37625.

### DIFF
--- a/src/sage/rings/lazy_series_ring.py
+++ b/src/sage/rings/lazy_series_ring.py
@@ -2716,6 +2716,13 @@ class LazyCompletionGradedAlgebra(LazySeriesRing):
             ...
             ValueError: basis should be in GradedAlgebrasWithBasis
 
+        Check that :issue:`37625` is fixed::
+
+            sage: R = algebras.Free(QQ, ('a', 'b'), degrees=(1, 2))
+            sage: L = R.completion()
+            sage: a, b = R.gens()
+            sage: (L(a) + L(b))^2
+            a^2 + (a*b+b*a) + b^2
         """
         base_ring = basis.base_ring()
         self._minimal_valuation = 0
@@ -2738,11 +2745,7 @@ class LazyCompletionGradedAlgebra(LazySeriesRing):
         Parent.__init__(self, base=base_ring, category=category)
         self._sparse = sparse
         self._laurent_poly_ring = basis
-        if self._laurent_poly_ring not in Rings().Commutative():
-            from sage.algebras.free_algebra import FreeAlgebra
-            self._internal_poly_ring = FreeAlgebra(self._laurent_poly_ring, 1, "DUMMY_VARIABLE")
-        else:
-            self._internal_poly_ring = PolynomialRing(self._laurent_poly_ring, "DUMMY_VARIABLE", sparse=sparse)
+        self._internal_poly_ring = PolynomialRing(self._laurent_poly_ring, "DUMMY_VARIABLE", sparse=sparse)
 
     def _repr_(self):
         """


mantepse:
There might be further issues when using the completion of the `FreeAlgebra`, because the arity is not detected correctly.  The test we currently use is
```
        if basis in Algebras.TensorProducts:
            self._arity = len(basis._sets)
        else:
            if basis not in GradedAlgebrasWithBasis:
                raise ValueError("basis should be in GradedAlgebrasWithBasis")
            self._arity = 1
```
Therefore, we have
```
            sage: R = algebras.Free(QQ, ('a', 'b'), degrees=(1, 2))
            sage: L = R.completion()
            sage: L._arity
            1
```
Here is a concrete bug that entails:
```
            sage: e = L(R.gen(0))
            sage: e, e.valuation(), e.shift(1), e.shift(1).valuation()
            (a, 1, a, 2)
```
(The method `shift` should actually raise an error.)

I think, however, that this bug is less problematic than not having the completion of the free algebra at all.

tscrim:
Arity doesn’t make much sense for general graded algebras. It makes some sense for the free algebra, as we can consider that a tensor product of univariate free algebras (from the functorial properties of tensor algebras). We might want to just special case the free algebra in most (lazy) completions since it has extra properties.

mantepse:
> Arity doesn’t make much sense for general graded algebras.  It makes some sense for the free algebra, as we can consider that a tensor product of univariate free algebras (from the functorial properties of tensor algebras). We might want to just special case the free algebra in most (lazy) completions since it has extra properties.

I just checked the `LazyCompletionGradedAlgebraElement` class - this has only the inherited methods from `LazyCauchyProductSeries`, and the only method that check `_arity` is `shift`.  Other than that, `_arity` is used in the `_element_constructor`.

I guess that there are some methods that would use `_arity` and could be applied to the completion of the free algebra.  For example, composition (from `LazyPowerSeries`) should just work.

Maybe it would be best to make arity a mandatory argument to `LazyCompletionGradedAlgebra.__init__`?

I just realised that the `shift` bug I mentioned above is also present for `LazySymmetricFunction`.

I would suggest not to worry too much about this right now.  I am sure we can find a solution once some feature is requested.  If you want to, I can turn this discussion into a new issue, so it doesn't get lost.

tscrim (approved):
LGTM. Thanks.

tscrim:
It probably would be good to make a separate issue for potentially changing the completion of the free algebra into its own special case (I think there will be a number of things that would benefit from this) and one for the arity issues.

Perhaps it would be useful for the arity to be an optional `__init__` argument, so each parent could more easily handle this (with possibly the user giving it too). However, I am not entirely sure about the best way to handle this.

mantepse:
I think we are using `_arity` for two things which are only superficially related, and I guess we should simply unentangle that, perhaps as follows:
* `_arity`, for any code related to composition of series.  In particular, we could very likely reuse `PowerSeries.__call__` for the completion of the `FreeAlgebra`.
* `_shiftable` (or something better), a boolean that tells us whether a coefficient in the stream knows its degree or doesn't.

Perhaps this also has something to do with `_internal_poly_ring` and `_laurent_poly_ring`, but I haven't thought about it.

In any case, other ticket.

PR #37634, opened 2024-03-19
The Glaisher-Franklin bijections on integer partitions

mantepse:
The Glaisher-Franklin bijections, for a positive integer $s$, map the set of parts divisible by $s$ to the set of parts which occur at least $s$ times, see https://www.findstat.org/MapsDatabase/Mp00312 for the $s=2$ case.

This generalizes and streamlines the code from findstat.

### DIFF
--- a/src/sage/combinat/partition.py
+++ b/src/sage/combinat/partition.py
@@ -2485,6 +2485,95 @@ class Partition(CombinatorialElement):
         par = Partitions_n(sum(self))
         return par.element_class(par, conjugate(self))
 
+    def franklin_glaisher(self, s):
+        r"""
+        Apply the Franklin-Glaisher bijection to ``self``.
+
+        INPUT:
+
+        - ``s`` -- positive integer
+
+        OUTPUT:
+
+        The Franklin-Glaisher bijection, with parameter `s`, returns
+        a partition whose set of parts that are repeated at least `s`
+        times equals the set of parts divisible by `s` in ``self``,
+        after dividing each part by `s`.
+
+        EXAMPLES::
+
+            sage: Partition([4, 3, 2, 2, 1]).franklin_glaisher(2)
+            [3, 2, 2, 1, 1, 1, 1, 1]
+
+        TESTS::
+
+            sage: d = lambda la, s: set(p / s for p in la if p % s == 0)
+            sage: r = lambda la, s: set(p for p in la if list(la).count(p) >= s)
+            sage: all(d(mu, s) == r(mu.franklin_glaisher(s), s)
+            ....:     for n in range(20) for mu in Partitions(n)
+            ....:     for s in range(1, 5))
+            True
+        """
+        mu = []
+        for p, m in enumerate(self.to_exp(), 1):
+            if not p % s:
+                mu.extend([p//s]*(m*s))
+            else:
+                mu.extend(p * v * s**i for i, v in enumerate(m.digits(s)) if v)
+
+        P = self.parent()
+        return P.element_class(P, sorted(mu, reverse=True))
+
+    def franklin_glaisher_inverse(self, s):
+        r"""
+        Apply the inverse of the Franklin-Glaisher bijection to ``self``.
+
+        INPUT:
+
+        - ``s`` -- positive integer
+
+        OUTPUT:
+
+        The inverse of the Franklin-Glaisher bijection, with
+        parameter `s`, returns a partition whose set of parts that
+        are divisible by `s`, after dividing each by `s`, equals the
+        equals the set of parts repeated at least `s` times in
+        ``self``.
+
+        EXAMPLES::
+
+            sage: Partition([4, 3, 2, 2, 1]).franklin_glaisher(2)
+            [3, 2, 2, 1, 1, 1, 1, 1]
+            sage: Partition([3, 2, 2, 1, 1, 1, 1, 1]).franklin_glaisher_inverse(2)
+            [4, 3, 2, 2, 1]
+
+        TESTS::
+
+            sage: d = lambda la, s: set(p / s for p in la if p % s == 0)
+            sage: r = lambda la, s: set(p for p in la if list(la).count(p) >= s)
+            sage: all(r(mu, s) == d(mu.franklin_glaisher_inverse(s), s)
+            ....:     for n in range(20) for mu in Partitions(n)
+            ....:     for s in range(1, 5))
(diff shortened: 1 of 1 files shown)


src/sage/combinat/partition.py
```diff
@@ -2485,6 +2485,122 @@ def conjugate(self):
...
+            [4, 3, 2, 2, 1]
+
+        TESTS:
+
+        The map is inverse to :meth:`glaisher_franklin`::
+
+            sage: all(mu.glaisher_franklin(s).glaisher_franklin_inverse(s) == mu
+            ....:     for n in range(20) for mu in Partitions(n)
+            ....:     for s in range(1, 5))
+            True
```
```suggestion
            sage: all(mu.glaisher_franklin(s).glaisher_franklin_inverse(s) == mu
            ....:     mu.glaisher_franklin_inverse(s).glaisher_franklin(s) == mu
            ....:     for n in range(20) for mu in Partitions(n)
            ....:     for s in range(1, 5))
            True
```
I like having both compositions to be safe. Of course, the test in the GF method checking the size of the output makes this redundant. However, the explicitness of this has some value IMO.

src/sage/combinat/partition.py
```diff
@@ -2485,6 +2485,122 @@ def conjugate(self):
...
+        """
+        s = ZZ(s)
+        if s.is_one():
+            return self
+        mu = []
+        for p, m in enumerate(self.to_exp(), 1):
+            p = ZZ(p)
+            mu.extend([p * s]*(m // s))
+            m1, p1 = p.val_unit(s)
+            mu.extend([p1]*((m % s) * s**m1))
```
```suggestion
            mu.extend([p * s] * (m // s))
            m1, p1 = p.val_unit(s)
            mu.extend([p1] * ((m % s) * s**m1))
```
PEP8 things.

src/sage/combinat/partition.py
```diff
@@ -2485,6 +2485,122 @@ def conjugate(self):
...
+        if s.is_one():
+            return self
+        mu = []
+        for p, m in enumerate(self.to_exp(), 1):
+            if not p % s:
+                mu.extend([p // s]*(m*s))
+            else:
+                mu.extend(p1 for i, v in enumerate(m.digits(s))
+                          if (p1 := p * s**i)
+                          for _ in range(v))
```
```suggestion
                mu.extend([p // s] * (m*s))
            else:
                mu.extend([p1]*v for i, v in enumerate(m.digits(s))
                          if (p1 := p * s**i))
```
The first is PEP8, the second is because it is faster:
```
sage: def test1():
....:     L = [1,2,3]
....:     L.extend([1] * 100)
....:
sage: def test2():
....:     L = [1,2,3]
....:     L.extend(1 for _ in range(100))
....:
sage: %timeit test1()
1.11 µs ± 71.7 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
sage: %timeit test2()
9.21 µs ± 652 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

tscrim (changes requested):
Some minor specific things. The other thing is I think the OUTPUT block is better as a description block. The output (in the programing sense) is a partition, which doesn't need to be stated. Otherwise LGTM.

src/sage/combinat/partition.py
mantepse: Wow, I missed that one!  Thank you!

tscrim (approved):
Thanks. LGTM.

vbraun:
Please don't set tickets to positive review with obvious CI failures

src/sage/combinat/partition.py
```diff
@@ -2485,6 +2485,118 @@ def conjugate(self):
...
+            [3, 2, 2, 1, 1, 1, 1, 1]
+            sage: Partition([3, 2, 2, 1, 1, 1, 1, 1]).glaisher_franklin_inverse(2)
+            [4, 3, 2, 2, 1]
+
+        TESTS:
+
+        The map is inverse to :meth:`glaisher_franklin`::
+
+            sage: all(mu.glaisher_franklin(s).glaisher_franklin_inverse(s) == mu
+            ....:     mu.glaisher_franklin_inverse(s).glaisher_franklin(s) == mu
```
```suggestion
            ....:     and mu.glaisher_franklin_inverse(s).glaisher_franklin(s) == mu
```

src/sage/combinat/partition.py
```diff
@@ -2485,6 +2485,118 @@ def conjugate(self):
...
+        s = ZZ(s)
+        if s.is_one():
+            return self
+        mu = []
+        for p, m in enumerate(self.to_exp(), 1):
+            if not p % s:
+                mu.extend([p // s] * (m*s))
+            else:
+                mu.extend([p1]*v for i, v in enumerate(m.digits(s))
+                          if (p1 := p * s**i))
```
```suggestion
                mu.extend([p1]*v for i, v in enumerate(m.digits(s))
                          if (p1 := p * s**i))
```
This doesn't quite work here. Sorry for that. How you had it before is necessary as it extends by adding the lists over the `for` loop. Also, isn't `p1` always nonzero since $p, s \in \mathbb{Z}_{>0}$?

tscrim (changes requested):
@vbraun Sorry, this was passing before and these were supposed to be innocuous changes.

src/sage/combinat/partition.py
mantepse: Yes, `p1` is nonzero, the `if` is only to define it, because it is the same for all `v`, and previously I was looping over `v`.

mantepse: I guess
```
                for i, v in enumerate(m.digits(s)):
                    mu.extend([p * s**i]*v)
```
is better, right?

tscrim: Possibly. A quick timing check should be done. Although it is a mirco-optimization at the end of the day, so I might side with the more readable one here.

For `p1`, what I am saying is it will *never* be nonzero, no matter what `p` and `s` are as it is a product of positive numbers.

mantepse: Yes, 'p1' is nonzero, why are you asking?

tscrim: But by its formula, it is always nonzero, so the `if` check is vacuous here.

mantepse: The if check was only to be able to define 'p1'.

tscrim: I see. Then yes, the `for` loop with the `mu.extend` is better.

tscrim (approved):
LGTM.

PR #37636, opened 2024-03-19
remove unused deps for sagelib

dimpase:
This cleans up a bit the list of deps for sagelibs. Namely, after we got cypari2, pplpy, and primecountpy, pari, ppl,  primesieve, and primecount stopped being used in sagelib itself  

### :memo: Checklist


dimpase:
@orlitzky

orlitzky:
For sage-the-distro this is probably fine since all that `dependencies` file does is change the order or installation, but I notice that e.g. libpari is still linked to some of my sage bits:

```
$ ldd local/var/lib/sage/venv-python3.11/lib/python3.11/site-packages/sage/libs/pari/convert_flint.cpython-311-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007ffd5cef3000)
	libpari-gmp-tls.so.8 => /usr/lib64/libpari-gmp-tls.so.8 (0x00007fb0b57dc000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fb0b5633000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb0b6289000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fb0b5588000)
	libgmp.so.10 => /usr/lib64/libgmp.so.10 (0x00007fb0b5504000)
```

I'm not sure yet why that is. It could be that cython isn't respecting the `--as-needed` in my `LDFLAGS`, or it could be that cypari is somehow exposing internals of libpari and we also (somehow) wind up with `-lpari` in our link commands.

orlitzky:
Maybe something I don't understand about cypari2 then. The other two (pplpy and primecountpy) don't cause libppl or libprimecount to be linked in to sage.

I guess what I really want to know (for this PR) is: if pari is installed somewhere where sage can't find it, and then cypari2 is installed by pointing it to that copy of pari, will sagelib still be able to build without knowing where pari is?

dimpase:
pplpy and primecountpy are not invoked via cimports, but cypari2 is.
So the former are fully dynamic (no build-time linking), the latter is not.

As to how exactly the linking of cimported cypari2 modules is happening. I don't know. It does not seem to be documented.

orlitzky:
sagelib is linking to pari itself, it seems: https://github.com/sagemath/sage/blob/develop/src/sage/misc/cython.py#L53

Ideally we should be linking to cypari2 instead (my cypari2 binaries themselves link to pari), but I don't know if cython supports something like that. Maybe better to skip pari for now and remove the other dependencies?

dimpase:
> sagelib is linking to pari itself, it seems: https://github.com/sagemath/sage/blob/develop/src/sage/misc/cython.py#L53
> 

I tried removing 'pari' there and rebuiding sagelib, it has no effect - the linkage as in
https://github.com/sagemath/sage/pull/37636#issuecomment-2009400723
stays the same, and everything still works

dimpase:
some CI fails due to broken giac, it's not relevant for this PR

orlitzky:
> I tried removing 'pari' there and rebuiding sagelib, it has no effect - the linkage as in [#37636 (comment)](https://github.com/sagemath/sage/pull/37636#issuecomment-2009400723) stays the same, and everything still works

I confirmed this but I still don't know where the pari linkage is coming from. If sage needs to be linked to pari and if it's not happening via cypari2, then sage should still depend on pari. It's probably not a deep mystery I just haven't had much time this week.

mkoeppe:
`sage.misc.cython` is only for *runtime* use of Cython by users (and by some doctests) and does not affect the build of Sage at all.

Removing `pari` there thus only affects user code. The list of "standard libraries" that is used there is pretty arbitrary anyway, so I don't think there would be a major compatibility concern to removing the `-lpari`.

mkoeppe:
However, linking any extension module that `cimport`s from cypari2 will explicitly link through with `-lpari` -- because the `cypari2` pxd files contain the directive `# distutils: libraries = gmp pari`.

Therefore, you cannot just remove `pari` from the dependencies of sagelib here. In particular, on distributions with -devel subpackages, you will need the .so file from the -devel subpackage.

tornaria:
> However, linking any extension module that `cimport`s from cypari2 will explicitly link through with `-lpari` -- because the `cypari2` pxd files contain the directive `# distutils: libraries = gmp pari`.
> 
> Therefore, you cannot just remove `pari` from the dependencies of sagelib here. In particular, on distributions with -devel subpackages, you will need the .so file from the -devel subpackage.

Would adding a trivial `paridecl.pyx` and `paripriv.pyx` into cypari2 allow us to get the link to the matching `-lgmp -lpari` via `paridecl.so` and `paripriv.so` and avoid the need for the distutils line?

In any case a cython file that `cimport`s from cypari2 will need access to the pari headers, but these have to match as well. Maybe the simplest solution is to include the headers in cypari2 (either the actual files or a symlink). The binary wheels for cypari include the libraries anyway as they should to match. If the hack I suggested above is not good, the `libgmp.so` and `libpari.so` symlinks could be included as well.

Issue #37417, opened 2024-02-21
Quaternion `maximal_order` fails 


grhkm21:
### Steps To Reproduce

```sage
B = QuaternionAlgebra(-4, -28)
B.maximal_order()
```

### Expected Behavior

work

### Actual Behavior

not

### Additional Information

Not sure if this is normal, as I don't have enough quaternion knowledge to determine.

@yyyyx4 

### Environment

```markdown
N/A
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

grhkm21:
Sorry, I missed #37217 .

PR #37239, opened 2024-02-04
Fix wrong answer given by `normalize_basis_at_p` function of quaternion_algebra

RuchitJagodara:
This patch fixes #37217. I have also added a doctest for this issue and 
removed an wrong example.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.

### :hourglass: Dependencies

Not applicable.

PR #37644, opened 2024-03-21
Corrections in `.normalize_basis_at_p` and `.maximal_order()` of `quaternion_algebra.py`

### LINKED ISSUE #37217: Quaternion algebra: bugs in `maximal_order` and `normalize_basis_at_p`
### Steps To Reproduce

1. On any sage terminal run: 
```
a,b = -292, -732
A = QuaternionAlgebra(QQ,a,b)
A.maximal_order()
```

2. On any sage terminal run: 
```
from sage.algebras.quatalg.quaternion_algebra import normalize_basis_at_p
A.<i,j,k> = QuaternionAlgebra(-1,-7)
e = [A(1), k, j, 1/2 + 1/2*i + 1/2*j + 1/2*k]
e_norm = normalize_basis_at_p(e, 2)
V = QQ**4
V.span([V(x.coefficient_tuple()) for (x,_) in e_norm]).dimension()
```

### Expected Behavior

1. I expect the code to return a maximal order of the quaternion algebra. 
2. I expect a basis to be of rank 4

### Actual Behavior

1. It fails. 
2. Return a list of rank 3

### Additional Information

The error can be traced to a supposedly bug for the function `normalize_basis_at_p` when $p = 2$. In fact in the faulty execution it returns a supposed basis of rank $3$ instead of $4$. 

Trying to debug the function I discovered that actually in some situations it return a list of $4$ vectors with the last two being equal, clearly not a basis! This strange behaviour can be noted also in the examples inserted for the function:
```
sage: A.<i,j,k> = QuaternionAlgebra(-1,-7)
sage: e = [A(1), k, j, 1/2 + 1/2*i + 1/2*j + 1/2*k]
sage: normalize_basis_at_p(e, 2)
[(1, 0), (1/2 + 1/2*i + 1/2*j + 1/2*k, 0), (-34/105*i - 463/735*j + 71/105*k, 1),
         (-34/105*i - 463/735*j + 71/105*k, 1)]
```

I already tried to solve the issue myself so I insert here some possibly ***useful notes:***
- it is easy to find more errors (just take random $a,b$ until maximal order crashes), but here 20 of them:
`[(-292, -732), (-48, -564), (-436, -768), (-752, -708), (885, 545), (411, -710), (-411, 593), (805, -591), (-921, 353), (409, 96), (394, 873), (353, -722), (730, 830), (-466, -427), (-213, -630), (-511, 608), (493, 880), (105, -709), (-213, 530), (97, 745)]`,
- the two functions are *Algorithms 7.10, 7.9 and 3.12* from https://arxiv.org/abs/1004.0994 
- `maximal_order` to me seem to be implemented correctly and I believe the error is only traceable to `normalize_basis_at_p` (that in fact contains some differences from the original algorithm)
- trying to remove the previous differences I noted creates several crashes for even more quaternion algebras, so I suppose they where necessaries. 

I'll still try to solve the issue, so add here eventual updates.

### Environment

```markdown
- **OS**: MacOS
- **Sage Version**: Version: 9.2
```

### Checklist

S17A05:
- Corrected in `.normalize_basis_at_p` the wrong assignments of `f0, f1`: The idea is to replace `f0` by `f0 + f1` and `f1` by `f0` _simultaneously_ - the original code, however, first replaced `f0` by `f0 + f1` and then copied the value to `f1`, hence causing a decrease in dimension
- Corrected an index error just before the recursive call of `normalize_basis_at_p` in the off-diagnonal case for `p = 2` (the `else`-block)
- In `.maximal_order()` the intermediate basis `e_n` might not define an order anymore (as seen in the comments, this was known to the author(s) of the method) - therefore we need to manually compute the discriminant in the loop instead of relying on the `.discriminant()`-method of quaternion orders
- Adapted an example in `.maximal_order()` to use the new method `.is_maximal()`

Fixes #37217 (both parts) and #37417.

Disclaimer:
I am aware of #37239, but since progress on that PR seems to have halted, I took the liberty to fix the issues myself :)


src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -766,9 +765,24 @@ def maximal_order(self, take_shortcuts=True):
...
+            ....:           (-511, 608), (493, 880), (105, -709), (-213, 530),
+            ....:           (97, 745)]
+            sage: all(QuaternionAlgebra(a, b).maximal_order().is_maximal()
+            ....:           for (a, b) in invars)
+            True
         """
         if self.base_ring() != QQ:
-            raise NotImplementedError("maximal order only implemented for rational quaternion algebras")
+            raise NotImplementedError("maximal order only implemented for"
+                                        " rational quaternion algebras")
```
mkoeppe:
```suggestion
            raise NotImplementedError("maximal order only implemented for "
                                      "rational quaternion algebras")
```

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -892,13 +908,23 @@ def maximal_order(self, take_shortcuts=True):
...
                 e = e_n
 
+                # Since e might not define an order at this point, we need to
+                # manually calculate the updated discriminant
+                L = []
+                for x in e:
+                    MM = []
+                    for y in e:
+                        MM.append(x.pair(y))
+                    L.append(MM)
```
mkoeppe: This could be rewritten as list comprehensions

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -892,13 +908,23 @@ def maximal_order(self, take_shortcuts=True):
...
+                disc = (MatrixSpace(QQ, 4, 4)(L)).determinant().sqrt()
```
mkoeppe:
```suggestion
                disc = matrix(QQ, 4, 4, L).determinant().sqrt()
```

mkoeppe (approved):
LGTM

PR #37645, opened 2024-03-22
Fix a couple of doctests that trigger when building from sdist

tornaria:
Some failures I found when building and running as in https://github.com/sagemath/sage/pull/37138#issuecomment-2014038968

The first one is a missing `# needs tdlib` which is not noticed when building a wheel from the github source instead of via sdist, since `tdlib.pyx` is included in the former but not in the latter.

The second one triggers when one installs the wheel in a random directory (not site-packages, not a venv) and runs via PYTHONPATH. Then module names like `sage.doctest.tests.tolerance` are instead just `tolerance` so there's a mismatch. Very easy to fix writing `...tolerance` instead.

### :memo: Checklist

### DIFF
--- a/src/sage/doctest/test.py
+++ b/src/sage/doctest/test.py
@@ -54,7 +54,7 @@ Check handling of tolerances::
     Doctesting 1 file.
     sage -t --warn-long 0.0 --random-seed=0 tolerance.rst
     **********************************************************************
-    File "tolerance.rst", line ..., in sage.doctest.tests.tolerance
+    File "tolerance.rst", line ..., in ...tolerance
     Failed example:
         print(":-(")    # abs tol 0.1
     Expected:
@@ -62,7 +62,7 @@ Check handling of tolerances::
     Got:
         :-(
     **********************************************************************
-    File "tolerance.rst", line ..., in sage.doctest.tests.tolerance
+    File "tolerance.rst", line ..., in ...tolerance
     Failed example:
         print("1.0 2.0 3.0")  # abs tol 0.1
     Expected:
@@ -70,7 +70,7 @@ Check handling of tolerances::
     Got:
         1.0 2.0 3.0
     **********************************************************************
-    File "tolerance.rst", line ..., in sage.doctest.tests.tolerance
+    File "tolerance.rst", line ..., in ...tolerance
     Failed example:
         print("Hello")  # abs tol 0.1
     Expected:
@@ -78,7 +78,7 @@ Check handling of tolerances::
     Got:
         Hello
     **********************************************************************
-    File "tolerance.rst", line ..., in sage.doctest.tests.tolerance
+    File "tolerance.rst", line ..., in ...tolerance
     Failed example:
         print("1.0")  # abs tol 0.1
     Expected:
@@ -86,7 +86,7 @@ Check handling of tolerances::
     Got:
         1.0
     **********************************************************************
-    File "tolerance.rst", line ..., in sage.doctest.tests.tolerance
+    File "tolerance.rst", line ..., in ...tolerance
     Failed example:
         print("Hello 1.1")  # abs tol 0.1
     Expected:
@@ -94,7 +94,7 @@ Check handling of tolerances::
     Got:
         Hello 1.1
     **********************************************************************
-    File "tolerance.rst", line ..., in sage.doctest.tests.tolerance
+    File "tolerance.rst", line ..., in ...tolerance
     Failed example:
         print("Hello 1.0")  # rel tol 1e-6
     Expected:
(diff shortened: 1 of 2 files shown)


mkoeppe:
> Then module names like `sage.doctest.tests.tolerance` are instead just `tolerance` so there's a mismatch. Very easy to fix writing `...tolerance` instead.

I think we can find a better fix for this, let me take a look. This may be the same issue as with all the PEP-420 issues -- finding the root is not well-defined. (But it's well-defined in the Sage layout because of our all*.py files.)

mkoeppe:
Indeed this is happening in `sage.doctest.sources.get_basename`, which is one of the last holdouts of our use of `SAGE_SRC` and `SAGE_LIB`.

tornaria:
> `age.doctest.sources.get_basename`,

No, it's not fixed by that.

Is my fix here unreasonable?

mkoeppe:
> Is my fix here unreasonable?

If the basename was used only for printing these messages, I would probably not have bothered to look for a proper fix.
But it's

tornaria:
> > Is my fix here unreasonable?
> 
> If the basename was used only for printing these messages, I would probably not have bothered to look for a proper fix. But it's also used as the identifier for the "stats", and thus for the "baseline_stats" mechanism.

But this is a "second level" test. We doctest `sage.doctest.test` whose identifier is ok, some doctest in that file changes directory to `.../sage/doctest/tests/` and runs `sage -t tolerance.rst` in there (it passes the filename without any directory) and it's this run of `sage -t tolerance.rst` which fails because the identifier is printed incorrectly.

I don't think this "second level" is relevant for the "stats", or is it?

mkoeppe:
> some doctest in that file changes directory to `.../sage/doctest/tests/` and runs `sage -t tolerance.rst` in there (it passes the filename without any directory) and it's this run of `sage -t tolerance.rst` which fails because the identifier is printed incorrectly.

You are describing it correctly; except it's not just "printed" incorrectly. There's no such code that abbreviates identifiers further somehow, as far as I know.

mkoeppe:
I agree that in this situation (running `sage -t tolerance.rst`), just the plain filename would conceivably also be the correct identifier -- but it really should not depend on the SAGE_LIB business. That's what #37647 attempts to fix. 

I also agree that none of this is very important, and that nothing would be wrong with fixing it the way that you do here. It's just that I cannot resist opportunities for synergetic fixes.

tornaria:
> I agree that in this situation (running `sage -t tolerance.rst`), just the plain filename would conceivably also be the correct identifier -- but it really should not depend on the SAGE_LIB business. That's what #37647 attempts to fix.
> 
> I also agree that none of this is very important, and that nothing would be wrong with fixing it the way that you do here. It's just that I cannot resist opportunities for synergetic fixes.

Makes sense, but since I wasn't able to reproduce it using `get_basename(..)` we are kind of blind on that. Sorry about that.

mkoeppe:
No worries, I'll test it later today or tomorrow.

tornaria:
Ping @mkoeppe 

I'm switching our package to build from pypi sdist, and I was hit by this:
```
sage -t --warn-long 4.0 --random-seed=306581033401793295942652738315483415678 /usr/lib/python3.12/site-packages/sage/misc/package_dir.py
File "/usr/lib/python3.12/site-packages/sage/misc/package_dir.py", line 116, in sage.misc.package_dir.read_distribution
Failed example:
    read_distribution(os.path.join(SAGE_SRC, 'sage', 'graphs', 'graph_decompositions', 'tdlib.pyx'))
Exception raised:
      File "/usr/lib/python3.12/site-packages/sage/doctest/forker.py", line 714, in _run
      File "/usr/lib/python3.12/site-packages/sage/doctest/forker.py", line 1146, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.misc.package_dir.read_distribution[2]>", line 1, in <module>
        read_distribution(os.path.join(SAGE_SRC, 'sage', 'graphs', 'graph_decompositions', 'tdlib.pyx'))
      File "/usr/lib/python3.12/site-packages/sage/misc/package_dir.py", line 121, in read_distribution
        with open(src_file, encoding='utf-8', errors='ignore') as fh:
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/python3.12/site-packages/sage/graphs/graph_decompositions/tdlib.pyx'
1 item had failures:
   1 of   5 in sage.misc.package_dir.read_distribution
    [35 tests, 1 failure, 0.02 s]
```

kiwifb:
I had not noticed that the doctest error about tdblib.pyx disappeared when I started to apply the line `graft sage` to fix the great sdist debacle for 10.4.beta3. It does cause this file to be shipped which is an issue in and off itself but as feedback to another issue.

tornaria:
For the record, I switched void-linux package to pypi sdist and it's working fine with this PR as seen in https://github.com/void-linux/void-packages/pull/49571

kiwifb:
The `# needs tdlib` stuff does work and I am happy with that bit. I do not see the other test failures in `sage/doctest/test.py` which means that for me that part makes no difference.

PR #37650, opened 2024-03-22
src/sage/features/sagemath.py: Add feature SAGE_SRC

mkoeppe:
This feature is for conditionalizing some doctests.

Fixes doctest failure reported in https://github.com/sagemath/sage/pull/37645.

### :memo: Checklist

### DIFF
--- a/src/sage/misc/package_dir.py
+++ b/src/sage/misc/package_dir.py
@@ -111,6 +111,7 @@ def read_distribution(src_file):
 
     EXAMPLES::
 
+        sage: # needs SAGE_SRC
         sage: from sage.env import SAGE_SRC
         sage: from sage.misc.package_dir import read_distribution
         sage: read_distribution(os.path.join(SAGE_SRC, 'sage', 'graphs', 'graph_decompositions', 'tdlib.pyx'))
--- a/src/sage/features/sagemath.py
+++ b/src/sage/features/sagemath.py
@@ -43,6 +43,26 @@ from . import PythonModule, StaticFile
 from .join_feature import JoinFeature
 
 
+class SAGE_SRC(StaticFile):
+    r"""
+    A :class:`~sage.features.Feature` which describes the presence of the
+    monolithic source tree of the Sage library.
+
+    """
+    def __init__(self):
+        r"""
+        TESTS::
+
+            sage: from sage.features.sagemath import SAGE_SRC
+            sage: isinstance(SAGE_SRC(), SAGE_SRC)
+            True
+        """
+        from sage.env import SAGE_SRC
+        StaticFile.__init__(self, 'SAGE_SRC',
+                            filename='VERSION.txt',  # a file that is never installed
+                            search_path=(SAGE_SRC,) if SAGE_SRC else ())
+
+
 class sagemath_doc_html(StaticFile):
     r"""
     A :class:`~sage.features.Feature` which describes the presence of the documentation
@@ -1095,7 +1115,8 @@ def all_features():
         sage: list(all_features())
         [...Feature('sage.combinat'), ...]
     """
-    return [sagemath_doc_html(),
+    return [SAGE_SRC(),
+            sagemath_doc_html(),
             sage__combinat(),
             sage__geometry__polyhedron(),
             sage__graphs(),
--- a/src/sage/features/sagemath.py
+++ b/src/sage/features/sagemath.py
@@ -58,8 +58,13 @@ class SAGE_SRC(StaticFile):
             True
         """
         from sage.env import SAGE_SRC
+        # We check the file bin/sage-src-env-config.in, which by design is:
+        # - never installed,
+        # - not included in the sagemath-standard sdist,
+        # - included only in one modularized sdist, of pkgs/sage-conf_pypi,
+        #   where it appears in a subdirectory (sage_root/src/bin/)
         StaticFile.__init__(self, 'SAGE_SRC',
-                            filename='VERSION.txt',  # a file that is never installed
+                            filename='bin/sage-src-env-config.in',
                             search_path=(SAGE_SRC,) if SAGE_SRC else ())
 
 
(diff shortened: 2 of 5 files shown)


kiwifb:
That's an interesting solution, using features more is a good idea. Here `SAGE_SRC` is `/usr/lib/python3.12/site-packages` and I certainly hope no one put a file 'VERSION.txt` at that level of the tree.

It is cool that it addresses the failures in `package_dir.py` and ` sage_setup/find.py` but I have no issues with the the other two files. Nevertheless, it is probably a good idea, future change may cause breakage.

kiwifb (approved):
LGTM

mkoeppe:
> Here `SAGE_SRC` is `/usr/lib/python3.12/site-packages`

Yes, this feature test does take care of this situation, but at some point we will probably also want to get rid of the SAGE_SRC fallback to SAGE_LIB.

kiwifb:
That'd be great.

mkoeppe:
That's now:
- #37666

mkoeppe:
Setting to "positive review" based on @kiwifb's review.

src/sage/misc/package_dir.py
```diff
@@ -111,6 +111,7 @@ def read_distribution(src_file):
 
     EXAMPLES::
 
+        sage: # needs SAGE_SRC
         sage: from sage.env import SAGE_SRC
         sage: from sage.misc.package_dir import read_distribution
         sage: read_distribution(os.path.join(SAGE_SRC, 'sage', 'graphs', 'graph_decompositions', 'tdlib.pyx'))
```
tornaria, line 117: This line `# needs tdlib` as explained in #37645. You act as if `SAGE_SRC` indicates that you have the whole source tree, but in fact `sagemath-standard` sdist contains `VERSION.txt` afaict.

kiwifb: A fair point that I had not thought about. I was only thinking of the case where you test with an installation of sage that is not in source. In which case `VERSION.txt` is not installed and cannot be found.

mkoeppe: > in fact `sagemath-standard` sdist contains `VERSION.txt` afaict.

Very good point. And in fact all of the modularized distributions have VERSION.txt too. 
Suggestions what we should test?

kiwifb: Presence of the `bin` or `doc` folders? It also has to be something that will be stable - for some time at least.

kiwifb: Producing a sdist right now to make sure I do not say something stupid.

kiwifb: OK, `bin` is in the sdist. But we could go for any `.m4` files.

mkoeppe: I think `src/bin/sage-src-env-config.in` may be a good candidate

kiwifb: Indeed, it would be adequate.

tornaria: Why not add a sentinel file in `src/sage/features/` that is never included in any distribution? Using an arbitrary existing file is just asking for trouble when later that file is changed for an unrelated reason, etc.

OTOH, it feels weird that testing sage-setup needs SAGE_SRC. The `read_distribution()` function seems like a part of `sage-setup` not of sagelib, and why can't it be unit tested instead of testing with a `.pyx` from the sage source?

kiwifb: I totally get what you mean about the test, but it is meant to test "distribution", so stuff like sagemath-bliss or as in this case sagemath-tdlib. If you want to keep things in source tree, you have to provide the test suite with a "fake" mini distribution to test.

As for the file to test, it is arbitrary and I would have probably chosen `pyproject.tml.m4` myself. But I think the file chosen is there for a long time which is a criterion.

tornaria: What is wrong with a file `src/sage/features/this-file-only-in-the-source.txt` which IIUC it's never included in a sdist by default (https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html)

The file can have a simple explanation of its purpose, being a text file.

tornaria:
> [Documentation preview for this PR](https://deploy-preview-37650--sagemath.netlify.app/html/en) (built with commit [9a761f7](https://github.com/sagemath/sage/commit/9a761f7f98c37ad0ecd76fd3504d42e0414cde53); [changes](https://deploy-preview-37650--sagemath.netlify.app/CHANGES.html)) is ready! 🎉

Please stop the spam bot: #37740

src/sage/misc/package_dir.py
mkoeppe: > Using an arbitrary existing file is just asking for trouble when later that file is changed for an unrelated reason, etc.

Well, this particular file I chose is not arbitrary but by its design has exactly the right properties.

mkoeppe:
@tornaria Please stop with these repeated off-topic comments.

src/sage/misc/package_dir.py
mkoeppe: > The `read_distribution()` function seems like a part of `sage-setup` not of sagelib

No, it's in the right place. This module, like `sage.features`, is part of sagemath-environment.

mkoeppe:
Let's get this in please

kiwifb:
I was reminded of this because of #37645 which address only the `tdlib.pyx` tests occurring in sagemath-standard without covering the ones in sage_setup. I confirm my positive review now that the one contentious point has been solved.

tornaria:
image

I'm not arguing that my comment was not offtopic. It was.

But there's no record of who marked it as offtopic (it wasn't me); I don't like that someone can anonymously hide comments. FWIW, I *do NOT* have this ability. It seems can mark *my own* comments as offtopic, but not others (not even in my own PRs).

It's the same asymmetry where some developers can modify PRs they don't own, but others don't have this ability (I don't).

mkoeppe:
Thanks, Francois.

PR #37651, opened 2024-03-22
use parent in asymptotic ring

fchapoton:
trying to use the modern `Parent` and categories in the asymptotic ring, instead of the auld `Algebra`

### :memo: Checklist


mezzarobba (approved):
lgtm, thanks!

PR #35404, opened 2023-03-31
Drop support for Python 3.8

tobiasdiez:
### :books: Description

In preparation for important package upgrades, we remove support for Python v3.8. (Almost) all occurrences of "Python 3.8" are removed from the codebase. There are still some comments left that essentially say that in newer versions things can be simplified, but I didn't touch those and leave them for follow-up PRs.

This will help with:
-  PRs that require Python 3.9 or above, such as https://github.com/sagemath/sage/pull/34973 and https://github.com/sagemath/sage/pull/35188. 
- Upgrades of Python packages that have dropped support for Python 3.8 because they follow NEP 29 and have already released a new major version that drops support according to the NEP 29 schedule.
  - NetworkX 3.2 (expected 2023-??; version 3.1 was released 2023-04, see https://github.com/sagemath/sage/pull/35671)
  - #34816
  - #35703
  - See also #32074.

Test run: https://github.com/tobiasdiez/sage/actions/runs/4586981626

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

PR #35203, opened 2023-02-25
sagemath-standard: Add dependencies `typing_extensions`, `importlib_resources`, `importlib_metadata`

mkoeppe:
### 📚 Description

Resolves #34831 

### 📝 Checklist

- [ ] I have made sure that the title is self-explanatory and the description concisely explains the PR.
- [x] I have linked an issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37654, opened 2024-03-22
Developer guide: Improve discussion of backport packages

mkoeppe:
- #35203 added these packages as unconditional dependencies
- prompted by a sporadic use of typing_extensions in the Sage library (see #34831)
- motivated further by the demand to immediately drop support for Python 3.8 so that newer typing features can be used in the Sage library (https://github.com/sagemath/sage/pull/35404)
- https://github.com/sagemath/sage/pull/36776 reduced the packages to conditional dependencies

Here we improve the documentation in the section "Language standard" of the developer's guide so that it aligns with how the conditional dependencies are declared.

Based on changes split out from #37399.

### :memo: Checklist

### DIFF
--- a/src/doc/en/developer/coding_in_python.rst
+++ b/src/doc/en/developer/coding_in_python.rst
@@ -56,7 +56,9 @@ using one of two mechanisms:
   For example, on Linux distributions that only ship newer versions
   of Python, the backport packages may not be available as
   system packages. Hence, if there is a need to use the backport
-  packages, they should be imported like this::
+  packages, they should be imported like this:
+
+  .. CODE-BLOCK:: python
 
         import sys
 
@@ -102,7 +104,7 @@ eventually be changed so that the trailing underscore is always
 included. This section describes these special methods.
 
 All objects in Sage should derive from the Cython extension class
-``SageObject``:
+:class:`SageObject`:
 
 .. CODE-BLOCK:: python
 
--- a/src/doc/en/developer/coding_in_python.rst
+++ b/src/doc/en/developer/coding_in_python.rst
@@ -82,7 +82,7 @@ into the structure of Sage. In particular, much of Sage is implemented
 in the object-oriented language Python, and there is a hierarchy of
 classes that organize code and functionality. For example, if you
 implement elements of a ring, your class should derive from
-``sage.structure.element.RingElement``, rather than starting from
+:class:`sage.structure.element.RingElement`, rather than starting from
 scratch. Try to figure out how your code should fit in with other Sage
 code, and design it accordingly.
 
@@ -90,15 +90,15 @@ code, and design it accordingly.
 Special Sage functions
 ======================
 
-Functions with leading and trailing double underscores ``__XXX__`` are
+Functions with leading and trailing double underscores :meth:`__XXX__` are
 all predefined by Python. Functions with leading and trailing single
-underscores ``_XXX_`` are defined for Sage. Functions with a single
+underscores :meth:`_XXX_` are defined for Sage. Functions with a single
 leading underscore are meant to be semi-private, and those with a
 double leading underscore are considered really private. Users can
 create functions with leading and trailing underscores.
 
 Just as Python has many standard special methods for objects, Sage
-also has special methods. They are typically of the form ``_XXX_``.
+also has special methods. They are typically of the form :meth:`_XXX_`.
 In a few cases, the trailing underscore is not included, but this will
 eventually be changed so that the trailing underscore is always
 included. This section describes these special methods.
@@ -122,7 +122,7 @@ or from some other already existing Sage class:
     class MyFavoriteAlgebra(Parent):
         ...
 
-You should implement the ``_latex_`` and ``_repr_`` method for every
+You should implement the :meth:`_latex_` and :meth:`_repr_` method for every
 object. The other methods depend on the nature of the object.
 
 
@@ -132,7 +132,7 @@ LaTeX representation
 Every object ``x`` in Sage should support the command ``latex(x)``, so
 that any Sage object can be easily and accurately displayed via
 LaTeX. Here is how to make a class (and therefore its instances)
-support the command ``latex``.
+support the command :func:`latex`.
 
 #. Define a method ``_latex_(self)`` that returns a LaTeX
    representation of your object. It should be something that can be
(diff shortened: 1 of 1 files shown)


mkoeppe:
Other follow-ups:
- #37649
- #37650

src/doc/en/developer/coding_in_python.rst
```diff
@@ -37,14 +37,34 @@ using one of two mechanisms:
 - Backport packages
 
   - `importlib_metadata <../reference/spkg/importlib_metadata>`_
-    (to be used in place of ``importlib.metadata``),
+    (to be used in place of ``importlib.metadata`` when Python older
```
kwankyu, line 40: Not related with this PR, but you may want to fix the line 18.

kwankyu (approved):
LGTM.

src/doc/en/developer/coding_in_python.rst
```diff
@@ -37,14 +37,34 @@ using one of two mechanisms:
...
+  versions of the packages that provide features of Python 3.11
+  are available.
+
+  The backport packages should not be imported unconditionally.
+  For example, on Linux distributions that only ship newer versions
+  of Python, the backport packages may not be available as
+  system packages. Hence, if there is a need to use the backport
+  packages, they should be imported like this::
+
+        import sys
```
kwankyu, line 63: Instead of `::`:

```sphinx
.. CODE-BLOCK:: python
```

src/doc/en/developer/coding_in_python.rst
mkoeppe: I've pushed a change in 52001daab13, not sure if this is what you had in mind

src/doc/en/developer/coding_in_python.rst
mkoeppe: Done in 6f02bda7b8d

tornaria:
Can you make a list of the features from backport packages not available on python 3.9 that could be useful?

I think right now nothing in sagelib uses the backport pkgs, seems better to keep it that way.

mkoeppe:
That's out of scope of this PR.

src/doc/en/developer/coding_in_python.rst
kwankyu: > I've pushed a change in [52001da](https://github.com/sagemath/sage/commit/52001daab13c74cfa3291c8270c2140457ff96b9), not sure if this is what you had in mind

No. I thought I saw a sentence mentioning `src/setup.cfg.m4`, which is gone. 

Thanks, anyway :-)

src/doc/en/developer/coding_in_python.rst
kwankyu: Thanks.

src/doc/en/developer/coding_in_python.rst
mkoeppe: Ah, this may have disappeared in the rebasing that I did.

mkoeppe:
Thanks for the review!

src/doc/en/developer/coding_in_python.rst
```diff
@@ -38,14 +38,36 @@ using one of two mechanisms:
...
+
+  .. CODE-BLOCK:: python
+
+        import sys
+
+        if sys.version_info < (3, 11):
+            # Use backport package providing Python 3.11 features
+            from importlib_resources import files
+        else:
+            from importlib.resources import files
```
tornaria, line 69: This gives the impression that we recommend using the backport packages. IMO, we should discourage the backport packages. Right now, afaict, nothing in the sage library requires the backport packages and we should keep it that way.

The risk is that if someone uses the backport packages it might use a newer feature not available on current python. We've already pointed out that distros are not shipping some of the backport packages.

Please remove this suggestion.

tornaria:
> Can you make a list of the features from backport packages not available on python 3.9 that could be useful?
> 
> I think right now nothing in sagelib uses the backport pkgs, seems better to keep it that way.

My point here is: if there is a concrete feature that justifies using the backport packages for python <= 3.11, it would be reasonable to compromise. But what is that feature?

I know `importlib.resources` does not support namespace packages on 3.9, but we can avoid that for the time being and 3.9 is on its way out soon. That is the only feature I'm aware of that might need a backport package. Am I missing something?

src/doc/en/developer/coding_in_python.rst
mkoeppe: It's documenting the status quo, which was established by positively reviewed PRs in the past.

mkoeppe:
@tornaria Once more, this is outside of the scope of this PR. It is improving the documentation of what is the current policy.

You are asking for a justification of the status quo and/or to change the policy. That's a separate issue -- and, most importantly, the burden of justifying the change is on you there.

tornaria:
> @tornaria Once more, this is outside of the scope of this PR. It is improving the documentation of what is the current policy.
> 
> You are asking for a justification of the status quo and/or to change the policy. That's a separate issue -- and, most importantly, the burden of justifying the change is on you there.

What you document here *is not* the status quo: nothing in sagelib uses any backport package afaict. It's already agreed that the backport packages should not be used beyond python 3.11, but I don't think we really need them even for 3.9.

If there's a concrete feature that is available on 3.12 but not on 3.9 that we might want to use, it would be fine to refer to the specific feature.

mkoeppe:
> What you document here _is not_ the status quo: nothing in sagelib uses any backport package afaict

The status quo is that nothing currently uses it and that developers can use it because it's a declared dependency.

tornaria:
> The status quo is that nothing currently uses it and that developers can use it because it's a declared dependency.

I don't understand your position. If you need some feature for the modularization, just say so, and we can mention the particular feature. Otherwise, we all benefit from reducing dependencies, modularization as well.

At the very least add a warning that features not available on python 3.12 should not be used. But really: since we are not using it, just dropping it seems the best approach.

mkoeppe:
> > The status quo is that nothing currently uses it and that developers can use it because it's a declared dependency.
> 
> I don't understand your position.

What "position"? You disagree that it is a correct description of the status quo?

tornaria:
> > I don't understand your position.
> 
> What "position"? You disagree that it is a correct description of the status quo?

Yes, I disagree. These packages were added as requirements to some other packages (jsonschema, virtualenv) on older versions of python. There is no status quo or policy about *using* these packages on sagemath itself.

Meanwhile, I believe it has already been agreed not to use them beyond python 3.11, so developers have to be at least warned about this (this is `needs work`).

If there is a useful feature available on 3.11 or 3.12 but not on older python, that might be a good reason. Otherwise, it's just simpler to discourage them altogether to avoid somebody inadvertently using some feature that causes trouble later. That is my position, but we don't know what is yours.

mkoeppe:
> Yes, I disagree. These packages were added as requirements to some other packages (jsonschema, virtualenv) on older versions of python. There is no status quo or policy about _using_ these packages on sagemath itself.

But @tornaria, this is obviously false, as very first link in the PR description shows.

mkoeppe:
Irrelevant, as I am sure you understand.

Issue #37656, opened 2024-03-23
Fix quadratic_defect


nbruin:
### Expected Behavior

From https://groups.google.com/g/sage-devel/c/KfvLDTqf4rk/m/_4SAoCDtAgAJ :
```
Qx.<x> = QQ[]
K.<t> = NumberField(x^10 - x^8 - 2*x^7 - x^6 + 2*x^5 + 2*x^4 - x^2 + 1)
p = K.prime_factors(2)[0]
pi = K.uniformizer(p)
a = 1 + pi^3
K.quadratic_defect(a,p)
```
should work

### Actual Behavior

It actually produces a
```
TypeError: cannot convert non-integral float to integer
```

### Additional Information

Reason is clear:
the line that produces the error is:
https://github.com/sagemath/sage/blob/ccc11b6285ce0ee89dbbc3a7645d3b1af5db7534/src/sage/rings/number_field/number_field.py#L2605
That code can never have worked properly, but in python 3, `1/2` will be a float and the finite field element constructed here cannot be raised to a float power. A line that could work is:
```
s = self(q((a - 1) / pi**w).sqrt())
```
although it would probably worth checking if there are no other bugs in the code. Doctests need to be improved for this code too, since this error must have been lurking since the implementation was merged.

The post has another suggestion to change the computation of a uniformizing element, presently:

https://github.com/sagemath/sage/blob/ccc11b6285ce0ee89dbbc3a7645d3b1af5db7534/src/sage/rings/number_field/number_field.py#L2583-L2587

and use `pi = K.uniformizer(p)` instead. It looks like that is indeed a more robust way (and internally does actually much the same thing but with more knowledge of the internal representation)

### Environment

```markdown
- **Sage Version**: 10.3
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

PR #37657, opened 2024-03-23
fix quadratic_defect

nbruin:
should resolve Fix quadratic_defect #37656 

changes suggested on issue are implemented

### :memo: Checklist


nbruin:
I've made a pull request with the minimal changes implemented as a start.  It would be a nice `easy` ticket for someone to do further testing and improve test coverage of this routine.
(searching the source, this routine isn't actually used elsewhere in sage, but as the report shows, the routine is apparently used by people who use sage!)

mkoeppe:
@nbruin note https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

rburing:
A doctest can be taken from the original report on Ask SageMath: https://ask.sagemath.org/question/76622/bug-in-quadratic_defect/

nbruin:
question raised in https://groups.google.com/g/sage-devel/c/KfvLDTqf4rk/m/Vlj6D1AtAwAJ : would the method be more at home as a method on an element rather than on the parent (the number field)? The latter is a `pyx` file, so it could be that the author wasn't comfortably writing cython and therefore opted to put it in a "python" file.

fchapoton:
looks good enough to me. Could somebody please confirm, so we can move forward here ?

fchapoton:
@nbruin ? @rburing ? a simple approval would be welcome

PR #37659, opened 2024-03-24
fix ruff codes UP012 and UP023

fchapoton:
scripted using `ruff` ; only two minor changes

see https://docs.astral.sh/ruff/rules/#pyupgrade-up

### :memo: Checklist


mkoeppe (approved):
LGTM.

PR #37661, opened 2024-03-24
some cross-references btw doc of RR, RBF, ...

mezzarobba:
(mainly intended to make RBF and CBF, which are the most powerful
implementations for most purposes, more discoverable)


mkoeppe:
LGTM.

Such cross-references could also be added in `sage.rings.complex_interval` and `real_interval_absolute`

mezzarobba:
> LGTM.

Thanks!

> Such cross-references could also be added in `sage.rings.complex_interval` and `real_interval_absolute`

Yes, I left them aside because they are largely internal modules that the casual user is less likely to discover. (This PR was prompted by talking to an experienced Sage user who wished that Sage had an interface to Arb.)

PR #37673, opened 2024-03-26
upgrade msolve to 0.6.5

dimpase:
routine upgrade, dropping already applied patch


dimpase:
https://github.com/algebraic-solving/msolve/issues/131 (non-0 exit code from `msolve -h`) breaks `sage.feature.msolve`

mezzarobba (approved):
lgtm, thank you!

src/sage/features/msolve.py
```diff
@@ -55,10 +55,10 @@ def is_functional(self):
         """
         msolve_out = subprocess.run(["msolve", "-h"], capture_output=True)
 
-        if msolve_out.returncode != 0:
-            return FeatureTestResult(self, False, reason="msolve -h returned "
-                                f"non-zero exit status {msolve_out.returncode}")
-        elif (msolve_out.stdout[:46] !=
+#        if msolve_out.returncode != 0:
+#            return FeatureTestResult(self, False, reason="msolve -h returned "
+#                                f"non-zero exit status {msolve_out.returncode}")
```
mkoeppe, line 60:
```suggestion
```
Normally we don't leave commented out stuff in our files.

mezzarobba: What do you mean by “normally”? I would generally agree that it is not a good idea to use comments instead of git for keeping track of history, but for a change like this one that will hopefully be reverted with the next msolve upgrade, it seems very natural to me to keep the commented out code in the file.

dimpase: I have submitted a PR to upstream fixing their regression. They literally changed exit(0) to exit(1)  for `msolve -h`, as a part of "cleanup"
https://github.com/algebraic-solving/msolve/pull/132

PR #37675, opened 2024-03-26
Allow `.maximal_order()` of `quaternion_algebra.py` to extend a given order

S17A05:
1. Added an optional argument `order_basis` to `.maximal_order()` to allow for extension of orders: If `order_basis` defines an order of the given quaternion algebra, then the method now returns a maximal order that contains the order specified by the given basis.
~~2. Added comparison functions `__le__`, `__ge__`, `__lt__` and `__gt__` for quaternion orders.~~
~~3. Small edits of the comparison functions `__eq__` and `__ne__` for quaternion orders.~~
2.  Added a general comparison function `__richcmp__` for quaternion orders and removed the explicit implementation of `__eq__` and `__ne__`.

More details on 1.:
The algorithm used in `.maximal_order()` currently uses the standard basis `(1,i,j,k)` as a starting basis, but it extends to any starting basis whose first entry is equal to `1`. Using the helper method `basis_for_quaternion_lattice`, we transform any quaternion order basis input by the user to a basis of this form.

Depends on #37644: Naturally, the extended algorithm can only work correctly if the base algorithm works correctly in the first place.


src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -1634,6 +1689,74 @@ def __ne__(self, other):
         """
         return not self.__eq__(other)
 
+    def __le__(self, other):
+        """
+        Compare orders self and other.
```
tscrim:
```suggestion
        Compare orders ``self`` and ``other``.
```
You should also use a `__richcmp__` instead. (If not, I forget if Python also requires all of the others to be implemented if you have just a, e.g., `__lt__`.)

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -3577,8 +3711,7 @@ def normalize_basis_at_p(e, p, B=QuaternionAlgebraElement_abstract.pair):
 
             # Ensures that (B(f0,f0)/2).valuation(p) <= B(f0,f1).valuation(p)
             if B(f0, f1).valuation(p) + 1 < B(f0, f0).valuation(p):
-                f0 += f1
-                f1 = f0
+                f0, f1 = f0 + f1, f0
```
tscrim: This change is makes the code behave in a very subtle way. It would be better to make it explicit.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -766,9 +783,24 @@ def maximal_order(self, take_shortcuts=True):
...
+            ....:           (-511, 608), (493, 880), (105, -709), (-213, 530),
+            ....:           (97, 745)]
+            sage: all(QuaternionAlgebra(a, b).maximal_order().is_maximal()
+            ....:           for (a, b) in invars)
+            True
         """
         if self.base_ring() != QQ:
-            raise NotImplementedError("maximal order only implemented for rational quaternion algebras")
+            raise NotImplementedError("maximal order only implemented for"
+                                        " rational quaternion algebras")
```
tscrim: Please revert, the 80 char/line is a guideline and you make the code harder to read by the manual break.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -766,9 +783,24 @@ def maximal_order(self, take_shortcuts=True):
...
+        Check that :issue:`37417` and the first part of :issue:`37217` are fixed::
+
+            sage: invars = [(-4, -28), (-292, -732), (-48, -564), (-436, -768),
+            ....:           (-752, -708), (885, 545), (411, -710), (-411, 593),
+            ....:           (805, -591), (-921, 353), (409, 96), (394, 873),
+            ....:           (353, -722), (730, 830), (-466, -427), (-213, -630),
+            ....:           (-511, 608), (493, 880), (105, -709), (-213, 530),
+            ....:           (97, 745)]
+            sage: all(QuaternionAlgebra(a, b).maximal_order().is_maximal()
+            ....:           for (a, b) in invars)
```
tscrim:
```suggestion
            sage: all(QuaternionAlgebra(a, b).maximal_order().is_maximal()
            ....:     for (a, b) in invars)
```
PEP8 suggests this alignment.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -687,14 +687,19 @@ def __init__(self, base_ring, a, b, names='i,j,k'):
         self._gens = (self([0, 1, 0, 0]), self([0, 0, 1, 0]), self([0, 0, 0, 1]))
 
     @cached_method
-    def maximal_order(self, take_shortcuts=True):
+    def maximal_order(self, order_basis=None, take_shortcuts=True):
```
tscrim: This is a backwards incompatible change. I would instead put `order_basis` last because of this.

tscrim (changes requested):

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Regarding `__richcmp__`: I tried implementing that, but it resulted in multiple errors in testing; I'm not sure why, though - maybe because I also removed `__eq__` and `__ne__` in the process?

S17A05:
The testing timeout of `algebras.fusion_rings.fusion_ring` in the CI should be unrelated; I can't reproduce it locally, though.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: The following is how I tried to implement it:
```
def __richcmp__(self, right, op):
    return self.unit_ideal()._richcmp_(right.unit_ideal(), op)
```
However, building and testing with this gives me the error 
`TypeError: '<=' not supported between instances of 'QuaternionOrder_with_category' and 'QuaternionOrder_with_category'`

Maybe I'm misunderstanding how `__richcmp__` is supposed to be implemented though.

tscrim: You would  use
```python
def __richcmp__(self, other, op):
    from sage.structure.richcmp import richcmp, op_NE
    if not isinstance(other, QuaternionOrder):
        return op == op_NE
    return richcmp(self.unit_ideal(), other.unit_ideal(), op)
```

tscrim: Also you mark the class as `@richcmp_method` from `from sage.structure.richcmp importrichcmp_method`.

S17A05: Thanks, that worked! Does the `_richcmp_`-method of the class `QuaternionFractionalIdeal_rational` need similar changes?

tscrim: I don't know. I'm not familiar with that code. So you will have to test, but it might be okay.

S17A05: It seems to work fine, so I'll leave it as is - I'm just surprised that it does work, given the extra steps that were necessary for the `__richcmp__` here.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -689,14 +691,19 @@ def __init__(self, base_ring, a, b, names='i,j,k'):
...
-        The algorithm used is from [Voi2012]_.
+        If ``order_basis`` is specified, the resulting maximal order
+        will contain the order of the quaternion algebra given by this
+        basis. The algorithm used is from [Voi2012]_.
 
         INPUT:
 
+        - ``order_basis`` -- (optional, default: ``None``) a basis of an
+          order of this quaternion algebra
+
```
tscrim: Since this is now the second argument, it might be better to put this after.

src/sage/algebras/quatalg/quaternion_algebra.py
```diff
@@ -793,7 +813,8 @@ def maximal_order(self, take_shortcuts=True):
         # (every quaternion algebra of prime discriminant has a representation
         #  of such a form though)
         a, b = self.invariants()
-        if take_shortcuts and d_A.is_prime() and a in ZZ and b in ZZ:
+        if (not order_basis and take_shortcuts and d_A.is_prime()
+        and a in ZZ and b in ZZ):
```
tscrim:
```suggestion
        if (not order_basis and take_shortcuts and d_A.is_prime()
            and a in ZZ and b in ZZ):
```
PEP8

tscrim (approved):
Thank you. Two trivial things that I would appreciate. Once done (or rejected), then you can set a positive review.

src/sage/algebras/quatalg/quaternion_algebra.py
S17A05: Oops, I forgot to do this when I switched the order of the arguments; it's fixed now.

PR #37677, opened 2024-03-27
some simplifications in moment-angle complex

### LINKED ISSUE #36217: Improve Euler characteristic implementation for moment-angle complexes
### Problem Description

The current implementation of the Euler characteristic just computes it as an alternating sum of Betti numbers (which is the most natural way to compute it), but this is possibly easier to do for moment-angle complexes.

### Proposed Solution

We can see from the Corollary 4.6.3 in the book given below that the Euler characteristic of a moment-angle complex is 0, if the associated simplicial complex is not a simplex. This means that we can avoid using the Betti numbers (and consequently the ``homology()`` method), and just simply check whether the associated simplicial complex is a simplex (in the negative case, the Euler characteristic is 1, which is not hard to see).

### Alternatives Considered

This method is not inefficient as it is, and perhaps the better approach is to leave it as is, because that's the way it is defined mathematically. But, the proposed approach allows us to compute it very efficiently, even for large simplicial complexes associated with the moment-angle complex.

### Additional Information

The Corollary mentioned above can be found in [this book](https://arxiv.org/abs/1210.2368v3).
This issue is regarding the PR #35875, and, consequently, is a part of #35640 (GSoC2023). 

fchapoton:
a bunch of code details in the modified file

fixes #36217

### :memo: Checklist

### DIFF
--- a/src/sage/topology/moment_angle_complex.py
+++ b/src/sage/topology/moment_angle_complex.py
@@ -483,10 +483,12 @@ class MomentAngleComplex(UniqueRepresentation, SageObject):
         n = len(vertices)
         invfac = []
 
+        in_field = base_ring in Fields()
+
         for j in range(n + 1):
             for x in combinations(vertices, j):
                 S = self._simplicial_complex.generated_subcomplex(x)
-                if base_ring in Fields():
+                if in_field:
                     invfac.append(S.homology(i - j - 1, base_ring=base_ring,
                                              cohomology=cohomology, algorithm=algorithm,
                                              verbose=verbose, reduced=True).dimension())
@@ -495,7 +497,7 @@ class MomentAngleComplex(UniqueRepresentation, SageObject):
                                              cohomology=cohomology, algorithm=algorithm,
                                              verbose=verbose, reduced=True)._original_invts)
 
-        if base_ring in Fields():
+        if in_field:
             return HomologyGroup(sum(invfac), base_ring)
 
         m = len(invfac)
@@ -700,7 +702,7 @@ class MomentAngleComplex(UniqueRepresentation, SageObject):
         except AttributeError:
             return H.dimension()
 
-    def euler_characteristic(self) -> int:
+    def euler_characteristic(self):
         """
         Return the Euler characteristic of ``self``.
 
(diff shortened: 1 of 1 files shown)


src/sage/topology/moment_angle_complex.py
```diff
@@ -488,19 +483,19 @@ def _homology_group(self, i, base_ring, cohomology, algorithm, verbose, reduced)
         n = len(vertices)
         invfac = []
 
-        for j in range(n+1):
+        for j in range(n + 1):
             for x in combinations(vertices, j):
                 S = self._simplicial_complex.generated_subcomplex(x)
-                if base_ring.is_field():
-                    invfac.append(S.homology(i-j-1, base_ring=base_ring,
+                if base_ring in Fields():
```
tscrim: We should pull this test outside the `for` loops.

src/sage/topology/moment_angle_complex.py
```diff
@@ -488,19 +483,19 @@ def _homology_group(self, i, base_ring, cohomology, algorithm, verbose, reduced)
...
                                              cohomology=cohomology, algorithm=algorithm,
                                              verbose=verbose, reduced=True).dimension())
                 else:
-                    invfac.extend(S.homology(i-j-1, base_ring=base_ring,
+                    invfac.extend(S.homology(i - j - 1, base_ring=base_ring,
                                              cohomology=cohomology, algorithm=algorithm,
                                              verbose=verbose, reduced=True)._original_invts)
 
-        if base_ring.is_field():
+        if base_ring in Fields():
```
tscrim: Which we can then reuse here as well.

src/sage/topology/moment_angle_complex.py
```diff
@@ -724,8 +721,8 @@ def euler_characteristic(self):
             sage: Z.euler_characteristic()
             1
         """
-        betti_numbers = self.betti()
-        return ZZ.sum((-1)**n * betti_numbers[n] for n in range(self.dimension() + 1))
+        sc = self.simplicial_complex()
+        return 1 if sc.dimension() + 1 == len(sc.vertices()) else 0
```
tscrim:
```suggestion
        return ZZ.one() if sc.dimension() + 1 == len(sc.vertices()) else ZZ.zero()
```
We shouldn't return `int`s.

tscrim (changes requested):
A few little things.

fchapoton:
Thanks for the good idea. All done.

tscrim (approved):
Thanks. LGTM.

PR #37684, opened 2024-03-27
fix ascii_art of empty chain complex

### LINKED ISSUE #37678: ascii_art fails on empty chain complex
### Steps To Reproduce

1. I'm on Sage 10.2 on MacOS 12.4, working in Jupyter via sage --notebook.
2. Creating the trivial chain complex `E=ChainComplex(base_ring=ZZ)` works, but `ascii_art(E)` raises the index error listed below.

### Expected Behavior

It should return either `0` or an empty string.

### Actual Behavior

```---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In [81], line 1
----> 1 ascii_art(E)

File /private/var/tmp/sage-10.2-current/local/var/lib/sage/venv-python3.11.1/lib/python3.11/site-packages/sage/typeset/ascii_art.py:288, in ascii_art(*obj, **kwds)
    286     raise ValueError('unknown keyword arguments: {0}'.format(list(kwds)))
    287 if len(obj) == 1:
--> 288     return _ascii_art_factory.build(obj[0], baseline=baseline)
    289 if not isinstance(separator, AsciiArt):
    290     separator = _ascii_art_factory.build(separator, baseline=sep_baseline)

File /private/var/tmp/sage-10.2-current/local/var/lib/sage/venv-python3.11.1/lib/python3.11/site-packages/sage/typeset/character_art_factory.py:123, in CharacterArtFactory.build(self, obj, baseline)
    121     return obj
    122 if isinstance(obj, SageObject):
--> 123     return self.build_from_magic_method(obj, baseline)
    124 if baseline is None:
    125     baseline = 0

File /private/var/tmp/sage-10.2-current/local/var/lib/sage/venv-python3.11.1/lib/python3.11/site-packages/sage/typeset/character_art_factory.py:171, in CharacterArtFactory.build_from_magic_method(self, obj, baseline)
    154 """
    155 Return the character art object created by the object's magic method
    156 
   (...)
    168     <class 'sage.typeset.ascii_art.AsciiArt'>
    169 """
    170 magic_method = getattr(obj, self.magic_method_name)
--> 171 ret = magic_method()
    172 if baseline is not None:
    173     ret._baseline = baseline

File /private/var/tmp/sage-10.2-current/local/var/lib/sage/venv-python3.11.1/lib/python3.11/site-packages/sage/homology/chain_complex.py:1806, in ChainComplex_class._ascii_art_(self)
   1804         result_ordered += arrow_art(n) + module_art(n)
   1805     result = [result_ordered] + result
-> 1806 concatenated = result[0]
   1807 for r in result[1:]:
   1808     concatenated += AsciiArt([' ... ']) + r

IndexError: list index out of range```

### Additional Information

_No response_

### Environment

```markdown
- **MacOS 10.2**:
- **Sage 10.2**:
```

### Checklist

dimpase:
This fixes #37678 (ascii_art() fails on a trivial chain complex)

### DIFF
--- a/src/sage/homology/chain_complex.py
+++ b/src/sage/homology/chain_complex.py
@@ -1889,7 +1889,7 @@ class ChainComplex_class(Parent):
         try:
             concatenated = result[0]
             for r in result[1:]:
-                concatenated += AsciiArt([' ... ']) + r
+                concatenated += UnicodeArt([' ... ']) + r
             return concatenated
         except IndexError:
             return []
--- a/src/sage/homology/chain_complex.py
+++ b/src/sage/homology/chain_complex.py
@@ -478,6 +478,10 @@ class Chain_class(ModuleElement):
                d_2       d_1  ⎛4⎞  d_0  ⎜2⎟  d_-1
             0 <──── (0) <──── ⎝5⎠ <──── ⎝3⎠ <───── 0
             sage: unicode_art(C())
+                                        ⎛0⎞
+               d_2       d_1  ⎛0⎞  d_0  ⎜0⎟  d_-1
+            0 <──── (0) <──── ⎝0⎠ <──── ⎝0⎠ <───── 0
+            sage: unicode_art(ChainComplex())
             []
         """
         from sage.typeset.unicode_art import UnicodeArt
--- a/src/sage/homology/chain_complex.py
+++ b/src/sage/homology/chain_complex.py
@@ -509,7 +509,7 @@ class Chain_class(ModuleElement):
         try:
             concatenated = result[0]
             for r in result[1:]:
-                concatenated += AsciiArt([' ... ']) + r
+                concatenated += UnicodeArt([' ... ']) + r
             return concatenated
         except IndexError:
             return []
--- a/src/sage/homology/chain_complex.py
+++ b/src/sage/homology/chain_complex.py
@@ -418,6 +418,15 @@ class Chain_class(ModuleElement):
                d_2       d_1       d_0  [1]  d_-1
             0 <---- [0] <---- [4] <---- [2] <----- 0
                               [5]       [3]
+
+        TESTS:
+
+        check that :issue:`37678` is fixed::
+
+            sage: C = ChainComplex(base_ring=ZZ)
+            sage: ascii_art(C())
+            []
+
         """
         from sage.typeset.ascii_art import AsciiArt
 
@@ -468,6 +477,8 @@ class Chain_class(ModuleElement):
                                         ⎛1⎞
                d_2       d_1  ⎛4⎞  d_0  ⎜2⎟  d_-1
             0 <──── (0) <──── ⎝5⎠ <──── ⎝3⎠ <───── 0
+            sage: unicode_art(C())
+            []
         """
         from sage.typeset.unicode_art import UnicodeArt
 
@@ -495,10 +506,13 @@ class Chain_class(ModuleElement):
             for n in ordered:
                 result_ordered += arrow_art(n) + vector_art(n)
             result = [result_ordered] + result
-        concatenated = result[0]
-        for r in result[1:]:
-            concatenated += UnicodeArt([u' ... ']) + r
-        return concatenated
+        try:
+            concatenated = result[0]
+            for r in result[1:]:
+                concatenated += AsciiArt([' ... ']) + r
+            return concatenated
+        except IndexError:
+            return []
 
     def is_cycle(self):
         """
@@ -1778,6 +1792,8 @@ class ChainComplex_class(Parent):
             sage: ascii_art(D)
                         [1]                             [1]       [0]       [1]
              0 <-- C_7 <---- C_6 <-- 0  ...  0 <-- C_3 <---- C_2 <---- C_1 <---- C_0 <-- 0
+             sage: ascii_art(ChainComplex(base_ring=ZZ))
+             []
         """
         from sage.typeset.ascii_art import AsciiArt
 
@@ -1834,6 +1850,14 @@ class ChainComplex_class(Parent):
             sage: unicode_art(D)
                       (1)                           (1)     (0)     (1)
             0 <── C_7 <── C_6 <── 0  ...  0 <── C_3 <── C_2 <── C_1 <── C_0 <── 0
+
+        TESTS:
+
+        check that :issue:`37678` is fixed::
+
+            sage: C = ChainComplex(base_ring=ZZ)
+            sage: unicode_art(C)
+            []
         """
         from sage.typeset.unicode_art import UnicodeArt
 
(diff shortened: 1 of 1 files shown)


jhpalmieri:
In principle it looks good. Could you add a doctest showing the new behavior?

dimpase:
Should we really return  `[]`, or rather `''` ?

There are similar troubles for `unicode_art()`, I'll fix them too

jhpalmieri:
> Should we really return `[]`, or rather `''` ?

I don't have a preference. It should not be a common situation.

src/sage/homology/chain_complex.py
```diff
@@ -492,10 +506,13 @@ def vector_art(d):
...
                 result_ordered += arrow_art(n) + vector_art(n)
             result = [result_ordered] + result
-        concatenated = result[0]
-        for r in result[1:]:
-            concatenated += UnicodeArt([u' ... ']) + r
-        return concatenated
+        try:
+            concatenated = result[0]
+            for r in result[1:]:
+                concatenated += AsciiArt([' ... ']) + r
```
jhpalmieri:
```suggestion
                concatenated += UnicodeArt([' ... ']) + r
```

src/sage/homology/chain_complex.py
```diff
@@ -1856,10 +1886,13 @@ def module_art(n):
...
                 result_ordered += arrow_art(n) + module_art(n)
             result = [result_ordered] + result
-        concatenated = result[0]
-        for r in result[1:]:
-            concatenated += UnicodeArt([u' ... ']) + r
-        return concatenated
+        try:
+            concatenated = result[0]
+            for r in result[1:]:
+                concatenated += AsciiArt([' ... ']) + r
```
jhpalmieri:
```suggestion
                concatenated += UnicodeArt([' ... ']) + r
```

src/sage/homology/chain_complex.py
```diff
@@ -465,6 +477,8 @@ def _unicode_art_(self):
                                         ⎛1⎞
                d_2       d_1  ⎛4⎞  d_0  ⎜2⎟  d_-1
             0 <──── (0) <──── ⎝5⎠ <──── ⎝3⎠ <───── 0
+            sage: unicode_art(C())
+            []
```
jhpalmieri: This doctest fails for me.

jhpalmieri (commented):
One failing doctest that I'm not sure what to do with, plus two minor corrections.

kwankyu:
> Should we really return `[]`, or rather `''` ?
> 
> There are similar troubles for `unicode_art()`, I'll fix them too

The same object as below?

```
 sage: from sage.typeset.ascii_art import AsciiArt
sage: AsciiArt.empty()

sage: from sage.typeset.unicode_art import UnicodeArt
sage: UnicodeArt.empty()

```

tscrim:
Thanks for fixing this. I had noticed it earlier this month, but I didn't create an issue about it as I had planned to fix it quickly and then forgot...

tscrim:
IMO, the trivial chain complex should just be `0` to match the latex output.

dimpase:
CI is clean (modulo irrelevant warnings on codelines not covered by tests - they are)

tscrim:
I really think this should return `0` to match the latex output.

Moreover, it should return an ascii/unicode art object to (easily) work with other such objects.

I think checking for an index error for the empty case is not good a programming control method. You should just check `if not result` as the only thing that will fail with the `IndexError` is the `result[0]` (the slice doesn't fail when I checked this for length 0 and 1 lists).

dimpase:
returning 0 seems to be like a wrong thing to do. Should it rather return $\emptyset$, i.e. `\emptyset` ?

tscrim:
No, it is not the empty complex. Essentially, the chain complexes consider themselves to be infinite in both directions with an infinite number of trivial modules with trivial morphisms. (Of course, internally we cannot do that.) This is also important for operations such as taking tensor products.

```
sage: MS = MatrixSpace(ZZ, 0)
sage: C = ChainComplex({1:MS.zero(), 2:MS.zero()}, degree_of_differential=-1)
sage: C
Trivial chain complex over Integer Ring
sage: C.differential(5)
[]
sage: R.<x,y,z> = QQ[]
sage: C1 = ChainComplex({1:matrix([[x]])}, degree_of_differential=-1)
sage: C.tensor([C1])
Trivial chain complex over Integer Ring
```

Since we ascii/unicode/latex the chain complex as the chain of modules, returning `0` makes sense and is correct. It is a bit better than $0 \to 0$ since that would indicate a nontrivial differential.

antieau:
> No, it is not the empty complex. Essentially, the chain complexes consider themselves to be infinite in both directions with an infinite number of trivial modules with trivial morphisms. (Of course, internally we cannot do that.) This is also important for operations such as taking tensor products.

I agree with this: the category of chain complexes has a zero object but no $\emptyset$ object.

dimpase:
I'll try to update tonight

dimpase:
I feel that $0$ is confusing. I looked in Weibel's  "Introduction to homological algebra'' to see his notation for the trivial chain complex, and he uses  on [p3, Ex 1.1.5](https://people.math.rochester.edu/faculty/doug/otherpapers/weibel-hom.pdf)   “0” (which is [“](https://www.compart.com/en/unicode/U+201C)0[”](https://www.compart.com/en/unicode/U+201D) ).
Not $0$. 

More precisely, he uses both :-) 
But these must be LaTeX typos, as it's a hell to do this without Unicode in maths mode. In fact, GitHub (in my recent Firefox on Linux)  shows these double quotes equal, but these are NOT! The opening one is U+201C, the closing one in U+201D. YMMV. Perhaps my fonts are buggy

dimpase:
How does one do  “0”  in `ascii_art`? Just
```
``0''
```
?

tscrim:
I think `0` by itself is clear. On the other hand, the quotes just make it more confusing to me. In the context of a book where you need to distinguish between the trivial chain and other zeros, that is reasonable (although I would have chosen something different like $\vec{0}$ or $0 to 0$). However, within Sage a user can directly inquire about the object with `parent(C)` or `type(C)`. IMO, it is easier and more natural for a user to add additional symbols to a string than remove them.

@jhpalmieri, do you have an opinion on this?

kwankyu:
+1 to `ascii_art(0)` and `unicode_art(0)`, to denote the zero object in the category.

jhpalmieri:
> @jhpalmieri, do you have an opinion on this?

No. We're seriously bikeshedding at this point.

jhpalmieri:
> I feel that 0 is confusing. I looked in Weibel's "Introduction to homological algebra'' to see his notation for the trivial chain complex, and he uses “0” on [p3, Ex 1.1.5](https://people.math.rochester.edu/faculty/doug/otherpapers/weibel-hom.pdf) (which is [“](https://www.compart.com/en/unicode/U+201C)0[”](https://www.compart.com/en/unicode/U+201D) ). Not 0.

I think that Weibel is using the quotes just to emphasize the notation he is explaining (namely: 0). I do not read the quotes as being part of his notation, just 0. But maybe I'm misunderstanding your point.

dimpase:
OK, this should address all the concerns here. Sorry for being slow.

tscrim (approved):
No problem. Thank you.

If anyone else has any comments/issues, please feel free to revert the positive review.

PR #37686, opened 2024-03-28
provide a construction functor using individual functors

mantepse:
We enable sage to find common parents also when symmetric functions are involved. In particular, we provide a functorial construction that takes a commutative (coefficient) ring and produces a (commutative) ring of symmetric functions over this ring.

For Macdonald polynomials and other symmetric functions that involve parameters we cheat a bit: the categories should actually be commutative rings with distinguished elements (namely, the parameters). However, having these is very likely overkill.

This is an alternative to #37220, relying on individual construction functors instead of passing around strings.

### DIFF
--- a/src/sage/combinat/sf/character.py
+++ b/src/sage/combinat/sf/character.py
@@ -37,6 +37,23 @@ from sage.rings.integer import Integer
 
 
 class generic_character(SFA_generic):
+
+    def construction(self):
+        """
+        Return a pair ``(F, R)``, where ``F`` is a
+        :class:`SymmetricFunctionsFunctor` and `R` is a ring, such
+        that ``F(R)`` returns ``self``.
+
+        EXAMPLES::
+
+            sage: ht = SymmetricFunctions(QQ).ht()
+            sage: ht.construction()
+            (SymmetricFunctionsFunctor[induced trivial symmetric group character],
+             Rational Field)
+        """
+        return (CharacterSymmetricFunctionsFunctor(self, self.basis_name()),
+                self.base_ring())
+
     def _my_key(self, la):
         r"""
         A rank function for partitions.
@@ -149,6 +166,61 @@ class generic_character(SFA_generic):
         if k > 0:
             return ~k * self._p.linear_combination((self._p([d]),moebius(k//d))
                                     for d in divisors(k))
+
+
+from sage.combinat.sf.sfa import SymmetricFunctionsFunctor
+class CharacterSymmetricFunctionsFunctor(SymmetricFunctionsFunctor):
+    def __init__(self, basis, name):
+        r"""
+        Initialise the functor.
+
+        INPUT:
+
+        - ``basis`` -- the basis of the Character symmetric function algebra
+        - ``name`` -- the name of the basis
+
+        .. WARNING::
+
+            The codomain of this functor could actually be
+            :class:`CommutativeAlgebras` over the given ring, but
+            parameterized functors are currently not available.
+
+        EXAMPLES::
+
+            sage: from sage.combinat.sf.character import CharacterSymmetricFunctionsFunctor
+            sage: B = SymmetricFunctions(QQ).ht()
+            sage: CharacterSymmetricFunctionsFunctor(B, B.basis_name())
+            SymmetricFunctionsFunctor[induced trivial symmetric group character]
+
+        """
+        super().__init__(basis, name)
+        self._prefix = basis._prefix
+
+    def _apply_functor(self, R):
+        """
+        Apply the functor to an object of ``self``'s domain.
+
+        EXAMPLES::
+
+            sage: B = SymmetricFunctions(QQ).ht()
+            sage: F, R = B.construction()  # indirect doctest
+            sage: F(QQbar)
+            Symmetric Functions over Algebraic Field in the induced trivial
+             symmetric group character basis
+        """
+        from sage.combinat.sf.sf import SymmetricFunctions
+        return self._basis(SymmetricFunctions(R), self._prefix)
+
+    def __eq__(self, other):
+        """
+        EXAMPLES::
+
+            sage: B1 = SymmetricFunctions(QQ).ht()
+            sage: B2 = SymmetricFunctions(QQ).st()
+            sage: B1.construction()[0] == B2.construction()[0]
+            False
+        """
+        return super().__eq__(other) and self._prefix == other._prefix
 
 
 class induced_trivial_character_basis(generic_character):
--- a/src/sage/combinat/sf/dual.py
+++ b/src/sage/combinat/sf/dual.py
@@ -26,7 +26,16 @@ from . import classical
 
 
 class SymmetricFunctionAlgebra_dual(classical.SymmetricFunctionAlgebra_classical):
-    def __init__(self, dual_basis, scalar, scalar_name="", basis_name=None, prefix=None):
+    @staticmethod
+    def __classcall__(cls, dual_basis, scalar, scalar_name="", basis_name=None, prefix=None):
+        """
+        Normalize the arguments.
+        """
+        if prefix is None:
+            prefix = 'd_'+dual_basis.prefix()
+        return super().__classcall__(cls, dual_basis, scalar, scalar_name, basis_name, prefix)
+
+    def __init__(self, dual_basis, scalar, scalar_name, basis_name, prefix):
         r"""
         Generic dual basis of a basis of symmetric functions.
 
@@ -72,9 +81,9 @@ class SymmetricFunctionAlgebra_dual(classical.SymmetricFunctionAlgebra_classical
         EXAMPLES::
 
             sage: e = SymmetricFunctions(QQ).e()
-            sage: f = e.dual_basis(prefix = "m", basis_name="Forgotten symmetric functions"); f
+            sage: f = e.dual_basis(prefix="m", basis_name="Forgotten symmetric functions"); f
             Symmetric Functions over Rational Field in the Forgotten symmetric functions basis
-            sage: TestSuite(f).run(elements = [f[1,1]+2*f[2], f[1]+3*f[1,1]])
+            sage: TestSuite(f).run(elements=[f[1,1]+2*f[2], f[1]+3*f[1,1]])
             sage: TestSuite(f).run() # long time (11s on sage.math, 2011)
 
         This class defines canonical coercions between ``self`` and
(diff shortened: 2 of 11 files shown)


src/sage/combinat/sf/dual.py
```diff
@@ -26,7 +26,16 @@
 
 
 class SymmetricFunctionAlgebra_dual(classical.SymmetricFunctionAlgebra_classical):
-    def __init__(self, dual_basis, scalar, scalar_name="", basis_name=None, prefix=None):
+    @staticmethod
+    def __classcall__(cls, dual_basis, scalar, scalar_name="", basis_name=None, prefix=None):
```
tscrim, line 30: This needs a doctest (e.g., checking the uniqueness).

src/sage/combinat/sf/dual.py
```diff
@@ -890,6 +927,72 @@ def expand(self, n, alphabet='x'):
...
+    A constructor for algebras of symmetric functions constructed by
+    duality.
+
+    EXAMPLES::
+
+        sage: w = SymmetricFunctions(ZZ).witt()
+        sage: w.dual_basis().construction()
+        (SymmetricFunctionsFunctor[dual Witt], Integer Ring)
+    """
+    def __init__(self, basis):
```
tscrim, line 942: This needs a doctest (as always, a `TestSuite(F).run()` is good).

src/sage/combinat/sf/hall_littlewood.py
```diff
@@ -77,7 +77,14 @@ def __repr__(self):
         """
         return self._name + " over %s" % self._sym.base_ring()
 
-    def __init__(self, Sym, t='t'):
+    @staticmethod
+    def __classcall__(cls, Sym, t='t'):
```
tscrim, line 81: This needs a doctest (such as testing uniqueness).

src/sage/combinat/sf/hall_littlewood.py
```diff
@@ -387,6 +394,24 @@ def __init__(self, hall_littlewood):
...
+        that ``F(R)`` returns ``self``.
+
+        EXAMPLES::
+
+            sage: P = SymmetricFunctions(QQ).hall_littlewood(t=2).P()
+            sage: P.construction()
+            (SymmetricFunctionsFunctor[Hall-Littlewood P with t=2], Rational Field)
+        """
+
+        return (HallLittlewoodSymmetricFunctionsFunctor(self,
```
tscrim, line 410:
```suggestion
        """
        return (HallLittlewoodSymmetricFunctionsFunctor(self,
```

src/sage/combinat/sf/hall_littlewood.py
```diff
@@ -673,6 +698,80 @@ def scalar_hl(self, x, t=None):
                                                        orthogonal=True)
 
 
+from sage.combinat.sf.sfa import SymmetricFunctionsFunctor
+class HallLittlewoodSymmetricFunctionsFunctor(SymmetricFunctionsFunctor):
+    def __init__(self, basis, name, t):
+        r"""Initialise the functor.
```
tscrim, line 704:
```suggestion
        r"""
        Initialize the functor.
```

src/sage/combinat/sf/hecke.py
```diff
@@ -121,8 +121,14 @@ class HeckeCharacter(SymmetricFunctionAlgebra_multiplicative):
     - [Ram1991]_
     - [RR1997]_
     """
+    @staticmethod
+    def __classcall__(cls, Sym, q='q'):
```
tscrim, line 125: This needs a doctest (test for the uniqueness).

src/sage/combinat/sf/jack.py
```diff
@@ -50,8 +50,14 @@
 
 
 class Jack(UniqueRepresentation):
+    @staticmethod
```
tscrim, line 53: This needs a doctest.

src/sage/combinat/sf/llt.py
```diff
@@ -91,8 +91,14 @@ class LLT_class(UniqueRepresentation):
         sage: HS3x(HC3t2[3,1])
         2*HSp3[3, 1] + (-2*x+1)*HSp3[4]
     """
+    @staticmethod
+    def __classcall__(cls, Sym, k, t='t'):
```
tscrim, line 95: This needs a doctest.

src/sage/combinat/sf/macdonald.py
```diff
@@ -97,7 +97,14 @@ def __repr__(self):
         """
         return self._name
 
-    def __init__(self, Sym, q='q', t='t'):
+    @staticmethod
+    def __classcall__(cls, Sym, q='q', t='t'):
```
tscrim, line 101: This needs a doctest.

src/sage/combinat/sf/orthotriang.py
```diff
@@ -49,8 +49,14 @@ class SymmetricFunctionAlgebra_orthotriang(sfa.SymmetricFunctionAlgebra_generic)
 
     class Element(sfa.SymmetricFunctionAlgebra_generic.Element):
         pass
+    @staticmethod
+    def __classcall__(cls, Sym, base, scalar, prefix, basis_name, leading_coeff=None):
```
tscrim, line 53: This doesn't serve any purpose, so you can revert this change to this file.

src/sage/combinat/sf/sf.py
```diff
@@ -976,8 +975,6 @@ def witt(self, coerce_h=True, coerce_e=False, coerce_p=False):
         from . import witt
         return witt.SymmetricFunctionAlgebra_witt(self, coerce_h=coerce_h, coerce_e=coerce_e, coerce_p=coerce_p)
     w = witt
-    # Currently needed by sfa.GradedSymmetricFunctionsBases.corresponding_basis_over
-    Witt = witt
```
tscrim, line 980: This is breaking backwards compatibility and creating a slightly inconsistency.

src/sage/combinat/sf/witt.py
```diff
@@ -396,8 +396,14 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
         ....:      #this holds for all odd i and is easily proven by induction
         True
     """
+    @staticmethod
+    def __classcall__(cls, Sym, coerce_h=True, coerce_e=False, coerce_p=False):
```
tscrim, line 400: This change can be reverted here.

src/sage/combinat/sf/character.py
```diff
@@ -151,6 +168,61 @@ def _b_power_k(self, k):
...
+            parameterized functors are currently not available.
+
+        EXAMPLES::
+
+            sage: from sage.combinat.sf.character import CharacterSymmetricFunctionsFunctor
+            sage: B = SymmetricFunctions(QQ).ht()
+            sage: CharacterSymmetricFunctionsFunctor(B, B.basis_name())
+            SymmetricFunctionsFunctor[induced trivial symmetric group character]
+
+        """
```
tscrim, line 195:
```suggestion
            SymmetricFunctionsFunctor[induced trivial symmetric group character]
        """
```

src/sage/combinat/sf/character.py
```diff
@@ -151,6 +168,61 @@ def _b_power_k(self, k):
                                     for d in divisors(k))
 
 
+from sage.combinat.sf.sfa import SymmetricFunctionsFunctor
+class CharacterSymmetricFunctionsFunctor(SymmetricFunctionsFunctor):
+    def __init__(self, basis, name):
```
tscrim, line 173: I would move the `__init__` docstring to the class level.

src/sage/combinat/sf/character.py
```diff
@@ -37,6 +37,23 @@
 
 
 class generic_character(SFA_generic):
+
+    def construction(self):
```
tscrim, line 41:
```suggestion
class generic_character(SFA_generic):
    def construction(self):
```

src/sage/combinat/sf/character.py
```diff
@@ -151,6 +168,61 @@ def _b_power_k(self, k):
                                     for d in divisors(k))
 
 
+from sage.combinat.sf.sfa import SymmetricFunctionsFunctor
+class CharacterSymmetricFunctionsFunctor(SymmetricFunctionsFunctor):
```
tscrim, line 172: Every basis should come with a `_prefix` (since they are all subclasses of CFM). So I don't think this functor is necessary.

tscrim (changes requested):
There is one annoying thing: when we override `__eq__` we also need to override `__hash__` to keep things hashable. Perhaps we should take a middle ground to your other proposal. My proposal is we have `SymmetricFunctionsFunctor` that is an endofunctor on `CommutativeRings` with comparison of bases, but then a subclass `SymmetricFunctionsFixedElementsFunctor` (better name likely necessary) that also takes a fixed number of extra arguments for the comparison (and hash). This would then be the base class for, e.g., `MacdonaldSymmetricFunctionsFunctor` and reduce the number of `__eq__` and `__hash__` methods (maybe also `__init__`, but I do like the explicitness), and `.. WARNING::`s necessary. It also fits more mathematically what is going on. What do you think?

Also, we use American English spelling, so `initialise` -> `initialize`.

mantepse:
I am not completely sure whether I did what you had in mind, @tscrim.  In any case, there are no `*args`, as mentioned in https://github.com/sagemath/sage/pull/37220#issuecomment-2021288584, anymore.

To answer the question
> Alternatively, why have the, e.g., `DualBasisFunctor`?

In https://github.com/sagemath/sage/compare/develop...mantepse:sage:sf/construction-functional there are four functors:

* `SymmetricFunctionsFunctor` for bases that are created as `SymmetricFunctions(R).basis(*args)`
* `SymmetricFunctionsFamilyFunctor` for bases that are created as `SymmetricFunctions(R).family(*args).basis()`
* `DualBasisFunctor` for bases that are created as the dual of a basis
* `OrthotriangBasisFunctor`, for bases that are created by orthogonality and triangularity.

It seems to me that these four are quite distinct.

tscrim:
This is what I was thinking.

It seems to me like the last two are basically just the first one that can be passed from `construction` through the arguments.

I am thinking you are right with attaching a list of arguments to the functor with some generality, but I think we should assume it is non-empty.

mantepse:
> This is what I was thinking.
> 
> It seems to me like the last two are basically just the first one that can be passed from `construction` through the arguments.
> 
> I am thinking you are right with attaching a list of arguments to the functor with some generality, but I think we should assume it is non-empty.

Sorry, it's not clear to me what you mean.  I'm in a meeting now for something like half an hour, but maybe we could talk afterwards?

src/sage/combinat/sf/character.py
mantepse: The strange thing here is that `induced_trivial_character_basis.__init__` takes prefix as a required argument.  Maybe this could or should be made more uniform.  (My previous approach required that the arguments were provided exactly as needed for this `__init__`.)

PR #37687, opened 2024-03-28
provide a construction functor using a single functor for families

mantepse:
We enable sage to find common parents also when symmetric functions are involved. In particular, we provide a functorial construction that takes a commutative (coefficient) ring and produces a (commutative) ring of symmetric functions over this ring.

For Macdonald polynomials and other symmetric functions that involve parameters we cheat a bit: the categories should actually be commutative rings with distinguished elements (namely, the parameters). However, having these is very likely overkill.

This is an alternative to #37220, relying on individual construction functors instead of passing around strings, and #37686, which provides an individual functor for each family.

### DIFF
--- a/src/sage/combinat/sf/dual.py
+++ b/src/sage/combinat/sf/dual.py
@@ -292,7 +292,7 @@ class SymmetricFunctionAlgebra_dual(classical.SymmetricFunctionAlgebra_classical
         Return the name of the basis of ``self``.
 
         This is used for output and, for the classical bases of
-        symmetric functions, to connect this basis with Symmetrica.
+        symmetric functions, to connect this basis with :ref:`Symmetrica <spkg_symmetrica>`.
 
         EXAMPLES::
 
--- a/src/sage/combinat/sf/sfa.py
+++ b/src/sage/combinat/sf/sfa.py
@@ -641,7 +641,8 @@ class SymmetricFunctionsBases(Category_realization_of_parent):
                 sage: Sym = SymmetricFunctions(P)
                 sage: mj = Sym.macdonald().J()
                 sage: mj.corresponding_basis_over(Integers(13)['q','t'])
-                Symmetric Functions over Multivariate Polynomial Ring in q, t over Ring of integers modulo 13 in the Macdonald J basis
+                Symmetric Functions over Multivariate Polynomial Ring in q, t over 
+                 Ring of integers modulo 13 in the Macdonald J basis
 
             TESTS:
 
--- a/src/sage/combinat/sf/sfa.py
+++ b/src/sage/combinat/sf/sfa.py
@@ -641,7 +641,7 @@ class SymmetricFunctionsBases(Category_realization_of_parent):
                 sage: Sym = SymmetricFunctions(P)
                 sage: mj = Sym.macdonald().J()
                 sage: mj.corresponding_basis_over(Integers(13)['q','t'])
-                Symmetric Functions over Multivariate Polynomial Ring in q, t over 
+                Symmetric Functions over Multivariate Polynomial Ring in q, t over
                  Ring of integers modulo 13 in the Macdonald J basis
 
             TESTS:
--- a/src/sage/combinat/sf/witt.py
+++ b/src/sage/combinat/sf/witt.py
@@ -23,7 +23,6 @@ from . import multiplicative
 from sage.arith.misc import divisors
 from sage.combinat.integer_lists.invlex import IntegerListsLex
 from sage.combinat.partitions import ZS1_iterator
-from sage.matrix.constructor import matrix
 from sage.misc.cachefunc import cached_method
 
 


mantepse:
There is a problem with the idea for the Witt basis (using the code below).  Running the tests for `sf.py` gives:
```
sage -t --random-seed=200594915698172272573741126794711739352 src/sage/combinat/sf/sf.py
File "src/sage/combinat/sf/sf.py", line 970, in sage.combinat.sf.sf.SymmetricFunctions.witt
Failed example:
    SymmetricFunctions(QQ).witt(coerce_p=True)
Exception raised:
      File "/home/martin/sage/src/sage/doctest/forker.py", line 712, in _run
      File "/home/martin/sage/src/sage/doctest/forker.py", line 1147, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.combinat.sf.sf.SymmetricFunctions.witt[1]>", line 1, in <module>
        SymmetricFunctions(QQ).witt(coerce_p=True)
      File "/home/martin/sage/src/sage/combinat/sf/sf.py", line 976, in witt
        return witt.SymmetricFunctionAlgebra_witt(self, coerce_h=coerce_h, coerce_e=coerce_e, coerce_p=coerce_p)
      File "sage/misc/classcall_metaclass.pyx", line 320, in sage.misc.classcall_metaclass.ClasscallMetaclass.__call__
        return cls.classcall(cls, *args, **kwds)
      File "/home/martin/sage/src/sage/combinat/sf/witt.py", line 408, in __classcall__
        b._init_extra_()
      File "/home/martin/sage/src/sage/combinat/sf/witt.py", line 1075, in _init_extra_
        self._h.register_coercion(self._module_morphism(self._w_to_h_on_basis, codomain=self._h))
      File "sage/structure/parent.pyx", line 1607, in sage.structure.parent.Parent.register_coercion
        cpdef register_coercion(self, mor) noexcept:
      File "sage/structure/parent.pyx", line 1660, in sage.structure.parent.Parent.register_coercion
        assert not (self._coercions_used and D in self._coerce_from_hash and
    AssertionError: coercion from Symmetric Functions over Rational Field in the Witt basis to Symmetric Functions over Rational Field in the homogeneous basis already registered or discovered
```
I could not yet localize the problem further.

```diff
diff --git a/src/sage/combinat/sf/witt.py b/src/sage/combinat/sf/witt.py
index 2a22ccec58b..c52762b18b4 100644
--- a/src/sage/combinat/sf/witt.py
+++ b/src/sage/combinat/sf/witt.py
@@ -401,9 +401,14 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
         """
         Normalize the arguments.
         """
-        return super().__classcall__(cls, Sym, coerce_h, coerce_e, coerce_p)
-
-    def __init__(self, Sym, coerce_h, coerce_e, coerce_p):
+        b = super().__classcall__(cls, Sym)
+        b._coerce_h = coerce_h
+        b._coerce_e = coerce_e
+        b._coerce_p = coerce_p
+        b._init_extra_()
+        return b
+
+    def __init__(self, Sym):
         """
         Initialize ``self``.
 
@@ -413,9 +418,6 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
             sage: TestSuite(w).run(skip=['_test_associativity', '_test_distributivity', '_test_prod'])
             sage: TestSuite(w).run(elements = [w[1,1]+w[2], w[1]+2*w[1,1]])
         """
-        self._coerce_h = coerce_h
-        self._coerce_e = coerce_e
-        self._coerce_p = coerce_p
         multiplicative.SymmetricFunctionAlgebra_multiplicative.__init__(self, Sym, "Witt", 'w')
 
     def _precompute_cache(self, n, to_self_cache, from_self_cache, transition_matrices, inverse_transition_matrices, to_self_gen_function):
@@ -975,11 +977,18 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
         self._precompute_p(n)
         return self._p._from_dict(self._p_from_self_cache[lam])
 
-    def __init_extra__(self):
+    def _init_extra_(self):
         """
         Sets up caches for the transition maps to other bases, and registers
         them as coercions.
 
+        .. WARNING::
+
+            We do not use the generic `__init_extra__` protocol here,
+            because we do not want the specific choice of which
+            coercions to be set up be part of the identity of the
+            basis.
+
         EXAMPLES::
 
             sage: Sym = SymmetricFunctions(QQ) # indirect doctest
```

mantepse:
I incorporated your suggestions from #37686.

mantepse:
I did some debugging because of the coercion thing, and found the following doctest:
```
        sage: Sym = SymmetricFunctions(QQ)
        sage: p = Sym.p()
        sage: wh = Sym.w()
        sage: we = Sym.w(coerce_h=False, coerce_e=True)
        sage: wp = Sym.w(coerce_h=False, coerce_p=True)
        sage: all( p(wh(lam)) == p(we(lam)) == p(wp(lam)) for lam in Partitions(4) )
        True
        sage: all ( wh(p(lam)).monomial_coefficients()
        ....:       == we(p(lam)).monomial_coefficients()
        ....:       == wp(p(lam)).monomial_coefficients() for lam in Partitions(4) )
        True
```
So, apparently, it was indeed the intention that different settings of `coerce_h`, `coerce_e` and `coerce_p` yield *different* bases.  In particular, they would also have different caches then.

I have no idea how large these caches might become.  Should we remove the `h`-cache when we say `Sym.w(coerce_h=False, coerce_p=True)`?  I am guessing that this would be closest to current expectations.  Doing so, all Witt bases over a given ring will have the caching set by the last call.

However, calling `Sym.w()` without any arguments should then probably *not* switch to the `h`-cache, right?  In other words, the logic would be:
* `coerce_h is None and coerce_e is None and coerce_p is None`
    * no cache set yet: set `coerce_h=True`.
    * otherwise do nothing
* otherwise, remove old cache and set up new one.

tscrim:
I think what we should do is expand the caches based on the input (provided they are unpopulated). I am pretty sure they are not meant to be different bases. The point of that doctest is to check consistency through the caches (which were previously constructed separately). We probably need to update such doctests to test what they were meant to test.

For the issue at hand, we probably aren't being careful enough with its initialization and `__classcall__`. Mainly we probably shouldn't rerun `_init_extra_` by setting some sort of flag in the class or through examination of the caches. I will look at this tomorrow.

mantepse:
The reason for the failure ia that a different (less efficient) coercion is already discovered.

Thus, we just have to remove this coercion before registering the new one. Since there is only a single basis now, we *know* that a previously registered coercion cannot be one set up by the Witt class.

The only question is, whether we want to accumulate caches or not.  I think that not doing that gives greater flexibility at little cost. Setting up the caches or not is probably only an important decision if one wants to do seriously large computations. In this case it may be useful to be able to remove a cache.

tscrim:
Hmm...I probably would make using all the caches the default. I don't see a reason to have these extra complications.

@darijgr Do you have an opinion about the caches in the Witt basis? (I believe you wrote this, right?)

mantepse:
Here is what I had in mind (the doctests mentioned above would still need updating, and I'd rename `_init_extra_` to `_setup_caches`).
```
diff --git a/src/sage/combinat/sf/witt.py b/src/sage/combinat/sf/witt.py
index 2a22ccec58..c5d8ec9020 100644
--- a/src/sage/combinat/sf/witt.py
+++ b/src/sage/combinat/sf/witt.py
@@ -400,10 +400,28 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
     def __classcall__(cls, Sym, coerce_h=True, coerce_e=False, coerce_p=False):
         """
         Normalize the arguments.
-        """
-        return super().__classcall__(cls, Sym, coerce_h, coerce_e, coerce_p)
 
-    def __init__(self, Sym, coerce_h, coerce_e, coerce_p):
+        TESTS::
+
+            sage: R.<q, t> = QQ[]
+            sage: B1 = SymmetricFunctions(R).w()
+            sage: B2 = SymmetricFunctions(R).w(coerce_e=True)
+            sage: B1 is B2
+            True
+        """
+        b = super().__classcall__(cls, Sym)
+        if not coerce_h and not coerce_e and not coerce_p:
+            if not hasattr(b, "_friendly") or self._friendly is None:
+                coerce_h = True
+                b._init_extra_()
+        else:
+            b._coerce_h = coerce_h
+            b._coerce_e = coerce_e
+            b._coerce_p = coerce_p
+            b._init_extra_()
+        return b
+
+    def __init__(self, Sym):
         """
         Initialize ``self``.
 
@@ -413,9 +431,6 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
             sage: TestSuite(w).run(skip=['_test_associativity', '_test_distributivity', '_test_prod'])
             sage: TestSuite(w).run(elements = [w[1,1]+w[2], w[1]+2*w[1,1]])
         """
-        self._coerce_h = coerce_h
-        self._coerce_e = coerce_e
-        self._coerce_p = coerce_p
         multiplicative.SymmetricFunctionAlgebra_multiplicative.__init__(self, Sym, "Witt", 'w')
 
     def _precompute_cache(self, n, to_self_cache, from_self_cache, transition_matrices, inverse_transition_matrices, to_self_gen_function):
@@ -975,11 +990,18 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
         self._precompute_p(n)
         return self._p._from_dict(self._p_from_self_cache[lam])
 
-    def __init_extra__(self):
+    def _init_extra_(self):
         """
         Sets up caches for the transition maps to other bases, and registers
         them as coercions.
 
+        .. WARNING::
+
+            We do not use the generic `__init_extra__` protocol here,
+            because we do not want the specific choice of which
+            coercions to be set up be part of the identity of the
+            basis.
+
         EXAMPLES::
 
             sage: Sym = SymmetricFunctions(QQ) # indirect doctest
@@ -1030,13 +1052,16 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
             # the elements of the Witt basis with respect to the powersum basis
             self._p_inverse_transition_matrices = {}
 
+            self._remove_from_coerce_cache(self._p)
             self.register_coercion(self._p._module_morphism(self._p_to_w_on_basis, codomain=self))
             from sage.rings.rational_field import RationalField
             if self.base_ring().has_coerce_map_from(RationalField):
+                self._p._remove_from_coerce_cache(self)
                 self._p.register_coercion(self._module_morphism(self._w_to_p_on_basis, codomain=self._p))
                 self._friendly = self._p
             else:
                 # self._w_to_p_on_basis is a partial map at best
+                self._p._remove_from_coerce_cache(self)
                 self._p.register_conversion(self._module_morphism(self._w_to_p_on_basis, codomain=self._p))
                 if (not self._coerce_e) and (not self._coerce_h):
                     # ensure that self has coercion at least to one other basis,
@@ -1062,7 +1087,9 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
             # cache for transition matrices which contain the coordinates of
             # the elements of the Witt basis with respect to the homogeneous basis
             self._h_inverse_transition_matrices = {}
+            self._remove_from_coerce_cache(self._h)
             self.register_coercion(self._h._module_morphism(self._h_to_w_on_basis, codomain=self))
+            self._h._remove_from_coerce_cache(self)
             self._h.register_coercion(self._module_morphism(self._w_to_h_on_basis, codomain=self._h))
             if self._friendly is None:
                 self._friendly = self._h
@@ -1084,7 +1111,10 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
             # cache for transition matrices which contain the coordinates of
             # the elements of the Witt basis with respect to the elementary basis
             self._e_inverse_transition_matrices = {}
+            # a coercion (outside of this class) may have been registered already
+            self._remove_from_coerce_cache(self._e)
             self.register_coercion(self._e._module_morphism(self._e_to_w_on_basis, codomain=self))
+            self._e._remove_from_coerce_cache(self)
             self._e.register_coercion(self._module_morphism(self._w_to_e_on_basis, codomain=self._e))
             if self._friendly is None:
                 self._friendly = self._e
```

mantepse:
And, of course, I should not delete a cache which is requested, just to recreate it again.

mantepse:
I just realized that I am not removing the caches...

tscrim:
I think having these options for the different caches are not so useful. I would probably rewrite the entire coercion as `@cached_method` for the $p, h, e, \to w$ and vice versa[^1], and let the multiplication in the other bases handle the rest (without caching). Or we then cache the products as well. In either case, I think the caching should be handled by `@cached_method`s, rather than manually, and with no options for (dis)allowing the cache. (This also gets rid of inverting a potentially big matrix even for relatively small degree.)

[^1]: Or do we not know enough explicitly to do $w_n = \sum_{\lambda \vdash n} c_n^{\lambda} h_{\lambda}$ (or similar)? If not, then perhaps we can use the inverse of a triangular module morphism to compute this without caching and inverting the full matrix.

darijgr:
Sorry, this is not a changeset I can easily understand...
Is the question at hand about whether the caches should be on the basis or on the Sym? Or whether they should be on the general class or on the instance corresponding to a given base ring?
And yes, I don't know of any explicit formula for `w_n` in terms of the h's. The easiest way to compute it is to recursively solve the generating-function equation `(1 - w_1 t^1) (1 - w_2 t^2) (1 - w_3 t^3) ... = (1 - x_1 t) (1 - x_2 t) (1 - x_3 t) ... = \sum_i (-1)^i e_i t^i` for `w_1, w_2, w_3, ...`. Do you see a simple way to do that? Maybe a cached function on the level of the module file, just like the a's and b's in the symmetric group algebra?

tscrim:
@darijgr The main question is should the Witt basis even have input options (which if not the default, gives distinct bases of Sym) of having caches to/from the $e,h,p$ bases? Although asking if the cache should be on Sym is an interesting twist on it, but not quite the same question I think.

Actually, given the recursive method of solving for $w_n$ in terms of $w_k$ for $k < n$ should give a much more efficient way of computing $w_n$. The caching for $w_n$ (using `@cached_method`) would be good for this as well.

darijgr:
I'm afraid my memory here is useless or worse. If the new code doesn't break any tests and doesn't slow things down, it's good as far as I'm concerned. Sorry for not being more useful!

tscrim:
@darijgr No problem.

@mantepse Then shall we just remove the inputs and always use the cached versions, at least for now? I will see if I can think of a quick way to recursively solve for $w_n$ in terms of the $e$ and $h$ basis.

mantepse:
> @mantepse Then shall we just remove the inputs and always use the cached versions, at least for now? I will see if I can think of a quick way to recursively solve for wn in terms of the e and h basis.

I am all for it.  So, what you mean is *effectively*
```
diff --git a/src/sage/combinat/sf/witt.py b/src/sage/combinat/sf/witt.py
index 2a22ccec58..e33d2aba1c 100644
--- a/src/sage/combinat/sf/witt.py
+++ b/src/sage/combinat/sf/witt.py
@@ -397,13 +397,13 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
         True
     """
     @staticmethod
-    def __classcall__(cls, Sym, coerce_h=True, coerce_e=False, coerce_p=False):
+    def __classcall__(cls, Sym, coerce_h=True, coerce_e=True, coerce_p=True):
         """
         Normalize the arguments.
         """
-        return super().__classcall__(cls, Sym, coerce_h, coerce_e, coerce_p)
+        return super().__classcall__(cls, Sym)
 
-    def __init__(self, Sym, coerce_h, coerce_e, coerce_p):
+    def __init__(self, Sym):
         """
         Initialize ``self``.
 
@@ -413,9 +413,9 @@ class SymmetricFunctionAlgebra_witt(multiplicative.SymmetricFunctionAlgebra_mult
             sage: TestSuite(w).run(skip=['_test_associativity', '_test_distributivity', '_test_prod'])
             sage: TestSuite(w).run(elements = [w[1,1]+w[2], w[1]+2*w[1,1]])
         """
-        self._coerce_h = coerce_h
-        self._coerce_e = coerce_e
-        self._coerce_p = coerce_p
+        self._coerce_h = True
+        self._coerce_e = True
+        self._coerce_p = True
         multiplicative.SymmetricFunctionAlgebra_multiplicative.__init__(self, Sym, "Witt", 'w')
 
     def _precompute_cache(self, n, to_self_cache, from_self_cache, transition_matrices, inverse_transition_matrices, to_self_gen_function):
```
except that I guess I should get rid of the three `_coerce_*` attributes throughout, and remove the doctests testing them?  (Also, if `self._coerce_p` can only be used if the base ring contains `QQ`, I think).

Do we have to deprecate?  Or should we just silently ignore the arguments?

tscrim:
Yes, that's what I'm thinking. I think we are safe to just remove them, but perhaps we should be safe and put a deprecation in the `__classcall_private__` that just eats them and doesn't pass them along (set them to `None` as the default and check against that for the deprecation).

tscrim:
Okay, I think we can easily do the recursion for the $w \leftrightarrow e$ case. As Darij said, we have the generating function identity:
$$\prod_{n=1}^{\infty} (1 - w_n t^n) = \sum_{n=0}^{\infty} (-1)^n e_n t^n.$$
Hence, for fixed degree $n$ we have
$$\sum_{\lambda \vdash_s n} (-1)^{\ell(\lambda)} w_{\lambda} = (-1)^n e_n,$$
where the sum is over strict partitions, which clearly gives us $e_n$ in terms for $w_{\lambda}$. However, since this is a multiplicative basis, we can write
$$w_n = e_n - \sum_{\lambda \vdash_s n; \ell(\lambda) > 1} (-1)^{n-\ell(\lambda)} \prod_{i=1}^{\ell(\lambda)} w_{\lambda_i}.$$
It's not super pretty, but it means we don't have to invert the full matrix and can take advantage of the cache. (The smartest way to do this is would be to multiply by the smallest part of the partition.) We could only compute what we need at any given moment (which is not necessarily the full cache; say `e(w[2,2,2,2,2])` would only need $w_2$ and recurse on $w_{2^k}$).

The case for $w \leftrightarrow h$ is similar.

Since the Witt case is growing, I think we should split that off as a dependency. Do you understand the recursion I am talking about? If it isn't quite clear, I can do it. Maybe tonight before I fall asleep, but otherwise tomorrow.

Oh, here is something else to fix:
```
sage: Sym = SymmetricFunctions(QQ)
sage: e = Sym.e()
sage: e.completion() is LazySymmetricFunctions(e)
False
```

mantepse:
> Since the Witt case is growing, I think we should split that off as a dependency. Do you understand the recursion I am talking about? If it isn't quite clear, I can do it. Maybe tonight before I fall asleep, but otherwise tomorrow.

I think it will be much easier for you to do it than for me.  I'd rather put my energy into the `subs` of the `InfinitePolynomialRing`.
 
> Oh, here is something else to fix:
> 
> ```
> sage: Sym = SymmetricFunctions(QQ)
> sage: e = Sym.e()
> sage: e.completion() is LazySymmetricFunctions(e)
> False
> ```
Furthermore, `e.completion()` does not have plethysm - the `Element` is wrong.

tscrim:
The Witt coercion changes are now in #37736.

tscrim (approved):
LGTM. Thank you.

mantepse:
I'm sorry, this was modified by accident.  The last four commits are only cosmetics, and would also be contained in #37033.

PR #37688, opened 2024-03-28
typo

DennisYurichev:
Small typo

### DIFF
--- a/src/sage/logic/logictable.py
+++ b/src/sage/logic/logictable.py
@@ -62,7 +62,7 @@ Create a truth table of a boolean formula::
     True   True   False  True
     True   True   True   True
 
-Get the latex code for a truth table::
+Get the LaTeX code for a truth table::
 
     sage: latex(s.truthtable(5,11))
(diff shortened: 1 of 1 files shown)


src/sage/logic/logictable.py
```diff
@@ -62,7 +62,7 @@
     True   True   False  True
     True   True   True   True
 
-Get the letex code for a truth table::
+Get the latex code for a truth table::
```
mantepse: LaTeX seems more common in the documentation, I'd suggest to use that instead.

Found using
```grep -r --include=*.{py,pyx} --color -nH --null -e "[^.(_]LaTeX[^_('\`\"=]" *```
and
```grep -r --include=*.{py,pyx} --color -nH --null -e "[^.(_]latex[^_('\`\"=]" *```

DennisYurichev:
So? Should I do something? Please advise.

grhkm21:
Yes, replace "latex" with "LaTex".

DennisYurichev:
OK, did so: https://github.com/sagemath/sage/pull/37688/commits/3b137b50a0112ef56ac7fa63c64cb0a1c5671db5

mantepse (approved):
Thank you!

PR #37690, opened 2024-03-29
using pathlib.Path to handle attached files

fchapoton:
Let us use `pathlib.Path` to handle the attached files. This make the code cleaner.

### :memo: Checklist


fchapoton:
there seems to be an issue in conda :

`ValueError: file '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpysx6hdf6.py'`

versus

`['/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpysx6hdf6.py']`

which has 'private' in front. Any suggestion ?

fchapoton:
I just learned that

_In Mac OS X, /var is a symlink to /private/var._

src/sage/repl/attach.py
```diff
@@ -385,13 +396,13 @@ def add_attached_file(filename):
         []
     """
     sage.repl.inputhook.install()
-    fpath = os.path.abspath(filename)
-    attached[fpath] = os.path.getmtime(fpath)
+    fpath = Path(filename).resolve()
+    attached[fpath] = fpath.stat().st_mtime
```
kwankyu, line 399: Couldn't we remove `resolve()` from line 399 to fix `/var` vs `/private/var` issue? Do we need `resolve()`?

src/sage/misc/remote_file.py
```diff
@@ -17,8 +16,8 @@ def get_remote_file(filename, verbose=True):
 
     OUTPUT:
 
-    creates a file in the temp directory and returns the absolute path
-    to that file.
+    This creates a file in the temp directory and returns the absolute path
+    to that file as a ``Path`` object.
```
kwankyu, line 21: We can use ``` :class:`Path` ``` everywhere. Then we have a nice link.

src/sage/repl/attach.py
fchapoton: The idea is to store the attached files in fully resolved form, <s>as before</s>. The issues we have come from the bad behaviour of "tmp_filename" that should return Path objects in resolved form instead of strings. This should be changed later, but not here.

src/sage/misc/remote_file.py
fchapoton: done, let's check if this works

src/sage/repl/attach.py
kwankyu: > The idea is to store the attached files in fully resolved form, as before. 

Does `os.path.abspath(filename)` give a fully resolved form, or is `filename` already a fully resolved form?

kwankyu:
```
sage: t= tmp_filename(ext='.py')
sage: t
'/var/folders/ym/nw_hc8550mgc0v4r43lvswqh0000gn/T/tmpsnhsu7u8/tmp_3lmn7es3.py'
```
It seems that `tmp_filename` does not give a fully resolved form, before...

kwankyu: I am on mac.

src/sage/repl/load.py
```diff
@@ -28,7 +28,7 @@ def is_loadable_filename(filename):
 
     INPUT:
 
-    - ``filename`` -- a string
+    - ``filename`` -- a string or Path object
```
kwankyu, line 32: At least
```
``Path``
```
or better
```
:class:`Path`
```

src/sage/repl/load.py
```diff
@@ -86,12 +90,12 @@ def load(filename, globals, attach=False):
 
     INPUT:
 
-    - ``filename`` -- a string denoting a filename or URL.
+    - ``filename`` -- a string (denoting a filename or URL) or a Path object
```
kwankyu, line 94: The same here.

src/sage/repl/attach.py
kwankyu: > The idea is to store the attached files in fully resolved form, as before.

Before this PR:
```
sage: fn
'/var/folders/td/fw1q9ljs311ggyph77rs53_40000gn/T/tmp5hz2j1pi/tmp_asix90ys.py'
sage: attached_files()
['/var/folders/td/fw1q9ljs311ggyph77rs53_40000gn/T/tmp5hz2j1pi/tmp_asix90ys.py']
```
So the attached_files are not stored in fully resolved form. Right?

Then what is the reason (or benefits) of storing the attached files in fully resolved form, with this PR?

fchapoton: We should at least store an absolute path, and not a relative one, as the user may change the current directory at any time. It seems to me that it is safe and a good idea to use `Path.resolve` to have the correct path.

Another possibility would be, according to

https://docs.python.org/fr/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

to use instead `Path.absolute()`, which should behave more like the existing code. Would you prefer that ?

kwankyu: I think we should do either

(1) use `Path.absolute()` and keep current `tmp_filename()`

or

(2) use `Path.resolve()` and fix (here) `tmp_filename()` to use fully resolved filename

instead of "fixing" doctests for disconcerting `Path` and `tmp_filename()`.

(1) seems easier.

src/sage/repl/attach.py
```diff
@@ -469,22 +486,24 @@ def detach(filename):
...
-        if not os.path.isabs(fpath):
+        fpath = filename.expanduser()
+        if not fpath.is_absolute():
             for path in load_attach_path():
-                epath = os.path.expanduser(path)
-                fpath = os.path.join(epath, filename)
-                fpath = os.path.abspath(fpath)
+                fpath = path.expanduser() / filename
+                fpath = fpath.resolve()
                 if fpath in attached:
```
kwankyu, line 498: Here still `resolve()`?

src/sage/repl/attach.py
```diff
@@ -469,22 +486,24 @@ def detach(filename):
...
                     break
+        r_fpath = fpath.resolve()
         if fpath in attached:
```
kwankyu, line 501: Here too?

kwankyu:
Finally, let's remove ugly doctests:
```diff
diff --git a/src/sage/misc/reset.pyx b/src/sage/misc/reset.pyx
index ba6a2be5f6e..e8247bb8f89 100644
--- a/src/sage/misc/reset.pyx
+++ b/src/sage/misc/reset.pyx
@@ -44,7 +44,7 @@ def reset(vars=None, attached=False):
         sage: attach(fn)
         sage: af = attached_files(); len(af)
         1
-        sage: fn in af[0]
+        sage: af == [fn]
         True
         sage: reset()
         sage: af = attached_files(); len(af)
diff --git a/src/sage/repl/attach.py b/src/sage/repl/attach.py
index aef8b003d34..00a171081ea 100644
--- a/src/sage/repl/attach.py
+++ b/src/sage/repl/attach.py
@@ -341,9 +341,9 @@ def attach(*files):
         hi there xxx
         sage: af = attached_files(); len(af)
         2
-        sage: any(t1 in f for f in af)
+        sage: t1 in af
         True
-        sage: any(t2 in f for f in af)
+        sage: t2 in af
         True
 
     .. SEEALSO::
diff --git a/src/sage/repl/load.py b/src/sage/repl/load.py
index f7a20d77dff..2d93214cc12 100644
--- a/src/sage/repl/load.py
+++ b/src/sage/repl/load.py
@@ -167,7 +167,7 @@ def load(filename, globals, attach=False):
         ....:     _ = f.write("print('hello world')")
         sage: sage.repl.load.load(t, globals(), attach=True)
         hello world
-        sage: any(t in f for f in attached_files())
+        sage: t in attached_files()
         True
 
     You cannot attach remote URLs (yet)::
@@ -300,15 +300,13 @@ def load_wrap(filename, attach=False):
 
     INPUT:
 
-    - ``filename`` - a string or Path object; the argument
+    - ``filename`` -- a string or :class:`Path` object; the argument
       to the load or attach command
 
-    - ``attach`` - a boolean (default: ``False``); whether to attach
+    - ``attach`` -- a boolean (default: ``False``); whether to attach
       ``filename``, instead of loading it
 
-    OUTPUT:
-
-    - a string
+    OUTPUT: a string
 
     EXAMPLES::
 
diff --git a/src/sage/repl/preparse.py b/src/sage/repl/preparse.py
index 59f09ff1a18..c74174ca40d 100644
--- a/src/sage/repl/preparse.py
+++ b/src/sage/repl/preparse.py
@@ -2192,7 +2192,7 @@ def handle_encoding_declaration(contents, out):
             return '\n'.join(lines[:num] + lines[(num + 1):])
 
     # If we did not find any encoding hints, use explicit utf-8.
-    # According to PEP 3120, this could be omitted
+    # According to PEP 3120, this could be omitted.
     out.write("# -*- coding: utf-8 -*-\n")
     return contents
```

kwankyu (approved):
Thanks.

LGTM.

PR #37607, opened 2024-03-14
src/sage/matrix: Doctest cosmetics

mkoeppe:
Standard reformatting of doctests and their outputs

Split out from #35095 

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37692, opened 2024-03-29
`matrix`, `Graph.incidence_matrix`, `LinearMatroid.representation`: Support constructing `Hom(CombinatorialFreeModule)` elements

mkoeppe:
We use morphisms of `CombinatorialFreeModule`s (each of which has a distinguished finite or enumerated basis indexed by arbitrary objects) as matrices whose rows and columns are indexed by arbitrary objects (`row_keys`, `column_keys`).

Example:
```
        sage: M = matrix([[1,2,3], [4,5,6]],
        ....:            column_keys=['a','b','c'], row_keys=['u','v']); M
        Generic morphism:
          From: Free module generated by {'a', 'b', 'c'} over Integer Ring
          To:   Free module generated by {'u', 'v'} over Integer Ring
```

Example application done here on the PR: The incidence matrix of a graph or digraph. Returning it as a morphism instead of a matrix has the benefit of keeping the vertices and edges with the result. This new behavior is activated by special values for the existing parameters `vertices` and `edges`.
```
            sage: D12 = posets.DivisorLattice(12).hasse_diagram()
            sage: phi_VE = D12.incidence_matrix(vertices=True, edges=True); phi_VE
            Generic morphism:
              From: Free module generated by
                      {(1, 2), (1, 3), (2, 4), (2, 6), (3, 6), (4, 12), (6, 12)}
                    over Integer Ring
              To:   Free module generated by {1, 2, 3, 4, 6, 12} over Integer Ring
            sage: print(phi_VE._unicode_art_matrix())
                         (1, 2)  (1, 3)  (2, 4)  (2, 6)  (3, 6) (4, 12) (6, 12)
                      1⎛     -1      -1       0       0       0       0       0⎞
                      2⎜      1       0      -1      -1       0       0       0⎟
                      3⎜      0       1       0       0      -1       0       0⎟
                      4⎜      0       0       1       0       0      -1       0⎟
                      6⎜      0       0       0       1       1       0      -1⎟
                     12⎝      0       0       0       0       0       1       1⎠
```

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37607 
- Depends on #37514
- Depends on #37606 
- Depends on #37646


gmou3:
It looks ok. I am thinking however that it may complicate the code. E.g., in `linear_matroid.pyx`, do we gain something by having the `morphism` option internally coded?

Another small comment: in `action.pyx` there appear to be a few docstring formatting inconsistencies. Lines (30, 92), 141-144, 199.

mkoeppe:
> in `action.pyx` there appear to be a few docstring formatting inconsistencies. Lines (30, 92), 141-144, 199.

Thanks; these formatting changes are coming from the dependency https://github.com/sagemath/sage/pull/37607. I agree that some further improvements are possible to make the formatting more consistent, but I don't want to do it on this PR.

mkoeppe:
> I am thinking however that it may complicate the code. E.g., in `linear_matroid.pyx`, do we gain something by having the `morphism` option internally coded?

I don't think the code got more complicated. (The description of the options did get more complicated.) What is gained is the improved connection to the central structure using which algebraic combinatorics is expressed in Sage, the `CombinatorialFreeModule`.

mkoeppe:
@gmou3 By the way, what's still missing is the implementation of the roundtrip:
```
sage: M = matroids.catalog.Fano()
sage: A = M.representation(order=True)
Generic morphism:
  From: Free module generated by {'a', 'b', 'c', 'd', 'e', 'f', 'g'} over Finite Field of size 2
  To:   Free module generated by {0, 1, 2} over Finite Field of size 2
sage: Matroid(A)  # error
```

gmou3:
> > I am thinking however that it may complicate the code. E.g., in `linear_matroid.pyx`, do we gain something by having the `morphism` option internally coded?
> 
> I don't think the code got more complicated. (The description of the options did get more complicated.) What is gained is the improved connection to the central structure using which algebraic combinatorics is expressed in Sage, the `CombinatorialFreeModule`.

Ok. This is the more general comment I would raise, that things appear to be getting a bit baroque. Also in `graph`, `matrix`. But you can judge this better than me.

> @gmou3 By the way, what's still missing is the implementation of the roundtrip:
> 
> ```
> sage: M = matroids.catalog.Fano()
> sage: A = M.representation(order=True)
> Generic morphism:
>   From: Free module generated by {'a', 'b', 'c', 'd', 'e', 'f', 'g'} over Finite Field of size 2
>   To:   Free module generated by {0, 1, 2} over Finite Field of size 2
> sage: Matroid(A)  # error
> ```

I could fix that. Similarly, with `A = M.representation(labels=true)` we get an error.

mkoeppe:
> the more general comment I would raise, that things appear to be getting a bit baroque. Also in `graph`, `matrix`.

That's definitely a valid concern that is also on my radar. But it has to be weighed against other factors such as namespace management and discoverability. On the latter, for example we could make the `representation` or `incidence_matrix` methods' arguments less baroque by just creating separate methods `representation_morphism` and `incidence_morphism`. But I argue that this would introduce terminology that reduces discoverability because the higher-level mathematical language creates a barrier. I want users who expect a matrix just read the documentation of the method and see from the examples that "this can return things like matrices where rows and columns are labeled".

mkoeppe:
Lint and Build&Test failures are unrelated.

gmou3 (approved):
Fine by me. I'll play around with adding the matroid morphism input after this is merged.

src/sage/matrix/constructor.pyx
```diff
@@ -78,16 +83,15 @@ def matrix(*args, **kwds):
...
 
     - ``immutable`` -- (boolean) make the matrix immutable. By default,
       the output matrix is mutable.
 
 
-    OUTPUT:
-
-    a matrix
+    OUTPUT: a matrix or, more generally, a homomorphism between free
+    modules.
```
gmou3:
```suggestion
    modules
```

src/sage/matrix/args.pyx
```diff
@@ -714,13 +738,67 @@ cdef class MatrixArgs:
...
 
+    cpdef element(self, bint immutable=False) noexcept:
+        r"""
+        Return the matrix or morphism.
+
+        INPUT:
+
+        - ``immutable`` -- boolean; if ``True``, the result will be immutable.
+
+        OUTPUT: an element of ``self.space``.
```
gmou3:
```suggestion
        OUTPUT: an element of ``self.space``
```

src/sage/matrix/matrix_space.py
```diff
@@ -434,9 +434,73 @@ def get_matrix_class(R, nrows, ncols, sparse, implementation):
...
+      - ``'m4ri'`` -- for characteristic 2 using the :ref:`spkg_m4ri` library
+
+      - ``'linbox-float'`` -- for integer mod rings up to `2^8 = 256`
+
+      - ``'linbox-double'`` -- for integer mod rings up to
+        `floor(2^26*sqrt(2) + 1/2) = 94906266`
+
+      - ``'numpy'`` -- for real and complex floating point numbers
+
+    OUTPUT: a matrix space or, more generally, a homspace between free modules.
```
gmou3:
```suggestion
    OUTPUT: a matrix space or, more generally, a homspace between free modules
```

gmou3 (commented):
As per the Developer Guide.

mkoeppe:
Merged the latest version of the dependency https://github.com/sagemath/sage/pull/37514.

Any remaining concerns, or may I set to "positive review"?

gmou3:
None from me.

mkoeppe:
Thanks!

vbraun:
doesn't work, ci fails

mkoeppe:
All tests pass. Failure of "test modularized distributions" is unrelated.

PR #37695, opened 2024-03-29
Matrix Constructor

Veeransh14:
Updating imports: I used Sage's Cython-compatible format to ensure the code works well with Sage. 
Adding type declarations: I made sure that variables and function arguments have their types clearly specified, making the code easier to understand and less error-prone. 
Implementing method bodies: I completed the code for all methods so that each one now does what it's supposed to, making the class more functional and reliable. 
Adding docstrings: I wrote clear documentation for each method, explaining what it does and how it should be used. This helps users understand and use the class more effectively.


mkoeppe:
Check your branch

tscrim:
Also be aware that this can cause slowdowns as well because of the extra type checking. It also reduces the flexibility as we cannot do ducktyping. Both of these will need to be checked once the branch is clean.

tscrim:
I am going to close this PR as it doesn't have a clear objective and the branch is foobar. If @Veeransh14 pushes a clean and correct branch, then we can reopen this.

PR #37696, opened 2024-03-29
Remove trailing spaces in latex representations.

Eloitor:
While rendering elliptic curves in markdown, I found that rendering the following f-string didn't work as expected:
```
f"${latex(EllipticCurve("38.b1"))}$"
```

This is because the latex representation has a trailing space.
Using `rg -U -A 2 "sage: .*_latex_.*\\n.* '.* '"` I found another object that has the same problem.
Another reason for this PR is to have more consistency in the latex representations.

Let me know if I missed some other instance.

### :memo: Checklist


tscrim:
Trailing whitespace is not a problem in latex and can serve a purpose for readability. I think it is the responsibility of anything using the latex code to strip trailing whitespace as necessary. So I am -1 on this change.

PR #37699, opened 2024-03-30
Add `approximate_closest_vector` to `IntegerLattice`, using Babai's nearest plane algorithm

TheBlupper:
This PR implements Babai's nearest plane algorithm to find vectors close to a given vector in an integer lattice.

I would appreciate if someone could verify the bound stated in the docstring, it is inferred from Babai's original paper but parameterized to use the `delta` parameter of the LLL algorithm.

Ideas for improvements:
 - The doctest is currently the same as for `closest_vector`, this could be expanded
 - The `delta` parameter for LLL needs to be kept track of by the user. How "much" the basis has been reduced (i.e to how high a `delta` value) could be kept track of by the class and if a further reduction is needed for some given bound then it could be performed.

### :memo: Checklist


src/sage/modules/free_module_integer.py
```diff
@@ -804,3 +804,41 @@ def CVPP_2V(t, V, voronoi_cell):
             t_new = t_new - CVPP_2V(t_new, V_scaled, ZZ(2 ** (i - 1)) * voronoi_cell)
             i -= 1
         return t - t_new
+
+    def approximate_closest_vector(self, t):
+        r"""
+        Compute a vector `w` such that `|t-w| < (\frac{1}{\delta-\frac{1}{4}})^{d/2} |t-u|`
+        where `u` is the closest lattice point to `t`, `\delta` is the LLL reduction parameter
+        used to reduce the lattice basis, and `d` is the dimension of the lattice.
```
grhkm21: The word part of the docstring should be wrapped to 80 characters, c.f. [here](https://github.com/sagemath/sage/pull/37118#pullrequestreview-1897378170) for example.

src/sage/modules/free_module_integer.py
```diff
@@ -804,3 +804,41 @@ def CVPP_2V(t, V, voronoi_cell):
...
+
+            sage: from sage.modules.free_module_integer import IntegerLattice
+            sage: L = IntegerLattice([[1, 0], [0, 1]])
+            sage: L.approximate_closest_vector((-6, 5/3))
+            (-6, 2)
+
+        ALGORITHM:
+
+        Uses the algorithm from [Bab86]_.
+
```
grhkm21:
```suggestion
```

src/sage/modules/free_module_integer.py
```diff
@@ -804,3 +804,41 @@ def CVPP_2V(t, V, voronoi_cell):
             t_new = t_new - CVPP_2V(t_new, V_scaled, ZZ(2 ** (i - 1)) * voronoi_cell)
             i -= 1
         return t - t_new
+
+    def approximate_closest_vector(self, t):
```
grhkm21: (As suggested in your "ideas for improvement") I think `delta` should be a user-specified argument. You might also want to add a note that this method changes the lattice basis (`self.LLL()` seems to modify `self`).

grhkm21: In fact, why not just add `*args, **kwds` that is passed on to the `LLL` call? That will include e.g. `delta` and `algorithm`. Though that will probably complicate the bound in the docstring below.

grhkm21 (changes requested):
More doctest is optional IMO

src/sage/modules/free_module_integer.py
TheBlupper: We don't want to reduce the lattice every time we call this method, so the way I envision it is that instead of checking `_basis_is_LLL_reduced` it keeps track of the delta value the basis has been reduced with. That becomes a little problematic if the basis has been reduced with BKZ or HKZ since we don't have an analog to delta (i think?). How do you suggest we deal with that?

TheBlupper: HKZ can reasonably be defined as using `delta=1` (though the given bound will likely be unnecessarily strict) or even a `delta` >1. BKZ does take a `delta` parameter but I have no experience using it, the block size should also affect the reduced basis and I'm not sure how to handle that.

TheBlupper: I'm leaning towards only reducing if the basis hasn't already been reduced, similar to how it is now, but we also allow the user to pass on arguments to LLL. These will be ignored if the basis is already LLL reduced.

Alternatively we only reduce if the user specifies reduction params.

*Alternatively* we simply leave it to the user to handle, potentially raising a warning if the basis hasn't been reduced at all.

grhkm21: What about the user can call the function with any `*args, **kwds` and with an `algorithm="*"` argument. If the lattice basis hasn't been reduced by any algorithm yet (so there should be a `.is_reduced` flag?), then `algorithm` argument is required. Otherwise, it's optional

TheBlupper: There are cases when an already reduced basis is passed to `IntegerLattice`, I think this might introduce unnecessary complexity.

I now realize this should be the perfect use-case for `is_LLL_reduced`. It should run fairly quickly and accepts a `delta`-parameter. Then we could have the function accept a `delta` which has some default value (3/4 or 0.99), it checks if the basis is reduced to that degree and otherwise it runs LLL to make sure that it is. Sound good?

grhkm21: Yes that works fine, I suggested mine since you wanted to handle BKZ too but I suppose it's output will also be LLL-reduced (I think)?

TheBlupper: I think it is a fairly universal measure of basis "goodness"

TheBlupper:
Fixed some typos/inaccuracies in the docstring of the LLL method as well

src/sage/modules/free_module_integer.py
```diff
@@ -804,3 +805,59 @@ def CVPP_2V(t, V, voronoi_cell):
...
+        OUTPUT:
+
+        The vector `w` described above.
+
+        EXAMPLES::
+
+            sage: from sage.modules.free_module_integer import IntegerLattice
+            sage: L = IntegerLattice([[1, 0], [0, 1]])
+            sage: L.approximate_closest_vector((-6, 5/3))
+            (-6, 2)
```
grhkm21, line 844: Add tests for the `delta` etc arguments

src/sage/modules/free_module_integer.py
```diff
@@ -804,3 +805,59 @@ def CVPP_2V(t, V, voronoi_cell):
...
+            self.LLL(*args, delta=delta, **kwargs)
+
+        B = self._reduced_basis
+        G = B.gram_schmidt()[0]
+        t = vector(t)
+
+        b = t
+        for i in reversed(range(G.nrows())):
+            b -= B[i] * ((b * G[i]) / (G[i] * G[i])).round("even")
+        return (t - b).change_ring(ZZ)
```
grhkm21: Also I *think* adding `.babai` as an alias (so just add a line `babai = approximate_shortest_vector`) would be a good idea.

grhkm21 (changes requested):
add tests

TheBlupper:
@grhkm21 thanks for the feedback, all concerns raised should be fixed now, think you could take another look?

src/sage/modules/free_module_integer.py
```diff
@@ -805,3 +806,73 @@ def CVPP_2V(t, V, voronoi_cell):
...
```
grhkm21, line 877:
```suggestion
        return (t - b).change_ring(ZZ)
        
```
Sorry I missed this. There should be an extra line. Otherwise it is good. Thanks :)

grhkm21 (changes requested):

Issue #33380, opened 2022-02-19
matrix constructor: Accept "base_ring" keyword


mkoeppe:
<div id="comment:0"></div>

Right now, if not passed as a positional argument, it has to be provided as `ring`.

```
sage: matrix(5, base_ring=QQ)
TypeError: __classcall__() got multiple values for argument 'base_ring'
sage: matrix(5, ring=QQ)
[0 0 0 0 0]
[0 0 0 0 0]
[0 0 0 0 0]
[0 0 0 0 0]
[0 0 0 0 0]
```

CC:  @orlitzky @dcoudert @tscrim

Component: **linear algebra**

_Issue created by migration from https://trac.sagemath.org/ticket/33380_

SandwichGouda:
Hello ! 
I am motivated to tackle this issue.
Can I be assigned to it ?

mkoeppe:
Fixed in #37721.

PR #37700, opened 2024-03-30
Added "base_ring" keyword to the matrix constructor

SandwichGouda:
The ``matrix`` constructor only accepts ``ring`` as (positional) argument. 
It is now possible to specify the base ring of matrices using the ``base_ring`` (keyword only) argument.

This solves @mkoeppe's #33380 issue.

The first commit adds some code to the ``MatrixArgs`` class defined in ``src/sage/matrix/constructor.pyx``.
The second commit adds appropriate documentation (with tests and examples) accordingly. 

If the user specifies both arguments (``ring`` and ``base_ring``), everything's fine as long as their values are the same.
(The value of ``base_ring`` is then discarded.)
If the user specifies only one or none of them, no problem. 
If the user specifies both ``ring`` and ``base_ring`` arguments with different values, 
``TypeError: positional argument ring and keyword argument base_ring have different values`` is raised. 

Note : I could have chosen to prioritize ``ring`` over ``base_ring`` if case both arguments are given.
I have chosen not to since this is not the expected usage of these arguments. 
If a user mistakenly specifies both arguments with different values, it is most likely an error, and they better
be informed directly, preventing unnecessary debugging. 

This is my first contribution to sage, I hope it works out !

### :memo: Checklist


mantepse:
Why not simply rename the parameter (along with deprecation)? Also, what's wrong with ring in the first place?

SandwichGouda:
Nothing's wrong with ``ring``, but @mkoeppe asked to implement this in #33380 so I tackled it :)

I'm not sure I understand what you mean exactly my "renaming the parameter along with deprecation". :O
The idea is to allow the user to use _either_  ``ring`` or ``base_ring``. If I rename the parameter ``ring`` to ``base_ring`` in ``__init__``, users wouldn't be able to use ``ring`` anymore (which would probably be a huge problem, as many existring sage scripts wouldn't work anymore). This is probably not what you meant but then I don't get it.

mantepse:
@mkoeppe, could you please clarify?  I am assuming that this is about unification - class factories should take `base_ring` as argument, not `ring`, right?

SandwichGouda:
I think @mkoeppe is okay with this because he accepted the pull request.
Unfortunately there were errors so I had to push another version (I hope it'll work out this time).

mantepse:
> I think @mkoeppe is okay with this because he accepted the pull request.

I don't understand what you are saying here.  Could you please clarify?

SandwichGouda:
I'm just saying he's okay with my pull request and that he's not againt adding my contribution to the code source.
My PR is currently awaiting approval ("5 workflows awaiting approval") so if you have the possibility to approve it it would be cool

mantepse:
> I'm just saying he's okay with my pull request and that he's not againt adding my contribution to the code source. My PR is currently awaiting approval ("5 workflows awaiting approval") so if you have the possibility to approve it it would be cool

I am actually against inclusion, if there is no explanation for why it is needed.  Moreover, I am quite sure that ad-hoc aliasing of argument names is not a good idea.

mkoeppe:
> @mkoeppe, could you please clarify? I am assuming that this is about unification - class factories should take `base_ring` as argument, not `ring`, right?

@mantepse Exactly. `base_ring` is what users expect from their experience with virtually every other class. Apologies for the overly telegraphic style in #33380. I also didn't specify whether I want the old name `ring` deprecated or not. I would be in favor of doing deprecation.

SandwichGouda:
Closing this since it I had not understood mkoeppe's intentions in his issue. 
I will open another PR soon enough with different and appropriate changes.

PR #37702, opened 2024-03-31
Workaround for multiprocessing shared memory limits

tornaria:
This affects musl libc and macos, see:
https://github.com/sagemath/sage/pull/36741#issuecomment-2027357414

We are working on a better fix to avoid using multiprocessing in sage.features (cf https://github.com/sagemath/sage/pull/36741#issuecomment-2028145753), but this will do meanwhile.

There is agreement that disabling the lock is completely harmless, since the counter is not really used for anything; disabling the lock fixes the issues that we had on 10.4.beta0.

### :memo: Checklist


soehms (approved):
LGTM, Thanks!

PR #37703, opened 2024-03-31
Implement an is_nilpotent() method for matrices

tscrim:
Checking if a matrix is nilpotent is a fundamental test that Sage is (surprisingly) missing. We check this by using the char poly.

### :memo: Checklist

### DIFF
--- a/src/sage/matrix/matrix2.pyx
+++ b/src/sage/matrix/matrix2.pyx
@@ -9915,8 +9915,11 @@ cdef class Matrix(Matrix1):
             ....:             [1,1,7,1,-5], [1,1,1,7,-5]])
             sage: B.is_nilpotent()
             True
-            sage: C = matrix(GF(7), [[1, 2], [2, 4]])
+            sage: C = matrix(GF(7), [[1, 2], [2, 6]])
             sage: C.is_nilpotent()
+            False
+            sage: D = matrix([[1,0],[0,0]])
+            sage: D.is_nilpotent()
             False
             sage: Z = matrix.zero(QQ, 5)
             sage: Z.is_nilpotent()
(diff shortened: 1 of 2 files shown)


fchapoton:
a failing doctest in src/sage/structure/element.pyx

tscrim:
Indeed. I have simply updated the test.

fchapoton:
maybe it would be worth checking the trace first, for speed ?

src/sage/matrix/matrix2.pyx
```diff
@@ -9898,6 +9898,44 @@ cdef class Matrix(Matrix1):
         self.cache(key, normal)
         return normal
 
+    def is_nilpotent(self) -> bool:
+        r"""
+        Return if ``self`` is a nilpotent matrix.
+
+        A matrix `A` is *nilpotent* if there exists a positive integer
+        `k` such that `A^k = 0`. We test this by using the Cayley-Hamilton
+        theorem to see if the characteristic polynomial is `x^n = 0`.
```
mantepse, line 9907: shouldn't that be `x^k = 0`?

mantepse: actually, it might be more precise to say:
```
We test this by checking whether the characteristic polynomial is a monomial, `x^n`.
The Cayley-Hamilton theorem asserts that then the minimal `k` such that `A^k = 0` equals `n`.
```

fchapoton: Wot ? the minimal k can be anything between 1 and n...

mantepse: upsi, sorry, in delirium I confused minimal and characteristic polynomial.  now going back to sleep.

tscrim:
> maybe it would be worth checking the trace first, for speed ?

Good idea. It crossed my mind to do this, but then I forgot at some point.

fchapoton (approved):
ok, good enough

tscrim:
Thank you.

Issue #37704, opened 2024-03-31
`_repr_` description of sum morphisms


grhkm21:
It seems that the `repr` of sum and composite morphisms are different. Is that intentional?
```sage
sage: set_random_seed(13337)
....: E = EllipticCurve(GF(103), [31, 84])
....: assert E.order() % 7 == 0
....: P = E.random_point() * (E.order() // 7)
....: assert P.order() == 7
....: phi = E.isogeny(P)
sage: phi
Isogeny of degree 7 from Elliptic Curve defined by y^2 = x^3 + 31*x + 84 over Finite Field of size 103 to Elliptic Curve defined by y^2 = x^3 + 42*x + 35 over Finite Field of size 103

sage: phi + phi
Sum morphism:
  From: Elliptic Curve defined by y^2 = x^3 + 31*x + 84 over Finite Field of size 103
  To:   Elliptic Curve defined by y^2 = x^3 + 42*x + 35 over Finite Field of size 103
  Via:  (Isogeny of degree 7 from Elliptic Curve defined by y^2 = x^3 + 31*x + 84 over Finite Field of size 103 to Elliptic Curve defined by y^2 = x^3 + 42*x + 35 over Finite Field of size 103, Isogeny of degree 7 from Elliptic Curve defined by y^2 = x^3 + 31*x + 84 over Finite Field of size 103 to Elliptic Curve defined by y^2 = x^3 + 42*x + 35 over Finite Field of size 103)

sage: E1 = phi.codomain()
....: psi = E1.isogeny(E1.random_point())
....: psi * phi
Composite morphism of degree 49 = 7^2:
  From: Elliptic Curve defined by y^2 = x^3 + 31*x + 84 over Finite Field of size 103
  To:   Elliptic Curve defined by y^2 = x^3 + 11*x + 52 over Finite Field of size 103
```

I think the "Via: " part should be hidden away. What do you think @yyyyx4

yyyyx4:
Indeed, unifying and streamlining the `repr()` and `str()` of various types of elliptic-curve morphisms (i.e., `EllipticCurveHom` children) has been an item on my list for a long time. Feel free to take care of it if you like.

grhkm21:
I'll remove the via part & move all `str/repr` methods to SchemeMorphism (or higher) in #37705 then.

PR #37705, opened 2024-03-31
Implementing `SchemeMorphism_sum`

grhkm21:
This PR implements a `SchemeMorphism_sum` class, which is a sum of scheme morphisms. It is refactored out from @yyyyx4's `EllipticHom_sum` class for sum of elliptic curve homomorphisms (zero or isogeny). As a result, we can now mess around with sum of jacobian homomorphisms. This is a step towards defining the endomorphism ring of um *things* (elliptic curves and jacobians?).

cc: @kwankyu 

Four questions:
- [ ] How to implements [richcmp](https://github.com/sagemath/sage/pull/37705/files#diff-f0685ad69ab5a3292d9a9c7e9deb3f0e8406f4d36c6b616071b509e23d1703f6R861-R899)? Should I do it stupidly and let specified classes (e.g. `EllipticCurveHom`) overload it?
- [ ] Should the "Via:" argument (e.g. [here](https://github.com/sagemath/sage/pull/37705/files#diff-f0685ad69ab5a3292d9a9c7e9deb3f0e8406f4d36c6b616071b509e23d1703f6R362-R364)) be removed? c.f. #37704
- [ ] Should the [addition check](https://github.com/sagemath/sage/pull/37705/files#diff-f0685ad69ab5a3292d9a9c7e9deb3f0e8406f4d36c6b616071b509e23d1703f6R789-R794V) exist at all?
- [ ] Should this class be moved upwards, e.g. "next to" `sage.categories.homset.Homset`?

TODO:
- [X] Move/copy tests from elliptic curves to morphism.py
- [X] ~~Use `FormalSum` or similar to store the summands~~ no, as not all morphisms are hashable.
- [x] Figure out what's happening with `SchemeMorphism` vs `Morphism` - inheriting elliptic curve morphisms from `SchemeMorphism` fails.


src/sage/structure/parent.pyx
```diff
@@ -1438,6 +1438,40 @@ cdef class Parent(sage.structure.category_object.CategoryObject):
...
+
+        The endomorphism set of ``self`` in the category ``category``.
+
+        .. SEEALSO:: :func:`~sage.categories.homset.End`, :func:`~sage.categories.homset.Hom`
+
+        EXAMPLES::
+
+            sage: R.<x,y> = PolynomialRing(QQ, 2)
+            sage: R.End()
+            Set of Homomorphisms from Multivariate Polynomial Ring in x, y over Rational Field to Rational Field
```
mkoeppe: This doesn't look right

grhkm21: Ah yes. I will fix it after Build&Test CI is done

src/sage/schemes/generic/morphism.py
```diff
@@ -2146,3 +2408,17 @@ def specialization(self, D=None, phi=None, ambient=None):
                 ambient = ambient.change_ring(phi.codomain().base_ring())
         psi = ambient.ambient_space().coordinate_ring().hom([0 for i in range(ambient.ambient_space().ngens())], ambient.base_ring())
         return ambient([psi(phi(t)) for t in self])
+
+    def _add_(self, point):
+        r"""
+        Add two point schemes.
```
mkoeppe: scheme points?

src/sage/schemes/generic/morphism.py
```diff
@@ -100,8 +102,7 @@ def is_SchemeMorphism(f):
 
     OUTPUT:
 
-    Boolean. Return ``True`` if ``f`` is a scheme morphism or a point
```
mkoeppe, line 103: When touching any of these `is_...` functions, best to deprecate them. We are trying to get rid of all of them in favor of using `isinstance` (or categories) in the code. See #32414 for pointers

src/sage/schemes/generic/morphism.py
```diff
@@ -337,13 +342,43 @@ def _repr_(self):
             s += "\n  To:   %s" % self._codomain
         d = self._repr_defn()
         if d != '':
-            s += "\n  Defn: %s" % ('\n        '.join(self._repr_defn().split('\n')))
+            s += "\n  Defn: %s" % '\n        '.join(d.split('\n'))
         return s
 
+    def _add_(self, other):
+        r"""
```
mkoeppe, line 350: I think this `_add_` method should only be provided when the underlying scheme is in the category of additive magmas. 
```
sage: E = EllipticCurve(j=42)
sage: E.categories()
[Category of abelian varieties over Rational Field,
 Category of commutative additive groups,
 Category of additive groups,
 Category of additive inverse additive unital additive magmas,
 Category of commutative additive monoids,
 Category of additive monoids,
 Category of additive unital additive magmas,
 Category of commutative additive semigroups,
 Category of additive commutative additive magmas,
 Category of additive semigroups,
 Category of additive magmas,
 Category of schemes over Rational Field,
 Category of schemes,
 Category of sets,
 Category of sets with partial maps,
 Category of objects]
```

mkoeppe: The problem: For Jacobians of HyperellipticCurve, nothing is declared. What laws does the addition of points of it satisfy?
```
sage: R.<x> = QQ[]
            sage: J = HyperellipticCurve(x^5 + x + 1).jacobian()
            sage: phi = End(J).identity()
sage: EJ = End(J)
sage: EJ.categories()
[Category of endsets of schemes over Rational Field,
 Category of endsets,
 Category of monoids,
 Category of semigroups,
 Category of unital magmas,
 Category of magmas,
 Category of homsets of schemes over Rational Field,
 Category of homsets,
 Category of sets,
 Category of sets with partial maps,
 Category of objects]
```

grhkm21: It has an abelian group structure - it's basically a $\mathbb{Z}$-module (with some group quotients).

grhkm21: I will look into how to add this categories structure, I'm not too familiar with it.

grhkm21: I have now made Jacobians (over fields) abelian varieties:
```diff
# a.patch
# this is ok since Jacobian constructor only accepts curves over fields
-         Scheme.__init__(self, C.base_scheme())
+         Scheme.__init__(self, C.base_scheme(), category=AbelianVarieties(C.base_ring()))
```
Then
```sage
sage: EllipticCurve(QQ, [0, 1]).categories() == HyperellipticCurve(polygen(QQ)^5 + 1).jacobian().categories()
True
```

grhkm21: I have fixed the jacobian issue so that (nice) jacobians have an abelian variety structure.
```sage
sage: EllipticCurve(j=42).categories()
[Category of abelian varieties over Rational Field,
 Category of commutative additive groups,
 Category of additive groups,
 Category of additive inverse additive unital additive magmas,
 Category of commutative additive monoids,
 Category of additive monoids,
 Category of additive unital additive magmas,
 Category of commutative additive semigroups,
 Category of additive commutative additive magmas,
 Category of additive semigroups,
 Category of additive magmas,
 Category of schemes over Rational Field,
 Category of schemes,
 Category of sets,
 Category of sets with partial maps,
 Category of objects]
sage: Jacobian_generic(HyperellipticCurve(x^5 + 1)).categories() == EllipticCurve(j=42).categories()
True
```

You mentioned that
> I think this `_add_` method should only be provided when the underlying scheme is in the category of additive magmas.

How should I implement this? Should I check if `self in AdditiveMagmas() and other in AdditiveMagmas()`?

src/sage/schemes/generic/morphism.py
grhkm21: I will do this in a separate PR

src/sage/schemes/curves/projective_curve.py
```diff
@@ -220,6 +220,37 @@ def __init__(self, A, X, category=None):
 
         Curve_generic.__init__(self, A, X, category=category)
 
+    def _an_element_(self):
+        r"""
+        Return an element of this projective curve. We try to find variables
+        that never appear as a pure monomial, and fail otherwise.
```
mkoeppe:
```suggestion
        Return an element of this projective curve. 
        
        We try to find variables that never appear as a pure monomial, and fail otherwise.
```
(First line of docstring should be 1-line summary)

src/sage/schemes/generic/morphism.py
mkoeppe: I think you can just provide it via the category, as `AbelianVarieties.MorphismMethods._add_` etc.

mkoeppe:
I think you'll be better off with making a class for arbitrary finite `ZZ`-linear combinations of the morphisms, not just sums. Then you can implement not just _add_ but also _lmul_, _rmul_ with integers, etc.

grhkm21:
> I think you'll be better off with making a class for arbitrary finite `ZZ`-linear combinations of the morphisms, not just sums. Then you can implement not just `_add_` but also `_lmul_`, `_rmul_` with integers, etc.

Okay, that should just be `FormalSums` then?

mkoeppe:
sort of, but the implementation of `FormalSums` is not in very good shape

mkoeppe:
What `FormalSums` is lacking: control over what elements are allowed.

grhkm21:
I can overload the constructor of `SchemeMorphism_sum` and check the correctness of types there, I guess. Let me try it first.

Just a heads up, this PR is my first time writing any structural code, and I had 0 experience with Parent/Element/Category's. I'm just reading the [wiki](https://doc.sagemath.org/html/en/reference/structure/sage/structure/sage_object.html#sage.structure.sage_object.SageObject.category) as I go..

src/sage/schemes/generic/morphism.py
```diff
@@ -669,6 +710,201 @@ def glue_along_domains(self, other):
...
+                raise ValueError(f"addition is not implemented for {codomain}")
+        for phi in phis:
+            if phi.domain() != domain:
+                raise ValueError(f'summand {phi} has incorrect domain (need {domain})')
+            if phi.codomain() != codomain:
+                raise ValueError(f'summand {phi} has incorrect codomain (need {codomain})')
+
+        self._phis = phis
+        self._domain = domain
+        SchemeMorphism.__init__(self, self._domain, codomain)
```
grhkm21, line 809: This line is wrong... The second argument is the `parent` (homset) not the domain 😓

grhkm21:
I don't understand how multi-hierarchy should be done in Sage. Here I want the `Homset_sum` (let's say `X`) to be both a FormalSum and a SchemeMorphism. If I put `FormalSum` first, then `X.parent()` would be `FormalSums()`, which is undesirable. If I put SchemeMorphism first, then `X + X` will call the `_add_` method of SchemeMorphism (as a fallback) instead of FormalSum, which is undesirable.

I think I won't inherit from FormalSum, but instead modify the previous implementation so that instead of `self._phis` being a list of summands, it is a defaultdict(int). I have an implementation that I can push soon

mkoeppe:
Yes, don't try to inherit from `FormalSum`. 
(If you really want to reuse any functionality from `FormalSum`, just store an instance of it and delegate to it, i.e., use the "has-a", not the "is-a" pattern.)

grhkm21:
Update: I don't think this is possible, since not everything is hashable. For example, consider hashing an endomorphism of an elliptic curve over $\mathbb{Q}_3$, where you map $x$ to $ax$, $a \in \mathbb{Q}_3$ or whatever. You eventually have to hash $a$, which is not okay due to #11895.

grhkm21:
Something about "Test modularized distributions" is failing, and not just for me

nbruin:
Concerning the motivation in the ticket: it would look to me you'd *first* need endomorphism rings of abelian varieties and *then* homomorphisms to/from them can be modules over them (by pre- and/or post-composition).

Concerning changes to the coercion framework: I guess the main issue with morphisms sets that also form Z-modules is that there's the very general (partial) composition operation and that there is a distinguished set of endomorphisms somewhere that is canonically isomorphic to Z, which produces an action of Z on the set as well. It may well be possible to tweak things so that this works. It works for matrices ... However, whenever you tweak the coercion system you have to be very careful about performance: it tries quite hard to stay out of the way of low-level operations (or at least provide a fast path for it) and that's easily ruined by small changes.

The other thing is a general design: it's of course nice if an object can be everything it is at once, but we have forgetful functors for a reason in maths: if you overload all meaning onto an object you could end up with a very restricted notion of morphism. You'll have to be able to dress down the information you're keeping track of so that you can also consider the object in a category with more permissive morphisms. So I think in many cases it will be more flexible to just implement the particular view you're interested in and facilitate easy conversion to objects that emphasize other aspects, and I suspect in some cases you may have to go with that approach. It's not clear to me whether this is one of those cases.

mkoeppe:
A relevant old ticket:
- #14279

mkoeppe:
> Something about "Test modularized distributions" is failing

Yes, unrelated to this PR and on my list of things to fix.

src/sage/schemes/generic/homset.py
```diff
@@ -380,11 +445,8 @@ def _element_constructor_(self, x, check=True):
             sage: A.<x,y> = AffineSpace(R)
             sage: C = A.subscheme(x*y - 1)
             sage: H = C.Hom(C); H
-            Set of morphisms
-              From: Closed subscheme of Affine Space of dimension 2 over Rational Field
-                    defined by: x*y - 1
-              To:   Closed subscheme of Affine Space of dimension 2 over Rational Field
-                    defined by: x*y - 1
+            Set of scheme endomorphisms of Closed subscheme of Affine Space of dimension 2 over Rational Field defined by:
+             x*y - 1
```
mkoeppe, line 450: Please wrap long doctest output lines, see https://doc.sagemath.org/html/en/developer/coding_basics.html#writing-testable-examples

grhkm21:
@nbruin 

> Concerning the motivation in the ticket: it would look to me you'd first need endomorphism rings of abelian varieties and then homomorphisms to/from them can be modules over them (by pre- and/or post-composition).

Do you mind explaining this a bit more? What should the classes inheritance look like? There is `End(A) == Hom(A, A)` already, and I can make it a ring for abelian varieties, if that's what you mean.

Thanks for the other two comments, I will think about these. I agree that sometimes it's more appropriate to separate functionalities, it will just be the most convenient if both functionalities exist at the same time, but maybe in this case that'll make things a lot messier.

grhkm21:
> A relevant old ticket:
> 
>     * [Coercion for homsets #14279](https://github.com/sagemath/sage/issues/14279)

I will have a look at that. This line seems interesting:

> One important change: I let `sage.categories.map.Map` inherit from `sage.structure.element.ModuleElement`. Not all homsets are modules, of course, but when they are, then one can simply define single underscore `_add_`, `_rmul_` and `_composition_` (the latter is for composition of maps) as usual, so that one does not need to overload `__mul__` so often.

They basically do a mathematically wrong thing, but say "when it's mathematically wrong, just don't define the relevant methods"...

grhkm21:
Very nice to see that all examples in #14279 have been sped up by ~5x though, either by my CPU or by Sage

src/sage/schemes/generic/homset.py
```diff
@@ -582,7 +644,7 @@ def _coerce_map_from_(self, other):
             except AttributeError:  # no .ambient_space
                 return False
         elif isinstance(other, SchemeHomset_points):
-        #we are converting between scheme points
+            #we are converting between scheme points
```
mkoeppe:
```suggestion
            # we are converting between scheme points
```

src/sage/schemes/generic/morphism.py
```diff
@@ -669,6 +714,236 @@ def glue_along_domains(self, other):
...
+        sage: phi = End(J).identity(); phi
+        Scheme endomorphism of Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
+          Defn: Identity map
+        sage: psi = phi + phi; psi
+        Sum morphism:
+          From: Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
+          To:   Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
+          Via:  (Scheme endomorphism of Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
+          Defn: Identity map, Scheme endomorphism of Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
+          Defn: Identity map)
```
mkoeppe:
```suggestion
        Sum morphism:
          From: Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
          To:   Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
          Via:  (Scheme endomorphism of Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
                   Defn: Identity map, 
                 Scheme endomorphism of Jacobian of Hyperelliptic Curve over Rational Field defined by y^2 = x^5 - 8*x
                   Defn: Identity map)
```

grhkm21:
I will fix this after my exams are done, but I think a more productive and better way to approach this is to fix the issue with SchemeMorphism and Morphism being incompatible first, just so we don't duplicate code between the two classes... Otherwise, we will need both `Morphism_sum` and `SchemeMorphism_sum`

grhkm21:
draft until I can think of a sane way to implement this that does not copy and pasting more code and maintaining the same code on two different places

grhkm21:
I think this PR is not the correct way to implement the feature I want (sum of morphisms of *abelian varieties*), since of course you can't just add two arbitrary scheme morphisms (unless we want to form formal sums, but that's not a good idea as a default implementation...) so I will close this PR, but I believe the code here can be reused in another PR.

PR #37713, opened 2024-04-01
`sage.sets`: Doctest cosmetics

mkoeppe:
Cherry-picked from #35095.

### :memo: Checklist

### DIFF
--- a/src/sage/sets/recursively_enumerated_set.pyx
+++ b/src/sage/sets/recursively_enumerated_set.pyx
@@ -209,7 +209,7 @@ or even::
 We can then create the :class:`RecursivelyEnumeratedSet` object with either::
 
     sage: S = RecursivelyEnumeratedSet([''],
-    ....:     lambda x: [x+letter for letter in ['a', 'b', 'c']]
+    ....:     lambda x: [x + letter for letter in ['a', 'b', 'c']]
     ....:               if len(x) < 2 else [],
     ....:     structure='forest', enumeration='depth',
     ....:     category=FiniteEnumeratedSets())
@@ -248,8 +248,8 @@ convention is that the generated elements are the ``s := f(n)``, except when
     ....:     st = set(st) # make a copy
     ....:     if st:
     ....:        el = st.pop()
-    ....:        for i in range(len(lst)+1):
-    ....:            yield (lst[0:i]+[el]+lst[i:], st)
+    ....:        for i in range(len(lst) + 1):
+    ....:            yield (lst[0:i] + [el] + lst[i:], st)
     sage: list(children(([1,2], {3,7,9})))
     [([9, 1, 2], {3, 7}), ([1, 9, 2], {3, 7}), ([1, 2, 9], {3, 7})]
     sage: def post_process(node):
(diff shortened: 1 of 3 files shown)


mantepse:
I am against replacing short `lambda` expressions with two line `def` in docstrings: they are a pain to copy into the repl.

mkoeppe:
> I am against replacing short `lambda` expressions with two line `def` in docstrings: they are a pain to copy into the repl.

Are you referring to copying from the source code or from the formatted HTML documentation? Where exactly does it hurt?

mantepse:
> > I am against replacing short `lambda` expressions with two line `def` in docstrings: they are a pain to copy into the repl.
> 
> Are you referring to copying from the source code or from the formatted HTML documentation? Where exactly does it hurt?

Copying from source code.    It does work, but it is a painful.   For example, modifying does not work as well, and also emacs doesn't offer autocompletion for functions defined in this way.

Also, I don't see a good reason for the replacement.  What is the purpose?

mkoeppe:
> What is the purpose?

Coding style. In library code, such lambda assignments are flagged by pycodestyle, and they have been replaced already. Best to use a consistent style also in code examples.

mkoeppe:
> Copying from source code.

You know that you can copy the whole lines including prompts, yes?

mantepse:
> > What is the purpose?
> 
> Coding style. In library code, such lambda assignments are flagged by pycodestyle, and they have been replaced already. Best to use a consistent style also in code examples.

I would agree that in very many cases the assignment is unnecessary, and could simply be inlined.

Otherwise, you could simply use the one-line form of `def`:
```
def f(x): return 2*x 
```
works just fine.

> > Copying from source code.
> 
> You know that you can copy the whole lines including prompts, yes?

Is there an editor that would not let you copy full lines?

mkoeppe:
We appear to be miscommunicating. What's the exact problem that you are facing when copying multi-line definitions?

mkoeppe:
> use the one-line form of `def`:
> 
> ```
> def f(x): return 2*x 
> ```

Codestyle linters also don't like that

mantepse:
It is a bit hard to explain if you do not use emacs.

I can go to the first line of the test in the documentation buffer, and press `C-d`, to run the snippet in the attached sage session.  It will look as follows:
```
sage: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:def rotate_permutation(p):
:    cycle = Permutation(tuple(range(1, len(p)+1)))
:    return Permutation([cycle.inverse()(p(cycle(i))) for i in range(1, len(p)+1)])
:--
```
If I do it this way (rather than copy-pasting), I at least get tab completion.  However, if I now want to modify the definition slightly, it gets a bit cumbersome, at least for me.

I also find that the increased verbosity is another drawback.

I do see the point of using pycodestyle, but I would rather turn off that particular recommendation for doctests, or relax it to allow single line `def` statements.

mkoeppe:
Is this `sage-shell-mode`?

mantepse:
Yes

mkoeppe:
I've reduced this PR to the subset of the changes that do not involve changing `lambda` assignments to `def`s.

mantepse (approved):
LGTM, thank you!

mkoeppe:
Thanks.

PR #37719, opened 2024-04-01
Refactor ring categories

fchapoton:
A somewhat large refactoring of some of the auld classes for rings, and related categories

- introducing a new category of Noetherian rings
- moving some methods in appropriate categories
- fixing necessary details

### :memo: Checklist

Depends on #37778


fchapoton:
damn, there remains annoying failing doctests in `src/sage/matrix`  !

fchapoton:
One of the remaining bug is
```
sage: A = Zmod(1)
sage: B = A['x']
sage: B.one()
crash in coercion somehow
```

tscrim:
Okay, I've figured it out. The change in the category from Algebras to Rings means that the base ring coercion is no longer provided by default. It's actually somewhat amazing that this doesn't end up going through an infinite loop, but that is how `coerce_map_from` works with its caching. If you add
```python
        if P == base_ring:
            return self._coerce_map_from_base_ring()
```
to the beginning of `PolynomialRing_general._coerce_map_from_()`, then the coercion will work. It's a bit of a dirty workaround, but I don't see how to do it any better. The other option I considered is doing a second `base_ring.is_zero()` at the end of the `__init__` and registering the coercion manually. Yet this felt even less clean, but I can't really say why though. (I am also surprised the multivariate case works; I don't understand how it does at this point.)

There are some other issues/inconsistencies with polynomials over $F_1$:
```
sage: B = A['x,y']
sage: B.gens()
(0, 0)
sage: B = A['x']
sage: B.gens()
(x,)
```
The univariate case is wrong IMO.

The change in `multi_polynomial.pyx` is for the polynomial ring with no generators, correct? Can we add a test for this?

I am not sure about the change for `LocalGeneric` to `Parent` from `CommutativeRing` because the latter class seems to have methods not provided by the category (e.g., `localization`).

fchapoton:
Thanks a lot, Travis ! I was worrying to be unable to debug that issue about `Zmod(1)[x]`.

I propose to keep the other inconsistencies of polynomials over the zero ring for later.

I have added the suggested doctest in `multi_polynomial`.

For the change in `local_generic`, I do not remember why I was forced to do that. In any case, it contributes to the future possible removal of `CommutativeRing`, one day.

I have disabled a doctest in the testsuite of Laurent polynomials over `Zmod(2)` about gcd.

fchapoton:
no idea what the CI complains about. Something about "modularity", it seems......

mkoeppe:
> no idea what the CI complains about

The failure is unrelated to this PR and tracked in #37734.

mkoeppe (approved):
LGTM

grhkm21:
Setting to positive since Matthias approved, unless Travis disagrees.

tscrim:
Could you say a little bit more about the Laurent polynomial test suite failure?

Also, did you check that the change for `LocalGeneric` didn’t cause anything to lose the methods defined in `CommutativeRing`? I don’t think they are all defined in the category. So we need to make these methods available, and the easiest way is to not change the base class.

fchapoton:
I have undone the change in `local_generic.py`, although it's a step bakcwards. It can be re-done elsewhere more carefully.

The disactivated doctest in TestSuite of Laurent polynomials is about `gcd` versus `xgcd` in Laurent polynomials. I am not sure which one is changed. And also not sure if there is a clear definition of either, because monomials are invertible.

tscrim:
I think I can fix the xgcd for the Laurent polynomials.

tscrim:
The Laurent polynomial tests should now pass with #37778.

fchapoton:
I am not sure how to handle dependencies here ; I try hard to avoid them, but now #37778 is a reasonable one.. By the way, thanks for investigating and finding the cure, Travis !

tscrim:
I would prefer to have #37778 as a dependency since it fixes a regression here. I know it is annoying for you, especially since it is just to prevent a doctest from failing otherwise. Yet, it would be better to not disable the test. I could also make #37778 a dependency on this that undoes the skipped test, but that is somewhat contrary to the logical flow of the PRs. Anyways, I will leave the decision up to you.

tscrim (approved):
Thank you. Once the bot comes back green, positive review.

fchapoton:
As expected, the lights are not green. As I said, I have no idea if it is possible to declare a dependency in such a way that the CI will merge the branches before testing.

mantepse:
> As expected, the lights are not green. As I said, I have no idea if it is possible to declare a dependency in such a way that the CI will merge the branches before testing.

As far as I know, you have to merge the dependency into your branch.

mkoeppe:
I've repaired the CI. It complains:
```
sage -t --random-seed=232623843441133993949100432847335930207 src/sage/matrix/matrix_space.py
File "src/sage/matrix/matrix_space.py", line 493, in sage.matrix.matrix_space.MatrixSpace
Failed example:
    MatrixSpace(ZZ, 10, 5).category()
Expected:
    Category of infinite enumerated finite dimensional modules with basis over
     (Dedekind domains and euclidean domains
      and infinite enumerated sets and metric spaces)
Got:
    Category of infinite enumerated finite dimensional modules with basis over (Dedekind domains and euclidean domains and noetherian rings and infinite enumerated sets and metric spaces)
File "src/sage/matrix/matrix_space.py", line 497, in sage.matrix.matrix_space.MatrixSpace
Failed example:
    MatrixSpace(ZZ, 10, 10).category()
Expected:
    Category of infinite enumerated finite dimensional algebras with basis over
     (Dedekind domains and euclidean domains
      and infinite enumerated sets and metric spaces)
Got:
    Category of infinite enumerated finite dimensional algebras with basis over (Dedekind domains and euclidean domains and noetherian rings and infinite enumerated sets and metric spaces)
1 item had failures:
   2 of  50 in sage.matrix.matrix_space.MatrixSpace
    [4[60](https://github.com/sagemath/sage/actions/runs/8948624007/job/24713068325?pr=37719#step:12:61) tests, 2 failures, 2.97 s]
```

src/sage/categories/rings.py
```diff
@@ -339,6 +339,101 @@ def is_commutative(self) -> bool:
...
+            INPUT:
+
+            - ``proof`` -- (default: ``True``) Determines what to do in unknown
+              cases
+
+            ALGORITHM:
+
+            If the parameter ``proof`` is set to ``True``, the returned value is
+            correct but the method might throw an error.  Otherwise, if it is set
+            to ``False``, the method returns ``True`` if it can establish that self
```
mkoeppe:
```suggestion
            to ``False``, the method returns ``True`` if it can establish that ``self``
```

src/sage/categories/commutative_rings.py
```diff
@@ -248,6 +281,19 @@ class Finite(CategoryWithAxiom):
             ....:                    GF(5)]) in Rings().Commutative().Finite()
             True
         """
+        def extra_super_categories(self):
+            r"""
+            Let Sage knows that finite commutative rings
+            are Noetherian.
```
mkoeppe: check whether you want to use "Noetherian" or "noetherian" throughout the text

mkoeppe (approved):
LGTM.

PR #37721, opened 2024-04-01
Renamed "ring" argument of matrix contructor to "base_ring"

SandwichGouda:
The matrix constructor used to accept ``ring`` as argument to specify a the base ring of a matrix. 
It now takes ``base_ring`` instead. ``ring`` is still accepted but deprecated (I added a deprecation in the code). 
This is for unification (``base_ring`` is what users expect from their experience with other classes).
This solves @mkoeppe's issue #33380. 

### :memo: Checklist

### DIFF
--- a/src/sage/matrix/args.pyx
+++ b/src/sage/matrix/args.pyx
@@ -28,7 +28,7 @@ from sage.structure.element cimport Element, RingElement, Vector
 from sage.arith.long cimport pyobject_to_long
 from sage.misc.misc_c import sized_iter
 from sage.categories import monoids
-from sage.misc.decorators import rename_keyword
+from sage.misc.superseded import deprecation_cython
 
 
 try:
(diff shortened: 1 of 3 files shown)


mkoeppe (approved):
LGTM, thanks for working on it!

SandwichGouda:
My pleasure ! 
Quick question : 
You may have noticed that I attempted to use the [@rename_keyword](https://doc.sagemath.org/html/en/developer/coding_in_python.html#deprecation) in the first commit. 
This did not work : the ``matrix`` constructor worked normally when using ``base_ring``, but when using ``ring``, it raised an ``unexpected keyword`` error, as if the ``@rename_keyword`` decorator wasn't there. (The expected behaviour was that it should've returned matrices as if I had used ``base_ring``, printing a deprecation)
Do you have any insights as to why it did not work ? Do you think it's because these decorators are intended only for use (and only work in) in Python (``.py`` files) and not Cython (``.pyx``) ?

mkoeppe:
> Do you think it's because these decorators are intended only for use (and only work in) in Python (`.py` files) and not Cython (`.pyx`) ?

Possibly, or perhaps it only does not work for `__init__` methods in Cython. That would be worth investigating and documenting. The definition of the decorator (`src/sage/misc/decorators.py`) currently does not mention or test any of this.

`git grep rename_keyword` does show a small number of uses in .pyx files but they look like they may have been applied mechanically, and I'm not sure if they work correctly.

PR #37736, opened 2024-04-03
Reimplementing the Witt (Sym Func) change of basis, caches, and omega involution

tscrim:
We reimplement the coercion from the Witt basis of symmetric functions to the $h$, $e$, and $p$ bases and their inverses by using the generating function identities and recursion for $w_n, h_n, e_n, p_n$ and the mulitplicative basis property for the general shape (in particular, we attempt to minimize the number of multiplications and maximize the use of the cache by recursively splitting the partition by its smallest part). We cache the result using `@cached_method` instead of custom classes. We also deprecate the `coerce_*` inputs.

We also take advantage that $\omega w_n = w_n$ when $n$ is odd to compute the omega involution. This extends to a faster computation of the antipode.

### :memo: Checklist


mantepse:
I'm impressed!  (I'll just wait for the bot.)

mantepse (approved):
This is a massive improvement.  I checked the change of basis from witt to elementary for all partitions of 20, which took 6.2s previously (with `coerce_e=True`) and now 4s.

Thank you!

darijgr:
Nice work, thanks Travis! (I've checked the math and it's all good. I assume the programming part can't be bad, as it is replacing complicated code by simple code.)

darijgr (approved):
Mathematics correct!

tscrim:
Thank you.

tscrim:
I made one trivial change that I forgot to do: I updated the issue/PR number referenced in the deprecation warning to this PR.

Issue #37739, opened 2024-04-03
CI Docbuild: Silence documentation preview comment when there is no change


mkoeppe:
Wishlist item:

- Suppress the preview comment when there is no change to the documentation.

- Download a previously deployed documentation archive from netlify, and then suppress the preview comment if there was no change compared to that.

_Originally posted by @kwankyu in https://github.com/sagemath/sage/issues/37387#issuecomment-2028718420_

orlitzky:
This was done in https://github.com/sagemath/sage/pull/37740

PR #37387, opened 2024-02-18
Remove noisy documentation preview comment

tobiasdiez:
The comment linking to the built docs is noisy and doesn't add much value: the docs can also easily be accessed by clicking on the "Successfully deploy preview" status check or via the summary of the deploy workflow.

On the other hand, it always triggers a notification. Hence, we remove the comment.

### :memo: Checklist

- [ ] The title is concise, informative, and self-explanatory.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

Issue #36803, opened 2023-12-03
Developer's guide: Include resources for managing cognitive overload in Sage development


mkoeppe:
Problem description:
- too much cognitive noise, all these packages and PRs and discussions [...]
- It doesn't help that the "fun" and "no fun" [packages are] mixed up in one place, in alphabetical order, no structure whatsoever.
- In a too noisy workplace one gets easily distracted and irritated

_Originally posted by @dimpase in https://github.com/sagemath/sage/issues/36777#issuecomment-1837277961_

As part of our goal to make the Sage development community more inclusive, we expand the developer's guide with strategies, resources, tools for managing cognitive overload in Sage development.

dimpase:
yeah yeah, let's meditate the software bugs away   😛

as long as PRs and issues are all piled up in the same place, and, more importantly, the releases are only repo-wide, our monorepo would not fly.

the flat spkgs tree structure, with dependencies and purposes not reflected in it, is not helping either.

mkoeppe:
Let's please take this Issue seriously.

mkoeppe:
From my personal experience, GitHub Issues and PRs are excellent tools for managing attention / focus. 

An important part is that the **scope** of a PR is defined clearly in its description, and narrowed whenever necessary. Experienced developers know how to set the scope properly, and can recognize when the scope (in their own or others' PRs) needs to be narrowed. 

We may need to revise our Reviewer's Guide to emphasize the importance of scope of a PR. 

In particular, we should explain that reviewers need to use restraint when discussing topics outside of the scope of the PR in the comments of the PR. Although sometimes the "big picture" needs to inform decisions and discussions on a PR, there is the danger of creating distractions -- which can create cognitive overload for PR author and other reviewers.

dimpase:
I am very serious here, I think it's totally wrong to think that you can leave the workplace highly unoptimised, yet make it more agreeable, by making noises about managing cognitive overload.

mkoeppe:
I wrote: "we should explain that reviewers need to use restraint when discussing topics outside of the scope of the PR [...] there is the danger of creating distractions -- which can create cognitive overload for PR author and other reviewers."

Here's an example:

- #36726. Where one well-documented line of a POSIX shell command running in a Linux Docker container is to be reviewed; but a reviewer starts to talk about OpenBSD, about gfortran, about macOS, etc.

mkoeppe:
[...]

PR #37740, opened 2024-04-03
Disable most documentation notifications on GitHub

saraedum:
Previously, the documentation comment on GitHub would be removed and then recreated whenever there is a change to a PR which creates notification emails and to some feels like it's polluting their GitHub inbox.

Here, we change this behavior so that only the initial build of the documentation causes such a notification. Further updates to the documentation do update the links in the comment but do not trigger a notification.

This is meant as an alternative to #37387. See #37739 for futher ideas about a smarter behavior here.

An actual preview of the result can be seen here: https://github.com/saraedum/sage/pull/2

Note that the links on that PR do not work because my fork does not have netlify credentials configured.

### :memo: Checklist

- [x] I have created tests covering the changes; this is fairly hard to test automatically. I'll test this manually if there is interest in merging this.
- [x] I have updated the documentation accordingly; kind of, I updated the PR checklist.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@
 - [ ] The description explains in detail what this PR is about.
 - [ ] I have linked a relevant issue or discussion.
 - [ ] I have created tests covering the changes.
-- [ ] I have updated the documentation and checked the generated documentation. <!-- You do not need to build the documentation locally. A bot is recreating the documentation whenever there is a change to this PR and will maintain a link to the latest preview once it's ready. -->
+- [ ] I have updated the documentation and checked the generated documentation.
 
 ### :hourglass: Dependencies
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@
 - [ ] The description explains in detail what this PR is about.
 - [ ] I have linked a relevant issue or discussion.
 - [ ] I have created tests covering the changes.
-- [ ] I have updated the documentation accordingly.
+- [ ] I have updated the documentation and checked the generated documentation. <!-- You do not need to build the documentation locally. A bot is recreating the documentation whenever there is a change to this PR and will maintain a link to the latest preview once it's ready. -->
 
 ### :hourglass: Dependencies
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@
 - [ ] The description explains in detail what this PR is about.
 - [ ] I have linked a relevant issue or discussion.
 - [ ] I have created tests covering the changes.
-- [ ] I have updated the documentation and checked the generated documentation.
+- [ ] I have updated the documentation and checked the documentation preview.
 
 ### :hourglass: Dependencies
 
--- a/.github/workflows/doc-publish.yml
+++ b/.github/workflows/doc-publish.yml
@@ -72,7 +72,8 @@ jobs:
           header: preview-comment
           recreate: false
           message: |
-            [Documentation preview for this PR](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/html/en) (built with commit ${{ steps.source-run-info.outputs.sourceHeadSha }}; [changes](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/CHANGES.html)) is ready! :tada: This preview will update shortly after each push to this PR.
+            [Documentation preview for this PR](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/html/en) (built with commit ${{ steps.source-run-info.outputs.sourceHeadSha }}; [changes](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/CHANGES.html)) is ready! :tada:
+            This preview will update shortly after each push to this PR.
 
       - name: Update deployment status PR check
         uses: myrotvorets/set-commit-status-action@v2.0.0
@@ -134,4 +135,3 @@ jobs:
       - name: Report deployment url
         run: |
           echo "::notice::The live documentation has been deployed - ${{ steps.deploy-netlify.outputs.NETLIFY_URL }}"
-
--- a/.github/workflows/doc-publish.yml
+++ b/.github/workflows/doc-publish.yml
@@ -72,7 +72,7 @@ jobs:
           header: preview-comment
           recreate: false
           message: |
-            [Documentation preview for this PR](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/html/en) (built with commit ${{ steps.source-run-info.outputs.sourceHeadSha }}; [changes](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/CHANGES.html)) is ready! :tada:
+            [Documentation preview for this PR](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/html/en) (built with commit ${{ steps.source-run-info.outputs.sourceHeadSha }}; [changes](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/CHANGES.html)) is ready! :tada: This preview will update shortly after each push to this PR.
 
       - name: Update deployment status PR check
         uses: myrotvorets/set-commit-status-action@v2.0.0
--- a/.github/workflows/doc-publish.yml
+++ b/.github/workflows/doc-publish.yml
@@ -70,7 +70,7 @@ jobs:
         with:
           number: ${{ steps.source-run-info.outputs.pullRequestNumber }}
           header: preview-comment
-          recreate: true
+          recreate: false
           message: |
             [Documentation preview for this PR](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/html/en) (built with commit ${{ steps.source-run-info.outputs.sourceHeadSha }}; [changes](${{ steps.deploy-netlify.outputs.NETLIFY_URL }}/CHANGES.html)) is ready! :tada:
 


saraedum:
Setting this to "disputed" since @mkoeppe already said on #37387 that he is against this approach.

saraedum:
@tornaria @kwankyu @mantepse @tobiasdiez feel free to have a look :)

tornaria:
I think this is a reasonable compromise so :+1:

Alternatively: is the bot able to edit the description of the PR? Wouldn't it make more sense to put the link at the end of the description (and update it there every time)? This could be a more visible location for the link.

Maybe easier: is there an easy URL that will redirect to the latest version of the documentation for a given PR? Maybe this link could be added statically into the PR template?

saraedum:
> Maybe easier: is there an easy URL that will redirect to the latest version of the documentation for a given PR? Maybe this link could be added statically into the PR template?

Yeah, actually that's probably even easier.

grhkm21:
+1, *please*. Preferably I want to get 0 email notifications from github-actions[bot] for documentations.

grhkm21:
Another option is to learn from how Gitpod does it. For example, see [this Mathlib issue](https://github.com/leanprover-community/mathlib/pull/18998). They have a button that redirects to Gitpod, then (I think) Gitpod checks the referral header to see which repo & issue it comes from. We can do something similar, have an endpoint that takes the issue number and redirect to the built documentation.

tornaria:
> > Maybe easier: is there an easy URL that will redirect to the latest version of the documentation for a given PR? Maybe this link could be added statically into the PR template?
> 
> ~Yeah, actually that's probably even easier.~ Actually, no. We can't replace the PR number in the PR template easily.

Is it possible to have a static file in a single fixed URL that will read the PR from the referrer and redirect to the correct link? I assume this is what happens with the `https://gitpod.io/from-referrer/` url in the mathlib issue.

With some luck, this can be done in javascript without server side involvement.

tornaria:
Here's a test: https://www.cmat.edu.uy/~tornaria/ref.html

Let's try the link to see if javascript can figure out the PR number.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -15,7 +15,7 @@
 - [ ] The description explains in detail what this PR is about.
 - [ ] I have linked a relevant issue or discussion.
 - [ ] I have created tests covering the changes.
-- [ ] I have updated the documentation accordingly.
+- [ ] I have updated the documentation and checked the generated documentation. <!-- You do not need to build the documentation locally. A bot is recreating the documentation whenever there is a change to this PR and will maintain a link to the latest preview once it's ready. -->
```
kwankyu, line 19: The comment `<!-- ... -->` is helpful for a new sage developer but a visual noise to all others. Please remove the comment. 

Otherwise, I like the modified checklist.

kwankyu: Or you may want "... and checked the documentation preview".

saraedum: > The comment `<!-- ... -->` is helpful for a new sage developer but a visual noise to all others. Please remove the comment.

I moved the remark to the generated comment. I think it's important to explain how these things work. I fear that if you are not contributing constantly you might not understand that these things actually update (without you having to interact with the system in any way.)

kwankyu: Thanks.

Would you show a capture of the changed comment? 

Before I modified the comment to fit in a single line for easy read. I fear that the changed comment overflows to another line :-)

kwankyu: Since the comment itself is not updated, it might not matter though.

saraedum: Yes, it does overflow. (Depends on your browser of course, but on my desktop setup I get this.)

image

I don't see the problem with this overflowing to a second line though.

saraedum: I can change it to this if you prefer

image

kwankyu: OK with two lines.

kwankyu: Is the commit number updated?

saraedum: > Is the commit number updated?

Sure.

saraedum:
I tested that this actually works on https://github.com/saraedum/sage/pull/2. (The link there is broken because there is no netlify configured.) But the notification updates correctly and looks correct to me.

saraedum:
I certainly want to wait for a week or so before I tally the votes here. But if you'd like to, I would appreciate a formal review of this PR.

saraedum:
> [referer approach]

Browsers (and their privacy extensions) can decide not to send the referer at all or mangle it. From my understanding of how this works, I think this has not much hope to lead to a robust solution.

kwankyu:
Of this change `recreate: false`, I am not sure.

Before this change, I can easily see that the doc build workflow is completed since the comment reappears at the bottom of the thread but others get annoyed by the useless notifications in github inbox (assuming they have email filters).

After this change, I would be annoyed for the need to look into the Checks list to see that the doc build workflow is completed but others won't receive the useless notifications in github inbox. 

I vote -1 because I want this PR to get merged only if there are many people to care about the github inbox (I don't).

tornaria:
Before this change, I can easily see that the doc build workflow is completed since the comment reappears at the bottom of the thread but others get annoyed by the useless notifications in github inbox (assuming they have email filters).
> 
> After this change, I would be annoyed for the need to look into the Checks list to see that the doc build workflow is completed but others won't receive the useless notifications in github inbox.

Maybe it's possible to move up the "deploy documentation" so it's the first thing visible there ?

An advantage of the check box is that it will always be there as the last thing in the PR, while the comment by the bot gets lost if there are other comments after.

grhkm21:
> > [referer approach]
> 
> Browsers (and their privacy extensions) can decide not to send the referer at all or mangle it. From my understanding of how this works, I think this has not much hope to lead to a robust solution.

Will it be reasonable to have a giant <input> box just for the user to type the PR number in manually? I mean, if you click that "open PR" button or whatever, it'll open a new tab, so the GitHub PR page is still open, and you can copy the PR number over easily.

tornaria:
> > > [referer approach]
> > 
> > 
> > Browsers (and their privacy extensions) can decide not to send the referer at all or mangle it. From my understanding of how this works, I think this has not much hope to lead to a robust solution.
> 
> Will it be reasonable to have a giant box just for the user to type the PR number in manually? I mean, if you click that "open PR" button or whatever, it'll open a new tab, so the GitHub PR page is still open, and you can copy the PR number over easily.

Maybe we can use the referrer url, which will work for most people (e.g. for @kwankyu as shown in https://github.com/sagemath/sage/pull/37740#issuecomment-2037435731). But when the referrer url is mangled (for people like me, using privacy plugins) show a box to let me input the PR number.

There's possibly some (mild) correlation between people that get annoyed by spam bots and people that use privacy plugins. I wouldn't mind at all that the link in the PR comment doesn't work for me without manual intervention, typing the PR number is nil compared to receiving lots of spam from the bot which hides relevant notifications.

grhkm21:
That sounds great! Just to enlarge the sample size, referral doesn't show PR number for me.

dimpase:
+1 here

orlitzky:
+1

This is a simple solution that eliminates most of the noise.

For what it's worth, I blocked these in my mail system to see how hard it would be. It's not terribly easy, and I'm the mail server administrator. The notifications come from the usual github address, with all of the same headers that accompany any other comment. To dig into the body using server-side filtering, you need the [sieve body extension](https://datatracker.ietf.org/doc/html/rfc5173.txt). Then of course you want to test it by filtering the messages into junk for a while before you discard them entirely. All of this takes time, multiplied by the apparently large number of users who hate these notifications.

In the future, I would remain open to any of the better solutions that have been proposed but not implemented yet.

saraedum:
@dimpase @orlitzky thanks for your feedback. This still needs review (apart from votes.)

dimpase (approved):

grhkm21:
Wait, so are we disabling the notifications without a replacement (e.g. the redirection button or whatever)?

kwankyu:
Please do counting of votes before status labelling. That's the rule for disputed PRs.

dimpase:
@kwankyu this vote counting is getting ludicrous. As far as I know, @mkoeppe does not get email notifications from GitHub - and such people must be disqualified from voting here, as their mailboxes don't get a email notification from the same PR just cause docs got rebuilt.

dimpase:
> Wait, so are we disabling the notifications without a replacement (e.g. the redirection button or whatever)?

you'll get one notification, 1st time docs are/attempted to be built. You can revisit the PR if you want to see how it's going.

.github/workflows/doc-publish.yml
```diff
@@ -134,4 +135,3 @@ jobs:
       - name: Report deployment url
         run: |
           echo "::notice::The live documentation has been deployed - ${{ steps.deploy-netlify.outputs.NETLIFY_URL }}"
```
grhkm21, line 137: Can we also copy [this](https://github.com/sagemath/sage/pull/37387/files#diff-7282d89989d4658ef93d9d1933ae0079effcb523f9723c84130e0c1508d94992R84) over so that it updates the documentation comment with the link to the "changes" page?

tornaria:
+1: @saraedum @grhkm21 @dimpase @orlitzky @tornaria 
-1: @kwankyu

saraedum:
> @kwankyu this vote counting is getting ludicrous.

@dimpase if you want to change the process, I believe that sage-devel is the appropriate place to discuss this.

> such people must be disqualified from voting here

I strongly disagree with this. Everybody gets to vote.

grhkm21:
@tornaria i think mkoeppe is also -1, from one of the first comments. Maybe @mkoeppe can vote himself directly (I don’t think that’s required though?)

tornaria:
> @tornaria i think mkoeppe is also -1, from one of the first comments. Maybe @mkoeppe can vote himself directly (I don’t think that’s required though?)

I had initially placed him on the -1 side. However, I couldn't find his vote on this PR. He was tagged in the first comment, and also this PR was mentioned in the other PR he was participating in, so I would say he was given reasonable notification.

The vote would still be 5 to 2, so it wouldn't make a difference.

mkoeppe:
> As far as I know, @mkoeppe does not get email notifications from GitHub

@dimpase Well, how *would* you know this? 

As a matter of fact, I do get the email notifications, and they are exactly what I use in my notification-driven workflow. I said this clearly in the original discussion - https://github.com/sagemath/sage/pull/37387#issuecomment-2023870786

mkoeppe:
My vote is -1.

I have already explained clearly why in https://github.com/sagemath/sage/pull/37387#issuecomment-2036145172: "This proposal is not suitable as a compromise -- because it completely ignores the use of this notification as a workflow management tool. https://github.com/sagemath/sage/pull/37387#issuecomment-2023711674"

tobiasdiez:
+1, people interested in getting notifications when the doc workflow is finished can subscribe via built-in github functionality (https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs) or by quickly hacking together there own notification system using the public and easy-to-use github api (https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#about-workflow-runs-in-github-actions)

dimpase:
by the way, these notifications appear to be broken, e.g. I got https://github.com/dimpase/sage/pull/6#issuecomment-2050620587
long before docs actually got built, and the link provded there gives 404

dimpase:
> > As far as I know, @mkoeppe does not get email notifications from GitHub
> 
> @dimpase Well, how _would_ you know this?

that's what you suggested in a discussion around #36803 - to disable email notifications (IIRC). This assumes that you [eat your own dog food](https://en.wikipedia.org/wiki/Eating_your_own_dog_food), certainly.

saraedum:
> by the way, these notifications appear to be broken, e.g. I got [dimpase#6 (comment)](https://github.com/dimpase/sage/pull/6#issuecomment-2050620587) long before docs actually got built, and the link provded there gives 404

Most likely you do not have netlify credentials configured. When creating a PR into a fork where these are not configured, the notification gets created but is not functional. While that could be fixed, I don't really think it's relevant at the moment since I don't think anybody is creating PRs into other forks regularly.

PR #37747, opened 2024-04-04
be more specific when raising "not a constant polynomial"

mantepse:
Currently, if retracting a polynomial to its base ring fails, the error message is `TypeError: not a constant polynomial`.

In some circumstances, this is not very helpful, because it may be hard to find out where the nonconstant polynomial comes from.  In particular, this happens when the lazy implicit equations solver (#37033) fails.

Therefore, it seems sensible to add the offending polynomial.


mkoeppe:
doctest failure because one more test needs updating in: src/doc/en/thematic_tutorials/coercion_and_categories

mkoeppe (approved):
LGTM.

mantepse:
Thank you!

mantepse:
> Bit of bikeshedding, but I think `f is not a constant polynomial` would be better in terms of English. Not a big deal if you don't want to change it.

I actually put the polynomial last on purpose: it may be a long string.  I thought that in this case the error message will likely be easier to read if the polynomial comes last, because any linebreak will break the polynomial and not the `not a constant polynomial`.

mantepse:
OK

mantepse:
I included another unspecific error message in the same spirit.

mkoeppe:
Should we be concerned about performance implications if this ends up being called within the coercion framework on polynomials with many terms?

tscrim:
I don’t think we need to, but it wouldn’t hurt to make it a lazy string.

mantepse:
I am willign to do this, but on the condition of showing an example (even if artificial) where the performance degrades.  (I would then turn that example as a doctest.)

mkoeppe:
I don't have such tests at hand, so let's just merge it.

mkoeppe (approved):

vbraun:
fails as in the CI

mkoeppe (approved):

Issue #37734, opened 2024-04-03
Test modularised distributions causes CI failure 


GiacomoPope:
### Steps To Reproduce

Check the CI for recently run PR. For example: 

1. https://github.com/sagemath/sage/pull/37441
2. https://github.com/sagemath/sage/pull/37513
3. https://github.com/sagemath/sage/pull/37615

```
Run make V=0 tox && make pypi-wheels-check
make -j2 --output-sync=recurse build/make/Makefile --stop
make[1]: 'build/make/Makefile' is up to date.
build/bin/sage-logger \
	"cd build/make && ./install 'tox'" logs/install.log
Nothing to (re)build / all up-to-date.
make -j2 --output-sync=recurse build/make/Makefile --stop
make[1]: 'build/make/Makefile' is up to date.
build/bin/sage-logger \
	"cd build/make && ./install 'pypi-wheels-check'" logs/install.log
*** ALL ENVIRONMENT VARIABLES BEFORE BUILD: ***
sage_sws2rst-10.4.beta1
sage_setup-10.4.beta1
[sage_sws2rst-10.4.beta1] + set -e
[sage_sws2rst-10.4.beta1] + sage-sws2rst -h
sagemath_environment-10.4.beta1
sagemath_objects-10.4.beta1
make[1]: *** [Makefile:3584: sagemath_environment-SAGE_VENV-check] Error 1
make[1]: *** Waiting for unfinished jobs....
[sagemath_objects-10.4.beta1] ERROR: Could not find a version that satisfies the requirement cysignals>=1.10.2 (from versions: none)
[sagemath_objects-10.4.beta1] ERROR: No matching distribution found for cysignals>=1.10.2
[sagemath_objects-10.4.beta1] .pkg-sagepython-sagewheels-nopypi: remove tox env folder /sage/pkgs/sagemath-objects/.tox/.pkg-sagepython-sagewheels-nopypi
[sagemath_objects-10.4.beta1] .pkg-sagepython-sagewheels-nopypi: find interpreter for spec PythonSpec(path=/bin/python3)
[sagemath_objects-10.4.beta1] .pkg-sagepython-sagewheels-nopypi: proposed PythonInfo(spec=CPython3.8.10.final.0-64, exe=/bin/python3, platform=linux, version='3.8.10 (default, Nov 22 2023, 10:22:35) \n[GCC 9.4.0]', encoding_fs_io=utf-8-utf-8)
[...]

PR #37748, opened 2024-04-04
dft for symmetric group algebra when p|n!

jacksonwalters:
The DFT for the SGA throws a ZeroDivisionError when p|n. Uses the idempotents as computed in [Murphy '83] to build the Pierce decomposition of the symmetric group algebra over a finite field. There is a homomorphic projection onto each block via v |--> v*e_i. The change of basis matrix is the modular Fourier transform, and works even when p|n.

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/symmetric_group_algebra.py
+++ b/src/sage/combinat/symmetric_group_algebra.py
@@ -1950,8 +1950,8 @@ class SymmetricGroupAlgebra_n(GroupAlgebra_class):
 
         EXAMPLES::
 
-            sage: GF3S3 = SymmetricGroupAlgebra(GF(2),3)
-            sage: GF3S3._dft_modular()
+            sage: GF2S3 = SymmetricGroupAlgebra(GF(2),3)
+            sage: GF2S3._dft_modular()
             [1 0 0 0 1 0]
             [0 1 0 0 0 1]
             [0 0 1 0 0 1]
@@ -1977,7 +1977,7 @@ class SymmetricGroupAlgebra_n(GroupAlgebra_class):
             for pair in list(b):
                 coord_vector[sym_group_list.index(pair[0])] = pair[1]
             change_of_basis_matrix.append(coord_vector)
-        return Matrix(self.base_ring(),change_of_basis_matrix)
+        return matrix(self.base_ring(),change_of_basis_matrix)
 
     def epsilon_ik(self, itab, ktab, star=0, mult='l2r'):
         r"""
--- a/src/sage/combinat/symmetric_group_algebra.py
+++ b/src/sage/combinat/symmetric_group_algebra.py
@@ -1977,7 +1977,7 @@ class SymmetricGroupAlgebra_n(GroupAlgebra_class):
             for pair in list(b):
                 coord_vector[sym_group_list.index(pair[0])] = pair[1]
             change_of_basis_matrix.append(coord_vector)
-        return matrix(self.base_ring(),change_of_basis_matrix)
+        return matrix(self.base_ring(),change_of_basis_matrix).transpose()
 
     def epsilon_ik(self, itab, ktab, star=0, mult='l2r'):
         r"""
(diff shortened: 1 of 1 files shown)


jacksonwalters:
doctest is currently failing. I think that has something to do with the preloaded examples in the comments. I am attempting to build sage locally for the first time to run doctest with the changed `symmetric_group_algebra.py` file.

jacksonwalters:
the output in the example I had was wrong. that's been corrected. now test modularized distributions is failing with:

[sagemath_objects-10.4.beta1] ERROR: Could not find a version that satisfies the requirement cysignals>=1.10.2 (from versions: none)

jacksonwalters:
https://github.com/sagemath/sage/pull/37750

jacksonwalters:
https://github.com/sagemath/sage/issues/37734

jacksonwalters:
https://github.com/sagemath/sage/issues/37751

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -1909,12 +1909,14 @@ def dft(self, form="seminormal", mult='l2r'):
         """
         if form == "seminormal":
             return self._dft_seminormal(mult=mult)
```
S17A05: It might be a good idea to catch the error in the case $p \mid n$ here and raise an error with a clearer error message, or to remove the standard argument for `form` and choose it depending on whether the characteristic divides $n$ (depending on whether `._dft_seminormal` always fails in the modular case)

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -1941,6 +1943,42 @@ def _dft_seminormal(self, mult='l2r'):
         snb = self.seminormal_basis(mult=mult)
         return matrix([vector(b) for b in snb]).inverse().transpose()
 
+    def _dft_modular(self):
+        """
+        Return the discrete Foruier transform when the characterisc divides the order of the group.
+        The usual .dft() throws ZeroDivisionError when p|n.
```
S17A05: I'd remove the reference to the `.dft()`-method here; there's also a small typo.
```suggestion
        Return the discrete Foruier transform when the characteristic divides the order of the group.
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -1941,6 +1943,42 @@ def _dft_seminormal(self, mult='l2r'):
...
+        EXAMPLES::
+
+            sage: GF2S3 = SymmetricGroupAlgebra(GF(2),3)
+            sage: GF2S3._dft_modular()
+            [1 0 0 0 1 0]
+            [0 1 0 0 0 1]
+            [0 0 1 0 0 1]
+            [0 0 0 1 1 0]
+            [1 0 0 1 1 0]
+            [0 1 1 0 0 1]
```
S17A05: This example should go into the `.dft`-method, so that it shows up in the documentation, with the small change given below:

```suggestion
            sage: GF2S3 = SymmetricGroupAlgebra(GF(2),3)
            sage: GF2S3.dft(form="modular")
            [1 0 0 0 1 0]
            [0 1 0 0 0 1]
            [0 0 1 0 0 1]
            [0 0 0 1 1 0]
            [1 0 0 1 1 0]
            [0 1 1 0 0 1]
```

S17A05 (commented):
Admittedly I don't know enough about modular representation theory to check this code mathematically, but here are some programming comments.

Also, the documentation for the `.dft`-method should be rewritten a bit to explain the `form` argument, to make the explanation of `mult` dependent on `form = "seminormal"`, and to also explain the different approaches used depending on `form`. For the explanation of the modular approach, you can also refer to [Murphy '83], which needs to be formatted as `[Mur1983]_` in the docstring to give a hyperlink in the documentation.

src/sage/combinat/symmetric_group_algebra.py
jacksonwalters: I agree, at the very least the failing cases p|n should at least be handled with a clear error message. I believe it only fails when p | n! (n was a type - it's the order of the symmetric group, so n!). I've chosen to just remove the standard argument for form and return the modular DFT in that case.

jacksonwalters:
Accidentally closed this PR when I deleted the fork on my end. Should be reopened.

PR #37761, opened 2024-04-06
rings/infinite polynomial ring fixes

### LINKED ISSUE #37756: silently wrong result in coercion between InfinitePolynomialRing and LazyPowerSeriesRing
### Steps To Reproduce

```
sage: L.<x, y> = LazyPowerSeriesRing(QQ)
sage: R.<a> = InfinitePolynomialRing(QQ)
sage: o = L(0); r = a[0]
sage: o + r
y
sage: _.parent()
Infinite polynomial ring in a over Multivariate Lazy Taylor Series Ring in x, y over Rational Field
```

### Expected Behavior

The result should either be `a[0]` (ideally with parent `LazyPowerSeriesRing(InfinitePolynomialRing(QQ))`) or an error should be raised.

### Actual Behavior

A nonsense result, 

### Additional Information

The parent is explained as follows:
```
sage: cm.explain(o, r, op="add")
Coercion on left operand via
    Generic morphism:
      From: Multivariate Lazy Taylor Series Ring in x, y over Rational Field
      To:   Infinite polynomial ring in a over Multivariate Lazy Taylor Series Ring in x, y over Rational Field
Coercion on right operand via
    Coercion map:
      From: Infinite polynomial ring in a over Rational Field
      To:   Infinite polynomial ring in a over Multivariate Lazy Taylor Series Ring in x, y over Rational Field
Arithmetic performed after coercions.
Result lives in Infinite polynomial ring in a over Multivariate Lazy Taylor Series Ring in x, y over Rational Field
Infinite polynomial ring in a over Multivariate Lazy Taylor Series Ring in x, y over Rational Field
```

It is unclear to me, why we would want to have
```
sage: L.construction()[0]
Completion[('x', 'y'), prec=+Infinity]
sage: _.rank
4
sage: R.construction()[0]
InfPoly{[a], "lex", "dense"}
sage: _.rank
9.5
```

mantepse:
Fix #37756


src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -560,6 +560,44 @@ def is_nilpotent(self):
...
         return self._p.is_nilpotent()
 
+    def numerator(self):
+        r"""
+        Return a numerator of ``self``, computed as ``self * self.denominator()``.
+
+        Note that some subclasses may implement its own numerator
+        function.
+
+        .. warning::
```
tscrim:
```suggestion
        .. WARNING::
```

src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -578,9 +616,70 @@ def variables(self):
...
+            sage: p.monomials()
+            [x_1^3, x_3*x_1, x_2]
+        """
+        P = self.parent()
+        return [InfinitePolynomial(P, m) for m in self._p.monomials()]
+
+    def monomial_coefficient(self, mon):
+        """
+        Return the coefficient in the base ring of the monomial mon in
+        ``self``, where mon must have the same parent as self.
```
tscrim:
```suggestion
        Return the coefficient in the base ring of the monomial ``mon`` in
        ``self``, where ``mon`` must have the same parent as ``self``.
```

src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -578,9 +616,70 @@ def variables(self):
...
+
+        This function contrasts with the function ``coefficient``
+        which returns the coefficient of a monomial viewing this
+        polynomial in a polynomial ring over a base ring having fewer
+        variables.
+
+        INPUT:
+
+        - ``mon`` - a monomial
```
tscrim:
```suggestion
        - ``mon`` -- a monomial
```

src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -1018,42 +1117,42 @@ def coefficient(self, monomial):
...
         """
+        P = self.parent()
         if self._p == 0:
-            res = 0
-        elif isinstance(monomial, self.__class__):
-            if not (self.parent().has_coerce_map_from(monomial.parent())):
-                res = 0
+            return P.zero()
+        if isinstance(monomial, self.__class__):
+            if not (P.has_coerce_map_from(monomial.parent())):
```
tscrim:
```suggestion
            if not P.has_coerce_map_from(monomial.parent()):
```
https://xkcd.com/297/ :)

src/sage/rings/polynomial/infinite_polynomial_ring.py
```diff
@@ -922,7 +931,7 @@ def _element_constructor_(self, x):
             if isinstance(self._base, MPolynomialRing_polydict):
                 x = sage_eval(repr(), next(self.gens_dict()))
             else:
-                x = self._base(x)
+                x = self._base.coerce(x)
```
tscrim: I find this change to be slightly dangerous. The element construction is the conversion, but now you are requiring the input to satisfy a coercion into the base. So it might not be able to convert rational numbers into integers (e.g., `2/1`). Please add some appropriate examples showing that `InfinitePolynomialRing(ZZ, 'x')` works as expected with integers realized as rational numbers and similar polynomials.

src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -578,9 +616,70 @@ def variables(self):
...
+
+            sage: X.<x> = InfinitePolynomialRing(QQ, order='deglex')
+            sage: p = x[1]^3 + x[2] - 2*x[1]*x[3]
+            sage: p.monomials()
+            [x_1^3, x_3*x_1, x_2]
+        """
+        P = self.parent()
+        return [InfinitePolynomial(P, m) for m in self._p.monomials()]
+
+    def monomial_coefficient(self, mon):
```
tscrim, line 633: Are you sure we need this method? Shouldn't the default redirection in`__getattr__` handle this?

src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -560,6 +560,44 @@ def is_nilpotent(self):
...
+        Note that some subclasses may implement its own numerator
+        function.
+
+        .. warning::
+
+           This is not the numerator of the rational function
+           defined by ``self``, which would always be self since ``self`` is a
+           polynomial.
+
+        EXAMPLES:
```
tscrim:
```suggestion
        EXAMPLES::
```

tscrim (changes requested):

src/sage/rings/polynomial/infinite_polynomial_element.py
mantepse: that was not me!  (and lisp always has the function first)!  :-)

src/sage/rings/polynomial/infinite_polynomial_ring.py
mantepse: Do you meant as follows:
```
             sage: R = InfinitePolynomialRing(ZZ, "a")
             sage: R(GF(5)(2))
             2
```
?  (There is something odd, because the 2 above has parent `ZZ`,  not `InfinitePolynomialRing(ZZ, "a")`.

However, the change to `x = self._base.coerce(x)` is actually what resolved the bug
```
            sage: L.<x, y> = QQ[]
            sage: R.<a> = InfinitePolynomialRing(QQ)
            sage: M = InfinitePolynomialRing(L, names=["a"])
            sage: c = a[0]
            sage: M(c)
            a_0
```
Moreover, this is also done in `MPolynomialRing_polydict.__call__`.

src/sage/rings/polynomial/infinite_polynomial_element.py
mantepse: It seems so.  Without the method `monomial_coefficient` I have:
```
sage: X.<x> = InfinitePolynomialRing(QQ)
sage: f = 2*x[0]*x[2] + 3*x[1]^2
sage: c = f.monomial_coefficient(x[1]^2); c
Cell In[4], line 1
----> 1 c = f.monomial_coefficient(x[Integer(1)]**Integer(2)); c

TypeError: Argument 'mon' has incorrect type (expected sage.rings.polynomial.multi_polynomial_libsingular.MPolynomial_libsingular, got InfinitePolynomial_dense)
```

src/sage/rings/polynomial/infinite_polynomial_ring.py
tscrim: Before your changes, we have this:
```
sage: R = InfinitePolynomialRing(ZZ, "a")
sage: R(QQ(2)).parent()
Infinite polynomial ring in a over Integer Ring
sage: R(GF(5)(2)).parent()
Infinite polynomial ring in a over Integer Ring
```
The base ring can do these conversions, so the polynomial ring should also do these conversions:
```
sage: R.<x> = ZZ[]
sage: R(QQ(2)).parent()
Univariate Polynomial Ring in x over Integer Ring
sage: R(GF(5)(2)).parent()
Univariate Polynomial Ring in x over Integer Ring
sage: R.<x,y> = ZZ[]
sage: R(QQ(2)).parent()
Multivariate Polynomial Ring in x, y over Integer Ring
sage: R(GF(5)(2)).parent()
Multivariate Polynomial Ring in x, y over Integer Ring
```
However, these fail (as they should!):
```
sage: ZZ.coerce(GF(5)(2))
sage: ZZ.coerce(QQ(5)(2))
```
So this change at least completely changes how these are handled, but they should be processed here. My guess is something less straightforward with a conversion-without-coercion that cannot be done using string representations will break.

Note that the example works without your changes once the internal poly ring of has more variables than `L`.

We probably need to check the input is a "compatible" infinite polynomial ring before trying to see if it belongs to the base ring. I am thinking compatible means the generator names are a subset of `self`'s. Likely a complete rewrite of `_element_constructor_` is needed.

This might work as a short-term hack, but it is 100% a hack solution IMO.

src/sage/rings/polynomial/infinite_polynomial_element.py
tscrim: Ah, right, yes, we need to extract out the underlying polynomial.

src/sage/rings/polynomial/infinite_polynomial_ring.py
mantepse: I agree that a complete rewrite of `_element_constructor_` is needed, but I'm a bit afraid that this is not easy.  In particular, comments like "it's even more of a shame that MPolynomialRing_polydict does not work in complicated settings" don't help improving my confidence.

tscrim: That might be the next thing we try to do after we get the functional solver working. As such, I can accept this if we put a (code) comment saying this should instead do a conversion instead of a coercion as a `# FIXME`.

Also note that I don't think this works if I want to be evil and have my base ring have variable names the same as the inf poly ring.

tscrim: As mentioned in our recent email exchange, we've effectively found an example where this change produces undesirable results.

mantepse: I guess it would be good to make this example available somehow for the afterworld, I am not extremely motivated, though.

mantepse:
I reverted the original fix, and left a warning.  I don't have the knowledge to fix this properly.  However, the remaining fixes are good for #38108 (superseding #37033).

mantepse:
@tscrim, anything else you would like me to do here?

src/sage/rings/polynomial/infinite_polynomial_ring.py
```diff
@@ -888,6 +888,17 @@ def _element_constructor_(self, x):
             Traceback (most recent call last):
             ...
             ValueError: cannot convert 1/3 into an element of Infinite polynomial ring in x over Integer Ring
+
+        .. WARNING:
+
+        The :issue:`37756` is not yet fixed::
+
+            sage: L.<x, y> = QQ[]
```
tscrim:
```suggestion
        .. WARNING::

            The :issue:`37756` is not yet fixed::

                sage: L.<x, y> = QQ[]
```

src/sage/rings/polynomial/infinite_polynomial_element.py
```diff
@@ -577,9 +612,70 @@ def variables(self):
...
+            sage: c.parent()
+            Rational Field
+
+            sage: c = f.coefficient(x[2]); c
+            2*x_0
+            sage: c.parent()
+            Infinite polynomial ring in x over Rational Field
+
+        """
+        return self._p.monomial_coefficient(mon._p)
```
tscrim:
```suggestion
            Infinite polynomial ring in x over Rational Field
        """
        return self._p.monomial_coefficient(mon._p)
```

tscrim (changes requested):
Some last little details. Otherwise LGTM.

src/sage/rings/polynomial/infinite_polynomial_ring.py
```diff
@@ -884,6 +884,17 @@ def _element_constructor_(self, x):
...
+        .. WARNING::
+
+            The :issue:`37756` is not yet fixed::
+
+                sage: L.<x, y> = QQ[]
+            sage: R.<a> = InfinitePolynomialRing(QQ)
+            sage: M = InfinitePolynomialRing(L, names=["a"])
+            sage: c = a[0]
+            sage: M(c)  # known bug
+            a_0
```
tscrim:
```suggestion
                sage: R.<a> = InfinitePolynomialRing(QQ)
                sage: M = InfinitePolynomialRing(L, names=["a"])
                sage: c = a[0]
                sage: M(c)  # known bug
                a_0
```
The rest of these need to be adjusted too; sorry I wasn't explicit about it.

mantepse: I'm sorry, that was my fault.

tscrim (approved):
Thanks. Let's get this in.

PR #37526, opened 2024-03-02
Fix `monomial_coefficients` for submodules of free modules from sage.modules

mkoeppe:
- Rebased version of #34455
- Part of https://github.com/sagemath/sage/issues/30309

Fixes #34455

### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37766, opened 2024-04-08
modules with basis/map coefficients

### LINKED ISSUE #18264: make map_coefficients in modules_with_basis safer or more general

mantepse:
This is a replacement for #37271, without dependency on the construction functor for symmetric functions.

Fixes #18264


src/sage/categories/modules_with_basis.py
```diff
@@ -2062,17 +2063,25 @@ def trailing_term(self, *args, **kwds):
...
+            ``new_base_ring`` to override this behaviour.
+
+            An error is raised if the coefficients are not in the new base ring.
 
             INPUT:
 
-            - ``f`` -- an endofunction on the coefficient ring of the
-              free module
+            - ``f`` -- a callable that will be applied to the
+              coefficients of ``self``.
```
tscrim:
```suggestion
            - ``f`` -- a callable that will be applied to the
              coefficients of ``self``
```

src/sage/categories/modules_with_basis.py
```diff
@@ -2062,17 +2063,25 @@ def trailing_term(self, *args, **kwds):
...
 
-            Return a new element of ``self.parent()`` obtained by applying the
-            function ``f`` to all of the coefficients of ``self``.
+            - ``new_base_ring`` (optional) -- if given, the resulting element
+              will be defined over this ring.
```
tscrim:
```suggestion
            - ``new_base_ring`` -- (optional) if given, the resulting element
              will be defined over this ring
```

src/sage/categories/modules_with_basis.py
```diff
@@ -2062,17 +2063,25 @@ def trailing_term(self, *args, **kwds):
...
-            Mapping a function on coefficients.
+            Return the element obtained by applying ``f`` to the non-zero
+            coefficients of ``self``.
+
+            If ``f`` is a :class:`sage.categories.map.Map`, then the resulting
+            polynomial will be defined over the codomain of ``f``. Otherwise, the
+            resulting polynomial will be over the same ring as ``self``. Set
+            ``new_base_ring`` to override this behaviour.
+
+            An error is raised if the coefficients are not in the new base ring.
```
tscrim: As written, I don't think we want this error to be raised if the callable returns elements in, say `ZZ` but `new_base_ring=GF(2)`. We want to automatically convert the elements for consistency:
```
sage: R.<x,y> = ZZ[]
sage: x.map_coefficients(lambda p: 2*p, GF(2))
0
sage: x.map_coefficients(lambda p: GF(2)(p), QQ)
x
```
I think what you meant to say is if the coefficients must be able to be converted into `new_base_ring`.

Additionally, if `f` is a map, then `new_base_ring` should still be the ultimate target ring for consistency:
```
sage: R.<a> = QQ[]
sage: f = R.hom([a^2])
sage: S.<x,y> = R[]
sage: (a*x).map_coefficients(f)
a^2*x
sage: (a*x).map_coefficients(f, R['q'])
a^2*x
sage: _.parent()
Multivariate Polynomial Ring in x, y over Univariate Polynomial Ring in q over Univariate Polynomial Ring in a over Rational Field
```

src/sage/categories/modules_with_basis.py
```diff
@@ -2096,8 +2105,32 @@ def map_coefficients(self, f):
...
+            R = self.parent()
+            if new_base_ring is not None:
+                B = new_base_ring
+                R = R.change_ring(B)
+            elif isinstance(f, Map):
+                B = f.codomain()
+                R = R.change_ring(B)
+            else:
+                B = self.base_ring()
+            return R.sum_of_terms((m, B(f(c))) for m, c in self)
```
tscrim:
```suggestion
            mc = self.monomial_coefficients(copy=False)
            return R.sum_of_terms((m, B(f(c))) for m, c in mc.items())
```
We should not assume specific `__iter__` output.

src/sage/categories/modules_with_basis.py
```diff
@@ -2096,8 +2105,32 @@ def map_coefficients(self, f):
...
+            """
+            R = self.parent()
+            if new_base_ring is not None:
+                B = new_base_ring
+                R = R.change_ring(B)
+            elif isinstance(f, Map):
+                B = f.codomain()
+                R = R.change_ring(B)
+            else:
+                B = self.base_ring()
```
tscrim:
```suggestion
            if isinstance(f, Map):
                B = f.codomain()
            else:
                B = self.base_ring()
            if new_base_ring is not None:
                B = new_base_ring
            if B is not self.base_ring():
                R = R.change_ring(B)
```
To be consistent with what (multivariate) polynomials do.

tscrim (changes requested):
Why is #37687 a dependency? This seems completely independent.

mantepse:
> Why is #37687 a dependency? This seems completely independent.

The doctest only works with #37687.

tscrim:
It would be better to use a doctest that does not depend on #37687. (In fact, this change is very mildly bad because it depends on something that is not implemented in the category.) Just use a regular CFM:
```
sage: M = CombinatorialFreeModule(ZZ, 'abc')
```

src/sage/categories/modules_with_basis.py
mantepse: Do I understand correctly that you are asking me to adapt the documentation, the code is doing what it should, right?

mantepse:
> It would be better to use a doctest that does not depend on #37687. (In fact, this change is very mildly bad because it depends on something that is not implemented in the category.) Just use a regular CFM:
> 
> ```
> sage: M = CombinatorialFreeModule(ZZ, 'abc')
> ```

OK

src/sage/categories/modules_with_basis.py
tscrim: Yes, that's correct.

tscrim:
Yes, that's what I was thinking. You don't need a new PR (it would make it *much* easier for me if you didn't); you can interactive rebase and just remove the commits from #37687 (among other ways) and force push the result.

mantepse:
need more help with removing the offending commits, sorry.

mantepse:
success!

src/sage/categories/modules_with_basis.py
```diff
@@ -2062,17 +2063,26 @@ def trailing_term(self, *args, **kwds):
             """
             return self.parent().term(*self.trailing_item(*args, **kwds))
 
-        def map_coefficients(self, f):
+        def map_coefficients(self, f, new_base_ring=None):
             """
-            Mapping a function on coefficients.
+            Return the element obtained by applying ``f`` to the non-zero
+            coefficients of ``self``.
+
```
mkoeppe, line 2071: This method will need clarification because of:
- #37526

mantepse: Could you please elaborate?  I don't see a change to `map_coefficients` in #37526, nor a deprecation.

mantepse: ping?

mkoeppe: Sorry for the delay. I was concerned regarding the meaning of "coefficients" when submodules are involved.
The code looks fine -- as it's just mapping monomial coefficients to monomial coefficients.

mkoeppe (approved):
LGTM

tscrim (approved):

PR #37770, opened 2024-04-09
implement matrix method `is_symplectic`

grhkm21:
It just calls the `contain` method of the existing `Sp` (symplectic matrices) class


tscrim:
On its own, this is not well-defined. The group $Sp_{2n}(F)$ secretly has a fixed skew-symmetric form that defines the set (although there is a typical one, almost canonical, it is not universal). For example, if we take the matrices
$$\Omega = \begin{bmatrix} &&&1 \\ &&&1 \\-1&&& \\ &-1&&\end{bmatrix}, \Theta = \begin{bmatrix}&&1& \\ -1&&& \\ &&&1 \\ &&-1&\end{bmatrix}$$
which define two bilinear forms, then we get different symplectic matrices, i.e., that satisfy $M^T \Omega M = \Omega$ versus $M^T \Theta M = \Theta$. In particular:
```
sage: Omega = matrix([[0,0,1,0],[0,0,0,1],[-1,0,0,0],[0,-1,0,0,]]); Omega
[ 0  0  1  0]
[ 0  0  0  1]
[-1  0  0  0]
[ 0 -1  0  0]
sage: Theta = matrix([[0,1,0,0],[-1,0,0,0],[0,0,0,1],[0,0,-1,0]]); Theta
[ 0  1  0  0]
[-1  0  0  0]
[ 0  0  0  1]
[ 0  0 -1  0]
sage: M = matrix([[0,1,0,0],[1,0,0,0],[0,0,0,1],[0,0,1,0]])  # permutation (1 2)(2 4) matrix
sage: M.det()
1
sage: M.transpose() * Omega * M == Omega
True
sage: M.transpose() * Theta * M == Theta
False
```
So I think you need an extra option to specify the skew-symmetric bilinear form's Gram matrix. There can be a default of that of the `Sp` group, which is $\sum_{i=1}^n E_{i,2n+1-i} - E_{2n+1-i,i}$. However, all of this needs to be explained in the docstring.

grhkm21:
I see (FYI I don't know the theory of symplectic *things* at all). I will add an argument `invariant_form=...` (that's what it's called for `Sp(...)`, and also note the default one in the docstring.

mkoeppe:
Note also `sage.modules.free_quadratic_module`, which surprisingly makes no mention of the symplectic case.

grhkm21:
> Note also `sage.modules.free_quadratic_module`, which surprisingly makes no mention of the symplectic case.

How should this be incorporated to this PR? Should it also accept a quadratic form / modules rather than just matrices? I'm not too familiar with them though.

grhkm21:
`__contains__` is broken lol, give me a second

src/sage/matrix/matrix0.pyx
```diff
@@ -4546,6 +4546,49 @@ cdef class Matrix(sage.structure.element.Matrix):
...
+            sage: M.is_symplectic(invariant_form=E)
+            True
+            sage: M * E * M.T == E
+            True
+        """
+        if self._ncols != self._nrows:
+            raise ValueError("the matrix is not a square matrix")
+
+        from sage.groups.matrix_gps.symplectic import Sp
+        return self in Sp(self._nrows, self._base_ring, invariant_form=invariant_form)
```
tscrim:
```suggestion
        if invariant_form is None:
            from sage.groups.matrix_gps.symplectic import Sp
            invariant_form = Sp(self._nrows, self._base_ring).invariant_form()
        return self * invariant_form * self.transpose() == invariant_form
```

tscrim (changes requested):
I propose changing it so that `Sp._check_matrix` calls 
```
M.is_symplectic(self.invariant_form())
```
This should allow you to remove the `__contains__`, where the generic one seems to be handling things better as yours seems to be causing tests to break in `src/sage/algebras/iwahori_hecke_algebra.py`. I think this is the more natural programming/mathematical flow.

grhkm21:
sorry for the delay

src/sage/matrix/matrix0.pyx
```diff
@@ -4561,6 +4561,51 @@ cdef class Matrix(sage.structure.element.Matrix):
             raise ValueError("The matrix is not a square matrix")
         return self._check_symmetrizability(return_diag=return_diag, skew=True, positive=positive)
 
+    def is_symplectic(self, invariant_form=None):
+        r"""
+        Return ``True`` if this matrix `M` is symplectic with respect to the
+        invariant form `E`.
```
tscrim, line 4567:
```suggestion
        Return whether ``self`` is symplectic with respect to the given
        invariant form.
```

tscrim (approved):
Thank you. One last little docstring thing.

PR #37771, opened 2024-04-09
some C4 fixes in combinat folder

fchapoton:
Just fixing a few of the `ruff --select=C4` suggestions in the `combinat` folder

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/sloane_functions.py
+++ b/src/sage/combinat/sloane_functions.py
@@ -859,7 +859,7 @@ class A003418(SloaneSequence):
             sage: [sloane.A003418._eval(n) for n in range(1,11)]
             [1, 2, 6, 12, 60, 60, 420, 840, 2520, 2520]
         """
-        return arith.lcm([i for i in range(1, n+1)])
+        return arith.lcm(list(range(1, n + 1)))
 
 
 class A007318(SloaneSequence):
--- a/src/sage/combinat/sloane_functions.py
+++ b/src/sage/combinat/sloane_functions.py
@@ -859,7 +859,7 @@ class A003418(SloaneSequence):
             sage: [sloane.A003418._eval(n) for n in range(1,11)]
             [1, 2, 6, 12, 60, 60, 420, 840, 2520, 2520]
         """
-        return arith.lcm(list(range(1, n + 1)))
+        return arith.lcm(range(1, n + 1))
 
 
 class A007318(SloaneSequence):
--- a/src/sage/combinat/tiling.py
+++ b/src/sage/combinat/tiling.py
@@ -1811,13 +1811,10 @@ class TilingSolver(SageObject):
             True
             sage: all(B[A[i]] == i for i in A)
             True
-
         """
         if self._reusable:
             return dict(enumerate(self.space()))
-
-        number_of_pieces = len(self._pieces)
-        return {i + number_of_pieces: c for i, c in enumerate(self.space())}
+        return dict(enumerate(self.space(), start=len(self._pieces)))
 
     @cached_method
     def rows_for_piece(self, i, mod_box_isometries=False):
--- a/src/sage/combinat/tiling.py
+++ b/src/sage/combinat/tiling.py
@@ -991,12 +991,13 @@ class Polyomino(SageObject):
         """
         if mod_box_isometries:
             L = ncube_isometry_group_cosets(self._dimension, orientation_preserving)
-            P_cosets = set(frozenset((m * self).canonical() for m in coset) for coset in L)
+            P_cosets = {frozenset((m * self).canonical() for m in coset)
+                        for coset in L}
             P_cosets_representents = [min(s, key=lambda a: a.sorted_list()) for s in P_cosets]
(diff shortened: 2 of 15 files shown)


src/sage/combinat/sloane_functions.py
```diff
@@ -859,7 +859,7 @@ def _eval(self, n):
             sage: [sloane.A003418._eval(n) for n in range(1,11)]
             [1, 2, 6, 12, 60, 60, 420, 840, 2520, 2520]
         """
-        return arith.lcm([i for i in range(1, n+1)])
+        return arith.lcm(list(range(1, n + 1)))
```
mkoeppe: Shouldn't we rather fix the documentation of `sage.arith.functions.lcm` to clarify that the argument can be any iterable, not just a list or tuple?

src/sage/combinat/tiling.py
```diff
@@ -1814,11 +1814,10 @@ def int_to_coord_dict(self):
...
         if self._reusable:
-            return dict((i, c) for i, c in enumerate(self.space()))
-        else:
-            number_of_pieces = len(self._pieces)
-            return dict((i+number_of_pieces, c)
-                        for i, c in enumerate(self.space()))
+            return dict(enumerate(self.space()))
+
+        number_of_pieces = len(self._pieces)
+        return {i + number_of_pieces: c for i, c in enumerate(self.space())}
```
mkoeppe:
```suggestion
        return dict(enumerate(self.space(), start=number_of_pieces))
```

src/sage/combinat/words/morphism.py
```diff
@@ -1727,9 +1729,8 @@ def is_uniform(self, k=None):
             True
         """
         if k is None:
-            return len(set(w.length() for w in self.images())) == 1
-        else:
-            return all(w.length() == k for w in self.images())
+            k = self.images()[0].length()
```
mkoeppe: is there always at least one element?

fchapoton: no, indeed

src/sage/combinat/words/morphism.py
```diff
@@ -1725,11 +1727,19 @@ def is_uniform(self, k=None):
             False
             sage: tau.is_uniform(k=2)
             True
+
+        TESTS:
```
mkoeppe:
```suggestion
        TESTS::
```

src/sage/combinat/words/morphism.py
```diff
@@ -1725,11 +1727,19 @@ def is_uniform(self, k=None):
...
+
+            sage: phi = WordMorphism('')
+            sage: phi.is_uniform()
+            False
```
mkoeppe: I know that this is the existing behavior, but is it correct? After all, we have
```
sage: phi.is_uniform(0)
True
```

src/sage/combinat/words/morphism.py
```diff
@@ -1725,11 +1727,19 @@ def is_uniform(self, k=None):
...
         """
         if k is None:
-            return len(set(w.length() for w in self.images())) == 1
-        else:
-            return all(w.length() == k for w in self.images())
+            try:
+                k = self.images()[0].length()
+            except IndexError:
+                return False
```
mkoeppe: Do we consider these 4 lines an improvement over the 1 line that was there before?

Taking into account my comment above, I would just suggest:
```
        if k is None:
            return len(set(w.length() for w in self.images())) <= 1
```

src/sage/combinat/words/morphism.py
fchapoton: indeed, I changed the answer to True

src/sage/combinat/words/morphism.py
fchapoton: This avoids to compute all of them in case of failure (similar to `all` and `any`).

mkoeppe: Thanks, I missed that.

src/sage/combinat/words/morphism.py
```diff
@@ -1725,11 +1728,19 @@ def is_uniform(self, k=None):
...
         """
         if k is None:
-            return len(set(w.length() for w in self.images())) == 1
-        else:
-            return all(w.length() == k for w in self.images())
+            try:
+                k = self.images()[0].length()
+            except IndexError:
+                return True
+        return all(w.length() == k for w in self.images())
```
mkoeppe, line 1743: How about rewriting it using an iterator? Just bikeshedding, feel free to ignore.
```
images = iter(self.images())
if k is None:
    try:
        k = next(images).length()
    except StopIteration:
        return True
return all(w.length() == k for w in images)

mkoeppe (approved):

src/sage/combinat/words/morphism.py
fchapoton: well, this is a list, so there is no point, I think

mkoeppe: Fine

PR #37773, opened 2024-04-09
Allow construction of classical Lie algebras over finite fields

tscrim:
Many classical Lie algebras are unable to be constructed because they multiply a rational number times an element in the base ring. If the base ring is, say, `GF(3)`, then there is no coercion (since not every element is defined), so we need to force all elements to be in the base ring during the computation (which might perform a division that goes to the fraction field, but it will eventually brings us back to a $\mathbb{Z}$-basis).

Before, this would fail with a `QQ` to `GF(3)` coercion error:
```
sage: sl4 = lie_algebras.sl(GF(3), 4)
```

### :memo: Checklist

### DIFF
--- a/src/sage/algebras/lie_algebras/classical_lie_algebra.py
+++ b/src/sage/algebras/lie_algebras/classical_lie_algebra.py
@@ -1700,6 +1700,26 @@ class LieAlgebraChevalleyBasis(LieAlgebraWithStructureCoefficients):
             [((alpha[1], -alpha[1]), {alphacheck[1]: 1}),
              ((alpha[1], alphacheck[1]), {alpha[1]: -2}),
              ((alphacheck[1], -alpha[1]), {-alpha[1]: -2})]
+
+        TESTS:
+
+        Check that we can construct Lie algebras over positive characteristic
+        fields (:issue:`37773`)::
+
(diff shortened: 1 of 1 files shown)


src/sage/algebras/lie_algebras/classical_lie_algebra.py
```diff
@@ -1700,6 +1700,26 @@ def _construct_struct_coeffs(self, R, p_roots):
             [((alpha[1], -alpha[1]), {alphacheck[1]: 1}),
              ((alpha[1], alphacheck[1]), {alpha[1]: -2}),
              ((alphacheck[1], -alpha[1]), {-alpha[1]: -2})]
+
+        TESTS:
+
```
fchapoton, line 1705: maybe add a word of caution for GF(2) and a failing example in that case ?

tscrim: That's a good point. It actually should work for `GF(2)`. So I implemented a better algorithm using the fact that the Chevalley basis is a $\mathbb{Z}$-basis of the Lie algebra.

tscrim:
This now works over `GF(2)` in my test cases and added some checks against some known results.

fchapoton (approved):
ok, good.

tscrim:
Merci!

PR #37778, opened 2024-04-10
Make univariate Laurent polynomials over a field pass the TestSuite

tscrim:
The output of `xgcd` does not match the `gcd` output, but this is required by #17671. We change the computation to make these match. We also fix another bug as it should also take non-Laurent polynomial inputs to `xgcd`.

We also implement the `euclidean_domain()` method.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/polynomial/laurent_polynomial.pyx
+++ b/src/sage/rings/polynomial/laurent_polynomial.pyx
@@ -1318,22 +1318,59 @@ cdef class LaurentPolynomial_univariate(LaurentPolynomial):
             return ~self
         raise ArithmeticError("element is not a unit")
 
+    @coerce_binop
     def xgcd(self, other):
-        """
+        r"""
         Extended `gcd` for univariate Laurent polynomial rings over a field.
+
+        OUTPUT:
+
+        A triple ``(g, p, q)`` such that ``g`` is the :meth:`gcd` of
+        ``self`` (`= a`) and ``other`` (`= b`), and ``p`` and ``q`` are
+        cofactors satisfying the Bezout identity
+
+        .. MATH::
+
+            g = p \cdot a + q \cdot b.
 
         EXAMPLES::
 
             sage: S.<t> = LaurentPolynomialRing(QQ)
-            sage: (t^-2 + 1).xgcd(t^-3 + 1)
-            (1, 1/2*t^2 - 1/2*t^3 - 1/2*t^4, 1/2*t^3 + 1/2*t^4)
+            sage: a = t^-2 + 1
+            sage: b = t^-3 + 1
+            sage: g, p, q = a.xgcd(b); (g, p, q)
+            (t^-3, 1/2*t^-1 - 1/2 - 1/2*t, 1/2 + 1/2*t)
+            sage: g == p * a + q * b
+            True
+            sage: g == a.gcd(b)
+            True
+            sage: t.xgcd(t)
+            (t, 0, 1)
+            sage: t.xgcd(5)
+            (1, 0, 1/5)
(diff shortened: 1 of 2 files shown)


enriqueartal:
Before looking at how is coded, it is right that `p.gcd(q)` should be equal to `p.xgcd(q)[0]`. What is less clear to me is the choice of the `gcd` up to unit. Whenever there is a canonical choice this should be done (for field coefficients monic with non-zero independent term). For example, now `(-1).gcd(-1)` gives `1`. Should `x.gcd(x)` be `1` for Laurent polynomials?

src/sage/rings/polynomial/laurent_polynomial.pyx
```diff
@@ -1318,22 +1318,59 @@ cdef class LaurentPolynomial_univariate(LaurentPolynomial):
...
-        f, df = self.monomial_reduction()
-        g, dg = other.monomial_reduction()
-        h, p, q = f.xgcd(g)
-        return R(h), p / df, q / dg
+        cdef LaurentPolynomial_univariate elt = other
+        cdef LaurentPolynomial_univariate ret_gcd, ret_p, ret_q
+        cdef long n = min(self.__n, elt.__n)
+
+        h, p, q = self.__u.xgcd(elt.__u)
+
```
enriqueartal, line 1357: I think mathematically it would be better to take out the `x` factors of `h` and change accordingly `p` and `q` to have Bézout identity. It would imply to change `gcd`.

enriqueartal:
I am not sure if my proposal is better. In sage for the integers, the `gcd` of a negative number and itself is positive, and I guess it was done since there is a canonical choice of an integer up to a unit. As you notice, for a non-zero Laurent polynomial (with coefficients in a field or in general in a ring where the canonical choice  can be done), there is a canonical choice of valuation 0. More than computational issues, I think of that the canonical choice is better and it has been done in other case.

tscrim:
It's basically just you like it better than anything mathematical (say, something algebro-geometric related). That's not unreasonable.

However, I think we should keep it the way it is because it will match on polynomials. This is parallel to how the gcd for `QQ` is done to match `ZZ` (and other fraction fields). I would say it is less surprising for the more casual user.

I am not sure which would be faster or less code. I imagine both versions would be basically the same...

enriqueartal:
I basically agree with you. Maybe not right now, but I wonder that if we may have both with a `keyword` as `normalize` on both `gcd` and `xgcd` for Laurent polynomials.

src/sage/rings/polynomial/laurent_polynomial.pyx
```diff
@@ -1318,22 +1318,59 @@ cdef class LaurentPolynomial_univariate(LaurentPolynomial):
             return ~self
         raise ArithmeticError("element is not a unit")
 
+    @coerce_binop
     def xgcd(self, other):
-        """
+        r"""
         Extended `gcd` for univariate Laurent polynomial rings over a field.
```
enriqueartal: Maybe it was my fault to use \`gcd\` instead of gcd, which looks better

tscrim: Shall I change it then?

enriqueartal: It is not important, but either gcd or \`\gcd\` will look better. Otherwise, LGTM

tscrim: Thanks. I opted to link to the method.

enriqueartal (approved):

tscrim:
Thank you.

Issue #37643, opened 2024-03-21
`SetPartitions()` should not error out in degenerate cases


maxale:
### Steps To Reproduce

In contrast to `Partitions()` and `OrderedSetPartitions`, `SetPartitions()` imposes unnecessary restrictions on arguments and errors out when they are not satisfied:
```
sage: list( Partitions(3,length=4) )
[]
sage: list( OrderedSetPartitions(range(3),4) )
[]
sage: list( SetPartitions(range(3),4) )
...
ValueError: part must be <= len(set)
```

### Expected Behavior

In the above example, `SetPartitions(range(3),4)` should simply return an empty object (as there are no suitable set partitions). It is a degenerate case, but not an error.

### Actual Behavior

`SetPartitions(range(3),4)` errors out.

### Environment

```markdown
- **OS**: Ubuntu 22.04.4 LTS
- **Sage Version**: 10.3
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

DitoluT:
This problem can actually be solved by returning an empty setPartition, if length is greater it means there exists no partitions

Nolord:
On could remove the `if` condition that raises [here](https://github.com/sagemath/sage/blob/develop/src/sage/combinat/set_partition.py#L2067) and place it [here](https://github.com/sagemath/sage/blob/develop/src/sage/combinat/set_partition_iterator.pyx#L132) (at the level of the iterator in the Cython code) as such
```Python
if n >= k: #here
   for P in _set_partition_block_gen(n, k, a): #additional indent
      yield from_word(<list> P, base) #additional indent
```

S17A05:
Fixed in 10.4beta4 via #37781.

PR #37781, opened 2024-04-10
`SetPartition` return empty iterator instead of erroring out in degenerate cases

Nolord:
Replaced the error raised by `SetPartition` in a degenerate case (see below) by an empty iterator, to match the behaviour of similar functions such as `Partitions` and `OrderedSetPartitions`. Old behaviour:
```Python
sage: list( Partitions(3,length=4) )
[]
sage: list( OrderedSetPartitions(range(3),4) )
[]
sage: list( SetPartitions(range(3),4) )
...
ValueError: part must be <= len(set)
```
New behaviour
```Python
sage: list( SetPartitions(range(3),4) )
[]
```
This solves issue #37643.

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/set_partition_iterator.pyx
+++ b/src/sage/combinat/set_partition_iterator.pyx
@@ -129,5 +129,6 @@ def set_partition_iterator_blocks(base_set, Py_ssize_t k):
     cdef Py_ssize_t n = len(base)
     cdef list a = list(range(n))
     # TODO: implement _set_partition_block_gen as an iterative algorithm
-    for P in _set_partition_block_gen(n, k, a):
-        yield from_word(<list> P, base)
+    if n >= k:
+        for P in _set_partition_block_gen(n, k, a):
+            yield from_word(<list> P, base)
--- a/src/sage/combinat/set_partition_iterator.pyx
+++ b/src/sage/combinat/set_partition_iterator.pyx
@@ -129,6 +129,7 @@ def set_partition_iterator_blocks(base_set, Py_ssize_t k):
     cdef Py_ssize_t n = len(base)
     cdef list a = list(range(n))
     # TODO: implement _set_partition_block_gen as an iterative algorithm
-    if n >= k:
-        for P in _set_partition_block_gen(n, k, a):
-            yield from_word(<list> P, base)
+    if n < k:
+        return
+    for P in _set_partition_block_gen(n, k, a):
+        yield from_word(<list> P, base)
--- a/src/sage/combinat/set_partition.py
+++ b/src/sage/combinat/set_partition.py
@@ -2061,18 +2061,17 @@ class SetPartitions(UniqueRepresentation, Parent):
                 pass
             s = frozenset(s)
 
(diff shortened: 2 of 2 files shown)


Nolord:
Also, I'm not sure if a similar fix would be needed for the case when the second argument is a list. The element in the list should add up to the length of the set:
```Python
sage: SetPartitions(range(4), [2, 1, 1])
Set partitions of {0, 1, 2, 3} with sizes in [2, 1, 1]
sage: SetPartitions(range(4), [2, 1])
...
ValueError: part must be an integer partition of 4
```
We have the equivalence
```
sage: sorted(list(SetPartitions(range(4), 3))) == sorted(list(SetPartitions(range(4), [2, 1, 1])))
True
```
Maybe the error above should also be replaced by an empty iterator.

grhkm21:
> Also, I'm not sure if a similar fix would be needed for the case when the second argument is a list.

yes

src/sage/combinat/set_partition_iterator.pyx
```diff
@@ -129,5 +129,6 @@ def set_partition_iterator_blocks(base_set, Py_ssize_t k):
     cdef Py_ssize_t n = len(base)
     cdef list a = list(range(n))
     # TODO: implement _set_partition_block_gen as an iterative algorithm
-    for P in _set_partition_block_gen(n, k, a):
-        yield from_word(<list> P, base)
+    if n >= k:
+        for P in _set_partition_block_gen(n, k, a):
+            yield from_word(<list> P, base)
```
tscrim:
```suggestion
    if n < k:
        return
    for P in _set_partition_block_gen(n, k, a):
        yield from_word(<list> P, base)
```
Basically bikeshedding, but I find this easier to understand (and easier to work with if we extend it).

tscrim (changes requested):
Let me also mention as a counterpoint, having the errors here are useful for catching logic bugs. So `OrderedSetPartitions` would error out for such clearly bad input.

Personally, I agree with your change here. However, it is something to consider.

Also, in case you haven't already, please check to make sure every place within Sage this is being called, the code isn't trying to catch the `ValueError` as program flow. (I don't think this is done anywhere, but we should check as it might not be covered by doctests.)

Nolord:
> Let me also mention as a counterpoint, having the errors here are useful for catching logic bugs. So `OrderedSetPartitions` would error out for such clearly bad input.

I agree that getting an empty iterator in this case would most likely be unintended.

> Also, in case you haven't already, please check to make sure every place within Sage this is being called, the code isn't trying to catch the `ValueError` as program flow. (I don't think this is done anywhere, but we should check as it might not be covered by doctests.)

The only other file where the keyword `set_partition_iterator_blocks` is used is in `sage/src/sage/combinat/combinat_cython.pyx`. But it is only imported via `LazyImport`.

tscrim (approved):
That's not quite what I asked, but I checked and didn't see any cases of what I was looking for either. Anyways, this is good for me. Thank you.

PR #37784, opened 2024-04-11
Speed up symmetric group simple_module_rank()

tscrim:
Since we are going to work with a matrix $f^{\lambda} \times f^{\lambda}$ (recall $f^{\lambda}$ is the number of standard Young tableaux of shape $\lambda$), then we should just store the list of standard tableaux rather than iterating over them repeatedly. This results in a major speedup:
```
sage: from sage.combinat.specht_module import simple_module_rank
sage: %time simple_module_rank([5,3,2], GF(2))
CPU times: user 1.34 s, sys: 35.7 ms, total: 1.38 s
Wall time: 1.38 s
200
```
versus before
```
sage: %time simple_module_rank([5,3,2], GF(2))
CPU times: user 4.97 s, sys: 65 ms, total: 5.03 s
Wall time: 5.06 s
200
```

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/specht_module.py
+++ b/src/sage/combinat/specht_module.py
@@ -1351,12 +1351,12 @@ def tabloid_gram_matrix(la, base_ring):
         [4 1 1 2 4]
     """
     from sage.combinat.tableau import StandardTableaux
-    ST = StandardTableaux(la)
+    ST = list(StandardTableaux(la))
 
     def bilinear_form(p1, p2):
         if len(p2) < len(p1):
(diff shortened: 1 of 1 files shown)


tscrim:
Some very simple, but useful optimizations. Should be an easy review. @darijgr @fchapoton @AndrewMathas, if you have time.

darijgr (approved):
Yep, good change. LGTM

tscrim:
Thank you. I will let the tests run then change to a positive review, but I doubt anything will fail.

PR #37391, opened 2024-02-18
Make installation of "wheel" packages less noisy

mkoeppe:
Some care for `build/bin/sage-spkg`, reducing unhelpful verbosity. In particular, for "wheel" packages (https://deploy-livedoc--sagemath.netlify.app/html/en/developer/packaging#package-source-types), there is no build step, so there is no point in talking about what C compiler is in use.
 
Also saving a few lines of output (and making it clearer who does what) by prefixing the output from the various `spkg-...` scripts.

Example:
```
$ make packaging-no-deps
[packaging-23.2] Using cached file /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/upstream/packaging-23.2-py3-none-any.whl
[packaging-23.2] Setting up build directory /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/tmp/sage/build/packaging-23.2
[packaging-23.2] [spkg-piprm] Found existing installation: packaging 23.2
[packaging-23.2] [spkg-piprm] Uninstalling packaging-23.2:
[packaging-23.2] [spkg-piprm]   Successfully uninstalled packaging-23.2
[packaging-23.2] Removing stamp file /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/installed/packaging-23.2
[packaging-23.2] [spkg-install] Staged wheel file, staged /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/scripts/packaging/spkg-requirements.txt
[packaging-23.2] Moving package files from temporary location /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/tmp/sage/build/packaging-23.2/inst to /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11
[packaging-23.2] [spkg-pipinst] Using pip 23.3.1 from /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/lib/python3.11/site-packages/pip (python 3.11)
[packaging-23.2] [spkg-pipinst] Looking in links: /Users/mkoeppe/s/sage/sage-rebasing/worktree-pristine/local/var/lib/sage/venv-python3.11/var/lib/sage/wheels
[...]

PR #37785, opened 2024-04-11
`sage-logger`: Suppress "No such file or directory" messages

mkoeppe:
A minor follow-up after #37391:

Seen for example in 
https://github.com/sagemath/sage/actions/runs/8626379356/job/23644513913?pr=37570#step:11:3227:
```
#24 4091.8 /sage/build/bin/sage-logger: line 66: /usr/bin/time: No such file or directory
#24 4091.8 [sagemath_doc_html-none] installing. Log file: /sage/logs/pkgs/sagemath_doc_html-none.log
#24 5807.5 cat: /sage/logs/pkgs/sagemath_doc_html-none.time: No such file or directory
```
We suppress these (harmless) messages which show up when `/usr/bin/time` is not available.

### :memo: Checklist

### DIFF
--- a/build/bin/sage-logger
+++ b/build/bin/sage-logger
@@ -63,7 +63,7 @@ fi
 
 timefile="$logdir/$logname.time"
 rm -f "$timefile"
-if /usr/bin/time -h -o /dev/null true; then
+if /usr/bin/time -h -o /dev/null true 2>/dev/null; then
     TIME="/usr/bin/time -h -o $timefile"
 else
     TIME=""
@@ -71,7 +71,7 @@ fi
 
 report_time ()
 {
-    time=$(echo $(cat $timefile))
+    time=$(echo $(cat $timefile 2>/dev/null))
     case "$time" in
         *m*real*|*h*real*|*[1-9][0-9].*real*|*[1-9][0-9],*real*)
             # at least 10 seconds wall time


DavidAyotte (approved):
It looks good to me!

mkoeppe:
Thanks!

build/bin/sage-logger
```diff
@@ -63,15 +63,15 @@ fi
 
 timefile="$logdir/$logname.time"
 rm -f "$timefile"
-if /usr/bin/time -h -o /dev/null true; then
+if /usr/bin/time -h -o /dev/null true 2>/dev/null; then
```
tornaria, line 66: I'm sure ubuntu has `/usr/bin/time` available. The problem is that you are using nonstandard arguments to `time(1)`:
```
$ /usr/bin/time -h -o /dev/null
/usr/bin/time: invalid option -- 'h'
Try '/usr/bin/time --help' for more information.
$ echo $?
125
```
Also, if you want to parse the time you should pass `-p` to ensure the standard format is used (https://www.man7.org/linux/man-pages/man1/time.1p.html), and then you can just use the bash builtin.

For what you are trying to acomplish, you could also use bash variables `$EPOCHSECONDS` (or `$EPOCHREALTIME` for more precision).

mkoeppe: The reviewer specifically asked for the human-readable format to be used.
The whole thing is just a convenience.

tornaria: > The reviewer specifically asked for the human-readable format to be used. The whole thing is just a convenience.

I don't know what is "the human-readable format" but `-h` is not supported by gnu `time(1)` nor by the bash builtin.

Gnu time does support `-o` but not `-h`.

If you want to bypass the bash builtin, the better approach is to use `\time` which will still search in `PATH`.

> For what you are trying to acomplish, you could also use bash variables $EPOCHSECONDS (or $EPOCHREALTIME for more precision).

There's also `$SECONDS`.

tornaria: > The reviewer specifically asked for the human-readable format to be used. The whole thing is just a convenience.

I see, but then the bash builtin time is just what the reviewer asked for.

You could
 - do the redirect `2>&1` inside the `sh -c "..."`
 - capture with `2>` the output of time

(since the bash builtin doesn't support -o)

mkoeppe: If you want to prepare a follow-up PR, I can test it for its portability. I don't have time to work on it myself at the moment.

tornaria: > If you want to prepare a follow-up PR, I can test it for its portability. I don't have time to work on it myself at the moment.

Please don't tell me what I should do with my time.

My comment before was just remarking that the current PR does not fix the issue but only hides it. Reverting cb4e7184b for the time being seems to be easier workaround, until you or someone else gets the time to improve the output in a portable way.

mkoeppe: > Please don't tell me what I should do with my time.

@tornaria And I am not, so don't tell me what I shouldn't tell you.

tornaria:
Please revert https://github.com/sagemath/sage/commit/cb4e7184b3af09a1ed3c2353aba77d63f206276d so this works on all supported platforms.

mkoeppe:
No, that's outside of the scope of this PR.

mkoeppe:
Per PR description, this PR resolves the messages displayed `/usr/bin/time: No such file or directory` when there is no `/usr/bin/time` binary.

tornaria:
> No, that's outside of the scope of this PR.

It's in scope for #37391 which was incorrect in this regard. The follow-up should be fixing, not hiding it.

The error you showed is
```
cat: /sage/logs/pkgs/sagemath_doc_html-none.time: No such file or directory
```
this has nothing to do with the absence of `/usr/bin/time`, rather with the fact that `time -h` fails because of the non-standard option.
 
Reverting cb4e718 seems easy enough; this ends up with your original PR which is the one I assume you tested most. Trying to satisfy a very reasonable request by @kwankyu is great, but not at the expense of breaking in a lot of supported systems.

mkoeppe:
Please, Gonzalo, the PR author sets scope of the PR.

mkoeppe:
If you look at the actual ticket description, line 1:

> #24 4091.8 /sage/build/bin/sage-logger: line 66: /usr/bin/time: No such file or directory
> #24 4091.8 [sagemath_doc_html-none] installing. Log file: /sage/logs/pkgs/sagemath_doc_html-none.log
> #24 5807.5 cat: /sage/logs/pkgs/sagemath_doc_html-none.time: No such file or directory

So yes, this PR is about absence of `/usr/bin/time` as indicated.

tornaria:
> If you look at the actual ticket description, line 1:
> 
> > #24 4091.8 /sage/build/bin/sage-logger: line 66: /usr/bin/time: No such file or directory
> > #24 4091.8 [sagemath_doc_html-none] installing. Log file: /sage/logs/pkgs/sagemath_doc_html-none.log
> > #24 5807.5 cat: /sage/logs/pkgs/sagemath_doc_html-none.time: No such file or directory
> 
> So yes, this PR is about absence of `/usr/bin/time` as indicated.

But #37391 is also broken on boxes where `/usr/bin/time` is available but 
```
$ /usr/bin/time -h -o /dev/null true
/usr/bin/time: invalid option -- 'h'
Try '/usr/bin/time --help' for more information.
$ echo $?
125
```
This includes most linux boxes.

I grant you that changing this to use the builtin `time` or `$SECONDS` is more work, so I understand not doing that. But changing this to not use `-h` seems very easy.

I'll show myself the exit. I was only trying to help since I know (and appreciate) you are always concerned about portability.

PS: note that `time` bash builtin output cannot be captured:
```
$ time 2> /dev/null

real	0m0.000s
user	0m0.000s
sys	0m0.000s
$
```
so the obvious using `time` or `time -p` and capture with `2>` doesn't seem to work. What you *can* do is run the time in a subshell like so:
```
$ ( time true ) 2> /dev/null
$
```
In fact, you could do something like the following:
```
$ cmd="echo some stdout; sleep 1; echo and some stderr; exit 42"
$ timeit=$( (>>logfile.txt 2>&1 time sh -c "$cmd") 2>&1 ) ; status=$?
$ echo $status
42
$ echo $timeit
real 0m1.004s user 0m0.003s sys 0m0.002s
$ cat logfile.txt 
some stdout
and some stderr
$
```

PS2: when you want to skip a bash builtin, the common idiom is `\time` which will still search for `time` in `PATH` so you avoid hardcoding it and allow the user to replace the command if they so wish.

mkoeppe:
> I'll show myself the exit. I was only trying to help

That's appreciated, and I will benefit from the details and suggestions that you made when I work on the follow-up PR. Thank you.

tornaria:
> > I'll show myself the exit. I was only trying to help
> 
> That's appreciated, and I will benefit from the details and suggestions that you made when I work on the follow-up PR. Thank you.

I think my last suggestion might be useful. It seems to me that the format for the bash builtin `time` is similar to the output of `time -h` that you parse, so possibly you don't need to change `report_time`.

Since all of our scripts already depend on bash (as the shebang shows) depending on bashisms of the `time` builtin is "kind of" ok (at least it should be the same everywhere).

PR #37787, opened 2024-04-11
modernize coercion for Hecke algebras

fchapoton:
Small changes that switch Hecke algebras to the modern coercion framework.

<s>Also avoiding the auld class `CommutativeRing` and use the categories instead.</s>

### :memo: Checklist

### DIFF
--- a/src/sage/modular/hecke/algebra.py
+++ b/src/sage/modular/hecke/algebra.py
@@ -35,13 +35,13 @@ from sage.arith.functions import lcm
 from sage.arith.misc import gcd
 from sage.misc.latex import latex
 from sage.matrix.matrix_space import MatrixSpace
+from sage.rings.ring import CommutativeRing
 from sage.rings.integer_ring import ZZ
 from sage.rings.rational_field import QQ
 from sage.structure.element import Element
 from sage.structure.unique_representation import CachedRepresentation
 from sage.misc.cachefunc import cached_method
 from sage.structure.richcmp import richcmp_method, richcmp
-from sage.structure.parent import Parent
 
 
 def is_HeckeAlgebra(x) -> bool:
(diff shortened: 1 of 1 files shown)


tscrim:
Are we sure we are not loosing the methods:
- `ideal_monoid`
- `extension`
- `frobenius_endomorphism`
- `derivation_module`
- `derivation`
(as they are all provided by `CommutativeRing` generically)? Unfortunately I think we have to address this now, and I don't think they are provided. We probably just need to copy them to the `CommutativeRings()` category.
 
(`localization` and `fraction_field` are for integral domains, so not having them is an improvement (unless Hecke algebras are always an integral domain, but I don't think so?). Actually, are these provided by the category? I don't see them in there. `krull_dimension` is generically a `NotImplementedError`.)

fchapoton:
I would prefer to keep this one small, as there is already the big #37719 blocking the view. I am trying to move by little steps only. Here we are getting rid of the last class using `CommutativeRing` outside of the `ring` folder.
I understand your concerns, but there is no doctest. Of course, people could use that in the wilderness.

Would it be ok to perform the move of these methods to the category later ? Probably copying them is easier, but there is no doctest ready for testing these copied methods.

tscrim:
I don't think a subclass must include doctests/examples showing the methods defined in a base class work; so I don't find that to be a convincing argument. For `ideal_monoid` and `derivation_module`, we can just use the classes at hand here. We could ask one of the number theory people if they have any thoughts (for this specific case) and/or suggestions for such doctests.

I am also happy to do the changes on a separate PR that would become a dependency of this one.

fchapoton:
ok, I will do another PR, but probably not soon

tscrim (approved):
LGTM. Thank you.

PR #37494, opened 2024-02-28
`build/pkgs`: Update scip to 9, scip_sdp to 4.4, soplex to 7, papilo to 2.2, pyscipopt to 5, onetbb to 2021.11 

mkoeppe:
https://www.scipopt.org/doc-9.0.0/html/RN90.php

- [x] https://github.com/scipopt/SCIP-SDP/issues/12
- [ ] https://github.com/scipopt/soplex/issues/35
- [x] needs pyscipopt update

### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #37237 
- Depends on #37351 
- Depends on #36970
- Depends on #37392

Issue #25, opened 2006-09-12
make sage_eval only take strings as input.


williamstein:
<div id="comment:0"></div>

Component: **basic arithmetic**

_Issue created by migration from https://trac.sagemath.org/ticket/25_

PR #37801, opened 2024-04-14
Drop custom GLPK patch and error handling

### LINKED ISSUE #29829: Remove GLPK error recovery patch
<div id="comment:0"></div>

In #29493, it was proposed to remove the remove the `sage.libs.glpk.error` module and `error_recovery.patch` for GLPK, which was rejected by upstream.

This ticket is split out from the larger changes in #29493.

Depends on #29493

CC:  @orlitzky @kliem @jdemeyer @mkoeppe @jpflori @embray @sagetrac-gouezel @kiwifb

Component: **packages: standard**

Author: **Michael Orlitzky**

Branch/Commit: **[u/mjo/ticket/29829](https://github.com/sagemath/sagetrac-mirror/tree/u/mjo/ticket/29829) @ [`c49ef62`](https://github.com/sagemath/sagetrac-mirror/commit/c49ef6277f3faffd83e70c50bfd1fb66d6160489)**

_Issue created by migration from https://trac.sagemath.org/ticket/29829_

orlitzky:
The sage distribution patches GLPK to be able to recover from errors. Upstream rejected this patch, said what it does is unsupported, and no one else has ever adopted it. Sage works fine without it.

If someone really wants to add error recovery to GLPK, it has to be done in a way that upstream does not object to. Otherwise, it cannot be done reliably -- no linux distros or conda or homebrew are going to ship our patch.

These changes date back to https://github.com/sagemath/sage/issues/29493, and  this PR will close https://github.com/sagemath/sage/issues/29829.


mkoeppe:
> 2024-04-15T12:54:14.3132521Z #25 3690.5 cat: logs/install.time: No such file or directory

This one is cosmetic and does not cause failures. Fixed in:
- #37785

mkoeppe:
> is (AFAICT) eventually causing
> 
> ```
> 2024-04-15T14:01:25.4315886Z #25 7721.7 make[1]: *** [Makefile:240: test] Error 1
> 2024-04-15T14:01:25.4316823Z #25 7721.7 make[1]: Leaving directory '/sage'
> 2024-04-15T14:01:25.4317485Z #25 7721.7 make: *** [Makefile:255: ptest] Error 2
> 2024-04-15T14:01:25.4318987Z #25 ERROR: process "/bin/sh -c make SAGE_SPKG=\"sage-spkg -y -o\" ${USE_MAKEFLAGS} ${TARGETS}" did not co\
> mplete successfully: exit code: 2
> ```

Yes, the doc-html build failure is what causes this.

mkoeppe:
OK, let's get rid of it. But why remove `src/sage/libs/glpk/env.pxd`?

orlitzky:
> OK, let's get rid of it. But why remove `src/sage/libs/glpk/env.pxd`?

IIRC it's unused and I only noticed it because I was removing `error.pyx` at the time.

mkoeppe:
But I think we often carry .pxd files that just provide the full API whether it's used or not.

orlitzky:
I would guess that it'd be better to regenerate it however-many years from now based on the latest GLPK, if/when someone wants it. But I don't really care if you think we should keep it. It was removed for housekeeping and not to solve any problem.

mkoeppe:
Yes, let's please keep it. (By the way, I don't expect any API developments in GLPK to happen.)

mkoeppe:
By the way, I hope to downgrade GLPK to "optional" soon and replace it by SCIP (see #37494)

mkoeppe (approved):
LGTM, thanks.

orlitzky:
Thanks.

> By the way, I hope to downgrade GLPK to "optional" soon and replace it by SCIP (see #37494)

GLPK isn't well-maintained these days, and is not exactly at the forefront of mathematical programming, so I think that would be an improvement.

vbraun:
merge conflict

PR #37804, opened 2024-04-15
Fix sagemath-standard sdist

mkoeppe:
Fixes https://github.com/sagemath/sage/pull/37286#issuecomment-2054464011 @kiwifb 

### :memo: Checklist

### DIFF
--- a/src/MANIFEST.in
+++ b/src/MANIFEST.in
@@ -2,6 +2,7 @@ include VERSION.txt
 
 recursive-include sage *.pxi *.pxd *.h *.hpp
 
+graft sage
 prune sage/ext/interpreters   # In particular, __init__.py must not be present in the distribution; or sage_setup.autogen.interpreters.rebuild will not generate the code
 prune sage_setup
 prune sage_docbuild


kiwifb:
OK I got .py files now. Still trying to bisect the origin, once I fixed my container issues to run checkout commits.

tornaria:
Can we please figure out what causes the issue and why the python files are not being included, before adding a line that should not be necessary afaik? Also, doesn't `graft sage` means to add *all* files inside `sage` that are not later excluded?

Here: https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html is the list of files that should be included by default:
 - [pure Python module](https://packaging.python.org/en/latest/glossary/#term-Pure-Module) files implied by the py-modules and packages configuration parameters in pyproject.toml and/or equivalent in setup.cfg/setup.py;
From that wording, this may mean something was missed in the switch from setup.cfg to pyproject.toml.

mkoeppe:
The `graft` brings it in line with what I'm doing in the other distribution packages (sagemath-categories etc).

kiwifb:
I am not keen on going through individual commit myself to find out - I just checked at PR commits level which may or may not be a correct way of doing it, but if someone has a good process to go through those to check the presence of .py files they are welcome to it.

kcrisman:
> Can we please figure out what causes the issue and why the python files are not being included, before adding a line that should not be necessary afaik? Also, doesn't `graft sage` means to add _all_ files inside `sage` that are not later excluded?

It does seem like fixing whatever went wrong might be best in this case, rather than adding something new that may also have unintended side effects.  I'd ask @tornaria @mkoeppe @kiwifb to explain it to me, but actually I think that is not a good use of their time.  Unfortunately possibly necessitated by the new voting mechanism, as I suppose this one is fated to become disputed as well.

mkoeppe:
> Can we please figure out what causes the issue and why the python files are not being included

I already did that, which enabled me to fix it. There's no mystery, and I'll be happy to explain it, or any other aspect of Python packaging for that matter.

mkoeppe:
The explanation: `pkgs/sagemath-standard/setup.py` was changed in commit https://github.com/sagemath/sage/commit/865e6a0da63e6e656a1fa589a1b255917eb76dcb (part of #36964). This change should have been accompanied by the change now added here.
That's all.

mkoeppe:
> am I wrong in thinking that `graft sage` means to add all files inside sage that are not later excluded?

Yes, that's what it means.

tornaria:
> > Commit [865e6a0](https://github.com/sagemath/sage/commit/865e6a0da63e6e656a1fa589a1b255917eb76dcb) seems very strange.
> 
> What precisely is your concern about it?

For a start, it's unreviewed.

mkoeppe:
> things are getting much better in python packaging.

I know, Gonzalo, because doing the Python distribution packaging for Sage and bringing it to modern practice (PEP 517/518) is what I have been doing for the past 4 years.

kiwifb:
So, as I commented in another issue just now, the graft causes `sage/graphs/graph_decomposition/tdlib.pyx`  to be shipped as part of the sdist when it is supposed to be in sagemath-tdlib only. May be some other - it seems not to include any files from sagemath-bliss, sagemath-meataxe and sagemath-sirocco which are packages I am supporting in sage-on-gentoo.

mkoeppe:
@kiwifb Thanks for catching this. Actually only tdlib is affected, and it's because of a typo in the `exclude` line responsible for it, which also leads to this warning in `make sagelib-sdist`.
```
warning: no previously-included files found matching 'sage/graphs/graph_decomposition/tdlib.p*'
```
I'll fix this in:
- #37901

... after cherry-picking the commit from here.

PR #37805, opened 2024-04-15
Implement algorithm for enumerating ideals of a poset in O(log n) amortized time

### LINKED ISSUE #37749: generating the ideals of a poset in logarithmic time complexity
### Problem Description

Currently, SageMath lacks efficient algorithms for enumerating the ideals of a partially ordered set (poset) in logarithmic time complexity. An enumeration of the ideals of a poset can be used to solve various scheduling and reliability problems. Current algorithms for generating ideals require an amortized time of `O(n)` per ideal in the worst case, where `n` is the number of elements in the poset. Through this issue, we can generate ideals in an amortized time of `O(log n)` per ideal. 

### Proposed Solution

The proposed solution involves implementing an efficient algorithm for enumerating the ideals of a partially ordered set (poset) in logarithmic time complexity within SageMath. One approach to achieve this is by adapting the algorithm described in the research paper "Enumerating the Ideals of a Poset" by Matthew B. Squire. This algorithm presents a divide-and-conquer strategy that efficiently enumerates the ideals of a post. 

Specifically, the implementation would involve creating a recursive procedure to generate ideals based on subsets of elements in the poset. By strategically partitioning the poset and exploiting its structure, the algorithm can achieve logarithmic time complexity for ideal enumeration.

### Alternatives Considered

1. **Naive Enumeration:** One alternative is to continue using the current methods for enumerating ideals, which typically have polynomial time complexity. However, these methods may become impractical for large posets or when performing intensive computations.
2. **External Libraries:** Another option is to rely on external libraries or tools that provide efficient algorithms for enumerating ideals. However, this approach may introduce dependencies and compatibility issues, and it may not integrate seamlessly with SageMath's existing functionality.
3. **Custom Implementations:** Users could develop custom implementations of ideal enumeration algorithms outside of SageMath. While feasible, this approach may require significant effort and expertise, and it may not benefit from the full integration and support offered by SageMath.

### Additional Information

_No response_

saatvikraoIITGN:
### Description
This PR implements an algorithm for enumerating ideals of a poset in the amortized time complexity of O(log n). The algorithm is based on a new approach that efficiently generates upsets using a recursive procedure. It reorders the elements of the poset and recursively generates upsets, significantly reducing the time complexity compared to previous methods.

This change is required to improve the efficiency of enumerating ideals of a poset, which is a common operation in various applications such as computational geometry, combinatorics, and optimization problems.

Fixes #37749 

### :memo: Checklist

### :hourglass: Dependencies

#37345 : the acyclic orientations function that is implemented here will have to be modified for better efficiency


src/sage/categories/posets.py
```diff
@@ -305,6 +305,56 @@ def order_filter(self, elements):
...
+            """
+            if not Y:
+                yield U
+            else:
+                Xm = Y[len(Y) // 2]
+                Y1 = []
+                Y2 = []
+                for Xi in Y:
+                    if Poset[(Xi, Xm)] == 1:
+                        U[Y.index(Xi)] = 0
```
dcoudert: A call to `Y.index(Xi)` takes time `O(|Y|)`. You can avoid it using enumerate:
```py
for i, Xi in enumerate(Y):
    if Poset[(Xi, Xm)] == 1:
        U[i] = 0
    else:
        Y1.append(Xi)
```

src/sage/categories/posets.py
```diff
@@ -305,6 +305,56 @@ def order_filter(self, elements):
...
+                    else:
+                        Y1.append(Xi)
+                yield from Posets.generate_upsets(U.copy(), Y1, Poset)
+                for Xj in Y:
+                    if Poset[(Xm, Xj)] == 1:
+                        U[Y.index(Xj)] = 1
```
dcoudert: same here

src/sage/categories/posets.py
```diff
@@ -305,6 +305,56 @@ def order_filter(self, elements):
...
+
+            INPUT:
+
+            - ``U`` -- a list of the current upset being constructed.
+            - ``Y`` -- a list of remaining elements to be considered for inclusion.
+            - ``Poset`` -- a dictionary representing the partial order.
+
+            OUTPUT:
+
+            - A subset of Y that satisfies the upset condition.
```
dcoudert: Y -> `Y`

tscrim:
@saatvikraoIITGN Please use proper Python OOP. Mainly start methods with `self` and use `self.generate_upsets`. Although you are probably better utilizing the underlying digraph (i.e., Hasse diagram).

I am also not sure of your claim about the speed. How Sage constructs a single ideal is using a DFS on the Hasse diagram, which should be $O(\log n)$ amortized (with $n = |P|$ the size of the poset).  The abstract of the reference (which I can't find, please provide a link or full citation info) makes a contradictory claim: amortized means you don't consider the worst case. I doubt you will be able to do better than that unless you are really taking advantage of the DAG property. On the other hand, if you want to enumerate *all* ideals, then I don't believe it can be done better than $O(n)$ since you have at least one ideal for each element in the poset. Please clarify this.

@kevindilks You might be interested in this PR too.

Issue #37782, opened 2024-04-10
Need for FPT Algorithm for Vertex Cover of a Graph


MumukshTayal:
### Problem Description

The goal is to implement a function that determines whether a given graph has a vertex cover of size less than or equal to a user-provided parameter k. This approach utilizes a [Fixed Parameter Tractable (FPT)](https://en.wikipedia.org/wiki/Parameterized_complexity) algorithm, which provides better performance compared to finding the exact minimum vertex cover size, especially for smaller values of k.

### Proposed Solution

The proposed solution is an implementation of an FPT algorithm for the Vertex Cover problem. The algorithm checks if a graph has a vertex cover of size less than or equal to the user-provided parameter k. It does not return the exact size of the minimum vertex cover. The time complexity of this implementation is O(n^2 * 1.71^k), where n represents the size of the graph.
The implementation of the proposed FPT algorithm is available in the following [GitHub Gist](https://gist.github.com/MumukshTayal/279d02dc1188dea55cdad8dfd6335c0e)

### Alternatives Considered

**Minimum Vertex Cover**: The SageMath library already implements an algorithm for finding the minimum vertex cover size. However, this implementation uses Cython, which may outperform the proposed Python implementation in certain cases, particularly when the size of the minimum vertex cover is close to the parameter k.

### Additional Information

- The FPT approach is generally applicable for relatively smaller values of k, as it converts the problem from an exponential time complexity to a polynomial time complexity by handling the exponential part of the algorithm with the parameter k as O*(c^k), where O* represents a polynomial multiplied by c^k, and c is a constant.

- If the user provides a small value of k and the size of the graph increases from n to 2n vertices, the time complexity of the proposed FPT implementation remains O*(c^k), with a slight performance hit due to the polynomial time impact. In contrast, a precise solution providing algorithm may grow from O*(c^n) to O*(c^(2n)), an exponential increase.

- The proposed FPT implementation may be useful for smaller values of k, as it offers better performance compared to the brute-force solution, while sacrificing the exactness of the minimum vertex cover size.

### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

dcoudert:
May be you should join forces with @SandwichGouda to get a better code ? I don't know if it is possible to have 2 developers for the same PR.

dimpase:
Surely, a developer can open a PR on the fork of another developer - then, if accepted, the corresponding commits will have correct authorship.

PR #37806, opened 2024-04-16
Added has_vertex_cover_of_size method to Graph

SandwichGouda:
I implemented a method to ``Graph`` that tells whether a given graph has a vertex cover of size less than or equal to a user-provided parameter ``k``. This provides better performance compared to finding the exact minimum vertex cover size, especially for smaller values of k.
This solves issue #37782.

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/graph.py
+++ b/src/sage/graphs/graph.py
@@ -89,6 +89,8 @@ AUTHORS:
 - Jean-Florent Raymond (2019-04): is_redundant, is_dominating,
    private_neighbors
 
+- William Driot (04-2024) : added has_vertex_cover_of_size.
+
 Graph Format
 ------------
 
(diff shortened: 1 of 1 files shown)


dcoudert:
Welcome to Sagemath.

I have several concern about this PR:
1. I don't understand the algorithm you implement. You should document the code.
2. You use recursion and so the method is limited in the size of graphs it can accept.
3. I'm not sure at all of the efficiency of the proposed algorithm compared to using our methods for computing a vertex cover and checking it's size. Is this method really faster ?  Vertex cover is usually very efficiently solved by exact methods and your method should at least be compared to a ILP solution in which you add a constraint on the size of the solution. 
4. why not adding a parameter k to method `vertex_cover` to check if the graph has a vertex cover of size at most k ? this way you could benefit from the preprocessing methods and the (rather fast) algorithms it proposes.

SandwichGouda:
1. The algorithm is a [Fixed Parameter Tractable (FPT)](https://en.wikipedia.org/wiki/Parameterized_complexity) algorithm.
I will add this information to the docs, together with some references. However, having a deep understanding of the algorithm requires understanding of Parametrized Complexity, which is a high level master's degree subject that I can't squash in a few lines. 
You are totally right, though, that I should add some references to the docs as to where the algirithm comes from (although I won't be able to explain it fully)(and that's not the point IMO : most algorithms in Sage aren't explained in detail in the code itself, they have external links to external references that explain them). 

2. The complexity of this algorithm is $O(n^2 * 1.71^k)$ where $n$ is the size of the graph and $k$ the input parameter. 
In some use cases, the complexity won't beat ``vertex_cover`` (when k is too big)
But the powerful advantage of the FPT (_Fixed Parameter_) approach is that it allows to fix ``k`` to completely outperform
exact methods : if ``k`` is fixed, and/or remains not too big, the algorithm is polynomial in $n$.

For instance, if the size of the graph increases from n to 2n vertices, the time complexity of the proposed FPT implementation remains $O^* (c^k)$, In contrast, a precise solution providing algorithm may grow from $O^* (c^n)$ to $O^* (c^{2n})$, an exponential increase.

Here's an example :

```
sage: t = time.time() ; graphs.CircularLadderGraph(10**4).vertex_cover() ; time.time() - t
...
387.23221015930176 # 5 minutes
sage: t = time.time() ; graphs.CircularLadderGraph(10**4).has_vertex_cover_of_size(10) ; time.time() - t
...
0.18092012405395508 # 180 milliseconds
```

3. After some testing, it appeared to me that the recursion problem is not really one, because the recursion depth limit is beyond the "algorithms-takes-way-too-much-time-to-complete-anyway" time limit (when the values are too big, and that none of the approaches yield a result in reasonable time anyway). 

4. The algorithms have completely different approaches, it would be weird to add $k$ as a parameter of ``vertex_cover`` and distinguish cases (if $k$ not given, exact algorithm, if $k$ given, FPT approach)

dimpase:
What is $k$ ? The size of the cover?

I know an easy algorithm with complexity $n^2 2^k$. 1.7 instead of 2 must be harder, and I suppose it is published somewhere.
Put this reference in the docs for your code.

PR #37809, opened 2024-04-16
gambit: Re-create as a pip package `pygambit` (help wanted!)

mkoeppe:
[Gambit](http://www.gambit-project.org/), the package for computation in game theory, is being actively developed again and has made several releases in 2024. Congratulations to @tturocy for obtaining support from the Alan Turing Institute for the development of this package.

In this PR, we update our use of Gambit to make it compatible with Gambit 16.x via its Python package [pygambit](https://pypi.org/project/pygambit/).

Steps:
- [x] Package and Feature
- [x] Update imports (first pass: rename `gambit` -> `pygambit`)
- [x] Sphinx role for linking to new documentation https://gambitproject.readthedocs.io/en/latest/pygambit.user.html
- [ ] Update API use (Help wanted - send PRs)

Cc: @drvinceknight @kcrisman @fchapoton 

### :memo: Checklist

### :hourglass: Dependencies

- #37810


kcrisman:
Just as a point of order, how does one see which parts of the diff are *just* from this particular PR and not from (say) #37810?

This is a welcome development, though, I second the congrats to the developer.  I don't know if we want to make it a standard package, is this draft currently adding it as optional?

mkoeppe:
> how does one see which parts of the diff are _just_ from this particular PR and not from (say) #37810?

Actually #37810 is not merged here yet. I prepared it after preparing the PR here because it became clear that more work is needed to finish it but I want the removal #37810 done as quickly as possible.

mkoeppe:
> I don't know if we want to make it a standard package, is this draft currently adding it as optional?

Yes, it's adding it as "optional" (see file `build/pkgs/pygambit/type`). I'd think adding it as a standard package would have to go through sage-devel again.

tturocy:
Many thanks @mkoeppe for initiating this.  We've been trying to do work in Gambit on standardising packaging to make it easier for downstream integration (with SageMath high on that list).  Please let us know if we can help because, as noted, we're actually resourced now!

mkoeppe:
The new packaging looks all good, and installing in Sage is already working with the present branch.

On our side just the uses of the old `gambit` Python API need to be updated to the new `pygambit` API; but I don't have first-hand experience with either of the two.

tturocy:
> The new packaging looks all good, and installing in Sage is already working with the present branch.
> 
> On our side just the uses of the old `gambit` Python API need to be updated to the new `pygambit` API; but I don't have first-hand experience with either of the two.

With respect to this, we should disclose that we are currently very much in a "move fast and fix things" mode.  There will be breaking API improvements across minor versions.  For at least this calendar year, we probably won't always be able to provide deprecation warnings (but we will try to do so where we can).  To compensate for this our plan is to maintain at least the two previous minor releases (starting from 16.1) with bug fixes, so users won't be forced to adapt their code to API changes simply to get around some bug.

By about this time next year, we will be hoping to have settled into a more mature lifecycle where there'll be fewer API changes and a deprecation path for the ones that do happen.

aliresool621:
I got to say, will this work like gap or there will be a missing pieces?

edwardchalstrey1:
Hello @mkoeppe - there's now a fair bit of momentum underway on Gambit around development and documentation. Do you know if this is likely to be picked up in the near future?

kcrisman:
> There's now a fair bit of momentum underway on Gambit around development and documentation. Do you know if this is likely to be picked up in the near future?

My guess is that, at the current time, this would require at least some Gambit developers to modify this pull request to the most recent head, possibly with some other modifications relative to packages.  An email to the sage-devel list might clarify which pieces of this can be used as-is (hopefully most!).  I would love to see this return, though I'm not currently in a position to help directly, my apologies.

Davda-James:
@mkoeppe I had previously reached out to the Gambit organization and maintainers regarding SageMath interoperability project, and I received their response outlining the suggested next steps. Following I went through commits that has marked 3 steps, and now I am analyzing api level changes, it would be great help if you can suggest which parts of the existing work can be reused as is, and where new or updated implementation would be needed (as far I explored the sage codebase its `src/sage/game_theory/normal_form_game.py` and `src/sage/game_theory/parser.py`) ..

itvi-1234:
Hi @mkoeppe, @kcrisman, @edwardchalstrey1,

I am a GSoC 2026 applicant working with the Gambit project (mentors: Ed Chalstrey, Ted Turocy) on the "Working on Gambit Interoperability with SageMath" project.

I have studied this PR in detail and have a few questions before I start working on the API update (Step 4):

`Package declaration`: I noticed build/pkgs/pygambit/ is missing package-version.txt and version_requirements.txt. Should I add these targeting pygambit>=16.1,<17, or is there a preferred version constraint?

`Rebase strategy`: Since this branch is ~2 years behind sagemath:develop, would it be better to rebase the existing branch or open a fresh PR based on the current state of develop?

`format_gambit() in parser.py` : Since ExternalLCPSolver and NashProfile no longer exist in pygambit 16.x, I plan to remove the format_gambit() method entirely and inline the equilibrium extraction directly in _solve_LCP(). Does that approach sound right to you?

Thank you for the foundational work already done here — I plan to build on it.

— Sumit Goyal

Steve5829:
I am currently testing PR #37809 with Gambit 16.5 and SageMath 10.x to identify the specific integration bottlenecks.

kcrisman:
Glad there is work on this.  With Gambit's reinvigoration, it's a natural time to glue these worlds back together.

> `Package declaration`: I noticed build/pkgs/pygambit/ is missing package-version.txt and version_requirements.txt. Should I add these targeting pygambit>=16.1,<17, or is there a preferred version constraint?

I think that would really depend on what versions of pygambit are "up to date" and compatible with whatever version of Python Sage is currently using.  The Gambit mentors would be able to speak to that.

> `format_gambit() in parser.py` : Since ExternalLCPSolver and NashProfile no longer exist in pygambit 16.x, I plan to remove the format_gambit() method entirely and inline the equilibrium extraction directly in _solve_LCP(). Does that approach sound right to you?

Well, you still want to call Gambit's LCP-computed Nash solution, correct?  Looks like there's a [host of ways](https://gambitproject.readthedocs.io/en/v16.1.0/pygambit.api.html#computation-of-nash-equilibria) Gambit can do that now, so presumably you'd have to call [`lcp_solve`](https://gambitproject.readthedocs.io/en/v16.1.0/api/pygambit.nash.lcp_solve.html#pygambit.nash.lcp_solve) and see whether it gives output in a format Sage can use.  If yes, then great!  Otherwise you might still need to format things somehow, and the `Parser` is where that could happen.

PR #37835, opened 2024-04-20
Optimize `DisjointSet`

gmou3:
The methods `find` and `union` of the class `DisjointSet_of_hashables` pass through an unnecessary layer of input checks that slow down the code. I avoided these checks, cythonized more of the code, and edited the docstrings.

I can attest to the fact that these changes improve speed, and in a follow-up PR (#37839) I mention a test that confirms this.


mantepse:
Wouldn't it make sense to simply remove the input checks?

gmou3:
Yes, I would prefer to remove them altogether. However, `union` is a reserved word in C, so I was tempted to create these alternatives. Any suggestions welcome.

src/sage/sets/disjoint_set.pyx
```diff
@@ -397,7 +397,7 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
             sage: d.__getstate__()
             [0, 1, 0, 0, 4]
         """
-        l = []
+        cdef list l = []
         cdef int i
         for i in range(self.cardinality()):
             l.append(self._nodes.parent[i])
```
tscrim:
```suggestion
        cdef Py_ssize_t card = self.cardinality()
        cdef list l = [None] * card
        cdef int i
        for i in range(card):
            l[i] = self._nodes.parent[i]
```
It is faster to set the length of the list than to constantly append. (Here it is a micro-optimization.)

src/sage/sets/disjoint_set.pyx
```diff
@@ -779,8 +826,44 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
             ...
             KeyError: 5
         """
-        i = self._el_to_int[e]
-        r = self._d.find(i)
+        cdef int i = self._el_to_int[e]
+        cdef int r = self._d.find(i)
+        return self._int_to_el[r]
```
tscrim: I don't think there is a benefit to typecasting here. It might actually be slower since it has to check the types.

src/sage/sets/disjoint_set.pyx
```diff
@@ -779,8 +826,44 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
...
+            sage: [e._find(i) for i in range(5)]
+            [0, 1, 1, 1, 1]
+            sage: e._find(5)
+            Traceback (most recent call last):
+            ...
+            KeyError: 5
+        """
+        cdef int i = self._el_to_int[e]
+        cdef int r = self._d._find(i)
         return self._int_to_el[r]
```
tscrim: Same comment here about typecasting.

src/sage/sets/disjoint_set.pyx
```diff
@@ -797,23 +880,46 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
...
             {{'a', 'b'}, {'c', 'e'}, {'d'}}
-            sage: e.union('b','e')
-            sage: e
+            sage: e.union('b','e'); e
             {{'a', 'b', 'c', 'e'}, {'d'}}
         """
-        i = self._el_to_int[e]
-        j = self._el_to_int[f]
+        cdef int i = self._el_to_int[e]
+        cdef int j = self._el_to_int[f]
```
tscrim:
```suggestion
        cdef int i = <int> self._el_to_int[e]
        cdef int j = <int> self._el_to_int[f]
```
Avoid Cython checking the type.

src/sage/sets/disjoint_set.pyx
```diff
@@ -797,23 +880,46 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
...
+            {{'a'}, {'b'}, {'c'}, {'d'}, {'e'}}
+            sage: e._union('a','b'); e
+            {{'a', 'b'}, {'c'}, {'d'}, {'e'}}
+            sage: e._union('c','e'); e
+            {{'a', 'b'}, {'c', 'e'}, {'d'}}
+            sage: e._union('b','e'); e
+            {{'a', 'b', 'c', 'e'}, {'d'}}
+        """
+        cdef int i = self._el_to_int[e]
+        cdef int j = self._el_to_int[f]
```
tscrim:
```suggestion
        cdef int i = <int> self._el_to_int[e]
        cdef int j = <int> self._el_to_int[f]
```

tscrim (commented):
Here's some things to try and make it a bit faster.

gmou3:
@mantepse I only kept the version with no input checks and renamed `union` to `join`. Note that the `Disjoint Set` data structure is also called a `union-find` data structure or `merge-find` data structure (is `merge` a better name? - see https://en.wikipedia.org/wiki/Disjoint-set_data_structure).

@tscrim Thanks for the suggestions; I want to make it as fast as possible. I did implement the recommended changes and I think it is a tiny bit faster. (It is not easy to come up with good tests.)

tscrim:
Thank you. Although I am -1 on renaming `union` to `join`. I believe `union` is a fairly standard name for this and a name conflict with a reserved C keyword is a poor argument IMO (as we are in Cython/Python and it is a method, which can have the same name as reserved keywords).

mantepse:
Yes, please keep union-find.

gmou3:
Ok, I managed a workaround and now `union` works.

videlec:
Calling `OP_find(OrbitPartition * p, int n)` without being sure that `n` is in the range of `p` might result in a `SEGFAULT`. While this is fine for `C` code it is **not** fine for pure Python code. Did you try something like
```python
sage: e = DisjointSet(5)
sage: e.find(2**30)
```

gmou3:
It causes a segmentation fault indeed. What should be done?

gmou3:
@videlec I think you will be satisfied with the last changes.
Doing these I also found out that `cdef` is faster than `cpdef`.

tscrim:
I believe @videlec's point is that user-facing code should not segfault on bad input (of course, internal code should not either, but there we can sanitize/control the input). So I believe your changes should have addressed this.

gmou3:
Yes, I think it is good now.

videlec:
Looking at it a bit more, I don't quite understand the purpose of this PR. If you want something fast(er), why don't you use directly the `OrbitPartition` C class? The `DisjointSet` provides one extra level of indirection which is a lot of overhead.

src/sage/sets/disjoint_set.pyx
```diff
@@ -379,28 +376,29 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
...
             [0, 1, 0, 0, 4]
         """
-        l = []
+        cdef Py_ssize_t card = self.cardinality()
+        cdef list l = [None] * card
         cdef int i
-        for i in range(self.cardinality()):
-            l.append(self._nodes.parent[i])
+        for i in range(card):
+            l[i] = self._nodes.parent[i]
```
mantepse, line 408: wouldn't `l = [self._nodes.parent[i] for i in range(self.cardinality())]` be better?

Hm, probably not: https://github.com/cython/cython/issues/2844

gmou3:
> Looking at it a bit more, I don't quite understand the purpose of this PR. If you want something fast(er), why don't you use directly the `OrbitPartition` C class? The `DisjointSet` provides one extra level of indirection which is a lot of overhead.

To some extent this is a fair point. However, the `DisjointSet` provides some extra functionality - e.g., the handling of hashables. Without any compromise in input checking, this PR still optimizes (at least) the `union` and `find` methods of the `DisjointSet_of_hashables` class. These methods performed unnecessary checks by passing through the `union` and `find` of `DisjointSet_of_integers`.

Given that now I call `OP_find` and `OP_union` directly from the `find` and `union` of `DisjointSet_of_hashables`, I wonder whether the `_find` and `_union` of `DisjointSet_of_integers` are still useful. Maybe if someone was to find them useful they should directly use the `OrbitPartition` class instead?

src/sage/sets/disjoint_set.pyx
gmou3: Yes, I doubt it would help.

videlec:
> > Looking at it a bit more, I don't quite understand the purpose of this PR. If you want something fast(er), why don't you use directly the `OrbitPartition` C class? The `DisjointSet` provides one extra level of indirection which is a lot of overhead.
> 
> To some extent this is a fair point. However, the `DisjointSet` provides some extra functionality - e.g., the handling of hashables. Without any compromise in input checking, this PR still optimizes (at least) the `union` and `find` methods of the `DisjointSet_of_hashables` class. These methods performed unnecessary checks by passing through the `union` and `find` of `DisjointSet_of_integers`.

Indeed, that looks like a good optimization!
 
> Given that now I call `OP_find` and `OP_union` directly from the `find` and `union` of `DisjointSet_of_hashables`, I wonder whether the `_find` and `_union` of `DisjointSet_of_integers` are still useful. Maybe if someone was to find them useful they should directly use the `OrbitPartition` class instead?

I would not bother keeping them.

mantepse:
> > Given that now I call `OP_find` and `OP_union` directly from the `find` and `union` of `DisjointSet_of_hashables`, I wonder whether the `_find` and `_union` of `DisjointSet_of_integers` are still useful. Maybe if someone was to find them useful they should directly use the `OrbitPartition` class instead?
> 
> I would not bother keeping them.

Maybe you could add a note in the docstring of `find` and `union` that these are the methods to use if no checking is needed?

gmou3:
@videlec Cool!

@mantepse:
> > > Given that now I call `OP_find` and `OP_union` directly from the `find` and `union` of `DisjointSet_of_hashables`, I wonder whether the `_find` and `_union` of `DisjointSet_of_integers` are still useful. Maybe if someone was to find them useful they should directly use the `OrbitPartition` class instead?
> > 
> > 
> > I would not bother keeping them.
> 
> Maybe you could add a note in the docstring of `find` and `union` that these are the methods to use if no checking is needed?

I removed them and referred to `OP_find` and `OP_union`. One should be able to use those directly and almost as easily.

src/sage/sets/disjoint_set.pyx
```diff
@@ -6,26 +6,25 @@ The main entry point is :func:`DisjointSet` which chooses the appropriate
...
 - Sébastien Labbé (2010-01) - Inclusion into sage (:issue:`6775`).
+- Giorgos Mousa (2024-04-22): Optimize
 
 EXAMPLES:
 
 Disjoint set of integers from ``0`` to ``n - 1``::
 
-    sage: s = DisjointSet(6)
-    sage: s
+    sage: s = DisjointSet(6); s
```
videlec: It is better to avoid semi-colon to improve readability [see E702](https://www.flake8rules.com/rules/E702.html).

src/sage/sets/disjoint_set.pyx
```diff
@@ -379,28 +373,29 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
...
+            sage: d.union(0, 3)
             sage: d.__getstate__()
             [0, 1, 2, 0, 4]
-            sage: d.union(2,0)
+            sage: d.union(2, 0)
             sage: d.__getstate__()
             [0, 1, 0, 0, 4]
         """
-        l = []
+        cdef Py_ssize_t card = self.cardinality()
```
videlec: Why not use directly `self._nodes.degree` here?

src/sage/sets/disjoint_set.pyx
```diff
@@ -531,57 +525,56 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
...
             [(1, [1]), (2, [0, 2, 3]), (4, [4])]
             sage: d
             {{0, 2, 3}, {1}, {4}}
         """
-        s = {}
-        cdef int i
+        cdef dict s = {}
+        cdef int i, o
         for i in range(self.cardinality()):
             o = self.find(i)
```
videlec: I thought it was faster to use `OP_find` :-)

src/sage/sets/disjoint_set.pyx
```diff
@@ -531,57 +525,56 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
...
             sage: e = d.element_to_root_dict(); e
             {0: 0, 1: 4, 2: 2, 3: 2, 4: 4}
             sage: WordMorphism(e)                                                       # needs sage.combinat
             WordMorphism: 0->0, 1->4, 2->2, 3->2, 4->4
         """
-        d = {}
+        cdef dict d = {}
         cdef int i
         for i in range(self.cardinality()):
             d[i] = self.find(i)
```
videlec: Again, use `OP_find`

src/sage/sets/disjoint_set.pyx
```diff
@@ -531,57 +525,56 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
...
+            sage: d.union(4, 1)
             sage: e = d.element_to_root_dict(); e
             {0: 0, 1: 4, 2: 2, 3: 2, 4: 4}
             sage: WordMorphism(e)                                                       # needs sage.combinat
             WordMorphism: 0->0, 1->4, 2->2, 3->2, 4->4
         """
-        d = {}
+        cdef dict d = {}
         cdef int i
         for i in range(self.cardinality()):
```
videlec: Again, use `self._nodes.degree`.

src/sage/sets/disjoint_set.pyx
gmou3: Not so elegant, but ok.

src/sage/sets/disjoint_set.pyx
gmou3: Indeed :O. I think these functions are not called as often as find and union.

videlec:
Looks good to me. Thanks.

tscrim (approved):
I will take @videlec's comment to mean a positive review.

PR #37836, opened 2024-04-20
fricas/do not get inputform twice

mantepse:
By mistake, we were previously converting the InputForm to a string twice, which would be quite costly especially for large expressions.


dimpase (approved):
lgtm

PR #37839, opened 2024-04-20
Cythonize `GraphicMatroid`

gmou3:
A metamorphosis of `graphic_matroid.py` into `graphic_matroid.pyx` and `graphic_matroid.pxd`.
Things should be somewhat faster; test, e.g., the `GraphicMatroid`'s `_circuit` function. An indirect and diluted test that shows improvement can be seen by running `%time matroids.CompleteGraphic(8).circuits()` (~half the time).

`For the reviewer:` The changes of this PR are only inside the `matroids` module. The rest come from #37835.

### :hourglass: Dependencies

#37835: Deeper improvements of `DisjointSet`.


src/sage/matroids/graphic_matroid.pyx
```diff
@@ -380,7 +371,7 @@ def __eq__(self, other):
 
         INPUT:
 
-        - ``other`` -- a matroid
+        - ``other`` -- matroid
```
kwankyu, line 375: You are systematically removing the articles "a" or "an" from INPUT descriptions. However, it is a common practice to add articles. I think it reads better with an article. For example, 
```
- ``other`` -- a matroid
```
is read as "``other`` is a matroid".

tscrim: I don't think it is so common, as evidenced by the file (e.g., `_minor` has simply `frozenset`). It is to convey the type of the object, not plain English. I would leave it to the PR author to decide.

kwankyu: If it is new code, I would leave it to the author. But the PR author is changing existing code. Then it should be an improvement. I don't think it is an improvement.

kwankyu: If it is not a personal taste but a clear improvement, then we may think of making the same changes to the whole codebase. But do we?

gmou3: I am following the `INPUT` example of the [guide](https://doc.sagemath.org/html/en/developer/coding_basics.html). However, there does not seem to be a clear rule concerning articles, and they are used in many places. My intent behind these changes is to at least achieve some consistency within the `matroids` module.

kwankyu: > ... My intent behind these changes is to at least achieve some consistency within the `matroids` module.

If it is, then no problem. I saw only your changes not the whole file.

kwankyu: Achieving consistency is an improvement :-)

gmou3: :D (maybe a rule would be good too; I think that the docstrings should be parsed through some program.)

kwankyu: Even if we could devise a program (though I doubt),  we could not get a necessary consensus on the issue, that is, a rule for the program.

Thus we have https://doc.sagemath.org/html/en/developer/coding_basics.html#fine-points-on-styles.

gmou3: True. A program can incorporate such laxity though.

src/sage/matroids/graphic_matroid.pyx
```diff
@@ -277,40 +271,37 @@ def _rank(self, X):
             sage: M.rank([0,3])
             1
         """
-        from sage.sets.disjoint_set import DisjointSet
-
-        edges = self.groundset_to_edges(X)
-        vertices = set([u for (u, v, l) in edges]).union(
-            [v for (u, v, l) in edges])
+        cdef DisjointSet_of_hashables DS_vertices
+        cdef list edges = self.groundset_to_edges(X)
```
tscrim, line 275:
```suggestion
        cdef list edges = self.groundset_to_edges(X)
```
Wouldn't it be better for the method to know its output type is a `list`? This applies in many places/methods. This helps Cython know to avoid the type checking when generating its C code.. Although I guess this isn't done at-large in the matroid Cython code...

src/sage/matroids/graphic_matroid.pyx
```diff
@@ -720,32 +707,32 @@ def _closure(self, X):
...
-        g = self._subgraph_from_set(X)
-        V = g.vertices(sort=False)
-        components = g.connected_components_number()
+        cdef set XX = set(X)
+        cdef frozenset Y = self.groundset().difference(XX)
+        cdef list edgelist = self._groundset_to_edges(Y)
+        cdef GenericGraph_pyx g = self._subgraph_from_set(XX)
+        cdef list V = g.vertices(sort=False)
+        cdef int components = g.connected_components_number()
         for e in edgelist:
```
tscrim:
```suggestion
        cdef frozenset Y = self._groundset.difference(XX)
        cdef list edgelist = self._groundset_to_edges(Y)
        cdef GenericGraph_pyx g = self._subgraph_from_set(XX)
        cdef list V = g.vertices(sort=False)
        cdef int components = g.connected_components_number()
        cdef tuple e
        for e in edgelist:
```
Since the edge is going to be a tuple and perhaps explicit type casting to avoid unnecessary type checking.

tscrim (commented):
It might be good to look at the C code generated in performance critical sections to see if there are additional places where type casting and other such declarations could simplify the C code (and hence, increase performance). Although I am basically okay to include this in the current state and do other general improvements in later PRs. It is up to you.

src/sage/matroids/graphic_matroid.pyx
gmou3: I am not sure whether it is better, but I can do it all over the matroid code.

gmou3:
> It might be good to look at the C code generated in performance critical sections to see if there are additional places where type casting and other such declarations could simplify the C code (and hence, increase performance). Although I am basically okay to include this in the current state and do other general improvements in later PRs. It is up to you.

I hadn't thought of doing that and I'll keep the idea in mind for future PRs.

src/sage/matroids/graphic_matroid.pyx
tscrim: My understanding and intuition (i.e., without explicitly checking) is that this involves a Cython type check that the output is indeed a `list` before assigning it to `edges`. While on any individual operation, this is nominal, but in functions that can be called in tight loops, these little bits can add up and take away from the full Cython speed. Anyways, it's not something that necessarily needs to be done here.

tscrim (commented):
Okay, then let's get this in. It seems like the tester failures are unrelated, but it would be good to have them be run... I don't know how to kick it to run it again.

tscrim (approved):

src/sage/matroids/graphic_matroid.pyx
gmou3: I see. I may do this for the whole of the `matroids` module on some moonlit night.

gmou3:
Please don't kick it as it may further break things.
There are many errors indeed; alas, what a state of disrepair!

mkoeppe:
I've repaired the CI, which complains:
```
sage -t --random-seed=109326302233194564031839857779385647036 src/sage/matroids/graphic_matroid.py
File "src/sage/matroids/graphic_matroid.py", line 490, in sage.matroids.graphic_matroid.GraphicMatroid._minor
Failed example:
    M._minor(deletions=frozenset([0,1,2]))
Exception raised:
      File "/sage/src/sage/doctest/forker.py", line 714, in _run
      File "/sage/src/sage/doctest/forker.py", line 1146, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.matroids.graphic_matroid.GraphicMatroid._minor[1]>", line 1, in <module>
        M._minor(deletions=frozenset([Integer(0),Integer(1),Integer(2)]))
      File "sage/matroids/graphic_matroid.pyx", line 462, in sage.matroids.graphic_matroid.GraphicMatroid._minor
        cpdef _minor(self, contractions, deletions):
    TypeError: _minor() takes exactly 2 positional arguments (0 given)
File "src/sage/matroids/graphic_matroid.py", line 492, in sage.matroids.graphic_matroid.GraphicMatroid._minor
Failed example:
    M._minor(contractions=frozenset([0,1,2]))
Exception raised:
      File "/sage/src/sage/doctest/forker.py", line 714, in _run
      File "/sage/src/sage/doctest/forker.py", line 11[46](https://github.com/sagemath/sage/actions/runs/8959227503/job/24712775461?pr=37839#step:12:47), in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.matroids.graphic_matroid.GraphicMatroid._minor[2]>", line 1, in <module>
        M._minor(contractions=frozenset([Integer(0),Integer(1),Integer(2)]))
      File "sage/matroids/graphic_matroid.pyx", line 462, in sage.matroids.graphic_matroid.GraphicMatroid._minor
        cpdef _minor(self, contractions, deletions):
    TypeError: _minor() takes exactly 2 positional arguments (1 given)
```

tscrim:
Something is wrong with the CI. There is no more `graphic_matroid.py` file as it is changed to a `.pyx` file.

mkoeppe:
Indeed, sorry for the noise.

PR #37840, opened 2024-04-20
`sage-logger`: Replace use of `/usr/bin/time` by bash keyword `time`

### LINKED ISSUE #37832: /usr/bin/time should be a pre-req
from #37785

Instead of suppressing things, just make `time` a pre-req. In particular, as it appears to cause breakage beyond warnings only, see

https://groups.google.com/g/sage-devel/c/l2MFDazkpMQ/m/_t9LFC06BAAJ
```
file:///home/pi/sage/local/var/lib/sage/venv-python3.12/var/lib/sage/wheels/cypari2-2.1.5-cp312-cp312-linux_x86_64.whl->-r
/home/pi/sage/local/var/lib/sage/venv-python3.12/var/lib/sage/scripts/cypari/spkg-requirements.txt
(line 1)) (1.11.4)
[cypari-2.1.5] [spkg-pipinst] Installing collected packages: cypari2
[cypari-2.1.5] [spkg-pipinst] Successfully installed cypari2-2.1.5
[cypari-2.1.5] [spkg-pipinst] cat: ./spkg-pipinst.time: No such file or
directory
[cypari-2.1.5] Deleting build directory
/home/pi/sage/local/var/lib/sage/venv-python3.12/var/tmp/sage/build/cypari-2.1.5
[cypari-2.1.5] Finished installing cypari-2.1.5
[cypari-2.1.5] cat: /home/pi/sage/logs/pkgs/cypari-2.1.5.time: Aucun
fichier ou dossier de ce nom
make[2]: *** [Makefile:3052 : all-build] Erreur 2
make[2] : on quitte le répertoire « /home/pi/sage/build/make » 
```

mkoeppe:
Follow-up after 
- https://github.com/sagemath/sage/pull/37391
- #37785

... to make the display of package build times available on Linux too.

Fixes https://github.com/sagemath/sage/pull/37785#issuecomment-2060170610, closes #37832. 

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37785

### DIFF
--- a/build/bin/sage-logger
+++ b/build/bin/sage-logger
@@ -63,21 +63,19 @@ fi
 
 timefile="$logdir/$logname.time"
 rm -f "$timefile"
-if /usr/bin/time -h -o /dev/null true 2>/dev/null; then
-    TIME="/usr/bin/time -h -o $timefile"
-else
-    TIME=""
-fi
+time_cmd() {
+    local max=$(($SECONDS+10))
+    exec 3>&1 4>&2
+    local out
+    TIME_OUTPUT=$((time 1>&3 2>&4 sh -c "$1") 2>&1)
+    local retstat=$?
+    [ "$SECONDS" -lt "$max" ] || echo "$TIME_OUTPUT" > "$timefile"
+    return $retstat
+}
 
 report_time ()
 {
-    time=$(echo $(cat $timefile 2>/dev/null))
-    case "$time" in
-        *m*real*|*h*real*|*[1-9][0-9].*real*|*[1-9][0-9],*real*)
-            # at least 10 seconds wall time
-            echo "$time"
-            ;;
-    esac
+    [ -r "$timefile" ] && echo $(< $timefile)
 }
 
 mkdir -p "$logdir"
@@ -94,8 +92,7 @@ if [ -n "$SAGE_SILENT_BUILD" -a ${use_prefix} = true ]; then
     # Silent build.
     # Similar to https://www.gnu.org/software/automake/manual/html_node/Automake-Silent-Rules.html#Automake-Silent-Rules
     echo "[$logname] installing. Log file: $logfile"
-    ( exec>> $logfile 2>&1 ; $TIME sh -c "$cmd"; status=$?;
-      [ -r $timefile ] && cat $timefile; exit $status )
+    ( exec>> $logfile 2>&1; time_cmd "$cmd"; status=$?; report_time; exit $status )
     status=$?
     if [[ $status != 0 ]]; then
         echo "  [$logname] error installing, exit status $status. End of log file:"
@@ -115,9 +112,7 @@ else
     # We trap SIGINT such that SIGINT interrupts the main process being
     # run, not the logging.
 
-    ( exec 2>&1;
-      $TIME sh -c "$cmd"; status=$?; report_time;
-      exit $status ) | \
+    ( exec 2>&1; time_cmd "$cmd"; status=$?; report_time; exit $status ) | \
         ( trap '' SIGINT; if [ -n "$GITHUB_ACTIONS" -a -n "$prefix" ]; then echo "::group::${logname}"; fi; tee -a "$logfile" | $SED; if [ -n "$GITHUB_ACTIONS" -a -n "$prefix" ]; then echo "::endgroup::"; fi )
 
     pipestatus=(${PIPESTATUS[*]})


tornaria:
Do we really need to have different functions for different systems? It seems this kind of case by case handling might lead to issues later.

May I suggest something like this, which depends only on `$SECONDS` (available on bash) and will use whatever the user has as `time` (usually bash builtin)
```
time_silent_if_less_than() { 
  local max=$(($SECONDS+$1))
  shift;
  exec 3>&1 4>&2
  local out=$((time 1>&3 2>&4 "$@") 2>&1)
  [ "$SECONDS" -lt "$max" ] || echo "$out" >&2
}
```

Usage:
```
$ time_silent_if_less_than 10 sh -c "sleep 5 ; echo stdout && echo stderr >&2" 
stdout
stderr
$ time_silent_if_less_than 10 sh -c "sleep 15 ; echo stdout && echo stderr >&2" 
stdout
stderr

real	0m15.006s
user	0m0.001s
sys	0m0.005s
$ time_silent_if_less_than 2 sh -c "sleep 65 ; echo stdout && echo stderr >&2" 
stdout
stderr

real	1m5.007s
user	0m0.001s
sys	0m0.006s
```

Variation:
```
time_silent_if_less_than() { 
  local max=$(($SECONDS+$1))
  shift;
  exec 3>&1 4>&2
  local out=$((time 1>&3 2>&4 "$@") 2>&1)
  [ "$SECONDS" -lt "$max" ] || echo $out >&2
}
$ time_silent_if_less_than 10 sh -c "sleep 15 ; echo stdout && echo stderr >&2" 
stdout
stderr
real 0m15.006s user 0m0.000s sys 0m0.006s
```

mkoeppe:
> It seems this kind of case by case handling might lead to issues later.

What "issues" could that possibly be?

tornaria:
> > It seems this kind of case by case handling might lead to issues later.
> 
> What "issues" could that possibly be?

I don't know, someone wants to change something about this code, and has to be aware of the 3-4 different versions of the code. Just like #37391 was only tested on mac and seemed to work fine, having different implementations for different platforms is a risk. Since everything can e implemented with bash, we just defer to bash having the same behavior on the different platforms.

mkoeppe:
> having different implementations for different platforms is a risk

Based on my experience with portability work, let me just say that I don't find this believable.

mkoeppe:
> Since everything can be implemented with bash, we just defer to bash having the same behavior on the different platforms.

Let's check in with our most prominent antibashist (#29345, #29344, #30128) in our project. @orlitzky, any thoughts?

tornaria:
Note that `$SECONDS` can be replaced by `$(date +%s)` if you are concerned about that.

orlitzky:
Bash is fine so long as the shebang doesn't say `#!/bin/sh`. If bash makes this easy to fix, then I would use bash.

In case it helps, the comments to the effect that bash _time_ can't redirect to a file aren't quite right. For example (in bash),

```
$ (time echo "hello world!") 2> derp
hello world!
$ cat derp

real	0m0.000s
user	0m0.000s
sys	0m0.000s
```

Problem is, this will redirect stderr for the inner command, too. I'm sure there's a way around it by juggling file descriptors.

tornaria:
> Bash is fine so long as the shebang doesn't say `#!/bin/sh`. If bash makes this easy to fix, then I would use bash.
> 
> In case it helps, the comments to the effect that bash _time_ can't redirect to a file aren't quite right. For example (in bash),
> 
> ```
> $ (time echo "hello world!") 2> derp
> hello world!
> $ cat derp
> 
> real	0m0.000s
> user	0m0.000s
> sys	0m0.000s
> ```
> 
> Problem is, this will redirect stderr for the inner command, too. I'm sure there's a way around it by juggling file descriptors.

Here: https://github.com/sagemath/sage/pull/37840#issuecomment-2068136415

kwankyu:
A compromise
```bash
time_cmd() {
    exec 3>&1 4>&2
    local time=$({ time sh -c "$@" 1>&3 2>&4; } 2>& 1 )
    echo "$time" > $timefile
}

report_time() {
    local time=$(echo $(cat $timefile 2>/dev/null))
    if [[ $time =~ real\ ([0-9]+)m([0-9]+)\.[0-9]+s ]]; then
        local minutes="${BASH_REMATCH[1]}"
        local seconds="${BASH_REMATCH[2]}"
        total_seconds=$((minutes * 60 + seconds))
        if (( total_seconds >= 10 )); then
            echo "$time"
        fi
    fi
}  
```
Would this work?

tornaria:
> A compromise
> 
> ```shell
> time_cmd() {
>     exec 3>&1 4>&2
>     local time=$({ time sh -c "$@" 1>&3 2>&4; } 2>& 1 )
>     echo "$time" > $timefile
> }
> 
> report_time() {
>     local time=$(echo $(cat $timefile 2>/dev/null))
>     if [[ $time =~ real\ ([0-9]+)m([0-9]+)\.[0-9]+s ]]; then
>         local minutes="${BASH_REMATCH[1]}"
>         local seconds="${BASH_REMATCH[2]}"
>         total_seconds=$((minutes * 60 + seconds))
>         if (( total_seconds >= 10 )); then
>             echo "$time"
>         fi
>     fi
> }  
> ```
> 
> Would this work?

LGTM. Two comments:
1. if we are using bash specifics anyway, using `$SECONDS` to detect whether 10 seconds have elapsed seems easier and cleaner (no parsing)
2. the reporting of time can be done inside `time_cmd()` itself, avoiding saving through a file, and the caller doesn't need to know about it. You just get a `time_cmd` which prints the time only if more than 10 seconds.

This is more or less what I proposed here:
https://github.com/sagemath/sage/pull/37840#issuecomment-2068136415

if you don't like the extra parameter, you can hardcode "10" there. Something like:
```bash
time_cmd() { 
    local start_seconds=$SECONDS
    exec 3>&1 4>&2
    local time=$({ time sh -c "$@" 1>&3 2>&4; } 2>& 1 )
    local total_seconds=$(($SECONDS - $start_seconds))
    if (( total_seconds >= 10 )); then
        echo "$time"
    fi
}
```

kwankyu:
> 1. if we are using bash specifics anyway, using `$SECONDS` to detect whether 10 seconds have elapsed seems easier and cleaner (no parsing)
> 2. the reporting of time can be done inside `time_cmd()` itself, avoiding saving through a file, and the caller doesn't need to know about it. You just get a `time_cmd` which prints the time only if more than 10 seconds.

My variant is a compromise, in the sense that the suggested commands can be replaced with Matthias' `time_cmd()` and `report_time()`. 

Anyway, this PR is Matthias's. He is free to adopt whatever we suggest or not.

mkoeppe:
Finally coming back to this PR. Thanks all for the suggestions. I'm preparing a version that adapts from the suggested solutions that use the bash keyword `time` and output redirection.

kwankyu (approved):
Works well. LGTM. Thanks.

mkoeppe:
Thanks!

build/bin/sage-logger
```diff
@@ -63,21 +63,19 @@ fi
...
-else
-    TIME=""
-fi
+time_cmd() {
+    local max=$(($SECONDS+10))
+    exec 3>&1 4>&2
+    local out
+    TIME_OUTPUT=$((time 1>&3 2>&4 sh -c "$1") 2>&1)
+    local retstat=$?
+    [ "$SECONDS" -lt "$max" ] || echo "$TIME_OUTPUT" > "$timefile"
```
tornaria, line 72:
```suggestion
    [ "$SECONDS" -lt "$max" ] || echo "$TIME_OUTPUT" >&2
```
you don't really need to go via "$timefile", do you?

It seems more natural if `time_cmd` has similar semantics/usage as regular `time` (except the time is only printed out if >= 10s).

build/bin/sage-logger
```diff
@@ -94,8 +92,7 @@ if [ -n "$SAGE_SILENT_BUILD" -a ${use_prefix} = true ]; then
     # Silent build.
     # Similar to https://www.gnu.org/software/automake/manual/html_node/Automake-Silent-Rules.html#Automake-Silent-Rules
     echo "[$logname] installing. Log file: $logfile"
-    ( exec>> $logfile 2>&1 ; $TIME sh -c "$cmd"; status=$?;
-      [ -r $timefile ] && cat $timefile; exit $status )
+    ( exec>> $logfile 2>&1; time_cmd "$cmd"; status=$?; report_time; exit $status )
```
tornaria, line 95:
```suggestion
    ( exec>> $logfile 2>&1; time_cmd "$cmd" )
```
or even maybe
```suggestion
    time cmd "$cmd" >> $logfile 2>&$1
```

Also, in L104 below, the time is reported again.

build/bin/sage-logger
```diff
@@ -115,9 +112,7 @@ else
     # We trap SIGINT such that SIGINT interrupts the main process being
     # run, not the logging.
 
-    ( exec 2>&1;
-      $TIME sh -c "$cmd"; status=$?; report_time;
-      exit $status ) | \
+    ( exec 2>&1; time_cmd "$cmd"; status=$?; report_time; exit $status ) | \
```
tornaria, line 115: same here...

build/bin/sage-logger
```diff
@@ -63,21 +63,19 @@ fi
...
 rm -f "$timefile"
-if /usr/bin/time -h -o /dev/null true 2>/dev/null; then
-    TIME="/usr/bin/time -h -o $timefile"
-else
-    TIME=""
-fi
+time_cmd() {
+    local max=$(($SECONDS+10))
+    exec 3>&1 4>&2
+    local out
```
tornaria, line 69: unused

build/bin/sage-logger
```diff
@@ -63,21 +63,19 @@ fi
...
+    TIME_OUTPUT=$((time 1>&3 2>&4 sh -c "$1") 2>&1)
+    local retstat=$?
```
tornaria, line 71:
```suggestion
    local TIME_OUTPUT=$((time 1>&3 2>&4 sh -c "$1") 2>&1) retstat=$?
```
if you want't `TIME_OUTPUT` to be local, this is one way.

OTOH, if you keep `TIME_OUTPUT` global you can use it instead of `timefile`.

build/bin/sage-logger
mkoeppe: Yes, I do. This thing runs in a subshell.

tornaria: My point is: print to stderr instead of saving to a file. What's wrong with that? Saves a point of failure (communication via filesystem).

build/bin/sage-logger
mkoeppe: thanks, I'll clean this one up in a follow up.

build/bin/sage-logger
mkoeppe: One goes to the logfile. One goes to the terminal.

kwankyu:
@tornaria Thanks for reviewing and for additional comments. But let me express my worries about certain attitude of reviewers that I am observing recently. First my apologies for making remarks toward fellow reviewers, which I usually don't.

From our doc on "reviewing", we see 
> “The perfect is the enemy of the good”
>
> The point of the review is to ensure that the Sage code guidelines are followed and that the implementation is mathematically correct. Please refrain from additional feature requests or open-ended discussion about alternative implementations. If you want the code written differently, your suggestion should be a clear and actionable request. 

We develop sage "incrementally". A PR should be good enough to be accepted, but it need not be "perfect" from one reviewer's point of view. If the reviewer has an idea to improve the code further, then he can open up his own PR building on the present PR, implementing the idea.

If I have mistaken your good intention, then my apologies again.

build/bin/sage-logger
mkoeppe: This is done in #38026.

build/bin/sage-logger
mkoeppe: From my experience as the principal maintainer of the Sage distro build system, let me just say that this concern ("point of failure") is wildly beside the point.

tornaria:
> If I have mistaken your good intention, then my apologies again.

I don't understand what you mean.

I'll make sure in the future to keep my 2c to myself.

PR #37841, opened 2024-04-21
`.ci/write-dockerfile.sh`: Move here from build/bin/

mkoeppe:
... as suggested by @kwankyu in https://groups.google.com/g/sage-devel/c/dEa3i2Fn3ZY/m/nzUr2wc_AAAJ

To test: Commands such as `tox -e docker-ubuntu-jammy-minimal -- config.status` should still work.

### :memo: Checklist


.github/workflows/docker.yml
```diff
@@ -183,8 +183,10 @@ jobs:
           sudo apt-get clean
           df -h
       - name: Update Sage packages from upstream artifact
+        # Handle both the old and new location of write-dockerfile.sh,
+        # because docker.yml is a reusable workflow.
         run: |
```
kwankyu, line 188: Sorry, but would you explain this: "because docker.yml is a reusable workflow"?

kwankyu: I think I know what "a reusable workflow" is, but why need to handle both the old and new location?

kwankyu:
> To test: Commands such as `tox -e docker-ubuntu-jammy-minimal -- config.status` should still work.

Yes, this works well.

.github/workflows/docker.yml
mkoeppe: Reusable = invoked by other workflows, which includes workflows in other repositories, see links in 
- https://github.com/sagemath/sage/issues/33338

mkoeppe: I cannot simultaneously update all of these workflows, so our reusable workflow needs to support both versions until such updates are completed.

kwankyu: I see. 

The upstream repos use the reusable workflow, `.github/workflows/docker.yml`. But they do not directly use `write-dockerfile.sh`, which is the only file moved to the new location. Right?

Then, as long as the location of  `.github/workflows/docker.yml` doesn't change, the new location of `write-dockerfile.sh` doesn't matter for the upstream repos. Does it?

mkoeppe: That's basically true, but the URL of the workflow and the branch of the repository that it's using are specified separately, see for example:
- https://github.com/scipopt/SCIP-SDP/blob/main/.github/workflows/ci-sage.yml#L87
- https://github.com/scipopt/SCIP-SDP/blob/main/.github/workflows/ci-sage.yml#L97

They may be out of sync for some of the workflows.

kwankyu: This business is more complicated in practice than it should be...

OK. Thanks for the explanation.

kwankyu (approved):
LGTM.

mkoeppe:
Thanks for the review!

PR #36269, opened 2023-09-15
`coefficients` for `QuasiModularFormsElement`

seewoo5:
This patch implements `_compute` and `coefficients` methods to `QuasiModularFormsElement`, which already exists in `ModularFormsElement`. I made this for my own usage.
(This is my first PR for Sage, so comments are welcome!)

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation accordingly.

PR #37842, opened 2024-04-21
Lazy q-expansions for modular forms

DavidAyotte:
This PR aims to implement q-expansion of modular forms as Lazy power series as suggested by @mantepse in PR #36269. In short we want to make the following code work:
```
sage: M = ModularForms(1, 12)
sage: f = M.0
sage: fqexp = f.q_expansion(oo)
sage: fqexp
q - 24*q^2 + 252*q^3 - 1472*q^4 + 4830*q^5 - 6048*q^6 + O(q^7)
sage: fqexp.parent()
Lazy Taylor Series Ring in q over Rational Field
sage: fqexp[1000]
-30328412970240000
```
I'm marking this as a draft right now, but the initial commit does implement the desired functionality. Note that I also plan to implement this for quasimodular forms and graded modular forms (elements of modular forms ring).

#### Quick facts about modular forms (if you don't know about them)

Modular forms are holomorphic functions $f:\mathcal{H} \rightarrow \mathbb{C}$, where $\mathcal{H}$ is the complex upper half plane, satisfying two conditions:
* an invariance condition under the action of $\mathrm{SL}_2(\mathbb{Z})$ on $\mathcal{H}$ (given by linear fractional transformations)
* it admits a $q$-expansion of the form: $f = \sum_{n \geq 0} a_n q^n$

The "weight" of a modular form is an integer related with the first condition. Lastly, an important fact about spaces of modular forms of fixed weight is that they are finite dimensional vector spaces (over $\mathbb{C}$). We can furthermore compute a basis such that its elements have $q$-expansion with algebraic (even rational in some case) coefficients.

In this PR, we want to implement the $q$-expansion of any modular forms $f$ as a Lazy power series.

CC: @mantepse @tscrim I'm tagging you because you both are involved with Lazy power series and I'm hoping that you bring valuable input (I'm not an expert of Lazy power series). 

### :memo: Checklist

### DIFF
--- a/src/sage/modular/modform/element.py
+++ b/src/sage/modular/modform/element.py
@@ -557,8 +557,15 @@ class ModularForm_abstract(ModuleElement):
             ...
             ValueError: prec (= -1) must be non-negative
         """
+        from sage.rings.infinity import PlusInfinity
         if prec is None:
             prec = self.parent().prec()
+        if prec is PlusInfinity():
+            from sage.rings.lazy_series_ring import LazyPowerSeriesRing
+            L = LazyPowerSeriesRing(self.base_ring(),
+                                    names=defaults.DEFAULT_VARIABLE)
+            an = lambda n: self._compute([n])[0]
+            return L(an)
         prec = Integer(prec)
         try:
             current_prec, f = self.__q_expansion


src/sage/modular/modform/element.py
```diff
@@ -557,8 +557,15 @@ def q_expansion(self, prec=None):
...
             ValueError: prec (= -1) must be non-negative
         """
+        from sage.rings.infinity import PlusInfinity
         if prec is None:
             prec = self.parent().prec()
+        if prec is PlusInfinity():
+            from sage.rings.lazy_series_ring import LazyPowerSeriesRing
+            L = LazyPowerSeriesRing(self.base_ring(),
+                                    names=defaults.DEFAULT_VARIABLE)
+            an = lambda n: self._compute([n])[0]
```
mantepse, line 567: It seems to me that `self._compute([n])[0]` just does `self.q_expansion(n + 1)[n]`, and this in turn really calls `self._compute_q_expansion` and caches the result in `self.__q_expansion`.

So I guess it would make more sense to make `self.__q_expansion` a `LazyPowerSeries` from the beginning, because then you can get rid of all the bookkeeping.

I don't know what `self.modular_symbols(1).q_eigenform(prec, names=self._name())` does, but maybe `q_eigenform` should accept infinity as precision.

I don't understand right now how `q_eigenform` is computed, but I imagine that it could be much easier if done lazily.  Can you write it in math terms for me?  This might also be a good performance test for the lazy power series.

tscrim: +1 on @mantepse's suggestion.

tscrim: Actually, I might go further and say that the caching that is done with `q_eigenform` instead be a lazy series (which is cached).

(I find having the name parameter being part of the caching information a little strange; I think it would be better to do this generically and just convert the output data as it is just printing output. Perhaps there is a reason for this I don't see from looking at the code locally.)

DavidAyotte: Hello both of you and thank you for the suggestions!
One thing here is that the current implementation computes the $q$-expansion of a modular form at the _parent level_ and not at the element level. An important fact is that the space of all modular forms of fixed weight $k$ and fixed congruence subgroup $\Gamma$ is a **finite** $K$-vector space $M_k(\Gamma)$ ($K$ a number field). Thus, the implementation fixes a basis and then simply sees a modular form as a linear combination of that basis.

To compute the expansion of any modular form, the code computes the expansion of each element of that basis and then sum the result. Here's an example to see this in action:
```
sage: M = ModularForms(DirichletGroup(5).0, 7)
sage: M
Modular Forms space of dimension 4, character [zeta4] and weight 7 over Cyclotomic Field of order 4 and degree 2
sage: f = M.0
sage: f.coefficient(100)  # long time
-216600*zeta4 + 668800
sage: g = M.1
sage: g.coefficient(100)  # quick
50525*zeta4 - 75075
```
In the above example, the computation of the 100-th coefficient of $f$ actually computes the coefficients of the other elements in the basis so that why the computation of the 100-th coefficient of $g$ is very fast. 

This was designed as such because the algorithm relies heavily on the computations of matrices of some operators, known as the Hecke operators, and this computation essentially implies the computations of the coefficients of all basis elements. If you want to dig deeper, the actual code in sage that does this is the method `sage.modular.modsym.space.ModularSymbolsSpace._q_expansion_basis_hecke_dual`. I skipped a lot of maths here, so if you want to dig deeped I would suggest that you look into Stein's [book](https://wstein.org/books/modform/stein-modform.pdf) (more or less section 3.6).

I don't know whether it would be possible to change this design of computing the coefficients of all basis elements and, if possible, how much work this would involve (surely more than the scope of this PR), but I would be interested in exploring this avenue.

tscrim: @DavidAyotte You mean the vector space is finite dimensional (since number fields must contain $\mathbb{Q}$ as I understand it), right?

What we are suggesting is to use the lazy series as the cache. This won't change the design of how the coefficients are computed, but how they are stored. Although I guess our current design isn't so good at computations that involve simultaneous updates. So perhaps this isn't so feasible? Or we just have some duplicated cache information, but I guess that is already there with the `__coefficients` in the elements. At the very least, it makes sense to move that caching into the lazy series. So the `q_expansion(prec=oo)` is cached in `__q_expansion` and
```python
def coefficient(self, n):
    return self.__q_expansion[n]

def coefficients(self, n):
    if n in ZZ:
        n = range(n)
    return [self.__q_expansion[i] for i in n]

def q_expansion(self, prec):
    if prec is PlusInfinity:
        return self.__q_expansion
    PSR = self.parent()._q_expansion_ring()
    f = PSR(self.coefficients(prec-1))
    return f.add_bigoh(prec)
```
However, it would be better to have the parent `_compute_q_expansion_basis()` return a sequence of lazy power series ring elements (i.e., have infinite precision). Then `self.__q_expansion` could just be set to `self.parent()._q_expansion(self.element())`. I can probably sketch everything out for this in the next few days if the idea isn't clear.

tscrim: @mantepse We should add an `add_bigoh` method to the lazy power series ring elements and a coercion to the finite precision series.

DavidAyotte: > @DavidAyotte You mean the vector space is finite dimensional (since number fields must contain Q as I understand it), right?

Yes and no, sorry I was unclear before, let me be more precise. The space of modular forms $M_k(\Gamma)$ for a fixed weight $k\in \mathbb{Z}$ and congruence subgroup $\Gamma\leq \mathrm{SL}_2(\mathbb{Z})$ is a finite dimensional $K$-vector space, but it is not because $K$ is finite dimensional over $\mathbb{Q}$. This fact is definitely not trivial, but important for the current implementation. I could go further and say that one can decompose the space of modular form as a direct sum:
$$M_k(\Gamma; K) \cong E_k(\Gamma) \oplus S_k(\Gamma)$$
where $E_k(\Gamma)$ is the space of Eisenstein forms and $S_k(\Gamma)$ (generated by Eisenstein series) is the space of cusps forms. The difficulty in computing a basis mainly relies in computing a basis of the subspace $S_k(\Gamma)$, as one usually know the expansion of Eisenstein series (implementing Lazy power series for those shouldn't be a problem).

> What we are suggesting is to use the lazy series as the cache. This won't change the design of how the coefficients are computed, but how they are stored. Although I guess our current design isn't so good at computations that involve simultaneous updates. So perhaps this isn't so feasible? Or we just have some duplicated cache information, but I guess that is already there with the `__coefficients` in the elements. At the very least, it makes sense to move that caching into the lazy series. So the `q_expansion(prec=oo)` is cached in `__q_expansion` and
> 
> ```python
> def coefficient(self, n):
>     return self.__q_expansion[n]
> 
> def coefficients(self, n):
>     if n in ZZ:
>         n = range(n)
>     return [self.__q_expansion[i] for i in n]
> 
> def q_expansion(self, prec):
>     if prec is PlusInfinity:
>         return self.__q_expansion
>     PSR = self.parent()._q_expansion_ring()
>     f = PSR(self.coefficients(prec-1))
>     return f.add_bigoh(prec)
> ```
> 
> However, it would be better to have the parent `_compute_q_expansion_basis()` return a sequence of lazy power series ring elements (i.e., have infinite precision). Then `self.__q_expansion` could just be set to `self.parent()._q_expansion(self.element())`. I can probably sketch everything out for this in the next few days if the idea isn't clear.

Ok thanks, I better understand your suggestions and I think that it makes sense. I will evaluate in the coming days if this can be achieved and come back to you.

mantepse: Taking the risk of stating the obvious, there are essentially two ways to specify a `LazyPowerSeries`:
1. by supplying an explicit formula for the coefficients, or
2. using recursion.

The first option, in combination with asking for a `sparse` cache, has the advantage that only required terms are computed, that is, asking for the n-th term will compute exactly this term and no others.

The second option has the advantage that you have a wide range of possibilities of specifying the series.  In particular, you can definite a series `s` implicitly using a functional equation of the form `s = F(s)`, and this may be very efficient.  Let me note that most operations we could think of are meanwhile implemented.  For example, we support infinite sums (and, mutatis mutandis, products) of the form $\sum_k s_k(q)$ provided that the valuation of $s_k$ is weakly increasing.

Finally, let me point out that we are not limited to formal power series in one (or several) variables. The infrastructure works for any $\mathbf N$ graded ring.

I had a brief look at algorithm 3.26 in the book, but not yet enough to understand it.

DavidAyotte: Hi Martin, thank you for the precisions. About the algorithm, the idea is that there exists a sequence of linear maps known as Hecke operators and denoted $T_n$ for $n\geq 1$. These linear maps act on the vector space of cusps forms and satisfies various interesting and useful properties.

One of those properties is that there is a way for computing the matrix representation of these maps, denoted $[T_n]$ under some basis (a basis of the space of "modular symbols", a space on which the Hecke operators also acts). Note that this matrix $[T_n]$ is of dimension $d\times d$ where $d$ is the dimension of the space of cusps forms, which one can compute before running the algorithm. Let's denote by $a_{ij}([T_n])$ the coefficient of $[T_n]$ at position $(i, j)$

One of the key fact on which the algorithm relies is that the following expansion
$$f_{ij} := \sum_{n\geq 1} a_{ij}([T_n]) q^n$$
**defines** a cusps form **for any** $i$ and $j$ fixed.

Then, what the algorithm does, it iterates over $1\leq i,j\leq d$, defines $f_{i,j}$ up to a given precision and then consider the span of these forms and stops adding them when the span is of dimension $d$. Lastly, it computes a basis of this span space, which must be a basis of the cusp forms space (since the dimensions match). I hope this clears up a bit how the algorithm works.

tscrim: Then what to store as lazy series are the $f_{ij}$ series, which can then be defined by using the (I imagined cached) matrices $[T_n]$. In an ideal world, we would simultaneously compute the $n$-th coefficient of all $f_{ij}$ (so as to not hold onto the redundant $[T_n]$ matrices), but the lazy series code doesn't have a good mechanism for such simultaneous updating. Then we would return a set that forms a basis. Does that make sense?

mantepse: It is still a bit unclear to me what you mean with "simultaneous updating".

tscrim: We have a matrix (mathematically) `[f_{i,j}]_{i,j}` (sorry it is not latex formatting, GH wants to think the `_` should be italics) whose entries are lazy series. They are all based on the same $[T_n]$, so in order to get the $n$-th coefficient of $f_{ij}$ for fixed $i,j$, we compute it for all $f_{ab}$ (for all $a,b$). However, we cannot tell $f_{ab}$ to also update its coefficients (simultaneously).

mantepse: Don't we simply have a series (in $q$) of matrices $[T_n]$, and $f_{i,j}$ is a coefficient?

tscrim: We could, but that still has the information repeated, at the expense of "hiding" (more precisely, the cache) the matrices themselves.

mantepse: Sorry (truly!), I don't understand - in particular: which information is repeated?

Do I understand correctly that the expensive computation which is cached (and which we want to cache) is `self.dual_hecke_matrix(n)`?

Or is there other, additional caching?

(I am guessing that updating the row space may also be expensive.)

tscrim: My understanding is the `dual_hecke_matrix()` is the expensive part (and should be cached in some fashion). However, the exact same information for the $n$-th matrix $[T_n]$ is contained in the $n$-th coefficients of the $\{f_{ij}\}$. Since we cache the coefficients of the lazy series and the matrices, this would be repeated. The issue is that we could not define the $f_{ij}$ without either repeating the computation of $[T_n]$ or storing the $[T_n]$ data. (Well, we could really break into the internals of the streams, but I don't think this is worth it.) The duplication is okay since it shouldn't be big relative to the rest of the computation. The point is we don't have a (good) mechanism for such updates to avoid this, and I am not sure such a feature is beneficial for the complexity.

mantepse: Ah, OK, thank you for this explanation!

Actually, we could avoid the duplication very easily by having an option to `Stream_function` to not cache the results of the function (or create a class `Stream_function_without_cache`.  But I don't think it is worth it.

tscrim: Yea, I don't think so as well. It doesn't even really increase the order of the memory usage ($O(d^2+d) = O(d^2)$).  Trying to avoid this redundancy is just a premature optimization. @DavidAyotte Sorry for a bit of a digression here.

DavidAyotte: Hello Martin and Travis. Sorry for the long answer, but I appreciate your participation in my PR :

> Do I understand correctly that the expensive computation which is cached (and which we want to cache) is self.dual_hecke_matrix(n)?

Indeed, the $n$-th "dual hecke matrix" (which is just the transpose of the $n$-th "hecke matrix") is the complicated part to compute. However, they are already cached by the code as they are quite important, see the method `sage.modular.hecke.module.HeckeModule_free_module.dual_hecke_matrix`.

> (I am guessing that updating the row space may also be expensive.)

Indeed, as updating the row space one need to compute the reduce row echelon form to find the linear independant elements. I'll illustrate that below.

> My understanding is the `dual_hecke_matrix()` is the expensive part (and should be cached in some fashion). However, the exact same information for the n-th matrix [Tn] is contained in the n-th coefficients of the fij. Since we cache the coefficients of the lazy series and the matrices, this would be repeated. The issue is that we could not define the fij without either repeating the computation of [Tn] or storing the [Tn] data. (Well, we could really break into the internals of the streams, but I don't think this is worth it.) The duplication is okay since it shouldn't be big relative to the rest of the computation. The point is we don't have a (good) mechanism for such updates to avoid this, and I am not sure such a feature is beneficial for the complexity.

Yes your understanding is right. Let me also add something: there is an other layer were the code perform linear algebra to find linear independent elements and those might not be the forms $f_{ij}$. I'll illustrate that by computing "by hand" a basis of a certain space of cusp forms of dimension $d = 4$:
```
sage: S = CuspForms(DirichletGroup(5).0, 11)
sage: v = [S.dual_hecke_matrix(n).column(S.dimension() - 1) for n in range(1, 7)]
sage: X = Matrix(v).transpose(); X
[               0                0                0                1        zeta4 + 2         -3*zeta4]
[               0                1         -3*zeta4     -5*zeta4 - 5   -14*zeta4 - 33   -90*zeta4 + 90]
[               0         -3*zeta4              -18   -90*zeta4 + 90   57*zeta4 + 219 -765*zeta4 - 765]
[               1     -5*zeta4 - 5   -90*zeta4 + 90        698*zeta4  143*zeta4 - 684             5124]
```
Each row of the above matrix corresponds to the $q$-expansion of the forms $f_{4, j}$ (here $i = d = 4$ and $1 \leq j \leq 4$). Then, the code compute the row space the the matrix $X$:
```
sage: X.row_space()
Vector space of degree 6 and dimension 4 over Cyclotomic Field of order 4 and degree 2
Basis matrix:
[             1              0              0              0 -23*zeta4 - 26           -120]
[             0              1              0              0   19*zeta4 - 7 -45*zeta4 + 45]
[             0              0              1              0   -7*zeta4 + 6  20*zeta4 + 20]
[             0              0              0              1      zeta4 + 2       -3*zeta4]
```
As we see, in this case the basis matrix is really just the RREF of the matrix X (which one needs to compute to find the linear independent elements). Moreover, each row of the basis matrix corresponds to the $q$-expansion of the basis of cusp forms, as computed by SageMath:
```
sage: S.basis()
[q + (-23*zeta4 - 26)*q^5 + O(q^6),
 q^2 + (19*zeta4 - 7)*q^5 + O(q^6),
 q^3 + (-7*zeta4 + 6)*q^5 + O(q^6),
 q^4 + (zeta4 + 2)*q^5 + O(q^6)]
```

From the above calculations, I was wondering whether it would be a good idea to define the $f_{i,j}$ as lazy series *before* computing the row space of the matrix X. This extra step makes me I'm a bit hesitant in using the lazy series framework with the $f_{i,j}$. Do you think it would be do-able to compute the RREF with the lazy series? For example, given a finite set of lazy power series $h_{i} = \sum_{n\geq 0} a_{n, i} q^n$ for $1\leq i \leq N$, can you find a lazy series basis of the subspace spanned by the $h_i$ ? The coefficients of the basis elements should be in "RREF", meaning that, in the case were the span is generated by four elements, I would want the lazy series to be of these form:
```
[             1              0              0              0               a               ... ]
[             0              1              0              0               b               ... ]
[             0              0              1              0               c               ... ]
[             0              0              0              1               d               ... ]
```

EDIT: the above "matrix" (a matrix with infinite columns) is just an example, but it could be really of any form like:
```
[ 0   0   0   1   0   0   a   ... ]
[ 0   0   0   0   1   0   b   ... ]
[ 0   0   0   0   0   1   c   ... ]
```
(I hope what I wrote is not too confusing)

tscrim: Yes, this is definitely possible. If you compute the $LU$ decomposition of the matrix of coefficients (up to some fixed degree), then you use the $L$ to rewrite your original series into echelon form, then you use the $U$ matrix to group these into RREF form. The resulting RREF lazy series would remember (as part of its construction) the linear algebra needed to get the corresponding coefficient. So it wouldn't need to recompute the RREF every time you want to grow the number of columns (at least, I am guess this is what you are doing); it would just create the next column from this data.

For speed, I am not sure which would be faster in a "typical" case. It might depend on how it is coded and utilized. Of course, you can directly implement the Gaussian elimination on the lazy power series, but I think this would be slower as the series being computed might end up with a complicated internal structure.[^1]

[^1]: We don't have any optimizations for simplifying or realizing generically that, e.g., $(f + g) + h = f + (g + h)$. So if you start chaining together additions and multiplications, even though the result might be simple mathematically, the lazy series code doesn't know currently how to do all such "obvious" simplifications. There's some very simple things that could be done (e.g., linear combinations), but such speed improvements haven't yet been warranted. We've mostly been trying to get features out first  (mainly the functional equations).

PR #37848, opened 2024-04-22
Implement all G(r,p,m) complex reflection groups

tscrim:
We provide an implementation of the whole family of `G(r, p, n)` complex reflection groups by realizing them as a subgroup of `ColoredPermutations`.

### :memo: Checklist


tscrim:
This has a merge conflict with the splitting of the `combinat/all.py`. I will deal with that once this is reviewed (if that is still an issue at that time).

fchapoton:
where does the merge conflict comes from ? which pull request ?

fchapoton:
maybe the choice of the name `ComplexReflectionGroup` is a bit confusing with a category.

tscrim:
It came from the one that split the `all.py` files, which has then been reverted. So there shouldn’t be a conflict anymore.

It’s the same situation with, e.g., `WeylGroup(s)`. This is most naturally called a complex reflection group at this level of generality; I can’t really call it $G(r,p,n)$. Of course, this isn’t all such groups, but it is the unique infinite family. I guess the $p > 1$ case could be called the `ImprimitiveComplexReflectionGroup` (with $p=1$ being the colored permutations). What do you think?

fchapoton:
no, ok, I was just asking. It is indeed reasonable to keep your proposed name, with appropriate documentation

fchapoton:
There is a typo "Klien" instead of Klein.

For the name, I would suggest to keep it shorter. What about "ComplexReflectionG" ?

tscrim:
That makes it seem like I just couldn't type `roup`. How about `ShephardToddFamilyGroup` or `ShephardToddFamilyCRG`?

fchapoton:
right. :smile: 

What about `ShephardToddFamilyGroup` in the class and `ShephardToddFamily` in the catalog ?

tscrim:
Changed. I also removed the global (renamed) import `ComplexReflectionGroup` and added the hook under `groups.misc.ShephardToddFamily`. At some point, I want to change the `ReflectionGroup` to dispatch to these implementations (as well as via `CoxeterGroup` in the real case), but that is for later.

fchapoton (approved):
ok, looks good, Thanks

tscrim:
Thank you!

Issue #37587, opened 2024-03-11
Knot plot() returns questionable non-planar knot diagram (moreover, with overlapping segments)


PratherConid:
### Steps To Reproduce

```
kn = Knots().from_dowker_code([30, 18, 20, 24, 22, 26, 28, 32, 2, 4, 6, 8, 12, 10, 16, 14])
kn.plot()
```

### Expected Behavior

A knot diagram, which should be a planar graph with gaps at each vertex indicating over-crossing/under-crossing.

### Actual Behavior

The `plot` function returns a knot diagram corresponding to a non-planar graph, and with overlapping segments.
![image](https://github.com/sagemath/sage/assets/58621828/88859207-a55c-4a74-b52c-b22d5eda4697)

### Environment

```markdown
- **OS**: Ubuntu 20.04
- **Sage Version**: 10.2
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

tscrim:
@soehms

soehms:
The good news: There is a workaround which produces a correct result in the case of the knot in question:

```
kn = Knots().from_dowker_code([30, 18, 20, 24, 22, 26, 28, 32, 2, 4, 6, 8, 12, 10, 16, 14])
sage: kn.plot(solver='PPL')
```

![kn_ppl](https://github.com/sagemath/sage/assets/47305845/6ce24490-2e1c-492f-98b1-d50a4ba8de02)

The bad news: I still need time to fully understand the implementation of this method. It uses mixed integer linear programming twice. The problem occurs in the first step, which is determining the number of bends required for the segments. The constraints on the MLP at this step are probably not strong enough. I suspect that this is due to taking the exterior region out of consideration:

```
        regions = sorted(self.regions(), key=len)
        regions = regions[:-1]
```


@amitjamadagni, @miguelmarco do you remember the ideas behind this?

DaveWitteMorris:
I think this issue should be closed, because it was fixed by #37851.

tscrim:
Indeed.

PR #37851, opened 2024-04-22
Fix issue 37587 regarding the Link class plot method

soehms:
As suspected in https://github.com/sagemath/sage/issues/37587#issuecomment-2051141073 the issue can be fixed by adding the contraint for the exterior region.

In the differences of this branch just the lines below belong to the corresponding correction of the algorithm:

```
index 9d988d877c..33d0cae051 100644
--- a/src/sage/knots/link.py
+++ b/src/sage/knots/link.py
@@ -3607,36 +3607,55 @@ class Link(SageObject):
         # such that the resulting regions are in fact closed regions
         # with straight angles, and using the minimal number of bends.
         regions = sorted(self.regions(), key=len)
-        regions = regions[:-1]
         edges = list(set(flatten(pd_code)))
...
+            else:
+                # capacity of exterior region, only sink
+                capacity = len(r) + 4
...
         nregions = []
-        for r in regions:
+        for r in regions[:-1]:  # interior regions
             nregion = []
```

All other changes are intended to improve the readability of the code. To this end I change and introduce new variable names to make the origin of the ideas coming from the class [OrthogonalLinkDiagram](https://github.com/3-manifolds/Spherogram/blob/3062478dd69a52a16e55b967dd46dfe940af9ea7/spherogram_src/links/orthogonal.py#L118) of [Spherogram](https://github.com/3-manifolds/Spherogram) more explicit. Furthermore I make some codestyle fixes and change the sign of the capacity such that the sink appears to be positive and thus according to the usage in Spherogram and the literature cited there.

With this branch the knot reported in the issue is plotted like this:

```
sage: kn = Knots().from_dowker_code([30, 18, 20, 24, 22, 26, 28, 32, 2, 4, 6, 8, 12, 10, 16, 14])
sage: arcs = sorted(kn.arcs())
sage: cols = {tuple(arcs[i]): i for i in range(len(arcs))}
sage: kn.plot(color=cols)
```

### New plot

kn_new_cols

To make comparison with other results more convenient I used different colours for the arcs. The corresponding result for `PPL` is this:

```
sage: kn.plot(color=cols, solver='PPL')
```

### New plot with `PPL`

kn_new_cols_ppl

Without this branch the `PPL` version with colours has been this:

### Old plot with `PPL`

kn_cols_ppl

The coloured wrong result has been this:

### Old plot

kn_cols

Note, that this branch also changes other plot results in the documentation. For example:

Previously:

### Old documentation

grafik

With this branch:

### New documentation

grafik

### :memo: Checklist

### DIFF
--- a/src/sage/knots/link.py
+++ b/src/sage/knots/link.py
@@ -3607,36 +3607,55 @@ class Link(SageObject):
         # such that the resulting regions are in fact closed regions
         # with straight angles, and using the minimal number of bends.
         regions = sorted(self.regions(), key=len)
-        regions = regions[:-1]
         edges = list(set(flatten(pd_code)))
         edges.sort()
         MLP = MixedIntegerLinearProgram(maximization=False, solver=solver)
         # v will be the list of variables in the MLP problem. There will be
-        # two variables for each edge: number of right bendings and number of
-        # left bendings (at the end, since we are minimizing the total, only one
-        # of each will be nonzero
+        # two variables for each edge: number of bendings concerning the
+        # left-hand-side region (source, even index if positive oriented)
+        # and number of bendings concerning the right-hand-side region
+        # (sink, odd index if positive oriented). At the end, since
+        # we are minimizing the total, only one of each will be nonzero.
         v = MLP.new_variable(nonnegative=True, integer=True)
+
+        def flow_from_source(e):
+            r"""
+            Return the flow variable from the source.
+            """
+            if e > 0:
+                return v[2*edges.index(e)]
+            else:
+                return v[2*edges.index(-e)+1]
+
+        def flow_to_sink(e):
+            r"""
+            Return the flow variable to the sink.
+            """
+            return flow_from_source(-e)
+
         # one condition for each region
-        for i in range(len(regions)):
-            cond = 0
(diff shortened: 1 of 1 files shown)


miguelmarco:
Sorry for not seeing this before. I will try to review it in the next days.

As you mention, we first use integer programming to compute the number of bendings of the segments. I am surprised that a lack of checking the outer region has only caused problems in very few cases.

miguelmarco:
Looks good.

As a side comment for the future, I think it would be worth considering an approach that is based on [this](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/schnyder.html).

soehms:
> Looks good.
> 
> As a side comment for the future, I think it would be worth considering an approach that is based on [this](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/schnyder.html).

Sounds like a good idea!

soehms:
> One minor comment though: I don't understand the rationale behind the "source" and "sink" naming that you use in the comments.

I made another attempt in the current commit. I hope I have found a better wording for this.

soehms:
Thanks for the review, @miguelmarco!

PR #37853, opened 2024-04-22
Work around inconsistent iteration in `multi_polynomial_sequence.py`

### LINKED ISSUE #37837: PolynomialSequence_gf2.coefficients_monomials is broken
### Steps To Reproduce

Run

```python
for F in (GF(3), GF(2)):
    R = PolynomialRing(F, ['a', 'b', 'c'])
    a, b, c = R.gens()
    f = a + b + c
    M, monos = Sequence([f]).coefficients_monomials()
    print(F)
    print(M)
    print(monos)
```

### Expected Behavior

Print

```
[1 1 1]
(a, b, c)
```

for both `GF(3)` and `GF(2)`

### Actual Behavior

`GF(3)` works correctly
`GF(2)` result in this error:

```
KeyError                                  Traceback (most recent call last)
File /usr/lib/python3.11/site-packages/sage/rings/polynomial/multi_polynomial_sequence.py:1746, in PolynomialSequence_gf2.coefficients_monomials(self, order, sparse)
   1745 try:
-> 1746     A[x, y[m]] = one
   1747 except KeyError:

KeyError: (1, a)

Cell In[209], line 5
      3 a, b, c = R.gens()
      4 f = a + b + c
----> 5 M, monos = Sequence([f]).coefficients_monomials()
      6 print(F)
      7 print(M)

File /usr/lib/python3.11/site-packages/sage/rings/polynomial/multi_polynomial_sequence.py:1748, in PolynomialSequence_gf2.coefficients_monomials(self, order, sparse)
   1746             A[x, y[m]] = one
   1747         except KeyError:
-> 1748             raise ValueError("order argument does not contain all monomials")
   1749 return A, vector(v)

ValueError: order argument does not contain all monomials
```

### Additional Information

I think it comes from [here](https://github.com/sagemath/sage/blob/15fcf52fbd6a72f0d0f226c02fe1a74b21bbfaf5/src/sage/rings/polynomial/multi_polynomial_sequence.py#L1745), iterating directly on a `GF(2)` polynomial is still a pair of `(coefficient, monomial)` instead of just `monomial`.

### Environment

```markdown
- **OS**: Arch Linux on WSL2
- **Sage Version**: 10.3
```

### Checklist

S17A05:
When iterating over a `pbori.BooleanPolynomial`, the iterator generates just the monomials, whereas iterating over a `MPolynomial_libsingular` defined over `GF(2)` yields a tuple of the monomial together with its coefficient $1$. This causes issues in the method `.coefficients_monomials()` of `PolynomialSequence_gf2`; we fix said issue by checking ~~whether the iteration yields tuples, in which case we simply replace the tuple by its second element (the actual monomial) in each iteration.~~ which type of polynomial ring we are working in before iterating.

Fixes #37837 (but there might be a more elegant solution).

### DIFF
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -1752,13 +1752,13 @@ class PolynomialSequence_gf2(PolynomialSequence_generic):
             else:
                 raise ValueError("order argument can only accept list or tuple")
 
-        PolyRing = self.ring()
-        BaseRing = PolyRing.base_ring()
+        R = self.ring()
+        K = R.base_ring()
         y = dict(zip(v, range(len(v))))  # construct dictionary for fast lookups
-        A = Matrix(BaseRing, len(self), len(v), sparse=sparse)
+        A = Matrix(K, len(self), len(v), sparse=sparse)
 
-        if isinstance(PolyRing, BooleanPolynomialRing_base):
-            one = BaseRing.one()
+        if isinstance(R, BooleanPolynomialRing_base):
+            one = K.one()
             for x, poly in enumerate(self):
                 for m in poly:
                     try:
--- a/src/sage/rings/polynomial/multi_polynomial_sequence.py
+++ b/src/sage/rings/polynomial/multi_polynomial_sequence.py
@@ -1741,6 +1741,8 @@ class PolynomialSequence_gf2(PolynomialSequence_generic):
         """
         from sage.modules.free_module_element import vector
         from sage.matrix.constructor import Matrix
+        from sage.rings.polynomial.multi_polynomial_ring_base import \
+            BooleanPolynomialRing_base
 
         if order is None:
             v = sorted(self.monomials(), reverse=True)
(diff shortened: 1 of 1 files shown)


src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -1743,6 +1756,8 @@ def coefficients_monomials(self, order=None, sparse=True):
         A = Matrix(R, len(self), len(v), sparse=sparse)
         for x, poly in enumerate(self):
             for m in poly:
+                if isinstance(m, tuple):
+                    m = m[1]
                 try:
                     A[x, y[m]] = one
                 except KeyError:
```
mantepse: I'd rather use `isinstance(self.ring(), BooleanPolynomialRing_base)`, and perhaps (don't know) do the check outside of the loops.

S17A05: I've changed it up a bit to check the type of polynomial ring outside of the loops.

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -1737,16 +1752,27 @@ def coefficients_monomials(self, order=None, sparse=True):
             else:
                 raise ValueError("order argument can only accept list or tuple")
 
-        R = self.ring().base_ring()
-        one = R.one()
+        PolyRing = self.ring()
+        BaseRing = PolyRing.base_ring()
```
mantepse: This is extremely minor: I'd like to ask you to stick to the conventions in this module:
```
        R = self.ring()
        K = R.base_ring()
```
Otherwise perfect!

mantepse (approved):
Thank you!

mantepse:
In case you are interested in the boolean polynomial ring, maybe you want to resurrect #26929.

Issue #29169, opened 2020-02-09
Maximal and critical angles for cones


orlitzky:
<div id="comment:0"></div>

My paper with two algorithms for finding the maximal/critical angles between polyhedral convex cones was accepted:

  http://www.optimization-online.org/DB_HTML/2019/01/7048.html

I've already written these algorithms in sage, so it would be nice to add them to the library. I just need to clean up some of the documentation.

Depends on #30605
Depends on #30654
Depends on #32877

CC:  @jplab @novoselt @tscrim @fchapoton

Component: **geometry**

Author: **Michael Orlitzky**

Branch/Commit: **[u/mjo/ticket/29169](https://github.com/sagemath/sagetrac-mirror/tree/u/mjo/ticket/29169) @ [`fb55e7f`](https://github.com/sagemath/sagetrac-mirror/commit/fb55e7fc7e04ebc904f1b4c16f7716533be3e36d)**

Reviewer: **Matthias Koeppe**

_Issue created by migration from https://trac.sagemath.org/ticket/29169_

orlitzky:
This is now https://github.com/sagemath/sage/pull/37854

PR #37854, opened 2024-04-23
Maximal angles between convex cones

orlitzky:
My colleagues are writing about this stuff again. Let's migrate https://github.com/sagemath/sage/issues/29169 from the old Trac branch.

Relative to _develop_:
* Add a new reference
* Add a new module `src.sage.geometry.cone_critical_angles` for the implementation
* Add a `max_angle()` method to the class in `src.sage.geometry.cone` for the user interface

Relative to the old trac branch:

* Update the algorithm to take https://www.heldermann.de/JCA/JCA31/JCA311/jca31015.htm into consideration
* Add that paper as a reference
* Drop the more-general critical angles computation. It isn't nearly as nice as the maximal angle method.

TODO:

* ~~Revert commit 95e57636c9 after checking the docs~~


src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,978 @@
...
+
+
+def solve_gevp_zero(M, I, J):
+    r"""
+    Solve the generalized eigenvalue problem in Theorem 3 for a zero
+    eigenvalue using Propositions 3 and 4 [Or2020]_.
+
+    INPUT:
+
+    - ``M`` -- the matrix whose `(i,j)^{th}` entry is the inner
```
mkoeppe:
```suggestion
    - ``M`` -- the matrix whose `(i,j)`-th entry is the inner
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,978 @@
...
+    return K
+
+
+def gevp_licis(G):
+    r"""
+    Return all nonempty subsets of indices for the columns of
+    ``G`` that correspond to linearly-independent sets (of columns of
+    ``G``).
+
+    Mnemonic: linearly-independent column-index subsets (LICIS).
```
mkoeppe:
```suggestion
    ``G`` that correspond to linearly independent sets (of columns of
    ``G``).

    Mnemonic: linearly independent column-index subsets (LICIS).
```
(throughout)

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,978 @@
...
+        sage: I = [0, 1]
+        sage: J = [0, 1]
+        sage: all( v in [-1,1] and m == 2
+        ....:      for (v,_,_,m) in _solve_gevp_naive(GG,HH,M,I,J) )
+        True
+    """
+    A = matrix.block([ [ ZZ(0),              M[I,J] ],
+                       [ M.transpose()[J,I], ZZ(0)  ] ])
+    B = matrix.block([ [ GG[I,I],    ZZ(0) ],
+                       [ ZZ(0), HH[J,J]  ] ])
```
mkoeppe: align blocks?

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,978 @@
...
+    TESTS:
+
+    This function should return the same solutions (with zero included,
+    of course) as the naive implementation even for random cones::
+
+        sage: # long time
+        sage: from itertools import chain
+        sage: from sage.geometry.cone_critical_angles import (
+        ....:   _normalize_gevp_solution,
+        ....:    _random_admissible_cone,
```
mkoeppe:
```suggestion
        ....:   _random_admissible_cone,
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,978 @@
...
+                        eta.set_immutable()
+                        yield (l, xi, eta, m)
+
+
+def compute_gevp_M(gs, hs):
+    r"""
+    Compute the matrix `M` whose `(i,j)^{th}` entry is the
+    inner product of ``gs[i]`` and ``hs[j]``.
+
+    This is the "generalized gram matrix" appearing in Proposition 6
```
mkoeppe:
```suggestion
    This is the "generalized Gram matrix" appearing in Proposition 6
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,978 @@
...
+        sage: gs = [g.change_ring(AA).normalized() for g in P]
+        sage: G = matrix.column(gs)
+        sage: GG = G.transpose()*G
+        sage: hs = [h.change_ring(AA).normalized() for h in Q]
+        sage: H = matrix.column(hs)
+        sage: HH = H.transpose()*H
+        sage: M = G.transpose()*H
+        sage: G_index_sets = list(gevp_licis(G))
+        sage: H_index_sets = list(gevp_licis(H))
+        sage: all( set( _normalize_gevp_solution(s) for s in
```
mkoeppe: Beware the style police

src/sage/geometry/cone_critical_angles.py
orlitzky: Done.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done, but are you sure this works? I started with `` `(i,j)`th `` and later changed it because it was being mis-parsed.

src/sage/geometry/cone_critical_angles.py
orlitzky: I'm not sure what you meant so I've reindented these two examples and have probably made them worse.

src/sage/geometry/cone_critical_angles.py
kwankyu: The hyphen does the trick.

orlitzky: Ok, thanks for the confirmation.

src/sage/geometry/cone.py
```diff
@@ -6194,6 +6194,205 @@ def Z_operators_gens(self):
...
+        Return the maximal angle between ``self`` and ``other``.
+
+        The maximal angle between two closed convex cones is the
+        unique largest angle formed by any two unit-norm vectors in
+        those cones. In pathological cases, this computation can fail.
+
+        If it fails when ``exact`` is ``True`` and if each of your
+        cones :meth:`is_strictly_convex`, then a second attempt will
+        be made using inexact arithmetic. (This sometimes avoids the
+        problem [Or2024]_). If the computation fails when your cones
```
tscrim:
```suggestion
        If it fails when ``exact`` is ``True`` and if each of the
        cones :meth:`is_strictly_convex`, then a second attempt will
        be made using inexact arithmetic. (This sometimes avoids the
        problem noted in [Or2024]_). If the computation fails when the cones
```

src/sage/geometry/cone.py
```diff
@@ -6194,6 +6194,205 @@ def Z_operators_gens(self):
...
+        - ``epsilon`` -- (default: ``0``) the tolerance to use when
+          making comparisons
+
+        .. WARNING::
+
+          Using inexact arithmetic (``exact=False``) is faster, but
+          this computation is only known to be stable when both of
+          your cones are strictly convex (or when one of them is the
+          entire space, but the maximal angle is obviously `\pi` in
+          that case).
```
tscrim:
```suggestion
            Using inexact arithmetic (``exact=False``) is faster, but
            this computation is only known to be stable when both of
            your cones are strictly convex (or when one of them is the
            entire space, but the maximal angle is obviously `\pi` in
            that case).
```

src/sage/geometry/cone.py
```diff
@@ -6194,6 +6194,205 @@ def Z_operators_gens(self):
...
+
+        OUTPUT:
+
+        A triple `\left( \theta_{\text{max}}, u, v \right)`
+        containing,
```
tscrim:
```suggestion
        A triple `\left( \theta_{\max}, u, v \right)`
        containing:
```

src/sage/geometry/cone.py
```diff
@@ -6194,6 +6194,205 @@ def Z_operators_gens(self):
...
+                                 "must agree")
+            if other.is_trivial():
+                raise ValueError("other cone cannot be trivial")
+
+        # Use the names from the paper to keep things understandable.
+        P = self
+        Q = other
+
+        from sage.geometry.cone_critical_angles import max_angle
+        return max_angle(P, Q, exact, epsilon)
```
tscrim:
```suggestion
        from sage.geometry.cone_critical_angles import max_angle
        return max_angle(self, other, exact, epsilon)
```
I don't think that helps. Since the main algorithm is in the `max_angle` function, that is where matching the paper is useful.

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
+r"""
+Find maximal angles between polyhedral convex cones
+
+Finding the maximal (or equivalently, the minimal) angle between two
+polyhedral convex cones is a hard nonconvex optimization problem. The
+problem for a single cone was introduced in [IS2005]_, and was later
+extended in [SS2016]_ to two cones as a generalization of the
+principal angle between two vector subspaces.
+
+Seeger and Sossa proposed an algorithm to find maximal angles, and
```
tscrim:
```suggestion
Seeger and Sossa proposed an algorithm in [SS2016]_ to find maximal angles, and
```
Good to be explicit IMO.

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+[Or2024]_ is taken into consideration to avoid pathological cases.
+
+This module is internal to SageMath; the interface presented to users
+consists of a public method,
+:meth:`sage.geometry.cone.ConvexRationalPolyhedralCone.max_angle` for
+polyhedral convex cones. Even though all of the functions in this
+module are internal, some are more internal than others. There are a
+few functions that are used only in doctests, and not by any code that
+an end-user would run. Breaking somewhat with tradition, only those
+methods have been prefixed with an underscore.
```
tscrim, line 31: I don't understand the last sentence. Although there are benefits to exposing the functions to the public doc, even for technical functions. They still wouldn't be accessible from Sage's API. So I would probably not make them hidden in general (unless perhaps they are just a trivial helper function).

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        sage: G = matrix.column(K.rays())
+        sage: all( len(s) == K.rays(s).dimension() for s in gevp_licis(G) )
+        True
+    """
+    from sage.matroids.linear_matroid import LinearMatroid
+
+    # There's a fast implementation of this for matroids, but we need
+    # to drop the empty set from its output and convert the rest to
+    # lists that are all sorted in the same order.
+    return map(sorted, filter(bool,LinearMatroid(G).independent_sets()))
```
tscrim:
```suggestion
    return map(sorted, filter(bool, LinearMatroid(G).independent_sets()))
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        sage: J = [0, 1]
+        sage: all( v in [-1,1] and m == 2
+        ....:      for (v,_,_,m) in _solve_gevp_naive(GG,HH,M,I,J) )
+        True
+    """
+    A = matrix.block([ [ ZZ(0),              M[I,J] ],
+                       [ M.transpose()[J,I], ZZ(0)  ] ])
+    B = matrix.block([ [ GG[I,I], ZZ(0)   ],
+                       [ ZZ(0),   HH[J,J] ] ])
+    M = B.inverse() * A
```
tscrim:
```suggestion
    A = matrix.block(ZZ, [[ ZZ.zero(), M[I,J] ],
                       [M[I,J].transpose(), ZZ.zero()]])
    Binv = matrix.block_diagonal(ZZ, [GG[I,I].inverse(), HH[J,J].inverse()])
    M = Binv * A
```
These are mostly mirco-optimizations with more direct cache usage. It is better to transpose and invert the smaller matrices too. PEP8 also doesn't like such alignments... (yes, I think there should be exceptions too...)

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+            xi = z[0:len(I)]
+            xi.set_immutable()
+            eta = z[len(I):]
+            eta.set_immutable()
+            yield (evalue, xi, eta, multiplicity)
+
+
+def solve_gevp_zero(M, I, J):
+    r"""
+    Solve the generalized eigenvalue problem in Theorem 3 for a zero
```
tscrim:
```suggestion
    Solve the generalized eigenvalue problem in Theorem 3 [Or2020]_ for a zero
```
Otherwise, which Theorem 3?

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    INPUT:
+
+    - ``M`` -- the matrix whose `(i,j)`-th entry is the inner product
+      of `g_{i}` and `h_{j}` as in Proposition 6 [Or2020]_
+
+    - ``I`` -- a linearly independent column-index set for the matrix
+      `G` that appears in Theorem 3
+
+    - ``J`` -- a linearly independent column-index set for the matrix
+      `H` that appears in Theorem 3
```
tscrim:
```suggestion
    - ``I`` -- a linearly independent column-index set for the matrix
      `G` that appears in Theorem 3 [Or2020]_

    - ``J`` -- a linearly independent column-index set for the matrix
      `H` that appears in Theorem 3 [Or2020]_
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        sage: I = [0]
+        sage: J = [1]
+        sage: list(solve_gevp_zero(GG,I,J))
+        [(0, (1), (0), 2), (0, (0), (1), 2)]
+    """
+    # A Cartesian product would be more appropriate here, but Sage
+    # isn't smart enough to figure out a basis for the product. So,
+    # we use the direct sum and then chop it up.
+    xi_space = (M.transpose()[J,I]).right_kernel()
+    eta_space = (M[I,J]).right_kernel()
```
tscrim:
```suggestion
    MIJ = M[I,J]
    xi_space = MIJ.left_kernel()
    eta_space = MIJ.right_kernel()
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+
+    INPUT:
+
+    - ``GG`` -- the matrix whose `(i,j)`-th entry is the inner product
+      of `g_{i}` and `g_{j}`, which are in turn the `i`-th and `j`-th
+      columns of the matrix `G` in Theorem 3
+
+    - ``HH`` -- the matrix whose `(i,j)`-th entry is the inner product
+      of `h_{i}` and `h_{j}`, which are in turn the `i`-th and `j`-th
+      columns of the matrix `H` in Theorem 3
```
tscrim:
```suggestion
      columns of the matrix `G` in Theorem 3 [Or2020]_

    - ``HH`` -- the matrix whose `(i,j)`-th entry is the inner product
      of `h_{i}` and `h_{j}`, which are in turn the `i`-th and `j`-th
      columns of the matrix `H` in Theorem 3 [Or2020]_
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+
+    - ``M`` -- the matrix whose `(i,j)`-th entry is the inner product
+      of `g_{i}` and `h_{j}` as in Proposition 6 in [Or2020]_
+
+    - ``I`` -- a linearly independent column-index set for the matrix
+      `G` that appears in Theorem 3
+
+    - ``J`` -- a linearly independent column-index set for the matrix
+      `H` that appears in Theorem 3
```
tscrim:
```suggestion
      `G` that appears in Theorem 3 [Or2020]_

    - ``J`` -- a linearly independent column-index set for the matrix
      `H` that appears in Theorem 3 [Or2020]_
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    - ``xi`` is the first (length ``len(I)``) component of an
+      eigenvector associated with ``eigenvalue``
+
+    - ``eta`` is the second (length ``len(J)``) component of an
+      eigenvector associated with ``eigenvalue``
+
+    - ``multiplicity`` is the dimension of the eigenspace associated
+      with ``eigenvalue``
+
+    Note that we don't return a basis for each eigenspace along with
```
tscrim:
```suggestion
    Note that we do not return a basis for each eigenspace along with
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        ....:     _normalize_gevp_solution(s)
+        ....:     for s in
+        ....:     _solve_gevp_naive(GG,HH,M,I,J)
+        ....:   )
+        ....:   for I in G_index_sets
+        ....:   for J in H_index_sets
+        ....: )
+        True
+
+    According to Proposition 7, the only eigenvalues that arise when
```
tscrim:
```suggestion
    According to Proposition 7 [Or2020]_, the only eigenvalues that arise when
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    """
+    if len(J) < len(I):
+        # We can always opt to solve the smaller problem. Reading the
+        # first three assignments below, you should be able to
+        # convince yourself that switching GG <-> HH, I <-> J, and
+        # transposing M does in fact switch from the "xi problem" to
+        # the "eta problem."
+        yield from ( (l, xi, eta, m)
+                     for (l, eta, xi, m)
+                     in solve_gevp_nonzero(HH,GG,M.transpose(),J,I) )
```
tscrim:
```suggestion
        yield from ((l, xi, eta, m)
                     for (l, eta, xi, m)
                     in solve_gevp_nonzero(HH, GG, M.transpose(), J, I))
```
PEP8

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    else:
+        G_I_pinv_H_J = GG[I,I].inverse_positive_definite()*M[I,J]
+        H_J_pinv_G_I = HH[J,J].inverse_positive_definite()*M.transpose()[J,I]
```
tscrim:
```suggestion
        MIJ = M[I,J]
        G_I_pinv_H_J = GG[I,I].inverse_positive_definite() * MIJ
        H_J_pinv_G_I = HH[J,J].inverse_positive_definite() * MIJ.transpose()
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        L = (G_I_pinv_H_J * H_J_pinv_G_I)
+
+        for (sigma, xis, m) in L.eigenvectors_right():
+            for xi in xis:
+                if sigma > 0:
+                    for l in [-sigma.sqrt(), sigma.sqrt()]:
+                        eta = ~l * H_J_pinv_G_I*xi
+                        eta.set_immutable()
+                        yield (l, xi, eta, m)
```
tscrim:
```suggestion
        for (sigma, xis, m) in L.eigenvectors_right():
            if sigma > 0:
                sigma_sqrt = sigma.sqrt()
                inv_sqrt = ~sigma_sqrt
                sqrts = [-sigma_sqrt, sigma_sqrt]
                inv_sqrts = [-inv_sqrt, inv_sqrt]
                for xi in xis:
                    for l, li in zip(sqrts, inv_sqrts):
                        eta = li * H_J_pinv_G_I * xi
                        eta.set_immutable()
                        yield (l, xi, eta, m)
```
Trying to avoid repeating computations.

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        sage: H_index_sets = list(gevp_licis(H))
+        sage: all( _test_indexing(I,J) for I in G_index_sets
+        ....:                          for J in H_index_sets )
+        True
+    """
+    min_u = gs[0]
+    min_v = hs[0]
+    min_ip = min_u.inner_product(min_v)
+
+    M = []
```
tscrim, line 696:
```suggestion
    M = [None] * len(gs)
```
It is faster to initialize the list to the correct size if known.

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    for i in range(len(gs)):
+        M_i = []
+        for j in range(len(hs)):
+            M_i.append(gs[i].inner_product(hs[j]))
+            if (M_i[j] < min_ip):
+                min_ip = M_i[j]
+                min_u = gs[i]
+                min_v = hs[j]
+        M.append(M_i)
```
tscrim, line 706:
```suggestion
        M[i] = M_i
```

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+
+    M = []
+    for i in range(len(gs)):
+        M_i = []
+        for j in range(len(hs)):
+            M_i.append(gs[i].inner_product(hs[j]))
+            if (M_i[j] < min_ip):
+                min_ip = M_i[j]
+                min_u = gs[i]
+                min_v = hs[j]
```
tscrim:
```suggestion
        M_i = [None] * len(hs)
        for v in hs:
            val = M_i.append(gs[i].inner_product(v))
            if val < min_ip:
                min_ip = val
                min_u = gs[i]
                min_v = v
```
Similar such changes.

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    v = H_J*eta_hat
+    rhs = v - cos_theta*G_I*xi_hat
+
+    if any( x < -epsilon for x in G_I_c_T*rhs ):
+        return infeasible_result
+
+    u = G_I*xi_hat
+    rhs = u - cos_theta*H_J*eta_hat
+    if any( x < -epsilon for x in H_J_c_T*rhs):
+        return infeasible_result
```
tscrim:
```suggestion
    v = H_J * eta_hat
    rhs = v - cos_theta * G_I * xi_hat

    if any(x < -epsilon for x in G_I_c_T * rhs):
        return infeasible_result

    u = G_I * xi_hat
    rhs = u - cos_theta * H_J * eta_hat
    if any(x < -epsilon for x in H_J_c_T * rhs):
        return infeasible_result
```
PEP8

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+
+    if Q_is_P:
+        GG = M
+        H = G
+        HH = M
+        H_index_sets = G_index_sets
+    else:
+        GG = G.transpose()*G
+        H = matrix.column(hs)
+        HH = H.transpose()*H
```
tscrim:
```suggestion
        GG = G.transpose() * G
        H = matrix.column(hs)
        HH = H.transpose() * H
```
PEP8

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+    # Keep track of the (cos-theta, xi, eta, multiplicity) tuples with
+    # multiplicity > 1. These are only a problem if they could
+    # potentially be maximal. Therefore, we want to inspect them AFTER
+    # we've checked all of the multiplicity=1 angles and found the
+    # largest. This allows us to ignore most of the problematic
+    # eigenspaces, independent of the order in which we run through I,J.
+    big_eigenspaces = []
+
+    for I in G_index_sets:
+        G_I = G[range(n),I]
```
tscrim: Wouldn't `G.matrix_from_columns(I)` be faster here? Similar for other places below.

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+        I_complement = [i for i in range(P.nrays()) if i not in I]
+        G_I_c_T = G[range(n),I_complement].transpose()
+
+        for J in H_index_sets:
+            J_complement = [j for j in range(Q.nrays()) if j not in J]
+            H_J = H[range(n),J]
+            H_J_c_T = H[range(n),J_complement].transpose()
+
+            for (cos_theta,xi,eta,mult) in solve_gevp_nonzero(GG,HH,M,I,J):
```
tscrim:
```suggestion
            for (cos_theta,xi,eta,mult) in solve_gevp_nonzero(GG, HH, M, I, J):
```
PEP8

src/sage/geometry/cone_critical_angles.py
```diff
@@ -0,0 +1,998 @@
...
+                                                             epsilon)
+
+                if is_feasible:
+                    min_ip = cos_theta
+                    min_u = u
+                    min_v = v
+                elif mult > 1:
+                    # Save this for later. The eigenvalue cos_theta
+                    # might be so big that we can ignore it.
+                    big_eigenspaces.append( (cos_theta, xi, eta, mult) )
```
tscrim:
```suggestion
                    big_eigenspaces.append((cos_theta, xi, eta, mult))
```
PEP8

tscrim (changes requested):
Hopefully these changes are all speed improvements. I didn't check carefully that all statements are equivalent, but hopefully the intent is clear. Let me know if you have questions.

src/sage/geometry/cone_critical_angles.py
orlitzky: Fixed

src/sage/geometry/cone_critical_angles.py
orlitzky: Fixed

src/sage/geometry/cone_critical_angles.py
orlitzky: No problem, done

src/sage/geometry/cone.py
orlitzky: In hindsight I agree. Fixed!

src/sage/geometry/cone.py
orlitzky: Done

src/sage/geometry/cone.py
orlitzky: Done, and changed "your cones" to "the cones" here too.

src/sage/geometry/cone_critical_angles.py
orlitzky: I fixed some of these, but PEP8 actually recommends writing products without spaces when addition/subtraction (or any lower-precedence operation) are involved: https://peps.python.org/pep-0008/#other-recommendations

As a result, `v - cos_theta*G_I*xi_hat` is probably correct -- it's one of the examples they give. The expression,

```python
x < -epsilon for x in H_J_c_T*rhs
```

is more ambiguous, but in theory you could be multiplying the result of the list comprehension by `rhs`. Writing `H_J_c_T*rhs` with no spaces helps indicate (to me, anyway) that it's one quantity and binds stronger than the `for foo in bar`. I'm still happy to change it if you care, though.

src/sage/geometry/cone_critical_angles.py
orlitzky: I'm happy to change it either way, but does PEP8 actually suggest putting spaces after the commas? I've been looking and don't see it.

src/sage/geometry/cone.py
orlitzky: When the subscript is a word and not a variable I prefer to see it in text face, but I've no idea if I'm being (in)consistent and don't want to waste time checking, so I've gone ahead and changed this.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: I've changed `ZZ(0)` to `ZZ.zero()` since that is a good idea in any case. But this code is inside the `_solve_gevp_naive` function, whose only purpose is to test the two non-naive GEVP functions. As the docs say,

> This is used only for testing, to ensure that the smart way of solving the generalized eigenvalue problem ... returns the same answers as the dumb way.

For that reason I have avoided using any mathematical tricks (no matter how obvious) to simplify the problem. Is it OK to leave the block matrices? I can trim some of the spaces to conform to PEP8.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
mkoeppe: > As a result, `v - cos_theta*G_I*xi_hat` is probably correct

I strongly agree. It's correct per PEP 8 and much more readable than the fully "spaced out" version.

> Writing `H_J_c_T*rhs` with no spaces

This one is trickier to justify per PEP 8, but I still support it; not with the explanation that you give, but rather because these terms (products) appear right next to polynomial expression where the monomials are written in this way without spaces. Keeping the same style for the same terms aids readability.

tscrim: At the risk of continuing to bikeshed, the `H_J_c_T * rhs` needs the spacing by PEP8 as the `*` is the highest priority operator. It's not a consistency thing.

I  have no objections to not having the spaces for the other as it is correct. (I think it is a bit of an overreach to say PEP8 prefers it, much less recommends it.) I probably wasn't paying close enough attention here. (Although I do find it easier to read with all the underscores in the variable names, but its not important.)

src/sage/geometry/cone_critical_angles.py
tscrim: I'm surprised this is not in there. I could have sworn it was a Python style coding convention. It is definitely a very widely used principle that commas should generally be followed by a whitespace. This might be a Sage convention...

src/sage/geometry/cone_critical_angles.py
tscrim: It's not really a trick, it's simply realizing it is a block diagonal matrix. Anyways, if you prefer it, it is okay.

src/sage/geometry/cone.py
tscrim: Note that I had `\max`, not `max`. So it will display as text. I agree with you that text formatting is better.

orlitzky: Ah, I missed the `\` in the diff. Thanks, I'll put it back.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done

src/sage/geometry/cone_critical_angles.py
orlitzky: OK, I made this change in a few places.

src/sage/geometry/cone_critical_angles.py
orlitzky: Ok, thanks. I've removed the extra spacing and reindented:

```python
    A = matrix.block([
        [ZZ.zero(), M[I,J]],
        [M.transpose()[J,I], ZZ.zero()]
    ])
    B = matrix.block([
        [GG[I,I], ZZ.zero()],
        [ZZ.zero(), HH[J,J]]
    ])
```

PEP8 is okay with that.

src/sage/geometry/cone_critical_angles.py
orlitzky: Do you think it's worth pulling out the `zip` as well? Something like,

```python
pm_sqrt_inv_pairs = [
    (-sigma_sqrt, -inv_sqrt),
    (sigma_sqrt, inv_sqrt),
]
for xi in xis:
    for l, li in pm_sqrt_inv_pairs:
```

I can't say for sure now that `zip` returns an iterator.

src/sage/geometry/cone_critical_angles.py
orlitzky: Typically a module will have functions that are part of the public API, and internal (underscore) methods that are not. But in this case, the entire module should be considered internal:

> This module is internal to SageMath

Moreover it contains an extra class of functions beyond just the public and internal ones, namely functions used only for testing. Those testing-only functions are certainly internal, so I've given them an underscore prefix. But now what should we do with the non-testing functions, all of which are *also* internal?

If we give them all underscores, then there's no way to differentiate between the testing-only functions and the ones that are called in real code by way of `max_angle()`. So, in this case, I've left the underscore off the internal functions that are used in real code.

Nothing is truly internal in python anyway so this seemed like the least-bad option to me. I'm open to suggestions though.

src/sage/geometry/cone_critical_angles.py
orlitzky: Avoiding `zip` looks like it saves about a microsecond, but at that scale it's impossible to say for sure.

orlitzky: Done, thanks

src/sage/geometry/cone_critical_angles.py
orlitzky: Yep. I didn't know that method was there.

src/sage/geometry/cone_critical_angles.py
orlitzky: Done, thanks

src/sage/geometry/cone_critical_angles.py
orlitzky: Done, thanks

src/sage/geometry/cone_critical_angles.py
tscrim: Pulling out the zip (as a zip) wouldn't work because of the iterator, but you're right that it should be better to just do it manually since it is a very small list. In either case, the biggest thing is making sure the `if sigma > 0:` is outside the `for` loops.

src/sage/geometry/cone_critical_angles.py
orlitzky: It doesn't seem to be faster in this case?

src/sage/geometry/cone_critical_angles.py
orlitzky: list append doesn't return a value... or am I missing something? I could still assign `M_i[j]` to a variable to avoid re-indexing but surely by itself that's overkill.

src/sage/geometry/cone_critical_angles.py
tscrim: Okay, thanks. Now I understand.

I would probably lean towards making them all visible so they appear in the documentation. That way someone looking at this module's doc will know they are there and could potentially use them as part of their own testing. It seems reasonable to expose them in the documentation (especially considering how much effort you spent documenting these hidden functions). Anyways, it's okay to leave it the way it is too. I don't have a very strong opinion on this.

src/sage/geometry/cone_critical_angles.py
tscrim: Ah, sorry, that was a mistake. It should have been
```python
for j, v in enumerate(hs):
    val = gs[i].inner_product(v)
    M_i[j] = val

File: src/sage/geometry/cone_critical_angles.py (hunk 25 above)
Comment (tscrim) [reply src/sage/geometry/cone_critical_angles.py:696]: Definitely a micro-optimization, and the benefit depends on the size of the list:
```python
sage: def test1(n):
....:     L = [None]*n
....:     for i in range(n):
....:         L[i] = i
sage: def test2(n):
....:     L = []
....:     for i in range(n):
....:         L.append(i)
sage: %timeit test1(5)
516 ns ± 14.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
sage: %timeit test2(5)
426 ns ± 9.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
sage: %timeit test1(20)
923 ns ± 25.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
sage: %timeit test2(20)
969 ns ± 38.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
sage: %timeit test1(1000)
34.6 µs ± 1.92 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit test2(1000)
43.9 µs ± 2.18 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

src/sage/geometry/cone_critical_angles.py
orlitzky: I've just added the spaces.

src/sage/geometry/cone_critical_angles.py
orlitzky: They won't appear in the documentation anyway unless we add the module to the doc index (see the reverted TODO in the PR description). I personally like seeing the docs but I don't want to give anyone the impression that these are supported/stable interfaces. Not that I foresee major changes, but I'd like to reserve the right to hack away parts of  `cone_critital_angles.py` without deprecation warnings so long as `max_angle()` remains untouched.

src/sage/geometry/cone_critical_angles.py
orlitzky: Ah, all my examples were small. In that case I'd rather leave this alone. The algorithm is combinatorial in the size of `gs` and `hs`,  so you're not going be solving problems where `M` is large enough to see any benefit here.

tscrim: Okay, no problem.

src/sage/geometry/cone_critical_angles.py
tscrim: I would add it to the doc index. You're right that having it exposed like that would mean, strictly speaking, we should deprecate stuff. Although we can be flexible, but you can also add some redundancy with an additional statement saying everything except `max_angle()` in this file might change without a deprecation.

src/sage/geometry/cone_critical_angles.py
orlitzky: Since I'm keeping `M = []` and the append, I was able to omit the indices entirely. Looks like a small improvement, thanks!

src/sage/geometry/cone_critical_angles.py
orlitzky: I may have accidentally made this acceptable to PEP8. In any case, the style police haven't filed any charges.

src/sage/geometry/cone_critical_angles.py
orlitzky: I envy your optimism :)

I've added it back to the doc index along with a big warning at the top of the module doc stating that everything is subject to change.

I think that's it, thanks for taking a look at this.

tscrim (approved):
Okay, thank you. Then let it be so.

orlitzky:
Thanks for the review!

PR #37855, opened 2024-04-23
`sage.misc.package`: Remove deprecated code

mkoeppe:
Remove code deprecated in:
- #31013 (2022)
- #30747 (2020)
- #30607 (2020)

### :memo: Checklist

### DIFF
--- a/src/sage/misc/all.py
+++ b/src/sage/misc/all.py
@@ -40,8 +40,7 @@ lazy_import('sage.misc.profiler', 'Profiler', deprecation=34259)
 lazy_import('sage.misc.dist', 'install_scripts', deprecation=34259)
 lazy_import('sage.misc.trace', 'trace', deprecation=34259)
 lazy_import('sage.misc.package', ('installed_packages', 'is_package_installed',
-                                  'standard_packages', 'optional_packages',
-                                  'experimental_packages', 'package_versions'),
+                                  'package_versions'),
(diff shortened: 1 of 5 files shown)


src/sage/misc/all.py
```diff
@@ -40,8 +40,7 @@
 lazy_import('sage.misc.dist', 'install_scripts', deprecation=34259)
 lazy_import('sage.misc.trace', 'trace', deprecation=34259)
 lazy_import('sage.misc.package', ('installed_packages', 'is_package_installed',
-                                  'standard_packages', 'optional_packages',
-                                  'experimental_packages', 'package_versions'),
+                                  'package_versions'),
```
kiwifb, line 43: I suggest we remove `lazy_import('sage.misc.package',...)` altogether. Those functions are always explicitly imported and only in limited context (setup.py and feature.pkg_systems.py). `package_versions` is never ever called anywhere anymore which suggests it should be at least deprecated if not removed.

mkoeppe: The import is already deprecated, but it's short of 1 year of deprecation

kiwifb: Fair enough.

kiwifb (approved):
Getting rid of that stuff is good, wish we could get rid of more faster.

mkoeppe:
Thank you!

PR #37856, opened 2024-04-23
`sage.misc.dist`: Remove; deprecated in #30207 (2022)

mkoeppe:
Deprecated in:
- #30207 (2022)

### :memo: Checklist


jhpalmieri:
Looks good to me.

mkoeppe:
Thank you!

soehms:
This broke our `docker/Dockerfile` for Docker Hub (see [this workflow run](https://github.com/sagemath/sage/actions/runs/8931531496/job/24537083286#step:8:26693)) in [line 234](https://github.com/sagemath/sage/blob/ffbbea9cb2360340f1c5517b590883fdaf02575c/docker/Dockerfile#L233):

```
# Put scripts to start gap, gp, maxima, ... in /usr/bin
RUN sudo $SAGE_ROOT/sage --nodotsage -c "install_scripts('/usr/bin')"
```

Can I just delete these two lines or must they be replaced by something else?

mkoeppe:
Sorry about missing this!
Perhaps `$SAGE_ROOT/local/bin` should just be prepended to PATH in this image.

soehms:
> Sorry about missing this! 

No problem!

> Perhaps `$SAGE_ROOT/local/bin` should just be prepended to PATH in this image.

I'm implementing this in #38049, but I have to confess that it's not clear to me for what purpose we need to have `$SAGE/local/bin` in the system `$PATH`? I manually built the two missing Docker images for 10.4.beta5 and 10.4.beta6 and pushed them to Docker Hub. In the first image I removed the line with `install_scripts`. In `beta6` I followed your suggestion. In both images, `$SAGE_ROOT/local/bin` is contained in `$PATH` within Sage. Accordingly, `gap_console()`, for example, works for both images. The only difference I've observed occurs when I execute a `bash` terminal in a container running on these images: In a container for `beta6`, I can run `gap` from the `bash` prompt what is not possible in a container on `beta5`. On the other hand, `sage -gap` works in both cases.

Therefore, it seems to me that the purpose of adding `$SAGE_ROOT/local/bin` to the system `$PATH` is not necessary for Sage to work. Leaving it out could eliminate a source of confusion.

mkoeppe:
> adding `$SAGE_ROOT/local/bin` to the system `$PATH` is not necessary for Sage to work

I agree.

I can't comment on what expectations people might have regarding what's available from bash in these Docker images.

soehms:
> I can't comment on what expectations people might have regarding what's available from bash in these Docker images.

I do not know that either. If in doubt, we should leave it as it is. I make the comment clearer on this.

PR #37737, opened 2024-04-03
Follow up #36741: improve code structure

soehms:
From https://github.com/sagemath/sage/pull/36741#issuecomment-2027705010

> Overall, I would suggest to remove the direct and unconditional use of multiprocessing from sage.features.
> Perhaps sage.doctest can put such Value attributes into features that are to be hidden.

This PR implements the suggestion made in https://github.com/sagemath/sage/pull/36741#issuecomment-2028145753

> I looked a little bit into it and here's an idea:
> 
> * In features, implement a simple hide() / unhide() / is_hidden() interface.
> * The only state is _hidden, not shared (for parallel doctesting this will be set before fork so it's ok)
> * Implement AvailableSoftware.hidden() similar to AvailableSoftware.seen() so the logic stays in that class and internals don't leak to sage.doctest.control (if necessary use _seen[idx] to record hidden state and/or number of hidings, or add another shared array)

Fixes #37905

### :memo: Checklist

- [ ] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37857, opened 2024-04-23
Replace special handling of optional extensions (bliss, coxeter3, ....)

mkoeppe:
We relieve the distribution **sagemath-standard** (in both versions - `SAGE_ROOT/pkgs/sagemath-standard` and `SAGE_ROOT/src`) from the duty to build "optional extensions" based on what Sage packages are installed.

The installation is now done uniformly using the modularized distributions **sagemath-bliss**, **sagemath-coxeter3** etc. 

We introduce the missing features `coxeter3` and `sirocco` so that the doctester does not have to rely on the sage-the-distro installation records any more.

The wheels of the distributions now build correctly even when not going through building an sdist first, which previously was required to apply MANIFEST-based filtering. This is achieved using the new `sage_setup.command.build_py`.

User-visible change: 
- To install these options, use `./configure --enable-sagemath_bliss` before building, or use `./sage -i sagemath_bliss` or `make sagemath_bliss`.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37737 (merged here)
- Depends on #37973 (merged here)

### DIFF
--- a/build/pkgs/sagelib/spkg-install.in
+++ b/build/pkgs/sagelib/spkg-install.in
@@ -43,8 +43,6 @@ unset SAGE_PKG_CONFIG_PATH
 
 SITEPACKAGESDIR=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
 
-export SAGE_OPTIONAL_PACKAGES_WITH_EXTENSIONS=""
-
 if [ "$SAGE_EDITABLE" = yes ]; then
     # In an incremental build, we may need to uninstall old versions installed by distutils
     # under the old distribution name "sage" (before #30912, which switched to setuptools
--- a/build/pkgs/sagelib/spkg-install.in
+++ b/build/pkgs/sagelib/spkg-install.in
@@ -43,12 +43,10 @@ unset SAGE_PKG_CONFIG_PATH
 
 SITEPACKAGESDIR=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
 
+# Make sure that an installed old version of sagelib in which sage is an ordinary package
+# does not shadow the namespace package sage during the build.
+(cd "$SITEPACKAGESDIR" && rm -f sage/__init__.py)
 if [ "$SAGE_EDITABLE" = yes ]; then
-    # In an incremental build, we may need to uninstall old versions installed by distutils
-    # under the old distribution name "sage" (before #30912, which switched to setuptools
-    # and renamed the distribution to "sagemath-standard").  There is no clean way to uninstall
-    # them, so we just use rm.
-    (cd "$SITEPACKAGESDIR" && rm -rf sage sage-[1-9]*.egg-info sage-[1-9]*.dist-info)
     # Until https://github.com/sagemath/sage/issues/34209 switches us to PEP 660 editable wheels
     export SETUPTOOLS_ENABLE_FEATURES=legacy-editable
     sdh_pip_editable_install .
@@ -59,12 +57,8 @@ if [ "$SAGE_EDITABLE" = yes ]; then
     fi
 else
     # Now implied: "$SAGE_WHEELS" = yes
-    # Make sure that an installed old version of sagelib in which sage is an ordinary package
-    # does not shadow the namespace package sage during the build.
-    (cd "$SITEPACKAGESDIR" && rm -f sage/__init__.py)
-    # Likewise, we should remove the egg-link that may have been installed previously.
+    # We should remove the egg-link that may have been installed previously.
     (cd "$SITEPACKAGESDIR" && rm -f sagemath-standard.egg-link)
-
     # Use --no-build-isolation to avoid rebuilds because of dependencies:
     # Compiling sage/interfaces/sagespawn.pyx because it depends on /private/var/folders/38/wnh4gf1552g_crsjnv2vmmww0000gp/T/pip-build-env-609n5985/overlay/lib/python3.10/site-packages/Cython/Includes/posix/unistd.pxd
     sdh_pip_install --no-build-isolation .
--- a/build/pkgs/sagemath_bliss/spkg-install.in
+++ b/build/pkgs/sagemath_bliss/spkg-install.in
@@ -3,13 +3,7 @@ cd src
 export PIP_NO_INDEX=true
 export PIP_FIND_LINKS="file://$SAGE_SPKG_WHEELS"
 
-if [ "$SAGE_EDITABLE" = yes ]; then
-    # SAGE_ROOT/src/setup.py installs everything, nothing to do...
-    if [ "$SAGE_WHEELS" = yes ]; then
-        # ... except we build the wheel if requested
-        sdh_setup_bdist_wheel && sdh_store_wheel .
-    fi
-else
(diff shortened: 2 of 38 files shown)


mkoeppe:
@soehms Here I'm running into a failure of the `--hide` tests. Would you be able to take a look?
```
Features detected for doctesting: mcqd,meataxe,networkx,numpy,sage.combinat,sage.geometry.polyhedron,sage.groups,sage.libs.flint,sage.libs.gap,sage.libs.linbox,sage.libs.m4ri,sage.libs.pari,sage.libs.singular,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.number_field,sage.rings.padics,sage.rings.real_mpfr,sage.symbolic,scipy,sympy
Features that have been hidden: meataxe
```

soehms:
> @soehms Here I'm running into a failure of the `--hide` tests. Would you be able to take a look?
> 
> ```
> Features detected for doctesting: mcqd,meataxe,networkx,numpy,sage.combinat,sage.geometry.polyhedron,sage.groups,sage.libs.flint,sage.libs.gap,sage.libs.linbox,sage.libs.m4ri,sage.libs.pari,sage.libs.singular,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.number_field,sage.rings.padics,sage.rings.real_mpfr,sage.symbolic,scipy,sympy
> Features that have been hidden: meataxe
> ```

If you mean [this one](https://github.com/sagemath/sage/actions/runs/8810499915/job/24182973760?pr=37857#step:18:29):


```
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/doctest/control.py
Error: Failed example:: Got: 714



    with open(filename, 'w') as f:
        f.write(test_hide)
        f.close()
Expected:
    729
Got:
    714
Error: Failed example:: Got: Running doctests with ID 2024-04-24-03-32-56-91e6ed54.
Running with SAGE_LOCAL='/sage/local' and SAGE_VENV='/sage/local/var/lib/sage/venv-python3.10'
```
This is because you made changes in `test_hide`. So this can be fixed by `729` -> `714`. However, this is obviously not the test that showed:
```
Features that have been hidden: meataxe
```
Where can I find it, or how to reproduce that?

mkoeppe:
Thanks for the fast reaction, and apologies for not providing enough context; it was late in the evening for me.

What I find puzzling is that `meataxe` is showing both in "Features detected for doctesting" and "Features that have been hidden".

In these tests (which you can see in the CI logs, or in the annotations shon in the ["Files changed" tab](https://github.com/sagemath/sage/pull/37857/files) here, the packages `meataxe` and `sagemath_meataxe` are installed.

You can reproduce it locally with this branch by doing `make sagemath_meataxe`.

mkoeppe:
Meanwhile I'll fix the trivial failure 729 -> 714.

kiwifb:
Thanks for pointing me here from #37905 I searched issues but not PR. It is rather odd and I cannot see at which point the hiding is done (we have some hidden hiding?).

soehms:
> Thanks for pointing me here from #37905 I searched issues but not PR. It is rather odd and I cannot see at which point the hiding is done (we have some hidden hiding?).

On which branch did you observe the failure? I can only reproduce it on the branch of this PR. As mentioned in my previous comment, it disappears if the following lines from it (in `control.py`) are omitted:

```
                        if pkg.name in ['bliss', 'coxeter3', 'mcqd', 'meataxe', 'sirocco', 'tdlib']:
                            continue
```

So what could you have had in your test that replaced what those two lines do?

kiwifb:
As I mentioned in #37905 it has a sage-on-gentoo self inflicted component. I only ship a very limited version of `sage/misc/package.py` and apply this patch https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sagemath-standard/files/sagemath-standard-10.4-neutering.patch which removes the part of `control.py` that depends on `list_packages` - and that's exactly where your own patch to `control.py` is located.

soehms:
> As I mentioned in #37905 it has a sage-on-gentoo self inflicted component. I only ship a very limited version of `sage/misc/package.py` and apply this patch https://github.com/cschwan/sage-on-gentoo/blob/master/sci-mathematics/sagemath-standard/files/sagemath-standard-10.4-neutering.patch which removes the part of `control.py` that depends on `list_packages` - and that's exactly where your own patch to `control.py` is located.

Thanks for the explanation. I was now able to locate the cause of the error.

This is because `AvailableSoftware.__contains__` caches its result via the `_seen` array. Here's what's happening: Due to the changes in this PR or the changes in the Gentoo patch, `meataxe` was not in the `options.optional` set. On the other hand there is a line in `control.py` with the tag `# optional - meataxe`. Therefore, in `SageDocTestParser.parse`, the test `'meataxe' in available_software` is called before the feature is set to hidden. This results in `_seen` being set to `1` for Meataxe. When later parsing the tests in the `test_hide` file, `'meataxe' in available_software` shows `True` again, even though the feature has now been hidden.

We currently have PR #37737 for an improvement to the implementation of the hide option. I'll see how to integrate a solution to the problem into that PR. For now, I would suggest adding the following two lines to `AvailableSoftware.__contains__` to fix the bug immediately:

```
diff --git a/src/sage/doctest/external.py b/src/sage/doctest/external.py
index 5dc1ed7b82..c396fbab38 100644
--- a/src/sage/doctest/external.py
+++ b/src/sage/doctest/external.py
@@ -445,6 +445,8 @@ class AvailableSoftware():
             idx = self._indices[item]
         except KeyError:
             return False
+        if self._features[idx]._hidden:
+            return False
         if not self._seen[idx]:
             if not self._allow_external and self._features[idx] in self._external_features:
                 self._seen[idx] = -1 # not available
```

soehms:
> We currently have PR #37737 for an improvement to the implementation of the hide option. I'll see how to integrate a solution to the problem into that PR.

I've implemented a solution to the issue there (see https://github.com/sagemath/sage/pull/37737#issuecomment-2102989850 and commit https://github.com/sagemath/sage/pull/37737/commits/f881e2b3fec578fbeeb29dfaf2e8badef7bd8900).

kiwifb:
Will test as soon as possible, my Friday is already marked has gone so I am not sure when it will be.

kiwifb:
I have started a private test workflow on my computer, I will let you know the results once it is done.

kiwifb:
The workflow thing did not go well because I had not taken into some "interference". Will try a different way.

kiwifb:
I'll also further note that I have been impacted by https://github.com/cschwan/sage-on-gentoo/issues/783 which is like the kind of stuff we see with setuptools_scm (but worse) and will only stop when either sagemath moves to meson or people bloody stop injecting their stuff in setuptools.

mkoeppe:
... or when downstream architects stop trying to circumvent build isolation ....

mkoeppe:
My modest proposal for this whole problem: A site-packages slurper that creates a PEP-660 editable wheel from a package installed in site-packages. Then build with proper build isolation instead of `gpep517` and use a `PIP_CONSTRAINT` that forces the use of these wheels.

kiwifb (approved):
I could build and test the thing. I think we need to move it forward now.

mkoeppe:
Thank you!

PR #37859, opened 2024-04-24
Implement the ladder idempotents of the symmetric group algebra

tscrim:
We implement the so-called ladder idempotents of the symmetric group algebra $F[S_n]$, defined by Ryom-Hansen, which can be used to construct the projective covers of simple modules when the field $F$ has positive characteristic.

### :memo: Checklist


tscrim:
@AndrewMathas This is something I came across while trying to devise a method to construct the projective covers. Thought it might be worthwhile to add.

tscrim:
Thank you. That's a good point. I think I lost track of what $e$ meant when writing the function. I also made the other change you suggested and added your code.

In the trac workflow, you would be allowed to push commits. However, with GH this has become very inconvenient for reviewers (or general collaborative code writing) to do... (The "standard" ways would be to either do a PR to the branch on my fork (not the main repo here);  just do your own PR with the additional commit(s); or I pull commits from a specific branch of yours) Well, no matter what I am okay with you pushing changes in one form or another (we might even set up something so you can just generically push to my fork, but it might not be worth the trouble)

src/sage/combinat/partition.py
```diff
@@ -2945,36 +2945,41 @@ def top_garnir_tableau(self,e,cell):
...
-        The *ladder tableau* is the standard Young tableau obtained by reading
-        the *ladders*, the set of cells `(i, j)` that differ from `(i+e, j-1)`,
-        of the partition `\lambda` from left-to-right.
+        The `e`-*ladder tableau* is the standard Young tableau obtained
+        by reading the *ladders*, the set of cells `(i, j)` that differ
+        from `(i+e-1, j-1)`, of the partition `\lambda` from left-to-right.
 
         INPUT:
 
         - ``e`` -- a positive integer
```
AndrewMathas: Sorry, should have caught this earlier too. You want `e` to be a non-negative integer, where `e=0` corresponds to (quantum) characteristic 0.

src/sage/combinat/partition.py
```diff
@@ -2945,36 +2945,41 @@ def top_garnir_tableau(self,e,cell):
...
-            sage: la.ladder_tableau(2, ladder_sizes=True)[1]
+            sage: la.ladder_tableau(3, ladder_lengths=True)[1]
             [1, 1, 2, 2, 3, 3, 3]
         """
         Tlad = [[None] * val for val in self]
         counter = 0
         start = 0
         n = sum(self)
         sizes = []
+        e -= 1
```
AndrewMathas: To cope with `e=0` this could be: `e = self.size() if e=0 else e-1`

AndrewMathas (changes requested):
Looks good but one more minor change that I should have picked up the first time.

tscrim:
What do you want `ladders()` to do when `e = 0`? Error out?

tscrim:
For the `ladder_tableau()` method, I agree that this is clear. However, the behavior of `ladders()` depends on the size.  While it isn't a major difference, it is not quite consistent as I add boxes to the bottom. Compare:
```
sage: Partition([3, 2]).ladders(0)
{0: [(0, 0)], 1: [(0, 1)], 2: [(0, 2)], 5: [(1, 0)], 6: [(1, 1)]}
sage: Partition([3, 2, 1]).ladders(0)
{0: [(0, 0)], 1: [(0, 1)], 2: [(0, 2)], 6: [(1, 0)], 7: [(1, 1)], 12: [(2, 0)]}
sage: Partition([3, 2, 1, 1]).ladders(0)
{0: [(0, 0)], 1: [(0, 1)], 2: [(0, 2)], 7: [(1, 0)], 8: [(1, 1)], 14: [(2, 0)], 21: [(3, 0)]}
```
In particular, see the ladder for the cell `(1, 0)` changing. It's just a matter of what the best behavior would be, but it is unlikely to be used in that setting I feel. So this might basically be bikeshedding...

tscrim:
Thank you. I added some more documentation and examples to illustrate these points about `e = 0`. If everything looks good to you, then please approve the PR and also set it to a positive review. (The former is more important as I cannot do that, but I can do the latter).

tscrim:
Thank you. Could you also please approve the PR by going to the "files changed" and do the "review changes" with selecting the "approve"?

tscrim:
(The last force push was just to remove some trailing whitespace I accidentally added.)

AndrewMathas (approved):
Everything looks good to me.

tscrim:
Thank you.

vbraun:
Docs don't build (see CI)

tscrim:
Sorry about that; fixed.

PR #37861, opened 2024-04-24
Fix segfault for Weil polynomials

### LINKED ISSUE #37860: weil_polynomials
### Steps To Reproduce

R.<T> = ZZ[]
R.weil_polynomials(0,1,-1)

### Expected Behavior

This should return an empty list.

### Actual Behavior

Currently it results in a segmentation failure. 

### Additional Information

_No response_

### Environment

```markdown
- **OS**: macOS Monterey 12.5.1
- **Sage Version**: 10.1
```

### Checklist

kedlaya:
Fix segfault for Weil polynomials

Eliminate a segfault that occurs when asking for Weil polynomials of degree 0 and sign -1 (which do not exist, so the iterator should return empty). This fixes #37860.


roed314:
LGTM.

kedlaya:
I noticed there is also a segfault when the degree is negative. The root cause is passing a negative degree to dfs_manager, so I moved the check to catch both issues at once, and added a second doctest.

tscrim (approved):
LGTM.

PR #37865, opened 2024-04-24
`sage.numerical.interactive_simplex_method`: improve support for base fields other than `QQ` and `RDF`

### LINKED ISSUE #31312: interactive_simplex_method: Improve support for base fields other than QQ, RDF
<div id="comment:0"></div>

We remove the ad-hoc conversion of data to `RDF` in `InteractiveLPProblem.feasible_set`.  This is outdated and problematic:
- the polyhedra code over `RDF` is not robust at all 
- Sage has good support for algebraic polyhedra via normaliz.

We also make the printing code in `_latex_product` more robust: 
- it no longer relies on `SR` to decide whether a coefficient needs to be wrapped in parens; instead, we use the same simple test that `sage.misc.repr.coeff_repr` already uses
- recognizing trivial coefficients (0, 1) is switched to comparing text instead of comparing elements - in order to support fields such as `SR` where `==` is magic

CC:  @yuan-zhou @novoselt

Component: **linear programming**

Author: **Matthias Koeppe, ...**

_Issue created by migration from https://trac.sagemath.org/ticket/31312_

mkoeppe:
We remove the ad-hoc conversion of data to `RDF` in `InteractiveLPProblem.feasible_set`.  This is outdated and problematic:
- the polyhedra code over `RDF` is not robust at all 
- Sage has good support for algebraic polyhedra, in particular via normaliz.

We also make the printing code in `_latex_product` more robust: 
- it no longer relies on `SR` to decide whether a coefficient needs to be wrapped in parens; instead, we use the same simple test that `sage.misc.repr.coeff_repr` already uses
- recognizing trivial coefficients (0, 1) is switched to comparing text instead of comparing elements - in order to support fields such as `SR` where `==` is magic

Rebased version of:
- #31312

Author: @mkoeppe, @ComboProblem

Resolves #31312.

### :memo: Checklist

### DIFF
--- a/src/sage/numerical/interactive_simplex_method.py
+++ b/src/sage/numerical/interactive_simplex_method.py
@@ -650,7 +650,7 @@ class InteractiveLPProblem(SageObject):
         b = vector(b)
         c = vector(c)
         if base_ring is None:
-            base_ring = vector(A.list() + list(b) + list(c)).base_ring()
+            base_ring = vector(A.list() + list(b) + list(c) + [objective_constant_term]).base_ring()
         base_ring = base_ring.fraction_field()
         A = A.change_ring(base_ring)
         A.set_immutable()
--- a/src/sage/numerical/interactive_simplex_method.py
+++ b/src/sage/numerical/interactive_simplex_method.py
@@ -304,11 +304,12 @@ def _latex_product(coefficients, variables,
     """
     entries = []
     for c, v in zip(coefficients, variables):
-        if c == 0:
+        t = latex(c)
+        if t == '0':
             entries.extend(["", ""])
             continue
         sign = "+"
-        if latex(c).strip().startswith("-"):
+        if t.strip().startswith("-"):
             sign = "-"
             c = - c
         if c == 1:
--- a/src/sage/numerical/interactive_simplex_method.py
+++ b/src/sage/numerical/interactive_simplex_method.py
@@ -1145,9 +1145,13 @@ class InteractiveLPProblem(SageObject):
             objective_constant_term=self._constant_term)
 
     @cached_method
-    def feasible_set(self):
+    def feasible_set(self, backend=None):
         r"""
         Return the feasible set of ``self``.
+
+        INPUT:
+
+        - (optional) a backend for :mod:`Polyhedron <sage.geometry.polyhedron.constructor>`
 
         OUTPUT: a :mod:`Polyhedron <sage.geometry.polyhedron.constructor>`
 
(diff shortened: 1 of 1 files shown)


src/sage/numerical/interactive_simplex_method.py
```diff
@@ -301,22 +300,23 @@ def _latex_product(coefficients, variables,
         sage: var("x, y")                                                               # needs sage.symbolic
         (x, y)
         sage: print(_latex_product([-1, 3], [x, y]))                                    # needs sage.symbolic
-        - \mspace{-6mu}&\mspace{-6mu} x \mspace{-6mu}&\mspace{-6mu} + \mspace{-6mu}&\mspace{-6mu} 3 y
+        - \mspace{-6mu}&\mspace{-6mu} 1 x \mspace{-6mu}&\mspace{-6mu} + \mspace{-6mu}&\mspace{-6mu} 3 y
```
mkoeppe: @ComboProblem I think this is an unintended change. Could you investigate what's causing this behavior?

ComboProblem: The issue is in Line 316-322. For the text input, with `c=-1` and `v = var("x")`, we have that `t.strip() = '-1'` and `c = 1`. So the if block here isn't executed and the else block code writes a coefficient of c=1into the expression. 

Furthermore, with fields that have a presentation of 1 that is not the string 1, we have that outputs are not as we'd expect for a coefficient of 1.   See the attached  example with RDF. Screenshot from 2024-05-14 11-07-58

For the `-1` as a coefficient example, I think a change to line 316 like `if t.strip() == '1' or t.strip() == '-1':` would fix this. For fields with different presentations of 1 other than the string 1, I'm unsure of how to approach this.

mkoeppe: I think we can live with 1.0 showing as 1.0; what's more surprising to me is that the objective constant shows as 0, not 0.0

ComboProblem: The default constructor for `InteractiveLPProblem` specifies the `objective_constant_term` term as the integer 0 rather than the base ring element 0. The code doesn't explicitly convert the  `objective_constant_term` to an element of the base ring.

mkoeppe: Thanks. That should probably be changed.

ComboProblem: I've just opened a PR that fixes this with the other changes discussed so far.

mkoeppe: Thanks! I've merged your PR https://github.com/mkoeppe/sage/pull/41 now

PR #37878, opened 2024-04-27
Run sage doc server for jupyterlab 

### LINKED ISSUE #34794: Run doc server along with jupyter server
<div id="comment:0"></div>

Sage documentation links in `help` menu have been long broken in jupyterlab notebook interface, although they work fine in jupyter notebook interface.

Jupyterlab loads local files (sage documentation files) in `iframe` tag. This is problematic and it seems hopeless to make this work. A proper way is to run local http server serving the doc along with jupyter(lab) notebook server.

The branch uses thread parallelism to run sage doc server.

Try the branch with `sage --notebook=jupyterlab`. This runs a light-weight sage doc server at `http://localhost:<SAGE_DOC_LOCAL_PORT>`, where the environment variable `SAGE_DOC_LOCAL_PORT` defaults to 8000.

Set `export SAGE_DOC_SERVER_URL=https://doc.sagemath.org` before running `sage --notebook=jupyterlab` to use the public doc server. This will be useful for jupyterhub users.

Component: **notebook**

Author: **Kwankyu Lee**

Branch/Commit: **[public/run_doc_server](https://github.com/sagemath/sagetrac-mirror/tree/public/run_doc_server) @ [`70bdf46`](https://github.com/sagemath/sagetrac-mirror/commit/70bdf46e811b15d9886077ebdc5ce544626dc963)**

_Issue created by migration from https://trac.sagemath.org/ticket/34794_

kwankyu:
Reopen #35139 because of the renewed interest: https://groups.google.com/g/sage-devel/c/kzSWB8ps7VA

New environment variables added to `sage.env`:
```
SAGE_DOC_SERVER_URL = var("SAGE_DOC_SERVER_URL")
SAGE_DOC_LOCAL_PORT = var("SAGE_DOC_LOCAL_PORT", "0")
```
When Jupyter notebook launches,
- If `SAGE_DOC_SERVER_URL` is set, the url is used; 
- else if local sage documentation is installed, a doc http server with port `SAGE_DOC_LOCAL_PORT` starts. If the port is 0 (the default), then a random port is selected by the system and stored to the environment variable;
- else the online official documentation website https://doc.sagemath.org is used

for the help menu of the Jupyter notebook.

Fixes #34794.

### :memo: Checklist


culler (commented):
Why do you use 8000 for the default port?  That is the http port > 1024 which is most likely to be used by another local http server.  If you use 0 as the default port then the system will assign a random unused port > 1024.  That way you won't have to worry about port conflicts.

I also would recommend using 127.0.0.1 in the url, rather than localhost.  The reason is that a (rogue) DNS nameserver could, in principle, assign an arbitrary IP address to the name localhost.  But the IP address 127.0.0.1 always refers to the loopback interface and would never trigger a DNS lookup.

kwankyu:
> Why do you use 8000 for the default port? That is the http port > 1024 which is most likely to be used by another local http server. If you use 0 as the default port then the system will assign a random unused port > 1024. That way you won't have to worry about port conflicts.

But then how can I know the port randomly assigned?

kwankyu:
`server.server_address[-1]` gives the port. Then I need to figure out how to pass this number to `kernel.py`...

kwankyu:
> Why do you use 8000 for the default port? That is the http port > 1024 which is most likely to be used by another local http server. If you use 0 as the default port then the system will assign a random unused port > 1024. That way you won't have to worry about port conflicts.
> 
> I also would recommend using 127.0.0.1 in the url, rather than localhost. The reason is that a (rogue) DNS nameserver could, in principle, assign an arbitrary IP address to the name localhost. But the IP address 127.0.0.1 always refers to the loopback interface and would never trigger a DNS lookup.

Done. Thanks for the tips!

src/bin/sage-notebook
```diff
@@ -227,4 +250,19 @@ if __name__ == '__main__':
...
+    banner()
+
+    # Start a Sage doc server if the Sage documentation is available locally.
+    # See the corresponding code in src/sage/repl/ipython_kernel/kernel.py.
+
+    from sage.env import SAGE_DOC_SERVER_URL
+    from sage.features.sagemath import sagemath_doc_html
+
+    if SAGE_DOC_SERVER_URL:
+        print(f'Sage doc server running at {SAGE_DOC_SERVER_URL}')
```
kwankyu:
```suggestion
        print(f'Sage doc server is running at {SAGE_DOC_SERVER_URL}')
```

kwankyu:
ping @culler @nbruin @mkoeppe

mkoeppe:
Looks like a reasonable approach. After #36730, one might discuss if the doc server should be the responsibility of the sagemath_doc_html distribution, but for now this looks good.

kwankyu:
Thanks!!

kwankyu:
Bump up priority since the original poster of the issue regards it as a bug.

Issue #37877, opened 2024-04-26
`FreeModule`, `FiniteDimensionalModulesWithBasis` disagree on what 'left' and 'right' mean


mkoeppe:
We have
```
class FiniteDimensionalModulesWithBasis(CategoryWithAxiom_over_base_ring):
...
    class MorphismMethods:
...
        def matrix(self, base_ring=None, side="left"):
            r"""
...
            INPUT:
...
            - ``side`` -- "left" or "right" (default: "left")

            If ``side`` is "left", this morphism is considered as
            acting on the left; i.e. each column of the matrix
            represents the image of an element of the basis of the
            domain.
```

.... but ....

```
class MatrixMorphism(MatrixMorphism_abstract):
...
    def __init__(self, parent, A, copy_matrix=True, side='left'):
...
    def matrix(self, side=None):
        r"""
        Return a matrix that defines this morphism.

        INPUT:

        - ``side`` -- (default: ``'None'``) the side of the matrix
          where a vector is placed to effect the morphism (function)

        OUTPUT:
...
        Internally, Sage represents a matrix morphism with the
        matrix multiplying a row vector placed to the left of the
        matrix.  If the option ``side='right'`` is used, then a
        matrix is returned that acts on a vector to the right of
        the matrix.
```

PR #37879, opened 2024-04-27
`ModulesWithBasis.MorphismMethods`: Add method for matrix of finite-dimensional restriction/projection, move more `MatrixMorphism` methods to category

mkoeppe:
We extend the technique of "support orders" for infinite-dimensional free modules to their morphisms.
The new helper method `_matrix_side_bases_orders` constructs a matrix suitable for throwing some linear algebra on the modules. (This goes in a similar direction as https://github.com/sagemath/sage/issues/34487.)
```
                sage: X = CombinatorialFreeModule(QQ, ZZ); x = X.basis()
                sage: def on_basis(i):
                ....:     return X.monomial(i+1) - X.monomial(i+2)
                sage: phi = X.module_morphism(on_basis=on_basis, codomain=X)
                sage: phi._matrix_side_bases_orders()
                ...
                ValueError: domain or codomain are not finite-dimensional;
                use row_order or column_order to obtain the matrix
                of a finite-dimensional restriction or projection
                sage: phi._matrix_side_bases_orders(row_order=range(5),
                ....:                               column_order=range(4))
                ( [ 0  0  0  0]
                  [ 1  0  0  0]
                  [-1  1  0  0]
                  [ 0 -1  1  0]
                  [ 0  0 -1  1], 'left',
                  Lazy family (Term map from Integer Ring to Free module generated by
                               Integer Ring over Rational Field(i))_{i in Integer Ring},
                  Lazy family (Term map from Integer Ring to Free module generated by
                               Integer Ring over Rational Field(i))_{i in Integer Ring},
                  range(0, 5), range(0, 4) )
```

As an immediate application of this method, we are able to move some methods (`is_injective`, `is_surjective`, `is_bijective`, `nullity`, `rank`) for the finite-dimensional case from `MatrixMorphism` to the category because we are sidestepping an issue of the existing `matrix` method:
- #37877

### :memo: Checklist


tscrim:
This can only be done reliably when the domain is a finite dimensional module, otherwise you run into the obvious problem with infinite matrices. Likewise `dense_coefficient_list()` should be for elements in that category as well (unless you want to have them be arbitrary-but-finite lists of coefficients reflecting the basis order).

mkoeppe:
The ideas here is that in the infinite-dimensional case, `order` will be a required argument.

mkoeppe:
`dense_coefficient_list` is generalized here from the existing method for the finite dimensional case.

PR #37881, opened 2024-04-27
Implement hypergeometric Euler factors for tame primes

kedlaya:
We continue the port of hypergeometric motives from Magma by implementing the computation of hypergeometric traces and Euler factors for tame primes. This splits into two essentially separate cases: when t has nonzero p-adic valuation ("tame"), and when t-1 has nonzero p-adic valuation ("multiplicative"). In the second case, the computation uses the local functional equation as in the good reduction case.


src/sage/modular/hypergeometric_motive.py
```diff
@@ -114,10 +113,18 @@ def characteristic_polynomial_from_traces(traces, d, q, i, sign):
     - ``i`` -- integer, the weight in the motivic sense
 
     - ``sign`` -- integer, the sign
+    
+    - ``deg`` -- an integer or None
+    
+    - ``use_fe`` -- a boolean (default True)
```
tscrim:
```suggestion

    - ``deg`` -- an integer or ``None``

    - ``use_fe`` -- a boolean (default: ``True``)
```
The whitespace issue is in many places and needs to be fixed (see the linter).

src/sage/modular/hypergeometric_motive.py
```diff
@@ -146,33 +153,42 @@ def characteristic_polynomial_from_traces(traces, d, q, i, sign):
...
-    series = series.O(d // 2 + 1).exp()
+    series = series.O(bound + 1).exp()
     coeffs = list(series)
-    coeffs += [0] * max(0, d // 2 + 1 - len(coeffs))
+    coeffs += [0] * max(0, bound + 1 - len(coeffs))
 
-    data = [0 for _ in range(d + 1)]
-    for k in range(d // 2 + 1):
+    fulldeg = d if deg is None else deg
+    data = [0 for _ in range(fulldeg + 1)]
```
tscrim:
```suggestion
    data = [0] * (fulldeg + 1)
```
Might be slightly faster, but at least IMO it is easier to read.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1633,6 +1649,8 @@ def euler_factor(self, t, p, cache_p=False):
         - `t` -- rational number, not 0 or 1
 
         - `p` -- prime number of good reduction
+        
+        - `deg` -- integer or None
```
tscrim:
```suggestion

        - ``deg`` -- integer or ``None``
```
The other parts of this doc are not correctly formatted, but let's not make it worse. ;) (Feel free to correct the doc for this method too.)

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1642,6 +1660,13 @@ def euler_factor(self, t, p, cache_p=False):
 
         For odd weight, the sign of the functional equation is +1. For even
         weight, the sign is computed by a recipe found in 11.1 of [Watkins]_.
+        
+        If ``deg`` is specified, then the polynomial is only computed up to degree
+        ``deg`` (inclusive).
+        
+        The prime `p` may not be wild. It may be tame if `t` is a `p`-adic unit.
+        When `v_p(t-1)` is nonzero and even, the Euler factor includes a linear term
+        described in 11.2 of [Watkins]_.
```
tscrim:
```suggestion
        described in Section 11.2 of [Watkins]_.
```
I think this is what you meant. Perhaps the reference makes this unambiguous though (I didn't check).

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1753,11 +1793,23 @@ def euler_factor(self, t, p, cache_p=False):
             raise ValueError('p not prime')
         if not all(x.denominator() % p for x in self._alpha + self._beta):
             raise NotImplementedError('p is wild')
-        if (t.numerator() * t.denominator() % p == 0 or (t - 1) % p == 0):
+        e = t.valuation(p)
+        if (t.numerator()*t.denominator() % p == 0):
```
tscrim:
```suggestion
        if t.numerator()*t.denominator() % p == 0:
```
Not needed in Python.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1753,11 +1793,23 @@ def euler_factor(self, t, p, cache_p=False):
...
             raise NotImplementedError('p is tame')
-        # now p is good
-        d = self.degree()
-        bound = d // 2
+        # now p is good, or p is tame and t is a p-adic unit
+        if (t-1) % p == 0:
+            d = self.degree()-1
```
tscrim:
```suggestion
            d = self.degree() - 1
```
PEP8

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1753,11 +1793,23 @@ def euler_factor(self, t, p, cache_p=False):
...
+            bound = d
+            if deg is not None:
+                bound = min(deg, bound)
+            deg = bound
+            use_fe = False
+        else:
+            d = self.degree()
+            bound = d // 2
+            if deg is not None:
+                bound = min(deg, bound)
```
tscrim: This could be pulled outside the `if`-`else` statement to avoid a little bit of duplication.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1766,4 +1818,13 @@ def euler_factor(self, t, p, cache_p=False):
...
-        return characteristic_polynomial_from_traces(traces, d, p, w, sign)
+        ans = characteristic_polynomial_from_traces(traces, d, p, w, sign, deg=deg, use_fe=use_fe)
+
+        # Add an extra factor when w is even and t-1 has nonzero even valuation.
+        if (w % 2 == 0 and (t-1) % p == 0 and (t-1).valuation(p) % 2 == 0):
+            m1 = self.cyclotomic_data()[1].count(1)
+            K = (-1)**((m1-1)//2)*2*prod(abs(x) for x in self.gamma_list())
+            t0 = (~t-1)/p**((t-1).valuation(p))
+            c = kronecker_symbol(K*t0, p)*p**(w//2)
+            ans *= 1 - c*ans.parent().gen()
```
tscrim:
```suggestion
        if w % 2 == 0 and (t-1) % p == 0 and (t-1).valuation(p) % 2 == 0:
            m1 = self.cyclotomic_data()[1].count(1)
            K = (-1) ** ((m1-1) // 2) * 2 * prod(abs(x) for x in self.gamma_list())
            t0 = (~t-1 ) / p**((t-1).valuation(p))
            c = kronecker_symbol(K*t0, p) * p**(w//2)
            ans *= 1 - c * ans.parent().gen()
```
PEP8 stuff. Although I think `(m1-1) // 2) * 2` is better written as `m1 - 1 - (m1 % 2)`.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1723,6 +1750,19 @@ def euler_factor(self, t, p, cache_p=False):
...
             p^f cannot exceed 2^31
 
+        Check handling of some tame cases::
+        
+            sage: H = Hyp(cyclotomic=[[4,2,2],[3,1,1]])
+            sage: H.euler_factor(8, 7)
+            -7*T^3 + 7*T^2 - T + 1
+            sage: H.euler_factor(50, 7)
+            -7*T^3 + 7*T^2 - T + 1
+            sage: H = Hyp(cyclotomic=[[4,2,2,2],[3,1,1,1]])
```
tscrim:
```suggestion
            sage: H = Hyp(cyclotomic=[[4,2,2], [3,1,1]])
            sage: H.euler_factor(8, 7)
            -7*T^3 + 7*T^2 - T + 1
            sage: H.euler_factor(50, 7)
            -7*T^3 + 7*T^2 - T + 1
            sage: H = Hyp(cyclotomic=[[4,2,2,2], [3,1,1,1]])
```

tscrim (changes requested):
I am assuming the math is good (I don't really know anything about it). I just have a few minor comments about the code. Once addressed, I am willing to set a positive review.

src/sage/modular/hypergeometric_motive.py
kedlaya: Just to clarify, according to PEP8
```
 ans *= 1 - c*ans.parent().gen()
```
is preferred over
```
 ans *= 1 - c * ans.parent().gen()
```
(see the section "If operators with different priorities are used").

That said, fixing some of the preexisting formatting issues might be useful to do at this point, let me see what I can do.

tscrim: Both are fine actually. PEP8 allows for higher precedence operators to not have spacing around them. It becomes a preference at that point. Feel free to not add that spacing if you prefer.

kedlaya:
I think all the content is (finally) here now, but I'm not sure what's going on with the build failures in CI.

tscrim:
Those build failures are happening across all PRs. I don't know what's going on there either.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+            raise ValueError('invalid t')
+        if not is_prime(p):
+            raise ValueError('p not prime')
+        if not all(x.denominator() % p for x in self._alpha + self._beta):
+            raise NotImplementedError('p is wild')
+
+        e = t.valuation(p)
+        t0 = t / p**e
+        if e % mo:
+            return 1
```
tscrim:
```suggestion
            return ZZ.one()
```
and similar below. Otherwise all of these would be Python `int`'s, which we should avoid returning.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+        if e > 0:
+            mul = self.cyclotomic_data()[1].count(mo)
+            if not mul:
+                return 1
+        elif e < 0:
+            mul = self.cyclotomic_data()[0].count(mo)
+            if not mul:
+                return 1
+        else:
+            return 1
```
tscrim:
```suggestion
        if e > 0:
            mul = self.cyclotomic_data()[1].count(mo)
        elif e < 0:
            mul = self.cyclotomic_data()[0].count(mo)
        else:
            mul = None
        if not mul:
            return ZZ.one()
```
It took me a minute to understand why the code below would be reached. I think this makes it more clear and removes a little bit of (albeit trivial) redundancy.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+        d = euler_phi(mo)
+        f = IntegerModRing(mo)(p).multiplicative_order()
+        if deg is None:
+            deg = d
+        if deg < f:
+            return 1
+        q = p**f
+        prec = ceil(deg*(self.weight()+1-mul)/2 + log(2*d + 1, p))
+        k = (q-1)//mo
```
tscrim:
```suggestion
            return ZZ.one()
        q = p ** f
        prec = ceil(deg*(self.weight()+1-mul)/2 + log(2*d + 1, p))
        k = (q-1) // mo
```
PEP8

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+            p_ring = Zp(p, prec, 'fixed-mod')
+        M = self.M_value()
+        teich = p_ring.teichmuller(M / t0)
+        m = {r: self._beta.count(QQ((r, q - 1))) for r in range(q - 1)}
+        D = -min(self.zigzag(x, flip_beta=True) for x in self._alpha + self._beta)
+        gamma = self.gamma_array()
+        l = []
+        for j in range(mo):
+            if gcd(j, mo) == 1:
+                r = j*k
```
tscrim:
```suggestion
                r = j * k
```
PEP8

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+                ct = 0
+                for v, gv in gamma.items():
+                    r1 = v * r % (q-1)
+                    ct += gv*sum(r1.digits(p))
+                    term *= p_ring(gtab[r1])**(-gv if flip else gv)
+                ct //= (p-1)
+                term *= ZZ(-1)**ct
+                ct += f * (D + m[0] - m[r])
+                l.append(term * p**ct)
+        traces = [0 if j % f else sum(i**(j//f) for i in l) for j in range(1,d+1)]
```
tscrim:
```suggestion
                ct //= p - 1
                term *= ZZ(-1) ** ct
                ct += f * (D + m[0] - m[r])
                l.append(term * p**ct)
        traces = [0 if j % f else sum(i ** (j//f) for i in l) for j in range(1, d+1)]
```
PEP8 stuff

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1751,25 +1931,85 @@ def euler_factor(self, t, p, cache_p=False):
...
-        d = self.degree()
+        if 0 in self._alpha:
+            return self._swap.euler_factor(~t, p)
+        P = PolynomialRing(ZZ, 'T')
+        if t.numerator() % p == 0 or t.denominator() % p == 0:
+            ans = P.one()
+            for m in set(j for i in self.cyclotomic_data() for j in i):
+                ans *= self.euler_factor_tame_contribution(t, p, m, deg)
+            if deg is not None:
+                ans = ans.truncate(deg+1)
```
tscrim:
```suggestion
                ans = ans.truncate(deg + 1)
```
PEP8

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1751,25 +1931,85 @@ def euler_factor(self, t, p, cache_p=False):
...
+                    u *= prod(v**gv for v, gv in self.gamma_array().items())
+                    c = kronecker_symbol(u, p) * p**((w-1)//2)
+                else:
+                    u = (-1) ** (1 + self.degree()//2 + (m1-1)//2)
+                    num, den = self.defining_polynomials()
+                    x = num.parent().gen()
+                    num = num(-x)
+                    num /= (x-1) ** num.valuation(x-1)
+                    den /= (x-1) ** den.valuation(x-1)
+                    u *= 2 * num(1) / den(1)
```
tscrim, line 1985: Wouldn't it be better to evaluate to `1` first, then do the exponentiation (of course, holding onto the valuations before specializing)? Same for the block below.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1751,25 +1931,85 @@ def euler_factor(self, t, p, cache_p=False):
...
+        if typ == "mult":
+            if self.degree() % 2 == 0:
+                ans *= tmp
+            if w % 2 == 0 and (t-1).valuation(p) % 2 == 0:
+                K = (-1) ** ((m1-1)//2)*2*prod(abs(x) for x in self.gamma_list())
+                t0 = (~t-1) / p**((t-1).valuation(p))
+                c = kronecker_symbol(K*t0, p) * p**(w//2)
+                ans *= 1 - c*P.gen()
+            if deg is not None:
+                ans = ans.truncate(deg+1)
```
tscrim:
```suggestion
                ans = ans.truncate(deg + 1)
```
PEP8

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1751,25 +1931,85 @@ def euler_factor(self, t, p, cache_p=False):
...
+                    u = (-1) ** (1 + self.degree()//2 + (m1-1)//2)
+                    num, den = self.defining_polynomials()
+                    x = num.parent().gen()
+                    num = num(-x)
+                    num /= (x-1) ** num.valuation(x-1)
+                    den /= (x-1) ** den.valuation(x-1)
+                    u *= 2 * num(1) / den(1)
+                    c = kronecker_symbol(u, p) * p**(w//2)
+                for j in range(len(traces)):
+                    traces[j] -= c ** (j+1)
```
tscrim:
```suggestion
                    c = kronecker_symbol(u, p) * p**(w//2)
                cpow = c
                for j in range(len(traces)):
                    traces[j] -= cpow
                    cpow *= c
```
I think this will be faster than repeatedly doing the exponentiation. Possibly more of a micro-optimization.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1751,25 +1931,85 @@ def euler_factor(self, t, p, cache_p=False):
...
+
+        # In the multiplicative case, we sometimes need to pull out a linear factor
+        # in order to apply the functional equation.
+        if typ == "mult":
+            if self.degree() % 2 == 0:
+                sign = 1
+                if w % 2:
+                    assert m1 % 2 == 0
+                    u = (-1) ** (m1//2)
+                    u *= prod(v**gv for v, gv in self.gamma_array().items())
```
tscrim:
```suggestion
                    u *= prod(v ** gv for v, gv in self.gamma_array().items())
```
PEP8 (I prefer the way you had it, but alas...)

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1647,12 +1766,18 @@ def euler_factor(self, t, p, cache_p=False):
         See [Benasque2009]_ for explicit examples of Euler factors.
 
         For odd weight, the sign of the functional equation is +1. For even
-        weight, the sign is computed by a recipe found in 11.1 of [Watkins]_.
+        weight, the sign is computed by a recipe found in section 11.1 of [Watkins]_.
```
tscrim:
```suggestion
        weight, the sign is computed by a recipe found in Section 11.1 of [Watkins]_.
```
At least I am used to this being considered a proper noun.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+            if gcd(j, mo) == 1:
+                r = j*k
+                term = teich**r * ZZ(-1)**m[0]
+                ct = 0
+                for v, gv in gamma.items():
+                    r1 = v * r % (q-1)
+                    ct += gv*sum(r1.digits(p))
+                    term *= p_ring(gtab[r1])**(-gv if flip else gv)
+                ct //= (p-1)
+                term *= ZZ(-1)**ct
```
tscrim:
```suggestion
                    ct += gv * sum(r1.digits(p))
                    term *= p_ring(gtab[r1]) ** (-gv if flip else gv)
                ct //= (p-1)
                term *= ZZ(-1) ** ct
```
PEP8

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
             sign = kronecker_symbol(t * (t - 1) * self._sign_param, p)
         return sign
 
+    def euler_factor_tame_contribution(self, t, p, mo, deg=None):
+        """
+        Return a contribution to the Euler factor of the motive `H_t` at a tame prime.
+
+        The output is only nontrivial when `t` has nonzero `p`-adic valuation.
+        The recipe is described in section 11.4.1 of [Watkins]_.
```
tscrim:
```suggestion
        The algorithm is described in Section 11.4.1 of [Watkins]_.
```
I am used to `Section X` being treated as a proper noun.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+
+        - ``mo`` -- integer
+
+        - ``deg`` -- integer (optional)
+
+        OUTPUT:
+
+        a polynomial
+
+        If ``deg`` is specified, the output is truncated to that degree (inclusive).
```
tscrim, line 1671: It would be good to have a doctest showing this. The `deg=1` of a degree 2 polynomial doesn't quite do this.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1629,16 +1647,117 @@ def sign(self, t, p):
...
+
+        EXAMPLES::
+
+            sage: from sage.modular.hypergeometric_motive import HypergeometricData as Hyp
+            sage: H = Hyp(cyclotomic=[[3,7],[4,5,6]])
```
tscrim:
```suggestion
            sage: H = Hyp(cyclotomic=[[3,7], [4,5,6]])
```
Since you added the spaces everywhere else.

src/sage/modular/hypergeometric_motive.py
```diff
@@ -1751,25 +1931,85 @@ def euler_factor(self, t, p, cache_p=False):
...
+            ans = P.one()
+            for m in set(j for i in self.cyclotomic_data() for j in i):
+                ans *= self.euler_factor_tame_contribution(t, p, m, deg)
+            if deg is not None:
+                ans = ans.truncate(deg+1)
+            return ans
+        # now p is good, or p is tame and t is a p-adic unit
+        elif (t-1) % p == 0:
+            typ = "mult"
+            d = (self.degree() - 1) // 2 * 2
```
tscrim: I still find it takes me a moment to realize that these operations should not simply cancel (and could follow my suggestion on the first round).

tscrim (changes requested):

fchapoton:
For a tentative to repair the currently very broken CI , see #37926

src/sage/modular/hypergeometric_motive.py
kedlaya: This is now resolved by the doctest:
```
            sage: H.euler_factor_tame_contribution(11^20, 11, 5, deg=3)
            161051*T^3 + 6171*T^2 + 121*T + 1
```

src/sage/modular/hypergeometric_motive.py
kedlaya: Evaluating first will typically yield zero over zero.

src/sage/modular/hypergeometric_motive.py
kedlaya: Maybe it's better to be clear than concise, reworked the logic.

src/sage/modular/hypergeometric_motive.py
tscrim: Ah, right, you're cancelling the zeros. I guess there isn't an easy way to evaluate it out first.

tscrim (approved):
Thank you. LGTM.

PR #37882, opened 2024-04-27
eclib: fix doctests after 20240408 update

tornaria:
With the update to 20240408 and later, the output of eclib has some extra information that makes some doctests fail. This PR adds some `...` in the right spots so the tests still pass after the update, without affecting previous versions of eclib

### :memo: Checklist


JohnCremona (approved):
Sorry about those dots ..

tornaria:
> Sorry about those dots ..

Nothing to be sorry about. The new mwrank adds output like
```
Checking saturation at [ 2 3 ]
```
which is just more diagnostic information. It's unfortunate that sagemath doctests are so sensitive to output changes.

For the record: the three dots `...` I added are just a wildcard to match anything, so the doctest works  ok on old eclib and on new eclib.

Thanks for the quick review!

PR #37883, opened 2024-04-27
gap: don't use deprecated LaTeX() and LaTeXObj()

tornaria:
These functions were removed in 4.13, but they weren't very good anyway.

This PR replaces its usage. It should work fine on old gap so this can be merged just fine.

### :memo: Checklist


tscrim (approved):
LGTM.

PR #37796, opened 2024-04-13
Revert PR with disputed dependencies

roed314:
#36964 was inappropriately merged, since two dependencies were still disputed.  This PR attempts to revert the merge.  It was created by

```
git revert -m 1 6ecb1d8532bc9b59ec03a1c7f3b1ce86c6b43cdb
```

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

### :hourglass: Dependencies

None, though it will also revert #36676 and #36951.

PR #37546, opened 2024-03-06
Sphinx ext links for Sage source files

mkoeppe:
In the documentation, we can now write `` :sage_root:`src/setup.py` `` or `` :sage_root:`src/doc/en/installation` ``, and it will be formatted uniformly, and a link to the file in the repository will be created.

- Rebased version of #33756

Fixes #33756.

### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37430, opened 2024-02-22
Refactor `bootstrap`, `bootstrap-conda`, `m4/sage_spkg_collect.m4`, `sage-spkg-info` through `sage --package properties`, `sage-get-system-packages`

mkoeppe:
Using the command `sage --package properties` added in https://github.com/sagemath/sage/pull/37018, we eliminate some direct references to `build/pkgs/` from various scripts and reduce code duplication in reading SPKG metadata.

Using other `sage --package` commands, we also simplify the code for `bootstrap -s` and `bootstrap -D`.

Review instructions: Note that there are changes to the `bootstrap`-generated file `m4/sage_spkg_configures.m4`: 
- The macro `SAGE_SPKG_COLLECT_INIT` is now called explicitly, with an argument (the full list of packages).
- Lines such as `m4_define([SPKG_INSTALL_REQUIRES_exceptiongroup], [['exceptiongroup; python_version<\"3.11\"',]])dnl` are added for use by the macro `SAGE_PYTHON_PACKAGE_CHECK` (replacing direct access to `version_requirements.txt` files) 

This is:
- Split out from #36740

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #37902, opened 2024-04-29
pkgs/sagemath-standard: Move metadata from setup.cfg to pyproject.toml

saraedum:
This PR recreates https://github.com/sagemath/sage/pull/36951 by @mkoeppe after it was reverted, see https://github.com/sagemath/sage/pull/37796.

---

Modernize the metadata. 

This is a trivial "chore" PR. It updates Python metadata to the latest format. No controversies about the current format are known about in the Python community. In a typical open source project, someone in a Maintainer role would open a PR and then immediately merge it, or when receiving such a PR from the outside, quickly review and merge it (examples: my PRs https://github.com/pytest-dev/pytest-mock/pull/410 (merged in within 1 day), https://github.com/pyodide/pyodide/pull/4472, https://github.com/pytest-dev/pytest-xdist/pull/1020,
https://github.com/sagemath/cypari2/pull/158, https://github.com/fplll/fpylll/pull/258, https://github.com/polymake/JuPyMake/pull/2, https://github.com/cvxpy/cvxpy/pull/2276, https://github.com/sagemath/cysignals/pull/177). 

~~Also fixes the `pyproject.toml` build requirements of **sagemath-standard**, broken since 10.2, hence "critical".~~

- Part of https://github.com/sagemath/sage/issues/33577
  - See also #36561

### :memo: Checklist

### :hourglass: Dependencies

- Depends on https://github.com/sagemath/sage/pull/36885 (merged here)
- Depends on #35986 (merged here to resolve merge conflict)
- Depends on #37546 (merged here to resolve merge conflict)
- Depends on #37430
- Depends on #37926


saraedum:
I'm counting the votes from #36951:

:+1: @jhpalmieri @mkoeppe @kwankyu @kiwifb

:-1: @tornaria @tobiasdiez 

(I am not the author of this PR, I didn't vote yet.)

tornaria:
I'll repeat my commment https://github.com/sagemath/sage/pull/36951#issuecomment-2041730101 since I'm not opposed to the idea of this PR, but only to its current execution.

I'm -1 on this for several reasons:
 - I think moving to `pyproject.toml` is a good idea, as it's the modern way python packages work; however, it should be `pyproject.toml` and not `pyproject.toml.m4`.
 - You need to fix build dependencies. Right now you require a lot of dependencies *just to create a sdist*.
 - This conflicts with other disputed PRs and I believe the way to move forward is to come together and figure out a reasonable compromise between all the different positions

mkoeppe:
> * I think moving to `pyproject.toml` is a good idea, as it's the modern way python packages work; however, it should be `pyproject.toml` and not `pyproject.toml.m4`.

@tornaria Once more, your demand to deviate here from our standing and documented practice is **outside of the scope of this PR**, which makes it improper by itself.

On the value and reasons for the current practice, as I explained in https://github.com/sagemath/sage/pull/36561#issuecomment-2081229564: I am the developer who has been maintaining these distribution packages in `pkgs/`. I have built and maintain the templating facility (via `m4`) because synchronizing the version information and other metadata simplifies this maintenance work; it's a carefully thought out and executed part of my plan for the modularized distributions.

> * You need to fix build dependencies. Right now you require a lot of dependencies _just to create a sdist_.

@tornaria I find it highly problematic that you are repeating this even after I explained it to you and we just had a detailed discussion on it in https://github.com/sagemath/sage/pull/37796#issuecomment-2081214321...
To summarize: **No, setuptools does not support it what you demand here, and no, this is not common practice at all.** One would need to implement one's own build backend, which either provides the wheel dependencies in a non-declarative way or you need to implement your own pyproject.toml parsing, duplicating what other backends (such as [hatch](https://hatch.pypa.io/latest/config/build/#dependencies)) allow to declare (but setuptools does not).

> * This conflicts with other disputed PRs and I believe the way to move forward is to come together and figure out a reasonable compromise between all the different positions

This PR is ready and has been ready since December. Other ideas that have been floated are neither improvements, nor are they ready to be merged (e.g., #36982). 

And nothing done here on the PR has to be the final form. Merging it is clear and overdue progress.

Finally I note that a call for "reasonable compromise" is simply a manipulation invoking the fallacy of the golden mean.

orlitzky:
I am -1 for now because I think https://github.com/sagemath/sage/pull/36982 is the better approach. But if that fails I would reconsider.

NathanDunfield:
+1: Moving from `setup.cfg.m4` to `pyproject.toml.m4` seems like a clear improvement.  Vote summary is now:

👍 @jhpalmieri @mkoeppe @kwankyu @kiwifb @NathanDunfield 
👎 @tornaria @tobiasdiez @orlitzky

culler:
I agree that generating pyproject.toml with m4 is better than generating setup.cfg with m4. It doesn't make sense to force the use of setup.cfg, and allow generating it with m4, in order to avoid generating pyproject.toml with m4.

The issue of how to generate pyproject.toml seems to me to be separate from the issue of whether tu use pyproject.toml.  Perhaps the scheme in #36982 is better than using m4.  If so then it should be possible to adopt that scheme later.

+1

mkoeppe:
I've reworked this on top of the change that made `src/pyproject.toml` a source file.

build/pkgs/conway_polynomials/version_requirements.txt
```diff
@@ -1 +1 @@
-conway-polynomials >=0.8
+conway-polynomials
```
kwankyu, line 1: may drop constraints for all packages here and below?

kwankyu:
This PR now does not seem to be the same PR which was disputed. If no one would object, I will remove the "disputed" label.

src/pyproject.toml
```diff
@@ -17,11 +17,192 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "sagemath-standard"
+description = "Sage: Open Source Mathematics Software: Standard Python Library"
+dependencies = [
+    'six >=1.15.0',
+    # From build/pkgs/sagelib/dependencies
```
tobiasdiez, line 26: These "section header" comments now make less sense and are better be replaced by proper ones (or deleted completely)

build/pkgs/cython/version_requirements.txt
```diff
@@ -0,0 +1 @@
+cython
```
tobiasdiez, line 1: The reintroduction of these version requirement files containing no actual requirements (which however are present in the pyproject.toml file) seems pretty pointless. This also reverts the design of https://github.com/sagemath/sage/pull/36982 - which already went through the "disputed" process.

Just add all packages whose version is now specified in the pyproject toml in the list in `bootstrap`.

tobiasdiez (changes requested):

tobiasdiez:
> This PR now does not seem to be the same PR which was disputed. If no one would object, I will remove the "disputed" label.

This PR reverts some of the changes made in https://github.com/sagemath/sage/pull/36982, because Matthias doesn't like the design of https://github.com/sagemath/sage/pull/36982 - which he expressed there of course already). Apart from these changes this PR looks good to me and I happily remove the "disputed" label once this is fixed.

build/pkgs/cython/version_requirements.txt
mkoeppe: > seems pretty pointless.

No. This defines the map between SPKG names and Python distribution names.

src/pyproject.toml
mkoeppe: They make full sense.

build/pkgs/cython/version_requirements.txt
mkoeppe: > Just add all packages whose version is now specified in the pyproject toml in the list in `bootstrap`.

No.

kwankyu: > ... This also reverts the design of #36982 - which already went through the "disputed" process.

As I understand it, by the design of #36982:

`src/pyproject.toml` specifies version constraints of the dependencies of the distribution package `sagemath-standard`, independently from the sage-distribution.

This PR respects that, and also reintroduces `build/pkgs/*/version_requirements.txt` files to specify further version constraints of the sage packages for the sage-distribution. 

@mkoeppe Am I right?

kwankyu:
I removed the "disputed" label since we are doing the normal review process.

When it is clear that there are still disagreements that cannot be resolved, we can add the "disputed" label.

Under the "disputed" label, we are supposed to do the voting instead.

build/pkgs/cython/version_requirements.txt
mkoeppe: @kwankyu That's correct.

build/pkgs/fpylll/version_requirements.txt
```diff
@@ -1 +1 @@
-fpylll >=0.5.9
+fpylll
```
kwankyu, line 1: So the constraint ">= 0.5.9" is needed for sagemath-standard distribution package. For sage-the-distribution, there is no further constraint. Hence it is just "fpylll" here. Right?

Is there a sage package for which sage-the-distribution needs more constraint than sagemath-standard? It seems there is none, currently...

src/pyproject.toml
```diff
@@ -18,11 +18,192 @@ requires = [
...
+    # According to https://github.com/python/typing_extensions/blob/main/CHANGELOG.md,
+    # version 4.4.0 adds another Python 3.11 typing backport
+    'typing_extensions >= 4.4.0; python_version<"3.11"',
+    # From Makefile.in: SAGERUNTIME
+    'ipython >=7.13.0',
+    'pexpect >=4.8.0',
+    # From Makefile.in: DOC_DEPENDENCIES
+    'sphinx >=7.4.7, <9'
+    'networkx >=2.4',
+    # 1.8 is known good version.
```
kwankyu, line 55: Missing comma in L53.

kwankyu:
The PR branch works well.

src/pyproject.toml
mkoeppe: Thanks, fixed in 443b5bb483c

build/pkgs/fpylll/version_requirements.txt
mkoeppe: As I said in https://groups.google.com/g/sage-devel/c/sVcJiJdUxRs/m/VMOu20agBgAJ, I don't plan to start maintaining two sets of constraints. (Setting the version constraint for the packages that are dependencies of sagelib in the `pyproject.toml` file instead of in the `build/pkgs/*/version_requirements.txt` file literally follows the design in Tobias's proof-of-concept PR #36982.)

kwankyu: > ... Setting the version constraint for the packages that are dependencies of sagelib in the `pyproject.toml` file instead of in the `build/pkgs/*/version_requirements.txt` file literally follows the design in Tobias's proof-of-concept PR #36982.)

You mean `src/pyproject.toml` file? It is supposed to maintain version constrains for the dependencies of sagelib. This is what Tobias' PR did? This PR does not revert that. No?

Now this PR makes `build/pkgs/*/version_requirements.txt`  maintain (additional) version constrains for the sage-distribution. No?

kwankyu: Two sets of constraints (those in `src/pyproject.toml` and those in `build/pkgs/*/version_requirements.txt `) serve different purposes. I think this is reasonable. Or I am misunderstanding things...

mkoeppe: > You mean `src/pyproject.toml` file? 

Yes.

> It is supposed to maintain version constrains for the dependencies of sagelib. This is what Tobias' PR did?

Consider a package A that is a build dependency of sagelib.
- Before Tobias's PR, the version constraints on A (for sagelib; for the modularized distributions; and for the Sage distribution in "configure --enable-system-site-packages" mode) were set in `build/pkgs/A/version_requirements.txt`.
- After Tobias's PR, the version constraints on A (for the same 3 purposes) are set in `src/pyproject.toml`, and `build/pkgs/A/version_requirements.txt` is no longer a source file.
- With the present PR, the file `build/pkgs/A/version_requirements.txt` becomes again a source file, but without actual version constraints.
- In principle, the file _could_ be used for additional version constraints on A, but I do not plan to use this mechanism. (Note that any version constraints there will take effect not only on the Sage distribution with "configure --enable-system-site-packages", but also for the modularized distributions.)

Now consider a package B that is not a build dependency of sagelib but a runtime dependency of sagelib.
- Tobias's PR made no changes regarding these packages.
- With the present PR, version constraints are applied in the same way as for package A.

kwankyu: Let

(L) version constraints for sagelib (= sagemath-standard)
(M) version constraints for modularized distributions 
(D) version constraints for the Sage distribution in "configure --enable-system-site-packages" mode

For package A, is this 

(D) >= (meaning more or the same constraints) (M) >= (L)

true?

or maybe (L) = (M)...

mkoeppe: There's really no good reason for (M) >= (L); it should rather be (L) >= (M) if anyone really wanted to maintain things on this level of detail.

kwankyu: OK. Then (L) = (M) in practice.

kwankyu: So we need to maintain two sets of constraints wherever they are.

mkoeppe: It's best to view the `version_requirements.txt` file in this case as simply providing the mapping between the SPKG name and the Python distribution name.

kwankyu: * In principle, the file [`build/pkgs/A/version_requirements.txt`] _could_ be used for additional version constraints on A, but I do not plan to use this mechanism. (Note that any version constraints there will take effect not only on the Sage distribution with "configure --enable-system-site-packages", but also for the modularized distributions.)

Now I see the dilemma.

kwankyu: > * In principle, the file [`build/pkgs/A/version_requirements.txt`] _could_ be used for additional version constraints on A, ...

The filename `version_requirements` suggests that the file is for ... version constraints. There seems to be nothing wrong with the use. It is the right place for version constraints necessary for sage-the-distribution.

> ... Note that any version constraints there will take effect not only on the Sage distribution with "configure --enable-system-site-packages", but also for the modularized distributions.

We may introduce a new syntax into "version_requirements.txt". For example
```
cython [<=3.0.2]
```
Then sage-bootstrap uses the bracketed constraints for sage-the-distribution. For modularized distributions, the bracketed part is ignored.

tobiasdiez: > OK. Then (L) = (M) in practice.

I don't think this is true. For example, the usage of sphinx in sage-lib is very minimal (essential only https://github.com/sagemath/sage/blob//src/sage/misc/sphinxify.py) and thus is compatible with a wide range of sphinx version's. However, building the actual documentation is very much tied to a specific sphinx version due to the deep autodoc integration; hence a very narrow version constraint is needed for the sage-doc package. Trying to homogenize these completely different use cases and version constraints is not purposeful in my opinion.

> It's best to view the `version_requirements.txt` file in this case as simply providing the mapping between the SPKG name and the Python distribution name.

Using `version_requirements.txt` files for this purpose seems wrong. These files were renamed to end on `requirements` so that standard tools pick them up. Its a misuse to use them for a sage-the-distro specific mapping. Why is such a mapping needed in the first place? Just rename the SPKG to match the python distro name, and be done.

kwankyu:
@tobiasdiez Please specify what needs to be worked on when you add "needs work" label, so that other people can see what is going on with the PR.

tobiasdiez:
> @tobiasdiez Please specify what needs to be worked on when you add "needs work" label, so that other people can see what is going on with the PR.

There are a few unresolved reviewer remarks above as comments directly against the changed files. That should be sufficient, or not?

src/pyproject.toml
tobiasdiez: Could you please expand on this? For example, build/pkgs/sagelib is downstream of sagelib in src; why would then the later refer to the former and not the other way around?

kwankyu:
No. Unresolved remarks just need more discussion.

src/pyproject.toml
mkoeppe: The comments explain where the dependencies are coming from; the audience for these explanations: people familiar with the build system.

build/pkgs/fpylll/version_requirements.txt
mkoeppe: > Using `version_requirements.txt` files for this purpose seems wrong. These files were renamed to end on `requirements` so that standard tools pick them up. Its a misuse to use them for a sage-the-distro specific mapping. 

No, it's not a "misuse". It's the documented design.

> Why is such a mapping needed in the first place? Just rename the SPKG to match the python distro name, and be done.

This does not work because the names sometimes clash with non-Python packages of the same name.

And, of course, it's beyond the scope of this PR to rename packages.

mkoeppe: > I don't think this is true. For example, the usage of sphinx in sage-lib is very minimal (essential only https://github.com/sagemath/sage/blob//src/sage/misc/sphinxify.py) and thus is compatible with a wide range of sphinx version's. [...]

But which ones? We have no mechanism to even test this, as we do basically all of our testing with the reference distribution (= the Sage distribution).

> Trying to homogenize these completely different use cases and version constraints is not purposeful in my opinion.

We are not "trying" to do this. We already do, and we do this by design to simplify maintenance. 

On the other hand, trying to manage these constraints in a very fine-grained way is extra work that I do not plan to take on, and no one else has stepped up to do this.

mkoeppe: > We may introduce a new syntax into "version_requirements.txt"

No, we cannot. These files must use the standard format of requirements files. Note we renamed them from the previous "install-requires.txt" because it is picked up by GitHub for SBOM purposes.

kwankyu: Then we may insert the extra info as comments:
```
cython  # <= 3.0.2 for distribution
```
or
```
cython  # [<= 3.0.2]
```
if there is no better solution. 

As we already have a case, it is likely that we always want more constraints for sage-the-distribution than for modularized distributions.

kwankyu: > On the other hand, trying to manage these constraints in a very fine-grained way is extra work that I do not plan to take on, and no one else has stepped up to do this.

I agree. Maintaining version constraints separately for all (there would be many) modularized distributions would be a hell. 

If a modularized distribution is so special, then it may have its own pyproject.toml (without .m4 suffix)...

mkoeppe: > If a modularized distribution is so special, then it may have its own pyproject.toml (without .m4 suffix)...

This can also be done in `pyproject.toml.m4` by just not using the macro that inserts the centralized dependency.

mkoeppe: > As we already have a case

Which one? I don't think we have any.

src/pyproject.toml
kwankyu: When pyproject.toml was generated from pyproject.toml.m4, it was clear where the dependencies info came from. Now that it is not, it is relevant to explain the source of the info.

Out of scope of this PR: we may restore pyproject.toml.m4 but also keep the generated pyproject.toml in repo. Then we may remove the comment.

build/pkgs/fpylll/version_requirements.txt
kwankyu:
```
    # per https://github.com/scipy/scipy/blob/maintenance/1.13.x/pyproject.toml
    'cython >=3.0.8,<3.1.0',
```
No?

kwankyu: Or according to https://groups.google.com/g/sage-devel/c/sVcJiJdUxRs/m/PZr7QjiQBgAJ ...

mkoeppe: Sure, these are constraints that we are setting, uniformly for easier maintainability.

What I mean is that we *don't* have a case for is any example where a modularized distribution really would benefit from a wider version range than we set centrally.

kwankyu: I don't know about "really"... In principle, wider version range is better.

mkoeppe: Of course. But without testing the wider range, it's just gambling.

kwankyu: This discussion is not necessary if we assume (D) = (M).

mkoeppe: Either way, all of this discussion is far outside the scope of the PR and far in speculative territory

kwankyu: OK.

kwankyu (approved):
LGTM.

tobiasdiez:
> No. Unresolved remarks just need more discussion.

Then please don't set it to positive review. I've said above that the reintroduction of committed version_requirements files is not okay with me; there is also no need to change this as part of a PR that aims to move metadata from setup.cfg to pyproject.toml. There is already a framework in place on how to specify infos about package dependencies in that file; just reuse exactly the same mechanism and create the version requirements files during bootstrap (and maybee improve it in the a follow up PR).

src/pyproject.toml
tobiasdiez: > When pyproject.toml was generated from pyproject.toml.m4, it was clear where the dependencies info came from. Now that it is not, it is relevant to explain the source of the info.

The dependencies are coming from the usage of these packages in sage-lib. If you want to specify the source, then you should mention the relevant subpackages. 

Now that pyproject.toml is the source for depenendency info, the downstream file (in build pkgs) should refer it it; not the other way around

build/pkgs/fpylll/version_requirements.txt
tobiasdiez: > > Using `version_requirements.txt` files for this purpose seems wrong. These files were renamed to end on `requirements` so that standard tools pick them up. Its a misuse to use them for a sage-the-distro specific mapping.
> 
> No, it's not a "misuse". It's the documented design.
> 
> > Why is such a mapping needed in the first place? Just rename the SPKG to match the python distro name, and be done.
> 
> This does not work because the names sometimes clash with non-Python packages of the same name.
> 

For which packages is this the case?

kwankyu:
I don't think this PR reverts the main of point of your PR #36982. The file `src/pyproject.toml`, with the dependencies info for the sage library (sagemath-standard), is still there.

This PR restores `version_requirements.txt` files, to establish the mapping between spkg names and pip package names, as Matthias explains. That info is not available from `src/pyproject.toml`. In this sense, this PR fixes a "bug" introduced by your PR #36982.

There was a need for additional version constraints (on cython, for example) for sage-the-distribution. I expected the `version_requirements.txt` file could be used for that purpose. But Matthias says that `src/pyproject.toml` serves the purpose, and  provisioning a mechanism for additional set of version constraints is out of scope of this PR. I agree with that.

Thus I am positive with this PR and added "positive review". 

In the discussions of you and Matthias, I see no items that you raised but Matthias didn't respond. It seemed that there is no item that Matthias will work on. It seems still so.

Now that you added "needs work" again with no new items to discuss with Matthias, this PR is qualified for "disputed PR". After waiting for Matthias' response to your "needs work" label, I will add "disputed" label to this PR.

tobiasdiez:
> This PR restores `version_requirements.txt` files, to establish the mapping between spkg names and pip package names, as Matthias explains. That info is not available from `src/pyproject.toml`. In this sense, this PR fixes a "bug" introduced by your PR #36982.

Per #36982 this map is defined in sage bootstrap, and is working well (e.g. the constraint for the python package cypari2 is correctly mapped to the sage package cypari - which by the way is the only package with a different slightly different name that I'm aware of [apart from some `-` vs `_`)). Which bugs did you experience?

In either way, changing the way the build dependencies are handled is out-of-scope of this PR, since this information is already in pyproject.toml and no longer in `setup.cfg`.

kwankyu:
> > This PR restores `version_requirements.txt` files, to establish the mapping between spkg names and pip package names, as Matthias explains. That info is not available from `src/pyproject.toml`. In this sense, this PR fixes a "bug" introduced by your PR #36982.
> 
> Per #36982 this map is defined in sage bootstrap, and is working well (e.g. the constraint for the python package cypari2 is correctly mapped to the sage package cypari - which by the way is the only package with a different slightly different name that I'm aware of [apart from some `-` vs `_`)). 

Matthias makes the mapping explicit using `version_requirements.txt` files. Hardcoding the mapping in python code is the "bug" I am saying.

> In either way, changing the way the build dependencies are handled is out-of-scope of this PR, since this information is already in pyproject.toml and no longer in `setup.cfg`.

The scope of a PR is set by the PR author. Neither you nor me. You may request the author to make the scope explicit in the PR description.

On the other hand, you may object to this PR on whatever ground.

build/pkgs/fpylll/version_requirements.txt
dimpase: non-python packages can be renamed if needed, this is not a problem

mkoeppe:
> the sage package cypari - which by the way is the only package with a different slightly different name that I'm aware of [apart from some `-` vs `_`)).

Wrong.

mkoeppe:
@kwankyu explains it all correctly.

build/pkgs/fpylll/version_requirements.txt
dimpase: one reason for name clashed between Python and non-Python packages is lack of a directory structure in build/pkgs/, everything is dumped into one flat namespace.

And no, don't tell me that a distribution XXX does this too, this is  irrelevant, and it suffices to see that such an XXX is forced to deal with the name clashes like this by inventing package name prefixes, which is obviously just a poor man replacement of a proper directory structure.

kwankyu:
+1 from me.

dimpase:
A very fat -1 from me. All sorts of wrong reasons are given here to get this PR through, such as

*  mkoeppe's "I am not going to maintain 2 sets of version constraints" - as if mkoeppe is the sole project owner
* flat refusal to consider normal ways to deal with package name clashes 

etc.

mkoeppe:
"to consider normal ways to deal with package name clashes" -- obviously outside of the scope of this PR.

build/pkgs/fpylll/version_requirements.txt
mkoeppe: Obviously outside of the scope of this PR to do any of this.

mkoeppe:
Closing in favor of:
- #38714

src/pyproject.toml
mkoeppe: More demands that are obviously outside of the scope of this PR.

tobiasdiez:
I've now cherry-picked the uncontroversial parts of this PR (or more specifically of #38714) to https://github.com/sagemath/sage/pull/38727. I think nobody objected that the metadata moved from the old `setup.cfg` to the modern `pyproject.toml`.

PR #37903, opened 2024-04-29
correct URL for symengine_py

dimpase:
Correct URL for the symengine_py package.

This is a 1-char typo made in https://github.com/sagemath/sage/pull/36677


mkoeppe:
duplicate

PR #37904, opened 2024-04-29
Change `SetSystem` representation

gmou3:
This is an attempt to represent `SetSystem` in a more informative and user-friendly way.

Previous:
```
sage: M = matroids.CompleteGraphic(7)
sage: M.bases()
Iterator over a system of subsets
```
Current:
```
sage: M = matroids.CompleteGraphic(7)
sage: M.bases()
SetSystem of 16807 sets over 21 elements
```

I think this is an improvement from resorting to calling, e.g., `len`, in order to to get some info, after you get struck with `Iterator over a system of subsets`.
Note that a `SetSystem` is not an iterator but simply an iterable (so `__repr__` is also misleading). The previous developers were planning to make it into an _actual_ iterator, as can be seen on [L66](https://github.com/sagemath/sage/blob/f1dc325cf01ac542d42257b77aa322cd92e74de8/src/sage/matroids/set_system.pyx#L66) of `set_system.pyx`.

### DIFF
--- a/src/sage/matroids/basis_exchange_matroid.pyx
+++ b/src/sage/matroids/basis_exchange_matroid.pyx
@@ -1923,7 +1923,7 @@ cdef class BasisExchangeMatroid(Matroid):
 
             sage: M = matroids.catalog.N1()
             sage: M._characteristic_setsystem()
-            Iterator over a system of subsets
+            SetSystem of 23 sets over 10 elements
             sage: len(M._characteristic_setsystem())
             23
         """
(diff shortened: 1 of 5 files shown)


tscrim:
This is an improvement given the current implementation. If it does become an actual iterator (or a "view" like `Python_dict.keys()`), it might know some additional information about its size, At that time, we can revisit the repr output.

tscrim:
@gmou3 If this is ready for review, just let me know. I am ready to set a positive review if ready.

gmou3:
Yeap, it is ready I think. Thanks for the review.

tscrim (approved):
Thanks.

PR #37916, opened 2024-05-01
structure/coerce actions

mantepse:
We modify `ModuleAction` to prefer coercion of the base ring over creating an action.  More precisely, before this branch, we have

```
sage: G = PolynomialRing(QQ, "x")
sage: S = PolynomialRing(MatrixSpace(QQ, 2), "x")
sage: G.gen() * S.gen()
[x 0]
[0 x]*x
```
instead of
```
            [1 0]
            [0 1]*x^2
```
and
```
sage: G = PolynomialRing(QQ, "x")
sage: S = PolynomialRing(InfinitePolynomialRing(QQ, "a"), "x")
sage: G.gen() * S.an_element()
x*x
```
instead of
```
x^2
```
which is at least surprising.

In the end, this is needed to make #37033 work.

Authors: @tscrim and @mantepse.

### DIFF
--- a/src/sage/categories/pushout.py
+++ b/src/sage/categories/pushout.py
@@ -35,7 +35,7 @@ lazy_import('sage.categories.commutative_additive_groups', 'CommutativeAdditiveG
 lazy_import('sage.categories.commutative_rings', 'CommutativeRings')
 lazy_import('sage.categories.groups', 'Groups')
 lazy_import('sage.categories.objects', 'Objects')
-lazy_import('sage.categories.rings', 'Rings', at_startup=True)
+lazy_import('sage.categories.rings', 'Rings')
 
 # TODO, think through the rankings, and override pushout where necessary.
 
--- a/src/sage/structure/coerce_actions.pyx
+++ b/src/sage/structure/coerce_actions.pyx
@@ -313,6 +313,23 @@ cdef class ModuleAction(Action):
             sage: 1/S.0
             1/x
 
+        If there is a coercion from ``G`` to ``S``, we do not create
+        the module action of ``G`` on the pushout of ``G`` and ``S``::
+
+            sage: G = PolynomialRing(QQ, "x")
+            sage: S = PolynomialRing(MatrixSpace(QQ, 2), "x")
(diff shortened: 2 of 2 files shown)


nbruin:
You might want to be careful with a change like that. Mathematically, for instance, a ZZ-algebra with unity is a ZZ-module with extra structure, so the action of ZZ on the algebra is more fundamental than the multiplication structure on the algebra itself. So mathematically, it makes sense to prefer the ZZ-action over coercion from ZZ into the algebra followed by multiplication.

For performance, this can also be a difference: multiplying by an integer (i.e., a scalar multiplication) is really quite a bit cheaper than using full algebra multiplication with an element that happens to be a coerced scalar.

Thank about the situation where you have  the full n x n matrix algebra over a field k. Then k coerces into the algebra as diagonal matrices. Would your change prefer that coercion over the scalar multiplication action? Multiplying two n x n matrices is significantly more expensive than a scalar multiplication of a matrix.

Note that the matrix algebra is simply an example. Perhaps there things are arranged in such a way that scalar multiplication gets recognized early or in a different way. But the principle applies in many cases. It may well be that your change is fine, but you should carefully review that it is, certainly for the reason above.

tscrim:
Martin’s description is oversimplified and missing a key component. We do **not** prefer coercion over *already defined* for the reason you stated. However, we instead prefer doing coercion over doing a pushout of the base ring.

tscrim:
The situation at hand is this. Suppose there is a coercion $\phi \colon R \to S$, where $S$ is an $F$-algebra. What the pushout construction does in $r \cdot s$ (via `r * s`) is create $S’$ as an extension of scalars of $S$ to an $R$-algebra. However, the coercion map should mean instead treat $r$ as $\phi(r) \in S$ to make sense of the multiplication instead of the natural $R$-action on $S’$. This agrees with what `common_parent` would find too.

nbruin:
Oh thanks. If you put it that way it sounds quite natural to do.

tscrim (approved):
Thank you. Than I will approve this.

Issue #37437, opened 2024-02-23
RSK on symmetric group element kills the kernel


yeetkode:
### Steps To Reproduce

2. Create an element of `SymmetricGroup(n)`, say `P`.
3. Call the function `RSK(P)`.

### Expected Behavior

ValueError: `P` is not one of the types supported by `RSK`, or some similar way of declaring an error.

### Actual Behavior

The kernel dies.

### Environment

```markdown
- **OS**: MacOS 14 Sonoma
- **Hardware**: M2 Silicon Chip MacBook Air 13'
- **Sage Version**: 10.0
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

mezzarobba:
Confirmed:
```
sage: RSK(SymmetricGroup(2).an_element())
[...]
#8  0x00007f1d9a95fd1d in __Pyx_PyObject_GetAttrStr (
    attr_name=<optimized out>, obj=<unknown at remote 0x7f1d947de700>)
    at sage/groups/perm_gps/permgroup_element.c:31049
        tp = <optimized out>
        tp = <optimized out>
#9  __pyx_pf_4sage_6groups_8perm_gps_17permgroup_element_23PermutationGroupElement_42__mul__ (
    __pyx_v_right=<sage.groups.perm_gps.permgroup_element.SymmetricGroupElement at remote 0x7f1d91717d00>, __pyx_v_left=[<unknown at remote 0xa4c3e9>])
    at sage/groups/perm_gps/permgroup_element.c:20169
        __pyx_t_2 = <optimized out>
        __pyx_t_7 = <optimized out>
        __pyx_lineno = 0
        __pyx_filename = 0x0
        __pyx_v_pl = 0x7f1d90c51c00
        __pyx_v_prod = 0x0
        __pyx_t_3 = 0x0
        __pyx_t_8 = <optimized out>
        __pyx_L1_error = <optimized out>
        __pyx_v_pr = 0x7f1d91717d00
        __pyx_v_Pleft = 0x7f1d947de700
        __pyx_t_4 = 0x0
        __pyx_v_i = <optimized out>
        __pyx_r = 0x0
        __pyx_t_5 = 0x0
        __pyx_clineno = 0
        __pyx_v_Pright = 0x7f1d94982990
        __pyx_t_1 = <optimized out>
        __pyx_t_6 = <optimized out>
[...]

PR #37918, opened 2024-05-01
structure/detect element action

mantepse:
We fix a typo in the documentation and a bug in the implementation of `structure.coerce_actions.detect_element_action`.

Previously, we had
```
sage: detect_element_action(S3, R, False)
Right action by Symmetric group of order 3! as a permutation group on Multivariate Polynomial Ring in x, y, z over Rational Field
```
which is wrong, because `detect_element_action` *should* (despite its docstring) return an action of the *second* argument on the *first*.

Authors: @tscrim, @mantepse


tscrim:
Interesting...The failure in `categories/action.pyx` is trivial; just need to swap the input order.

Now the failures in Lie algebras, that is because I didn't implement the `_act_on_` correctly for the PBW basis elements. This change fixes those failures:
```diff
diff --git a/src/sage/algebras/lie_algebras/poincare_birkhoff_witt.py b/src/sage/algebras/lie_algebras/poincare_birkhoff_witt.py
index 3b59303d0d..bb808d1c2d 100644
--- a/src/sage/algebras/lie_algebras/poincare_birkhoff_witt.py
+++ b/src/sage/algebras/lie_algebras/poincare_birkhoff_witt.py
@@ -609,24 +609,24 @@ class PoincareBirkhoffWittBasis(CombinatorialFreeModule):
                 return ret
             cm = get_coercion_model()
             L = self.parent()._g
-            if self_on_left:
-                if cm.discover_action(L, x.parent(), mul):
-                    ret = x.parent().zero()
+            X = x.parent()
+            action = X.get_action(L, self_on_left=not self_on_left)
+            if action:
+                assert action.actor() is L
+                ret = X.zero()
+                if self_on_left:
                     for mon, coeff in self._monomial_coefficients.items():
                         term = coeff * x
                         for k, exp in reversed(mon._sorted_items()):
                             for _ in range(exp):
                                 term = L.monomial(k) * term
                         ret += term
-                    return ret
-            else:
-                if cm.discover_action(x.parent(), L, mul):
-                    ret = x.parent().zero()
+                else:
                     for mon, coeff in self._monomial_coefficients.items():
-                        term = coeff * x
+                        term = x * coeff
                         for k, exp in reversed(mon._sorted_items()):
                             for _ in range(exp):
                                 term = term * L.monomial(k)
                         ret += term
-                    return ret
+                return ret
             return None
```

While investigating this, I also noticed that the documentation of `Parent.get_action()` is **also wrong**!! It is only checking if there is an action of `S` on `self`!

tscrim:
So the documentation for `get_action()` is not completely wrong as `_get_action_` can return an action going either way, which is reasonable. However, beyond that, the discovery process is one sided.

For the `argument_groups.py` failure, I would say that is because the testing is only one sided. We probably should expand that to do what the documentation says (so it covers, e.g., `int`'s). If we do this, then the `assert` in the diff above will need to become part of the `if` check.

I don't understand the change in output for the isometry group test. It is possibly related to the code not following the documentation of `get_action()`.

There is a small possibility that this could fix the segfault noticed  #37437, but I am very skeptical that it will.

tscrim:
The `src/sage/modules/free_quadratic_module_integer_symmetric.py` failure should be the same as for `matrix_gps/isometries.py`.

mantepse:
Shall I do the changes and push them here?

mantepse:
> Interesting...The failure in `categories/action.pyx` is trivial; just need to swap the input order.

Sorry, I don't know what you mean here.

mantepse:
> > > Interesting...The failure in `categories/action.pyx` is trivial; just need to swap the input order.
> > 
> > 
> > Sorry, I don't know what you mean here.
> 
> If you look at the failure and what the intended output should be, I think it is clear how to update the doctest (input).

I have, with this branch:
```
sage: cm = get_coercion_model()
sage: cm.get_action(ZZ, list, operator.mul)
sage: cm.get_action(list, ZZ, operator.mul)
```
that is, in both cases the output is `None`.

If I understand the doctring of `cm.get_action` correctly, it should be somewhat symmetric in `R` and `S` - although, for
```
sage: A = cm.get_action(QQx, ZZ, operator.truediv); A
```
it isn't.

tscrim:
It looked trivial, but perhaps it is a bit more complicated than I thought. It might come down to the requirement that both `X` and `Y` are now required to be `Parent` subclasses, which we probably want to weaken. I will try to look at this soon.

src/sage/structure/coerce_actions.pyx
```diff
@@ -224,15 +241,15 @@ def detect_element_action(Parent X, Y, bint X_on_left, X_el=None, Y_el=None):
...
-        if x._act_on_(y, X_on_left) is not None:
-            return ActOnAction(X, Y, X_on_left, False)
-    except CoercionException:
-        _record_exception()
-
-    # element x defining _acted_upon_
     if isinstance(Y, Parent):
+        # element y defining _act_on_
+        try:
+            if y._act_on_(x, not X_on_left) is not None:
```
mantepse, line 246: If we follow the docstring `as defined by elements of X`, this looks wrong.  Indeed, in the old code, we were only looking at methods of `x`.

src/sage/structure/coerce_actions.pyx
```diff
@@ -224,15 +241,15 @@ def detect_element_action(Parent X, Y, bint X_on_left, X_el=None, Y_el=None):
         except CoercionException as msg:
             _record_exception()
 
-    # element x defining _act_on_
-    try:
-        if x._act_on_(y, X_on_left) is not None:
-            return ActOnAction(X, Y, X_on_left, False)
```
mantepse, line 229: I 'traced' (using `print`)
```
sage: cm = get_coercion_model()
sage: c = cm.get_action(ZZ, list, operator.mul)
```
and it is line 230 that returns
```
Left action by Integer Ring on <class 'list'>
```

src/sage/structure/coerce_actions.pyx
tscrim: I don't quite think it is saying that it defined by the methods of `x` (the action is defined by `x` using the method of `y` is how I read this), but you can read it that way. Although I don't think the docstring is accurate in this regard either.

src/sage/structure/coerce_actions.pyx
tscrim: Yea, I think we are just being too strict with making sure `Y` is a parent. We need to allow that to not be a `Parent`, and maybe need to catch an error (or fix some other code). Not 100% sure at this point.

mantepse: I think, to be of any help I would need to know precisely what `detect_element_action` is supposed to be doing.  And, I guess, that's part of the problem, right?

tscrim: Indeed, we are kind of guessing at its purpose based on its code, its doc, and how it is used. Unfortunately there doesn’t seem to be any consistency with these three. @nbruin Do you have any recollections, thoughts, or opinions on this?

nbruin: I'm not sure which three routines are in play here. I don't think we can have an action of integers on a list, if that's what you're aiming for. Lists are way to coarse a "parent" to determine actions for. There may also be the problem that non-parents probably fail to have the right slots/attributes to properly participate in the coercion system. And there's the conflict  that `3*[1,2,3]` and `[1,2,3]*3` already means "repeated concatenation" in python.

There was recently a discussion on action discovery concerning `int` vs. `Integer` where `Parent` participating in actions also came up: https://groups.google.com/g/sage-devel/c/RvmljIs7wBI/
and in a different form in this relatively recent PR: https://github.com/sagemath/sage/pull/37369
(which  in the end did end up with a patch so that action discovery for int on ZZ-modules should also work!)

mantepse:
I just checked: the current branch does *not* fix #37976, but I think it should.

Using develop or this branch both give:
```
sage: v = vector([h[3]+h[2,1]])
sage: P = v.parent()
sage: from sage.structure.coerce_actions import detect_element_action
sage: a = detect_element_action(P, ZZ, True, v, -22)
sage: a
Right scalar multiplication by Integer Ring on Ambient free module of rank 1 over the integral domain Symmetric Functions over Rational Field in the homogeneous basis
sage: a.op
<built-in function mul>
```

PR #37923, opened 2024-05-02
Extend Drinfeld modules to support fields constructed as univariate quotients

ymusleh:
This PR implements some modifications to permit Drinfeld modules to be constructed over fields constructed as univariate quotient rings. Previously, the implementation required that the fields be constructed as `FiniteField` objects as in the following example:

```
sage: Fq = GF(4)
sage: A.<T> = Fq[]
sage: K.<z> = Fq.extension(2); K
Finite Field in z of size 2^4
```

This PR now allows Drinfeld modules to be built over fields that are built explicitly as univariate quotient rings, as in the example below.

```
sage: Fq = GF(4)
sage: A.<T> = Fq[]
sage: ip = A.irreducible_element(2)
sage: K.<z> = Fq.extension(ip); K
Univariate Quotient Polynomial Ring in z over Finite Field in z2 of size 2^2 with modulus z^2 + (z2 + 1)*z + z2 + 1
```

This features two main changes:

- the `in_base` method in `RingExtensionElement` has been extended to explicitly handle Univariate Quotient Rings.
- `FiniteDrinfeldModule` now keeps a private function for computing the frobenius action on field elements.

In addition, the caching behaviour of `frobenius_charpoly` has been modified to depend on the parameter `use_cache`.

@xcaruso @kryzar @DavidAyotte 

### :memo: Checklist


src/sage/rings/ring_extension_element.pyx
```diff
@@ -354,7 +379,12 @@ cdef class RingExtensionElement(CommutativeAlgebraElement):
             f = parent._backend_defining_morphism
             base = f.domain()
             ring = f.codomain()
+            parent_ring = self._backend.parent()
             if ring.has_coerce_map_from(base) and are_equal_morphisms(f, None):
+                if is_PolynomialQuotientRing(parent_ring):
+                    if base.has_coerce_map_from(parent_ring.base_ring()):
+                        if self._backend.lift().degree() <= 0:
+                            return parent.base()(self._backend.list()[0])
```
xcaruso, line 387: I would say that it is not the correct place for this piece of code.

You should instead make the following work:
```
sage: Fq.<a> = GF(5^2)
sage: A.<T> = Fq[]
sage: K.<b> = Fq.extension(T^2 + a)
sage: Fq(K(a))
...
TypeError: unable to coerce <class 'sage.rings.polynomial.polynomial_quotient_ring.PolynomialQuotientRing_field_with_category.element_class'>
```

ymusleh: While that type of coercion would be nice to have, the one we want in the Drinfeld module implementation is a bit different since we need to coerce an element from a ring extension over a base to a base element, so it would be something along the lines of 
```
sage: Fq.<a> = GF(5^2)
sage: A.<T> = Fq[]
sage: K.<b> = Fq.extension(T^2 + a)
sage: L = K.over(Fq)
sage: Fq(L(a))
...
TypeError: unable to coerce <class 'sage.rings.ring_extension_element.RingExtensionElement'>
```
This is necessary for `FiniteDrinfeldModule` methods like `invert` which need to map coefficients into $\mathbb{F}_q$, and that is already handled by the `in_base` method for `FiniteField` classes. It seems to me a bit odd to force a user to have to decide between these two different ways of mapping into the base for ring extensions depending on the type of the extension ring. One alternative option would be to implement that coercion more generally and replace the use of `in_base` to these explicit coercions. Another could be to not use RingExtensions for Drinfeld modules over univariate quotients, but I think that would likely be much more complicated to support.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -152,12 +152,37 @@ def __init__(self, gen, category):
...
         # have _frobenius_norm and _frobenius_trace attributes.
         super().__init__(gen, category)
-        self._base_degree_over_constants = self.base_over_constants_field().degree(self._Fq)
+        K = self.base_over_constants_field().backend()
+        self._over_quotient_ring = is_PolynomialQuotientRing(K)
+        char, q = self._Fq.characteristic(), self._Fq.cardinality()
+        qdeg = logb(q, char)
+        # Currently only support univariate quotient rings with coefficients
+        # in the base field Fq
+        if self._over_quotient_ring:
```
kryzar, line 177: I think one should not create a useless attribute (in this case, `self._over_quotient_ring`). Here, I believe it is only used once, so the benefits are minimal but it costs in maintenance. Better to delete it now.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -152,12 +152,37 @@ def __init__(self, gen, category):
             sage: ore_polring = phi.ore_polring()
             sage: phi._gen == ore_polring(gen)
             True
+
+            sage: Fq = GF(25)
+            sage: A.<T> = Fq[]
+            sage: z2 = Fq.gen()
+            sage: ip = A(T^6 + (z2 + 2)*T^5 + (2*z2 + 1)*T^4 + (4*z2 + 3)*T^3 + (z2 + 1)*T^2 + z2*T + 2)
```
kryzar, line 159: I think SageMath uses the naming convention `modulus` rather than `ip` but @xcaruso might correct me on this.

src/sage/rings/ring_extension_element.pyx
```diff
@@ -342,6 +343,30 @@ cdef class RingExtensionElement(CommutativeAlgebraElement):
...
+            sage: Fq = GF(25)
+            sage: A.<T> = Fq[]
+            sage: z2 = Fq.gen()
+            sage: ip = A(T^6 + (z2 + 2)*T^5 + (2*z2 + 1)*T^4 + (4*z2 + 3)*T^3 + (z2 + 1)*T^2 + z2*T + 2)
+            sage: K.<z12> = Fq.extension(ip)
+            sage: gen = [K.random_element() for _ in range(3)]
+            sage: phi = DrinfeldModule(A, gen)
+            sage: a = A(3*T^2 + 4*z2*T)
+            sage: phi.invert(phi(a)) == a
+            True
```
kryzar, line 369: I am not sure this test should be here.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -152,12 +152,37 @@ def __init__(self, gen, category):
...
         """
         # NOTE: There used to be no __init__ here (which was fine). I
         # added one to ensure that DrinfeldModule_finite would always
         # have _frobenius_norm and _frobenius_trace attributes.
         super().__init__(gen, category)
-        self._base_degree_over_constants = self.base_over_constants_field().degree(self._Fq)
+        K = self.base_over_constants_field().backend()
+        self._over_quotient_ring = is_PolynomialQuotientRing(K)
+        char, q = self._Fq.characteristic(), self._Fq.cardinality()
+        qdeg = logb(q, char)
```
kryzar, line 174: I think the name is not so great and the variable is used only once, so I would move the `char` definition to just before it's needed and use `logq(q, char)` in the lambda. Also, why `logb` and not `log`?

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -152,12 +152,37 @@ def __init__(self, gen, category):
...
+        # Currently only support univariate quotient rings with coefficients
+        # in the base field Fq
+        if self._over_quotient_ring:
+            if self._Fq == K.base_field():
+                self._base_degree_over_constants = K.degree()
+                self._frobenius = lambda a, i : a**(q**i)
```
kryzar, line 180: The name `_frobenius` can lead to a confusion with the Frobenius endomorphism. What about something like `base_field_frobenius`? Or `apply_base_field_frobenius`? (Because `base_field_frobenius` would be expected to be a callable while in your case you want to compute images under the callable.)

Tho tbh I'm not sure this is really necessary. This creates a new reliance on an internal attribute, which is not often needed (as of now, four occurrences, all in the same method). Or maybe make it a private method.

src/sage/rings/ring_extension_element.pyx
kryzar: > force a user to have to decide between these two different ways of mapping into the base for ring extensions depending on the type of the extension ring

What do you mean?

> Another could be to not use RingExtensions for Drinfeld modules over univariate quotients

Yeah, that would be a bit weird, no? Fundamentally, we used ring extensions to see the ground field $K$ as a $\mathbb F_q[T]$-field. We want to be able to do that without any assumption on $K$. Am I understanding correctly what you mean?

kryzar (commented):
So, I did a first basic review, and I left a few comments in the code.

More generally:
1. Although your code looks correct, I don't feel legitimate evaluating changes to ring extensions, given that it is a very general set of facilities, whose utilization goes far beyond the scope of Drinfeld modules. Maybe a separate PR should be created for that, to discuss all the options with other members of the organization. I acknowledge that doing so would give you more work, and if you cannot do that right now, that is totally okay.

2. I believe there are not enough tests. You only test `__init__`, `__frobenius_charpoly__` and `__invert__`. Maybe you should:

    - test various algorithms for the Frobenius charpoly, 
    - test the Frobenius trace and norm,
    - showcase the new construction in the docstring of the class (or that of `DrinfeldModule`),
    - test edge/trivial cases, etc.

PR #37925, opened 2024-05-02
Fix slow substitution in piecewise symbolic function

kwankyu:
Discussed in https://groups.google.com/g/sage-devel/c/i4b_0ooOxpQ

Before
```sage
sage:  f = piecewise([((0,1),1)])
sage: time f(0.1)
CPU times: user 1.51 s, sys: 204 ms, total: 1.72 s
Wall time: 1.33 s
1
sage: time f(0.1)
CPU times: user 75.5 ms, sys: 1.96 ms, total: 77.5 ms
Wall time: 76.3 ms
1
sage: time f(0.1)
CPU times: user 75.3 ms, sys: 1.83 ms, total: 77.2 ms
Wall time: 76.1 ms
1
```

After
```sage
sage: f = piecewise([((0,1),1)])
sage: time f(0.1)
CPU times: user 1.5 ms, sys: 80 µs, total: 1.58 ms
Wall time: 1.58 ms
1
sage: time f(0.1)
CPU times: user 280 µs, sys: 1e+03 ns, total: 281 µs
Wall time: 286 µs
1
sage: time f(0.1)
CPU times: user 276 µs, sys: 18 µs, total: 294 µs
Wall time: 300 µs
1
```

### :memo: Checklist

### DIFF
--- a/src/sage/functions/piecewise.py
+++ b/src/sage/functions/piecewise.py
@@ -217,7 +217,7 @@ class PiecewiseFunction(BuiltinFunction):
             if hasattr(point, 'pyobject'):
                 # unwrap any numeric values
                 point = point.pyobject()
-        elif point == x:  # this innocent-looking comparison takes a second
+        elif point == x:  # this comparison may be very slow (see #37925)
             # substitution only in auxiliary variables
             new_params = []
             for domain, func in parameters:
--- a/src/sage/functions/piecewise.py
+++ b/src/sage/functions/piecewise.py
@@ -213,17 +213,16 @@ class PiecewiseFunction(BuiltinFunction):
             piecewise(x|-->-x^sin(y) on (-2, 0), x|-->x - sin(y) on [0, 2]; x)
         """
         point = subs_map.apply_to(x, 0)
-        if point == x:
+        if ((point.is_numeric() or point.is_constant()) and (point.is_real())):
+            if hasattr(point, 'pyobject'):
+                # unwrap any numeric values
+                point = point.pyobject()
+        elif point == x:  # this innocent-looking comparison takes a second
             # substitution only in auxiliary variables
             new_params = []
             for domain, func in parameters:
                 new_params.append((domain, subs_map.apply_to(func, 0)))
             return piecewise(new_params, var=x)
-        if ((point.is_numeric() or point.is_constant())
-            and (point.is_real())):
-            if hasattr(point, 'pyobject'):
-                # unwrap any numeric values
-                point = point.pyobject()
         else:
             raise ValueError('substituting the piecewise variable must result in real number')
 


tscrim (changes requested):
LGTM overall as it makes sense to switch things in this way on its own. Although the comment about the test is misleading (and counterproductive). In part, it only takes so long on the first call as far as I can tell. Seems like a library being loaded. Subsequent calls are fast from what I saw. I think we should not have a comment about that comparison's speed.

kwankyu:
> Although the comment about the test is misleading (and counterproductive)...

I get
```bash
sage: time bool(x == 0.1)
CPU times: user 1.48 s, sys: 206 ms, total: 1.69 s
Wall time: 1.33 s
False
sage: time bool(x == 0.1)
CPU times: user 72.8 ms, sys: 1.79 ms, total: 74.6 ms
Wall time: 73.5 ms
False
sage: time bool(x == 0.1)
CPU times: user 73.7 ms, sys: 1.98 ms, total: 75.7 ms
Wall time: 74.4 ms
False
sage: time bool(x == 0.1)
CPU times: user 72.1 ms, sys: 1.73 ms, total: 73.9 ms
Wall time: 72.8 ms
False
sage: time bool(x == 0.1)
CPU times: user 73.6 ms, sys: 1.83 ms, total: 75.5 ms
Wall time: 74.4 ms
False
sage: time bool(x == 0.1)
CPU times: user 73.7 ms, sys: 1.85 ms, total: 75.5 ms
Wall time: 74.5 ms
False
```
What timings do you get? 

The comment is to warn someone who may make further changes to the code. Do you have a better phrase?

tscrim:
> > Although the comment about the test is misleading (and counterproductive)...
>
> What timings do you get?

Basically the same:
```shell
sage: %time bool(x == .1)
CPU times: user 824 ms, sys: 50.5 ms, total: 875 ms
Wall time: 424 ms
False
sage: %time bool(x == .1)
CPU times: user 13.7 ms, sys: 0 ns, total: 13.7 ms
Wall time: 13.5 ms
False
sage: %time bool(x == .1)
CPU times: user 14.9 ms, sys: 0 ns, total: 14.9 ms
Wall time: 14.6 ms
False
sage: %time bool(x == .1)
CPU times: user 20.8 ms, sys: 0 ns, total: 20.8 ms
Wall time: 20.3 ms
False
sage: %time bool(x == CC(.1,.1))
CPU times: user 15.2 ms, sys: 0 ns, total: 15.2 ms
Wall time: 15.1 ms
False
```
After the first time, I don't see any slowdown. Also, in a fresh session :
```
sage: %time bool(x == SR.var('y'))
CPU times: user 8.63 ms, sys: 3.59 ms, total: 12.2 ms
Wall time: 11.3 ms
False
sage: %time bool(x == SR.var('x'))
CPU times: user 82 µs, sys: 6 µs, total: 88 µs
Wall time: 92.3 µs
True
sage: %time bool(x == .1)
CPU times: user 850 ms, sys: 49.1 ms, total: 900 ms
Wall time: 442 ms
False
sage: %time bool(x == .1)
CPU times: user 17.6 ms, sys: 0 ns, total: 17.6 ms
Wall time: 16.9 ms
False
```
 
> The comment is to warn someone who may make further changes to the code. Do you have a better phrase?

Perhaps something like
```python
# this comparison can be slow the first time it is invoked
```

kwankyu:
My point is that it is slow in average, not only the first time...

kwankyu:
Compare with 
```
sage: time 2 == 0.1
CPU times: user 35 µs, sys: 1e+03 ns, total: 36 µs
Wall time: 40.3 µs
False
```

kwankyu:
Actually I found the "bug" in 
```sage
sage: f = piecewise([((0,1),1)])
sage: plot(f)
```

tscrim:
I think it depends on what is being compared:
```shell
sage: %timeit bool(pi == .1)
35.5 µs ± 117 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit bool(.1 == pi)
40.1 µs ± 59.4 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit bool(x == pi)
77.2 µs ± 187 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit bool(pi == x)
82.1 µs ± 287 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
So I don't think we cannot say it is universally slow.

An interesting thing that I was not expecting:
```
sage: type(pi.pyobject())
<class 'sage.symbolic.constants.Pi'>
sage: type(pi)
<class 'sage.symbolic.expression.Expression'>
sage: %timeit bool(pi.pyobject() == x)
1.84 µs ± 13 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

tscrim:
> Actually I found the "bug" in
> 
> ```python
> sage: f = piecewise([((0,1),1)])
> sage: plot(f)
> ```

I don't understand what you mean by this.

kwankyu:
> So I don't think we cannot say it is universally slow.

Right. What is slow is to compare the symbolic variable `x` with a numerical value.

kwankyu:
> > Actually I found the "bug" in
> > ```python
> > sage: f = piecewise([((0,1),1)])
> > sage: plot(f)
> > ```
> 
> I don't understand what you mean by this.

Try the code on your system.

tscrim:
> > So I don't think we cannot say it is universally slow.
> 
> Right. What is slow is to compare the symbolic variable `x` with a numerical value.

Yes, and my point is that we need to be careful with how we craft the comment to reflect this.

kwankyu:
Did you try the plotting?

tscrim:
> > > Actually I found the "bug" in
> > > ```python
> > > sage: f = piecewise([((0,1),1)])
> > > sage: plot(f)
> > > ```
> > 
> > 
> > I don't understand what you mean by this.
> 
> Try the code on your system.

The plotting is only slightly slower on my system before the change. Is this what you meant?

kwankyu:
Strange. It is very slow before this PR. More than 20 seconds.

I am busy now. I will be back some hours later. Sorry.

tscrim:
Plain:
```shell
sage: %time plot(f)
verbose 0 (3945: plot.py, generate_plot_points) WARNING: When plotting, failed to evaluate function at 101 points.
verbose 0 (3945: plot.py, generate_plot_points) Last error message: 'Unable to compute f(1.0)'
CPU times: user 304.4 ms, sys: 20.6 ms, total: 294.5 ms
Wall time: 293.8 ms
Launched png viewer for Graphics object consisting of 1 graphics primitive
```
With the PR:
```shell
sage: f = piecewise([((0,1),1)])
sage: %time plot(f)
verbose 0 (3945: plot.py, generate_plot_points) WARNING: When plotting, failed to evaluate function at 101 points.
verbose 0 (3945: plot.py, generate_plot_points) Last error message: 'Unable to compute f(1.0)'
CPU times: user 19.4 ms, sys: 307 µs, total: 19.7 ms
Wall time: 19.2 ms
Launched png viewer for Graphics object consisting of 1 graphics primitive
```
Well, 10x slower, but still on the order of ms. Perhaps my machine is significantly fast that it isn't so apparent.

No problem.

As I said, I completely agree with the change. We just need to tweak the comment to not mislead future readers.

kwankyu:
> Well, 10x slower, but still on the order of ms. Perhaps my machine is significantly fast that it isn't so apparent.

Not sure. I just guess that the difference of the machine architecture is affecting the software performance of, perhaps, maxima...

> As I said, I completely agree with the change. We just need to tweak the comment to not mislead future readers.

I think your last version is also inappropriate. Would you suggest another one?

Or how about ```# this comparison may be very slow```?

tscrim:
> > As I said, I completely agree with the change. We just need to tweak the comment to not mislead future readers.
> 
> I think your last version is also inappropriate. Would you suggest another one?
>
> Or how about `# This comparison may be very slow. See #37925`?

Yours is good. I can't think of anything better right now.

kwankyu:
Done.

tscrim (approved):
Thank you.

kwankyu:
Thanks!

PR #37919, opened 2024-05-02
Upgrades and configure fixes for macOS arm64

mkoeppe:
Also
- update for updated GitHub macOS runner labels

### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

### :hourglass: Dependencies

- Depends on #37726 (merged here to resolve merge conflict)

PR #37926, opened 2024-05-03
`Makefile`, `.ci/write-dockerfile.sh`: Update for `src/pyproject.toml` after #36982

mkoeppe:
Also some unrelated fixes:
- fix for `centos-7-devtoolset` needed to adjust for a small change in how Docker files are parsed (https://github.com/sagemath/sage/actions/runs/8931531493/job/24538795775).
- fix for `ubuntu-xenial-...` and other platforms that need special `configure` args
- fix for current macOS Docker Desktop 

### :memo: Checklist


fchapoton:
hmm, tagging doctests as "known bugs" does not look like a good permanent solution

mkoeppe:
That's right, this is a bad temporary solution.

This is fallout from the big revert in #37796.

mkoeppe:
I've removed the commits that made changes to `.github/workflows/*macos.yml` to reduce the scope of the PR; these commits are now on #37919.

.ci/retrofit-worktree.sh
```diff
@@ -36,6 +38,10 @@ git tag -f new
 # The changed files now show up as uncommitted changes.
 # The final "git add -N" makes sure that files that were added in "new" do not show
 # as untracked files, which would be removed by "git clean -fx".
+if [ -L $WORKTREE_NAME ]; then
+    rm -f $WORKTREE_NAME
+    git worktree prune --verbose
+fi
 git worktree add --detach $WORKTREE_NAME
```
kwankyu, line 45: Did #36982 make these four lines necessary? I wonder how...

kwankyu: Or this "retrofit-workreee" could be applied twice somewhere?

mkoeppe: Yes, in Build&Test it's applied three times: 1 time already in the Docker image that it's starting from, and then 2 more times.
This PR is good enough to fix the "CI Linux" workflow that runs on release tags. For "Build&Test", more fixing related to this is necessary.

kwankyu: Anyway, it is just cleaning up before actual work.

kwankyu (approved):

PR #37928, opened 2024-05-03
Add legend_label to implicit plots

Newtech66:
Fixes #15903.

- Introduces an internal `plot_subtype` parameter that is used to determine when legend_label should be added.
- If `plot_subtype == implicit`, adds legend_label by adding a label to the ContourSet artist and directly adding it to the plot.
- It is possible to do something similar for region_plot by using a proxy artist like [here](https://matplotlib.org/stable/users/explain/axes/legend_guide.html#creating-artists-specifically-for-adding-to-the-legend-aka-proxy-artists), but this isn't attempted in this PR because I wasn't sure what to do about the inside, outside, and border regions separately.

### :memo: Checklist

### :hourglass: Dependencies
None

### DIFF
--- a/src/sage/plot/contour_plot.py
+++ b/src/sage/plot/contour_plot.py
@@ -24,7 +24,6 @@ from sage.misc.decorators import options, suboptions
 from sage.plot.colors import rgbcolor, get_cmap
 from sage.arith.srange import xsrange
 
-
 class ContourPlot(GraphicPrimitive):
     """
     Primitive class for the contour plot graphics type.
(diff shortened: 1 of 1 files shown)


kcrisman:
I can't check the overall validity of the code (though it seems correct on
a first glance) at this time, but I do have a few questions/suggestions
before testing and positive review.

* Definitely needs some documentation examples for the implicit plots,
assuming there aren't any currently existing (since they wouldn't have had
legends, I guess) - perhaps both in the "tests" and "plot" for
documentation section.
* Am I correct that `'region_border'` is reserved for the future region use
you mention in the PR, so to speak?
* I'm probably just rusty with some Python conventions, but why is the
option `_subtype` when you pop it from the options (if present), but
`subtype` when it's input (and in the actual changed code for plain old
implicit plots)?

PR #37940, opened 2024-05-05
Support morphisms in `Matroid` constructor

gmou3:
This follows the merging of #37692 and it enables the (re-)feeding of a linear matroid's morphism representation into the `Matroid` constructor.

Example:
```
sage: M = matroids.catalog.Fano()
sage: A = M.representation(order=True); A
Generic morphism:
  From: Free module generated by {'a', 'b', 'c', 'd', 'e', 'f', 'g'} over Finite Field of size 2
  To:   Free module generated by {0, 1, 2} over Finite Field of size 2
sage: Matroid(A)
Binary matroid of rank 3 on 7 elements, type (3, 0)
```


tscrim (approved):
LGTM.

gmou3:
Thanks!

@mkoeppe you may wish to have a look.

src/sage/matroids/constructor.py
```diff
@@ -856,9 +876,15 @@ def Matroid(groundset=None, data=None, **kwds):
...
-    elif key in ['matrix', 'reduced_matrix']:
+    elif key in ['matrix', 'reduced_matrix', 'morphism']:
         A = data
         is_reduced = (key == 'reduced_matrix')
+        if key == 'morphism':
+            if isinstance(data, tuple):
+                A = data[0]
+                groundset = data[1]
+            if isinstance(A, sage.modules.with_basis.morphism.ModuleMorphism):
+                A = A.matrix()
```
mkoeppe: The point of the "morphism" output is that you can read off the groundset from the domain!

mkoeppe: (And in the reduced case, from the domain and codomain.)

gmou3: Right. I think it works nicely now.

src/sage/matroids/constructor.py
```diff
@@ -734,6 +778,10 @@ def Matroid(groundset=None, data=None, **kwds):
             key = 'graph'
         elif is_Matrix(data):
             key = 'matrix'
+        elif isinstance(data, sage.modules.with_basis.morphism.ModuleMorphism) or (
+             isinstance(data, tuple) and
```
mkoeppe, line 783: I'm not sure if I find "tuple" input very convincing. 
Does it have a clear benefit over just passing the groundset via the `groundset` parameter?

gmou3: The reason I did this is so that the following works:
```
sage: M = matroids.catalog.Fano()
sage: A = M.representation(order=True, labels=True)
sage: Matroid(A)
...
```

src/sage/matroids/constructor.py
```diff
@@ -734,6 +778,10 @@ def Matroid(groundset=None, data=None, **kwds):
             key = 'graph'
         elif is_Matrix(data):
             key = 'matrix'
+        elif isinstance(data, sage.modules.with_basis.morphism.ModuleMorphism) or (
```
mkoeppe, line 782: instead of using `isinstance` with this class, I think it's better to use `isinstance(data, Morphism) and data.category_for().is_subcategory(Modules(data.base_ring()).WithBasis())`

src/sage/matroids/constructor.py
mkoeppe: yes, ok

gmou3: I need to do something similar for matrices though.

src/sage/matroids/constructor.py
gmou3: Not sure what's best but the above doesn't seem to work.

src/sage/matroids/constructor.py
gmou3: I made it so that this also works and retains the groundset labels:
```
sage: M = matroids.catalog.Fano()
sage: A = M.representation(labels=True)
sage: Matroid(A)
...
```

mkoeppe:
LGTM, thanks!

Issue #35357, opened 2023-03-26
Cannot create GF(65537^12)


remyoudompheng:
### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

### Did you read the documentation and troubleshoot guide?

- [X] I have read the documentation and troubleshoot guide

### Environment

```markdown
- **OS**: Archlinux
- **Sage Version**: 9.8
```

### Steps To Reproduce

Try to create the finite field of order 65537^12 (without variable name)
```
sage: GF(65537^12)
```
Or any other exponent which is a multiple of 12.

### Expected Behavior

Create the finite field of order 65537^12

### Actual Behavior

Throws an exception:
```
  File "sage/rings/algebraic_closure_finite_field.py", line 1119, in _get_polynomial
    return self._pseudo_conway_lattice.polynomial(n)
  File "sage/rings/finite_rings/conway_polynomials.py", line 233, in polynomial
    _frobenius_shift(K, xi)
  File "sage/rings/finite_rings/conway_polynomials.py", line 432, in _frobenius_shift
    crt[(i,j)].append(_find_pow_of_frobenius(p, m, compatible[m][qlist[j]], compatible[m][qlist[i]]))
  File "sage/rings/finite_rings/conway_polynomials.py", line 318, in _find_pow_of_frobenius
    raise RuntimeError("No appropriate power of Frobenius found")
RuntimeError: No appropriate power of Frobenius found
```

### Additional Information

This was found randomly while trying to create a doctest.
It seems related to specific arithmetic properties and difficult to reproduce randomly with other values of p and the degree.

grhkm21:
Not-so-minimal reproducible example:

```python
sage: p = 65537
....: Fpbar = GF(p).algebraic_closure("z")
....: try: Fpbar._pseudo_conway_lattice.polynomial(12)
....: except RuntimeError: print("Error")
Error
sage: p = 65537
....: Fpbar = GF(p).algebraic_closure("z", use_database=False)
....: Fpbar._pseudo_conway_lattice.polynomial(12)
x^12 + 49224*x^11 + 29157*x^10 + 43350*x^9 + 35738*x^8 + 17809*x^7 + 41614*x^6 + 28700*x^5 + 11177*x^4 + 5221*x^3 + 38906*x^2 + 31504*x + 3
```

[...]

PR #37949, opened 2024-05-06
Adapt length check in `sage.databases.conway` in preparation for database update

S17A05:
Motivated by #35357, the database of Conway polynomials will be enlarged to ensure compatibility in the generation of finite fields (see https://github.com/sagemath/conway-polynomials/pull/5 for details). To prepare for this update, we slightly relax the length check in `sage.databases.conway` to be compatible with both the old and the new version of the database.

### DIFF
--- a/src/sage/databases/conway.py
+++ b/src/sage/databases/conway.py
@@ -138,11 +138,16 @@ class ConwayPolynomials(Mapping):
         """
         Return the number of polynomials in this database.
 
-        TESTS::
+        TESTS:
+
+        The database currently contains `35357` polynomials, but due to
+        :issue:`35357` it will be extended by Conway polynomials of
+        degrees `1`, `2` and `3` for primes between `65537` and `110000`,
+        thus leading to a new total of `47090` entries::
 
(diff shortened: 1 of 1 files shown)


tscrim:
Might be good to add a comment to the doctest showing why its output can be two different things. Otherwise LGTM.

src/sage/databases/conway.py
```diff
@@ -138,11 +138,16 @@ def __len__(self):
         """
         Return the number of polynomials in this database.
 
-        TESTS::
+        TESTS:
+
+        The database currently contains `35357` polynomials, but due to
+        :issue:`35357` it will be extended by Conway polynomials of
```
orlitzky, line 144: I was certainly expecting one of these `35357`s to be a typo, but nope!

orlitzky (approved):

src/sage/databases/conway.py
S17A05: I was also very surprised about the coincidence when I wrote it!

PR #37955, opened 2024-05-07
`Graph.{[weighted_]adjacency_matrix,kirchhoff_matrix}`: Support constructing `End(CombinatorialFreeModule)` elements

mkoeppe:
This is a follow-up after
- #37692 

... to cover a few more methods. The methods can now create endomorphisms of free modules whose bases are indexed by the vertices. To help with this, we make the `matrix` constructor a bit more flexible.

This is also preparation for making the spectral graph theory methods ready for `CombinatorialFreeModule`:
- #37943

### :memo: Checklist


src/sage/graphs/generic_graph.py
```diff
@@ -1900,6 +1900,23 @@ def to_dictionary(self, edge_labels=False, multiple_edges=False):
 
         return d
 
+    def _vertices_keys(self, vertices=None, *, sort=None):
```
mkoeppe, line 1903: @dcoudert Help with naming this method would be very welcome... I noticed that there is a similarly named method `_vertices_keys` (with a very different purpose)

dcoudert: I'm not aware of any `_vertices_keys` method (no match with git grep). 

Can you at least document the purpose of this method ?

src/sage/graphs/generic_graph.py
```diff
@@ -1900,6 +1900,68 @@ def to_dictionary(self, edge_labels=False, multiple_edges=False):
 
         return d
 
+    def _vertex_indices_and_keys(self, vertices=None, *, sort=None):
+        r"""
+        Process a ``vertices`` parameter.
+
```
dcoudert, line 1906: May be you could add that this is a helper function for methods like `adjacency_matrix`, `kirchhoff_matrix`, etc.

src/sage/graphs/generic_graph.py
mkoeppe: Sorry, meanwhile inspiration had struck and I reworked this method as `_vertex_indices_and_keys`.

src/sage/graphs/generic_graph.py
mkoeppe: Good idea, done in 98323cb719e

dcoudert (approved):
LGTM.

I assume that if needed we can change the name of the new method. I don't have better option in mind.

mkoeppe:
Thank you!

PR #37956, opened 2024-05-07
faster implementation of method `sage.graphs.generators.families.BalancedTree`

dcoudert:
We propose a direct implementation of the graph generator `BalancedTree` that is faster than the previous method using networkx.

Before (using networkx)
```
sage: %timeit G = graphs.BalancedTree(3, 5)
550 µs ± 1.17 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
sage: %timeit G = graphs.BalancedTree(3, 7)
5.33 ms ± 37.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
sage: %timeit G = graphs.BalancedTree(4, 5)
2.13 ms ± 18.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
sage: %timeit G = graphs.BalancedTree(4, 7)
54.5 ms ± 278 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

With this PR
```
sage: %timeit G = graphs.BalancedTree(3, 5)
157 µs ± 809 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit G = graphs.BalancedTree(3, 7)
1.44 ms ± 4.94 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
sage: %timeit G = graphs.BalancedTree(4, 5)
569 µs ± 6.78 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
sage: %timeit G = graphs.BalancedTree(4, 7)
9.52 ms ± 61.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/generators/families.py
+++ b/src/sage/graphs/generators/families.py
@@ -522,8 +522,9 @@ def BalancedTree(r, h):
 
         sage: G = graphs.BalancedTree(2, 1); G
         Balanced tree: Graph on 3 vertices
-        sage: G.order(); G.size()
+        sage: G.order()
         3
+        sage: G.size()
         2
         sage: r = 2; h = 1
         sage: v = 1 + r
@@ -534,7 +535,8 @@ def BalancedTree(r, h):
     Plot a balanced tree of height 5, whose root node has degree `r = 3`::
 
         sage: G = graphs.BalancedTree(3, 5)
-        sage: G.show()                          # long time                             # needs sage.plot
+        sage: G.plot()                          # long time                             # needs sage.plot
+        Graphics object consisting of 728 graphics primitives
 
     A tree is bipartite. If its vertex set is finite, then it is planar. ::
 
@@ -580,13 +582,12 @@ def BalancedTree(r, h):
     order = [r**l for l in range(h + 1)]
     # Compute the first index of the vertices of a level
     begin = [0]
-    for level in range(1, h + 2):
-        begin.append(begin[level - 1] + order[level - 1])
+    begin.extend(begin[-1] + val for val in order)
     # The number of vertices of the tree is the first index of level h + 1
     T = Graph(begin[-1], name="Balanced tree")
 
     # Add edges of the r-ary tree
-    for level in range(0, h):
+    for level in range(h):
         start = begin[level + 1]
         for u in range(begin[level], begin[level + 1]):
             T.add_edges((u, v) for v in range(start, start + r))
--- a/src/sage/graphs/generators/families.py
+++ b/src/sage/graphs/generators/families.py
@@ -496,7 +496,7 @@ def HammingGraph(n, q, X=None):
 
 def BalancedTree(r, h):
     r"""
-    Returns the perfectly balanced tree of height `h \geq 1`,
+    Return the perfectly balanced tree of height `h \geq 1`,
     whose root has degree `r \geq 2`.
 
     The number of vertices of this graph is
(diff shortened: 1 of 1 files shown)


src/sage/graphs/generators/families.py
```diff
@@ -513,22 +513,16 @@ def BalancedTree(r, h):
...
     EXAMPLES:
 
     A balanced tree whose root node has degree `r = 2`, and of height
     `h = 1`, has order 3 and size 2::
 
-        sage: G = graphs.BalancedTree(2, 1); G                                          # needs networkx
+        sage: G = graphs.BalancedTree(2, 1); G
         Balanced tree: Graph on 3 vertices
-        sage: G.order(); G.size()                                                       # needs networkx
+        sage: G.order(); G.size()
```
tscrim: It would be good to split this doctest into 2 lines.

src/sage/graphs/generators/families.py
```diff
@@ -563,17 +557,41 @@ def BalancedTree(r, h):
...
+        True
     """
-    import networkx
-    return Graph(networkx.balanced_tree(r, h), name="Balanced tree")
+    # Compute the number of vertices per level of the tree
+    order = [r**l for l in range(h + 1)]
+    # Compute the first index of the vertices of a level
+    begin = [0]
+    for level in range(1, h + 2):
+        begin.append(begin[level - 1] + order[level - 1])
```
tscrim:
```suggestion
    for level in range(h + 1):
        begin.append(begin[level] + order[level]) # level is 0-based
```
Micro-optimization. Although you could do
```python
    begin.extend(begin[-1] + val for val in order)
```
but that is more subtle.

src/sage/graphs/generators/families.py
```diff
@@ -563,17 +557,41 @@ def BalancedTree(r, h):
...
+    # The number of vertices of the tree is the first index of level h + 1
+    T = Graph(begin[-1], name="Balanced tree")
+
+    # Add edges of the r-ary tree
+    for level in range(0, h):
```
tscrim:
```suggestion
    for level in range(h):
```
Style preference.

src/sage/graphs/generators/families.py
```diff
@@ -539,8 +533,8 @@ def BalancedTree(r, h):
 
     Plot a balanced tree of height 5, whose root node has degree `r = 3`::
 
-        sage: G = graphs.BalancedTree(3, 5)                                             # needs networkx
-        sage: G.show()                          # long time                             # needs networkx sage.plot
+        sage: G = graphs.BalancedTree(3, 5)
+        sage: G.show()                          # long time                             # needs sage.plot
```
tscrim:
```suggestion
        sage: G.plot()                          # long time                             # needs sage.plot
```
This would provide a more descriptive output.

tscrim (approved):
All of these changes can be ignored if you prefer as they are either tangential or micro-optimizations. Once changed or ignored, you can set a positive review.

src/sage/graphs/generators/families.py
dcoudert: I like it.

src/sage/graphs/generators/families.py
dcoudert: done

src/sage/graphs/generators/families.py
dcoudert: ok

src/sage/graphs/generators/families.py
dcoudert: right

dcoudert:
I have implemented all proposed changes. Thanks.

tscrim:
Thank you.

PR #37967, opened 2024-05-08
Modify tests in `ell_point.py` to be more generic for Conway database update

S17A05:
As part of the Conway polynomials database update, two tests in `schemes.elliptic_curves.ell_point` caused issues due to their dependency on the precise implementation of the finite field `GF(65537^2)`:
- The first test (of `.weil_pairing`) checks whether the Sage implementation still works for an example of fixed `7282`-torsion points
- The second test (of `.tate_pairing`) checks whether the new implementation, which uses PARI up to reduction of the pairing, coincides with the old Sage implementation (still used e.g. in SageMathCell, but not available in the current code anymore)

To help with the database update, we adapt these tests to not be dependent on the exact minimal polynomial used for finite field generation:
- We turn both tests into examples (with fixed moduli) for their respective methods
- We modify the test of `.weil_pairing` to check for two randomly chosen `7282`-torsion points whether `algorithm="sage"` and `algorithm="pari"` yield the same result

Please let me know if you think other polynomial-independent tests are better suited as replacements.

See sagemath/conway-polynomials#5 (especially [comment 10](https://github.com/sagemath/conway-polynomials/pull/5#issuecomment-2101462886) and Update 2 in [comment 11](https://github.com/sagemath/conway-polynomials/pull/5#issuecomment-2101498241)) for more details.

CC: @yyyyx4 @GiacomoPope @JohnCremona

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_point.py
+++ b/src/sage/schemes/elliptic_curves/ell_point.py
@@ -1833,16 +1833,20 @@ class EllipticCurvePoint_field(SchemeMorphism_point_abelian_variety_field):
 
         TESTS:
 
-        Check that the original Sage implementation still works::
+        Check that the original Sage implementation still works and
+        that the result coincides with the PARI implementation::
 
             sage: # needs sage.rings.finite_rings
             sage: GF(65537^2).inject_variables()
             Defining z2
             sage: E = EllipticCurve(GF(65537^2), [0,1])
-            sage: P = E(22, 28891)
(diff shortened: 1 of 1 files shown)


tornaria:
LGTM. Otherwise, either one in https://github.com/sagemath/conway-polynomials/pull/5#issuecomment-2101629342 works and avoids removing the tests.

tornaria:
Nice, thanks.

PR #37972, opened 2024-05-09
Fix comparison for divisors of curves (and FormalSum commutativity)

### LINKED ISSUE #37966: Equality comparison gives incorrect result in divisor group of a curve
### Steps To Reproduce

Example code:

```python
E = EllipticCurve([1, 2])

# Two points on E
P = E(-1, 0)
Q = E(1, 2)

# Get the points as divisors
Pd = E.divisor(P)
Qd = E.divisor(Q)

# These should be equal, but they aren't
Pd + Qd == Qd + Pd

# Nevertheless, their difference is calculated to be 0
(Pd + Qd) - (Qd + Pd)
```

The same thing happens with `HyperellipticCurve` and `Curve`.

This example in particular violates commutativity of the divisor group, despite the divisor group being marked in Sage as part of the "Category of commutative additive groups".

### Expected Behavior

Comparison in the divisor group should give the correct result or at the very least return a not implemented error.

### Actual Behavior

Comparison in the divisor group gives an incorrect result.

### Additional Information

`__eq__` needs to be implemented for divisors. Something like this should work:

```python
def __eq__(self, other):
    return self - other == 0
```

but I'm not familiar enough with the internals of Sage to know where this should go.

### Environment

```markdown
- **OS**: Fedora 40
- **Sage Version**: 10.3, built from source in a conda environment
```

I ran `make ptestlong` and all tests passed, so I doubt it's an issue with my environment.

### Checklist

vincentmacri:
Fixes #37966.

### :memo: Checklist


tscrim:
Thank you for submitting the PR.

First, you should not be changing the `.gitignore`.

Second, as far as I can tell, this is just reimplementing one part of the `FormalSum` comparison with a more complicated check (more transient objects and subtraction). The underlying issue is the comparison that is happening with `FormalSum` is not commutative, but its implementation is (effectively) commutative. In `FormalSum.reduce()`, it should sort the input:
```python
try:
    self._data = [(c, x) for (x, c) in sorted(new.items()) if c]
    self._data = [(c, x) for (x, c) in sorted(new.items(), key=str) if c]
```

vincentmacri:
> First, you should not be changing the `.gitignore`.

The change to the `.gitignore` was some auto-generated files I think from running `./bootstrap` (I followed these instructions: https://doc.sagemath.org/html/en/installation/conda.html#using-conda-to-provide-all-dependencies-for-the-sage-library). I'll undo that though if it's out of scope (let me know if I should submit it as a separate PR).

> Second, as far as I can tell, this is just reimplementing one part of the `FormalSum` comparison with a more complicated check (more transient objects and subtraction). The underlying issue is the comparison that is happening with `FormalSum` is not commutative, but its implementation is (effectively) commutative. In `FormalSum.reduce()`, it should sort the input:
> 
> ```python
> try:
>     self._data = [(c, x) for (x, c) in sorted(new.items()) if c]
> except (TypeError, ValueError):
>     self._data = [(c, x) for (x, c) in sorted(new.items(), key=str) if c]
> ```

I looked at `FormalSum`, and that comparison is element-wise. In particular, comparing the formal sums `3 + 2 > 2 + 3` gives `True`, and if I implemented this behaviour in `FormalSum` then we would have both `3 + 2 == 2 + 3` and `3 + 2 > 2 + 3`, and I didn't think both of these should evaluate to `True`.

You're right that there is an issue where `FormalSum` isn't commutative. I'm not familiar enough with how `FormalSum` is used in Sage to know what to fix there.

Either way, I think that overriding the `FormalSum` implementation is needed anyway, because the divisors of a curve are a group, and shouldn't be treated exactly like formal sums. For instance, if two different curves happen to have the same point on both of them, as a group it doesn't make sense to say that those points are equal. That's what the `self.parent() == other.parent()` check is for in my PR. With formal sums though, these are treated as equal right now.

Here's an example of what I mean, which I think is a bug with comparison of divisors:

```python
C = EllipticCurve([2, 1])
R = C(1, 2)

E = EllipticCurve([1, 2])
Q = E(1, 2)

R == Q # Evaluates to False, which is correct

E.divisor(Q) == C.divisor(R) # Evaluates to True, which I think is incorrect
```

vincentmacri:
Some additional code showing that `E.divisor(Q) == C.divisor(R)` should not be `True`:

```python
C = EllipticCurve([2, 1])
R = C(1, 2)

E = EllipticCurve([1, 2])
Q = E(1, 2)

Qd = E.divisor(Q)
Rd = C.divisor(R)

Qd == Rd # Evaluates to True, which as I said above I think is incorrect

Qd.support() == Rd.support() # Evaluates to False, which I think is correct but doesn't make sense if Qd == Rd is evaluating to True.
```

tscrim:
Yes, the changes are out of scope. It's (very) bad practice to hide such changes in unrelated PRs. You can just remove those files yourself anyways; bootstrap is only run sparingly.

Okay, the `FormalSum` implementation is separate (but definitely an issue!). Running through your example I get:
```
sage: Rd.support()
[(1 : 2 : 1)]
sage: Qd.support()
[(1 : 2 : 1)]
sage: Rd == Qd
True
sage: Rd.support() == Qd.support()
True
```
The formal sum does also compare by parents. The subtlety is that there is a coercion map:
```
sage: Rd.parent().has_coerce_map_from(Qd.parent())
True
```
So the comparison is done after coercing both into a common parent (here, that is `Rd.parent()`). However, there should not be a coercion map. The parent `FormalSums` just does the "correct" thing of just comparing base rings for coercion, since that is the only input it knows about. However, the actual divisor group class should not allow coercions when there is not a coercion from the defining schemes. Hence, I think the correct fix would be to implement in `sage.schemes.generic.divisor_group.DivisorGroup_generic`
```python
def _coerce_map_from_(self, other):
    return (isinstance(other, type(self))
            and self._scheme == other._scheme
            and super()._coerce_map_from_(other))
```

vincentmacri:
> Okay, the `FormalSum` implementation is separate (but definitely an issue!). Running through your example I get:
> 
> ```
> sage: Rd.support()
> [(1 : 2 : 1)]
> sage: Qd.support()
> [(1 : 2 : 1)]
> sage: Rd == Qd
> True
> sage: Rd.support() == Qd.support()
> True

Strange, I get `Rd.support() == Qd.support()` is `False` both in 10.3 and on the `develop` branch...

Alternatively, I could do a larger rewrite of the `Divisor_generic` class using the same idea @mkoeppe suggested here: https://github.com/sagemath/sage/pull/37705#issuecomment-2028985619

That would involve making `Divisor_generic` no longer inherit from `FormalSum`, and instead have `Divisor_generic` have a `FormalSum` as a property. Then we could do stuff like not allow adding divisors that live in two different schemes together. I'd then have to implement `__add__`, `__eq__`, etc in `Divisor_generic`.

tscrim:
No, there's no reason for the element class to not inherit from `FormalSum`; it is a formal sum and you would quickly find you are duplicating or redirecting *everything* there. Even if you did that, you would still have a problem. The issue is in the coercion of the parents, and it is essentially a 1 line fix as I said.

vincentmacri:
To make sure I'm understanding before I go and implement your suggestions, you're suggesting that I make the following 3 changes?

1. Get rid of the `__eq__` implementation I added in `Divisor_generic` in favour of the next two points.
2. Change `reduce` in `FormalSum` to give some sort of "canonical representative" by sorting the terms. This would fix the issue with comparing not respecting commutativity in `Divisor_generic` (and all subclasses of `FormalSum` for that matter)
3. Implement `_coerce_map_from_` in `Divisor_generic` to handle correctly comparing divisors that come from different schemes.

Is that correct?

tscrim:
Yes, that's correct. Either 2 or the richcmp should convert the data to a dict for comparisons. Although I think the normalization of the representative is better for doctors output.

vincentmacri:
Okay, I made those changes.

The build will fail because this test is failing:

```python
sage: C = EllipticCurve([2, 1])
sage: R = C(1, 2)
sage: E = EllipticCurve([1, 2])
sage: Q = E(1, 2)
sage: Qd = E.divisor(Q)
sage: Rd = C.divisor(R)
sage: Qd == Rd
False
```

For some reason `Qd == Rd` is giving true, despite me having a test that `_coerce_map_from_` gives `False` on those inputs.

tscrim:
The `_coerce_map_from_` needs to be implemented in the parent class (`DivisorGroup_generic` in `sage.schemes.generic.divisor_group`), not the element class (`Divisor_generic`).

Also, I don't think it is worthwhile for `FormalSum.support` to change its output type. If you want to change its output, then a `frozenset` is best I think. There already is an implicit requirement the indices are hashable.

vincentmacri:
> The `_coerce_map_from_` needs to be implemented in the parent class (`DivisorGroup_generic` in `sage.schemes.generic.divisor_group`), not the element class (`Divisor_generic`).

That fixed it! Everything seems to be working now.

> Also, I don't think it is worthwhile for `FormalSum.support` to change its output type. If you want to change its output, then a `frozenset` is best I think. There already is an implicit requirement the indices are hashable.

My reasoning for this was that mathematically, the support is a set. In fact, the documentation for the support function in Sage calls it a "set of points". I'm not sure why `frozenset` would be preferred over `Set`. Changing it to `frozenset` seemed to break something I think involving enumerating over the set, and the Sage documentation says that for finite sets (which we have here) that `Set` is [a wrapper around Python’s enumerated immutable frozenset type with extra functionality](https://doc.sagemath.org/html/en/reference/sets/sage/sets/set.html).

If you disagree and still think it should remain as a list (or changed to `frozenset`) I'll make the necessary changes.

tscrim:
> > Also, I don't think it is worthwhile for `FormalSum.support` to change its output type. If you want to change its output, then a `frozenset` is best I think. There already is an implicit requirement the indices are hashable.
> 
> My reasoning for this was that mathematically, the support is a set. In fact, the documentation for the support function in Sage calls it a "set of points". I'm not sure why `frozenset` would be preferred over `Set`. Changing it to `frozenset` seemed to break something I think involving enumerating over the set, and the Sage documentation says that for finite sets (which we have here) that `Set` is [a wrapper around Python’s enumerated immutable frozenset type with extra functionality](https://doc.sagemath.org/html/en/reference/sets/sage/sets/set.html).
> 
> If you disagree and still think it should remain as a list (or changed to `frozenset`) I'll make the necessary changes.

Sage `Set`s are slow (especially compared to `list` and `frozenset`) and this is a fairly low-level operation (i.e., we care about the performance). The returned `list` is still a set, just represented as a `list`. You can still enumerate over `frozenset`; if something is breaking, that is indicating either something about the change (which features are depending on the output having certain features) or the broken code needs to change because it is making assumptions that it possibly shouldn't.

vincentmacri:
This was the error I got when running the tests with `frozenset`:

```
File "/[path to sage]/sage/src/sage/schemes/generic/divisor.py", line 483, in coefficient
        t, i = search(self.support(), P)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
File "sage/misc/search.pyx", line 28, in sage.misc.search.search
        cpdef search(object v, object x):
      File "sage/misc/search.pyx", line 63, in sage.misc.search.search
        i = bisect.bisect_left(v, x)
    TypeError: 'frozenset' object does not support indexing
```

I'll just leave it as a list.

src/sage/schemes/generic/divisor.py
```diff
@@ -218,6 +241,26 @@ def _repr_(self):
...
+            sage: Q = E(1, 2)
+            sage: Qd = E.divisor(Q)
+            sage: Rd = C.divisor(R)
+            sage: Qd._coerce_map_from_(Rd)
+            False
+            sage: Rd._coerce_map_from_(Qd)
+            False
+        """
+        return (isinstance(other, type(self)) and self.scheme() == other.scheme() and super()._coerce_map_from_(other))
+
```
tscrim:
```suggestion
```
Leftover to be removed.

src/sage/structure/formal_sum.py
```diff
@@ -93,7 +92,7 @@ def __init__(self, x, parent=None, check=True, reduce=True):
           coefficients into base ring, which can speed
           up constructing a formal sum.
         - ``reduce`` -- reduce (default: ``True``) if ``False``, do not
-          combine common terms
+          combine common terms. WARNING: setting this to ``False`` can cause commutativity issues when comparing sums that are equal but in different orders.
```
tscrim:
```suggestion
          combine common terms

      .. WARNING::

        Setting ``reduce`` to ``False`` can cause issues when comparing
        equal sums but with different orders and/or cancellations.
```

src/sage/structure/formal_sum.py
```diff
@@ -230,6 +229,15 @@ def _richcmp_(self, other, op):
             True
             sage: a == 0          # 0 is coerced into a.parent()(0)
             False
+
+        TESTS::
+
+            sage: a = FormalSum([(1,3),(2,5)])
+            sage: b = FormalSum([(2,5),(1,3)])
```
tscrim:
```suggestion
            sage: a = FormalSum([(1, 3), (2, 5)])
            sage: b = FormalSum([(2, 5), (1, 3)])
```

tscrim (changes requested):
I think that is good for now. Really, this is duplicating what `CombinatorialFreeModuleElement` is providing, but implemented as a `list` rather than a `dict` (which I would argue is the correct thing for commutative addition). However, that is a bigger change and for a separate PR.

Anyways, just a few minor things left to do.

vincentmacri:
> Really, this is duplicating what `CombinatorialFreeModuleElement` is providing, but implemented as a `list` rather than a `dict` (which I would argue is the correct thing for commutative addition). However, that is a bigger change and for a separate PR.

Yeah, there's definitely room for some bigger refactoring changes to `FormalSum`.

Thanks for your review!

tscrim (approved):
Thank you. We will just wait for the bots to come back, but if the checks all pass, then this will be a positive review.

vincentmacri:
The issue is with coercion. The change to `_coerce_map_from_` in `DivisorGroup_generic`  returning `False` when trying to coerce from divisors over the integers to divisors over the rational. Testing a fix, I'll push once all tests pass.

vincentmacri:
I think that should do it. I tested the scheme and structure modules since that's what this PR touches.

I had a couple test failures running locally, but those tests also fail when run in my installation of Sage 10.3 so I don't think it's because of this PR. The failing tests had to do with checking some real values.

Failures:

```
File "src/sage/schemes/projective/projective_morphism.py", line 1764, in sage.schemes.projective.projective_morphism.SchemeMorphism_polynomial_projective_space_field._number_field_from_algebraics
Failed example:
    f._number_field_from_algebraics()                                     # needs sage.symbolic
Expected:
    Scheme morphism:
      From: Projective Space of dimension 1 over Number Field in a
            with defining polynomial y^4 + 3*y^2 + 1
            with a = 0.?e-151 + 0.618033988749895?*I
      To:   Projective Space of dimension 2 over Number Field in a
            with defining polynomial y^4 + 3*y^2 + 1
            with a = 0.?e-151 + 0.618033988749895?*I
      Defn: Defined on coordinates by sending (x : y) to
            (x^2 + (a^3 + 2*a)*x*y + 3*y^2 : y^2 : (2*a^2 + 3)*x*y)
Got:
    Scheme morphism:
      From: Projective Space of dimension 1 over Number Field in a with defining polynomial y^4 + 3*y^2 + 1 with a = 0.?e-113 + 0.618033988749895?*I
      To:   Projective Space of dimension 2 over Number Field in a with defining polynomial y^4 + 3*y^2 + 1 with a = 0.?e-113 + 0.618033988749895?*I
      Defn: Defined on coordinates by sending (x : y) to
            (x^2 + (a^3 + 2*a)*x*y + 3*y^2 : y^2 : (2*a^2 + 3)*x*y)

```

```
File "src/sage/schemes/curves/zariski_vankampen.py", line 610, in sage.schemes.curves.zariski_vankampen.newton
Failed example:
    n.real().endpoints()
Expected:
    (-0.0460743801652894, 0.0291454081632654)
Got:
    (-0.0147727272727274, 0.00982142857142862)
```

vincentmacri:
For reviewing: it was `_coerce_map_from_` that I changed in the last commit (also added a couple more tests).

tscrim:
Looks like you have some numerical noise. Usually such tests are tested on different machines so they always pass. Although the second failure is strange. It might be good to open (separate) issues about those and provide detail about your computer.

src/sage/structure/formal_sum.py
```diff
@@ -95,6 +94,11 @@ def __init__(self, x, parent=None, check=True, reduce=True):
         - ``reduce`` -- reduce (default: ``True``) if ``False``, do not
           combine common terms
 
+      .. WARNING::
+
+        Setting ``reduce`` to ``False`` can cause issues when comparing
+        equal sums but with different orders and/or cancellations.
```
tscrim:
```suggestion
        .. WARNING::

            Setting ``reduce`` to ``False`` can cause issues when comparing
            equal sums but with different orders and/or cancellations.
```
The alignment is off here.

tscrim:
I don't really like your fix. I am actually surprised that it works.

Actually, I made the check too strong. Sorry about that. Instead, try
```python
    def _coerce_map_from_(self, other):
        return (isinstance(other, DivisorGroup_generic)
                and self.scheme() == other.scheme()
                and super()._coerce_map_from_(other))
```
Tests pass in the schemes folder for me.

vincentmacri:
Getting quite a few test failures now, I think coming from the changes to `FormalSum`. Some of the issues are just that `reduce` puts the terms in a different order (which is easy to fix those tests), but some of them are just actually different.

If I change `reduce` back to what it was before this PR then the tests pass (but of course then this PR wouldn't fix the bug it's trying to fix). I'm going to try doing the sorting in `richcmp` instead of in `reduce`.

vincentmacri:
Hopefully that fixes everything. I think this is a bit cleaner because now tests aren't reliant on the arbitrary string ordering I had for comparison purposes. Instead terms are in the order the you enter them in, which makes sense.

I didn't cache the results of sorting the `FormalSum` data, so the sorting will be computed on each `FormalSum` comparison. Let me know if you'd prefer I cache the results of the sorted data in `__init__` (or elsewhere).

src/sage/schemes/generic/divisor.py
```diff
@@ -45,6 +45,7 @@
 from sage.misc.repr import repr_lincomb
 from sage.misc.search import search
 from sage.rings.integer_ring import ZZ
+from sage.sets.set import Set
```
tscrim:
```suggestion
```
This is no longer needed.

src/sage/schemes/generic/divisor.py
```diff
@@ -384,6 +409,7 @@ def support(self):
         This checks that :issue:`10732` is fixed::
 
             sage: R.<x, y, z> = GF(5)[]
+            sage: PS = ProjectiveSpace(R)
```
tscrim:
```suggestion
```
This isn't used in the test.

src/sage/schemes/generic/divisor_group.py
```diff
@@ -190,6 +190,33 @@ def _element_constructor_(self, x, check=True, reduce=True):
         else:
             return Divisor_generic([(self.base_ring()(1), x)], check=False, reduce=False, parent=self)
 
+    def _coerce_map_from_(self, other):
+        r"""
+        Check if there is a coercion from ``other`` to ``self``.
+        This is used to prevent divisors on different schemes from comparing as equal to each other.
```
tscrim:
```suggestion
        Return if there is a coercion map from ``other`` to ``self``.

        There is a coercion from another divisor group if the
        schemes are equal and there is a coercion map from
        the base rings.
```
As I write this, I wonder if instead we want to check instead `self.scheme().has_coerce_map_from(other.scheme())` instead of the schemes being equal. This feels like a more natural condition.

tscrim (changes requested):
I feel like not normalizing the print representation is a small step backwards. It also means that we have to sort on every comparison (instead of just when calling `reduce()`). Although I don't know how these are used, so I don't know which could affect speed more. Well, I understand making a minimal change. I just want to mention that the fact that some are genuinely different might be exposing bugs. That might warrant further investigation later.

src/sage/schemes/generic/divisor_group.py
vincentmacri: Makes sense, yeah.

vincentmacri:
> I feel like not normalizing the print representation is a small step backwards. It also means that we have to sort on every comparison (instead of just when calling `reduce()`). Although I don't know how these are used, so I don't know which could affect speed more. Well, I understand making a minimal change. I just want to mention that the fact that some are genuinely different might be exposing bugs. That might warrant further investigation later.

I agree and I think `FormalSum` in general could use some significant changes. But I think what I've done is the best way to fix #37966 without a major rewrite. I could cache the result of sorting in its own variable in `__init__`, but then that would increase memory usage which is maybe worse than sorting the lists, since sorting is a fairly fast operation.

src/sage/schemes/generic/divisor_group.py
```diff
@@ -190,6 +190,36 @@ def _element_constructor_(self, x, check=True, reduce=True):
         else:
             return Divisor_generic([(self.base_ring()(1), x)], check=False, reduce=False, parent=self)
 
+    def _coerce_map_from_(self, other):
+        r"""
+        Return if there is a coercion map from ``other`` to ``self``.
+
+        There is a coercion from another divisor group if the
+        schemes are equal and there is a coercion map from
```
tscrim:
```suggestion
        There is a coercion from another divisor group if there is
        a coercion map from the schemes and there is a coercion map from
```
Now this needs to be updated slightly, but this should be the last change.

tscrim:
I wouldn't call the sorting as part of the `reduce` a major rewrite, but it can wait for now. Just one more little thing to do.

vincentmacri:
The commit I just pushed updated the warning in `FormalSum` for setting reduce to `False`. With the change to sorting in `richcmp`, there shouldn't be issued comparing sums with different orders, just with different cancellations.

The build from yesterday failed. For `test-new`, the following error occurred: `fatal: not a git repository: /new/.git/worktrees/worktree-image`. I'm not sure what would cause this.

tscrim:
Yea, our CI testing is currently massively broken...

tscrim (approved):
Okay, let's get this in.

PR #37980, opened 2024-05-10
Fix order computation for linear groups GL(n, R) and SL(n, R)

### LINKED ISSUE #37934: Wrong order of finite matrix groups
### Steps To Reproduce

On Sage 10.3 running in CoCalc, run
```
sage: GL(2,Integers(4)).order()
180
sage: GL(2,Integers(1)).order()
0
```

### Expected Behavior

```
sage: GL(2,Integers(4)).order()
96
sage: GL(2,Integers(1)).order()
1
```
First one works in sage version 10.2, second one returns recursion error.

### Actual Behavior

See above. Seems to be computing GL(2,Integers(q)).order() as (q-1)^2*q*(q+1).

### Additional Information

Explanation of correct computation: [here](https://math.stackexchange.com/questions/3357934/order-of-gl2-mathbbz-4).

### Environment

```markdown
- **OS**: CoCalc
- **Sage Version**: 10.3
```

### Checklist

S17A05:
#36881 introduced some wrong results for orders of linear groups since it did not check whether the base ring is a field. We add the necessary checks as well as formulas for some further cases, namely $\mathbb{Z}$ and $\mathbb{Z}/q\mathbb{Z}.$ Finally, we add a `NotImplementedError` for the cases that are not supported.

The formula for the general linear group over $\mathbb{Z}/q\mathbb{Z}$ is taken from https://math.stackexchange.com/a/2044571, and for the order of the special linear group we use the group isomorphism induced by the determinant.

Fixes #37934.


src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+                return ord
+
+            raise NotImplementedError("order computation of linear groups not "
+                                      "fully supported for arbitrary base rings")
+
+        if n > 1 or (R.is_field() and not self._special):
+            return Infinity
+
+        if self._special:
             return 1
```
tscrim:
```suggestion
            return ZZ.one()
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
 
-        if self._special and n == 1:
+        n = self.degree()
+        R = self.base_ring()
+
+        if R.is_finite():
+            q = R.order()
+
+            if q == 1:
+                return 1
```
tscrim:
```suggestion
                return ZZ.one()
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+
+        if n > 1 or (R.is_field() and not self._special):
+            return Infinity
+
+        if self._special:
             return 1
-        return Infinity
+
+        if R == ZZ:
+            return 2
```
tscrim:
```suggestion
            return ZZ(2)
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+
+            sage: GL(1, ZZ).order()
+            2
+        """
+        def order_over_finite_field(q, n, special):
             ord = prod(q**n - q**i for i in range(n))
-            if self._special:
-                return ord / (q-1)
+            if special:
+                ord = ord // (q-1)
```
tscrim:
```suggestion
            if self._special:
                return ord // (q-1)
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+            sage: GL(2, Integers(4)).order()
+            96
+
+            sage: GL(2, Integers(1)).order()
+            1
+
+            sage: GL(1, ZZ).order()
+            2
+        """
+        def order_over_finite_field(q, n, special):
```
tscrim:
```suggestion
        def order_over_finite_field(q, n):
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+        R = self.base_ring()
+
+        if R.is_finite():
+            q = R.order()
+
+            if q == 1:
+                return 1
+
+            if R.is_field():
+                return order_over_finite_field(q, n, self._special)
```
tscrim:
```suggestion
                return order_over_finite_field(q, n)
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+
+            if R == Integers(q):
+                ord = 1
+
+                # By the Chinese remainder theorem we need to build the product
+                # over the orders of GL(n, ZZ/p^e ZZ) (or SL) for all prime
+                # powers in the factorization of q
+                for (p,e) in q.factor():
+                    ord_base = order_over_finite_field(p, n, self._special)
```
tscrim:
```suggestion
                    ord_base = order_over_finite_field(p, n)
```

src/sage/groups/matrix_gps/linear.py
```diff
@@ -340,18 +359,71 @@ def order(self):
...
+                    if not self._special:
+                        ord *= p**((e-1)*n**2) * ord_base
+
+                    # We apply |SL(n, R)| = |GL(n, R)| / euler_phi(q), but since we
+                    # already iterate over the prime factorization of q, we divide
+                    # out euler_phi(q) iteratively. Noting that (p-1) is already
+                    # handled in the call to order_over_finite_field, we only
+                    # need to remove p^(e-1) compared to the above formula
+                    else:
+                        ord *= p**((e-1)*(n**2 - 1)) * ord_base
```
tscrim:
```suggestion
                        ord *= p**((e-1)*(n**2-1)) * ord_base
```
Being consistent about the spacing.

tscrim (changes requested):
Mostly a bunch of little things. We do not want to return Python `int`s. We can take advantage of the closure in Python.  It would be good to add a test verifying the result:
```python
sage: G = GL(2, Integers(6))
sage: G.order() == len(list(G))
True
sage: G = SL(2, Integers(6))
sage: G.order() == len(list(G))
True
```

tscrim (approved):
Thank you. LGTM.

S17A05:
Not sure why the tests failed, but the error makes it seem like a CI issue.

tscrim:
Indeed, something seems borked with it.

PR #37981, opened 2024-05-10
Reduce sugar abuse

mantepse:
We replace `R[x]` for a ring `R` and a string `x` with `PolynomialRing(R, x)` in library code.


kwankyu:
Is it also good for "health" (or performance)?

src/sage/schemes/elliptic_curves/mod_poly.py
```diff
@@ -116,8 +116,9 @@ def classical_modular_polynomial(l, j=None):
         except KeyError:
             pass
 
+        P = PolynomialRing(ZZ, ['X', 'Y'])
```
tscrim, line 119: Since all code paths use this polynomial ring, I would just use it everywhere. Also on line 128, you missed one.

Line 135 also is also an additional one to change.

tscrim:
Yes, this will be faster, both for avoiding redirection and additional checks. Although it will be a micro-optimization.

tscrim (changes requested):

mantepse:
There are very many more.  In particular in `combinat/sf`, `rings/qqbar`, `rings/padics`, etc.  For example, in the latter, I find
```
# Cache some commonly-used polynomial rings
QQx = QQ['x']
QQx_x = QQx.gen()
QQy = QQ['y']
QQy_y = QQy.gen()
QQxy = QQ['x', 'y']
QQxy_x = QQxy.gen(0)
QQxy_y = QQxy.gen(1)
```
Should I be thorough or leave at what we currently have?

What I did is:
```diff
diff --git a/src/sage/categories/rings.py b/src/sage/categories/rings.py
index 9685fb80dd..db2ba1c531 100644
--- a/src/sage/categories/rings.py
+++ b/src/sage/categories/rings.py
@@ -1191,6 +1191,15 @@ class Rings(CategoryWithAxiom):
                   To:   Real Lazy Field
                   Defn: a -> 1.414213562373095?
             """
+            import inspect
+            previous_frame = inspect.currentframe().f_back
+            (filename,
+             line_number,
+             function_name,
+             lines,
+             index) = inspect.getframeinfo(previous_frame)
+            print("sugar baby", filename, line_number, function_name, lines, index)
+
             def normalize_arg(arg):
                 if isinstance(arg, (tuple, list)):
                     # Allowing arbitrary iterables would create confusion,
```

tscrim:
I would be good and change them as well. The more we remove, the more it becomes the standard.

mantepse:
Unfortunately, there are many doctest failures.  What worries me a little: they do not complain about a missing import statement, but rather give completely different results.  Is this because exceptions are not caught properly, or for some other reason?

mkoeppe:
@mantepse Would you mind making an effort to use PR titles that are more descriptive?

tscrim:
> Unfortunately, there are many doctest failures. What worries me a little: they do not complain about a missing import statement, but rather give completely different results. Is this because exceptions are not caught properly, or for some other reason?

Welcome to Sage's import hell. It sounds like you have created a circular import.

mantepse:
I wouldn't be surprised about a circular import, but I am annoyed that there is no exception raised.

mantepse:
> @mantepse Would you mind making an effort to use PR titles that are more descriptive?

Yes.

src/sage/rings/polynomial/complex_roots.py
```diff
@@ -46,6 +46,7 @@
 from sage.rings.qqbar import AA, QQbar
 from sage.arith.misc import sort_complex_numbers_for_display
 from sage.rings.polynomial.refine_root import refine_root
+from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
```
tscrim:
```suggestion
```
My guess is this is causing the circular imports. I would make these imports locally.

src/sage/rings/polynomial/complex_roots.py
```diff
@@ -75,7 +76,7 @@ def interval_roots(p, rts, prec):
     """
 
     CIF = ComplexIntervalField(prec)
-    CIFX = CIF['x']
+    CIFX = PolynomialRing(CIF, 'x')
```
tscrim, line 79:
```suggestion
    from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
    CIFX = PolynomialRing(CIF, 'x')
```

src/sage/rings/polynomial/complex_roots.py
```diff
@@ -268,7 +269,7 @@ def complex_roots(p, skip_squarefree=False, retval='interval', min_prec=0):
     prec = 53
```
tscrim, line 269:
```suggestion
    prec = 53
    from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
```

src/sage/rings/real_lazy.pyx
```diff
@@ -46,6 +46,7 @@ from sage.categories.morphism cimport Morphism
 from sage.rings.ring cimport Field
 import sage.rings.infinity
 from sage.rings.integer import Integer
+from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
```
tscrim:
```suggestion
```
Let's be safe here too.

src/sage/rings/real_lazy.pyx
```diff
@@ -66,7 +67,7 @@ cdef _QQx = None
 cdef QQx():
     global _QQx
     if _QQx is None:
```
tscrim, line 69:
```suggestion
    from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
    if _QQx is None:
```

src/sage/rings/universal_cyclotomic_field.py
```diff
@@ -183,6 +183,7 @@
 from sage.rings.rational_field import QQ
 from sage.rings.infinity import Infinity
 from sage.rings.qqbar import AA, QQbar
+from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
```
tscrim:
```suggestion
```
For safety.

src/sage/rings/universal_cyclotomic_field.py
```diff
@@ -1281,7 +1282,7 @@ def minpoly(self, var='x'):
             construct the polynomial.
         """
         gap_p = libgap.MinimalPolynomial(libgap.eval("Rationals"), self._obj)
-        return QQ[var](QQ['x_1'](str(gap_p)))
+        return PolynomialRing(QQ, var)(PolynomialRing(QQ, 'x_1')(str(gap_p)))
```
tscrim, line 1284:
```suggestion
        from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
        return PolynomialRing(QQ, var)(PolynomialRing(QQ, 'x_1')(str(gap_p)))
```
Plus this import time will be negligible here.

tscrim (changes requested):
The failures you see are the best type of error you're going to get for circular imports. There isn't a way for Python to raise an error for that. At least, that would be something within Python upstream...

mantepse:
Interestingly,
```diff
diff --git a/src/sage/misc/functional.py b/src/sage/misc/functional.py
index c70d78800e0..720d4594646 100644
--- a/src/sage/misc/functional.py
+++ b/src/sage/misc/functional.py
@@ -209,7 +209,8 @@ def cyclotomic_polynomial(n, var='x'):
         sage: cyclotomic_polynomial(11)
         x^10 + x^9 + x^8 + x^7 + x^6 + x^5 + x^4 + x^3 + x^2 + x + 1
     """
-    return ZZ[var].cyclotomic_polynomial(n)
+    from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
+    return PolynomialRing(ZZ, var).cyclotomic_polynomial(n)
 
 
 def decomposition(x):
```
breaks everything.

tscrim:
What that change, I don't see major breakage, but some localized stuff in
```
src/sage/rings/polynomial/omega.py
src/sage/rings/polynomial/weil/weil_polynomials.pyx
```
(Although I didn't test everything with your branch.)

However, there is a slight difference. `ZZ` is special as it implements:
```python
def __getitem__(self, x):
        if x in self:
            return self

        if isinstance(x, NumberFieldElement_base):
            K, _ = parent(x).subfield(x)
            return K.order(K.gen())

        return CommutativeRing.__getitem__(self, x)
```
Furthermore, when the input is not a string, then the `__getitem__` can do something non-trivial, such as constructing an extension. So for those places that pass along inputs (e.g., `cyclotomic_polynomial`), it is not necessarily sugar abuse as the user could give an argument that is not a string. Hence, those changes we should revert.

src/sage/combinat/multiset_partition_into_sets_ordered.py
```diff
@@ -3399,7 +3400,7 @@ def val(self, q='q'):
             (q^2+q+1)*s[2, 1, 1] + q*s[2, 2]
         """
         H = [self._OMPs(list(b)) for b in self.highest_weight_vectors()]
-        Sym = SymmetricFunctions(ZZ[q])
+        Sym = SymmetricFunctions(PolynomialRing(ZZ, q))
```
mantepse, line 3405: So, for example, this one?  Only for `ZZ`, or also when the base ring *could be* `ZZ`?

Maybe we should revert all cases where it is not clear that the argument must be a string?

tscrim: To be safe, all cases where the argument is not guaranteed to be a string. That way we don't introduce any backwards-incompatible changes.

tscrim:
The tester is reporting some failures that look like they come from missing imports:
```
sage -t --long --random-seed=76127206414652364553524590949876643686 src/sage/categories/loop_crystals.py  # 1 doctest failed
sage -t --long --random-seed=76127206414652364553524590949876643686 src/sage/combinat/crystals/littelmann_path.py  # 5 doctests failed
sage -t --long --random-seed=76127206414652364553524590949876643686 src/sage/geometry/polyhedron/library.py  # 4 doctests failed
```

PR #37988, opened 2024-05-12
CI Build&Test: Replace use of `sage -t --new`

### LINKED ISSUE #37985: CI Build&Test: Replace use of `sage -t --new` by explicit file list
The current method, using `git reset` to mark what's new in the PR and then relying on `sage -t --new`, does not handle file deletions/renames correctly, as observed in https://github.com/sagemath/sage/pull/37839#issuecomment-2103562852

Instead we should obtain the list of files changed relative to the base of the PR.

mkoeppe:
Fixes #37985.

We also turn on  `--long --format github` for this run, so that annotations are shown in the "Files changed" tab.
(Previously, only the annotations from the full testsuite run were shown, out of the concern of duplicating the numerous doctest style warnings. But these warnings have been much reduced, and the last two warnings are fixed in #38029.)

Example run: https://github.com/sagemath/sage/actions/runs/9198070834/job/25299950120?pr=37988

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37926 (merged here)

### DIFF
--- a/.ci/write-dockerfile.sh
+++ b/.ci/write-dockerfile.sh
@@ -279,6 +279,7 @@ RUN if [ -d /sage ]; then                                               \
         printf '/src\n!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n!/src/VERSION.txt\n' >> /sage/.gitignore && \
         printf '/src\n!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n!/src/VERSION.txt\n' >> /new/.gitignore && \
         if ! (cd /new && ./.ci/retrofit-worktree.sh worktree-image /sage); then \
+            echo "retrofit-worktree.sh failed, falling back to replacing /sage"; \
             for a in local logs; do                                     \
                 if [ -d /sage/\$a ]; then mv /sage/\$a /new/; fi;       \
             done;                                                       \
@@ -339,6 +340,7 @@ $ADD .gitignore /new/.gitignore
 $ADD src /new/src
 RUN cd /new && rm -rf .git && \
     if ! /sage/.ci/retrofit-worktree.sh worktree-pre /sage; then \
+        echo "retrofit-worktree.sh failed, falling back to replacing /sage/src"; \
         rm -rf /sage/src;                                    \
         mv src /sage/src;                                    \
         cd /sage && ./bootstrap && ./config.status;          \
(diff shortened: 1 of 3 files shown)


.ci/write-dockerfile.sh
```diff
@@ -272,13 +272,14 @@ $ADD src/Pipfile.m4 src/pyproject.toml src/requirements.txt.m4 src/setup.cfg.m4
 $ADD m4 /new/m4
 $ADD pkgs /new/pkgs
 $ADD build /new/build
-$ADD .ci /new/.ci
 $ADD .upstream.d /new/.upstream.d
+ADD .ci /.ci
 RUN if [ -d /sage ]; then                                               \
```
kwankyu, line 277: What's the difference between `$ADD` and `ADD`?

kwankyu: Ah it is `ADD="ADD $__CHOWN"`...

kwankyu: Why here `ADD` and `$ADD` below (line 341)?

.github/workflows/build.yml
```diff
@@ -175,11 +175,20 @@ jobs:
...
+          files_ignore: src/{setup,conftest*}.py
+
       - name: Test changed files (sage -t --new)
+        if: steps.changed-files.outputs.all_changed_files
         run: |
           export MAKE="make -j2 --output-sync=recurse" SAGE_NUM_THREADS=4
           # We run tests with "sage -t --new"; this only tests the uncommitted changes.
-          ./sage -t --new -p4
+          # https://github.com/tj-actions/changed-files?tab=readme-ov-file#outputs-
+          ./sage -t --long --format github -p4 ${{ steps.changed-files.outputs.all_changed_files }}
```
kwankyu, line 191: Modify line 189 as well?

.github/workflows/build.yml
```diff
@@ -175,11 +175,20 @@ jobs:
           ./sage -python -m pytest -c tox.ini -qq --doctest --collect-only || true
         shell: sh .ci/docker-exec-script.sh BUILD /sage {0}
 
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v44
+        with:
+          files: src/**/*.{py,pyx,pxd,pxi,sage,spyx,rst,tex}
+          files_ignore: src/{setup,conftest*}.py
```
kwankyu, line 184: Is `tex` file also tested?

.ci/write-dockerfile.sh
mkoeppe: Thanks for catching this; fixed in f4f0a0ca80f

.github/workflows/build.yml
mkoeppe: I've added a comment

kwankyu: OK. It is status quo...

kwankyu (approved):
OK. LGTM.

mkoeppe:
Thanks!

PR #37991, opened 2024-05-12
The Burnside ring of a group

### LINKED ISSUE #35475: Burnside ring of permutation representations

### Problem Description

The Burnside ring of a group encodes the permutation representations of the group. This could be implemented as a parent in Sage.

### Proposed Solution

Implement a parent class for the Burnside ring of a finite group using GAP under the hood. This should also include a method which attempts to express a given representation in the Burnside ring by first checking that all the character values are positive (a necessary but not sufficient condition for a representation to be a permutation representation) and then attempting to express it as an element of the Burnside ring.

### Alternatives Considered

One could work with symmetric functions directly or by hand/in an ad hoc manner, but implementing a parent class for the Burnside ring of a group is the way that makes sense in Sage.

### Additional Information

_No response_

Newtech66:
Resolves #35475.

We implement the Burnside ring for a group $G$. We provide a few ways to construct elements, either directly as a formal sum of orbit types $[G/H]$ (where $H$ is a subgroup of $G$), or by providing a group action $\alpha: G \times X \rightarrow X$ on some set $X$. Finally, we implement addition and multiplication of ring elements.

Depends on #38371.

### :memo: Checklist

@mantepse


src/sage/rings/burnside.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+
+        EXAMPLES::
+
+            sage: G = SymmetricGroup(3)
+            sage: B = BurnsideRing(G)
+            sage: list(B)
+            [[()], [(2,3)], [(1,2,3)], 1]
+        """
+        return iter(self(H) for H in self._G.conjugacy_classes_subgroups())
```
mantepse: I made a mistake here, confusing the ring for its generators.  So, it should be
```suggestion
    def group(self):
        r"""
        Return the underlying group.

        EXAMPLES::

            sage: G = DiCyclicGroup(4)
            sage: B = BurnsideRing(G)
            sage: B.group()
            Dicyclic group of order 16 as a permutation group
        """
        return self._G

    @cached_method
    def gens(self):
        r"""
        Return the generators of ``self``.

        These are the conjugacy classes of subgroups of the
        underlying group :meth:`group`.

        EXAMPLES::

            sage: G = SymmetricGroup(3)
            sage: B = BurnsideRing(G)
            sage: B.gens()
            ([()], [(2,3)], [(1,2,3)], 1)
        """
        return tuple(self(H) for H in self._G.conjugacy_classes_subgroups())
```

You then also have to replace `for H in B` by `for H in B.gens()` throughout the doctests.

mantepse:
@trevorkarn, I'm adding you here because of your request at https://github.com/sagemath/sage/issues/35475#issuecomment-1507434563

The current branch is a prototype, not yet ready for review.

trevorkarn:
> @trevorkarn, I'm adding you here because of your request at [#35475 (comment)](https://github.com/sagemath/sage/issues/35475#issuecomment-1507434563)
> 
> The current branch is a prototype, not yet ready for review.

Thanks so much @mantepse! Just let me know when it is ready for review.

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __eq__(self, other):
+        return _is_conjugate(self.parent()._G, self._C, other._C)
+
+class ConjugacyClassesOfSubgroups(Parent):
+    def __init__(self, G):
+        self._G = G
+        self._cache = dict() # invariant to subgroups
+        self._names = dict() # dictionary mapping subgroups to names
+        Parent.__init__(self, category=Sets().Finite())
```
mantepse:
```suggestion
        Parent.__init__(self, category=FiniteEnumeratedSets())
```
allows us to get rid of our custom `gens` method

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
+from sage.misc.cachefunc import cached_method
+from sage.structure.unique_representation import UniqueRepresentation
+from sage.structure.parent import Parent
+from sage.structure.element import Element
+from sage.rings.integer_ring import ZZ
+from sage.structure.formal_sum import FormalSum
+from sage.categories.sets_cat import cartesian_product, Sets
```
mantepse:
```suggestion
from sage.categories.finite_enumerated_sets import FiniteEnumeratedSets
from sage.categories.sets_cat import cartesian_product
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
...
+        underlying group :meth:`group`.
+
+        EXAMPLES::
+
+            sage: G = SymmetricGroup(3)
+            sage: B = BurnsideRing(G)
+            sage: B.gens()
+            (B[((),)], B[((2,3),)], B[((1,2,3),)], B[1])
+        """
+        return tuple(self(H) for H in self._G.conjugacy_classes_subgroups())
```
mantepse: can be removed after the above changes :-)

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __hash__(self):
+        return hash(self._C)
+
+    def _repr_(self):
+        name = self.parent()._names.get(self._C, None)
+        if name is None:
+            return repr(self._C.gens())
+        else:
+            return name
```
mantepse:
```suggestion
        return name
```
is slightly preferred by a linter Frederic Chapoton uses.

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
...
+        self._names[H] = name
+        return H
+
+    def _element_constructor_(self, x):
+        if x.is_subgroup(self._G):
+            return self.element_class(self, self._normalize(x))
+        else:
+            raise ValueError("x must be a subgroup of " + repr(self._G))
+
+        raise ValueError(f"unable to convert {x} into {self}")
```
mantepse:
```suggestion
    raise ValueError(f"unable to convert {x} into {self}: not a subgroup of " + repr(self._G))
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
...
+        INPUT:
+
+        - ``x`` - data for an element
+
+        ``x`` can be a subgroup of `G` or a formal sum of such
+        subgroups.
+        """
+        if x in self.base_ring():
+            return x * self.one()
+
```
mantepse:
```suggestion
```
not needed anymore :-)

mantepse:
You are down to 250 lines with more functionality!

You can now use
```
sage -tox -- src/sage/rings/burnside.py
```
to run most (perhaps even all) checks locally.

I think it would be good to have one test checking that the cache is populated only as necessary.  Since `BurnsideRing` inerits from `UniqueRepresentation`, I guess it's best to use a group which doesn't occur in the tests otherwise, and maybe a big one.  For example:
```
sage: G = SymmetricGroup(6)
sage: B = BurnsideRing(G)
sage: X = Subsets(6, 2)
sage: b = B.construct_from_action(lambda g, x: X([g(e) for e in x]), X)
sage: B.basis().keys()._cache
{720: [Symmetric group of order 6! as a permutation group],
 48: [Subgroup generated by [(3,5,4,6), (1,2)(4,5,6)] of (Symmetric group of order 6! as a permutation group)]}
sage: b^2
B[((3,5,4,6), (1,2)(4,5,6))] + B[((5,6), (4,6,5))] + B[((5,6), (3,4), (1,2))]
sage: B.basis().keys()._cache
{720: [Symmetric group of order 6! as a permutation group],
 48: [Subgroup generated by [(3,5,4,6), (1,2)(4,5,6)] of (Symmetric group of order 6! as a permutation group)],
 6: [Subgroup generated by [(5,6), (4,6,5)] of (Symmetric group of order 6! as a permutation group)],
 8: [Subgroup generated by [(5,6), (3,4), (1,2)] of (Symmetric group of order 6! as a permutation group)]}
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,273 @@
...
+        if isinstance(x, list) or isinstance(x, FormalSum):
+            # if x is a list of pairs (coeff, subgroup) or FormalSum
+            # Turn all subgroups into elements of ConjugacyClassesOfSubgroups
+            x = [(self._indices(subgroup), coeff) for coeff, subgroup in x]
+            return self._from_dict(dict(x))
+        elif x.is_subgroup(self._G):
```
mantepse:
```suggestion
        if x.is_subgroup(self._G):
```
I think that this is not so useful, since we can simply construct an element using `sum`.

mantepse:
One other test:
```
sage: G = SymmetricGroup(4)
sage: B = BurnsideRing(G)
sage: X = Subsets(4, 2)
sage: b = B.construct_from_action(lambda g, x: X([g(e) for e in x]), X)
sage: b.tensor(b)
B[((3,4), (1,2), (1,2)(3,4))] # B[((3,4), (1,2), (1,2)(3,4))]
sage: (b.tensor(b))^2
4*B[((3,4), (1,2), (1,2)(3,4))] # B[((3,4), (1,2), (1,2)(3,4))] + 2*B[((),)] # B[((3,4), (1,2), (1,2)(3,4))] + 2*B[((3,4), (1,2), (1,2)(3,4))] # B[((),)] + B[((),)] # B[((),)]
```

mantepse:
I just realised that you can remove `BurnsideRing._element_constructor_` completely, and instead implement `ConjugacyClassesOfSubgroups.__contains__` (which we should provide anyway):
```python
    def __contains__(self, H):
        return H.is_subgroup(self._G)
```

mantepse:
Here are the next steps (in this order), as I see them:
1. add docstrings and doctests to make `sage -tox -- src/sage/rings/burnside.py` pass
2. create a class `MolecularDecompositionPolynomialSymmetricGroupAction` (for lack of a better name), which is the graded algebra $\bigoplus_{n\geq 0} B(\mathfrak S_n)$, where $B$ denotes the Burnside ring, and the multiplication is the Cauchy product.
3. optimize the Burnside ring, most likely by using the functionality in https://docs.gap-system.org/doc/ref/chap41.html
4. define `MolecularDecompositionSymmetricGroupAction` (for lack of a better name) as the `LazyCompletionGradedAlgebra(MolecularDecompositionPolynomialSymmetricGroupAction)`

I'm not completely sure, but I think that `MolecularDecompositionPolynomialSymmetricGroupAction` will again inherit from `CombinatorialFreeModule`.  The basis might be the set of pairs `(n, H)` such that `H` is a subgroup of $\mathfrak S_n$ - I think it is safer to remember `n`.

I am hesitant to call `MolecularDecompositionPolynomialSymmetricGroupAction` just `PolynomialSpecies`, although maybe we should do just that.  The reason for my doubt is that we can do very many things without knowing the decomposition of a group action, that is, its representation in the Burnside ring.  For example:
```
sage: Ep = species.SetSpecies(min=1)
sage: C = species.CycleSpecies()
sage: P = Ep(C)
sage: P.isotype_generating_series()
z + 2*z^2 + 3*z^3 + 5*z^4 + 7*z^5 + 11*z^6 + 15*z^7 + O(z^8)
```
should *not* create *any* conjugacy classes of subgroups - this would take forever.

Newtech66:
> I just realised that you can remove `BurnsideRing._element_constructor_` completely, and instead implement `ConjugacyClassesOfSubgroups.__contains__` (which we should provide anyway):
> 
> ```python
>     def __contains__(self, H):
>         return H.is_subgroup(self._G)
> ```

How does that work?

mantepse:
> > I just realised that you can remove `BurnsideRing._element_constructor_` completely, and instead implement `ConjugacyClassesOfSubgroups.__contains__` (which we should provide anyway):
> > ```python
> >     def __contains__(self, H):
> >         return H.is_subgroup(self._G)
> > ```
> 
> How is that working under the hood?

Not sure what exactly you mean here.

1. `__contains__` is python magic https://docs.python.org/3/reference/datamodel.html#object.__contains__
2. `CombinatorialFreeModule._element_constructor_` uses `__contains__` here:
https://github.com/sagemath/sage/blob/ffbbea9cb2360340f1c5517b590883fdaf02575c/src/sage/combinat/free_module.py#L763-L764

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,229 @@
...
+        CombinatorialFreeModule.__init__(self, base_ring, basis, category=category)
+
+    def __getitem__(self, H):
+        r"""
+        Return the basis element indexed by ``H``. ``H`` must be a subgroup of ``self._G``.
+        """
+        if H.is_subgroup(self._G):
+            return self._from_dict({self._indices(H): 1})
+        else:
+            raise ValueError(f"{H} must be a subgroup of {self._G}")
```
mantepse:
```suggestion
        raise ValueError(f"{H} must be a subgroup of {self._G}")
```

Newtech66:
I propose we implement the conjugacy class renaming like `B[Z3].rename_index("Z3")`. We can make it fail if we try to call it on something that is not a basis element. Any alternative suggestions? Can the function be named better?

mantepse:
> I propose we implement the conjugacy class renaming like `B[Z3].rename_index("Z3")`. We can make it fail if we try to call it on something that is not a basis element. Any alternative suggestions? Can the function be named better?

Maybe, I don't know.  I think we can delay thinking about this until we start to play with species.

src/sage/rings/burnside.py
Newtech66: I removed the error checking here entirely because it gets handled by `ConjugacyClassesOfSubgroups._element_constructor_` when we call `self._indices`.

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        sage: H1 = PermutationGroup([(1,2)])
+        sage: H2 = PermutationGroup([(2,3)])
+        sage: from sage.rings.burnside import _is_conjugate
+        sage: _is_conjugate(G, H1, H2)
+        True
+    """
+    return libgap.eval('fail') != libgap.RepresentativeAction(G, H1, H2)
+
+class ConjugacyClassOfSubgroups(Element):
+    def __init__(self, parent, C):
```
mantepse, line 71:
```suggestion
    def __init__(self, parent, C):
        r"""
        Initialize the conjugacy class.

        TESTS::

            sage: G = SymmetricGroup(4)
            sage: B = BurnsideRing(G)
            sage: Z4 = CyclicPermutationGroup(4)
            sage: TestSuite(B(Z4)).run()
        """
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        Element.__init__(self, parent)
+        self._C = C
+
+    def __hash__(self):
+        r"""
+        Return the hash of the representative of the conjugacy class.
```
mantepse:
```suggestion
        Return the hash of the representative of the conjugacy class.

        TESTS::

            sage: G = SymmetricGroup(3)
            sage: B = BurnsideRing(G)
            sage: H1 = B(PermutationGroup([(1,2)]))
            sage: H2 = B(PermutationGroup([(2,3)]))
            sage: hash(H1) == hash(H2)
            True
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        """
+        name = self.parent()._names.get(self._C, None)
+        return repr(self._C.gens()) if name is None else name
+
+    def __le__(self, other):
+        r"""
+        Return if this element is less or equal to ``other``.
+
+        ``self`` is less or equal to ``other`` if it is conjugate to
+        a subgroup of ``other`` in the parent group.
```
mantepse:
```suggestion
        a subgroup of ``other`` in the parent group.
        
        EXAMPLES:

        We recreate the poset of conjugacy classes of subgroups of
        `S_4`::

            sage: G = SymmetricGroup(4)
            sage: B = BurnsideRing(G)
            sage: P = Poset([B.basis(), lambda b, c: b <= c])
            sage: len(P.cover_relations())
            17
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
+from sage.misc.cachefunc import cached_method
+from sage.structure.unique_representation import UniqueRepresentation
```
mantepse:
```suggestion
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+from sage.structure.parent import Parent
+from sage.structure.element import Element
+from sage.rings.integer_ring import ZZ
+from sage.structure.formal_sum import FormalSum
```
mantepse:
```suggestion
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        return repr(self._C.gens()) if name is None else name
+
+    def __le__(self, other):
+        r"""
+        Return if this element is less or equal to ``other``.
+
+        ``self`` is less or equal to ``other`` if it is conjugate to
+        a subgroup of ``other`` in the parent group.
+        """
+        return libgap.eval('fail') != libgap.ContainedConjugates(self.parent()._G, other._C, self._C, True)
```
mantepse:
```suggestion
        return (isinstance(other, ConjugacyClassOfSubgroups)
                and (GAP_FAIL != libgap.ContainedConjugates(self.parent()._G,
                                                            other._C,
                                                            self._C,
                                                            True)))
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+from sage.structure.element import Element
+from sage.rings.integer_ring import ZZ
+from sage.structure.formal_sum import FormalSum
+from sage.categories.sets_cat import cartesian_product
+from sage.categories.finite_enumerated_sets import FiniteEnumeratedSets
+from sage.groups.perm_gps.permgroup import PermutationGroup
+from sage.libs.gap.libgap import libgap
+from sage.categories.algebras import Algebras
+from sage.combinat.free_module import CombinatorialFreeModule
+
```
mantepse, line 21:
```suggestion

GAP_FAIL = libgap.eval('fail')
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+    EXAMPLES:
+
+        sage: G = SymmetricGroup(3)
+        sage: H1 = PermutationGroup([(1,2)])
+        sage: H2 = PermutationGroup([(2,3)])
+        sage: from sage.rings.burnside import _is_conjugate
+        sage: _is_conjugate(G, H1, H2)
+        True
+    """
+    return libgap.eval('fail') != libgap.RepresentativeAction(G, H1, H2)
```
mantepse:
```suggestion
    return GAP_FAIL != libgap.RepresentativeAction(G, H1, H2)
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        """
+        return libgap.eval('fail') != libgap.ContainedConjugates(self.parent()._G, other._C, self._C, True)
+
+    def __eq__(self, other):
+        r"""
+        Return if this element is equal to ``other``.
+
+        Two elements compare equal if they are conjugate subgroups in the parent group.
+        """
+        return _is_conjugate(self.parent()._G, self._C, other._C)
```
mantepse:
```suggestion
        return (isinstance(other, ConjugacyClassOfSubgroups)
                and _is_conjugate(self.parent()._G, self._C, other._C))
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+
+class ConjugacyClassesOfSubgroups(Parent):
+    def __init__(self, G):
+        r"""
```
mantepse, line 208:
```suggestion
        r"""
        Initialize the set of conjugacy classes.
    
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        INPUT:
+
+        ``G`` -- a group.
+        """
```
mantepse:
```suggestion

        TESTS::

            sage: from sage.rings.burnside import ConjugacyClassesOfSubgroups
            sage: G = CyclicPermutationGroup(4)
            sage: C = ConjugacyClassesOfSubgroups(G)
            sage: TestSuite(C).run()
        """
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        self._G = G
+        self._cache = dict() # invariant to subgroups
+        self._names = dict() # dictionary mapping subgroups to names
+        Parent.__init__(self, category=FiniteEnumeratedSets())
+
```
mantepse:
```suggestion

    def __eq__(self, other):
        if not isinstance(other, ConjugacyClassesOfSubgroups):
            return False
        return self._G == other._G

    def __hash__(self):
        return hash(self._G)
```

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,329 @@
...
+        TESTS::
+
+            sage: G = SymmetricGroup(4)
+            sage: B = BurnsideRing(G)
+            sage: Z4 = CyclicPermutationGroup(4)
+            sage: Z4 in B._indices
+            True
+        """
+        return H.is_subgroup(self._G)
```
mantepse:
```suggestion
        if parent(H) == self:
            return True

        return (isinstance(H, PermutationGroup_generic)
                and H.is_subgroup(self._G))
```

dimpase:
Does this use tables of marks? (With which it's much easier). GAP does have functionality to deal with tables of marks,
and has a large library of them, packaged as tomlib package. (tomlib is a part of Sage)

mantepse:
> Does this use tables of marks? (With which it's much easier). GAP does have functionality to deal with tables of marks, and has a large library of them, packaged as tomlib package. (tomlib is a part of Sage)

Not yet, but eventually we will use it whenever available.  However, as you certainly know, the table of marks works is only available for relatively small groups, e.g. the symmetric groups up to $\mathfrak S_{13}$.  I hope that some operations will be possible even for larger groups, although, of course, they will be very slow.

src/sage/rings/burnside.py
Newtech66: `from sage.structure.element import parent as get_parent_of` was required, then `if get_parent_of(H) == self`.

mantepse: Oh, sorry, yes.  Except the renaming, this is unnecessary.

Newtech66:
Why are we adding `SymmetricGroupAction` in the name? The molecular decomposition of $F[n]$ is defined wrt $\mathfrak{S}_n$, so why the additional qualifier?

mantepse:
> Why are we adding `SymmetricGroupAction` in the name? The molecular decomposition of $F[n]$ is defined wrt $\mathfrak S_n$, so why the additional qualifier?

You are right, `PolynomialMolecularDecomposition` would be better.  I admit, I don't care much about the name for the moment, and I didn't think much about it.

For me, the only question is whether we need this class at all, and I think that we do.

mantepse:
> If we use the `GradedAlgebrasWithBasis` category then `completion`/`formal_series_ring` should be provided automatically.

Yes, that's correct, and that's the whole point of having a class `PolynomialMolecularDecomposition` (or whatever name we use).  Is it clear to you how to implement it?  It will inherit again from `CombinatorialFreeModule`, the only part which might involve work is to get the `basis_keys` right.   Possibly you have to define a class for these, i.e., `ConjugacyClassesOfSymmetricGroupSubgroups` - which is in `InfiniteEnumeratedSets`.

Newtech66:
The naming facility has been reworked. The class `SubgroupNameStore` maintains the subgroup <-> name associations and provides functions to get, set, and unset names. `BurnsideRing` and `PolynomialMolecularDecomposition` now inherit from `SubgroupNameStore` (is this a good idea? We can alternatively create an instance of it in __init__).

The name printing is done in `_repr_` of `ConjugacyClassOfSubgroups`. A `default_name` method is added for returning the default string if no name exists. The `_namegetter` attribute provides the interface between `BurnsideRing/PolynomialMolecularDecomposition` and their `basis_keys`. It is set to None by default, but if the relevant classes are created within one of the above, they are set to `get_name` in `SubgroupNameStore`. This lets `_repr_` work correctly even if the `ConjugacyClassesOfSubgroups(_SymmetricGroup)(_all)` classes are created as standalone objects (but we cannot rename anything).

Newtech66:
Naming facility:

The class `SubgroupStore` provides caching and renaming facilities. The classes `ConjugacyClassesOfSubgroups(_SymmetricGroup_all)` inherit from it. Then we set `_print_options['names'] = self._names` in `BurnsideRing` and `PolynomialMolecularDecomposition` to link changes in `SubgroupStore` to `CombinatorialFreeModule`'s printing system. This means the syntax for naming goes like `self._indices.set_name/get_name/...` which might be somewhat unintuitive.

Newtech66:
One doctest "_test_graded_components" is still failing because the numbers chosen for the degree of the symmetric group are chosen to be ``[0, 1, 3, 42]``. 42 is too large and GAP fails to generate the conjugacy classes of subgroups with the error: ``GAPError: Error, Range: <last> must be a small integer (not a large positive integer)``. Perhaps it will be fixed by #38300?

mantepse:
> One doctest `_test_graded_components` is still failing because the numbers chosen for the degree of the symmetric group are chosen to be `[0, 1, 3, 42]`. 42 is too large and GAP fails to generate the conjugacy classes of subgroups with the error: `GAPError: Error, Range: <last> must be a small integer (not a large positive integer)`. Perhaps it will be fixed by #38300?

You can do `TestSuite(XXX).run(skip="_test_graded_components")` for now.

mantepse:
I would suggest to make this pull request depend on #38371, and store instances of `ConjugacyClassOfSubgroups_SymmetricGroup` always in factored form.

I'm not sure whether this makes sense for Burnside rings of other groups than the symmetric group.

It's also important to me to prepare for conjugacy classes of subgroups of direct products of symmetric groups, corresponding to multivariate (in the book: "multisort") species.  As I detailed in some email, it was wrong of me to believe that we could get multivariate species as tensor products of univariate species.

I hope that this will not be difficult, although some thought should go into the design.  One thing I did not figure out yet: conjugacy classes of subgroups of $S_m \times S_n$ are essentially the same as the conjugacy classes of $S_n \times S_m$.  Can - and should - the code take advantage of that?

dimpase:
it's not realistic to compute the conjugacy classes of subgroups of $S_{42}$, these are way too many

Newtech66:
This PR might be becoming too large, I think we should keep the scope of this PR to BurnsideRing and PolynomialMolecularDecomposition as it is right now and build it further in another PR.

mantepse:
> This PR might be becoming too large, I think we should keep the scope of this PR to BurnsideRing and PolynomialMolecularDecomposition as it is right now and build it further in another PR.

Sorry, I probably miscommunicated.  I am *not* suggesting to add the multivariate part now (although this certainly should be added within the gsoc project).  All I'm saying is that this should be kept in mind, so we don't take design decisions we regret later.

Storing the (conjugacy classes of) subgroups in factored form should not add much code.  In fact, it should remove some, because multiplication becomes trivial.

Does this make more sense?

Newtech66:
Why do we want to do the factorisation within ConjugacyClassOfSubgroups? I was thinking I could make a separate element class for PolynomialMolecularDecomposition named AtomicDecomposition that would handle the factorisations. We could handle printing and lazily computing factorisations better that way by separating responsibilities.

mantepse:
> Why do we want to do the factorisation within ConjugacyClassOfSubgroups? I was thinking I could make a separate element class for PolynomialMolecularDecomposition named AtomicDecomposition that would handle the factorisations. We could handle printing and lazily computing factorisations better that way by separating responsibilities.

OK, I'm not sure.  I would have thought that it makes sense because the factorisation is very fast, and it is also easy to obtain the product from the factored object.  I thought it would be easier to only provide the factored objects, but of course, I may be mistaken.

I think I have mentioned before that, from a combinatorialists point of view, one will probably mostly want to use custom names for atomic species.  These are well-established, like $\mathcal E_n$ for the trivial action (perhaps with the exception of $\mathcal E_1 = X$), then $\mathcal C_n$ for the cyclic action, etc.  Does this fit well into your design?  Explicitly, it would be great to be able to do something like the following:
```
sage: for n in range(2, 100):
....:    AtomicConjugacyClass(PermutationGroup([], domain=list(range(1, n+1)))).rename("E_%s" % n)
```
and then get, automatically, output like
```
sage: PolynomialMolecularDecomposition().construct_from_action(a, X)
E_2 * E_4
```
(modulo many typos and omissions on my part)

mantepse:
To make it completely clear: I think it would be really good if the design would allow us to use the standard renaming facility of sage.  To make this work I think it would be sufficient to have a class `ConjugacyClassOfConnectedSubgroups`, such that a `ConjugacyClassOfSubgroups` is (essentially) a `Factorization` object of these.

If I made no mistake, `connected` makes sense for subgroups of arbitrary permutation groups.  In particular, for the multisort species, the atomic species are precisely the connected subgroups of $\mathfrak S_n$ which are subgroups of $\mathfrak S_{n_1}\times\dots\times\mathfrak S_{n_d}$.

Newtech66:
It's taking me a while to go through everything, but we can do something like this:

We have the set of atomic species. We could directly say we have the commutative ring on this set for characterising all species.

We can also take a more explicit approach. We can define the free commutative monoid on this set as the molecular species. Then finally the free module on that is the decomposition of any species.

mantepse:
Might the second approach be more practical? I think to implement a commutative ring you have to implement everything yourself, whereas the free module does  some things, like repr, for you.

dimpase:
Is "connected subgroup" a standard terminology, or something invented on the fly?

I, frankly, don't like it. By right, it means that there is an underlying topology in which this set is connected (this is how this term is used in areas such as algebraic group theory).
If there is no such topology, I would consider a renaming.

mantepse:
> Is "connected subgroup" a standard terminology, or something invented on the fly?

It was chosen by Götz Pfeiffer and Liam `Naughton,` see page 13 of https://cs.uwaterloo.ca/journals/JIS/VOL16/Naughton/naughton2.pdf and also https://oeis.org/A005226

I think that the terminology is now mostly established.  I asked for a more standard name on https://mathoverflow.net/q/329927.  If you do have a better name, please share!

> I, frankly, don't like it. By right, it means that there is an underlying topology in which this set is connected (this is how this term is used in areas such as algebraic group theory). If there is no such topology, I would consider a renaming.

I think it makes sense from a combinatorial point of view, when regarding permutation groups as a monoid with respect to the direct product.  For example, it is quite standard to call a Dyck path connected if it does not decompose as concatenation of shorter Dyck paths.

dimpase:
why not call it "indecomposable" then, instead of "connected"?

Newtech66:
Yes, it is the same.

mantepse:
> Yes, it is the same.

Apparently, one has to be very careful with terminology, though:

https://math.stackexchange.com/questions/913510/is-a-primitive-permutation-group-indecomposable

dimpase:
Chang-Jefferson's paper talks about disjoint direct product (d.d.p.) decomposition, and d.d.p. indecomposablility.

They don't even cite the paper where "connected" is defined.

mantepse:
I agree.  The drawback is that `indecomposable group` would not be the same as `indecomposable permutation group`.

Newtech66:
I understood my mistake. Primitivity and connectedness are defined wrt to the _group action_ of a group's permutation representation. Whereas [https://groupprops.subwiki.org/wiki/Directly_indecomposable_group](https://groupprops.subwiki.org/wiki/Directly_indecomposable_group) is a property of the abstract group.

For example, $G=[(),(1,2)(3,4),(3,4)(5,6),(1,2)(5,6)]$ is a connected permutation group. But, it is isomorphic to $Z\_2 \times Z\_2$, so it is a directly decomposable as an abstract group.

mantepse:
Well, that doesn't really count as mistake - I interpreted your comment as "as permutation groups" :-)

dimpase:
> I agree.  The drawback is that `indecomposable group` would not be the same as `indecomposable permutation group`.

Chang-Jefferson mean permutation groups in their definition. 

IMHO it's Naughton-Pfeiffer who deviated here with their "connected" from the usual maths practice of using "connected" in the topological sense.

Newtech66:
> sage: PolynomialMolecularDecomposition().construct_from_action(a, X)
> E_2 * E_4

What is the minimal input we need to work with something like this? We need to establish the number of sorts $k$, and then we can index the molecular/atomic species concentrated on the total cardinality $n$ using `IntegerVectors(n,k)`. So like, to establish any decomposition, at the very least we need the number of sorts.

So...pass the multicardinality and pass the group and we can find the atomic decomposition. But I'm still stuck because, let's say we have $G=[(1,2)(3,4),(3,4)(5,6),(1,2)(5,6)]$ as a subgroup of $S\_2 \times S\_2 \times S\_2$. Then how are we going to represent it? More generally, the parts produced by the disjoint direct product decomposition might have have overlaps across each sort. For example, $G=[(1,2)(3,4),(5,6),(7,8)(9,10)]$ as a subgroup of $S\_2 \times S\_6 \times S\_2$. I don't understand what to do in this case.

Newtech66:
Ok, I think I get it now. The multicardinality is determined by the number of points from each sort that is present in the domain. We could print something like `E_3(X,Y)[2,1]` `E_2(X,Y)[1,1]` `E_2(Y)[0,2]`. The atomic decomposition of the second example above, with `n=[2,6,2]`, after atomic decomposition would be like `[2,2,0]*[0,2,0]*[0,2,2]`. (excuse the notation)

Maybe we could also provide functionality for permuting the sets, like going from `E_3(X,Y)[2,1]` to `E_3(X,Y)[1,2]`.

mantepse:
> > sage: PolynomialMolecularDecomposition().construct_from_action(a, X)
> > E_2 * E_4
> 
> What is the minimal input we need to work with something like this? We need to establish the number of sorts k, and then we can index the molecular/atomic species concentrated on the total cardinality n using `IntegerVectors(n,k)`. So like, to establish any decomposition, at the very least we need the number of sorts.

Sorry, there must have been a misunderstanding: my example was intended as a univariate decomposition.

> Ok, I think I get it now. The multicardinality is determined by the number of points from each sort that is present in the domain. We could print something like `E_3(X,Y)[2,1]` `E_2(X,Y)[1,1]` `E_2(Y)[0,2]`. The atomic decomposition of the second example above, with `n=[2,6,2]`, after atomic decomposition would be like `[2,2,0]*[0,2,0]*[0,2,2]`. (excuse the notation)

There is no such thing as `E_3(X,Y)`, because `E_3` is a univariate species.  The notation for multivariate species is to use `X`, `Y`, `Z`, etc. as singleton species of different sorts.  From these we can create many multivariate species using composition, addition, multiplication, etc.

For example, `X*Y` is a species on two sorts.  Let `U` and `V` be arbitrary finite sets.  Then `X*Y[U, V]` is empty unless `U` and `V` both have exactly one element.  In this case, there is exactly one structure, which you can visualize as a thing with one `X`-label and one `Y`-label.  Note that `X*Y` is molecular but not atomic, because it is the product of `X` and `Y`.

`X+Y` is not even molecular, because it is a sum.

`E_2(X*Y)[U, V]` has no structures except if `U` and `V` contain exactly two elements each.  In this case, there are two structures, e.g., $E_2(X Y)[{1,2},{a,b}]$ consists of the set $1a,2b$ and the set $1b,2a$.

`E_2(X+Y)[U, V]` has no structures except is `U` and `V` have exactly two elements together.

Newtech66:
Oh ok, so `E_2(X*Y)` is the composition of `E_2` and `X*Y`. I had to refer to page 104 of the book.
In the book, page 108 eq(25), they state $C_{alt}(X,Y) = C(X \cdot Y)$. How so? The order of elements in $X \cdot Y$ is not fixed, right?

mantepse:
> Oh ok, so `E_2(X*Y)` is the composition of `E_2` and `X*Y`. I had to refer to page 104 of the book. In the book, page 108 eq(25), they state Calt(X,Y)=C(X⋅Y). How so? The order of elements in X⋅Y is not fixed, right?

Yes it is fixed!  `X\cdot Y[1;a]` contains exactly one structure: `1a`. The `1` is of sort `X` and the `a` is of sort `Y`.

Newtech66:
> To make it completely clear: I think it would be really good if the design would allow us to use the standard renaming facility of sage. To make this work I think it would be sufficient to have a class `ConjugacyClassOfConnectedSubgroups`, such that a `ConjugacyClassOfSubgroups` is (essentially) a `Factorization` object of these.
> 
> If I made no mistake, `connected` makes sense for subgroups of arbitrary permutation groups. In particular, for the multisort species, the atomic species are precisely the connected subgroups of Sn which are subgroups of Sn1×⋯×Snd.

Does using Sage's renaming system make sense? It operates on individual objects, like:
```
sage: Factorization([(2,1),(3,2)])
2 * 3^2
sage: a=Factorization([(2,1),(3,2)])
sage: a.rename('test1'); a
test1
sage: b=Factorization([(2,1),(3,2)]); b
2 * 3^2
sage: b.rename('test1'); b
test1
```

Newtech66:
We can do something like this for supporting multivariate species:

```
AtomicConjugacyClasses -> Infinite enumerated set of atomic conjugacy classes of S_[n_1,...n_k] where k in N
AtomicConjugacyClasses_k -> Infinite enumerated set of atomic conjugacy classes of S_[n_1,...n_k] where k is fixed
AtomicConjugacyClasses_nk -> Infinite enumerated set of atomic conjugacy classes of S_[n_1,...n_k] where k is fixed and n=sum(n_i) is fixed
AtomicConjugacyClass -> an atomic conjugacy class with [n_1,...,n_k] (and therefore also k) is specified

UnivariateAtomicConjugacyClasses (alias of AtomicConjugacyClasses(k=1))

class UnivariateMolecularConjugacyClasses(IndexedFreeAbelianMonoid)
IndexedFreeAbelianMonoid.__init__(indices=UnivariateAtomicConjugacyClasses,prefix='At')
def _element_constructor_ -> decompose input with DDPD and use super()._element_constructor_

PolynomialAtomicDecomposition (renamed PolynomialMolecularDecomposition) = CombinatorialFreeModule(basis_keys=UnivariateMolecularConjugacyClasses)
```

mantepse:
> Does using Sage's renaming system make sense? It operates on individual objects, like:

I think in this case, it does.  We want to give names to individual directly indecomposable groups up to conjugacy.  We want that these are unique objects anyway.

I admit that it might also be nice to be able to give a name to certain products, but I don't think that's going to work, and I don't think it is too important.

Newtech66:
I meant to say, in the `Factorisation` example I gave, `a` and `b` have identical values. Are we going to cache and return an object that we already constructed? That is, say we made `a`, and when we try to construct `b`, we will see if `a` already exists and we will just return that?

Newtech66:
I think we should actually use `CachedRepresentation`, because our `AtomicConjugacyClass` can have several different representatives in the conjugacy class. Then we can do some argument normalization like here: [https://doc.sagemath.org/html/en/reference/structure/sage/structure/unique_representation.html#normalising-the-arguments](https://doc.sagemath.org/html/en/reference/structure/sage/structure/unique_representation.html#normalising-the-arguments)

mantepse:
Sorry, I just realized we need neither, because the '_element_constructor_' of the parent can make sure it always returns the same object.

Newtech66:
What other work needs to be done here?

mantepse:
I'm slightly short of time this week, as announced, I'm sorry for that!

Could you double check whether the classes `UnivariateAtomicConjugacyClasses` and perhaps also `UnivariateMolecularConjugacyClasses` are really necessary?

The reason I'm asking is as follows: being directly indecomposable is an *intrinsic* property of a permutation group - it does not depend on an ambient symmetric group.

In particular:
* the "univariate" directly indecomposable permutation groups are *precisely all* directly indecomposable permutation groups,
* the "d-ary" directly indecomposable permutation groups are *precisely* pairs, consisting of
    - a set partition of the domain $\\{X_1,\dots,X_d\\}$ with $X_1\cup\dots\cup X_d = \\{1,\dots,n\\}$, and
    - a directly indecomposable permutation group, which is a subgroup of $S_{X_1}\times\dots\times S_{X_d}$.
 
This approach should also generalize to variants of species which are not part of this project.

Newtech66:
You're right...we can do better. (that is a brilliant observation...how do I make such observations myself...)

`AtomicConjugacyClass` can store 2 things: the permgroup and the multicardinality (we will normalize any input beforehand such that the domains are of the form [1...a] [a+1...b] ... [...n]). That takes care of the multivariate aspect.

We can get rid of `UnivariateMolecularConjugacyClasses` and merge its functionality into other classes.

Then, we do need `AtomicConjugacyClasses` for basis keys of `PolynomialMolecularDecomposition`, but then do we want to allow mixing $k$-variate species together? Like having univariate and bivariate species in the same sum? That doesn't feel like it makes sense, so for both classes we should add a variable like `k` as in `P=PolynomialMolecularDecomposition(k=2)` to denote that this is the ring of $k$-variate species. Input can be of the form `(permgroup, domains_list)` where `domains_list` must be something that can be turned into a valid `SetPartition` of `[1..n]` for some $n \geq$ `permgroup.degree()`.

Newtech66:
> AtomicConjugacyClass should just store a directly indecomposable permutation group, normalized. More precisely, the parent class AtomicConjugacyClasses has a cache and an _element_constructor_, that normalizes the input and puts it into the cache if it's not already there

Maintaining a cache here doesn't sound useful: we can only interpret it as a species and have some naming functionality and caching if we also have the set partition of its domain (how else can we distinguish the singleton species $X$ and $Y$ in the bivariate case, for example)? Instead we should have a cache in `MolecularConjugacyClasses`.

mantepse:
> > AtomicConjugacyClass should just store a directly indecomposable permutation group, normalized. More precisely, the parent class AtomicConjugacyClasses has a cache and an _element_constructor_, that normalizes the input and puts it into the cache if it's not already there
> 
> Maintaining a cache here doesn't sound useful: we can only interpret it as a species and have some naming functionality and caching if we also have the set partition of its domain (how else can we distinguish the singleton species X and Y in the bivariate case, for example)? Instead we should have a cache in `MolecularConjugacyClasses`.

I probably misled you with the name: I would not have used `AtomicConjugacyClass` for species, but rather for permutation groups.  So, I probably should have called it `ConjugacyClassOfDirectlyIndecomposableSubgroups` or whatever.  Thinking about it, probably the class `ConjugacyClassOfSubgroups` will also not be necessary.

Only the class `PolynomialMolecularDecomposition` should "interpret" a subgroup as a species, as follows:  For simplicity, suppose that we consider bivariate species, and suppose that the species `F` is atomic.  It is represented as a pair `(G_F, P)` where `G_F is an instance of `ConjugacyClassOfDirectlyIndecomposableSubgroups` (normalized, in particular with domain $1,\dots,n$) and `P` is an assignment of each element in the domain to one of the two "variables".  It was wrong of me to call this a set partition, because we need to remember which elements belong to which variable.

For example, `X` could be represented as $(\mathfrak S_1, (\\{1\\}, \\{\\}))$, and `Y` as $(\mathfrak S_1, (\\{\\}, \\{1\\}))$, and $\mathcal E_2(XY)$ as $([(1,3)(2,4)], (\\{1,3\\}, \\{2,4\\}))$. 

For a further example, suppose now that we look at univariate species, then $([(1,3)(2,4)], (\\{1,2,3,4\\}))$ would instead describe the species $\mathcal E_2(X^2)$, which is (of course!) also atomic.

mantepse:
Slightly puzzled as to which time of the day you happen to work :-)

Newtech66:
> It was wrong of me to call this a set partition, because we need to remember which elements belong to which variable.

I realise now that what I was calling as a "set partition" was actually in my mind an "ordered set partition". Which is also implemented in Sage: [Ordered Set Partitions](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/set_partition_ordered.html)

Newtech66:
I created a class `ConjugacyClassOfDirectlyIndecomposableSubgroups`, and `AtomicSpecies` (which pairs the conjugacy class with an integer vector) and `MultivariateAtomicSpecies` (the set of $k$-variate atomic species), and `PolynomialMolecularDecomposition` (now the ring of $k$-variate virtual species) now has `basis_keys=IndexedFreeAbelianMonoid(MultivariateAtomicSpecies)` (no more explicit molecular species class).

Newtech66:
@trevorkarn You can review this PR.

mantepse:
I think we should not rush to move this into sagemath, I would rather wait until the design of #38446 is fixed.

trevorkarn:
> @trevorkarn You can review this PR.

Thanks @Newtech66 I'll take a look!

trevorkarn:
I'm having trouble viewing the documentation. Did you add the file as described in https://doc.sagemath.org/html/en/developer/sage_manuals.html#section-add-file?

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,526 @@
...
+        The Burnside ring of the group ``G``.
+
+        INPUT:
+
+        ``G`` -- a group.
+        ``base_ring`` -- the ring of coefficients. Default value is ``ZZ``.
+
+        EXAMPLES::
+
+            sage: G = SymmetricGroup(6)
```
trevorkarn: It might be good to have some examples where `G` is not a `SymmetricGroup` just as a test

src/sage/rings/burnside.py
```diff
@@ -0,0 +1,526 @@
+from sage.misc.cachefunc import cached_method
```
trevorkarn, line 11: It could be nice to have a little more documentation for the whole file (as opposed to documentation for each class or method only)

Newtech66:
> I'm having trouble viewing the documentation. Did you add the file as described in https://doc.sagemath.org/html/en/developer/sage_manuals.html#section-add-file?

Oops, added it now. `en/reference/rings/sage/rings/burnside`

trevorkarn:
> Documentation preview for this PR

Thanks! It looks like there is still a "missing title". (Check out https://doc-pr-37991--sagemath.netlify.app/html/en/reference/rings/sage/rings/burnside.)

I am also wondering if there is a better place to put this. Of course I think Burnside rings are important, but probably they are not as important as (say) general quotient rings.  Maybe it would be better to put it inside the `combinat`reference? Just a thought.

Could you also add some tests for cases where the base ring is not `ZZ`? Maybe `QQ` and `ZZ.quo(6)` just as an example?

dimpase:
Burnside rings belong to group theory (or indeed rings). It's more of a coincidence that they have applications in combinatorics

mantepse:
The Burnside ring of the symmetric group has been implemented, with many improvements, in #38446 and #38371.

trevorkarn:
> The Burnside ring of the symmetric group has been implemented, with many improvements, in #38446 and #38371.

Are you suggesting that those PRs supersede this one? Or does this one complement them?

mantepse:
> > The Burnside ring of the symmetric group has been implemented, with many improvements, in #38446 and #38371.
> 
> Are you suggesting that those PRs supersede this one?

The code in this PR is mostly superseded, I would say.  I don't know whether it would be good to have a class for the Burnside ring of other groups.

PR #37900, opened 2024-04-29
Restructure sage.*.all for modularization, replace relative by absolute imports

saraedum:
This PR recreates https://github.com/sagemath/sage/pull/36676 by @mkoeppe after it was reverted, see https://github.com/sagemath/sage/pull/37796.

---

We restructure the `all.py` files for modularization.

Guided by the technical constraints outlined in https://groups.google.com/g/sage-devel/c/ozh-7ZZ848s, #35095 defines distribution packages (pip-installable packages) **sagemath-{brial, combinat, eclib, flint, gap, giac, glpk, graphs, groups, homfly, lcalc, libbraiding, libecm, linbox, modules, mpmath, ntl, pari, plot, polyhedra, schemes, singular, standard-no-symbolics, symbolics}**.

When a namespace package such as `sage.misc` is filled by modules from several distribution packages, we create modules named:
- `src/sage/misc/all__sagemath_environment.py` 
- `src/sage/misc/all__sagemath_objects.py`
- `src/sage/misc/all__sagemath_repl.py`

Import statements are moved from `src/sage/misc/all.py` to these files as appropriate, and `src/sage/misc/all.py` imports `*` from there.

Also some imports are replaced by lazy imports.

The new files provide the top level namespaces for the modularized distribution packages, thus [enabling modularized testing](https://doc.sagemath.org/html/en/developer/packaging_sage_library.html#testing-distribution-packages).

This design was introduced in #29865 (merged in Jan 2022, early in the Sage 9.6 development cycle).

- Copied from #35095.
- Part of #29705

Moreover, applied a one-line command to replace relative by absolute imports, thus complementing #36666, which does not touch `.all*` files.

The changes to other files in `sage.modules` etc. come from the PRs #36597, #36572, #36588, #36589 merged here and do not need review.

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

### :hourglass: Dependencies

- Depends on #36597 (merged here to resolve merge conflict)
- Depends on https://github.com/sagemath/sage/pull/36572 (merged here to resolve merge conflict)
- Depends on https://github.com/sagemath/sage/pull/36588 (merged here to resolve merge conflict)
- Depends on https://github.com/sagemath/sage/pull/36589 (merged here to resolve merge conflict)
[...]

PR #37993, opened 2024-05-12
`sage.calculus.expr`: Split out from `sage.calculus.all`

mkoeppe:
Cherry-picked from #36676/#37900.

### :memo: Checklist


mkoeppe:
Split out from https://github.com/sagemath/sage/pull/36676 / https://github.com/sagemath/sage/pull/37900, previously reviewed 6 months ago.

src/sage/calculus/expr.py
```diff
@@ -0,0 +1,203 @@
...
+        Traceback (most recent call last):
+        ...
+        TypeError: unable to convert <function function_with_keyword_only_arg at 0x...>
+        to a symbolic expression
+    """
+    if isinstance(x, Expression):
+        return x
+    elif hasattr(x, '_symbolic_'):
+        return x._symbolic_(SR)
+    elif isinstance(x, (tuple, list, FreeModuleElement)):
```
jhpalmieri, line 173: I'm curious, why move away from using `is_FreeModuleElement` but keep using `is_Matrix` (line 181)?

jhpalmieri: (Otherwise this looks fine to me.)

mkoeppe: Could as well also replace `is_Matrix`, but that is defined (along with the ABC `Matrix`) in the depths of `sage.structure.element`, which is shipped already with the lowest-level distribution, **sagemath-objects**, so it is not relevant for the modularization.

mkoeppe: ... also note that in the systematic deprecation of `is_...` functions, I have taken care of parent classes primarily so far.

jhpalmieri: Thanks for the explanation, that makes sense.

jhpalmieri:
Let's merge it.

mkoeppe:
Thanks again.

PR #37995, opened 2024-05-12
Remove interactive confirmation for installing experimental packages

### LINKED ISSUE #35670: sage build stops waiting for response

### Did you read the documentation and troubleshoot guide?

### Environment

```markdown
- **OS**: Ubuntu 22.04
- **Sage Version**: 9.8, 10.0 in fact all versions I have built.
```

### Steps To Reproduce

I am not sure if this is a bug or intended When building sage, the build process stops after installing a component with output likes

"Finished installing x".  To continue the user have to press return,

then sage will output

"Ok, will install y"

### Expected Behavior

build should continue without requiring user input. It takes a long time for sage to build, I would like it to just type "make" and let it run so I could do something else  instead of having to wait around to press return and wait for

"ok"

### Actual Behavior

See above

### Additional Information

_No response_

mkoeppe:
This removes some complexity and eliminates the pitfall that Sage is waiting for confirmation from the user but the query is hidden in long terminal output from other package installations. 
With this PR we just exit with an error and tell users what to do.

Fixes #35670.

### :memo: Checklist

### DIFF
--- a/build/bin/sage-spkg
+++ b/build/bin/sage-spkg
@@ -363,32 +363,10 @@ WRAPPED_SCRIPTS="build install preinst pipinst postinst $INSTALLED_SCRIPTS"
 warning_for_experimental_packages() { ############################
 if [ x`cat "$PKG_SCRIPTS/type"` = x"experimental" ]; then
     if [ $YES != 1 -a -n "$PKG_NAME_UPSTREAM" ]; then
-        # We use /dev/tty here because our output may be redirected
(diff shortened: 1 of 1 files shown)


mkoeppe:
> Is it worth keeping the old warning "This probably won't work at all for you" (etc.)?

I don't think this kind of warning adds anything

jhpalmieri:
Okay, positive review.

mkoeppe:
Thanks!

PR #38003, opened 2024-05-14
Sheaf cohomology for projective schemes (alternative implementation)

tscrim:
Alternative implementation of #37477 based on that branch.

### :memo: Checklist


tscrim:
I need to add a few more doctests, but I will do so tomorrow.

tscrim:
Okay, this is now ready.

I came across that the resolutions were not hashable with a good comparison, so I had to implement those.

I removed most of the marcos as they could conflict with other reasonable macros (e.g., the octonians or positive integers). IMO it is best to just write out a few extra characters in the docstrings.

PR #38005, opened 2024-05-15
Fix Help menu breakdown in external jupyterlab case

kwankyu:
#37878 introduced a local documentation server for jupyterlab. But it missed a case for which the Help menu of jupyterlab may break down.

In the setup explained in https://doc.sagemath.org/html/en/installation/launching.html#setting-up-sagemath-as-a-jupyter-kernel-in-an-existing-jupyter-notebook-or-jupyterlab-installation, external jupyterlab is launched and connected to sage kernel. In this case, the local doc server does not run, but the Help menu of the jupyterlab still expects a local doc server, and thus the Help menu does not work well.

In this PR, local doc server is detected by testing the port number, and the Help menu uses the online official documentation if local doc server is not running.

### :memo: Checklist

### DIFF
--- a/src/sage/repl/ipython_kernel/kernel.py
+++ b/src/sage/repl/ipython_kernel/kernel.py
@@ -108,14 +108,13 @@ class SageKernel(IPythonKernel):
         # src/bin/sage-notebook.
 
         from sage.env import SAGE_DOC_SERVER_URL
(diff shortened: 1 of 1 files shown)


kwankyu:
I cannot test this PR because my sage is broken...

mkoeppe:
Can't test it either, but LGTM.

kwankyu:
Thanks!

PR #38013, opened 2024-05-16
`DualMatroid`: Add `is_valid` and `relabel` methods

gmou3:
Implement class-specific `is_valid` and `relabel` methods.
Previously, such functionality was achieved through the generic and very slow rank-based algorithms.

The PR also does a touch-up of the docstrings of `dual_matroid.py`.

### DIFF
--- a/src/sage/matroids/dual_matroid.py
+++ b/src/sage/matroids/dual_matroid.py
@@ -507,13 +507,13 @@ class DualMatroid(Matroid):
         r"""
         Return an isomorphic matroid with relabeled groundset.
 
-        The output is obtained by relabeling each element ``e`` by
+        The output is obtained by relabeling each element `e` by
         ``mapping[e]``, where ``mapping`` is a given injective map. If
-        ``e not in mapping`` then the identity map is assumed.
+        ``mapping[e]`` is not defined, then the identity map is assumed.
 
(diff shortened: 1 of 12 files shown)


src/sage/matroids/dual_matroid.py
```diff
@@ -526,3 +502,64 @@ def __reduce__(self):
...
+        Return an isomorphic matroid with relabeled groundset.
+
+        The output is obtained by relabeling each element ``e`` by
+        ``mapping[e]``, where ``mapping`` is a given injective map. If
+        ``e not in mapping`` then the identity map is assumed.
+
+        INPUT:
+
+        - ``mapping`` -- a python object such that `mapping[e]` is the new
+          label of `e`
```
tscrim:
```suggestion
        - ``mapping`` -- a Python object such that ``mapping[e]`` is the new
          label of ``e``
```

tscrim (approved):
LGTM modulo one little thing.

src/sage/matroids/dual_matroid.py
gmou3: I think that `e` is better than ``e`` (change it elsewhere too).

tscrim:
LGTM. Thanks.

PR #38017, opened 2024-05-17
Installation guide: Do not recommend `--prefix=$CONDA_PREFIX`

mkoeppe:
Sage packages can clobber installed non-Python conda packages.
Instead we change it to recommend `--with-sage-venv=$CONDA_PREFIX` and to put SAGE_LOCAL in a subdirectory of the CONDA_PREFIX.

### :memo: Checklist

### DIFF
--- a/src/doc/en/installation/conda.rst
+++ b/src/doc/en/installation/conda.rst
@@ -70,7 +70,7 @@ Using conda to provide system packages for the Sage distribution
 If Conda is installed (check by typing ``conda info``), one can install SageMath
 from source as follows:
 
-  - Create a new conda environment including all standard packages
+-   Create a new conda environment including all standard packages
     recognized by sage, and activate it::
 
       $ conda env create --file environment-3.11-linux.yml --name sage-build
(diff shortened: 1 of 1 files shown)


kwankyu:
While we are at it, would you unindent the surrounding block (which is shown as literal block, unintentionally I think) here and below?

kwankyu:
Sorry but would you explain how "clobber" happens and why the change in this PR solves the problem?

kwankyu:
Now 
```
                    --with-sage-venv=$CONDA_PREFIX 
```
seems reasonable to me. 

Is there an argument to prefer 
```
                    --prefix=$CONDA_PREFIX/libexec/sage-local
```
than other choices, such as `$CONDA_PREFIX/share/sage/local`?

kwankyu:
> Sorry but would you explain how "clobber" happens and why the change in this PR solves the problem?

I think I can answer this. Previously sage installs files into `$CONDA_PREFIX/share`, and these files may conflict with files that conda itself manages.

> `$CONDA_PREFIX/share/sage/local` ?

Then this also could be problematic...

kwankyu:
We may just create a new directory instead of treading into conda areas. How about 
```
--prefix=$CONDA_PREFIX/sage/local
```
?

PR #38019, opened 2024-05-17
find integral solutions to matrix equations over PIDs

yyyyx4:
Even for matrices explicitly defined over ℤ, the `.solve_left()` and `.solve_right()` methods may return solutions over ℚ. In this patch, we add an optional flag `integral=` to these methods which forces the computation of an integral solution if one exists (and throws an exception otherwise).

The algorithm is completely standard, using the (pre-existing) `.smith_form()` method, but so far users had to implement the rest of the solution method on their own each time.

Computing integral solutions to underdetermined systems is a common task when working with lattices.


src/sage/matrix/matrix2.pyx
```diff
@@ -1071,6 +1071,57 @@ cdef class Matrix(Matrix1):
...
+        EXAMPLES::
+
+            sage: A = matrix(ZZ, 5, 7, [(1 + x^4) % 55 for x in range(5*7)]); A
+            [ 1  2 17 27 37 21 32]
+            [37 27 17 46 12  2 17]
+            [27 26 32 32 37 27  6]
+            [ 2 12  2 17 16 37 32]
+            [32 37 16 17  2 12  2]
+            sage: y = vector(ZZ, [-4, -1, 1, 5, 14, 31, 4])
+            sage: A.solve_left(y, integral=True)  # implicit doctest
```
mkoeppe:
```suggestion
            sage: A.solve_left(y, integral=True)  # indirect doctest
```
(throughout)

src/sage/matrix/matrix2.pyx
```diff
@@ -469,16 +469,16 @@ cdef class Matrix(Matrix1):
...
-            ``A.solve_right(B)``, that is, Sage implements "the
-            MATLAB/Octave backslash operator".
-
         INPUT:
 
         - ``B`` -- a matrix or vector
 
+        - ``integral`` -- boolean (default: ``False``); When the base
+          ring is a PID, request that a solution over the that ring
+          be returned. If there is no such solution, a ``ValueError``
```
mkoeppe:
```suggestion
        - ``integral`` -- boolean (default: ``False``); when the base
          ring is a PID, request that a solution over that ring
          be returned. If there is no such solution, a :class:`ValueError`
```

mkoeppe: Consider whether this could instead be expressed using an ``extend`` keyword, which is used in many other methods to control whether a larger ring/field than the base ring is allowed in results

yyyyx4:
Thanks, all done.

mkoeppe:
Of course, the same can be done with a weaker normal form, the Hermite form. matrix_integer_dense has many algorithm options for that one...

yyyyx4:
That's a good suggestion, thanks. I gave it a try.

mkoeppe (approved):
LGTM, thanks!

PR #38026, opened 2024-05-18
Do not create `spkg-*.log`, `spkg-*.time` files

### LINKED ISSUE #37887: Installing Sage's python3 spkg produces some untracked files
### Steps To Reproduce

Run `./configure --with-system-python3=no; make python3`. This is with Sage 10.3.beta4 on OS X.

### Expected Behavior

This should succeed, leaving a clean git state.

### Actual Behavior

`git status` returns
```
On branch develop
Your branch is up to date with 'upstream/develop'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	build/pkgs/python3/spkg-legacy-uninstall.log
	build/pkgs/python3/spkg-legacy-uninstall.time

nothing added to commit but untracked files present (use "git add" to track)
```

### Additional Information

Probably the log files should be in `logs/pkgs`, maybe called `python3-spkg-legacy-uninstall.*` in case there could be other `spkg-legacy-uninstall` scripts that might produce log files.

### Environment

```markdown
- **OS**: OS X 14.4.1
- **Sage Version**: 10.3.beta4
```

### Checklist

mkoeppe:
- Clean up after #37391, which used `sage-logger` for prefixing output; generating the log/time files was a side effect, which has not proven to be useful
- Fixes #37887 @jhpalmieri

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #37840 (merged here)

### DIFF
--- a/build/bin/sage-logger
+++ b/build/bin/sage-logger
@@ -1,13 +1,16 @@
 #!/usr/bin/env bash
 #
-# sage-logger [-p] COMMAND LOGFILE
+# sage-logger [-p] [-P PREFIX] COMMAND [LOGFILE]
 #
 # Evaluate shell command COMMAND while logging stdout and stderr to
-# LOGFILE. If either the command or the logging failed, return a
+# LOGFILE (if given). If either the command or the logging failed, return a
 # non-zero exit status.
 #
 # If the -p argument is given, each line printed to stdout is prefixed
 # with the name of the log file.
+#
+# If the -P PREFIX argument is given, each line printed to stdout is
+# prefixed with PREFIX.
 #
 # AUTHOR:
 #
@@ -24,11 +27,17 @@
 #*****************************************************************************
 
 use_prefix=false
+prefix=""
 
-if [[ "$1" = -p ]]; then
-    use_prefix=true
-    shift
-fi
+case "$1" in
+    -p) use_prefix=true
+        shift
+        ;;
+    -P) use_prefix=true
+        prefix="[$2] "
+        shift 2
+        ;;
+esac
 
 cmd="$1"
 logfile="$2"
@@ -36,10 +45,8 @@ export SAGE_LOGFILE="$logfile"
 logname="$(basename $logfile .log)"
 logdir=`dirname "$logfile"`
 
-if [[ $use_prefix = true ]]; then
+if [[ $use_prefix = true && -z "$prefix" ]]; then
     prefix="[${logname}] "
-else
-    prefix=""
 fi
 
 # Use sed option to reduce buffering, to make the output appear more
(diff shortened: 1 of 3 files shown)


jhpalmieri:
There is some code duplication. Would it make any sense to have a function that would replace each instance like
```
if [ -f spkg-pipinst ]; then
    sage-logger -P spkg-pipinst "$SAGE_SUDO ./spkg-pipinst"
    if [ $? -ne 0 ]; then
        error_msg "Error running the pipinst script for $PKG_NAME."
        exit 1
    fi
fi
```
by a call to a function defined by something like
```
run_script() {
if [ -f $1 ]; then
    sage-logger -P $1 "$SAGE_SUDO ./$1"
    if [ $? -ne 0 ]; then
        error_msg "Error running the $1 script for $PKG_NAME."
        exit 1
    fi
fi}
```
?

mkoeppe:
I've made it less verbose in a different way, please take a look

kwankyu:
It seems that some of `spkg-*.log` files are generated and some not. Which is which and why?

mkoeppe:
If any are still generated with this PR, that's a bug. Is this what you observed?

build/bin/sage-logger
```diff
@@ -61,15 +68,29 @@ else
...
+    pipestatus=(${PIPESTATUS[*]})
+
+    if [ ${pipestatus[1]} -ne 0 ]; then
+        exit ${pipestatus[1]}
+    else
+        exit ${pipestatus[0]}
+    fi
+fi
+
 timefile="$logdir/$logname.time"
```
kwankyu, line 87: Please a comment here saying `# if $logfille is given, then` or something, for readibility.

build/bin/sage-spkg
```diff
@@ -646,40 +618,24 @@ unset V
...
 
 # Run the pre-install script, if any
 if [ -f spkg-preinst ]; then
-    sage-logger -p "$SAGE_SUDO ./spkg-preinst" spkg-preinst.log
-    if [ $? -ne 0 ]; then
-        error_msg "Error running the preinst script for $PKG_NAME."
-        exit 1
-    fi
+    sage-logger -P spkg-preinst "$SAGE_SUDO ./spkg-preinst" spkg-preinst || exit_with_error_msg "Error running the preinst script for $PKG_NAME."
 fi
```
kwankyu, line 630: `spkg-preinst.log` is generated. No?

mkoeppe: Thanks for catching this; fixed in 7efeba390b3

build/bin/sage-logger
mkoeppe: Good idea, done in bfb292ae69a

kwankyu:
Can we test it here, with "CI Fix" tag?

I think there should be some explanation why you introduced the log files and why you think now they are not useful...

mkoeppe:
I don't have a better explanation than what I wrote in the ticket description: "I used sage-logger for prefixing output; generating the log/time files was a side effect, which has not proven to be useful"

kwankyu:
You certainly added the logging facility expecting that the logs would be useful. Then now you simply say "which has not proven to be useful". I am curious what happened in the meanwhile...

kwankyu:
I looked #37391 again. Right, you didn't add the logging feature. As you said, it is a side-effect of using `sage-logger` for prefixing. The generated log files are not useful, and hence here you add a feature to turn them off. 

OK. I misunderstood. Sorry.

mkoeppe:
No worries, these are all the right questions to ask!

kwankyu (approved):
Working well. LGTM.

mkoeppe:
Thank you!

PR #38027, opened 2024-05-19
Work on `FlatsMatroid`

gmou3:
Add class-optimized `_closure` and `_is_closed` methods.

Add input handling of a list of flats, and of a lattice of flats. Note that previously one could only define a `FlatsMatroid` from a dictionary of flats (indexed by their rank). The definition from a raw list requires more time upon definition as it computes and stores the lattice of flats from the input list. On the positive side, given this lattice, some methods become blazingly fast (e.g., `is_valid`, `whitney_numbers`).

The lattice of flats is now cached upon computation.

### :memo: Checklist

### DIFF
--- a/src/sage/matroids/flats_matroid.pxd
+++ b/src/sage/matroids/flats_matroid.pxd
@@ -1,4 +1,4 @@
-from sage.matroids.matroid cimport Matroid
+from .matroid cimport Matroid
 
 cdef class FlatsMatroid(Matroid):
     cdef frozenset _groundset  # _E
@@ -7,7 +7,8 @@ cdef class FlatsMatroid(Matroid):
     cpdef groundset(self)
     cpdef _rank(self, X)
     cpdef full_rank(self)
-    cpdef _is_independent(self, F)
+    cpdef _closure(self, X)
+    cpdef _is_closed(self, X)
 
     # enumeration
     cpdef flats(self, k)
@@ -18,4 +19,9 @@ cdef class FlatsMatroid(Matroid):
     cpdef relabel(self, mapping)
 
     # verification
+    cpdef is_valid(self)
+
+cdef class LatticeOfFlatsMatroid(FlatsMatroid):
+    cdef object _L  # lattice_of_flats
+    cpdef whitney_numbers(self)
     cpdef is_valid(self)
--- a/src/sage/matroids/flats_matroid.pxd
+++ b/src/sage/matroids/flats_matroid.pxd
@@ -1,9 +1,10 @@
 from .matroid cimport Matroid
 
 cdef class FlatsMatroid(Matroid):
-    cdef frozenset _groundset  # _E
-    cdef int _matroid_rank  # _R
+    cdef frozenset _groundset
+    cdef int _matroid_rank
     cdef dict _F  # flats
+    cdef object _L  # lattice of flats
     cpdef groundset(self)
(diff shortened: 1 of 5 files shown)


src/sage/matroids/flats_matroid.pyx
```diff
@@ -521,13 +562,243 @@ cdef class FlatsMatroid(Matroid):
...
                                 if flag:
                                     break
                         if not flag:
                             return False
 
         return True
+
+cdef class LatticeOfFlatsMatroid(FlatsMatroid):
+    r"""
+    INPUT:
```
tscrim:
```suggestion
cdef class LatticeOfFlatsMatroid(FlatsMatroid):
    r"""
    Matroid defined by its lattice of flats.
    
    INPUT:
```

src/sage/matroids/flats_matroid.pxd
```diff
@@ -7,7 +7,8 @@ cdef class FlatsMatroid(Matroid):
     cpdef groundset(self)
     cpdef _rank(self, X)
     cpdef full_rank(self)
-    cpdef _is_independent(self, F)
+    cpdef _closure(self, X)
+    cpdef _is_closed(self, X)
```
tscrim, line 12:
```suggestion
    cpdef bint _is_closed(self, X)
```
Perhaps it would be good to start adding such return type indicators in some key places? We probably have to see if the C code simplifies in some places.

tscrim (changes requested):
I am not sure about having a full subclass. I think `FlatsMatroid` should be able to take in the lattice as input and store it (and/or have an optional argument to compute it). If the lattice has been computed (which the user could ask for at some point), then the lattice should be stored and the matroid should chose the code path that take advantage of that. With the proposed change, the flats matroid cannot take advantage of knowing the lattice structure.

It's a bit of an overhaul, but I don't think it will be too difficult. What do you think?

Otherwise, I have some minor comments.

gmou3:
> It's a bit of an overhaul, but I don't think it will be too difficult. What do you think?

I like it. Quite easy and a good suggestion. Thanks!
Defining a subclass resulted in a lot of bloatcode (excuse my overly technical jargon).

src/sage/matroids/flats_matroid.pxd
gmou3: I am planning this on a massive scale on a future PR.

tscrim (changes requested):
Thank you.

The only other thing is I would explicitly set `self._L = None` and check `self._L is None`. This is because (in principle) the lattice might be empty, which is planned to eventually be changed so that `bool(empty_lattice)` is `False`. This might cause an issue with the corner case of the empty matroid. Plus it makes it a bit more explicit that we are checking if something is initialized.

gmou3:
> Thank you.
> 
> The only other thing is I would explicitly set `self._L = None` and check `self._L is None`. This is because (in principle) the lattice might be empty, which is planned to eventually be changed so that `bool(empty_lattice)` is `False`. This might cause an issue with the corner case of the empty matroid. Plus it makes it a bit more explicit that we are checking if something is initialized.

Done. It is more explicit indeed. I also saved some lines of code from the `__init__`.

Please have a look and don't hesitate to keep the corrections coming. :D

tscrim (approved):
I am happy with this as it stands. Positive review.

PR #38039, opened 2024-05-20
configure.ac: Check symlinks in the source tree

mkoeppe:
This is to alert users on WSL early about a source tree that has not been cloned correctly.

This will save users from running into build errors later, such as the one reported in https://groups.google.com/g/sage-devel/c/I6_yXdtMwQk/m/xQZoZFSGBgAJ

Fixes #37117

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

Issue #37117, opened 2024-01-20
failed to build source code 10.2


cyegger:
### Environment

```markdown
- **OS**: Ubuntu 22.04
- **Sage Version**: 10.2
```

### Steps To Reproduce

I met the error :

make[1]: Entering directory '/home/egger/sagemath/sage'
rm -rf config/install-sh config/compile config/config.guess config/config.sub config/missing configure build/make/Makefile-auto.in
rm -f src/doc/en/installation/*.txt
rm -rf src/doc/en/reference/spkg/*.rst
for a in environment environment-optional src/environment src/environment-dev src/environment-optional; do rm -f $a.yml $a-3.[89].yml $a-3.1[0-9].yml; done
rm -f src/Pipfile
rm -f src/pyproject.toml
rm -f src/requirements.txt
rm -f src/setup.cfg
make[1]: Leaving directory '/home/egger/sagemath/sage'
bootstrap:: installing 'm4/sage_spkg_configures.m4'
bootstrap:: installing 'm4/sage_spkg_versions.m4'
bootstrap:: installing 'm4/sage_spkg_versions_toml.m4'
./bootstrap: 1: src/bootstrap: not found
make: *** [Makefile:328: configure] Error 1

### Config log

make[1]: Entering directory '/home/egger/sagemath/sage'
rm -rf config/install-sh config/compile config/config.guess config/config.sub config/missing configure build/make/Makefile-auto.in
rm -f src/doc/en/installation/*.txt
rm -rf src/doc/en/reference/spkg/*.rst
for a in environment environment-optional src/environment src/environment-dev src/environment-optional; do rm -f $a.yml $a-3.[89].yml $a-3.1[0-9].yml; done
rm -f src/Pipfile
rm -f src/pyproject.toml
rm -f src/requirements.txt
rm -f src/setup.cfg
make[1]: Leaving directory '/home/egger/sagemath/sage'
bootstrap:: installing 'm4/sage_spkg_configures.m4'
bootstrap:: installing 'm4/sage_spkg_versions.m4'
bootstrap:: installing 'm4/sage_spkg_versions_toml.m4'
./bootstrap: 1: src/bootstrap: not found
make: *** [Makefile:328: configure] Error 1

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

fccoelho:
I had the same issue on 10.3

chrisxu3:
[...]

PR #38040, opened 2024-05-20
add important note in Step 2 about how WSL users should use the Linux git

chrisxu3:
### :memo: Checklist

### :hourglass: Dependencies

See #37117 #38039 for the details, which are regarding errors that you run into if you're using WSL and don't use Linux git to clone the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ in the Installation Guide.
 
 2.  Clone the sources with `git`:
 
+    - **NOTE: if you are using WSL, it is imperative that you follow these instructions, which use the Linux git. Otherwise you will run into confusing build errors.**
+
     - To check that `git` is available, open a terminal and enter
       the following command at the shell prompt (`$`):
 
--- a/README.md
+++ b/README.md
@@ -153,7 +153,9 @@ in the Installation Guide.
 
 2.  Clone the sources with `git`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,6 +153,8 @@ in the Installation Guide.
 
 2.  Clone the sources with `git`:
 
+    - **NOTE: if you are using WSL, it is imperative that you follow these instructions, which use the Linux git. Otherwise you will run into confusing build errors.**
+
```
mkoeppe, line 159: Maybe consolidate with the remark in lines 195-197 on Windows, symbolic links

mkoeppe:
@chrisxu3 I think our use of `-c core.symlinks=true` in the instructions is *intended* to actually make this work also with some types of `git` in Windows. But I'm not sure which ones. If you have a chance to try this out, this would be very helpful

README.md
```diff
@@ -153,6 +153,10 @@ in the Installation Guide.
 
 2.  Clone the sources with `git`:
 
+    - [Windows] **The Sage source tree contains symbolic links, which use UNIX line endings.
```
mkoeppe, line 156: "symbolic links" and "UNIX line endings" are two separate, unrelated things

chrisxu3: Maybe it should be clarified that it's using Windows line endings that ends up breaking the symlinks?

PR #38042, opened 2024-05-20
assign names to constructed Hadamard matrices

dimpase:
Hadamard and skew-Hadamard matrix generators are not telling you anything about the matrix you're getting.
This PR is an attempt to get this to work a bit better.

### :memo: Checklist

- [X ] The title is concise and informative.
- [ X] The description explains in detail what this PR is about.

### DIFF
--- a/src/sage/combinat/matrices/hadamard_matrix.py
+++ b/src/sage/combinat/matrices/hadamard_matrix.py
@@ -1663,6 +1663,9 @@ def hadamard_matrix(n, existence=False, check=True, construction_name=False):
     - ``check`` -- boolean (default: ``True``); whether to check that output is correct before
       returning it. As this is expected to be useless (but we are cautious
       guys), you may want to disable it whenever you want speed.
+    - ``construction_name`` -- boolean (default: ``False``); if it is ``True``,
+      ``existence`` is ``True``, and a matrix exists, output the construction name.
+      It has no effect if ``existence`` is set to ``False``.
 
     EXAMPLES::
 
@@ -1708,6 +1711,11 @@ def hadamard_matrix(n, existence=False, check=True, construction_name=False):
         [-----+-----+-----+-----+-----+-----]
         [ 1 -1| 1  1|-1 -1|-1 -1| 1  1| 1 -1]
         [ 1  1| 1 -1|-1  1|-1  1| 1 -1|-1 -1]
+
+    To find how the matrix is obtained, use ``construction_name`` ::
+
+        sage: matrix.hadamard(476, existence=True, construction_name=True)
+        'cooper-wallis small cases: 476'
 
     TESTS::
 
@@ -3099,6 +3107,9 @@ def skew_hadamard_matrix(n, existence=False, skew_normalize=True, check=True,
     - ``check`` -- boolean (default: ``True``); whether to check that output is
       correct before returning it. As this is expected to be useless (but we are
       cautious guys), you may want to disable it whenever you want speed
+    - ``construction_name`` -- boolean (default: ``False``); if it is ``True``,
+      ``existence`` is ``True``, and a matrix exists, output the construction name.
+      It has no effect if ``existence`` is set to ``False``.
 
     EXAMPLES::
 
@@ -3112,6 +3123,8 @@ def skew_hadamard_matrix(n, existence=False, skew_normalize=True, check=True,
         sage: skew_hadamard_matrix(2)
         [ 1  1]
         [-1  1]
+        sage: skew_hadamard_matrix(196, existence=True, construction_name=True)
+        'Williamson-Goethals-Seidel: 196'
 
     TESTS::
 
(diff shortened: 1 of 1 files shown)


dimpase:
I should still document new `construction_name=` parameters

src/sage/combinat/matrices/hadamard_matrix.py
```diff
@@ -1728,25 +1736,35 @@ def hadamard_matrix(n, existence=False, check=True):
...
-            return True
+            return report_name(name)
         M = matrix([[1, 1], [1, -1]])
     elif n == 1:
         if existence:
-            return True
+            return report_name(name)
         M = matrix([1])
     elif is_prime_power(n//2 - 1) and (n//2 - 1) % 4 == 1:
+        name = "paleyII " + name
```
dcoudert, line 1760: why setting this variable that is not reused ? you could put the string construction directly in the call of the function.

dcoudert (changes requested):
otherwise LGTM.

src/sage/combinat/matrices/hadamard_matrix.py
dimpase: I am not sure where you mean it to be re-used.

dimpase: There is a bit of a design deficiency in this code, which I am not going to address here - the matrices are just Sage matrices, you cannot attach names to them. A better way would be to have a derived class, I suppose. But not now. This is actually a quick fix to please a referee of a paper we wrote about these constructions.

dcoudert: I understand the motivation for this quick fix.

You do `name = "paleyI" + name` and then, if `existence` is `True`, you call `report_name(name)`. If `existence` is `False`, the modified name is not used in the method. So you could simply call `report_name("paleyI" + name)` and avoid `name = "paleyI" + name`.

src/sage/combinat/matrices/hadamard_matrix.py
```diff
@@ -1728,25 +1736,35 @@ def hadamard_matrix(n, existence=False, check=True):
...
         if existence:
-            return True
+            return report_name(name)
         M = hadamard_matrix_paleyII(n)
     elif n == 4 or n % 8 == 0 and hadamard_matrix(n//2, existence=True) is True:
+        name = "doubling" + name
```
MatteoCati: I think there should be a space at the end of the string here (`"doubling "`).

src/sage/combinat/matrices/hadamard_matrix.py
```diff
@@ -1755,44 +1773,53 @@ def hadamard_matrix(n, existence=False, check=True):
                        for i in range(R)])
         M = chad1.stack(chad2)
     elif is_prime_power(n - 1) and (n - 1) % 4 == 3:
+        name = "paleyI" + name
```
MatteoCati: As above, here I think it should be `"paleyI "`.

src/sage/combinat/matrices/hadamard_matrix.py
dimpase: I'd like to keep `name` as it is, as I plan to have more detailed name reporting, for recursive constructions, implemented,  as well. So `name` will get more "parents" info...

src/sage/combinat/matrices/hadamard_matrix.py
dimpase: done

src/sage/combinat/matrices/hadamard_matrix.py
dimpase: done

dimpase:
@MatteoCati - if you have time, until 28 May I'm revising the paper in dimpase/hadamard_sage (branch paper_specmats), have a look. 
[main.pdf](https://github.com/sagemath/sage/files/15409307/main.pdf)

src/sage/combinat/matrices/hadamard_matrix.py
dcoudert: ok.

dimpase:
@dcoudert - positive review?

dcoudert (approved):
LGTM.

MatteoCati (approved):

src/sage/combinat/matrices/hadamard_matrix.py
```diff
@@ -3136,52 +3169,62 @@ def skew_hadamard_matrix(n, existence=False, skew_normalize=True, check=True):
     if n < 1:
         raise ValueError("parameter n must be strictly positive")
 
-    def true():
-        _skew_had_cache[n] = True
+    def true(nam):
+        _skew_had_cache[n] = nam
+        if construction_name:
+            return nam+": "+str(n)
```
mkoeppe, line 3175:
```suggestion
            return f"{nam}: {n}"
```

dimpase: some other time then

PR #38045, opened 2024-05-21
simplify handling of dead oeis sequences

### LINKED ISSUE #38044: Ignore sequences with the keyword 'dead' when reading the OEIS database.
### Steps To Reproduce

_No response_

### Expected Behavior

Ignore sequences with the keyword 'dead' when reading the OEIS database.

### Actual Behavior

The opposite of the expected behavior.

### Additional Information

For ten years, we have asked to ignore sequences containing the keyword 'dead' when reading the OEIS database.

https://github.com/sagemath/sage/issues/17330

There are purely internal reasons why these sequences are in OEIS. Such sequences are not intended to be passed on. You cannot signal this more drastically than with the keyword 'dead'. And Sage should respect that.

What happened? Not what we intended; requests ('oeis(...,n)') still issue these false, misleading, and duplicated sequences. Every user still has to filter out the crap themselves.

Worse still, videlec had also noticed this (Mar 23, 2016) and wrote: "The patch provided does not do what is in the description. Namely, the dead sequences are not ignored!"

Unfortunately, in the same breath, he suggests adding an 'explicit error message' (which mo271 later toned down to a run-time warning).

The consequence is the opposite of what was intended: these sequences are not only not ignored but are particularly highlighted.

Even after adding a filter in some applications (like in 
https://github.com/PeterLuschny/BinaryQuadraticForms/blob/main/BinaryQF.ipynb
which we gladly recommend as a test case) these RuntimeWarnings continue to appear

    site-packages/sage/databases/oeis.py:1152:RuntimeWarning: This sequence is dead: "A141158: Duplicate of A038872." warn('This sequence is dead: "{}: {}"'.format(self.id(), self.name()), RuntimeWarning)

although this is now entirely misleading. I would have to continue patching even the warnings.

This absurd situation could be remedied with a single line: when reading the database put in:

    if 'dead' in seq.keywords() : continue

Please insert this line and change the warning to: "Move along, Nothing to See Here."

### Environment

```markdown
* OS: WSL: Ubuntu-22.04
* Sage Version 9.8
```

### Checklist

fchapoton:
Simplify the handling of dead OEIS sequences. Trying to ignore them, mostly.

This fixes #38044

### :memo: Checklist

### DIFF
--- a/src/sage/databases/oeis.py
+++ b/src/sage/databases/oeis.py
@@ -111,7 +111,7 @@ primes ?
     sage: b.formulas()[0]
     'E.g.f.: exp(exp(x) - 1).'
     sage: [i for i in b.comments() if 'prime' in i][-1]
-    'Number n is prime if ...'
+    'When n is prime, ...'
     sage: [n for n in range(2, 20) if (b(n)-2) % n == 0]
     [2, 3, 5, 7, 11, 13, 17, 19]
 
@@ -500,7 +500,8 @@ class OEIS:
                    'start': str(first_result)}
         url = oeis_url + "search?" + urlencode(options)
         sequence_list = _fetch(url).split('\n\n')[2:-1]
-        return FancyTuple([self.find_by_entry(entry=_) for _ in sequence_list])
+        T = [self.find_by_entry(entry=s) for s in sequence_list]
+        return FancyTuple([s for s in T if not s.is_dead()])
 
     def find_by_subsequence(self, subsequence, max_results=3, first_result=0):
         r"""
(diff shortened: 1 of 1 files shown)


kwankyu:
> Trying to ignore them, mostly.

I am not sure if this is desirable.

(1) When we search for oeis items for a given sequence, then dead items should be ignored, of course.

(2) But when we request data about a dead item, then it is right to warn the user.

To implement (1), `is_dead()` method should be an exception to (2) -- `is_dead()` should never raise a warning. But I think all other methods should behave according to (2).

Isn't this enough to solve #38044?

fchapoton:
I would prefer strongly to do exactly what I did. If somebody asks for data using the A-number of a dead sequence, it is his own responsability. No need for a warning.

kwankyu:
All right. I will leave the design to you and @PeterLuschny, as I have had no occasion to use OEIS.

I looked through the code, and have no other comments.

fchapoton:
so may I set to positive review in your name, if you cannot do it yourself ?

fchapoton:
ok, done.

PR #38053, opened 2024-05-22
work on letterplace algebras

fchapoton:
Moving the letterplace implementation of free algebras to the `Parent` framework

### :memo: Checklist


fchapoton:
There remains an annoying issue about `latex` methods. This needs to be investigated.

src/sage/algebras/letterplace/free_algebra_letterplace.pyx
```diff
@@ -591,20 +601,22 @@ cdef class FreeAlgebra_letterplace(Algebra):
...
-            tmp = [(j, E[base+j]) for j in range(ngens) if E[base+j]]
+            tmp = [(j, E[base + j]) for j in range(ngens) if E[base + j]]
             if not tmp:
                 continue
             var_ind, exp = tmp[0]
             if len(tmp) > 1 or exp > 1:
                 raise NotImplementedError("\n  Apparently you tried to view the letterplace algebra with\n  shift-multiplication as the free algebra over a finitely\n  generated free abelian monoid.\n  In principle, this is correct, but it is not implemented, yet.")
-
+            if self._nb_slackvars and var_ind == N:
+                continue
```
mkoeppe: what is this new code doing? is there an example that covers it?

fchapoton: There are two cases : with or without degrees argument. If there are degrees, there are "slack variables" and `_nb_slackvars` is set to 1. One slack variable appears at the end of each block in these loops. To properly get rid of them, we need this new code.

src/sage/algebras/letterplace/free_algebra_letterplace.pyx
```diff
@@ -624,15 +636,17 @@ cdef class FreeAlgebra_letterplace(Algebra):
...
             base = i*ngens
             tmp = [(j, E[base+j]) for j in range(ngens) if E[base+j]]
             if not tmp:
                 continue
             var_ind, exp = tmp[0]
             if len(tmp) > 1 or exp > 1:
                 raise NotImplementedError("\n  Apparently you tried to view the letterplace algebra with\n  shift-multiplication as the free algebra over a finitely\n  generated free abelian monoid.\n  In principle, this is correct, but it is not implemented, yet.")
-
+            if self._nb_slackvars and var_ind == N:
+                continue
```
mkoeppe: Same question here. Is this fixing a bug? Implementing a case currently not implemented?
Example / test case please

src/sage/algebras/letterplace/free_algebra_letterplace.pyx
```diff
@@ -233,9 +238,10 @@ cdef class FreeAlgebra_letterplace(Algebra):
 
         The free associative version of the given commutative ring.
 
-        NOTE:
+        .. NOTE::
 
-        One is supposed to use the `FreeAlgebra` constructor, in order to use the cache.
+            One is supposed to use the `FreeAlgebra` constructor,
```
tscrim:
```suggestion
            One is supposed to use the :class:`FreeAlgebra` constructor,
```

tscrim (commented):
I wonder a bit if we should have the single variable case redirect to a univariate polynomial ring, but I guess the other inconsistencies with coercion/type-checking/etc. would probably outweigh the extra functionality and speed in that case.

Is there a test that can be added to show that the change you made had an impact? I agree with the code, but to seems like there isn't any changes from before.

fchapoton:
What kind of change would you expect ? I am trying to make so that nothing changes, except that we get rid of the `Algebra` class inheritance. This means that all about gens is handled by category_object instead of the auld parent_with_gens.

tscrim:
I don’t see why changing the inheritance would change the print output (which is what the changes that Matthias is asking about is used for from what I understand). So I would expect something to be broken before this change that we could add as a test. If there isn’t anything visible, I  am curious as to why it isn’t necessary. Or are you saying its a performance boost?

fchapoton:
I cannot investigate now. What I did was
- change to Parent
- see that tests in latex repr were failing
- fix them as I could.
I have no idea what strange things were happening before.

tscrim:
Thank you. I can investigate what is going on. The change to `Parent` shouldn't have changed anything that would require adjustments like this...

tscrim:
Ah, I see what it is. By changing the type of the `for` loop in the, e.g., `exponents_to_string`, you've changed how it iterates. The `for i in range(n)` gets a new `i` every iteration but the `for i from 1 <= i < n` treats that like the condition on a `for` loop. The logic there seems a bit wonky and brittle. So now I understand why the added code is necessary.

fchapoton:
ah, great find. I had not seen that the loop variable was modified in the loop.. Eurk... What should we do ? Move back to the C-style for loop ?

tscrim:
I think we should just revert it to the C-style for now to keep this change more focused.

fchapoton:
ok, will do

tscrim (approved):
Thank you. Green bot => positive review.

fchapoton:
so I will allow myself to set to positive, thanks a lot

PR #38054, opened 2024-05-22
new iterators over the partitions of an integer

dcoudert:
As discussed in #37977, we implement new iterators over the partitions of an integer to get iterators generating partitions in increasing/decreasing lexicographic orders with partitions in ascending/descending orders.
We also implement methods yielding the next partition in a specific ordering.

### :memo: Checklist


dimpase:
linting errors?

src/sage/combinat/partitions.pyx
```diff
@@ -179,3 +335,400 @@ def ZS1_iterator_nk(int n, int k):
...
+        yield [1] * n
+        return
+    cdef list P = [0] * n
+    cdef int k = 1
+    cdef int y = n - 1
+    cdef int x, ell
+    while k:
+        k -= 1
+        x = P[k] + 1
+        while 2 * x <= y:
```
tscrim, line 683:
```suggestion
    cdef int x, twox, ell
    while k:
        k -= 1
        x = P[k] + 1
        twox = 2 * x
        while twox <= y:
```
Not sure if this will be faster or not, or if the compiler would optimizes this away anyways.

tscrim (changes requested):
At least it seems like we should not use `ZS2` over `AccelAsc` for increasing lex iteration by default. I am not opposed to keeping both implementations with some test for a sufficiently large `n` (say, between 50 and 100) showing the outputs are equal at every step.

I think we should add a `prev` method to partitions as well.

Other than that, I am good with this.

src/sage/combinat/partitions.pyx
dcoudert: I'm not sure we need this change. My timing don't show any improvement:

Before
```py
sage: %timeit sum(1 for _ in AccelAsc_iterator(20))
140 µs ± 481 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit sum(1 for _ in AccelAsc_iterator(50))
46.1 ms ± 117 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
After
```py
sage: %timeit sum(1 for _ in AccelAsc_iterator(20))
141 µs ± 810 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
sage: %timeit sum(1 for _ in AccelAsc_iterator(50))
46.1 ms ± 568 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

dcoudert:
> At least it seems like we should not use `ZS2` over `AccelAsc` for increasing lex iteration by default. I am not opposed to keeping both implementations with some test for a sufficiently large `n` (say, between 50 and 100) showing the outputs are equal at every step.

The orderings are not the same. Both generate partitions in lexicographic order, but the partitions of `ZS2` are in descending order while the partitions of `AccelAsc` are in ascending order. So both might be of interest for some users.
 
> I think we should add a `prev` method to partitions as well.

done.

src/sage/combinat/partition.py
```diff
@@ -7110,12 +7110,29 @@ def next(self, p):
...
+        q = ZS1_next(list(p))
+        if q:
+            return self.element_class(self, q)
+        return None
+
+    def prev(self, p):
+        r"""
+        Return the lexicographically previous partition before partition `p`.
+
+        EXAMPLES:
```
```suggestion
        Return the lexicographically previous partition before partition ``p``.

        EXAMPLES::
```

src/sage/combinat/partitions.pyx
```diff
@@ -45,52 +124,129 @@ def ZS1_iterator(int n):
...
+        h -= 1
+    # Let Q be such that Q[:m+1] == P and Q[h+1:] is an array of n-(h+1) ones
+    cdef list Q = P + [1] * (n - m - 1)
+    ZS1_step(Q, n, &m, &h)
+    return Q[:m+1]
+
+
+def ZS1_iterator_nk(int n, int k):
+    r"""
+    An iterator for the partitions of `n` of length at most `k` (in the
```
tscrim:
```suggestion
    An iterator for the partitions of ``n`` of length at most ``k`` (in the
```
and similar changes since the are inputs (and hence "code").

src/sage/combinat/partitions.pyx
```diff
@@ -179,3 +335,410 @@ def ZS1_iterator_nk(int n, int k):
...
+    - ``n`` -- integer; the sum of the elements of the partition stored in `P`
+
+    - ``m`` -- pointer to the index of the last element of the partition
+
+    - ``h`` -- pointer to the last value larger than 1 in `P`
+
+    .. WARNING::
+
+        The method assumes that input parameters are valid and consistent with
+        the AccelDesc algorithm as proposed in [YYYYYYY]_. It modifies the content of `P`
```
tscrim:
```suggestion
        the AccelDesc algorithm as proposed in [KS2012]_. It modifies the content of ``P``
```

src/sage/combinat/partitions.pyx
```diff
@@ -179,3 +335,410 @@ def ZS1_iterator_nk(int n, int k):
...
+
+    Check that :meth:`ZS1_iterator` and :meth:`AccelDesc_iterator` generate
+    partitions in the same order::
+
+        sage: from sage.combinat.partitions import ZS1_iterator
+        sage: from sage.misc.prandom import randint
+        sage: n = randint(1, 50)
+        sage: for p, q in zip(ZS1_iterator(n), AccelDesc_iterator(n)):  # long time
+        ....:     if p != q:
+        ....:         raise ValueError('something goes wrong')
```
tscrim:
```suggestion
        sage: all(p == q for p, q in zip(ZS1_iterator(n), AccelDesc_iterator(n)))  # long time
        True
```

src/sage/combinat/partitions.pyx
tscrim: Either it is such a mirco-optimization as to never matter or the compiler generally optimizes it away. Then let's leave it as is.

src/sage/combinat/partitions.pyx
```diff
@@ -1,16 +1,88 @@
...
+    sage: from sage.combinat.partitions import AccelDesc_next
+    sage: AccelDesc_next([2, 2])
+    [2, 1, 1]
+    sage: from sage.combinat.partitions import ZS2_next
+    sage: ZS2_next([2, 2])
+    [3, 1]
+    sage: from sage.combinat.partitions import AccelAsc_next
+    sage: AccelAsc_next([2, 2])
+    [4]
+
```
tscrim, line 57:
```suggestion
```
One too many blanklines.

tscrim (changes requested):
Thank you.

> > At least it seems like we should not use `ZS2` over `AccelAsc` for increasing lex iteration by default. I am not opposed to keeping both implementations with some test for a sufficiently large `n` (say, between 50 and 100) showing the outputs are equal at every step.
> 
> The orderings are not the same. Both generate partitions in lexicographic order, but the partitions of `ZS2` are in descending order while the partitions of `AccelAsc` are in ascending order. So both might be of interest for some users.

To me, a partition is always a weakly decreasing sequence (or a multiset), but yes, I agree that having the reverse order within the lists could be useful.

Just a few more minor things.

src/sage/combinat/partitions.pyx
dcoudert: I think I changed all occurrences.

src/sage/combinat/partitions.pyx
dcoudert: oups, good catch ;)

src/sage/combinat/partitions.pyx
dcoudert: done

dcoudert:
Thanks for the comments.

tscrim:
The bot is showing some failures:
```
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/groups/misc_gps/argument_groups.py  # 6 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/interfaces/gp.py  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/libs/pari/tests.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/lazy_series_ring.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/lazy_series.py  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/padics/padic_generic_element.pyx  # 2 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/qqbar.py  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/stats/basic_stats.py  # 2 doctests failed
```
I am not sure how many of them are real failures due to the changes here. Some seem unrelated from the log.

tscrim (approved):
Thanks for checking. I also saw today that that group of errors was reported on sage-release. So positive review.

dcoudert:
Thank you for the review.

PR #38056, opened 2024-05-22
`SetSystem`: Minor change to accommodate set input

gmou3:
This change makes the code cleaner in multiple places.
The PR also includes some docstring edits in `set_system.pyx`.

### :hourglass: Dependencies

- Depends on #37930 and #38027

PR #38057, opened 2024-05-22
`matroids`: Declare function input and return types

gmou3:
Especially for internal functions (starting with `_`).

The PR includes many changes, most of which are algorithmically inconsequential (simple type declarations and docstring edits).

Careful review is mostly needed for commit `f0fd595`, where some extra type castings are done within basic `matroid` functions.

### :hourglass: Dependencies

- Depends on #38056.


mkoeppe:
there are test failures

src/sage/matroids/matroid.pyx
```diff
@@ -63,9 +63,9 @@ additional functionality (e.g. linear extensions).
     - :meth:`circuit_closures() <sage.matroids.matroid.Matroid.circuit_closures>`
     - :meth:`nonspanning_circuit_closures() <sage.matroids.matroid.Matroid.nonspanning_circuit_closures>`
     - :meth:`bases() <sage.matroids.matroid.Matroid.bases>`
-    - :meth:`independent_r_sets() <sage.matroids.matroid.Matroid.independent_r_sets>`
+    - :meth:`independent_k_sets() <sage.matroids.matroid.Matroid.independent_k_sets>`
```
mkoeppe: This rename cannot be done without deprecation. See deprecated_function_alias

src/sage/matroids/advanced.py
```diff
@@ -63,6 +64,4 @@
 from . import lean_matrix
 from .extension import LinearSubclasses, MatroidExtensions
 from .union_matroid import MatroidUnion, MatroidSum, PartitionMatroid
-
-from sage.misc.lazy_import import lazy_import
-lazy_import('sage.matroids.graphic_matroid', 'GraphicMatroid')
+from .graphic_matroid import GraphicMatroid
```
mkoeppe: I made it a `lazy_import` for modularization purposes: `sage.matroids` is intended to be shipped as part of the distribution **sagemath-modules**; graph functionality becomes available at runtime if the distribution **sagemath-graphs** is installed

src/sage/matroids/basis_matroid.pyx
```diff
@@ -72,30 +64,29 @@ Methods
...
 from cpython.object cimport Py_EQ, Py_NE
-from sage.misc.decorators import rename_keyword
 from itertools import combinations
+from sage.data_structures.bitset_base cimport *
+from sage.misc.decorators import rename_keyword
+from sage.structure.richcmp cimport rich_to_bool
+from .matroid cimport Matroid
+from .basis_exchange_matroid cimport BasisExchangeMatroid
+from .set_system cimport SetSystem
+from .utilities import cmp_elements_key
```
mkoeppe: Note we got rid of relative cimports for because of technical restrictions of Cython (see #36228). Best not to undo it until the problem is fixed in Cython

src/sage/matroids/advanced.py
gmou3: I changed this when I cythonized `GraphicMatroid`. It doesn't compile.

gmou3: It works just fine actually. Can I run tests as if `graphs` isn't available?

src/sage/matroids/basis_exchange_matroid.pyx
```diff
@@ -1292,19 +1255,17 @@ cdef class BasisExchangeMatroid(Matroid):
                 self._whitney_numbers2_rec(f_vec, flats, todo, e + 1, i + 1)
             e = bitset_next(todo[i], e)
 
-    cpdef flats(self, r):
+    cpdef SetSystem flats(self, long k):
```
mkoeppe, line 1258: Technically this is a breaking change because user code may be passing `r` as a keyword argument

src/sage/matroids/matroid.pyx
```diff
@@ -2342,31 +2343,32 @@ cdef class Matroid(SageObject):
...
+                           self._rank(X.intersection(Y)) > rX + rY):
                             return False
         return True
 
     # enumeration
 
-    cpdef circuits(self):
+    cpdef SetSystem circuits(self, k=None):
         """
         Return the circuits of the matroid.
```
mkoeppe, line 2370: Document INPUT please

src/sage/matroids/basis_exchange_matroid.pyx
gmou3: What to do for this? `@rename_keyword` does not work for a `cpdef` function I think.

mkoeppe: It would be OK with me to just argue that given the very short name of this parameter, it was intended to be a positional-only argument but that Cython has historically not supported declaring positional-only arguments (do you know if it supports them now?), and therefore renaming without deprecation is fine.

But I'd welcome a second opinion on this; @tscrim?

gmou3: This argument seems convincing to me too. I don't know about the support of positional-only arguments.

tscrim: I highly doubt that anyone is using it as a keyword argument, but we should properly deprecate it if changing the name. Of course, we don't strictly have to do this, but it is the general policy. My preference would be to just keep the current argument name, but I think it is okay to go ahead and do it (without a deprecation) if you feel the renaming is warranted.

gmou3: I would like to rename it please. :D

In fact I would like to eventually incorporate the functionality of `independent_k_sets()` into `independent_sets()` with an optional argument `k`. Also I want to make `flats()` with no argument return all flats.

Let me know if I should do these here.

tscrim: Some justification would be good all the same. Right now it seems the name change is quite arbitrary (with `r` perhaps being sort for `rank`).

One reason for having separate methods is that the logic is more simple (and so marginally faster, which could matter in a tight loop). However, I would side with you on having just one method makes the API easier and that extra check/argument parsing likely is inconsequential.

tscrim: Of course, you can do the second change here too.

gmou3: `k` seems good to me for cardinalities (`k`-sets). `r` can be confused with the matroid's rank. Also, `k` is often not in use such as other indices (e.g., `i`). This argument holds well for independent sets, for flats the choice of `k` is a bit more arbitrary, but still fine by me.

Your concern about potential inefficiency is legitimate, however in this case it is indeed inconsequential as there is a lot of computation within the `if` statement.

gmou3: I'll leave the change of the functionality of flats for another PR.

src/sage/matroids/circuits_matroid.pyx
```diff
@@ -492,103 +467,105 @@ cdef class CircuitsMatroid(Matroid):
...
-    cpdef independent_r_sets(self, long r):
+    cpdef SetSystem independent_sets(self, long k=-1):
         r"""
-        Return the size-``r`` independent subsets of the matroid.
+        Return the size-`k` independent sets of the matroid.
 
         INPUT:
 
-        - ``r`` -- nonnegative integer
+        - ``k`` -- integer
```
mkoeppe: Shouldn't this explain what happens when `k` is not provided

mkoeppe (commented):
Otherwise LGTM now. 
Dropping the `r` (or `k`) in `independent_sets` is a nice solution.

src/sage/matroids/circuits_matroid.pyx
gmou3: Yes, thanks.

src/sage/matroids/basis_exchange_matroid.pyx
tscrim: I just want to make one final comment (mostly a repeat). I feel that is generally this is not a sufficient reason to make a backwards incompatible change. So I recommend doing such changes sparingly.

PR #38062, opened 2024-05-23
Implementation of algorithm for exact counting graph homomorphisms

guojing0:
In the past few months, we have been working towards a SageMath library for exact counting graph homomorphisms, based on the proof of Prop. 1.6 from "[Homomorphisms Are a Good Basis for Counting Small Subgraphs](https://arxiv.org/abs/1705.01595)" by Radu Curticapean, Holger Dell, and Dániel Marx: The algorithm is *deterministic* and runs in time $\exp(O(k)) + poly(k) \cdot n^{tw(H) + 1}$ for counting the number of homomorphisms from graph $H$ to graph $G$, where $k = |V(H)|$, $n = |V(G)|$, and $tw(H)$ is the treewidth of $H$.

The library is recently released on GitHub: https://github.com/guojing0/count-graph-homs

This commit is mainly for the sake of CI. More documentations and tests will follow.

### :memo: Checklist


src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+        self.density_threshold = density_threshold
+        self.graph_clr = graph_clr
+        self.target_clr = target_clr
+        self.colourful = colourful
+
+        # Bookkeeping for colourful mappings
+        self.actual_target_graph = target_graph
+        self.actual_target_size = len(self.actual_target_graph)
+
+        if not isinstance(graph, Graph):
```
dcoudert: Such test should be done before any assignments

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+        self.graph = graph
+        self.target_graph = target_graph
+        self.density_threshold = density_threshold
+        self.graph_clr = graph_clr
+        self.target_clr = target_clr
+        self.colourful = colourful
+
+        # Bookkeeping for colourful mappings
+        self.actual_target_graph = target_graph
+        self.actual_target_size = len(self.actual_target_graph)
```
dcoudert: could be `self.actual_target_size = target_graph.order()` ?

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
+from sage.graphs.graph import Graph
+
+from sage.graphs.homomorphisms.helper_functions import *
+
+from sage.graphs.graph_decompositions.tree_decomposition import make_nice_tree_decomposition, label_nice_tree_decomposition
```
dcoudert: these imports are used only once in `__init__` and so can be done just before using them.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+            sage: from sage.graphs.homomorphisms.count_homomorphisms import GraphHomomorphismCounter
+            sage: square = graphs.CycleGraph(4)
+            sage: bip = graphs.CompleteBipartiteGraph(2, 4)
+            sage: counter = GraphHomomorphismCounter(square, bip)
+            sage: counter.count_homomorphisms()
+            128
+        """
+        # Whether it's BFS or DFS, every node below join node(s) would be
+        # computed already, so we can go bottom-up safely.
+        for node in reversed(self.dir_labelled_TD.vertices()):
```
dcoudert: Can you clarify the order in which you want to visit the vertices. Here, you take the reverse order of the list returned by `vertices()` but `vertices()` does not ensure any kind of ordering.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+        return self.DP_table[0][0]
+
+    ### Main adding functions
+
+    def _add_leaf_node(self, node):
+        r"""
+        Add the computation result of processing leaf node(s) to the DP table.
+        """
+        node_index = get_node_index(node)
+        self.DP_table[node_index] = [1]
```
dcoudert, line 124: lies 69-80, you explained that `DP_table` is a list of dictionary and here you assign a list instead of a dictionary. Please clarify the data structure you use and why such assignment might be valid.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def _add_intro_node(self, node):
+        r"""
+        Add the computation result of processing introduce node(s) to the DP table.
+        """
+        # Basic setup
+        node_index, node_vertices = node
+        node_vtx_tuple = tuple(node_vertices)
+
+        child_node_index, child_node_vtx = self.dir_labelled_TD.neighbors_out(node)[0]
```
dcoudert: You could use `next(self.dir_labelled_TD.neighbor_out_iterator(node))`

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+        child_node_vtx_tuple = tuple(child_node_vtx)
+
+        mappings_length = self.actual_target_size ** len(node_vtx_tuple)
+
+        mappings_count = [0] * mappings_length
+
+        # Use the adjacency matrix when dense, otherwise the graph itself
+        target_density = self.actual_target_graph.density()
+        target = self.actual_target_graph.adjacency_matrix() if target_density >= self.density_threshold else self.actual_target_graph
```
dcoudert: when using `adjacency_matrix`, it is safer to specify the order of the vertices using parameter `vertices` (see the documentation) to ensure that the i-th row/column of the matrix corresponds to the I-th vertex in your ordering.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+
+        if colourful and (graph_clr is None or target_clr is None):
+            raise ValueError("Both graph_clr and target_clr must be provided when colourful is True")
+
+        self.graph._scream_if_not_simple()
+        self.target_graph._scream_if_not_simple()
+
+        self.tree_decomp = graph.treewidth(certificate=True)
+        self.nice_tree_decomp = make_nice_tree_decomposition(graph, self.tree_decomp)
+        self.root = sorted(self.nice_tree_decomp)[0]
```
dcoudert: You can use `self.root = min(self.nice_tree_decomp)`.

src/sage/graphs/homomorphisms/helper_functions.py
```diff
@@ -0,0 +1,95 @@
...
+    else:
+        # Assume that `target_graph` is an adjacency matrix
+        return all(target_graph[mapped_vtx, vtx] for vtx in mapped_nbhrs)
+
+
+### For integer representation of mappings
+
+def extract_bag_vertex(mapping, index, graph_size):
+    r"""
+    Extract the bag vertex at `index` from `mapping`
```
dcoudert: The method don't `extract` but returns an index

src/sage/graphs/homomorphisms/helper_functions.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    # Equivalent to taking the floor
+    return mapping // (graph_size ** index) % graph_size
+
+def add_vertex_into_mapping(new_vertex, mapping, index, graph_size):
+    r"""
+    Insert `new_vertex` at `index` into `mapping`
```
dcoudert: The method don't `insert` but returns the index where to do so.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+        r"""
+        INPUT:
+
+        - ``graph`` -- a Sage graph
+
+        - ``target_graph`` -- the graph to which ``graph`` is sent
+
+        - ``density_threshold`` (default: 0.25) -- the desnity threshold for `target_graph` representation
+
+        - ``graph_clr`` (default: None) -- a list of integers representing the colours of the vertices of `graph`
```
dcoudert: The proper form is
```
        - ``density_threshold`` -- float (default: ``0.25``); the density threshold for ``target_graph`` representation
```

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,236 @@
...
+
+        - ``target_clr`` (default: None) -- a list of integers representing the colours of the vertices of `target_graph`
```
dcoudert: The proper form is 
```
        - ``target_clr`` -- list (default: ``None``); a list of integers representing the colors of the vertices of ``target_graph``
```

dcoudert (changes requested):
This PR requires some work before. It must be properly documented, doctests are needed, etc. Here is a first list of needed changes.

I don't think we need this `helper_functions.py` file. I fact, I'm in favor of using a single file called `homomorphisms.py` and gathering all methods related to graph homomorphisms.

src/sage/graphs/homomorphisms/count_homomorphisms.py
guojing0: I just read the implementation code for `vertices()`, since `self.dir_labelled_TD` is a DAG, I assume in this case, the vertices returned would/should be from the source to the sink(s)?

src/sage/graphs/homomorphisms/count_homomorphisms.py
guojing0: Sorry, old comment. Just corrected.

src/sage/graphs/homomorphisms/count_homomorphisms.py
dcoudert: No, we cannot make any assumption on this ordering. It depends on internal data structures and not on the orientation of the graph. If you need a specific ordering, you must use an appropriate graph traversal method (BFS, DFS, etc.).

src/sage/graphs/homomorphisms/count_homomorphisms.py
guojing0: Because this is a one-time thing; we do not necessarily need to use an iterator? Which one would be more efficient and/or faster?

dcoudert: With the current code, `self.dir_labelled_TD.neighbors_out(node)[0]`, you first build the list of neighbors and then get the first one.  With `next(self.dir_labelled_TD.neighbor_out_iterator(node))`, you just get the first neighbor so it's faster.

src/sage/graphs/homomorphisms/count_homomorphisms.py
guojing0: Do you mean something like in the current code?

guojing0:
> This PR requires some work before. It must be properly documented, doctests are needed, etc. Here is a first list of needed changes.

Thank you for all the reviews! Of course, many tests and much documentation are needed.

> I don't think we need this `helper_functions.py` file. I fact, I'm in favor of using a single file called `homomorphisms.py` and gathering all methods related to graph homomorphisms.

I also consider the possibility of having only one file, but wouldn't it make the main file `count_homomorphisms.py` too long? It's already 230+ lines.

I think the name `count_homomorphisms.py` would better reflect what the goal(s) this file intends to achieve, i.e., counting homomorphisms; `homomorphisms.py` sounds somewhat general.

src/sage/graphs/homomorphisms/count_homomorphisms.py
dcoudert: Please check the documentation of `adjacency_matrix`. You can specify the order of the vertices `adjacency_matrix(vertices=my_order)`,  where `my_order` is a list of vertices indicating some order.

dcoudert:
> I also consider the possibility of having only one file, but wouldn't it make the main file `count_homomorphisms.py` too long? It's already 230+ lines.

We already have files with 20.000+ lines.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -151,7 +145,13 @@ def _add_intro_node(self, node):
 
         # Use the adjacency matrix when dense, otherwise the graph itself
         target_density = self.actual_target_graph.density()
-        target = self.actual_target_graph.adjacency_matrix() if target_density >= self.density_threshold else self.actual_target_graph
+
+        if target_density >= self.density_threshold:
+            target = self.actual_target_graph.adjacency_matrix(
+                vertices=self.actual_target_graph.vertices()
```
dcoudert, line 146: I think it's best if you fix the order in which you want to consider the vertices, assign it to a list and use it. Otherwise you are not sure what may happen.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,327 @@
...
+        self.density_threshold = density_threshold
+        self.graph_clr = graph_clr
+        self.target_clr = target_clr
+        self.colourful = colourful
+
+        # Bookkeeping for colourful mappings
+        self.actual_target_graph = target_graph
+        self.actual_target_size = self.actual_target_graph.order()
+
+        self.graph._scream_if_not_simple()
```
dcoudert: these tests should also be done prior to any assignment.

src/sage/graphs/homomorphisms/count_homomorphisms.py
```diff
@@ -0,0 +1,327 @@
...
+            sage: from sage.graphs.homomorphisms.count_homomorphisms import GraphHomomorphismCounter
+            sage: square = graphs.CycleGraph(4)
+            sage: bip = graphs.CompleteBipartiteGraph(2, 4)
+            sage: counter = GraphHomomorphismCounter(square, bip)
+            sage: counter.count_homomorphisms()
+            128
+        """
+        # Whether it's BFS or DFS, every node below join node(s) would be
+        # computed already, so we can go bottom-up safely.
+        for node in reversed(list(self.dir_labelled_TD.breadth_first_search(min(self.dir_labelled_TD)))):
```
dcoudert, line 101: Are you sure that `min(self.dir_labelled_TD)` is the root ? isn't it safer to search for a vertex with in-degree 0 ?

guojing0: > Are you sure that min(self.dir_labelled_TD) is the root ?

By construction, yes.

src/sage/graphs/homomorphisms/helper_functions.py
guojing0: Yes, but I could not think of a better name for this. Do you have any name suggestions?

src/sage/graphs/homomorphisms/helper_functions.py
guojing0: I belive this returns that vertex at some `index`

guojing0:
@dcoudert Thank you again for the code reviews! I think I have changed some codes based on your review.

Do you think we also need examples/tests for the small helper functions at the end of the file?

dcoudert:
Tests are needed in all methods, even the small helper functions.

PR #38067, opened 2024-05-24
`.ci/write-dockerfile.sh`: Quoting fix for `unminimize`

mkoeppe:
On docker-ubuntu, we attempted to remove the `unminimize` program after running it once, but this has failed because of  quoting mistake. https://github.com/mkoeppe/sage/actions/runs/9219405769/job/25364502464#step:9:311

Fixed here.

### :memo: Checklist

### DIFF
--- a/.ci/write-dockerfile.sh
+++ b/.ci/write-dockerfile.sh
@@ -67,7 +67,7 @@ case $SYSTEM in
                 cat <<EOF
 RUN if command -v unminimize > /dev/null; then  \
         (yes | unminimize) || echo "(ignored)"; \
-        rm -f "$(command -v unminimize)";       \
+        rm -f "\$(command -v unminimize)";       \
     fi
 EOF
                 if [ -n "$DIST_UPGRADE" ]; then


.ci/write-dockerfile.sh
```diff
@@ -67,7 +67,7 @@ case $SYSTEM in
                 cat <<EOF
 RUN if command -v unminimize > /dev/null; then  \
         (yes | unminimize) || echo "(ignored)"; \
-        rm -f "$(command -v unminimize)";       \
+        rm -f "\$(command -v unminimize)";       \
     fi
```
kwankyu, line 71: Are you sure? The old line (without `\`) seems correct.

kwankyu:
```bash
$ command -v unminimize
/usr/local/sbin/unminimize
$ rm -f "$(command -v unminimize)";  
rm: cannot remove '/usr/local/sbin/unminimize': Permission denied
$ rm -f "\$(command -v unminimize)";
$ command -v unminimize
/usr/local/sbin/unminimize
```

mkoeppe: This appears as part of a "here document" (`cat <<EOF`...`EOF`, https://www.gnu.org/software/bash/manual/bash.html#Here-Documents). The bug is that the `$(...)` is expanded at the time of running the script (in the host); it needs to be expanded when Docker executes the RUN command in the container.

kwankyu: Ah, okay. I failed to see the context. Thanks.

kwankyu (approved):
LGTM.

mkoeppe:
Thanks.

PR #38068, opened 2024-05-24
Correct developer guide's function template

gmou3:
Add spaces in tuples.

P.S. Should the default values `1` and `2` have single or double backticks?


src/doc/en/developer/coding_basics.rst
```diff
@@ -649,44 +649,44 @@ indentation:
...
           argument ``x`` goes here. If it contains multiple lines, all
           the lines after the first need to begin at the same indentation
           as the backtick.
 
-        - ``y`` -- integer (default: `2`); the description of the
+        - ``y`` -- integer (default: ``2``); the description of the
           argument ``y``
 
-        OUTPUT: the point as a tuple
+        OUTPUT: tuple; further description of the output
```
saraedum, line 664:
```suggestion
        OUTPUT: a tuple ``(x^5, y)``; further description of the output. This section
        can sometimes be omitted completely when it is immediately evident from
        the initial description.
```

saraedum (commented):
Sorry, I misunderstood where this goes in the entire documentation. I think the original `OUTPUT` was actually fine.

Then again, we have many examples of `OUTPUT: tuple` in the code base. So it's a good template to follow that.

gmou3:
Thank you.

src/doc/en/developer/coding_basics.rst
```diff
@@ -649,44 +649,44 @@ indentation:
...
     def point(self, x=1, y=2):
         r"""
-        Return the point `(x^5,y)`.
+        Return the point `(x^5, y)`.
 
         INPUT:
 
-        - ``x`` -- integer (default: `1`); the description of the
+        - ``x`` -- integer (default: ``1``); the description of the
           argument ``x`` goes here. If it contains multiple lines, all
```
kwankyu, line 657:
```
`1`
```

src/doc/en/developer/coding_basics.rst
```diff
@@ -649,44 +649,44 @@ indentation:
...
           the lines after the first need to begin at the same indentation
           as the backtick.
 
-        - ``y`` -- integer (default: `2`); the description of the
+        - ``y`` -- integer (default: ``2``); the description of the
           argument ``y``
```
kwankyu, line 662:
```
`2`
```

src/doc/en/developer/coding_basics.rst
```diff
@@ -649,44 +649,44 @@ indentation:
...
             ...
-            TypeError: unable to convert 'r' to an integer
+            TypeError: unable to convert 'x' to an integer
 
         .. NOTE::
 
-            This function uses the algorithm of [BCDT2001]_ to determine
-            whether an elliptic curve `E` over `Q` is modular.
+            This function uses :func:`pow` to determine the fifth
+            power of ``x``.
```
kwankyu, line 690:
```
`x`
```

src/doc/en/developer/coding_basics.rst
```diff
@@ -649,44 +649,44 @@ indentation:
...
         EXAMPLES:
 
         This example illustrates ... ::
 
-            sage: A = ModuliSpace()
-            sage: A.point(2,3)
-            xxx
+            sage: A = EuclideanSpace(2)
+            sage: A.point(2, 3)
+            (32, 3)
```
kwankyu, line 673: Why not 
```
(2, 3)
```
in the output?

src/doc/en/developer/coding_basics.rst
kwankyu: Reason: math for mathematical value.

gmou3:
@kwankyu I agree with these recommendations. It is clearer if one thinks of something more complicated as the default value, a field for example. We would like to show some nice math instead of some code I presume (the code would appear further up in the arguments).

I think I need to do a follow-up PR.

kwankyu:
If you agree, I may do it in #38117.

saraedum:
I'm confused. Should the defaults be typeset with single backticks and not double-backticks anymore?

gmou3:
Something from Python like None, or True, should have double backticks.
Something mathematical should have single backticks.

Let us know of any counter-argument.

saraedum:
> Something from Python like None, or True, should have double backticks. Something mathematical should have single backticks.
> 
> Let us know of any counter-argument.

My main argument is that the code base uses more double-backticks than single-backticks. I believe that this document should not set new standards but report what the code base is already doing.

Also, `1` is the actual default value as a Python `int`, it's not really a "mathematical object" arguably. I believe that the defaults should report the actual default of a function, imho this can never be a mathematical object. If it's ``None`` and ``None`` means the ring of integers, then `INPUT:` should say so instead of claiming that ``ring (default: `\ZZ`)`` simply because `` `\ZZ` `` is not something you can type into SageMath. I have no strong opinion on that latter part though.

gmou3:
I think the part after the "--" should be treated the same as the text body preceding the "INPUT". One can see how to type the default value from the function header.

I am unsure about what the rule is/was. The template had (and still has) single backticks however. We can determine here what is desirable based on some logic, and everything else can be adapted to comply.

PR #37873, opened 2024-04-25
`sage.libs.pynac`: Remove, deprecated in #32386 (2021)

mkoeppe:
### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38081, opened 2024-05-25
Fix sage.symbolic feature after libs/pynac removal

antonio-rojas:
Fixes https://github.com/sagemath/sage/pull/37873#issuecomment-2131371458


mkoeppe:
Thanks for fixing this!

PR #38106, opened 2024-05-28
Fix MPoly functors for single variable

### LINKED ISSUE #38089: Multiplying a multivariate polynomial in one variable by a constant unexpectedly changes its type to univariate
### Steps To Reproduce

The code:

```
K = PolynomialRing(ZZ,1,'z')
z = K.gen(0)
print( z.parent() )
print( (z*QQ.one()).parent() )
```

prints

> Multivariate Polynomial Ring in z over Integer Ring
> Univariate Polynomial Ring in z over Rational Field

### Expected Behavior

Multiplication by a constant should not turn a multivariate polynomial into univariate one. That is, `z*QQ.one()` should be an element of `Multivariate Polynomial Ring in z over Rational Field`.

### Actual Behavior

`z*QQ.one()` unexpectedly becomes a univariate polynomial.

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Ubuntu 22.04.4 LTS
- **Sage Version**: 10.3
```

### Checklist

grhkm21:
Fixes #38089. The issue comes from that the construction tower of functors for multivariate polynomial rings behaves incorrectly, since the `_apply_functor` method is incorrect. This PR fixes that.

### :memo: Checklist

### DIFF
--- a/src/sage/categories/pushout.py
+++ b/src/sage/categories/pushout.py
@@ -1055,9 +1055,18 @@ class MultiPolynomialFunctor(ConstructionFunctor):
             Multivariate Polynomial Ring in x, y, z over Integer Ring
             sage: F(RR)          # indirect doctest                                     # needs sage.rings.real_mpfr
             Multivariate Polynomial Ring in x, y, z over Real Field with 53 bits of precision
+
+        Verify that single-variable multivariate polynomial rings are created
+        correctly (:issue:`38089`)::
+
+            sage: R = PolynomialRing(ZZ, 1, "z")
+            sage: F = R.construction()[0]; F
+            MPoly[z]
+            sage: F(ZZ) == R
+            True
         """
         from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
-        return PolynomialRing(R, self.vars)
+        return PolynomialRing(R, len(self.vars), self.vars)
 
     def __eq__(self, other):
         """
--- a/src/sage/matroids/minorfix.h
+++ b/src/sage/matroids/minorfix.h
@@ -1,7 +1,3 @@
 // On some systems, macros "minor()" and "major()" are defined in system header files. This will undefine those:
-#ifdef major
-    #undef major
-#endif
-#ifdef minor
-    #undef minor
-#endif
+#undef major
+#undef minor
--- a/src/sage/schemes/elliptic_curves/hom_composite.py
+++ b/src/sage/schemes/elliptic_curves/hom_composite.py
@@ -754,9 +754,11 @@ class EllipticCurveHom_composite(EllipticCurveHom):
 
         TESTS::
 
-            sage: f = phi.codomain().defining_polynomial()                              # needs sage.rings.finite_rings
-            sage: g = E.defining_polynomial().subs({2:1})                               # needs sage.rings.finite_rings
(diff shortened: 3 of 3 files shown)


tscrim:
This is causing some doctest failures:
```
sage -t --long --random-seed=286735480429121101562228604801325644303 src/doc/en/reference/coercion/index.rst  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/categories/pushout.py  # 3 doctests failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/schemes/elliptic_curves/hom_composite.py  # 1 doctest failed
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/schemes/generic/morphism.py  # 3 doctests failed
```
However, some of these behavior changes are not good IMO. I think we should favor the univariate implementation, which has more features and speed, than the multivariate. There are some inconsistencies with the "multi"variate case having Singular as a backend that the univariate implementation doesn't have, but this change could have a lot of unintended consequences (including in performance and further coercions). So I'm not sure if we should change this.

grhkm21:
IMO if the user specifically wants multivariate polynomial rings (by constructing with `PolynomialRing(R, 1, "x")`) then we should stick with that. I agree that univariate polynomial rings have more features, but the issue is multivariate polynomial rings have different APIs. Also the original issue #38089 wants this.

There is only one "real" doctest failure at `hom_composite.py`, I will have a look at that

grhkm21:
The hom_composite test itself is a bit sketchy... [here](https://github.com/sagemath/sage/blob/develop/src/sage/schemes/elliptic_curves/hom_composite.py#L757-L760) `f(*phi_rational_maps(), 1)` is an element of $\mathrm{Frac}(\mathbb{F}_p[x, y])$, while `g` is an element of $\mathbb{F}_p[x, y, z]$. That's why the output depends on the coercion model. I have patched it.

tscrim:
However the failing doctests are not when the user is explicitly asking for the Singular univariate (i.e., "multi"variate) implementation. They are both univariate polynomials, just different implementations (output repr not withstanding as that should be changed IMO). Of course, the API's for the univariate and the multivariate should be compatible in an ideal world, but I don't know how feasible that is offhand. In any case, I think if you want a specific implementation, then you should manually convert things to the one you want and not rely on automated processes. Again, we should favor the univariate implementations when something is being done automatically, and changing this output is backwards-incompatible as the API's are different and some features will be missing.

On the other hand, this means the Singular univariate implementation would pass its `_test_construction`. Yet, I think we should do it such a way that the other pushouts still favor the univariate case.

I also feel compelled to remind you that just because there is an issue requesting it does not mean we should do it.

grhkm21:
I implemented this PR because the issue looked like a bug to me. I will close this.

tscrim:
There is a definite bug with the fact the applying the output of `construction()` does not yield the same ring back. I'm not sure right now how to fix that without doing things as you proposed in this PR.

Thank you for trying this. It did create useful data with seeing happened. Dealing with pushouts and polynomial rings is very tricky; I have run into many subtleties over the years most times I go into this.

I'm not sure we should close this outright either. We might just have to accept and deal with the effects of this change. I'm just currently of the opinion that this isn't the right way to do things, but not 100% sold on that either.

maxale:
As this fix did not happen, could you please reopen my issue #38089 ?

PR #38108, opened 2024-05-29
Implicit function solver for lazy series

mantepse:
This pull request provides a method `define_implicitly` for lazy rings that allows to produce lazy power series solutions for systems of functional equations, in relatively great generality, superseding #37033.

We use a special action to make sure that coercion between `CoefficientRing` and its base ring works as desired. However, it is a local change, if the current approach eventually shows its shortcomings. (Alternative approach previously considered: Using a special functor to control the pushout behaviors. The action approach was chosen as a way to prevent potential conflicts with other construction functors that may appear in the future.)

Dependencies: #38729


mantepse:
There is a time-sink I don't completely understand:
```
sage: L.<t, u> = LazyPowerSeriesRing(h.fraction_field())
sage: D = L.undefined()
sage: s1 = L.sum(lambda n: h[n]*t^(n+1)*u^(n-1), 1)
sage: L.define_implicitly([D], [u*D - u - u*s1*D - t*(D - D(t, 0))])
sage: D[0]
h[]
sage: D[1]
0
sage: D[2]
h[1]*t^2
sage: %prun D[3]
         3357653 function calls (3326279 primitive calls) in 3.081 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
[...]
  777/111    0.038    0.000    2.599    0.023 {method 'gcd' of 'sage.rings.polynomial.multi_polynomial.MPolynomial' objects}
[...]
  793/110    0.012    0.000    2.226    0.020 unique_factorization_domains.py:134(_gcd_univariate_polynomial)
[...]
sage: D[3]
0
```

mantepse:
That time sink is also present with `L.<t, u> = LazyPowerSeriesRing(QQbar["x"].fraction_field())`, so I guess it cannot be easily fixed.

mantepse:
Notes to myself, concerning the time sink.  If I understand correctly, it is dominated by the code in `Stream_uninitialized.__getitem__` that creates the new undetermined element of the stream.

I tried a bit to optimize this in the special case where `isinstance(self._coefficient_ring, MPolynomialRing_polydict)`, mainly by bypassing `MPolynomialRing_polydict.__call__` and the like.  Curiously, it turned out that creating the string representation of the monomials takes a lot of time (it invokes multiplication and gcd!).  However, even with all these optimizations, the performance was still quite bad.  Essentially, we are using a lot of time in `gcd` computations, and it is not clear which of these can be avoided.
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
...
  1886         2       1883.0    941.5      0.0          for n0 in range(len(self._cache) + self._approximate_order, n+1):
  1887                                                       # WARNING: coercing the new variable to self._PF slows
  1888                                                       # down the multiplication enormously
  1889         1      12524.0  12524.0      0.0              if self._coefficient_ring == self._base_ring:
  1890                                                           x = (self._pool.new_variable(self._name + "[%s]" % n0)
  1891                                                                * self._terms_of_degree(n0, self._P)[0])
  1892                                                       else:
  1893         2   26085266.0    1e+07     11.6                  x = sum(self._pool.new_variable(self._name + "[%s]" % m) * m
  1894         1    4885896.0    5e+06      2.2                          for m in self._terms_of_degree(n0, self._P))
  1895         1  194465377.0    2e+08     86.2              x = self._U(x)
  1896         1       1333.0   1333.0      0.0              self._cache.append(x)
  1897                                           
  1898         1       2886.0   2886.0      0.0          return x
```

mantepse:
Hm, somebody is calling `Rings.ParentMethods.__getitem__` an awful lot.

mantepse:
Oh wow.  I extracted the bit.  Creating `FESDUMMY_13/h[]*t^4 + FESDUMMY_12/h[]*t^3*u + FESDUMMY_11/h[]*t^2*u^2 + FESDUMMY_10/h[]*t*u^3 + FESDUMMY_9/h[]*u^4` takes more than a second on my computer!
```
def testit(n0, _P, _U, _terms_of_degree):
    """
    sage: from sage.data_structures.stream import CoefficientRing, VariablePool
    sage: _base_ring = SymmetricFunctions(QQ).h().fraction_field()
    sage: _coefficient_ring = _base_ring["t", "u"]
    sage: _PF = CoefficientRing(_base_ring)
    sage: _U = _coefficient_ring.change_ring(_PF); _U
    sage: _P = _PF.base(); _P
    sage: _terms_of_degree = lambda n, R: [m.change_ring(R) for m in _coefficient_ring.monomials_of_degree(n)]
    sage: %time testit(4, _P, _U, _terms_of_degree)
    CPU times: user 1.3 s, sys: 4.02 ms, total: 1.3 s
    Wall time: 1.3 s
    FESDUMMY_13/h[]*t^4 + FESDUMMY_12/h[]*t^3*u + FESDUMMY_11/h[]*t^2*u^2 + FESDUMMY_10/h[]*t*u^3 + FESDUMMY_9/h[]*u^4
    """
    _pool = VariablePool(_P)
    x = sum(_pool.new_variable("[%s]" % m) * m for m in _terms_of_degree(n0, _P))
    return _U(x)
```

mantepse:
Apparently, there is a hidden cost in the dense implementation of the infinite polynomial ring, I'll ask a question on sage-devel.

src/sage/combinat/sf/sfa.py
```diff
@@ -1877,6 +1877,9 @@ def __init__(self, Sym, basis_name=None, prefix=None, graded=True):
 
     _print_style = 'lex'
 
+    def is_prime_field(self):
+        return False
```
tscrim: This needs a doctest.

src/sage/rings/polynomial/multi_polynomial_sequence.py
```diff
@@ -298,7 +299,11 @@ def PolynomialSequence(arg1, arg2=None, immutable=False, cr=False, cr_str=None):
     except ImportError:
         BooleanMonomialMonoid = ()
 
-    is_ring = lambda r: isinstance(r, (MPolynomialRing_base, BooleanMonomialMonoid)) or (isinstance(r, QuotientRing_nc) and isinstance(r.cover_ring(), MPolynomialRing_base))
+    is_ring = lambda r: (isinstance(r, (MPolynomialRing_base,
+                                        BooleanMonomialMonoid,
+                                        InfinitePolynomialRing_sparse))
```
tscrim, line 318: It would be good to add a doctest showing that it now works with infinite polynomial rings.

src/sage/rings/lazy_series_ring.py
```diff
@@ -655,11 +657,415 @@ def undefined(self, valuation=None):
...
+        ``n`` such that coefficients are in the ring ``R``.
+
+        If ``self`` is a univariate Laurent, power, or Dirichlet
+        series, this is the list containing the one of the base ring.
+
+        If ``self`` is a multivariate power series, this is the list
+        of monomials of total degree ``n``.
+
+        If ``self`` is a lazy symmetric function, this is the list
+        of basis elements of total degree ``n``.
```
tscrim: This data/explanation is better with the concrete implementations as this could fall out of date as new implementations are added.

src/sage/rings/lazy_series_ring.py
```diff
@@ -2794,6 +3255,50 @@ def _monomial(self, c, n):
...
+        B = self._internal_poly_ring.base_ring()
+        B = B.change_ring(R)
+        if self._arity == 1:
+            return list(B.homogeneous_component_basis(n))
+        l = []
+        for c in IntegerVectors(n, self._arity):
+            for m in cartesian_product_iterator([F.homogeneous_component_basis(p)
+                                                 for F, p in zip(B.tensor_factors(), c)]):
+                l.append(tensor(m))
+        return l
```
tscrim:
```suggestion
        return [tensor(m) for c in IntegerVectors(n, self._arity)
                for m in cartesian_product_iterator([F.homogeneous_component_basis(p)
                                                     for F, p in zip(B.tensor_factors(), c)])]
```

src/sage/data_structures/stream.py
```diff
@@ -1203,14 +1259,267 @@ def iterate_coefficients(self):
...
+
+
+class CoefficientRing(UniqueRepresentation, FractionField_generic):
+    r"""
+    The class of unknown coefficients in a stream.
+    """
+    def __init__(self, base_ring):
+        """
+
+        EXAMPLES::
```
tscrim:
```suggestion
        """
        Initialize ``self``.

        EXAMPLES::
```

src/sage/data_structures/stream.py
```diff
@@ -1241,11 +1550,643 @@ def __init__(self, approximate_order, true_order=False):
...
+
+    def _subs_in_caches(self, var, val):
+        r"""
+        Substitute ``val`` for ``var`` in the caches of the input
+        streams and update ``self._good_cache``.
+
+        INPUT:
+
+        - ``var``, a variable in ``self._P``
+        - ``val``, the value that should replace the variable
```
tscrim:
```suggestion
        - ``var`` -- a variable in ``self._P``
        - ``val`` -- the value that should replace the variable
```

src/sage/data_structures/stream.py
```diff
@@ -2882,12 +3887,12 @@ def _approximate_order(self):
         EXAMPLES::
 
             sage: from sage.data_structures.stream import Stream_function, Stream_dirichlet_invert
-            sage: f = Stream_function(lambda n: n, True, 1)
+            sage: f = Stream_function(lambda n: QQ(n), True, 1)
```
tscrim, line 3896: Why is this necessary? This seems like a regression.

tscrim (changes requested):
Thank you. There just a few more details I have seen when giving this another look over.

I have one additional question. This gets rid of the old implicit definition, correct? I expect this to be slower than the "solver" we have before for `f = G(f)`. Thus, I am thinking we should keep that implementation for that special case. It's possible I missed something when looking at the code.

mantepse:
> Thank you. There just a few more details I have seen when giving this another look over.
> 
> I have one additional question. This gets rid of the old implicit definition, correct? I expect this to be slower than the "solver" we have before for `f = G(f)`. Thus, I am thinking we should keep that implementation for that special case. It's possible I missed something when looking at the code.

No, the special case is treated as before - although it is also handled by `Stream_uninitialized`.  The distinction is made by checking whether `self._target is None`, see `Stream_uninitialized.__getitem__`.

Yes, `f.define(G(f))` is *much* faster, whenever it works.

tscrim:
Ah, okay, thank you. Then once the other details are addressed, then this will be a positive review.

mantepse:
Cool, I'm on it.  Good evening? :-)

mantepse:
thank you, well, better indeed, but still somewhat so-so :-)

Currently running tests.  Short summary: `is_prime_field` is now indeed unnecessary, `QQ` in `Stream_function(lambda n: QQ(n), True, 1)` is necessary, because we expect elements of the stream to be in a correct parent, and `lambda n: n` in fact produces `int`.

src/sage/rings/lazy_series_ring.py
mantepse: The specifications are also in the concrete implementations.  I slightly reworded to make it clear that this is intended as an explanation of the purpose of the (abstract) method.

mantepse:
I cannot reproduce the failure in `symmetric_ideal.py`.

mantepse:
I marked the one doctest failure on macos as random.  Apparently, the order of the variables is different on macos than on linux.  However, it is more valuable to have the full output of the doctest, and, since it is an error message, I do not want to try to make the error message less random, which would involve some work and introduce more brittleness.

src/sage/rings/lazy_series_ring.py
```diff
@@ -655,11 +657,413 @@ def undefined(self, valuation=None):
...
+            sage: L.define_implicitly([F], [F(2*z) - (1+exp(x*z)+exp(y*z))*F - exp((x+y)*z)*F(-z)])
+            sage: F
+            <repr(...) failed: ValueError: could not determine any coefficients:
+                coefficient [3]: 6*series[3] + (-2*x - 2*y)*series[2] + (x*y)*series[1] == 0>
+
+            sage: F = L.undefined()
+            sage: L.define_implicitly([(F, [0, f1])], [F(2*z) - (1+exp(x*z)+exp(y*z))*F - exp((x+y)*z)*F(-z)])
+            sage: F  # random
+            <repr(...) failed: ValueError: could not determine any coefficients:
+                coefficient [3]: 6*series[3] + (-2*x - 2*y)*series[2] + (x*y*f1) == 0>
```
tscrim:
```suggestion
        The following test results in a failure, but at different points
        on different machines::

            mac: coefficient [3]: 6*series[3] + (-2*x - 2*y)*series[2] + (x*y*f1) == 0
            linux: ...
        
        Due to this, we only test that the base output::

            sage: F  # random
            <repr(...) failed: ValueError: could not determine any coefficients:
                coefficient ... == 0>
```
I don't think it is good to mark this as `# random` since that then only checks that an (uncaught) error is not raised. Instead, I think we should be more specific, both in the test and in the explanation.

tscrim (approved):
Thank you for the changes. This now LGTM. Thank you for all of your hard work in getting this implemented. This is a major new feature.

One last thing before a positive review.

src/sage/rings/lazy_series_ring.py
mantepse: I don't mind, but did you intend to remove#random in 812?

tscrim: Ah, yes, I meant to remove that.

mantepse: Oh, and it is only on some MacOS versions

mantepse: Oh, and it is only on some MacOS versions

mantepse: Actually, I think that "but at different points on different machines::" might be misunderstood.  The only thing that may differ is the order in which the monomials in the polynomial `6*series[3] + (-2*x - 2*y)*series[2] + (x*y*f1)` appear.  So here is another suggestion:

```suggestion
        We have thus specified that the degree 1 component equals `f1`.  This is still
        not enough to remove the ambiguity, so an error is raised.  However, we can
        see that the dependence on `series[1]` disappeared.  The equation which has
        no unique solution is now
        `6*series[3] + (-2*x - 2*y)*series[2] + (x*y*f1) == 0`::

            sage: F  # random
            <repr(...) failed: ValueError: could not determine any coefficients:
                coefficient [3]: ... == 0>
```

tscrim: If you're removing the `# random`, then okay. Although it would be good to explain a bit why the `...` is in the doctest here. Can you give the two different outputs, at least here on this PR for the record?

Also, it would be {{{``}}} on both sides for code formatting.

mantepse:
After  #39523, the doctests after `The Frobenius character of labelled Dyck words`, lines 895-905, will not be slow anymore.

tscrim (approved):
Thank you. Let's get this in. Any additional improvements or fixes can be done on later on followup PRs.

PR #37198, opened 2024-01-30
Translate "A Tour of Sage" into Greek

gmou3:
### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] I have updated the documentation accordingly.

PR #38109, opened 2024-05-29
Add Greek tutorial

gmou3:
A simple first version translated and edited from the English version.
Articles worth including in the future: `programming`.

Follows #37198. Resolves #19960.

P.S. The tutorial in English, and possibly other articles, are too verbose and one is lost in the noise. For example, they include many comments about Python in general, or they even describe basic `Tab` completion, using the up arrow, and they include a whole chapter on the standard order of operations (which I blindly translated and later removed, being struck by the endeavor's futility). I doubt any human benefits from this; on the contrary, they lose time and motivation.


kwankyu:
> A simple first version translated and edited from the English version. Articles worth including in the future: `programming`.

Are you Greek?

> P.S. The tutorial in English, and possibly other articles, are too verbose and one is lost in the noise. For example, they include many comments about Python in general, or they even describe basic `Tab` completion, using the up arrow, and they include a whole chapter on the standard order of operations (which I blindly translated and later removed, being struck by the endeavor's futility). I doubt any human benefits from this; on the contrary, they lose time and motivation.

Then why not edit the English version as well? 

On the other hand, as it is a tutorial, perhaps, we should assume very little in the user's background knowledge...

gmou3:
> Are you Greek?

Yes.

> Then why not edit the English version as well?

I may do that.

kwankyu:
> > Are you Greek?
> 
> Yes.

As far as I know, translated work is reviewed by someone who can read the language.

Do we have someone who can review this PR?
 
> > Then why not edit the English version as well?
> 
> I may do that.

Thanks. I suppose it needs to be updated anyway.

kwankyu:
@saraedum may do.

saraedum:
I'm a bit confused by the scope here. I understand thta you translated "Introduction" and a part of "The Interactive Shell."

Now, I am not sure if that's really helpful as a tutorial. Is there a plan to extend this further? Or should we link at least to the bits in English that are "missing"?

gmou3:
I stopped translating more because I didn't particularly like the old articles.
For example, I did rewrite the interactive shell in a more condensed manner.

If you can propose a minimal extension which can stand alone, I will try translating/rewriting.
Also let me know if you think that it won't work without much more work, or whether you think that this translating endeavor is somehow undesirable.

saraedum:
> Also let me know if you think that it won't work without much more work

I have no strong opinion here but I fear that what you have so far won't be too useful to the average Greek students who doesn't read English.

> or whether you think that this translating endeavor is somehow undesirable.

Translations are certainly not undesirable :)

I am not sure how useful they actually are though. I guess they are used but I have no idea what the audience is and what would be useful to them.

PR #37821, opened 2024-04-17
Fix category of ModularFormsRing and documentation cleanup

DavidAyotte:
The goal of this PR is twofold:

First, we fix the category of modular forms ring as per PR #37797 to make the following code work:
```
sage: M = ModularFormsRing(1)
sage: M['z', 'w']
Multivariate Polynomial Ring in z, w over Ring of Modular Forms for Modular Group SL(2,Z) over Rational Field
```

Second, we do a documentation and PEP8 cleanup of the class `ModularFormsRing`.

CC: @fchapoton 

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation and checked the documentation preview.

Issue #26254, opened 2018-09-11
Use Cython directive binding=True to get signatures for cython methods


kwankyu:
<div id="comment:0"></div>

In requests for help for cythonized built-in methods, the signature of the method is not shown, unlike normal python methods.  For an example,

```
sage: a=17 
sage: a.quo_rem?

Docstring:     
   Returns the quotient and the remainder of self divided by other.
   Note that the remainder returned is always either zero or of the
   same sign as other.

   INPUT:

   * "other" - the divisor

   OUTPUT:

   * "q" - the quotient of self/other

   * "r" - the remainder of self/other

   EXAMPLES:

      sage: z = Integer(231)
      sage: z.quo_rem(2)
      (115, 1)
...
``` 

To fix this, we set Cython directive `binding=True`. Thus we buy

- better support of introspection into cython methods, 
- consistent help messages for cython methods like python methods, 
- better tracebacks on exceptions, 
- better behaving for documentation

for slight performance degradation due to increased overhead cost of calling cython methods. 

Related tickets: #19100, #20860, #18192

Depends on #32509
Depends on #33864

CC:  @jdemeyer @tscrim @mkoeppe

Component: **user interface**

Author: **Kwankyu Lee, Tobias Diez**

Branch/Commit: **[public/26254](https://github.com/sagemath/sagetrac-mirror/tree/public/26254) @ [`326f19c`](https://github.com/sagemath/sagetrac-mirror/commit/326f19c6c36d7585cbb87e11bb3da81cfbc7ee19)**

Reviewer: **Tobias Diez, ...**

_Issue created by migration from https://trac.sagemath.org/ticket/26254_

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
Changed commit from **[`f5bc80d`](https://github.com/sagemath/sagetrac-mirror/commit/f5bc80d0d33902c088753ee9592972b72e6619e7)** to **[`f38dc83`](https://github.com/sagemath/sagetrac-mirror/commit/f38dc83cf09c8c1a4e784b97233f1d1a9d635fb0)**

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
<div id="comment:80"></div>

Branch pushed to git repo; I updated commit sha1. This was a forced push. New commits:
[...]

PR #38117, opened 2024-05-31
Add small examples of common docstring styles

kwankyu:
which I think are already established.  But it is good to recommend preferred styles to keep docstrings in consistent styles. 

motivated by the comment: https://github.com/sagemath/sage/pull/37821#discussion_r1621555198

### :memo: Checklist

### :hourglass: Dependencies

#38068 (merged here)

### DIFF
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -355,6 +355,8 @@ information. You can use the existing functions of Sage as templates.
 
        - ``p`` -- prime integer (default: `2`); coprime with `n`
 
+       - ``var`` -- string (default: ``'lambda'``)
+
        - ``check`` -- boolean (default: ``True``); specifies whether to check for primality
 
        - ``algorithm`` -- (default: ``None``) the name of the algorithm to use
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -353,7 +353,7 @@ information. You can use the existing functions of Sage as templates.
 
        - ``n`` -- integer
 
-       - ``p`` -- prime integer (default: `2`); coprime with ``n``
+       - ``p`` -- prime integer (default: `2`); coprime with `n`
 
        - ``check`` -- (default: ``True``) specifies whether to check primality
 
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -254,14 +254,14 @@ The top of each Sage code file should follow this format::
 
     AUTHORS:
 
-    - YOUR NAME (2005-01-03): initial version
+    - Your Name (2024-01-13): initial version
     - Alice Liddell (2024-05-31): added a method; cleaned docstrings
-    - Full name (year-month-day): short description
+    - Full name (YYYY-MM-DD): short description
 
     """
 
     # ****************************************************************************
-    #       Copyright (C) 2013 YOUR NAME <your email>
+    #       Copyright (C) 2024 Your Name <your email>
     #
     # This program is free software: you can redistribute it and/or modify
     # it under the terms of the GNU General Public License as published by
@@ -343,7 +343,7 @@ information. You can use the existing functions of Sage as templates.
 
    1. The type names should be descriptive, but do not have to represent the
       exact Sage/Python types. For example, use "integer" for anything that
-      behaves like an integer, rather than ``int`` or ``Integer``.
+      behaves like an integer, rather than "int" or "Integer".
 
    2. Mention the default values of the input arguments when applicable.
 
@@ -355,7 +355,9 @@ information. You can use the existing functions of Sage as templates.
 
        - ``p`` -- prime integer (default: `2`); coprime with `n`
 
-       - ``check`` -- (default: ``True``) specifies whether to check primality
+       - ``check`` -- boolean (default: ``True``); specifies whether to check for primality
+
+       - ``algorithm`` -- (default: ``None``) the name of the algorithm to use
 
    The OUTPUT block describes the expected output. This is required if the
    one-sentence description of the function needs more explanation.
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -255,8 +255,8 @@ The top of each Sage code file should follow this format::
     AUTHORS:
 
     - YOUR NAME (2005-01-03): initial version
-
-    - person (date in ISO year-month-day format): short desc
+    - Alice Liddell (2024-05-31): added a method; cleaned docstrings
+    - Full name (year-month-day): short description
 
     """
 
@@ -341,9 +341,9 @@ information. You can use the existing functions of Sage as templates.
 
    The INPUT block describes all arguments that the function accepts.
 
-   1. The type names should be descriptive, but do not have to represent
-      the exact Sage/Python types. For example, use "integer" for
-      anything that behaves like an integer, rather than ``int``.
+   1. The type names should be descriptive, but do not have to represent the
+      exact Sage/Python types. For example, use "integer" for anything that
+      behaves like an integer, rather than ``int`` or ``Integer``.
 
    2. Mention the default values of the input arguments when applicable.
 
@@ -354,6 +354,8 @@ information. You can use the existing functions of Sage as templates.
        - ``n`` -- integer
 
        - ``p`` -- prime integer (default: `2`); coprime with ``n``
+
+       - ``check`` -- (default: ``True``) specifies whether to check primality
 
    The OUTPUT block describes the expected output. This is required if the
    one-sentence description of the function needs more explanation.
--- a/src/doc/en/developer/coding_basics.rst
+++ b/src/doc/en/developer/coding_basics.rst
@@ -657,12 +657,12 @@ indentation:
 
         INPUT:
 
-        - ``x`` -- integer (default: ``1``); the description of the
+        - ``x`` -- integer (default: `1`); the description of the
           argument ``x`` goes here. If it contains multiple lines, all
           the lines after the first need to begin at the same indentation
           as the backtick.
 
-        - ``y`` -- integer (default: ``2``); the description of the
+        - ``y`` -- integer (default: `2`); the description of the
           argument ``y``
 
         OUTPUT: tuple; further description of the output
@@ -673,7 +673,7 @@ indentation:
 
             sage: A = EuclideanSpace(2)
             sage: A.point(2, 3)
-            (32, 3)
+            (2, 3)
 
         We now ... ::
 
@@ -690,7 +690,7 @@ indentation:
         .. NOTE::
 
             This function uses :func:`pow` to determine the fifth
-            power of ``x``.
+            power of `x`.
 
         ...
 
@@ -700,7 +700,7 @@ indentation:
 
         TESTS::
 
-            sage: A.point(42, 0)  # Check for corner case y = 0
+            sage: A.point(42, 0)  # check for corner case y = 0
             ...
         """
         <body of the function>


src/doc/en/developer/coding_basics.rst
```diff
@@ -255,8 +255,8 @@ The top of each Sage code file should follow this format::
     AUTHORS:
 
     - YOUR NAME (2005-01-03): initial version
```
gmou3:
```suggestion
    - Your Name (2024-01-13): initial version
```

src/doc/en/developer/coding_basics.rst
```diff
@@ -255,8 +255,8 @@ The top of each Sage code file should follow this format::
...
-
-    - person (date in ISO year-month-day format): short desc
+    - Alice Liddell (2024-05-31): added a method; cleaned docstrings
+    - Full name (year-month-day): short description
```
gmou3:
```suggestion
    - Full name (YYYY-MM-DD): short description
```

src/doc/en/developer/coding_basics.rst
```diff
@@ -341,9 +341,9 @@ information. You can use the existing functions of Sage as templates.
 
    The INPUT block describes all arguments that the function accepts.
 
-   1. The type names should be descriptive, but do not have to represent
-      the exact Sage/Python types. For example, use "integer" for
-      anything that behaves like an integer, rather than ``int``.
+   1. The type names should be descriptive, but do not have to represent the
+      exact Sage/Python types. For example, use "integer" for anything that
+      behaves like an integer, rather than ``int`` or ``Integer``.
```
gmou3:
```suggestion
      behaves like an integer, rather than "int" or "Integer".
```

src/doc/en/developer/coding_basics.rst
gmou3: Also change the year of the copyright from 2013 to 2024, and avoid all caps.

src/doc/en/developer/coding_basics.rst
```diff
@@ -353,7 +353,9 @@ information. You can use the existing functions of Sage as templates.
 
        - ``n`` -- integer
 
-       - ``p`` -- prime integer (default: `2`); coprime with ``n``
+       - ``p`` -- prime integer (default: `2`); coprime with `n`
+
+       - ``check`` -- (default: ``True``) specifies whether to check primality
```
gmou3:
```suggestion
       - ``check`` -- boolean (default: ``True``); specifies whether to check for primality
```

gmou3: I also prefer tight lists, but there is no consensus on that it seems.

gmou3 (changes requested):

src/doc/en/developer/coding_basics.rst
```diff
@@ -353,7 +353,9 @@ information. You can use the existing functions of Sage as templates.
 
        - ``n`` -- integer
 
-       - ``p`` -- prime integer (default: `2`); coprime with ``n``
+       - ``p`` -- prime integer (default: `2`); coprime with `n`
```
gmou3, line 356:
```suggestion
       - ``p`` -- prime integer (default: ``2``); coprime with `n`
```

gmou3: or no backticks at all?

gmou3 (changes requested):

src/doc/en/developer/coding_basics.rst
kwankyu: I prefer spaced lists. Right... no consensus.

Line 358 is by purpose. Type of an input argument may be (is often) omitted. In that case, semicolon seems redundant.

gmou3: Hm, I see. It would be nice to have an example for booleans too, because people may write it as "bool" or "Boolean". Maybe we should add an extra example with default value ``None`` that has no type.

src/doc/en/developer/coding_basics.rst
kwankyu: In sage, input values are often mathematical. This line shows an example: integer 2.

kwankyu: like n in "coprime with n".

src/doc/en/developer/coding_basics.rst
gmou3: Something like
```
- ``algorithm`` -- (default: ``None``) the name of the algorithm to use
```
However I think it is problematic to start a sentence with a parenthetical remark.

src/doc/en/developer/coding_basics.rst
gmou3: If that is the case I will need to revert my change in #38068.

Why isn't 
- ``n`` -- integer
written as
- `n` -- integer?

(with single instead of double backticks)

src/doc/en/developer/coding_basics.rst
gmou3: How about
```
- ``algorithm`` -- default: ``None``; the name of the algorithm to use
```
?

kwankyu: > Something like
> 
> ```
> - ``algorithm`` -- (default: ``None``) the name of the algorithm to use
> ```

Yes. This example may be better.

> However I think it is problematic to start a sentence with a parenthetical remark.

I regard that as a convention (for default value). We should put the parenthetical phrase somewhere in the line. I see no better place. I think that this choice is already established.

gmou3: Ok, let's leave it at that.

src/doc/en/developer/coding_basics.rst
kwankyu: The thing before `--` denotes the input argument (code snippet). 
```
The input argument ``n`` denotes an integer `n`.
```

gmou3: Shouldn't the default value also be considered as a code snippet?

kwankyu: The default value is a value of the input argument. The value is often mathematical in sage. Then we want to show it as math. It may be even
```
``m`` -- matrix (default: identity matrix)
```
or
```
``m`` -- matrix (default: `I`)
```

gmou3: Ok, I think it is indeed better to write everything after `--` in the nicest way available (math for math).

src/doc/en/developer/coding_basics.rst
kwankyu: Done.

src/doc/en/developer/coding_basics.rst
kwankyu: Done.

src/doc/en/developer/coding_basics.rst
kwankyu: Done.

src/doc/en/developer/coding_basics.rst
kwankyu: Thanks. Done.

src/doc/en/developer/coding_basics.rst
kwankyu: I merged #38068 here, made the suggested changes: https://github.com/sagemath/sage/pull/38068/files#issuecomment-2142935336

Thanks.

src/doc/en/developer/coding_basics.rst
```diff
@@ -254,14 +254,14 @@ The top of each Sage code file should follow this format::
 
     AUTHORS:
 
-    - YOUR NAME (2005-01-03): initial version
-
-    - person (date in ISO year-month-day format): short desc
+    - Your Name (2024-01-13): initial version
+    - Alice Liddell (2024-05-31): added a method; cleaned docstrings
+    - Full name (YYYY-MM-DD): short description
```
gmou3, line 259:
```suggestion
    - Full Name (YYYY-MM-DD): short description
```

src/doc/en/developer/coding_basics.rst
```diff
@@ -649,44 +653,44 @@ indentation:
...
             ...
-            TypeError: unable to convert 'r' to an integer
+            TypeError: unable to convert 'x' to an integer
 
         .. NOTE::
 
-            This function uses the algorithm of [BCDT2001]_ to determine
-            whether an elliptic curve `E` over `Q` is modular.
+            This function uses :func:`pow` to determine the fifth
+            power of ``x``.
```
gmou3:
```suggestion
            power of `x`.
```

gmou3 (changes requested):

src/doc/en/developer/coding_basics.rst
kwankyu: Done.

src/doc/en/developer/coding_basics.rst
```diff
@@ -661,32 +665,32 @@ indentation:
...
         EXAMPLES:
 
         This example illustrates ... ::
 
-            sage: A = ModuliSpace()
-            sage: A.point(2,3)
-            xxx
+            sage: A = EuclideanSpace(2)
+            sage: A.point(2, 3)
+            (2, 3)
```
gmou3, line 678:
```suggestion
            (32, 3)
```

gmou3: The function returns `(x^5, y)`.

gmou3 (changes requested):

gmou3 (approved):
Thanks. Looks good.

kwankyu:
Thanks. A good example of collaborative work :-)

src/doc/en/developer/coding_basics.rst
```diff
@@ -254,14 +254,14 @@ The top of each Sage code file should follow this format::
 
     AUTHORS:
 
-    - YOUR NAME (2005-01-03): initial version
-
-    - person (date in ISO year-month-day format): short desc
+    - Your Name (2024-01-13): initial version
+    - Alice Liddell (2024-05-31): added a method; cleaned docstrings
+    - Full Name (YYYY-MM-DD): short description
```
gmou3, line 260: Oh! one more thing please...
```suggestion
```

gmou3, line 260:
```suggestion
```

gmou3: Oh! one more thing please...

gmou3 (changes requested):

gmou3:
Let's give this a bit more time.

@saraedum we can continue the discussion here.

src/doc/en/developer/coding_basics.rst
kwankyu: OK.  

Actually I inserted a blank line after the list intentionally to surround the list by blank lines, though it is strictly not necessary.  This is another point where people (or one person at different times :-) use different conventions.

However, I feel that no blank line at the end of the text is the trend, and we can save a line (perhaps useless).

kwankyu:
> Let's give this a bit more time.
> 
> @saraedum we can continue the discussion here.

OK. We can. If we reach to a consensus that we can implement, then we can make a new PR.

> I am unsure about what the rule is/was. 

I am too. There is no rule in sage community to decide what is a rule :-)

> I think the part after the "--" should be treated the same as the text body preceding the "INPUT". One can see how to type the default value from the function header.

I agree with this.

> My main argument is that the code base uses more double-backticks than single-backticks.

Of course, since most default values are `None`, `True`, or `False`. These are non-mathematical values.

> I believe that ... should not set new standards but report what the code base is already doing.

I agree. The problem is that there are examples for each of different styles.

> ... 1 is the actual default value as a Python int, it's not really a "mathematical object" arguably. I believe that the defaults should report the actual default of a function, imho this can never be a mathematical object.

I would not say that any of

(A)
```
(default: 1) 
```

(B)
```
(default: ``1``)
```

(C)
```
(default: `1`)
```
does not follow a rule. It all depends on the author's intention about the python function he is writing. If the value has no mathematical meaning, (A) is okay. If the value is a python object, (B) is the only choice. If the value is mathematical, then (C) is nice. 

If the function calculates an arc length of an ellipse and takes input a and b which are radians, the default values of a and b are mathematical,  then clearly (C) is preferable.

saraedum:
> > My main argument is that the code base uses more double-backticks than single-backticks.
> 
> Of course, since most default values are `None`, `True`, or `False`. These are non-mathematical values.

No, that's not what I meant. Sorry, I should have been more specific. I grepped for numerical values only here. Even with just integers, I get:

```
git grep -E 'default: `[0-9]+`'|wc -l
75
git grep -E 'default: ``[0-9]+``'|wc -l
325
```

saraedum:
> > I am unsure about what the rule is/was.
> 
> I am too. There is no rule in sage community to decide what is a rule :-)

People have varying preferences in SageMath and we do not enforce any particular code style. I don't think this is a huge problem but I remember that it's confusing when you are working on your first contributions that it's a bit unclear which example to follow.

> > I think the part after the "--" should be treated the same as the text body preceding the "INPUT". One can see how to type the default value from the function header.
> 
> I agree with this.

Actually, in my REPL this does not work reliably. For functions I see the signature with `?` but for methods it's not always shown:

```
sage: ZZ.coerce_map_from?
Docstring:
   Return a "Map" object to coerce from "S" to "self" if one exists,
   or "None" if no such coercion exists.

   EXAMPLES:

   By https://github.com/sagemath/sage/issues/12313, a special kind of
   weak key dictionary is used to store coercion and conversion maps,
   namely "MonoDict". In that way, a memory leak was fixed that would
   occur in the following test:

      sage: import gc
      sage: _ = gc.collect()
      sage: K = GF(1<<55,'t')
      sage: for i in range(50):
      ....:   a = K.random_element()
      ....:   E = EllipticCurve(j=a)
      ....:   b = K.has_coerce_map_from(E)
      sage: _ = gc.collect()
      sage: len([x for x in gc.get_objects() if isinstance(x, type(E))])
      1
Init docstring: Initialize self.  See help(type(self)) for accurate signature.
File:           /usr/lib/python3.12/site-packages/sage/structure/parent.pyx
Type:           builtin_function_or_method
```

That's probably a separate problem. If the signature were always shown with the docstring, then I don't think it's so important to spell out the defaults in that technical form.

> [...]
> In this example,
> 
> ```
>        - ``p`` -- prime integer (default: `2`); coprime with `n`
> ```
> 
> we want to say (C) is recommended.

I see your point here. I had another look at the documentation, and arguably `` `1` `` looks bad from a typography point of view:

image

Ideally, limitations in the renderer should not inform how we write docstrings but I remember that in Wikipedia there were similar rules about not using LaTeX for something simple like a 1 since it changes the font and breaks the flow for the reader.

From that perspective, a simple plain `default: 1` would be best here. I find the output of ```default: ``1`` ``` also acceptable:

image

gmou3:
Your argumentation about simple values not being LaTeXed sounds good to me. Indeed it is ugly otherwise.
So `(default: 1)` is probably the best choice, with single backticks being reserved for more complicated math (as should also be the case in other docstring text).

Maybe we should add a guiding example for every usual scenario. What about strings for example? A) string, B) "string", C) ``"string"``, or D) ``'string'``? I think one often encounters ``string``, which doesn't seem right.

gmou3:
> Actually, in my REPL this does not work reliably. For functions I see the signature with `?` but for methods it's not always shown

It works with `??` however (which is supposed to yield the source code).

kwankyu:
> > I am too. There is no rule in sage community to decide what is a rule :-)
> 
> People have varying preferences in SageMath and we do not enforce any particular code style. I don't think this is a huge problem but I remember that it's confusing when you are working on your first contributions that it's a bit unclear which example to follow.

I fully agree.

> Actually, in my REPL this does not work reliably. For functions I see the signature with `?` but for methods it's not always shown ...

Unfortunately, cython method does not always show the signature. This is an old issue: https://github.com/sagemath/sage/issues/26254

In this regard, there is subtle difference between our definitions of "default value". For me, default value is the value that the function assumes for an input when the user does not give a specific value. On the other hand,  the default value  in the function signature is always a python object (code), as you said. In my view, the former default value is what the latter default value represents when it is mathematical.

> ... and arguably `` `1` `` looks bad from a typography point of view:

Perhaps yes. But we are writing math $n$ or $x$ in the docstring. Why default value is an exception? 

If there is not much math in the function or that 1 and other integers are only values, then simple 1 is preferable. This is all what the author should decide. We cannot recommend one style unilaterally.

kwankyu:
> What about strings for example? A) string, B) "string", C) `"string"`, or D) `'string'`? I think one often encounters `string`, which doesn't seem right.

```
``'string'``
```
is what @mkoeppe writes. I fully agree with this style.

kwankyu:
Recently I saw 
```
(default: empty list)
```
I think this is perfectly okay. This is a good example where the default value is not the literal value in the function signature.

gmou3:
I propose the addition of a line such as
```
- ``var`` -- string (default: ``'lambda'``)
```

The 1 or \`1\` point is somewhat subtle. Can we maybe write \`1\` for consistency (because it is something mathematical) and have a LaTeX rule or package (`mathastext`?) do the typography work for us? That is, simplifying very simple math like a natural number to plain text.

kwankyu:
> I propose the addition of a line such as
> 
> ```
> - ``var`` -- string (default: ``'lambda'``)
> ```
> 

OK. I will do it.

> The 1 or `1` point is somewhat subtle. Can we maybe leave it at `1` for consistency (because it is something mathematical) ...

If you mean to do nothing about existing instances of 1 or 0 as default values, I do agree. As it is subtle, they should be treated case by case.

> ... simplifying very simple math like a natural number to plain text. (The latter to be done in another PR if someone wishes to pursue it.)

We don't do that when we write a paper. There "1" is usually math and written `$1$` in latex. On the other hand, "1" in "Section 1" is not math. I think we should follow the same rule in our docstrings. If a python function is not mathematical, and expects an integer as input, then it is \`1\`.

kwankyu:
> I propose the addition of a line such as
> 
> ```
> - ``var`` -- string (default: ``'lambda'``)
> ```

Done.

gmou3 (approved):
Thanks.

PR #38119, opened 2024-05-31
Fix hashing for Weyl algebra elements

darijgr:
Weyl algebras elements cannot be hashed or compared, since the element class is too barren.
This patch lets the element class inherit from IndexedFreeModuleElement, which fixes this problem automatically.

### DIFF
--- a/src/sage/algebras/weyl_algebra.py
+++ b/src/sage/algebras/weyl_algebra.py
@@ -21,7 +21,6 @@ from sage.misc.cachefunc import cached_method
 from sage.misc.latex import latex, LatexExpr
 from sage.misc.lazy_attribute import lazy_attribute
 from sage.misc.misc_c import prod
-from sage.structure.richcmp import richcmp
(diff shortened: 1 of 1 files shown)


tscrim (changes requested):
The `_richcmp_` and `__init__` are no longer needed as it will be inherited from `IndexedFreeModuleElement`. Please keep the doctests somewhere though.

These methods are also unnecessary but can simply be removed: `_add_`, `__neg__`, `monomial_coefficients`, `support`, 

Please add a test showing the left and right multiplication of the Weyl algebra by the corresponding polynomial ring generates the correct results. I am slightly worried about the `_acted_upon_` versus `_lmul_` and `_rmul_` not being compatible...

darijgr:
Good points. All done -- except for the doctests, which already exist.

tscrim (approved):
Okay, thanks. Then let's get this in.

darijgr:
Thank you, Travis!

Issue #35538, opened 2023-04-19
WSL: Chrome on Windows does not find the `_static` style in built HTML doc 


mkoeppe:
In the parts of the reference manual, `_static` is a symlink to `../_static`. Browsers running in Windows (tested with Chrome) cannot resolve the symlink, so the furo style is not applied.

Replacing the symlink by a copy of the directory is a workaround for this problem; but there also remains a problem loading mathjax in this configuration.

mkoeppe:
Related to discussion in #31415 @tobiasdiez @kwankyu @haraldschilly

haraldschilly:
Not sure why I'm mentioned. For the online doc files, which are served via a CDN that doesn't read symlinks as well, there is a small routine to fix it by [changing the html output](https://github.com/sagemath/documentation/blob/5f451107898c6f75206cbe1ecb4d5c5d81737dd4/index.py#L305). Sounds like this is the same or at least very similar.

kwankyu:
Right. The small routine replaces every mention of the symlink "`_static`" in html files to the target of the symlink. Thus the routine takes a while to do its work.

For example, the routine converts the line in `local/share/doc/sage/html/en/reference/schemes/index.html`
```
 <link rel="stylesheet" type="text/css" href="_static/pygments.css" />     
```
to
```
 <link rel="stylesheet" type="text/css" href="../_static/pygments.css" />     
```
So the html files treated by the routine does not use the symlink.

Perhaps the same routine could be used to fix the problem on windows.

Then why do we use the symlink in the first place? This is hard-wired in how sphinx works (the static files should be in a directory in the root). Perhaps we cannot change this.

kwankyu:
By the way, @haraldschilly, please consider https://github.com/sagemath/documentation/pull/27 for your next release. (test first)

haraldschilly:
> By the way, @haraldschilly, please consider [sagemath/documentation#27](https://github.com/sagemath/documentation/pull/27) for your next release. (test first)

Oh, I certainly forgot about it, thx for reminding me. Let's see how well this works with the next release :+1:

mkoeppe:
> Perhaps the same routine could be used to fix the problem on windows.

Yes! +1 to just build this into our scripts.

mkoeppe:
> why do we use the symlink in the first place? This is hard-wired in how sphinx works (the static files should be in a directory in the root). Perhaps we cannot change this.

[...]

PR #38129, opened 2024-06-02
Fix live doc build by saving runner space

kwankyu:
as seen in: https://github.com/sagemath/sage/actions/runs/9333677258/job/25690941738

### :memo: Checklist

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -175,16 +175,18 @@ jobs:
           name: docs
           path: docs.zip
 
-      - name: Remove docs
-        if: (success() || failure()) && steps.upload.outcome == 'success'
+      - name: Save space
+        id: savespace
+        if: (success() || failure()) && steps.upload.outcome == 'success' && github.repository == 'sagemath/sage' && github.ref == 'refs/heads/develop'
         run: |
           set -ex
-          # We remove previous docs directory to save space on runner device
+          # Save space on runner device before we start to build livedoc
           rm -rf docs/
+          rm -f docs.zip
 
       - name: Build live doc
         id: buildlivedoc
-        if: (success() || failure()) && steps.copy.outcome == 'success' && github.repository == 'sagemath/sage' && github.ref == 'refs/heads/develop'
+        if: (success() || failure()) && steps.savespace.outcome == 'success'
         run: |
           export MAKE="make -j5 --output-sync=recurse" SAGE_NUM_THREADS=5
           export PATH="build/bin:$PATH"
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -206,8 +206,8 @@ jobs:
         run: |
           mkdir -p ./livedoc
           # We copy everything to a local folder
-          docker cp BUILD:/sage/local/share/doc/sage/html livedoc
-          docker cp BUILD:/sage/local/share/doc/sage/pdf livedoc
+          docker cp -L BUILD:/sage/local/share/doc/sage/html livedoc
+          docker cp -L BUILD:/sage/local/share/doc/sage/pdf livedoc
           docker cp BUILD:/sage/local/share/doc/sage/index.html livedoc
           zip -r livedoc.zip livedoc
 
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -168,11 +168,19 @@ jobs:
           zip -r docs.zip docs
 
       - name: Upload docs
+        id: upload
         if: (success() || failure()) && steps.copy.outcome == 'success'
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-build.yml
```diff
@@ -168,12 +168,20 @@ jobs:
...
         with:
           name: docs
           path: docs.zip
 
+      - name: Remove docs
+        if: (success() || failure()) && steps.upload.outcome == 'success'
+        run: |
+          set -ex
+          # We remove previous docs directory to save space on runner device
+          rm -rf docs/
```
mkoeppe, line 184: Simpler to do that right after line 168

kwankyu: The "Upload docs" needs `docs/` directory and its contents. No?

mkoeppe: It's uploading the zip file

kwankyu: Ah, the zip file is not in the directory. 

Then perhaps removing the zip file as well would help. What do you think?

.github/workflows/doc-build.yml
```diff
@@ -195,11 +203,11 @@ jobs:
...
           mkdir -p ./livedoc
-          cp -r -L /sage/local/share/doc/sage/html ./livedoc
-          cp -r -L /sage/local/share/doc/sage/pdf ./livedoc
-          cp  /sage/local/share/doc/sage/index.html ./livedoc
+          # We copy everything to a local folder
+          docker cp BUILD:/sage/local/share/doc/sage/html livedoc
+          docker cp BUILD:/sage/local/share/doc/sage/pdf livedoc
+          docker cp BUILD:/sage/local/share/doc/sage/index.html livedoc
           zip -r livedoc.zip livedoc
-        shell: sh .ci/docker-exec-script.sh BUILD . {0}
```
kwankyu, line 213: Am I doing things right here?

.github/workflows/doc-build.yml
kwankyu: We remove `docs.zip` file too.

.github/workflows/doc-build.yml
kwankyu: I am asking because there is `-L` previously, but `docker cp` does not have the flag.

mkoeppe: For the corresponding option for `docker cp`, see https://github.com/sagemath/sage/pull/36730/files

mkoeppe: But the right way to do this, I think, is to finally do https://github.com/sagemath/sage/issues/35538#issuecomment-1532477914 as part of the docbuild...

kwankyu: Thanks. I use `-L` equivalent to `--follow-link` for now.

kwankyu: Rather more explicit `--follow-link` as in #36730.

mkoeppe:
Is there a test run already somewhere?

kwankyu:
I ran it here: https://github.com/kwankyu/sage/actions/runs/9424131740 Let's see.

kwankyu:
It works well.

mkoeppe:
Then let's merge it.

kwankyu:
Thanks!

kwankyu:
I am confused by this: https://github.com/sagemath/sage/actions/runs/9357979101

That run succeeded in building live doc. Then this PR is not really needed...?

kwankyu:
In particular, this step
```
      - name: Copy live doc
        id: copylivedoc
        if: (success() || failure()) && steps.buildlivedoc.outcome == 'success'
        run: |
          mkdir -p ./livedoc
          cp -r -L /sage/local/share/doc/sage/html ./livedoc
          cp -r -L /sage/local/share/doc/sage/pdf ./livedoc
          cp  /sage/local/share/doc/sage/index.html ./livedoc
          zip -r livedoc.zip livedoc
        shell: sh .ci/docker-exec-script.sh BUILD . {0}
```
worked as it is in the run. I thought this step would not work correctly because the commands run inside the docker container and hence create the file `livedoc.zip` inside the container, and the upload step fails because the zip file is not accessible. I am misunderstanding something. What is it?

mkoeppe:
The container mounts the current directory (`.`) from the worker. These steps directly write to the worker file system.

kwankyu:
I see. That explains it. 

How about the space problem? Should we see the success of the last run as a sheer luck?

mkoeppe:
It's better to save the space. It's clear that we are operating very close to the edge with our builds on these runners

kwankyu:
I reviewed the last runs for from beta5 to beta8. Only the run beta8 failed by lack of space. 

> It's better to save the space. It's clear that we are operating very close to the edge with our builds on these runners

OK. Thanks.

PR #38161, opened 2024-06-06
fix green function in projective_ds

### LINKED ISSUE #36164: Possible bug in `green_function` that causes `canonical_height` to return incorrect results
We are trying to generalize Wells' algorithm (`canonical_height` function in `projective_ds.py`) to work over number fields with unique factorization in #35967, with modifications and relevant proofs provided by @bhutz.

    File "src/sage/dynamics/arithmetic_dynamics/projective_ds.py", line 2281, in sage.dynamics.arithmetic_dynamics.projective_ds.?.canonical_height
    Failed example:
        f.canonical_height(P(2 + v, 1))
    Expected:
        0.721817737589405
    Got:
        2.5228572277444373139081754230

In the [latest implementation](https://github.com/sagemath/sage/pull/35967/files#diff-acd67704b06c8ddd3ca1d874e4be554f37d4ed507ec591fb943c81a471acdc80), we encountered the test failure above, which we believe might be caused by a bug in `green_function`.

bhutz:
fixes #36164 

There were some errors in how the green function was calculated, especially in the error bound. These are corrected here. The potential error in the canonical height calculation was actually due to insufficient precision in the calculation rather than an error in the code. For example, the following returns the correct value

```
K.<v> = QuadraticField(5)
P1.<x,y> = ProjectiveSpace(K, 1)
f = DynamicalSystem([x^2, y^2])
P=P1(2+v,1)
pr=10000
f.canonical_height(P,prec=pr,error_bound=0.001)
```

### :memo: Checklist

documentation does not need to be updated. The existing tests were updated to reflect this change so accurately test this issue already.


tscrim (changes requested):
Do you have some way to verify that this is now correct (e.g., the outputs are correct or an example where this produced something definitively wrong before)? I don't know the math, but if there is something like this, then I can accept the change.

bhutz:
I could send you the book chapter I wrote working this out in detail, other than that I'm not sure I can find something definitive. Let me look at my files from when I was working on this and see if I can find something that demonstrates correctness.

bhutz:
I looked through my examples and have nothing that is obvious. Going all the way to the canonical height is the way I can think of the test final values via a different method. However, all values remain within the error bounds. Essentially, the canonical height is computed as a sum of the local green functions. The green functions are computed via an iterative process. The error bounds determines the number of iterations that are needed. These are all estimates designed to ensure the final value is within the specified error. The corrected error bounds typically result in no-change or a change of 1 iteration needed for accuracy. However, the final canonical height values in both the previous and correct errors are within the specified error bound of the "true" value.

tscrim:
Okay, thanks for the explanation. If you can resolve the merge conflicts, then we can merge this.

bhutz:
Merge conflict fixed.

tscrim:
Thanks. One last thing is needed to make the linter happy (I don't necessarily agree with it in terms of code clarity, but it is likely a micro-optimization).

bhutz:
ok, fixed the lint error.

tscrim (approved):
Thank you.

PR #38166, opened 2024-06-07
implement morphisms from free algebras

fchapoton:
trying to enhance the morphisms from free monoids and free algebras.

### :memo: Checklist


src/sage/algebras/free_algebra_element.py
```diff
@@ -280,6 +280,30 @@ def _acted_upon_(self, scalar, self_on_left=False):
...
+            sage: A = matrix(QQ, 2, 2, [2, 3, 4, 1])
+            sage: B = matrix(QQ, 2, 2, [1, 7, 7, 1])
+            sage: f = ring.hom([A, B])
+            sage: f(a*b+1)
+            [24 17]
+            [11 30]
+        """
+        n = self.parent().ngens()
+        if n == 0:
+            return codomain.coerce(self)
```
tscrim: I am not sure we should coerce `self` but the underlying coefficient (i.e., we coerce the base ring element)

tscrim (changes requested):
Big +1 on adding this. I probably should have done this a long time ago...

For polynomial rings, my understanding is the general principle is that the output parent should only depend on the parents of the input, not on any specific values. So the test you added would be breaking this. This consistency is useful for code to not have special cases in downstream code.

fchapoton:
there is an existing doctest that make sure one can plug non-square matrices.. That's the reason for the behaviour of the new doctest

tscrim:
I don't think what it does is good behavior. It is far too sensitive to the elements being passed in. Plus, that code is mostly untouched from 2007 (I believe well before we realized this consistency is good). I think we should full disentangle this with providing an example to reproduce the previous behavior but enforce consistency with `__call__` and the like.

fchapoton:
Travis, do you have an opinion on the current state ? It manages to keep the old behaviour and be more consistent, I hope.
Should we somehow require that the values that we substitute belongs to a ring ?

tscrim (approved):
Eventually I think we should make it belong to a common parent that is a ring (or monoid in the free monoid case). However, this is an improvement, and we can revisit dropping the general input later. Let’s get this in for now.

PR #38167, opened 2024-06-07
New methods for sparse graphs backend to enumerate neighbors in linear time

cyrilbouvier:
As noted in issue #37642, enumerating the neighbors of a vertex for a graph that use the sparse backend (the default in Sage) is **not** linear on the number of neighbors (there is an extra log factor).
The goal of this PR is to fix this.

The problem is that to enumerate the neighbors of a vertex, the current code calls iteratively _next_neighbor_unsafe (or one of its out or in variant). But this method return an int corresponding to a neighbor, so to go to the next neighbor, it was first necessary to retrieve this vertex* in the structure storing all the neighbors. As it is stored in a sorted binary tree, the cost of retrieving is O(log(#neighbors)).

*(or a close one if it was deleted in the meantime)

To obtain a linear complexity, I wrote a new method that do a simple tree traversal which is linear in the number of nodes in the tree (i.e., the number of neighbors). As this new method does not have a similar interface as the previous _next_neighbor_unsafe, some more rewriting was necessary to use it in other parts of the code.

1. I wrote new methods for the SparseGraph class: out_neighbors_unsafe and in_neighbors_unsafe. They overwrite the ones from the base class and rely on the new methods _neighbors_unsafe and _neighbors_BTNode_unsafe.
The _neighbors_BTNode_unsafe is a low-level method enumerating the neighbors in linear time.

2. I rewrote the two methods out_neighbors_BTNode_unsafe and in_neighbors_BTNode_unsafe to use the new _neighbors_BTNode_unsafe.

3. I wrote a new method _iterator_edges method for SparseGraphBackend to overwrite the one from the base class
  CGraphBackend, in order to expose the low-level code to the Graph class.
 

Question:
During the writing of this PR, I notice that the methods out_neighbors_BTNode_unsafe and in_neighbors_BTNode_unsafe are only defined for the sparse backend and are not used anywhere in the code.
Can I remove them ? Or should they be kept for compatibility ?

### :memo: Checklist

- [] I have created tests covering the changes.
- [] I have updated the documentation and checked the documentation preview.


cyrilbouvier:
This PR is necessary to fix #37642 but is not sufficient: PR #37662 is also necessary.

src/sage/graphs/base/sparse_graph.pyx
```diff
@@ -631,21 +663,12 @@ cdef class SparseGraph(CGraph):
...
-        p_pointers[0] = pointers
-
-        cdef SparseGraphBTNode* v = self.next_out_neighbor_BTNode_unsafe(u, -1)
-        while v:
-            pointers[num_nbrs] = v
-            num_nbrs += 1
-            v = self.next_out_neighbor_BTNode_unsafe(u, v.vertex)
-
-        return num_nbrs
+        p_pointers[0] = <SparseGraphBTNode **>check_allocarray(degree, sizeof(SparseGraphBTNode *))
```
dcoudert: here you allocate some memory but I think it is never freed. But may be it is done elsewhere...

src/sage/graphs/base/sparse_graph.pyx
```diff
@@ -1446,3 +1502,209 @@ cdef class SparseGraphBackend(CGraphBackend):
...
+                    v = self.vertex_label(v_int)
+                except StopIteration:
+                    return
+
+            # WARNING
+            # If you modify this, you must keep in mind the documentation in the
+            # corresponding method in `generic_graph.py` in the method `edge_iterator`.
+            # E.g. code assumes that you can use an iterator to relabel or delete arcs.
+
+            r = self._cg._neighbors_BTNode_unsafe (v_int, out, neighbors, maxdegree)
```
dcoudert: remove the space before `(`

src/sage/graphs/base/sparse_graph.pyx
```diff
@@ -753,21 +790,12 @@ cdef class SparseGraph(CGraph):
...
-
-        cdef SparseGraphBTNode* u = self.next_in_neighbor_BTNode_unsafe(v, -1)
-        while u:
-            pointers[num_nbrs] = u
-            num_nbrs += 1
-            u = self.next_in_neighbor_BTNode_unsafe(v, u.vertex)
-
-        return num_nbrs
+        p_pointers[0] = <SparseGraphBTNode **>check_allocarray(degree, sizeof(SparseGraphBTNode *))
+        return self._neighbors_BTNode_unsafe (v, 0, p_pointers[0], degree)
```
dcoudert: remove space before `(`

src/sage/graphs/base/sparse_graph.pyx
```diff
@@ -1446,3 +1502,209 @@ cdef class SparseGraphBackend(CGraphBackend):
...
+    OUTPUT: Boolean according the modus:
+
+    - ``modus == 0`` -- ``False``
+    - ``modus == 1`` -- ``True``
+    - ``modus == 2`` -- ``True
+    - ``modus == 3`` -- ``False if v <= u else True``
+    """
+    if modus == 0:
+        return False
+    if modus == 1 or modus == 2:
```
dcoudert: you could reorder the tests to check if `modus == 3` and otherwise return `True`

dcoudert (commented):
> Question:
During the writing of this PR, I notice that the methods out_neighbors_BTNode_unsafe and in_neighbors_BTNode_unsafe are only defined for the sparse backend and are not used anywhere in the code.
Can I remove them ? Or should they be kept for compatibility ?

These are old functions that are apparently no longer used. I think you can remove them.

src/sage/graphs/base/sparse_graph.pyx
cyrilbouvier: done

src/sage/graphs/base/sparse_graph.pyx
cyrilbouvier: The code was in the unused functions that are now deleted.
(For info, the code was written like this because the old functions assumed that an unallocated pointer was passed as argument, that the pointer was allocated inside the function and should be freed after the call.)

cyrilbouvier:
> These are old functions that are apparently no longer used. I think you can remove them.

done in [f04678c](https://github.com/sagemath/sage/pull/38167/commits/f04678cd5928ebd270eefd2779fe4613145b7767)

src/sage/graphs/base/sparse_graph.pyx
cyrilbouvier: Does 51933f0 correspond to what you had in mind ?

dcoudert:
> > These are old functions that are apparently no longer used. I think you can remove them.
> 
> done in [f04678c](https://github.com/sagemath/sage/pull/38167/commits/f04678cd5928ebd270eefd2779fe4613145b7767)

yes.

dcoudert:
something goes wrong with the CI. Before, this PR was inducing a segfault...

cyrilbouvier:
> something goes wrong with the CI. Before, this PR was inducing a segfault...

I fixed it: I wrote two line of code to compute the max degree of the graph but I used `num_verts` instead of the size of `active_vertices` (+ test if the bit is active in the bitset) to iterate over the vertices of the graph. It created out of bound memory access in the rest of the code. My bad. (see [e5f4299](https://github.com/sagemath/sage/pull/38167/commits/e5f4299775d42db611e0441ec13f3bbb1d1e4a58))

There is still some failing tests with labels. I am working on it.

src/sage/graphs/base/sparse_graph.pyx
```diff
@@ -1449,3 +1449,213 @@ cdef class SparseGraphBackend(CGraphBackend):
...
+        cdef int maxdegree = 0
+        cdef int *degrees = NULL
+        if out:
+            degrees = cg.out_degrees
+        else:
+            degrees = cg.in_degrees
+        for v_int in range(cg.active_vertices.size):
+            if bitset_in(cg.active_vertices, v_int):
+                maxdegree = max(degrees[v_int], maxdegree)
+        cdef SparseGraphBTNode **neighbors = <SparseGraphBTNode **>check_allocarray(maxdegree, sizeof(SparseGraphBTNode *))
```
dcoudert: this should be deallocated somewhere, before each return statement.

dcoudert (changes requested):

src/sage/graphs/base/sparse_graph.pyx
cyrilbouvier: Done, using MemoryAllocator in b7e644e

dcoudert (approved):
LGTM.

dcoudert:
This is a nice improvement. Thank you.

PR #38174, opened 2024-06-08
New algorithms to compute the characteristic polynomial of the Frobenius endomorphism of a Drinfeld module

kryzar:
This pull request implements two new algorithms to compute the characteristic polynomial of the Frobenius endomorphism of a Drinfeld $\mathbb F_q[T]$-module over a finite field $K$. Previously, only the algorithms based on crystalline cohomology (see [Musleh-Schost 2023](https://dl.acm.org/doi/10.1145/3597066.3597080)) or on Anderson motives (see [Caruso-Leudière 2023](https://arxiv.org/abs/2307.02879)) were implemented. We propose two new algorithms:

- The algorithm based on central simple algebras described in Chapter 4 of [Caruso-Leudière 2023](https://arxiv.org/abs/2307.02879).

- The algorithm described by Gekeler in [Gekeler 1991](https://www.sciencedirect.com/science/article/pii/002186939190211P).

**Acknowledgement.** This implementation was originally due to @xcaruso (see [here](https://github.com/xcaruso/sage/blob/d2e36bd18b51c93806b7a3b5c8261da7dc98c494/src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py)), and after a private discussion, I took the liberty of creating this PR.

I also propose to change the formula computed by `frobenius_norm`. Before, it computed

$$(-1)^n \mathrm{N}_{K/\mathbb F_q}(\Delta) p^{n / \mathrm{deg}(p)},$$

where $K$ is the ground field, $n$ is the degree of $K$ over $\mathbb F_q$, and $p$ is the monic generator of the $\mathbb F_q[T]$-characteristic of $K$. The docstring claimed this was $(-1)^r$ times the constant coefficient of the characteristic polynomial of the Frobenius endomorphism, $r$ being the rank of the Drinfeld module. I believe this was a mistake, and instead changed the formula, following Theorem 4.2.7 of [Papikian's book](https://link.springer.com/book/10.1007/978-3-031-19707-9), to
$$(-1)^{nr - n - r} \mathrm{N}_{K/\mathbb F_q}(\Delta) p^{n / \mathrm{deg}(p)}.$$

P.S. @davidayotte, given that you now work in the industry, please tell us if you still want to be tagged on these Drinfeld module stuff. Same question for you @ymusleh given that you defended your thesis (congratulations!).

### :memo: Checklist

(Dédicace à ABLCCN)


src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -329,33 +354,143 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+            sage: Fq = GF(2)
+            sage: A.<T> = Fq[]
+            sage: K.<z> = Fq.extension(2)
+            sage: phi = DrinfeldModule(A, [z, 0, 1])
+            sage: phi.frobenius_charpoly()
+            X^2 + X + T^2 + T + 1
+            sage: phi.frobenius_charpoly(var='Foo')
+            Foo^2 + Foo + T^2 + T + 1
+            sage: phi.frobenius_charpoly(var='Bar')
+            Bar^2 + Bar + T^2 + T + 1
```
DavidAyotte, line 427: I feel like there a little bit too much emphasis on this optional flag and the documentation of this method is already quite lenghty. I think this could either: be rephrased into a more "software reference manual documentation" way OR put into a `TEST` block in order to hide it from the manual completely.

What I mean by the first suggestion is: rephrase the sentence like "Use the optional ``var`` parameter in order to change the variable name of the resulting univariate polynomial" or somthing like this. Also, I would reuse the last `phi` to make it a one-liner doctest.

Anyway, this is only my preference as someone who don't really like to read documentation and prefer to have the most straightforward info.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -329,33 +354,143 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+            else:
+                algorithm = 'CSA'
         method_name = f'_frobenius_charpoly_{algorithm}'
         if hasattr(self, method_name):
             if self._frobenius_charpoly is not None:
-                return self._frobenius_charpoly
+                return self._frobenius_charpoly.change_variable_name(var)
             self._frobenius_charpoly = getattr(self, method_name)(var)
             return self._frobenius_charpoly
         raise NotImplementedError(f'algorithm \"{algorithm}\" not implemented')
```
DavidAyotte: I'm not sure, but I don't think the backslash are needed here (line 433).

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -329,33 +354,143 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+            phi_T = t * phi_T
+        chi = Matrix(rows).charpoly()
+        # We format the result
+        K = self.base_over_constants_field()
+        A = self.function_ring()
+        r = self.rank()
+        lc = chi[0][r]
+        coeffs = [ A([K(chi[i][j]/lc).in_base()
+                      for i in range((r-j)*n // r + 1)])
+                   for j in range(r+1) ]
```
DavidAyotte:
```suggestion
        coeffs = [A([K(chi[i][j]/lc).in_base()
                     for i in range((r-j)*n // r + 1)])
                  for j in range(r+1)]
```
Did I just correct your pep8?!

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: But this block (lines 408 to 420) are already in a `TESTS::` block, no? See line 359.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Thank you, I think you're right.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Yes 😭
(But it was Xavier's code...)

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -329,33 +354,143 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+        if algorithm is None:
+            if self.rank() < self._base_degree_over_constants:
+                algorithm = 'crystalline'
+            else:
+                algorithm = 'CSA'
         method_name = f'_frobenius_charpoly_{algorithm}'
         if hasattr(self, method_name):
             if self._frobenius_charpoly is not None:
-                return self._frobenius_charpoly
+                return self._frobenius_charpoly.change_variable_name(var)
```
xcaruso: With these lines, you use the cached value even if the user explicitly asks for another algorithm. I think that it is not what we want.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -512,11 +760,15 @@ def frobenius_norm(self):
...
+        p = self.characteristic()
         norm = K(self.coefficients()[-1]).norm()
-        self._frobenius_norm = ((-1)**n)*(char**(n/char.degree())) / norm
+        self._frobenius_norm = (-1) ** (n*r - n - r) \
+                               * norm**(-1) \
+                               * p ** (n//p.degree())
         return self._frobenius_norm
 
+    @cached_method
     def frobenius_trace(self):
```
xcaruso: In this method, you use the crystalline matrix in all cases. I think it is okay but aren't there any case where the motive matrix is faster to compute?

xcaruso (commented):
Two small comments.
Otherwise it looks good to me; however, my opinion is certainly biased since I am mostly the author of this code.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Yup. Will take care of that.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Agreed. Will also take care of it.
(Wasn't it already the case tho? I remember having to use some tricks in the past to not use the cached data, but maybe I was mistaken.)

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Any objection to making `DrinfeldModuleMorphism._motive_matrix` public? It's interesting in its own right and would be useful in our case.

xcaruso: What do you mean by public? Without the initial underscore?

kryzar: Yup.

xcaruso: Well, eventually, I plan to have a method `motive` to create the motive of a Drinfeld module or a morphism between those, and a method `matrix` to get the matrix of a morphism of Anderson motives.
So, if you make the method `_motive_matrix` public now, it means that it will need to be deprecated afterwards. Therefore, I prefer keeping it as it is.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -228,6 +230,30 @@ def companion(order):
...
+        is the characteristic polynomial of the Frobenius endomorphism,
+        which is crucial for :meth:`_frobenius_charpoly_CSA`.
+        """
+        E = self._base
+        EZ = PolynomialRing(E, name='Z')
+        n = self._base_degree_over_constants
+        phi_T = self.gen()
+        t = self.ore_variable()
+        rows = []
+        # Compute the reduced charpoly
```
xcaruso: Not sure that this comment is relevant.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -228,6 +230,30 @@ def companion(order):
...
+        respect to the maximal commutative subalgebra `K\{\tau^n\}` of
+        $K\{\tau\}$, where `K` is the base field and `n` is its degree
+        over `\mathbb F_q`. The characteristic polynomial of this matrix
+        is the characteristic polynomial of the Frobenius endomorphism,
+        which is crucial for :meth:`_frobenius_charpoly_CSA`.
+        """
+        E = self._base
+        EZ = PolynomialRing(E, name='Z')
+        n = self._base_degree_over_constants
+        phi_T = self.gen()
```
xcaruso: I propose to replace `phi_T` by `f` because it will then be updated in the loop below.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -291,9 +317,29 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+              on the crystalline cohomology (see [MS2023]_).
+            - ``'gekeler'`` -- it tries to identify coefficients by
+              writing that the characteristic polynomial annihilates the
+              Frobenius endomorphism; this algorithm may fail is some
+              cases (see [Gek2008]_).
+            - ``'motive'`` -- it uses the action of the Frobenius on the
+              Anderson motive (see Chapter 2 of [CL2023]_).
+
+        The method raises an exception if the user asks for an
+        unimplemented algorithm, even if the characteristic has already
```
xcaruso: the characteristic?

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -329,33 +378,147 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+                    algorithm = 'crystalline'
+                else:
+                    algorithm = 'CSA'
+        # If an algorithm is specified, do not use cached data, even
+        # if it is possible:
+        method_name = f'_frobenius_charpoly_{algorithm}'
+        if not hasattr(self, method_name):
+            raise NotImplementedError(f'algorithm "{algorithm}" not implemented')
+        self._frobenius_charpoly = getattr(self, method_name)(var)
+        return self._frobenius_charpoly
```
xcaruso:
```suggestion
        self._frobenius_charpoly = getattr(self, method_name)
        return self._frobenius_charpoly.change_variable_name(var)
```
To avoid problem if `var` is not a variable name.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -329,33 +378,147 @@ def frobenius_charpoly(self, var='X', algorithm='crystalline'):
...
+
+    @cached_method
+    def _frobenius_charpoly_CSA(self, var='X'):
```
xcaruso: I'm not sure we really need `var` here (and in the other `_frobenius_charpoly_*` methods. But, I'm fine with keeping it if you prefer.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
```diff
@@ -512,12 +787,16 @@ def frobenius_norm(self):
...
+        p = self.characteristic()
         norm = K(self.coefficients()[-1]).norm()
-        self._frobenius_norm = ((-1)**n)*(char**(n/char.degree())) / norm
+        self._frobenius_norm = (-1) ** (n*r - n - r) \
+                               * norm**(-1) \
+                               * p ** (n//p.degree())
         return self._frobenius_norm
 
-    def frobenius_trace(self):
+    @cached_method
```
xcaruso: Maybe caching in not needed here, since it's reimplemented in the method.

kryzar:
(Yeah, I have plenty of bugs. This was not ready for review. For now I have troubles with the machine I compile Sage on, so I pushed a first, bugged, version to have the results of the CI.)

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Fixed :+1:

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Done :+1:

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Comment removed.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: You're right. I removed them, it's cleaner and it was useless. Code should not be cluttered.

src/sage/rings/function_field/drinfeld_modules/finite_drinfeld_module.py
kryzar: Yep. I removed that on `frobenius_trace` and `frobenius_norm`.

xcaruso:
I merged with the latest release of sage and fixed the computation of the trace of the Frobenius.

xcaruso:
I don't understand anything to the automatic checks but I think that the ticket is ready for review. @kryzar, you can remove the draft tag if you agree.

xcaruso:
[Here](https://github.com/sagemath/sage/actions/runs/10339320986/job/28618521184?pr=38174), all tests passed... [Here](https://github.com/sagemath/sage/actions/runs/10339320986/job/28618521290?pr=38174), there were several failing doctests, but they don't seem to be related to this ticket.
All in all, it's quite confusing! I nevertheless give a positive review.

vbraun:
Documentation build fails with the same error as in the CI; please don't set to positive review if the CI fails.

xcaruso:
Strong +1.
Honestly, I'm really very confused (and unhappy) by all the CI. There are tons of them and many of them basically randomly fail for some weird reason I generally can't understand. I asked several times for clarification but very rarely got an answer.

xcaruso:
This time, for instance, I got the error for the check `Build & Test / test-new (pull_request)`:
```
Run docker run --name BUILD -dit \
  docker run --name BUILD -dit \
             --mount type=bind,src=$(pwd),dst=$(pwd) \
             --workdir $(pwd) \
             localhost:5000/sagemath/sage/sage-ubuntu-jammy-standard-with-targets:ci /bin/sh
  shell: /usr/bin/bash -e {0}
  env:
    TOX_ENV: docker-ubuntu-jammy-standard-incremental
    BUILD_IMAGE: localhost:5000/sagemath/sage/sage-ubuntu-jammy-standard-with-targets:ci
    FROM_DOCKER_REPOSITORY: ghcr.io/sagemath/sage/
    FROM_DOCKER_TARGET: with-targets
    FROM_DOCKER_TAG: dev
    EXTRA_CONFIGURE_ARGS: --enable-fat-binary
Unable to find image 'localhost:5000/sagemath/sage/sage-ubuntu-jammy-standard-with-targets:ci' locally
docker: Error response from daemon: manifest for localhost:5000/sagemath/sage/sage-ubuntu-jammy-standard-with-targets:ci not found: manifest unknown: manifest unknown.
See 'docker run --help'.
Error: Process completed with exit code 125.
```
I assume this is not due to this PR.

With `Build & Test using Conda / Conda (macos-latest, 3.11, environment) (pull_request)`, I got:
```
sage -t --random-seed=91866416175107778193950271457384847877 src/sage/modules/free_quadratic_module_integer_symmetric.py  # 1 doctest failed
sage -t --random-seed=91866416175107778193950271457384847877 src/sage/rings/number_field/galois_group.py  # 1 doctest failed [failed in baseline: unreported failure on macOS (sort order) in GaloisGroup_v2.artin_symbol, https://github.com/sagemath/sage/actions/runs/9525536510/job/26259809272?pr=37998]
sage -t --random-seed=91866416175107778193950271457384847877 src/sage/rings/number_field/number_field.py  # 2 doctests failed [failed in baseline: unreported failure on macOS (sort order) in NumberField_absolute.[optimized_]subfields, https://github.com/sagemath/sage/actions/runs/9525536510/job/26259809272?pr=37998]
sage -t --random-seed=91866416175107778193950271457384847877 src/sage/rings/qqbar.py  # 1 doctest failed [failed in baseline: unreported failure on macOS seen in https://github.com/sagemath/sage/actions/runs/9525536510/job/26259809272?pr=37998]
```
Should I care about this?

All `CI Linux incremental*` basically also fail. Is it important?

In any case, it seems that the issue with the documentation is now fixed. So I give again a positive review. I'm sorry if I'm wrong but I'm completely lost.

DavidAyotte:
Why is the git commit history so ugly? It's quite hard to follow what has been done in this PR.

xcaruso:
I think it's my fault but I don't know exactly what happened: at some point, I merged with the latest version of `develop` and got many commits from there is the history. It seems that github believes that PR is based on version 10.5.beta1, whereas it is based on version 10.5.beta1 (see the diff).

tobiasdiez:
The error in the conda workflow
sage -t --random-seed=91866416175107778193950271457384847877 src/sage/modules/free_quadratic_module_integer_symmetric.py  # 1 doctest failed

at least seems to be introduced by this PR or not? Could this be fixed please? Thanks! (Maybe as a follow-up since its already in volkers release branch)

kryzar:
> The error in the conda workflow sage -t --random-seed=91866416175107778193950271457384847877 src/sage/modules/free_quadratic_module_integer_symmetric.py # 1 doctest failed
> 
> at least seems to be introduced by this PR or not? Could this be fixed please? Thanks! (Maybe as a follow-up since its already in volkers release branch)

This bug is not even in the current version; I merged `develop` yesterday:

```
sage -t --random-seed=74903126027750647063226099060708583867 src/sage/modules/free_quadratic_module_integer_symmetric.py
```

So no, I don't think so, and I can hardly find it realistic that Xavier or me would have touched this file.

xcaruso:
Yo! All required tests pass! :tada:

kryzar:
[gif](https://i.giphy.com/media/v1.Y2lkPTc5MGI3NjExcWVuZjRld2U5ODJqMnZsczJtZ3MxOW93NnBxOWEzOWkwbXo3aDd1ayZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9dg/G96zgIcQn1L2xpmdxi/giphy.gif)

kryzar:
Ah bah non attends.

(Meaning: oh but wait but no.)

Some stuff related to coverage is failing. Oh well.

xcaruso:
The test is not required :-).
And more importantly, the error message is
```
...
FileExistsError: [Errno 17] File exists: '.coverage'
```
So the most likely is that there is a problem somewhere in the configuration of the container, but not in this PR.

kryzar:
Not all tests are required?!

xcaruso:
I'm not sure. But some of them are marked `Required` and others are not.

kwankyu:
coverage ci will be fixed by #38687. You may ignore the ci failure here.

kryzar:
> coverage ci will be fixed by #38687. You may ignore the ci failure here.

Many thanks for the heads-up!

xcaruso:
:champagne:

kryzar:
> 🍾

+100

PR #38176, opened 2024-06-09
fix and activate E305 in pyx files

fchapoton:
this is adding the pycodestyle E305 check to pycodestyle-minimal in pyx files

### :memo: Checklist

### DIFF
--- a/src/tox.ini
+++ b/src/tox.ini
@@ -183,7 +183,7 @@ description =
     # See https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
 deps = pycodestyle
 commands = pycodestyle --select E111,E21,E221,E222,E225,E227,E228,E25,E271,E303,E305,E306,E401,E502,E701,E702,E703,E71,E72,W291,W293,W391,W605 {posargs:{toxinidir}/sage/}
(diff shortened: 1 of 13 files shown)


src/tox.ini
```diff
@@ -183,7 +183,7 @@ description =
     # See https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
 deps = pycodestyle
 commands = pycodestyle --select E111,E21,E221,E222,E225,E227,E228,E25,E271,E303,E305,E306,E401,E502,E701,E702,E703,E71,E72,W291,W293,W391,W605 {posargs:{toxinidir}/sage/}
-       pycodestyle --select E111,E271,E301,E306,E401,E502,E703,E712,E713,E714,E72,W29,W391,W605, --filename *.pyx {posargs:{toxinidir}/sage/}
+       pycodestyle --select E111,E271,E301,E305,E306,E401,E502,E703,E712,E713,E714,E72,W29,W391,W605, --filename *.pyx {posargs:{toxinidir}/sage/}
```
mkoeppe, line 184: Also update SAGE_ROOT/.vscode/settings.json please

fchapoton: it seems that there is nothing in there to be modified. No idea if this has been moved or removed

mkoeppe: You are right; pycodestyle has been replaced by ruff there. 
Then let's please remove the comment in the file that instructs people to update settings.json ....

mkoeppe (approved):
Thanks, LGTM

PR #37858, opened 2024-04-24
`sage.features.topcom`

mkoeppe:
- Part of #35856

We also change how the SPKG index in the reference manual is generated.
`index.rst` is now a static file that uses `.. include` directives to read in generated files `index_standard.rst`, `index_optional.rst`, etc.

### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38185, opened 2024-06-10
Simplify doc build process by removing empty directories locally

kwankyu:
Now doc builder removes empty directories only in its own source directory.

### :memo: Checklist

### DIFF
--- a/src/doc/Makefile
+++ b/src/doc/Makefile
@@ -42,7 +42,7 @@ doc-inventory-reference: doc-src
 	$(eval BIBLIO = $(firstword $(DOCS)))
 	$(eval OTHER_DOCS = $(wordlist 2, 100, $(DOCS)))
 	$(MAKE) doc-inventory--$(subst /,-,$(BIBLIO))
-	$(MAKE) SAGE_DOCBUILD_OPTS="$(SAGE_DOCBUILD_OPTS) --no-prune-empty-dirs" $(foreach doc, $(OTHER_DOCS), doc-inventory--$(subst /,-,$(doc)))
+	$(MAKE) $(foreach doc, $(OTHER_DOCS), doc-inventory--$(subst /,-,$(doc)))
 	$(MAKE) SAGE_DOCBUILD_OPTS="$(SAGE_DOCBUILD_OPTS) --no-prune-empty-dirs" doc-inventory--reference_top
 endif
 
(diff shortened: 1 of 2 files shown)


vbraun:
Bootstrap download of the confball does not work (note the missing version):
```
$ ./bootstrap -u http://sagepad.org/spkg/upstream/configure
[...]
Error: downloading configure-.tar.gz from http://sagepad.org/spkg/upstream/configure failed
```

vbraun:
Needs work because it has #37858 merged in

kwankyu:
This PR is now separate from #37858. Positive review?

mkoeppe:
Has the code that removes empty directories been merged already?

kwankyu:
Before this PR, empty directories in all docs (each of them is a sphinx project) are removed at the same time when references (BIBLIO in src/doc/Makefile, one particular sphinx project) is made. This approach works but violates , well, the principle of separation of concerns(?).

With this PR, empty directories in each doc are removed when that doc is made. So one sphinx project does not touch empty directories in governance of another sphinx project. This approach also works and looks better.

mkoeppe:
Right, that's what I understood from the PR description, but in the diff I only see that `--no-prune-empty-dirs` is no longer used.

kwankyu:
In the phase of building inventory, `--no-prune-empty-dirs` is not used, so that empty directories are removed. In the phase of building html files, `--no-prune-empty-dirs` is used (as before) so that no time is wasted to remove empty directories.

src/doc/Makefile
```diff
@@ -63,7 +63,7 @@ doc-html-reference: doc-html-reference-sub
 doc-html-other: doc-html-reference
 	$(eval DOCS = $(shell sage --docbuild --all-documents all))
 	@if [ -z "$(DOCS)" ]; then echo "Error: 'sage --docbuild --all-documents' failed"; exit 1; fi
-	$(MAKE) SAGE_DOCBUILD_OPTS="$(SAGE_DOCBUILD_OPTS) --no-prune-empty-dirs" $(foreach doc, $(wordlist 2, 100, $(DOCS)), doc-html--$(subst /,-,$(doc)))
+	$(MAKE) $(foreach doc, $(wordlist 2, 100, $(DOCS)), doc-html--$(subst /,-,$(doc)))
```
mkoeppe, line 66: but this change here is for the "HTML" phase?

kwankyu: Above this part, it is for reference manual (that has inventory phase).

Here, it is for all other docs different from reference manual (that have only html phase).

kwankyu: Above this part, it is for sub docs of the reference manual (that has inventory phase).

mkoeppe: OK, thanks for the explanation.

kwankyu:
Thanks!

PR #38189, opened 2024-06-10
Improve DiGraph neighbor iterators

cyrilbouvier:
Currently, in the Digraph class, the two methods `neighbor_out_iterator` and `neighbor_in_iterator` are just wrappers around low-level methods `_backend.iterator_out_nbrs` and `_backend.iterator_in_nbrs`. To build the iterator, the low-level iterator is first converted into a set and then converted back into an iterator using `iter()`. It does not seem efficient.

This PR proposes to use `yield from` instead (see the very short commit 9a89251bc70c28ba)

As this modification changes the order of the output, the output of 4 tests need to be reorder (see commit 305b933ce6). One can check that the lists contains the same elements, but not in the same order.

### :memo: Checklist


grhkm21 (approved):
Looks good to me.

Was the previous version deterministic at all? I thought `set` has random order

cyrilbouvier:
> Was the previous version deterministic at all? I thought `set` has random order

I think `set` is deterministic (like dict): if you do `list(set([<input list>]))`, you always get the same order, just not the same as the input list.

dcoudert:
I think the `set(...)` was here to prevent reporting twice a same vertex, and this has been implemented long before Cython supports `yield` statements. It seems that the proposed change works well. However, could you double check that it's ok with all backends, and with digraphs with loops / multiple edges.

cyrilbouvier:
> I think the `set(...)` was here to prevent reporting twice a same vertex, and this has been implemented long before Cython supports `yield` statements. It seems that the proposed change works well. However, could you double check that it's ok with all backends, and with digraphs with loops / multiple edges.

I added tests to check that in the case of multiple edges, the neighbors are listed only once.

I also added tests to check that the closed neighborhood is listed in the presence of loop(s).

dcoudert (approved):
I did some tests with the current PR and this looks good to me. Thanks.

PR #38206, opened 2024-06-12
Lint: Install tox from ubuntu, as done in CI Build&Test etc.

mkoeppe:
... instead of via pip.
This is for uniformity and motivated by stability concerns prompted by #38205 

### :memo: Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,14 +28,12 @@ jobs:
         GH_TOKEN: ${{ github.token }}
         SAGE_CI_FIXES_FROM_REPOSITORIES: ${{ vars.SAGE_CI_FIXES_FROM_REPOSITORIES }}
 
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: 3.9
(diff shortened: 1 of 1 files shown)


kwankyu:
The situation is that pip-installed `tox` fails in auto-provisioning (upgrading) itself because of incompatible version of `filelock`, and the distribution-provided `tox` is supposed to be more stable from this kind of instability. Am I right?

kwankyu:
Is there a test run?

mkoeppe:
> The situation is that pip-installed `tox` fails in auto-provisioning (upgrading) itself because of incompatible version of `filelock`, and the distribution-provided `tox` is supposed to be more stable and immune from this kind of instability. Am I right?

Yes, that's right.

mkoeppe:
A test run (when there was still a broken filelock on PyPI): https://github.com/sagemath/sage/actions/runs/9488101498/job/26146233900

kwankyu (approved):
OK. LGTM.

mkoeppe:
Thanks!

PR #37768, opened 2024-04-08
Fix category of point homsets of EllipticCurves and Jacobians

grhkm21:
Fixes #37730.

There is a change to categories and homsets:
```sage
sage: Hom(ZZ, ZZ).category() # inherits from CommAdditiveGroups() now
Join of Category of endsets and Category of commutative additive groups and Category of homsets of Dedekind domains and euclidean domains and metric spaces
sage: # this used to be "Category of homsets of sets"
sage: class X(Parent):
....:     def _repr_(self): return "XXXXXX"
....: Hom(X(category=Sets()), ZZ).category()
Join of Category of commutative additive groups and Category of homsets of sets
```

More TestSuite passes now:
```sage
sage: for R in [Zmod(10), QQ, NumberField(x^3 + 5, "a"), Qp(3), GF(2), GF(3), GF((7, 3))]:
....:     params = [3, 5] if R.characteristic() not in [2, 3] else [1, 0, 1, 3, 5]
....:     TestSuite(EllipticCurve(R, params)).run(skip="_test_not_implemented_methods")
```
(This used to fail quite spectacularly)

There are also many changes in algebraic geometry, such as:
```sage
sage: J = HyperellipticCurve(polygen(QQ)^5 + 1).jacobian()
sage: J.categories() # jacobians are abelian varieties
[Category of abelian varieties over Spectrum of Rational Field,
 Category of schemes over Rational Field,
 Category of schemes,
 Category of sets,
 Category of sets with partial maps,
 Category of objects]
sage: J.point_homset().categories()
[Join of Category of commutative additive groups and Category of homsets of schemes over Rational Field,
 Category of commutative additive groups, ...
 Category of homsets of schemes over Rational Field, ...
 Category of sets with partial maps,
 Category of objects]

sage: from sage.schemes.jacobians.abstract_jacobian import Jacobian_generic
....: P2.<x, y> = ProjectiveSpace(QQ, 1)
....: J = Jacobian_generic(Curve(y^2 - (x^3 + x)))
....: J in AbelianVarieties(QQ) # works for Jacobian_generic
True
sage: J = Jacobian_generic(Curve(y^2 - x^3))
....: J in AbelianVarieties(QQ) # only if the curve is projective smooth
False

sage: AbelianVarieties(QQ) # changed base scheme to be a scheme (lol?)
Category of abelian varieties over Spectrum of Rational Field
sage: AbelianVarieties(QQ) is AbelianVarieties(Spec(QQ)) # input is normalised
True
```

[...]

PR #38207, opened 2024-06-13
Rename PolynomialRing_general to PolynomialRing_generic

grhkm21:
Class names should be *_generic, and indeed is the case for at least 140 classes, but PolynomialRing is incorrectly suffixed with _general. This PR fixes this problem.

### :memo: Checklist


grhkm21:
A note for future me and others: Most LSPs support renaming a symbol across an entire project. I use coc.nvim so just use coc-rename to all the names across Sage. (It doesn't detect comments though :( )

mkoeppe:
In a somewhat different direction, we may want to look into merging constructor functions such as `PolynomialRing` and the "generic" classes via __classcall__.

tscrim:
Normally I would completely agree with @mkoeppe about having the `PolynomialRing` constructor go through an ABC. However, in this case, I am quite apprehensive about it as this constructor is fairly complex, as well as currently the univariate and multivariate classes that are essentially separate:
```
sage: R.<x,y> = QQ[]
sage: R.__class__.__mro__
(<class 'sage.rings.polynomial.multi_polynomial_libsingular.MPolynomialRing_libsingular'>,
 <class 'sage.rings.polynomial.multi_polynomial_ring_base.MPolynomialRing_base'>,
 <class 'sage.rings.ring.CommutativeRing'>,
 <class 'sage.rings.ring.Ring'>,
 <class 'sage.structure.parent_gens.ParentWithGens'>,
 <class 'sage.structure.parent_base.ParentWithBase'>,
 <class 'sage.structure.parent_old.Parent'>,
 <class 'sage.structure.parent.Parent'>,
 <class 'sage.structure.category_object.CategoryObject'>,
 <class 'sage.structure.sage_object.SageObject'>,
 <class 'object'>)
```
(which doesn't even use the category framework properly too!).

@grhkm21 Just do a `grep -r SAGE_ROOT/src/sage/*` (possibly set to ignore generated `.c` files) :)

grhkm21:
> @grhkm21 Just do a `grep -r SAGE_ROOT/src/sage/*` (possibly set to ignore generated `.c` files) :)

Yeah that *finds* the appearance of the class, but the LSP method I mentioned renames the class + all references to it (including imports etc). Plus if there's hundreds of references, say I want to rename `Ring` to `Rang` or something, then grep will be very inaccurate, since e.g. `PolynomialRing` will include the word `Ring`. But LSP (*Language* + *Syntax* Protocol) has the context to deduce which Ring is the Ring I'm trying to rename :D

grhkm21:
Re category: yes, a lot of things are not in the category framework. I plan to work on that whenever I have time during the summer. e.g. by replacing some `isinstance(x, some ring base)` or `is_ring(x)` by `x in Rings()` (locally, not PRing them), then fixing doctests. That worked once already (#37768) so it should be a good approach. 

Re constructor/classcall thing: yes in this case I don't think it's feasible.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1776,7 +1777,38 @@ def monics( self, of_degree=None, max_degree=None ):
...
+        Check they provide the same functionalities for backward compatability::
+
+            sage: dir(PolynomialRing_general(QQ, name="a")) == dir(PolynomialRing_generic(QQ, name="a"))
+            True
+        """
+        deprecation(
+            38207, "The class PolynomialRing_general has been renamed to PolynomialRing_generic."
+        )
+        super().__init__(*args, **kwds)
+
```
kwankyu, line 1800: Do we need this deprecation? `PolynomialRing_general` is an internal class.

tscrim: Technically yes because it is in the public documentation. However, this is such a core class that it is quite probable that code in the wild is using this in some form. In either case, it doesn’t really hurt to have this deprecation.

grhkm21: I don't know, I am not familiar with deprecation of *classes*. Should there still be some kind of deprecation message for it? Or just let people git grep it? 😆

kwankyu: Deprecated code needs another PR to remove it later. It is nontrivial maintenance burden. I don't think it is a good idea to add deprecation in sage based on "it doesn't really hurt".

This is just one of common cases, for which we should apply our principles. What is the principle of determining a class is "internal" or "public"? Is it public (1) if it appears in our reference manual or (2) if it is accessible (or rather used) in repl without explicitly importing the class? I was thinking it is (2), but you are saying it is (1). I am not sure. What others think?

kwankyu: > I don't know, I am not familiar with deprecation of _classes_. 

I thought I knew. But I am not sure after hearing Travis :-) Let us hear Matthias too.

tscrim: Removing code is really basically nothing, and it isn’t something that needs to be done. So it is an epsilon maintenance burden at worst.

Strictly speaking, it is (1), but we don’t have to do that in all cases. In this case, we really should follow it.

kwankyu: By the way if it is (1), we are perhaps destined to add lots of deprecations whenever we refactor classes...

kwankyu: This question is separate from reviewing this PR.

I am positive to this PR, anyway.

kwankyu (approved):

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -1776,7 +1777,38 @@ def monics( self, of_degree=None, max_degree=None ):
         raise ValueError("you should pass exactly one of of_degree and max_degree")
 
 
-class PolynomialRing_commutative(PolynomialRing_general):
+# Placeholder class for deprecation
+class PolynomialRing_general(PolynomialRing_generic):
```
mkoeppe: This idea of creating a deprecated subclass for the old name does not work because it breaks user code that uses `isinstance(...,  PolynomialRing_general)`.

grhkm21: Please suggest a solution as well, as I don't know one. I can remove it, since kwankyu is positive on that as well :)

mkoeppe: For `isinstance` testing in user code, `lazy_import(..., deprecation=...)` works.
For subclassing in user code, the current solution works.

There isn't a solution that works both for `isinstance` testing and for subclassing.

So why not just make it an alias and write a comment.
```
PolynomialRing_general = PolynomialRing_generic  # deprecated alias, #38207
```

grhkm21: If it's just a comment then no users will see it, and if it's deprecated (meaning it will be removed in the future) then might as well remove it now.

Also I don't understand your `isinstance` code that "breaks". This works for me:
```sage
sage: from sage.rings.polynomial.polynomial_ring import PolynomialRing_generic, PolynomialRing_general
sage: R = PolynomialRing_general(QQ, name="a")
<ipython-input-3-1ad511612f9e>:1: DeprecationWarning: The class PolynomialRing_general has been renamed to PolynomialRing_generic.
See https://github.com/sagemath/sage/issues/38207 for details.
  R = PolynomialRing_general(QQ, name="a")
sage: isinstance(R, PolynomialRing_general)
True
sage: isinstance(R, PolynomialRing_generic)
True
```

mkoeppe: > If it's just a comment then no users will see it,

Except those that read the release notes

mkoeppe: What breaks is this:
```
sage: R = QQ['a']
sage: isinstance(R, PolynomialRing_general)
```

src/sage/rings/polynomial/polynomial_ring.py
kwankyu: The doc https://deploy-livedoc--sagemath.netlify.app/html/en/developer/coding_in_python.html#deprecation seems to have (2) in mind. I think our deprecation policy was designed to guide users, not the developers...

src/sage/rings/polynomial/polynomial_ring.py
tscrim: > > If it's just a comment then no users will see it,
> 
> Except those that read the release notes

I would not expect many people to read the release notes, especially for such a technical detail.

I think is is okay for subclassing to break provided the deprecation warning is emitted. We might be forced to accept a less-than-perfect solution here, but as long as it indicates what a user should do to change their code, then it is acceptable to me.

src/sage/rings/polynomial/polynomial_ring.py
tscrim: I don't see that interpretation at all. It only says "a backwards-incompatible change" with no reference to "internal" (or any synonym). The only time the global namespace is explicitly mentioned is for one example; note there is an "any other case" and even though it says nothing explicitly about classes, that is clearly implied. If we interpret it very strictly, then even hidden functions/methods/etc. are subject to deprecation. However, I would say for that if a user knows about such methods, then such a user can handle when such a change occurs. Moreover, users can easily see the online documentation, most will understand how to do imports, and they often subclass objects in their own code.

kwankyu: Let's not read the developer guide as like legal statements. 

Deprecation policy is intended to guide users when they encounter changes while they use sage. First I thought only of using sage through user interfaces. But I now agree that we should include the cases when they code using the sage library. If a class could be used by user code and there is change in how to use it, then yes we should add deprecation warning somehow. I thought `PolynomialRing_general` is not such a class unlike `PolynomialRing`. But perhaps I am wrong in this. 

hidden functions/methods/classes are by definition not to used outside the sage library. Deprecation policy is not for those using them.

mkoeppe: I think that classes with an underscore in them may have been intended to be "internal" (after all, this naming scheme violates Python naming conventions).
But in the big `is_ClassName` deprecation, we just started to ask users to switch to using `isinstance(..., ClassName_generic)` in many cases. So that's now definitely public API

kwankyu: Is someone who is writing `isinstance(..., ClassName_generic)` a sage user? Or is he a developer of the sage library?

mkoeppe: Could be either

kwankyu: By the way, there is a discussion on similar topic: https://github.com/sympy/sympy/issues/23037. But I cannot find a good answer there either.

kwankyu: > I think that classes with an underscore in them may have been intended to be "internal" (after all, this naming scheme violates Python naming conventions). 

This is a good idea for a criterion of "public" (being in API) or "internal" (being not part of API).

(C) A  name living in `sage.` is "internal" if it has an underscore in it, but "public" otherwise.

This is consistent with the policy that an attribute starting with an underscore is considered internal.

According to (C), `PolynomialRing_generic` is internal :-)

tscrim: > > I think that classes with an underscore in them may have been intended to be "internal" (after all, this naming scheme violates Python naming conventions).
> 
> This is a good idea for a criterion of "public" (being in API) or "internal" (being not part of API).
> 
> (C) A name living in `sage.` is "internal" if it has an underscore in it, but "public" otherwise.

Furthermore, all functions/methods with 2 words are therefore internal, such as `change_ring()`.

> This is consistent with the policy that an attribute starting with an underscore is considered internal.
>
> According to (C), `PolynomialRing_generic` is internal :-)

Of course, starting with an underscore was what was meant. :)

mkoeppe: No.

mkoeppe: We are talking about **class** names with an underscore anywhere in their name.

kwankyu: Sorry. As we have already rules for functions/methods/attributes, the updated rule is

(C) A class name living in `sage.` is "internal" if it has an underscore anywhere in it, but "public" otherwise

Good enough?

grhkm21: I am not entirely following the conversation, so maybe when you three reach a conclusion (and maybe put it somewhere in the developer guide) I can modify this PR accordingly. I'm fine with any solution, I came up with this placeholder class knowing it's ad-hoc-y

grhkm21: (And ping me for it :) )

kwankyu: The general solution being discussed in this thread is somewhat separate from reviewing this PR.

Whether or not we reach a conclusion for the general problem here, I am willing to continue this discussion in sage-devel. If the community reach to a solution (either by consensus or by voting), yes we should put it to the developer guide.

You, as the author, do not have to wait for that, but just resolve the work issue below according to the reviewer's request (which may change by the discussion here, of course).

mkoeppe:
> In a somewhat different direction, we may want to look into merging constructor functions such as `PolynomialRing` and the "generic" classes via **classcall**.

I've sketched this in #38228 (for a different class)

src/sage/rings/polynomial/polynomial_ring.py
grhkm21: > You, as the author, do not have to wait for that, but just resolve the work issue below according to the reviewer's request (which may change by the discussion here, of course).

Well, Matthias' review/suggestion would not be necessary if the conclusion in this discussion is that this class is treated as internal, since it then doesn't have to go through the deprecation process. Plus, as I replied to Matthias below, I don't agree with his proposed change, and he hasn't replied to me or Travis yet.

src/sage/rings/polynomial/polynomial_ring.py
mkoeppe: > I think is is okay for subclassing to break provided the deprecation warning is emitted.

No. By definition, during the deprecation phase nothing is broken, users just get the warning. 
Deprecation warnings warn about future breakage, not about current breakage.

mkoeppe:
In light of the discussion in 
- #38211 

... we should probably also discuss a possible direction to defining suitable abstract base classes without underscores in names.

One data point: For the elements, we have introduced an ABC `CommutativePolynomial`. (

Should we define an ABC `CommutativePolynomialRing`? (But would InfinitePolynomialRing be a subclass?)

grhkm21:
@tscrim @kwankyu Hello, I want this PR to go through soon. IIRC both of you were fine with the current deprecation solution? Do you still think so? If both of you are okay let's just mark this as positive review and remove the deprecation warning soon. Or if you two agree we can even remove the deprecation warning now, and just rename this somewhat internal base class

kwankyu:
According to the updated rules in https://doc-release--sagemath.netlify.app/html/en/developer/coding_in_python.html#deprecation, we should first decide if `PolynomialRing_general` is a public class or an internal class.

I think it is internal.

tscrim:
It's in our public documentation and should not be treated as internal. Moreover, it is also such an important class that we can reasonably expect people to have imported it in their local code. Hence, it should be subject to our usual deprecation policy.

I also don't really see the point/benefit of doing this...

kwankyu:
OK. Then it should be treated as a public class. The deprecation policy should be followed.

Otherwise I am positive to the PR.

grhkm21:
So according to the deprecation developer guideline, I should remove the
```python
# Placeholder class for deprecation
class PolynomialRing_general(PolynomialRing_generic):
...
```
placeholder, and just do
```python
class NewClass:
    ...

OldClass = NewClass   # OldClass is deprecated. See Issue 12345.
```
, correct? I assume I can remove it after a year as well (*adds to calendar*)

kwankyu:
Yes.

By the way, I think it should be
```sage
OldClass = NewClass   # OldClass is deprecated. See Issue #12345.
```
Note the added `#`. Indeed the typo in the developer guide seems my mistake in #38273...

mantepse (approved):
LGTM

src/sage/dynamics/arithmetic_dynamics/projective_ds.py
```diff
@@ -395,7 +395,7 @@ def __classcall_private__(cls, morphism_or_polys, domain=None, names=None):
             # homogenize!
             f = morphism_or_polys
             aff_CR = f.parent()
-            if (not isinstance(aff_CR, PolynomialRing_general) and not isinstance(aff_CR, FractionField_generic)
+            if (not isinstance(aff_CR, PolynomialRing_generic) and not isinstance(aff_CR, FractionField_generic)
```
mantepse:
```suggestion
            if (not isinstance(aff_CR, (PolynomialRing_generic, FractionField_generic))
```
should be the same, but that's really not important.

src/sage/ext/fast_callable.pyx
```diff
@@ -475,9 +475,9 @@ def fast_callable(x, domain=None, vars=None,
             x = x.function(*vars)
 
         if vars is None:
-            from sage.rings.polynomial.polynomial_ring import PolynomialRing_general
+            from sage.rings.polynomial.polynomial_ring import PolynomialRing_generic
             from sage.rings.polynomial.multi_polynomial_ring import MPolynomialRing_base
-            if isinstance(x.parent(), PolynomialRing_general) or isinstance(x.parent(), MPolynomialRing_base):
+            if isinstance(x.parent(), PolynomialRing_generic) or isinstance(x.parent(), MPolynomialRing_base):
                 vars = x.parent().variable_names()
```
mantepse:
```suggestion
            P = x.parent()
            if isinstance(P, (PolynomialRing_generic, MPolynomialRing_base)):
                vars = P.variable_names()
```

src/sage/rings/morphism.pyx
```diff
@@ -3150,7 +3150,7 @@ def _tensor_product_ring(B, A):
 
     def term_order(A):
         # univariate rings do not have a term order
-        if (isinstance(A, PolynomialRing_general) or isinstance(A, PolynomialQuotientRing_generic)
+        if (isinstance(A, PolynomialRing_generic) or isinstance(A, PolynomialQuotientRing_generic)
```
mantepse:
```suggestion
        if (isinstance(A, (PolynomialRing_generic, PolynomialQuotientRing_generic))
```

src/sage/rings/morphism.pyx
```diff
@@ -3166,7 +3166,7 @@ def _tensor_product_ring(B, A):
                        order=term_order(B) + term_order(A))
 
     def relations(A, R_gens_A):
-        if isinstance(A, MPolynomialRing_base) or isinstance(A, PolynomialRing_general):
+        if isinstance(A, MPolynomialRing_base) or isinstance(A, PolynomialRing_generic):
```
mantepse:
```suggestion
        if isinstance(A, (MPolynomialRing_base, PolynomialRing_generic)):
```

src/sage/rings/polynomial/flatten.py
```diff
@@ -160,14 +160,14 @@ def __init__(self, domain):
             sage: fl.section()(fl(p)) == p
             True
         """
-        if not isinstance(domain, PolynomialRing_general) and not isinstance(domain, MPolynomialRing_base):
+        if not isinstance(domain, PolynomialRing_generic) and not isinstance(domain, MPolynomialRing_base):
```
mantepse:
```suggestion
        if not isinstance(domain, (PolynomialRing_generic, MPolynomialRing_base)):
```

src/sage/rings/polynomial/flatten.py
```diff
@@ -347,14 +347,14 @@ def __init__(self, domain, codomain):
         """
         if not isinstance(domain, MPolynomialRing_base):
             raise ValueError("domain should be a multivariate polynomial ring")
-        if not isinstance(codomain, PolynomialRing_general) and not isinstance(codomain, MPolynomialRing_base):
+        if not isinstance(codomain, PolynomialRing_generic) and not isinstance(codomain, MPolynomialRing_base):
```
mantepse:
```suggestion
        if not isinstance(codomain, (PolynomialRing_generic, MPolynomialRing_base)):
```

src/sage/rings/polynomial/flatten.py
```diff
@@ -499,7 +499,7 @@ def __init__(self, domain, D):
               Defn: Defined on coordinates by sending (z) to
                     (z^2 + 1.00000000000000)
         """
-        if not isinstance(domain, PolynomialRing_general) and not isinstance(domain, MPolynomialRing_base):
+        if not isinstance(domain, PolynomialRing_generic) and not isinstance(domain, MPolynomialRing_base):
```
mantepse:
```suggestion
        if not isinstance(domain, (PolynomialRing_generic, MPolynomialRing_base)):
```

src/sage/rings/polynomial/multi_polynomial_ring_base.pyx
```diff
@@ -142,7 +142,7 @@ cdef class MPolynomialRing_base(CommutativeRing):
              Multivariate Polynomial Ring in x, y over Rational Field
         """
         base = self.base_ring()
-        if isinstance(base, MPolynomialRing_base) or isinstance(base, polynomial_ring.PolynomialRing_general):
+        if isinstance(base, MPolynomialRing_base) or isinstance(base, polynomial_ring.PolynomialRing_generic):
```
mantepse:
```suggestion
        if isinstance(base, (MPolynomialRing_base, polynomial_ring.PolynomialRing_generic)):
```

src/sage/symbolic/ring.pyx
```diff
@@ -219,7 +219,7 @@ cdef class SymbolicRing(sage.rings.abc.SymbolicRing):
             if R._is_numerical():
                 # Almost anything with a coercion into any precision of CC
                 return R not in (RLF, CLF)
-            elif isinstance(R, PolynomialRing_general) or isinstance(R, MPolynomialRing_base) or isinstance(R, FractionField_generic) or isinstance(R, LaurentPolynomialRing_generic):
+            elif isinstance(R, PolynomialRing_generic) or isinstance(R, MPolynomialRing_base) or isinstance(R, FractionField_generic) or isinstance(R, LaurentPolynomialRing_generic):
```
mantepse:
```suggestion
            elif isinstance(R, (PolynomialRing_generic, MPolynomialRing_base, FractionField_generic, LaurentPolynomialRing_generic)):
```

mantepse:
Since we changed `isinstance(x, ...) or isinstance(x, ...)` to `isinstance(x, (..., ...))` in the last beta (I think), I guess it makes sense to follow that spirit here.

grhkm21:
Thank you! I have applied the changes :)

mantepse (approved):
excellent, thank you!

PR #38209, opened 2024-06-13
Fix infinite loop bug in doctest from #36581

grhkm21:
Discovered by tornaria [here](https://github.com/sagemath/sage/pull/36581/files/5c51f54b1a193c3f9964da3e1388efd87623d45c#r1575566025).

Sorry for the incorrect implementation! Please merge this quickly, before the next beta (I think).

### DIFF
--- a/src/sage/stats/distributions/discrete_gaussian_lattice.py
+++ b/src/sage/stats/distributions/discrete_gaussian_lattice.py
@@ -440,8 +440,7 @@ class DiscreteGaussianDistributionLatticeSampler(SageObject):
             56.5486677646...
 
             sage: from collections import defaultdict
-            sage: counter = defaultdict(Integer)
-            sage: m = 0
+            sage: counter = defaultdict(Integer); m = 0
             sage: def add_samples(i):
             ....:     global counter, m
             ....:     for _ in range(i):
@@ -454,7 +453,7 @@ class DiscreteGaussianDistributionLatticeSampler(SageObject):
             sage: while abs(m*f(v)*1.0/nf/counter[v] - 1.0) >= 0.1:                     # needs sage.symbolic
             ....:     add_samples(1000)
 
-            sage: counter = defaultdict(Integer)
+            sage: counter = defaultdict(Integer); m = 0
             sage: v = vector(ZZ, n, (0, 0))
             sage: v.set_immutable()
             sage: while v not in counter:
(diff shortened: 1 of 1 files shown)


grhkm21:
Should "p: CI Fix" be added?

tscrim (changes requested):
Please be sure to test this locally:
```
File "src/sage/stats/distributions/discrete_gaussian_lattice.py", line 484, in sage.stats.distributions.discrete_gaussian_lattice.DiscreteGaussianDistributionLatticeSampler.__init__
Failed example:
    v = vector(ZZ, n, (0, 0))
Exception raised:
      File "/home/travis/sage-build/src/sage/doctest/forker.py", line 715, in _run
      File "/home/travis/sage-build/src/sage/doctest/forker.py", line 1147, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.stats.distributions.discrete_gaussian_lattice.DiscreteGaussianDistributionLatticeSampler.__init__[23]>", line 1, in <module>
        v = vector(ZZ, n, (Integer(0), Integer(0)))
      File "sage/modules/free_module_element.pyx", line 535, in sage.modules.free_module_element.vector
        raise ValueError("incompatible degrees in vector constructor")
    ValueError: incompatible degrees in vector constructor
```
It subsequently hangs on line 486.

tscrim:
See also the linter.

grhkm21:
My bad. I changed my editor config at some time and must have removed the trailing whitespace highlighting..

grhkm21:
(I'm fixing the hang(?), give me a second)

grhkm21:
`D.f` is not the most descriptive name for a class, so I should refactor it at some point. I will blame it on the original author though 🙈

tscrim (commented):
Some progress, but now line 501 hangs.

grhkm21:
It doesn't hang for me..? can you give me a seed

tscrim:
78175743252608910157834302274104155990

tscrim:
For the record, I am also testing with `--long --verbose`

grhkm21:
I will have a look. I only test with `sage -t --verbose` without long, but I tested it for 10 times without it looping.. maybe I was lucky

tscrim:
This one seed seems particularly bad; when I tried a few more times, everything passes.

grhkm21:
Okay. My `_normalisation_factor_zz` prints a warning message that it's a bad approximate right? It's *really* bad. For the vector `v` in the test case, it outputs 63.x when the true value is 78.x. That determines the computed expected value, so it's way off -> test case loops infinitely. I will just bump up the BOUND and put a TODO...

grhkm21:
Okay so as it turns out I can't do math. I accidentally coded the solution to $B^N \leq 10^4$ as $B = \log_N(10^4)$. I fixed that and now it works.

tscrim (approved):
Thank you. Then if the bots come back green, then we can set this to a positive review.

Issue #38216, opened 2024-06-13
On Decompositions, Generation Methods and related concepts in the theory of Matching Covered Graphs


janmenjayap:
### Problem Description

Matchings and perfect matchings have received considerable attention in graph theory as well as in other related domains (such as, but not limited to, algorithms and optimization). There still remain many open problems — such as Barnette’s conjecture, Berge-Fulkerson conjecture, and so on — due to which it continues to remain an active area of research. For problems concerning perfect matchings, it is well-known that it suffices to solve them for matching covered graphs (that is, those connected graphs wherein each edge belongs to some perfect matching). 

The objective of this issue is to implement efficient algorithms pertaining to the canonical partition, tight cut decomposition, dependency relations, (optimal) ear decomposition, brick and brace generation methods and related concepts in the theory of matching covered graphs, and to make all of these available freely to students, educators as well as researchers all across the world. This issue has been inspired by the book of C.L. Lucchesi and U.S.R. Murty: *PERFECT MATCHINGS: A THEORY OF MATCHING COVERED GRAPHS* <sup>1</sup>.

### Proposed Solution

To create a new class: `MatchingCoveredGraph`,  in the module `sage.graphs.matching_covered_graph` and to implement the following functions:

- `maximal_barrier()`: Return the (unique) maximal barrier of the (matching covered) graph containing the (provided) vertex.
- `canonical_partition()`: Return the canonical partition of the (matching covered) graph.
- `tight_cut_decomposition()`: Return <a maximal set of laminar nontrivial tight cuts, the vertex set partition as the shores wrt these cuts> of the (matching covered) graph.
- `bricks_and_braces()`: Return the list of (underlying simple graph of) the bricks and braces of the (matching covered) graph.
- `number_of_bricks()`: Return the number of bricks.
- `number_of_braces()`: Return the number of braces.
- `number_of_petersen_bricks()`: Return the number of Petersen bricks (the bricks the underlying simple graph of which is isomorphic to `graphs.PetersenGraph()`).
- `is_brick()`: Check if the (matching covered) graph is a brick.
- `is_brace()`: Check if the (matching covered) graph is a brace. 
- `is_removable_edge()`: Check if the edge of the (matching covered) graph is removable.
- `removable_edges()`: Return the list of all removable edges of the (matching covered) graph.
- `is_removable_doubleton()`: Check if the pair of edges constitute a removable doubleton in the (matching covered) graph.
- `removable_doubletons()`: Return the list of all removable doubletons.
[...]

PR #38218, opened 2024-06-14
graphs/matching: add `is_matching_covered()` and `is_bicritical()`

janmenjayap:
The objective of this issue is to implement  `is_matching_covered()` and `is_bicritical()` and to collect the matching related methods and to put them under `matching.py`.

This PR introduces two methods pertaining to matching and aims to list out all matching related functions in `matching.py`. The two new methods are described below:

- [x] `is_matching_covered()`: Check if the graph is matching covered.
- [x] `is_bicritical()`: Check if the graph is bicritical.

This PR addresses couple of missing functions and also addresses the lack of a separate file for matchings.

Fixes #38216. 

### :memo: Checklist

### :hourglass: Dependencies

Nothing as of now (up to my knowledge).

cc: @dcoudert.


dcoudert:
This is the shape of the class. You now have to code and document it. In particular, don't forget to add in the description of the module what a matching covered graph is.

Omegaconstant:
Sure. Thanks for the feedback. 😊👍

I was wondering regarding the following:
Can the `Undirected graphs` module be modified as follows:

1. Putting `matching_polynomial()`, `has_perfect_matching()`, `is_bicritical()`, `is_factor_critical()`, `is_matching_covered()`, `matching()` and `perfect_matchings()` under a section named **Matching**, and
2. Renaming **Leftovers** to **Miscellaneous**.

Please correct me in case I missed something.

dcoudert:
yes this can be done, but in another PR. I fear that you might create a giant PR that will be very hard to review.

Omegaconstant:
Thank you for the suggestion.

The code corresponding *is_bicritical()* and *is_matching_covered()* is
quite analogous to that of *is_factor_criritcal()*.
I was wondering about the following.

For the *is_factor_critical()*, the arguments provided in the code are
as follows:
def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None,
verbose=0,
*, integrality_tolerance=0.001):

But those, that appears in the documentation
<https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph.html#sage.graphs.graph.Graph.is_factor_critical>,
are as follows:
is_factor_critical(matching, algorithm=None, solver='Edmonds', verbose=None
, integrality_tolerance=0)

Could you please let me know why there has been a right shifting in the
default arguments?

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+
+        - [LR2004]_
+
+        .. SEEALSO::
+            :meth:`~sage.graphs.graph.Graph.is_factor_critical`,
+            :meth:`~sage.graphs.graph.Graph.is_matching_covered`,
+            :meth:`~sage.graphs.graph.Graph.is_bicritical`
+        """
+
+        G = self.to_simple()
```
dcoudert: This can be done after some tests

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+
+        # The input vertex must be a valid vertex of the graph
+        if vertex not in G.vertices():
```
dcoudert: use `if vertex not in self:`. the call to `G.vertices()` returns a list of vertices and there is no need to build this list.

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+
+        q = Queue()
+        q.put(vertex)
+
+        even = set([vertex])
+        odd = set()
+        predecessor = {vertex: vertex}
+        rank = {vertex: 0}
+
+        if vertex in M.vertices():
```
dcoudert: `if vertex in M:`, same reason as above

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+            u = M.neighbors(vertex)[0]
```
dcoudert: `u = next(M.neighbor_iterator(u))`

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+            x = q.get()
+            for y in G.neighbor_iterator(x):
+                if y in odd:
+                    continue
+                elif y in even:
+                    # Search t := LCA(x, y)
+                    ancestor_x = [x]
+                    ancestor_y = [y]
+
+                    while ancestor_x[-1] != ancestor_y[-1]:
```
dcoudert: add a comment saying that you are searching for the nearest common ancestor

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+                            ancestor_x.append(predecessor[ancestor_x[-1]])
+                            ancestor_y.append(predecessor[ancestor_y[-1]])
+
+                    lcs = ancestor_x.pop()
+                    ancestor_y.pop()
+                    # Set t as pred of all vertices of the chains and add
+                    # vertices marked odd to the queue
+                    for a in itertools.chain(ancestor_x, ancestor_y):
+                        predecessor[a] = lcs
+                        rank[a] = rank[lcs] + 1
```
dcoudert: you may put the value `rank[lcs] + 1` in a variable to avoid recomputing this value

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+
+                        if a in odd:
+                            even.add(a)
+                            odd.discard(a)
+                            q.put(a)
+
+                elif y in M.vertices():
```
dcoudert: `y in M`

src/sage/graphs/graph.py
```diff
@@ -4328,6 +4328,228 @@ def weight(x):
...
+        REFERENCES:
+
+        - [LR2004]_
+
+        .. SEEALSO::
+            :meth:`~sage.graphs.graph.Graph.is_factor_critical`,
+            :meth:`~sage.graphs.graph.Graph.is_matching_covered`,
+            :meth:`~sage.graphs.graph.Graph.is_bicritical`
+        """
+
```
dcoudert: no need for this empty line

src/sage/graphs/graph.py
```diff
@@ -4531,11 +4753,313 @@ def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None, ve
...
+        # Bipartite graphs of order at least three are not bicritical
+        if self.is_bipartite():
+            if not coNP_certificate:
+                return False
+
+            from sage.graphs.bipartite_graph import BipartiteGraph
+            H = BipartiteGraph(self)
+            color_class_A, color_class_B = H.bipartition()
+            del H
```
dcoudert: no need to `del H`, it will be released with the return

src/sage/graphs/graph.py
```diff
@@ -4531,11 +4753,313 @@ def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None, ve
...
+
+            if len(color_class_A) > 1:
+                return (False, set(list(color_class_A)[:2]))
+            else:
```
dcoudert: you can avoid `else` and simply do the default return

src/sage/graphs/graph.py
```diff
@@ -4531,11 +4753,313 @@ def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None, ve
...
+
+            # It must be a perfect matching
+            if self.order() != M.order():
+                return (False, set([M.edges[0][0], M.edges[0][1]])) if coNP_certificate else False
+
+        # G is bicritical iff for each vertex u with its M-matched neighbor being v,
+        # every vertex of the graph distinct from v must be reachable from u through an even length
+        # M-alternating uv-path starting with an edge not in M and ending with an edge in M
+
+        for u in self.vertices():
```
dcoudert: `for u in self:`

src/sage/graphs/graph.py
```diff
@@ -4531,11 +4753,313 @@ def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None, ve
...
+            v = next(M.neighbor_iterator(u))
+
+            even = self.M_alternating_even_mark(u, M)
+
+            for w in self.vertices():
```
dcoudert: `for w in self:`

dcoudert (changes requested):

src/sage/graphs/graph.py
janmenjayap: If we allow multiple edges in _G_, then the following error happens:

Consider the following example:

```
sage: G = graphs.CompleteBipartiteGraph(3, 3)
sage: G.allow_multiple_edges(True)
sage: G.add_edge(0, 3)
sage: M = G.matching()
sage: u = 0
sage: S = G.M_alternating_even_mark(u, M)
sage: S
```

It shows the following `ValueError` for `G.M_alternating_even_mark(u, M)`:

```
{
	"name": "ValueError",
	"message": "This method is not known to work on graphs with multiedges. Perhaps this method can be updated to handle them, but in the meantime if you want to use it please disallow multiedges using allow_multiple_edges().",
	"stack": "---------------------------------------------------------------------------
/tmp/ipykernel_987822/3075726170.py in ?()
      1 u = Integer(0)
----> 2 G.M_alternating_even_mark(u, M)

~/open-source/sage/src/sage/graphs/graph.py in ?(self, vertex, matching)
   4480         # The input matching must be a valid matching of the graph
   4481         M = Graph(matching)
   4482         if any(d != 1 for d in M.degree()):
   4483             raise ValueError(\"the input is not a matching\")
-> 4484         if not M.is_subgraph(self, induced=False):
   4485             raise ValueError(\"the input is not a matching of the graph\")
   4486 
   4487         # Build an M-alternating tree T rooted at vertex

~/open-source/sage/src/sage/graphs/generic_graph.py in ?(self, other, induced, up_to_isomorphism)
  15813         if not all(other.has_vertex(v) for v in self.vertex_iterator()):
  15814             return False
  15815 
  15816         self._scream_if_not_simple()
> 15817         other._scream_if_not_simple()
  15818 
  15819         if induced:
  15820             # Check whether ``self`` is contained in ``other``

~/open-source/sage/src/sage/graphs/generic_graph.py in ?(self, allow_loops, allow_multiple_edges)
   1495             msg = (\"This method is not known to work on graphs with \" + name + \". \"
   1496                    \"Perhaps this method can be updated to handle them, but in the \" +
   1497                    \"meantime if you want to use it please disallow \" + name + \" using \" +
   1498                    functions + \".\")
-> 1499             raise ValueError(msg)

ValueError: This method is not known to work on graphs with multiedges. Perhaps this method can be updated to handle them, but in the meantime if you want to use it please disallow multiedges using allow_multiple_edges()."

}
```

dcoudert: Yes, but you can at least do this operation after testing if `vertex in self` and if all vertices of `M` have degree 1.

Omegaconstant:
Also,
I have one more doubt, the `Build & Test / test-new (pull_request)` is failing majorly for two reasons:

1. For the following example under `M_alternating_even_mark()`:

```
sage: # K(4) ☉ K(3, 3) is matching covered
sage: G = Graph()
sage: G.add_edges([
....:    (0, 2), (0, 3), (0, 4), (1, 2),
....:    (1, 3), (1, 4), (2, 5), (3, 6),
....:    (4, 7), (5, 6), (5, 7), (6, 7)
....: ])
sage: M = Graph(G.matching())
sage: G.is_matching_covered(M)
True
sage: u = 0
sage: v = next(M.neighbor_iterator(u))
sage: S = G.M_alternating_even_mark(u, M)
sage: (set(G.neighbor_iterator(v))).issubset(S)
True
```

it throws the following error:

```
TypeError: Graph.is_matching_covered() takes 1 positional argument but 2 were given
```

which shall be resolved once `is_matching_covered()` is implemented. 

2. The following line in `is_bicritical()`:

```
4994            components = self.connected_components()
```

for the following example:

```
sage: cycle1 = graphs.CycleGraph(4)
sage: cycle2 = graphs.CycleGraph(6)
sage: cycle2.relabel(lambda v: v + 4)
sage: G = Graph()
sage: G.add_edges(cycle1.edges() + cycle2.edges())
sage: len(G.connected_components())
2
sage: G.is_bicritical()
False
```

throws this for `len(G.connected_components())`:

```
DeprecationWarning: parameter 'sort' will be set to False by default in the future See https://github.com/sagemath/sage/issues/35889 for details. 
2
```

This shows up in the local testing as well, but since it gave correct output `2`, I thought this should not be problem. Any comments on this?

dcoudert:
As I tried to explain earlier, I think that all methods related to matchings should be placed in a module called `matching.py` and then linked to `Graph` the same way we do for other methods. Indeed, I don't think that `M_alternating_even_mark` is needed in `Graph`. This method seems very specific and needed as a building block for other methods, no ?

grhkm21:
To fix "TypeError: `Graph.is_matching_covered()` takes 1 positional argument but 2 were given", you can make the argument list of `is_matching_covered` be `def is_matching_covered(self, *_, **__):`. Also do you mind replacing the `pass` with `raise NotImplementedError()`?

dcoudert:
> To fix "TypeError: `Graph.is_matching_covered()` takes 1 positional argument but 2 were given", you can make the argument list of `is_matching_covered` be `def is_matching_covered(self, *_, **__):`. 

You may use this trick during the development period, but it should not stay for the final version.

src/sage/graphs/graph.py
```diff
@@ -4530,6 +4754,683 @@ def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None, ve
...
+            sage: G = graphs.CircularLadderGraph(8)
+            sage: G.is_bipartite()
+            True
+            sage: G.is_matching_covered()
+            True
+            sage: A, B = G.bipartite_sets()
+            sage: import random
+            sage: a = random.choice(list(A))                                            # needs random
+            sage: b = random.choice(list(B))                                            # needs random
+            sage: G.delete_vertices([a, b])
```
janmenjayap: Could you please let me know why it is throwing the following error during the testing?

```
NameError: name 'a' is not defined
```

But in my local system, it is showing the output as expected.

Thank you.

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
```
dcoudert: something goes wrong here. variable a should be defined.

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+        if not self.is_connected():
+            raise ValueError("the graph is not connected")
+
+        # The graph must have an even order
+        if self.order() % 2:
+            return (False, self.edges()[0]) if coNP_certificate else False
+
+        # If the underlying simple graph is a complete graph of order two,
+        # the graph is matching covered
+        if self.order() == 2:
```
dcoudert: If I understand well, the graph must be simple, without loops or multiple edges.
Add at the beginning of the method `self._scream_if_not_simple()`

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+            return (True, None) if coNP_certificate else True
+
+        if matching:
+            # The input matching must be a valid perfect matching of the graph
+            M = Graph(matching)
+            if any(d != 1 for d in M.degree()):
```
dcoudert: why isn't it sufficient to check that the size of `matching` is twice the order of `M` ?

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+        else:
+            # A maximum matching of the graph is computed
+            M = Graph(self.matching(algorithm=algorithm, solver=solver, verbose=verbose,
+                                    integrality_tolerance=integrality_tolerance))
+
+            # It must be a perfect matching
+            if self.order() != M.order():
+                return (False, M.edges()[0]) if coNP_certificate else False
+
+        G = self.to_simple()
```
dcoudert: isn't it a prerequisite of the method ?

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+
+        '''
```
dcoudert: don't put ` ``` ` in the middle of the code. Use `#` instead.

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+
+        Given a connected bipartite graph G[A, B] with a perfect matching M.
+        Construct a directed graph D from G such that V(D) := V(G) and
+        for each edge in G direct the corresponding edge from A to B in D,
+        if it is in M or otherwise direct it from B to A. The graph G is
+        matching covered if and only if D is strongly connected.
+        '''
+        if G.is_bipartite():
+            A, _ = G.bipartite_sets()
+            color = dict()
```
dcoudert: `color = {u: u not in A for u in G}` ? it seems better to use booleans here. In fact, you may exchange the use of 0 and 1 to test: `u in A` instead.

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+
+            for u in G:
+                color[u] = 0 if u in A else 1
+
+            from sage.graphs.digraph import DiGraph
+            H = DiGraph()
+
+            for edge in G.edges():
```
dcoudert: `G.edge_iterator()`

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+                u, v = edge[0], edge[1]
+
+                if color[u] == 1:
```
dcoudert: `if color[u]:`

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+                    u, v = v, u
+
+                if M.has_edge(edge):
+                    H.add_edge(u, v)
+                else:
+                    H.add_edge(v, u)
+
+            # Check if H is strongly connected using Kosaraju's algorithm
+            def dfs(J, v, visited, p, parent, orientation):
```
dcoudert: I know it's more work, but can't you do a non-recursive version ?

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+                if orientation == 'in':
+                    for u in J.neighbors_out(v):
+                        if not visited[u]:
+                            dfs(J, u, visited, v, parent, orientation)
+                else:
+                    for u in J.neighbors_in(v):
+                        if not visited[u]:
+                            dfs(J, u, visited, v, parent, orientation)
+
+            visited_in = {v: False for v in H.vertices()}
```
dcoudert: simply use `for v in H`

src/sage/graphs/graph.py
```diff
@@ -5059,11 +5059,377 @@ def is_bicritical(self, matching=None, algorithm='Edmonds', coNP_certificate=Fal
...
+            parent_in = {v: None for v in H.vertices()}
+            dfs(H, H.vertices()[0], visited_in, None, parent_in, 'in')
```
dcoudert: use `next(H.neighbor_iterator())`. You can set a variable `root` that will be reused in the next DFS.

dcoudert (changes requested):

src/sage/graphs/graph.py
janmenjayap: Our focus is on only loopless undirected graphs; they might have multiple edges.

src/sage/graphs/graph.py
janmenjayap: This is required for graphs with multiple edges. I think I shall do a check for the non-existence of loops at the very beginning of the method.

src/sage/graphs/graph.py
janmenjayap: Yes, my bad. Thanks for pointing that out.

src/sage/graphs/graph.py
janmenjayap:
```
def dfs(J, v, visited, p, parent, orientation):
    stack = [(v, p)] # a stack of (vertex, parent)

    while stack:
        v, p = stack.pop()

        if not visited[v]:
            visited[v] = True
            parent[v] = p

        if orientation == 'in':
            for u in J.neighbors_out(v):
                if not visited[u]:
                    stack.append((u, v))
        else:
            for u in J.neighbors_in(v):
                if not visited[u]:
                    stack.append((u, v))
```

I suppose the above shall work.

src/sage/graphs/graph.py
janmenjayap: Basically, here we do three checks (if I am not mistaken, it's analogous to the one that is done in the `is_factor_critical()` one):
1. check if `matching` is a valid matching (so `M` shall be a 1-regular graph),
2. check if `matching` is a matching of the graph provided (to ensure it does not have any other edges/ vertices that are not there in the original graph), and
3. Finally, check if the order of the original graph is twice that of the size of the `M`.

src/sage/graphs/graph.py
dcoudert: variables `a` and `b` are created with annotations `# needs random`. Consequently, these variables can be used only for lines with the same annotation.
The solution is to use a block scope, that is add at the beginning of this block of example the line:
```
sage: # needs random
```

src/sage/graphs/graph.py
dcoudert: You can force the input graph to be simple using `G._scream_if_not_simple()`. This way the method will be valid only for simple graphs.

src/sage/graphs/graph.py
```diff
@@ -4524,6 +4757,680 @@ def is_factor_critical(self, matching=None, algorithm='Edmonds', solver=None, ve
...
+          If set to ``None``, a matching is computed using the other parameters.
+
+        - ``algorithm`` -- string (default: ``'Edmonds'``); the algorithm to be
+          used to compute a maximum matching of the graph among
+
+          - ``'Edmonds'`` selects Edmonds' algorithm as implemented in NetworkX,
+
+          - ``'LP'`` uses a Linear Program formulation of the matching problem.
+
+        - coNP_certificate -- boolean (default: ``False``); if set to ``True``
```
dcoudert: you must use ``` ``coNP_certificate`` ```

dcoudert (changes requested):
As I already said, it would be much better to create a new file `matching.py` to gather all methods related to matchings, and in particular the new methods you are adding.

janmenjayap:
Will be creating `matchin.py` shortly.

janmenjayap:
Will add `matching.py` shortly.

dcoudert:
> If we use `self._scream_if_not_simple()` this avoid the graphs with either multiple edges or loops or with both.
> 
> Is there any inbuilt function using which we may allow graphs with multiple edges, but raise `ValueError` in case it has loops. (Another possibility will be check for loops manually using for loop and raise `ValueError` if found any.)

Check the documentation of `_scream_if_not_simple`... It can do what you need.

janmenjayap:
I have a suggestion:
Would it be possible to consider renaming the `'Leftovers'` section in `Undirected graphs` module to `'Miscellaneous'`?

Also, considering the last commit, the way git compared the old and new changes in the last commit seems a bit unusual, especially since the main change was moving some methods from `graph.py` to `matching.py`.

janmenjayap:
```
5399 + 
5400 +            visited_in = {v: False for v in H.vertices()}
5401 +            parent_in = {v: None for v in H.vertices()}
5402 +            dfs(H, H.vertices()[0], visited_in, None, parent_in, 'in')
```

Could you please let me know how to use `next(H.neighbor_iterator())` here? I suppose I should set the 0th vertex as `root`.
Thank you.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1627 @@
+r"""
+Matching
+
+This module implements the functions pertaining to matching of undirected
+graphs. A `matching` in a graph is a set of pairwise nonadjacent links
```
dcoudert: don't use single ` for italic, it's for latex. You can use * or **.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1627 @@
...
+    :widths: 30, 70
+    :delim: |
+
+    :meth:`~has_perfect_matching` | Return whether this graph has a perfect matching
+    :meth:`~is_bicritical` | Check if the graph is bicritical
+    :meth:`~is_factor_critical` | Check whether this graph is factor-critical
+    :meth:`~is_matching_covered` | Check if the graph is matching covered
+    :meth:`~matching` | Return a maximum weighted matching of the graph represented by the list of its edges
+    :meth:`~perfect_matchings` | Return an iterator over all perfect matchings of the graph
+    :meth:`~M_alternating_even_mark` | Return the set of vertices each of which is reachable from the provided vertex through an (even length matching-alternating) path starting with an edge not in the matching and ending with an edge in the matching
```
dcoudert: please find a shorter form

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1627 @@
...
+            if not edges:
+                return False
+            p.add_constraint(p.sum(b[frozenset(e)] for e in edges) == 1)
+        try:
+            p.solve(log=verbose)
+            return True
+        except MIPSolverException:
+            return False
+    raise ValueError('algorithm must be set to "Edmonds", "LP_matching" or "LP"')
+
```
dcoudert, line 160: we need 2 empty lines between methods

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1627 @@
...
+          maximal cardinality, then check whether this cardinality is half the
+          number of vertices of the graph.
+
+        - ``'LP'`` uses a Linear Program formulation of the perfect matching
+          problem: put a binary variable ``b[e]`` on each edge `e`, and for
+          each vertex `v`, require that the sum of the values of the edges
+          incident to `v` is 1.
+
+    - ``solver`` -- string (default: ``None``); specifies a Mixed Integer
+        Linear Programming (MILP) solver to be used. If set to ``None``, the
```
dcoudert: you should have only 2 extra spaces of indentation, not 4.
```
  - ``solver`` -- string (default: ``None``); specifies a Mixed Integer
    Linear Programming (MILP) solver to be used. If set to ``None``, the
    default one is used. For more information on MILP solvers and which
  ```

The same everywhere

File: src/sage/graphs/matching.py [hunk 30]
```diff
@@ -0,0 +1,1627 @@
...
+
+    Providing with a graph with a self-loop::
+
+        sage: G = graphs.CompleteGraph(4)
+        sage: G.allow_loops(True)
+        sage: G.add_edge(0, 0)
+        sage: G.is_bicritical()
+        Traceback (most recent call last):
+        ...
+        ValueError: This method is not known to work on graphs with loops. Perhaps this method can be updated to handle them, but in the meantime if you want to use it please disallow loops using allow_loops().
```
dcoudert: you can put this text on multiple lines.

dcoudert (changes requested):

dcoudert:
> I have a suggestion: Would it be possible to consider renaming the `'Leftovers'` section in `Undirected graphs` module to `'Miscellaneous'`?
> 
> Also, considering the last commit, the way git compared the old and new changes in the last commit seems a bit unusual, especially since the main change was moving some methods from `graph.py` to `matching.py`.

May be, but not in this PR. It is bad practice to do too many things in a single PR as it makes it harder to review.

dcoudert:
> ```
> 5402 +            dfs(H, H.vertices()[0], visited_in, None, parent_in, 'in')
> ```
> 
> Could you please let me know how to use `next(H.neighbor_iterator())` here? I suppose I should set the 0th vertex as `root`. Thank you.

Yes, it should be possible to replace `H.vertices()[0]` with `next(H.neighbor_iterator())`.

dcoudert:
> ```
>         root = next(H.neighbor_iterator())
> ```
> 
> In the above piece of code, I got the following error:
> 
> ```
> TypeError: GenericGraph.neighbor_iterator() missing 1 required positional argument: 'vertex'
> ```

arghh, of course, it's not `neighbor_iterator` but `vertex_iterator` !!!

src/sage/graphs/matching.py
janmenjayap: The last line I have changed to the following one:
```
    :meth:`~M_alternating_even_mark` | Return the set of vertices reachable from the provided vertex via an even alternating path starting with a non-matching edge
```

It spans two lines, instead of four line as earlier in a default screen setting.

Please do let me know if any shorter form is required.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+
+    A nontrivial circular ladder graph whose order is not divisible by 4 is bicritical::
+
+        sage: G = graphs.CircularLadderGraph(5)
+        sage: G.is_bicritical()
+        True
+
+    The graph obtained by splicing two bicritical graph is also bicritical::
+
+        sage: # K(4) with one extra edge (say K(4)+) is bicritical
```
dcoudert: this comment can be in the description of this doctest.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        sage: G = graphs.CompleteGraph(4)
+        sage: G.allow_multiple_edges(True)
+        sage: G.add_edge(0, 1)
+        sage: G.is_bicritical()
+        True
+        sage: # Let H := K(4)+ ☉ #K(4)+ such that H has no multiple egde
```
dcoudert: This comment should be in the description of the doctest

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        # Check if there are at least two even components
+        components_of_even_order = [component for component in components if len(component) % 2 == 0]
+        if len(components_of_even_order) > 1:
+            return (False, set([components_of_even_order[0][0], components_of_even_order[1][0]]))
+
+        # Or otherwise there is at most one even component with at least two trivial odd components
+        u, v = None, None
+
+        for component in components:
+            if len(component) % 2 == 0 and u is not None:
```
dcoudert: change to `if u is not None and not len(component) % 2:`

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+    This method implements the algorithm proposed in [LR2004]_ and we assume
+    that a graph of order one is factor-critical. The time complexity of the
+    algorithm is linear if a near perfect matching is given as input (i.e.,
+    a matching such that all vertices but one are incident to an edge of the
+    matching). Otherwise, the time complexity is dominated by the time
+    needed to compute a maximum matching of the graph.
+
+    INPUT:
+
+    - ``matching`` -- (default: ``None``) a near perfect matching of the
```
dcoudert: add a `;` after `(default: ``None``);`

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        A, _ = G.bipartite_sets()
+        color = dict()
+
+        for u in G:
+            color[u] = 0 if u in A else 1
+
+        from sage.graphs.digraph import DiGraph
+        H = DiGraph()
+
+        for edge in G.edges():
```
dcoudert: since you don't use edge labels, you can do
```
for u, v in G.edge_iterator(labels:False):
    if color[u] == 1:
        u, v = v, u
....
```

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+                u, v = v, u
+
+            if M.has_edge(edge):
+                H.add_edge(u, v)
+            else:
+                H.add_edge(v, u)
+
+        # Check if H is strongly connected using Kosaraju's algorithm
+        def dfs(J, v, visited, p, parent, orientation):
+            stack = [(v, p)] # a stack of (vertex, parent)
```
dcoudert: 2 spaces before #

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+                elif orientation == 'out':
+                    for u in J.neighbors_in(v):
+                        if not visited[u]:
+                            stack.append((u, v))
+                else:
+                    raise ValueError('Unknown orientation')
+
+        root = next(H.vertex_iterator())
+
+        visited_in = {v: False for v in H}
```
dcoudert: you could instead use a set, initially empty. Indeed, instead of tests like `if not visited[u]`, you can do `if u not in visited`.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        parent_in = {v: None for v in H}
```
dcoudert: unless I'm mistaken, you feed `parent_in` but you never use it. If so, you can avoid maintaining parent.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        dfs(H, root, visited_in, None, parent_in, 'in')
+
+        visited_out = {v: False for v in H}
+        parent_out = {v: None for v in H}
+        dfs(H, root, visited_out, None, parent_out, 'out')
+
+        for edge in H.edges():
```
dcoudert: use `H.edge_iterator()`

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        sage: S = G.M_alternating_even_mark(v, M)
+        sage: S == set(G.vertices())
+        True
+
+    For a matching covered graph `G` with a perfect matching `M` and for
+    some vertex `u` with `v` being its `M`-matched neighbor, each neighbor
+    of `v` is reachable from `u` through an even length `M`-alternating
+    path as described above::
+
+        sage: # K(4) ☉ K(3, 3) is matching covered
```
dcoudert: this comment should be in the text above.

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        sage: S = G.M_alternating_even_mark(u, M)
+        sage: (set(G.neighbor_iterator(v))).issubset(S)
+        True
+
+    For a bicritical graph `G` with a perfect matching `M` and for some
+    vertex `u` with its `M`-matched neighbor being `v`, each vertex of the
+    graph distinct from `v` is reachable from `u` through an even length
+    `M`-alternating path as described above::
+
+        sage: # Petersen graph is bicritical
```
dcoudert: this comment should be in the text above

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+        sage: G = graphs.PetersenGraph()
+        sage: M = Graph(G.matching())
+        sage: G.is_bicritical(M)
+        True
+        sage: import random
+        sage: u = random.choice(G.vertices())                                       # needs random
+        sage: v = next(M.neighbor_iterator(u))
+        sage: S = G.M_alternating_even_mark(u, M)
+        sage: S == (set(G.vertices()) - {v})
```
dcoudert: `set(G)`

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+    # The result is equivalent for the underlying simple graph of the provided
+    # graph. So, the underlying simple graph is considered for implementational
+    # simplicity.
+    G = self.to_simple()
+
+    # The input matching must be a valid matching of the graph
+    M = Graph(matching)
+    if any(d != 1 for d in M.degree()):
+        raise ValueError("the input is not a matching")
+    if not M.is_subgraph(G, induced=False):
```
dcoudert: isn't it equivalent to `if any(not G.has_edge(e) for e in M.edge_iterator()):` ?

src/sage/graphs/graph.py
```diff
@@ -10228,6 +9612,13 @@ def bipartite_double(self, extended=False):
     from sage.graphs.graph_coloring import fractional_chromatic_number
     from sage.graphs.graph_coloring import fractional_chromatic_index
     from sage.graphs.hyperbolicity import hyperbolicity
+    has_perfect_matching = LazyImport('sage.graphs.matching', 'has_perfect_matching', at_startup=True)
+    is_bicritical = LazyImport('sage.graphs.matching', 'is_bicritical', at_startup=True)
+    is_factor_critical = LazyImport('sage.graphs.matching', 'is_factor_critical', at_startup=True)
+    is_matching_covered = LazyImport('sage.graphs.matching', 'is_matching_covered', at_startup=True)
+    matching = LazyImport('sage.graphs.matching', 'matching', at_startup=True)
+    perfect_matchings = LazyImport('sage.graphs.matching', 'perfect_matchings', at_startup=True)
+    M_alternating_even_mark = LazyImport('sage.graphs.matching', 'M_alternating_even_mark', at_startup=True)
```
dcoudert: why should this method be exposed in Graph ? is it needed for other purposes than where it is already used ?

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1635 @@
...
+    # For each unlabeled matching, we yield all its possible labelings
+    import itertools
+
+    for m in rec(G):
+        yield from itertools.product(*[edges[frozenset(e)] for e in m])
+
+
+def M_alternating_even_mark(self, vertex, matching):
+    r"""
+    Return the set of vertices each of which is reachable from the provided
```
dcoudert: This sentence is way too long. Consider having first a single line description and then more text.
For instance, `Return the vertices reachable from and even length alternating path starting from ``vertex``. `

dcoudert (changes requested):

src/sage/graphs/matching.py
janmenjayap: Perhaps we may consider rephrasing it as follows:

'Return the vertices reachable from `vertex` via an even alternating path starting with a non-matching edge'

This seems sufficient. (Since the path length is even, it is alternating and starts with an edge not in the matching, is should end with an edge that is in the matching.)

src/sage/graphs/graph.py
janmenjayap: Some other methods, for instance `matching_polynomial()`, `chromatic_polynomial()`, `geodetic_closure()`, `tutte_polynomial()`, `lovasz_theta()` and so on are defined in other files and have been imported to the `undirected graph` section.

Analogously, I thought it would be better if we could accumulate the matching related functions in `matching.py` with the two functions namely, `matching_polynomial()` and `complete_poly()`, defined in `matching_polynomial.py` and show it in `undirected graph` section.

src/sage/graphs/matching.py
janmenjayap: Seems so.

dcoudert:
some tests are failing.

src/sage/graphs/matching.py
```diff
@@ -1053,10 +1052,10 @@ def dfs(J, v, visited, orientation):
 
         root = next(H.vertex_iterator())
 
-        visited_in = {v: False for v in H}
+        visited_in = dict()
```
dcoudert: `set()` or `{}`,  not `dict()`.

dcoudert (changes requested):

src/sage/graphs/matching.py
```diff
@@ -0,0 +1,1629 @@
...
+                u, v = v, u
+
+            if M.has_edge(u, v):
+                H.add_edge(u, v)
+            else:
+                H.add_edge(v, u)
+
+        # Check if H is strongly connected using Kosaraju's algorithm
+        def dfs(J, v, visited, orientation):
+            stack = [v] # a stack of vertices
```
dcoudert: 2 spaces before #

dcoudert (changes requested):
one remaining minor comment. Otherwise LGTM.

src/sage/graphs/matching.py
janmenjayap: If I am not mistaken, by 2 spaces, it is meant two blank lines above the comment in line 1031, right?

janmenjayap: Please note that if it corresponds to two blank lines, I already got Lint error for the same in [this](https://github.com/sagemath/sage/pull/38218/commits/93a1b572c18434fe01ef7e1a0648b00f0e738b1c) commit.
Please kindly let me know if I am not mistaken.

src/sage/graphs/graph.py
dcoudert: some methods could be moved, but in a future PR.

src/sage/graphs/matching.py
dcoudert: ok

janmenjayap:
It seems like there are other files for instance `/sage/src/sage/doctest/forker.py` that import methods like `has_perfect_matching()` from graph.py. That's the reason after the removal of lazy_import corresponding these methods some tests are failing.

Shall we add them back in this PR itself?

dcoudert (changes requested):
For each method moved from `graph.py` to `matching.py`, you must expose the method in `graph.py`, as done for other methods at the bottom of the file.

I only doubt that exposing `M_alternating_even_mark` in `Graph` is useful. The method is too specific and can be accessed if needed anyway.

src/sage/graphs/graph.py
```diff
@@ -9612,6 +9612,8 @@ def bipartite_double(self, extended=False):
     from sage.graphs.graph_coloring import fractional_chromatic_number
     from sage.graphs.graph_coloring import fractional_chromatic_index
     from sage.graphs.hyperbolicity import hyperbolicity
+    from sage.graphs.matching import (has_perfect_matching, matching,
```
dcoudert: please do 1 line per import, as for other methods around. Add them in the doc as well. `is_bicritical` is missing.

dcoudert (changes requested):

janmenjayap:
> For each method moved from `graph.py` to `matching.py`, you must expose the method in `graph.py`, as done for other methods at the bottom of the file.
> 
> I only doubt that exposing `M_alternating_even_mark` in `Graph` is useful. The method is too specific and can be accessed if needed anyway.

I suppose the inclusion of `M_alternating_even_mark` is required for the successful completion of the test cases and examples of the same method as the method is used as an attribute of `Graph` class in those tests and examples.

src/sage/graphs/graph.py
janmenjayap: Shall I add include these methods under `Matching` or under `Leftovers` in the doc of `graph.py`?

dcoudert: Matching is certainly appropriate.

dcoudert:
> I suppose the inclusion of `M_alternating_even_mark` is required for the successful completion of the test cases and examples of the same method as the method is used as an attribute of `Graph` class in those tests and examples.

Or change calls from `G_simple.M_alternating_even_mark(u, M)` to `M_alternating_even_mark(G_simple, u, M)`...

dcoudert (approved):
This looks good to me.

Issue #37586, opened 2024-03-11
`build/pkgs/*/SPKG.rst`: Explain how SageMath makes use of the package


mkoeppe:
From https://github.com/sagemath/sage/pull/37581#discussion_r1518967653

I'd like to expand some of the package descriptions in the SPKG.rst with a description of how the package is being used in the Sage project. (That's part of a larger effort to improve the relation of Sage with upstream projects.)

kwankyu:
We may start by writing something about this direction in the developer guide. Then whenever a package is updated, the description may be improved.

mkoeppe:
Good idea!

PR #38219, opened 2024-06-14
`sage --package`: Multiple tarballs, binary wheels; `build/pkgs/jsonschema`: Update; `clarabel`: Change to binary wheel

mkoeppe:
In particular, `sage --package create --pypi --source wheel` can now do platform-dependent wheels

This enables, for example, the use of Rust-based packages, which is increasingly popular https://groups.google.com/g/sage-devel/c/JiSryImYGe0 @guojing0 

As illustrations, 
- we update `jsonschema`, which now has the Rust-based package `rpds_py` as a dependency, which we supply as a platform-dependent wheel;
- we change `clarabel`, the Rust-based package which will become the default interior point solver in a CVXpy update, to a platform-dependent wheel package.

We add a policy for standard packages from binary wheels to the Developer Guide. Discussion: https://groups.google.com/g/sage-devel/c/vFsNl87zLxo

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #38224 (merged here)
- Depends on #38468


kwankyu:
> we update jsonschema, which now has the Rust-based package rpds_py as a dependency, which we supply as a platform-dependent wheel;

So jsonschema of version 4.17.3, we are currently shipping, does not have rpds_py as a dependency, but newer version has it as dependency and the dependency is Rust-based. Right?

kwankyu:
And more and more our standard packages will itself be Rust-based or will have Rust-based dependencies. Is this what you speculate?

kwankyu:
Assuming yes to my questions, I think that the general direction of our project should be to make those packages prerequisites of the part of the sage library using the packages, and the part of the sage library itself an optional package.
When the user wants to install the part of the sage library, Sage will check the prerequisite if it is available from the system, if not, attempt to install from platform-dependent wheels, and if not possible, suggest the user to install the system package.

mkoeppe:
> So jsonschema of version 4.17.3, we are currently shipping, does not have rpds_py as a dependency, but newer version has it as dependency and the dependency is Rust-based.

That's right. And upgrading sonschema is a prerequisite for upgrading jupyter packages (to be done in a follow-up PR).

kwankyu:
So upgrading jupyter package, which is standard, forces us to have standard platform-dependent wheel packages, unless we introduce Rust compilers into sage, which is unlikely.

mkoeppe:
> [...] make those packages prerequisites of the part of the sage library using the packages, and the part of the sage library itself an optional package.

We'll really only need such a package when there is a compile-time dependency of a Cython module in the Sage library on the package provided by the platform-dependent wheel.

For the example package done here in this PR, `clarabel`, note there is no Cython module in the Sage library with a compile-time dependency on **clarabel**. The only dependency is `sage.numerical.backends.cvxpy_backend`, which has **cvxpy** as a runtime dependency, which then has **clarabel** as a runtime dependency.

mkoeppe:
> And more and more our standard packages are likely to be Rust-based or will have Rust-based dependencies. Is this what you speculate?

Yes, this is what I expect to happen.

mkoeppe:
> When the user wants to install [...], Sage will check the prerequisite if it is available from the system, if not, attempt to install from platform-dependent wheels, and if not possible, suggest the user to install the system package.

Yes; only the last part, suggesting the user to install the system package when no wheel fits, is a mechanism that is not implemented yet but could be added easily. Perhaps it could also suggest the other recourse (using the appropriate `--disable-...` flag).

mkoeppe:
> upgrading jupyter package, which is standard, forces us to have standard platform-dependent wheel packages, unless we introduce Rust compilers into sage, which is unlikely.

That's right, and this is an upgrade that I would like to make in the 10.5 cycle.

On a slightly longer timescale, I hope to make `cvxpy` standard, possibly replacing `glpk` as the default MIP solver. In this case, the dependency `clarabel` would also become a standard package.

kwankyu:
OK. It seems that we cannot avoid standard platform-dependent wheel packages. But as said before, enough wheels should be available for the officially-supported platforms. Do you agree? It it is optional, perhaps the range may be more loose (and we may ask the user to install the system package as prerequiste).

Then how "enough" would be enough?

mkoeppe:
> enough wheels should be available for the officially-supported platforms.

Yes, but that's a tautological condition.

kwankyu:
For example, the list in [build/pkgs/clarabel/checksums.ini](https://github.com/sagemath/sage/pull/38219/files#diff-89aea46795e03acfd21680103590a9b3d561d722c420eca746cb36d7acdcf78d) is enough for `clarabel` to become standard? If not, why not?

kwankyu:
> > enough wheels should be available for the officially-supported platforms.
> 
> Yes, but that's a tautological condition.

You mean that the list of available wheels will define the "officially-supported" platforms?

mkoeppe:
Yes, just like in current practice the set of supported platforms is simply the set of platforms on which all standard packages build successfully.

kwankyu:
OK. Then the release note for the current milestone may be consulted (or updated) whenever a platform-dependent package is created or upgraded.

mkoeppe:
> For example, the list in [build/pkgs/clarabel/checksums.ini](https://github.com/sagemath/sage/pull/38219/files#diff-89aea46795e03acfd21680103590a9b3d561d722c420eca746cb36d7acdcf78d) is enough for `clarabel` to become standard?

Yes, I think so, but I have not run the full portability tests.

mkoeppe:
> the release note for the current milestone may be consulted (or updated) whenever a platform-dependent package is created or upgraded.

Yes, same as for any other type of package. Changes in platform support are also tracked in #32074.

mkoeppe:
I have removed the improper use of the "disputed" label.

mantepse:
> I have removed the improper use of the "disputed" label.

Why is the use "improper"?

Apart from that, it seems to me that removing the disputed label should only be possible with consensus.

mkoeppe:
I've blocked @mantepse on GitHub.

kwankyu:
This is "off-topic", but is it total blocking? Why?

mantepse:
Thank you for being open about that.

mkoeppe:
@kwankyu The general purpose of marking comments as off-topic (by project maintainers), as well as of blocking users (by individuals, for commenting on the individual's Issues and PRs), is to reduce distracting noise that interferes with the review process.

At this point, I do not have more specific public announcements to make.

kwankyu:
I get
```bash
$ sage --package create pyo3_pack --pypi --source wheel 
Requires-Python: 
```
What am I doing wrong?

mkoeppe:
Nothing wrong. That's not an error message. (The output is admittedly not very enlightening.)

build/pkgs/rpds_py/SPKG.rst
```diff
@@ -0,0 +1,18 @@
...
+License
+-------
+
+MIT
+
+Upstream Contact
+----------------
+
+https://pypi.org/project/rpds-py/
+
```
kwankyu, line 21: It seems that rpds_py replaces pyrsistent. They are standard packages. We have a strict policy regarding changes on standard packages. As I said elsewhere, I think we are making a promise for standard packages that the standard packages are available in the standard install. Replacing a standard package with another silently breaks the promise.

I know that these packages are dependencies of a standard package. So I guess you apply the policy loosely to these. I understand this. So I think we should make the distinction between the main standard packages and the auxiliary standard packages. As there is no need to make the distinction on the technical side, I suggest to add a paragraph (perhaps under "Description") explaining how the package is used in sage and to which standard package it is a dependency and the like, hinting that there may be changes to this package without notice (on sage-devel).

I think these notices should be added to all auxiliary standard packages (later).

mkoeppe: Yes, that's a good idea and could be a good step towards
- #37586

kwankyu:
> Nothing wrong. That's not an error message. (The output is admittedly not very enlightening.)

Not only that. It is misleading. So the package is created? There's no git diff.

mkoeppe:
> There's no git diff.

The files aren't `git add`ed automatically (`sage --package create` never does that). They appear as untracked files.

kwankyu:
Ah I see.

build/pkgs/rpds_py/SPKG.rst
mkoeppe: I'll offer that to some extent, the tags `math` and `front-end` may be filling this role, as they are packages that we specifically advertise to users.

mkoeppe:
Only the `sage --package update...` commands have an option `--commit` that also runs `git`. It's helpful there because this is often fully automatic.

When I run `sage --package create`, on the other hand, I often edit details by hand or add files such as `distros/...`, so it's less useful to try to automate some `git` operation.

kwankyu:
Yes. That is very reasonable. 

By the way, what does this message `Requires-Python:` mean?

mkoeppe:
If the package only supports specific Python versions, this version constraint may appear there.
It's an information message that is very useful so as to avoid breaking support for supported Python versions.

mkoeppe:
I've pushed some little output improvements.

kwankyu:
Does the platform-dependent package clarabel cover all supported platforms? How can I (a reviewer) check that? This seems to require expertise that I don't have...

mkoeppe:
I was about to write that the change from `pip` package to `wheel` package does not affect what platforms are supported, but it's actually a bit more complicated:

 Theoretically for a package it could happen that some platform is only covered by some older version of the package on PyPI (and thus could be installed by a `pip` package, whereas the `wheel` package is only created with the wheels available for the latest release.

The details of the platform compatibility tags of wheels are quite complicated, so my only succinct advice can be that all optional packages are tested by the "CI Linux" and "CI macOS" workflows for the supported platforms.

kwankyu:
So I will take it that clarabel was tested (by "CI Linux" and "CI macOS" workflows) for already supported platforms. 

Perhaps "CI Linux Incremental" workflow here is also testing this PR (on some? platforms), but many of them already fail for other reasons.

mkoeppe:
Yes, CI Linux Incremental also tests the package, but it does not provide full coverage.

build/pkgs/rpds_py/SPKG.rst
kwankyu: Thanks. Good enough.

kwankyu (approved):
Thanks. LGTM.

kwankyu:
I am going to give positive review after examining the CI results.

With that, I will remove the "disputed" label, which I think was prematurely added. 

Then we may go on with the normal procedure, including raising any concerns or objections.

mkoeppe:
> I will remove the "disputed" label, which I think was prematurely added.

Yes, that label was clearly added without even looking at the PR's branch.

kwankyu:
Now voice any objections clearly so that the author may respond.

saraedum:
@kwankyu thanks for reviewing this PR. I have not looked into the details here yet, so let me reinstate the disputed label so this does not get picked up by the release manager without knowing whether the dispute has actually been resolved. (Note that you can tally votes after allowing some time for people to weigh in which restores the positive review.)

@kwankyu please let me know whether I am making a mistake here, I might have missed something in the discussion already.

I can see two points raised by Dima at least, I'll open a separate issue so Dima can voice any further concerns there.

kwankyu:
Good idea. Let's hear Dima's concerns through #38551, and also let the author respond to them.

mkoeppe:
I object to this ad-hoc procedure.

With the "disputed" label, we voted for a mechanism to resolve conflicts in the review process.
We did not vote for establishing a mechanism that legitimizes the display of hostility and obstructing work.

Pasechnik has not participated in the review of this PR: neither here, where he is blocked by me, nor in https://groups.google.com/g/sage-devel/c/vFsNl87zLxo, nor in Issue https://github.com/sagemath/sage/issues/38475 that he opened for the purpose of circumventing the block). There's not a hint of review, only a repeated false accusation.

mkoeppe:
I'll further note that CoCC does not have the authority to invent ad-hoc governance procedures. 

Therefore I'm removing the "disputed" label.

mkoeppe:
The manipulation of labels will stop.

saraedum:
> When there is disagreement about whether a PR should be merged, anyone may mark a PR as disputed.

Dima made two concrete objections. He apparently feels that they have not been addressed and that this PR should not be merged. Why would he not be entitled to mark it as disputed?

Why do you have the right to remove his disputed label?

> I'll further note that CoCC does not have the authority to invent ad-hoc governance procedures.

I didn't say in my comment that I acted on behalf of the CoCC. Also, nothing was invented here. I used the disputed procedure as I understand it.

> Pasechnik has not participated in the review of this PR

He made two objections. To me it appears that he participated in the review.

Since I feel that the disputed label was not removed according to the rules that we voted upon, I am reinstating it.

mkoeppe:
> > I'll further note that CoCC does not have the authority to invent ad-hoc governance procedures.
> 
> I didn't say in my comment that I acted on behalf of the CoCC. Also, nothing was invented here. I used the disputed procedure as I understand it.

Well, Julian, if you are not acting on behalf of the CoCC, then you should certainly not be operating the label. You are also not a participant in the review of this PR.

mkoeppe:
I've marked discussions of procedure as off-topic. See https://groups.google.com/g/sage-devel/c/Sm-HG9zQTQY/m/Ffs3CIPMAgAJ for explanation of the purpose.

kwankyu:
This is also "off-topic" according to Matthias' rule.

Let me be clear about my position and my view of the current situation.

1. Setting comments as "off-topic" may be used as an effective measure to hide offensive or distracting comments. Comments about proper handling of an individual PR is not "off-topic". This comment is not "off-topic".

2. Dima is a member of our community. He has the right to participate in the reviewing process of any PR. I know that he has been hostile to Matthias. But he submitted the concerns why he objects this PR.  They are not like "I object this PR because it is by Matthias".

3. Matthias rejected the concerns. Then this PR is in the state of a disputed PR. A disputed PR needs a "disputed" label.

4. I want this PR to be merged through the procedure upon which the community agreed. I don't want my own review to be misused, out of the procedure. So I withdraw my "positive review".

mkoeppe:
@kwankyu As you withdraw your positive review, the PR is in "needs review" status now.

mkoeppe:
> * Dima is a member of our community. He has the right to participate in the reviewing process of any PR. I know that he has been hostile to Matthias.

Note that participation in the project's process is subject to CoC.

orlitzky:
PR is stale, progress is being made on https://github.com/sagemath/sage/issues/41135 and https://github.com/sagemath/sage/pull/41535 instead.

Issue #37384, opened 2024-02-18
Version selector for the documentation


mkoeppe:
- We have the stable version at https://doc.sagemath.org/html/en/ 
- We have the development version at https://deploy-livedoc--sagemath.netlify.app/html/en (linked from https://github.com/sagemath/sage/wiki). 
- We have the version of a particular branch created by the doc-build workflow.
- We can build locally.

We should include a version selector in the generated HTML files that allows us to switch between these versions easily.

Multiple technical solutions:

Version switcher dropdowns — PyData Theme 0.15.2 documentation
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html

sphinx-versioned-docs · PyPI
https://pypi.org/project/sphinx-versioned-docs/

Quickstart — sphinx-multiversion 0.2.4 documentation
https://holzhaus.github.io/sphinx-multiversion/master/quickstart.html

steinwurf/versjon: Tool for injecting version selector in Sphinx's HTML output
https://github.com/steinwurf/versjon

My Sphinx Best Practice Guide for Multi-version Documentation in Different Languages — Coding With Thomas
https://www.codingwiththomas.com/blog/my-sphinx-best-practice-for-a-multiversion-documentation-in-different-languages

Versioned documentation — Read the Docs user documentation
https://docs.readthedocs.io/en/stable/versions.html

Omegaconstant:
Hello,
I would like to work on this.

Thank you :-)

mkoeppe:
Implemented in #38285

PR #38220, opened 2024-06-14
Fix doc deployment to netlify for releases

kwankyu:
Fix a bug that deployed a PR doc preview to the production site https://sagemath.netlify.app, which is supposed to serve the doc preview for a release.

The reason of the behavior seems that if PR number is missing somehow, the doc is deployed to the production site, as seen here
 
Screenshot 2024-06-15 at 12 33 28 AM

----
It started so at least. Now this PR ended in fixing a few things in doc-publish and doc-build-pdf, and overhauling doc-publish logic. 

Tests:

on pull request: https://github.com/kwankyu/sage/actions/runs/9635036620 
deployed site: https://doc-pr-42--sagemath-test.netlify.app

on push branch develop: https://github.com/kwankyu/sage/actions/runs/9640827756
deployed site: https://doc-develop--sagemath-test.netlify.app

on push tag: https://github.com/kwankyu/sage/actions/runs/9641539263
deployed site: https://doc-10-4-beta29--sagemath-test.netlify.app
deployed site: https://doc-release--sagemath-test.netlify.app

on pull request, build-doc-pdf: https://github.com/kwankyu/sage/actions/runs/9634975196

### :memo: Checklist

### DIFF
--- a/.ci/create-changes-html.sh
+++ b/.ci/create-changes-html.sh
@@ -19,7 +19,9 @@ echo '<script>hljs.highlightAll();</script>' >> CHANGES.html
 cat >> CHANGES.html << EOF
 <script>
 document.addEventListener('DOMContentLoaded', () => {
-const baseDocURL = 'https://sagemath.netlify.app'
+// This URL is hardcoded in the file .github/workflows/doc-publish.yml.
+// See NETLIFY_ALIAS of the "Deploy to Netlify" step.
+const baseDocURL = 'https://doc-develop--sagemath.netlify.app'
 const diffSite = 'https://pianomister.github.io/diffsite'
 const diffParagraphs = document.querySelectorAll('p.diff');
 diffParagraphs.forEach(paragraph => {
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -183,7 +183,8 @@ jobs:
           # Restore the new doc from changes by "wipe out"
           (cd doc && git checkout -f)
           # Sometimes rm -rf .git errors out because of some diehard hidden files
-          (cd doc && rm -rf .git 2>/dev/null && rm .gitattributes 2>/dev/null)
+          # So we simply move it out of the doc directory
+          (cd doc && mv .git ../git && mv .gitattributes ../gitattributes)
           mv CHANGES.html doc
           # Zip everything for increased performance
           zip -r doc.zip doc
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -122,7 +122,7 @@ jobs:
                      ${{ env.BUILD_IMAGE }} /bin/sh
 
       #
-      # On PRs and pushes to master or develop
+      # On PRs and pushes to develop
       #
 
       - name: Store old doc
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -13,7 +13,6 @@ on:
       - '[0-9]+.[0-9]+.rc[0-9]+'
       - '[0-9]+.[0-9]+.[0-9]+.rc[0-9]+'
     branches:
-      - master
       - develop
   workflow_dispatch:
     # Allow to run manually
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -4,6 +4,14 @@ on:
   pull_request:
   merge_group:
   push:
+    tags:
+      # Match all release tags including beta, rc
+      - '[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.beta[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+.beta[0-9]+'
+      - '[0-9]+.[0-9]+.rc[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+.rc[0-9]+'
     branches:
       - master
       - develop
@@ -34,7 +42,7 @@ env:
   EXTRA_CONFIGURE_ARGS:   --enable-fat-binary
 
 jobs:
-  build-docs:
+  build-doc:
     runs-on: ubuntu-latest
     services:
       # https://docs.docker.com/build/ci/github-actions/local-registry/
@@ -114,11 +122,13 @@ jobs:
                      --workdir $(pwd) \
                      ${{ env.BUILD_IMAGE }} /bin/sh
 
-      # Docs
+      #
+      # On PRs and pushes to master or develop
+      #
 
-      - name: Store old docs
+      - name: Store old doc
         id: worktree
-        if: (success() || failure()) && steps.container.outcome == 'success'
+        if: (success() || failure()) && steps.container.outcome == 'success' && !startsWith(github.ref, 'refs/tags/')
         run: |
           git config --global --add safe.directory $(pwd)
           git config --global user.email "ci-sage@example.com"
@@ -128,24 +138,24 @@ jobs:
           # mathjax path in new doc
           mathjax_path_to=$(docker exec -e SAGE_USE_CDNS=yes BUILD /sage/sage -python -c "from sage_docbuild.conf import mathjax_path; print(mathjax_path)")
           new_version=$(docker exec BUILD cat src/VERSION.txt)
-          mkdir -p docs/
-          docker cp BUILD:/sage/local/share/doc/sage/html docs/
-          # Wipe out chronic diffs of old doc against new doc
-          (cd docs && \
+          mkdir -p doc/
+          docker cp BUILD:/sage/local/share/doc/sage/html doc/
+          # Wipe out chronic diffs between old doc and new doc
+          (cd doc && \
            find . -name "*.html" | xargs sed -i -e '/class="sidebar-brand-text"/ s/Sage [0-9a-z.]* /Sage '"$new_version"' /' \
                                                 -e '/<link rel="stylesheet"/ s/?v=[0-9a-f]*"/"/' \
                                                 -e 's;'"$mathjax_path_from"';'"$mathjax_path_to"';' \
                                                 -e '\;<script type="application/vnd\.jupyter\.widget-state+json">;,\;</script>; d')
           # Create git repo from old doc
-          (cd docs && \
+          (cd doc && \
            git init && \
            (echo "*.svg binary"; echo "*.pdf binary") >> .gitattributes && \
            (echo ".buildinfo"; echo '*.inv'; echo '.git*'; echo '*.svg'; echo '*.pdf'; echo '*.png'; echo 'searchindex.js') > .gitignore; \
            git add -A && git commit --quiet -m "old")
 
-      - name: Build docs
+      - name: Build doc
         id: docbuild
-        if: (success() || failure()) && steps.worktree.outcome == 'success'
+        if: (success() || failure()) && steps.worktree.outcome == 'success' && !startsWith(github.ref, 'refs/tags/')
         # Always non-incremental because of the concern that
         # incremental docbuild may introduce broken links (inter-file references) though build succeeds
         run: |
@@ -157,50 +167,43 @@ jobs:
           ./config.status && make sagemath_doc_html-no-deps
         shell: sh .ci/docker-exec-script.sh BUILD /sage {0}
 
-      - name: Copy docs
+      - name: Copy doc
         id: copy
         if: (success() || failure()) && steps.docbuild.outcome == 'success'
         run: |
           set -ex
           # We copy everything to a local folder
-          docker cp BUILD:/sage/local/share/doc/sage/html docs
-          docker cp BUILD:/sage/local/share/doc/sage/index.html docs
-          (cd docs && git commit -a -m 'new')
+          docker cp --follow-link BUILD:/sage/local/share/doc/sage/html doc
+          docker cp --follow-link BUILD:/sage/local/share/doc/sage/index.html doc
+          (cd doc && git commit -a -m 'new')
           # Wipe out chronic diffs of new doc against old doc
-          (cd docs && \
+          (cd doc && \
            find . -name "*.html" | xargs sed -i -e '/This is documentation for/ s/ built with GitHub PR .*. Doc/. Doc/' \
                                                 -e '/<link rel="stylesheet"/ s/?v=[0-9a-f]*"/"/')
-          .ci/create-changes-html.sh $(cd docs && git rev-parse HEAD^) docs
+          .ci/create-changes-html.sh $(cd doc && git rev-parse HEAD^) doc
           # Restore the new doc from changes by "wipe out"
-          (cd docs && git checkout -f)
-          (cd docs && rm -rf .git)
-          mv CHANGES.html docs
-          # We also need to replace the symlinks because netlify is not following them
-          # CHECK IF STILL NEEDED
-          #cp -r -L $DOC_DIR ./docs
+          (cd doc && git checkout -f)
+          # Sometimes rm -rf .git errors out because of some diehard hidden files
+          (cd doc && rm -rf .git 2>/dev/null && rm .gitattributes 2>/dev/null)
+          mv CHANGES.html doc
           # Zip everything for increased performance
-          zip -r docs.zip docs
+          zip -r doc.zip doc
 
-      - name: Upload docs
+      - name: Upload doc
         id: upload
         if: (success() || failure()) && steps.copy.outcome == 'success'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: docs
-          path: docs.zip
+          name: doc
+          path: doc.zip
 
-      - name: Save space
-        id: savespace
-        if: (success() || failure()) && steps.upload.outcome == 'success' && github.repository == 'sagemath/sage' && github.ref == 'refs/heads/develop'
-        run: |
-          set -ex
-          # Save space on runner device before we start to build livedoc
-          rm -rf docs/
-          rm -f docs.zip
+      #
+      # On release tags: live doc and wheels
+      #
 
       - name: Build live doc
         id: buildlivedoc
-        if: (success() || failure()) && steps.savespace.outcome == 'success'
+        if: (success() || failure()) && startsWith(github.ref, 'refs/tags/')
         run: |
           export MAKE="make -j5 --output-sync=recurse" SAGE_NUM_THREADS=5
           export PATH="build/bin:$PATH"
@@ -227,7 +230,7 @@ jobs:
 
       - name: Upload live doc
         if: (success() || failure()) && steps.copylivedoc.outcome == 'success'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: livedoc
           path: livedoc.zip
--- a/.github/workflows/doc-publish.yml
+++ b/.github/workflows/doc-publish.yml
@@ -14,9 +14,11 @@ permissions:
   pull-requests: write
 
 jobs:
-  upload-docs:
+  publish-doc:
     runs-on: ubuntu-latest
     if: github.event.workflow_run.conclusion == 'success'
+    env:
+      CAN_DEPLOY: ${{ secrets.NETLIFY_AUTH_TOKEN != '' &&  secrets.NETLIFY_SITE_ID != '' }}
     steps:
       - name: Get information about workflow origin
         uses: potiuk/get-workflow-origin@v1_5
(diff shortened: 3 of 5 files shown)


.ci/create-changes-html.sh
```diff
@@ -19,7 +19,7 @@ echo '<script>hljs.highlightAll();</script>' >> CHANGES.html
 cat >> CHANGES.html << EOF
 <script>
 document.addEventListener('DOMContentLoaded', () => {
-const baseDocURL = 'https://sagemath.netlify.app'
+const baseDocURL = 'https://preview-release--sagemath.netlify.app'
```
mkoeppe: I don't understand the naming changes in this PR

kwankyu: We are uploading docs for a PR to, for example, https://preview-38220--sagemath.netlify.app, which is an "alias" of a non-human-readable url. 

On the other hand, docs for a release is uploaded to the production site https://sagemath.netlify.app, whose alias is, after this PR,  https://preview-release--sagemath.netlify.app. Before this PR, the alias was https://preview--sagemath.netlify.app but this alias url was not used.

Hence the naming change is just to make the url of the doc preview of the release look similar to the url for doc preview of a PR. But this change is not strictly necessary.

Do you prefer not to use the alias and keep using the shorter url?

kwankyu:
No problem. Then this test `github.ref == 'refs/heads/develop'` should be changed appropriately.

kwankyu:
Perhaps I missed the point of your question. Do you mean that with your change, docs (which is not livedoc) will not be uploaded on pushed release tags? This will break our "changes" mechanism totally...

kwankyu:
I see. So with your change, 'https://preview-release--sagemath.netlify.app' (that is 'https://sagemath.netlify.app') won't exist, only  'https://livedoc--sagemath.netlify.app' will.

I think I prefer to have  'https://preview-release--sagemath.netlify.app', which is better suited for comparison with PR doc preview (which is not livedoc)

.ci/create-changes-html.sh
kwankyu: I just checked that https://preview--sagemath.netlify.app does not exist...

This means that there is no alias for the production site. If so, the change on line 22 will not work...

kwankyu: There is a problem in using `--prod` together with `--alias`. That doesn't work as I described above. See https://answers.netlify.com/t/netlify-cli-deploy-with-alias-not-publishing/108047

Hence in https://github.com/sagemath/sage/pull/38220/commits/0fc6bbe46af30d381a3d8c1b463c717b9d9d1a08, I removed `--prod`.

Thus we will only have: https://preview-release--sagemath.netlify.app for doc preview for releases.

kwankyu: Then the production site https://sagemath.netlify.app remains unused. We may use it for other purpose. Perhaps we may deploy livedoc for stable versions there or replace current `https://livedoc--sagemath.netlify.app/ (livedoc for releases) with it. Something to consider later.

kwankyu: Or I can do it here if we agree on how to use `https://sagemath.netlify.app/`.

mkoeppe: Yes, let's try to figure it out, keeping also the following things in mind:
- this wishlist item: #37384;
- a possible future change in how the "develop" branch is operated, in which not only a beta or stable release is pushed to "develop";
- that with #36730, wheels of the livedoc are pushed to PyPI on beta or stable releases;
- possibly replacing doc.sagemath.org with the netlify deployment.

Some general thoughts: I think the word "preview" should only appear in what we build for PRs. Documentation that is built for a release tag (beta or stable), and documentation for the latest commit on the "develop" branch is not a "preview".

kwankyu: I think the doc built for a release tag (or on the "develop" branch) is still a "preview", since it is not an "official" deployment of the doc. Currently The deployment to https://doc.sagemath.org is of course official. The deployment to currently https://deploy-livedoc--sagemath.netlify.app (or https://livedoc--sagemath.netlify.app in future) might be called "official" too.

kwankyu: But I am okay with `release--sagemath.netlify.app` instead of `preview-release--sagemath.netlift.app`. If we plan to add `release-version--sagemath.netlify.app`, then prefix `preview` would look redundant.

mkoeppe:
Then let's do this bit, part of the change in #36730?

```diff
diff --git a/.github/workflows/doc-build.yml b/.github/workflows/doc-build.yml
index 9dfb8d7d221..57058ac8566 100644
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -4,6 +4,14 @@ on:
   pull_request:
   merge_group:
   push:
+    tags:
+      # Match all release tags including beta, rc
+      - '[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.beta[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+.beta[0-9]+'
+      - '[0-9]+.[0-9]+.rc[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+.rc[0-9]+'
     branches:
       - master
       - develop
@@ -114,11 +122,13 @@ jobs:
                      --workdir $(pwd) \
                      ${{ env.BUILD_IMAGE }} /bin/sh
 
-      # Docs
+      #
+      # On PRs and pushes to branches master/develop
+      #
 
       - name: Store old docs
         id: worktree
-        if: (success() || failure()) && steps.container.outcome == 'success'
+        if: (success() || failure()) && steps.container.outcome == 'success' && !startsWith(github.ref, 'refs/tags/')
         run: |
           git config --global --add safe.directory $(pwd)
           git config --global user.email "ci-sage@example.com"
@@ -145,7 +155,7 @@ jobs:
 
       - name: Build docs
         id: docbuild
-        if: (success() || failure()) && steps.worktree.outcome == 'success'
+        if: (success() || failure()) && steps.worktree.outcome == 'success' && !startsWith(github.ref, 'refs/tags/')
         # Always non-incremental because of the concern that
         # incremental docbuild may introduce broken links (inter-file references) though build succeeds
         run: |
@@ -182,9 +192,13 @@ jobs:
           name: docs
           path: docs.zip
 
+      #
+      # On release tags: live doc and wheels
+      #
+
       - name: Build live doc
         id: buildlivedoc
-        if: (success() || failure()) && steps.copy.outcome == 'success' && github.repository == 'sagemath/sage' && github.ref == 'refs/heads/develop'
+        if: (success() || failure()) && steps.container.outcome == 'success' && startsWith(github.ref, 'refs/tags/')
         run: |
           export MAKE="make -j5 --output-sync=recurse" SAGE_NUM_THREADS=5
           export PATH="build/bin:$PATH"
```

kwankyu:
But. Not harmful. OK.

mkoeppe:
For the tags, can we now arrange for the live doc to be deposited in versioned URLs?

kwankyu:
For comparison purpose with the docs for a PR, the regular (non-live) doc would be better.

regular doc or live doc for tags?

My suggestion: For tags, we may deploy regular doc to `release--sagemath.netlify.app` together with `release-version--sagemath.netlify.app`. And live doc for the latest stable release  to `sagemath.netlify.app`.

mkoeppe:
> For comparison purpose with the docs (non-live) for PRs, the regular (non-live) doc would be better.

You mean for preparing the diff, or for something else? The diff is always made against the documentation that is in the container image, and it unrelated to what is deployed to netlify, right?

kwankyu:
> > For comparison purpose with the docs (non-live) for PRs, the regular (non-live) doc would be better.
> 
> You mean for preparing the diff, or for something else? The diff is always made against the documentation that is in the container image, and it unrelated to what is deployed to netlify, right?

Right. 

But I use the deployed netlify doc site for online comparison in "changes.html" page.

mkoeppe:
OK, but that's not actually the correct version of the documentation to link, right? 

The container image is made on pushes to release tags, each of which is also pushed to `develop`. 
So for these also the regular doc is built because it is a push to the `develop` branch.

kwankyu:
> OK, but that's not actually the correct version of the documentation to link, right?

Why not? It is the latest release. 

> The container image is made on pushes to release tags, each of which is also pushed to `develop`. So for these also the regular doc is built because it is a push to the `develop` branch.

Ah, I see.

mkoeppe:
> For tags, we may deploy regular doc to `release--sagemath.netlify.app` together with `release-tag--sagemath.netlify.app`. And live doc for the latest stable release to `sagemath.netlify.app`.

In any case I would suggest to include the word "doc" in all of the URLs because nothing else in the URL explains that it is documentation.

- On push to the "develop" branch... perhaps deploy to doc-develop--sagemath.netlify.app
- On push to tags, deploy to livedoc-TAG--sagemath.netlify.app and additionally to livedoc-develop--sagemath.netlify.app; if a stable tag, additionally deploy to livedoc--sagemath.netlify.app

kwankyu:
> In any case I would suggest to include the word "doc" in all of the URLs because nothing else in the URL explains that it is documentation.

Another possibility is to change the domain name from "sagemath.netlify.app" to "sagemath-doc.netlify.app"or "sagedoc.netlify.app"

mkoeppe:
> Another possibility is to change the domain name from "sagemath.netlify.app" to "sagemath-doc.netlify.app"or "sagedoc.netlify.app"

I don't think that would be an improvement

kwankyu:
To me, `develop--sagemath-doc.netlify.app` seems better than `doc-develop--sagemath.netlify.app`. 

Anyway I am not strong about this though. 

`sagemath.netlify.app` is an "internal" ci service, and presence  or no presence of the word "doc" in the urls is not a big deal, I think.

mkoeppe:
For permanent deployment of doc, we can of course also look into using github.io, like the current doc.sagemath.org is deployed

mkoeppe:
Previous discussions on doc deployment: 
- https://github.com/sagemath/documentation/issues/24
- see also https://github.com/sagemath/sage/issues/37384

kwankyu:
> For permanent deployment of doc, we can of course also look into using github.io, like the current doc.sagemath.org is deployed

It seems necessary and useful to (P) keep some of previous docs for stable versions for some time. But I see no point of keeping the docs forever.

For (P), the current infrastructure seems enough.

kwankyu:
So do you insist:

doc-preview-PR_NUMBER--sagemath.netlify.app: for previewing docs of PRs.
doc-develop--sagemath.netlify.app: for push to develop and for online comparison with docs for PRs
livedoc-TAG--sagemath.netlify.app: for release (betas and rcs and stable versions)

?

mkoeppe:
> doc-preview-PR_NUMBER--sagemath.netlify.app: for previewing docs of PRs.

This one could also be "doc-pr-PR_NUMBER--sagemath.netlify.app". The "preview" does not do much for me, but I don't insist

kwankyu:
How about just "doc-PR_NUMBER--sagemath.netlify.app"?

kwankyu:
Also for "livedoc-TAG--sagemath.netlify.app", we may do just with "doc-TAG--sagemath.netlify.app". What do you think?

mkoeppe:
Fine with me to simplify "livedoc" -> "doc". 
Not sure about omitting "pr"

kwankyu:
OK. Let's go with

doc-pr-PR_NUMBER--sagemath.netlify.app: for previewing docs of PRs.
doc-develop--sagemath.netlify.app: for push to develop and for online comparison with docs for PRs
doc-TAG--sagemath.netlify.app: for release (betas and rcs and stable versions)

kwankyu:
Where should I put "pr" in
```
NETLIFY_ALIAS: doc-${{ github.ref == 'refs/heads/develop' && 'develop' || steps.source-run-info.outputs.pullRequestNumber }}
```
?

kwankyu:
Would this work?

```
          NETLIFY_ALIAS: doc-${{ github.ref.name }}  
```
for tags.

mkoeppe:
> Where should I put "pr"

Try https://docs.github.com/en/actions/learn-github-actions/expressions#format

kwankyu:
```
NETLIFY_ALIAS: doc-${{ github.ref == 'refs/heads/develop' && 'develop' || format('pr-{0}', steps.source-run-info.outputs.pullRequestNumber) }}

```
?

kwankyu:
For tags, according to chatgpt,
```
      - name: Get REF_NAME
        run: |
          echo "REF_NAME=${GITHUB_REF#refs/heads/}" >> $GITHUB_ENV
                                                               
```
and then
```
          NETLIFY_ALIAS: doc-${{ env.REF_NAME }}    
```
Is there a better way?

kwankyu:
```
  publish-live-doc:
    runs-on: ubuntu-latest
    if: github.event.workflow_run.conclusion == 'success' && github.repository == 'sagemath/sage' && github.event.workflow_run.head_branch == 'develop'
    steps:  
```
Here "if" condition should be changed?

kwankyu:
Testing in https://github.com/kwankyu/sage/actions/runs/9608481548

.ci/create-changes-html.sh
```diff
@@ -19,7 +19,7 @@ echo '<script>hljs.highlightAll();</script>' >> CHANGES.html
 cat >> CHANGES.html << EOF
 <script>
 document.addEventListener('DOMContentLoaded', () => {
-const baseDocURL = 'https://sagemath.netlify.app'
+const baseDocURL = 'https://doc-develop--sagemath.netlify.app'
```
mkoeppe, line 24: Add a similar comment here like you did in src/sage_docbuild/conf.py?

.github/workflows/doc-publish.yml
```diff
@@ -121,6 +120,10 @@ jobs:
       - name: Extract live doc
         run: unzip livedoc.zip -d doc && unzip doc/livedoc.zip -d doc/doc
 
+      - name: Get REF_NAME
+        run: |
+          echo "REF_NAME=${GITHUB_REF#refs/heads/}" >> $GITHUB_ENV
+
```
mkoeppe, line 131: but didn't we conditionalize on refs/tags?

.ci/create-changes-html.sh
kwankyu: Right. Thanks.

.github/workflows/doc-publish.yml
kwankyu: Right. Done. Could I do this without making another step "Get tag"?

kwankyu:
The testing didn't work as expected. There must be some bugs...

kwankyu:
In https://github.com/sagemath/sage/pull/38220/commits/e88fd79341a09f71e64e1203d99ec52e0a4dade5, I removed the check `github.repository == 'sagemath/sage'` for facilitate testing on my forked repo.

Do we need this check? The doc-publish will fail anyway if appropriate netlify secrets are not set. I do not see the point of the additional check.

mkoeppe:
Better to conditionalize on the presence of the secrets

kwankyu:
So why? "Skipping the workflow is better than running to fail" is it?

kwankyu:
Could we conditionalize on the presence of the secrets?

mkoeppe:
Yes, I do this for example in dist.yml for the PyPI secrets

kwankyu:
Do you worry that the secrets may be present for other purpose in the forked repo?

kwankyu:
> Yes, I do this for example in dist.yml for the PyPI secrets

Then okay that we do the same for doc-publish.yml?

mkoeppe:
Yes, I think that's the best solution.

kwankyu:
I will use 
```yml
    env:
      CAN_DEPLOY: ${{ secrets.NETLIFY_AUTH_TOKEN != '' &&  secrets.NETLIFY_SITE_ID != '' }} 

     ...
         if: env.CAN_DEPLOY == 'true'  
```

kwankyu:
> Better to conditionalize on the presence of the secrets

I was misreading your suggestion. Sorry, and thanks.

kwankyu:
Now this works well with me. Ready for review.

.github/workflows/doc-build.yml
```diff
@@ -114,11 +121,13 @@ jobs:
                      --workdir $(pwd) \
                      ${{ env.BUILD_IMAGE }} /bin/sh
 
-      # Docs
+      #
+      # On PRs and pushes to master or develop
```
mkoeppe: doesn't match the `on: push:` above

kwankyu:
```suggestion
      # On PRs and pushes to develop
```

kwankyu: Fixed. Thanks.

.github/workflows/doc-publish.yml
```diff
@@ -88,50 +79,54 @@ jobs:
...
-               artifact_id: matchArtifact.id,
-               archive_format: 'zip',
-            });
-            var fs = require('fs');
-            fs.writeFileSync('${{github.workspace}}/livedoc.zip', Buffer.from(download.data));
+          name: livedoc
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          repository: ${{ github.repository }}
+          run-id: ${{ github.event.workflow_run.id }}
+        if: steps.source-run-info.outputs.sourceEvent == 'push' && steps.source-run-info.outputs.targetBranch != 'develop'
```
mkoeppe, line 126: add a comment please that explains the purpose of this condition

.github/workflows/doc-publish.yml
```diff
@@ -14,54 +14,45 @@ permissions:
   pull-requests: write
 
 jobs:
-  upload-docs:
```
mkoeppe, line 17: How about adding a comment on the top of the file that gives an overview of the URLs deployed to (and by what event)

.github/workflows/doc-publish.yml
kwankyu: See https://github.com/sagemath/sage/pull/38220/commits/ea093f17c4791934cbef41d3837018358dad9285

.github/workflows/doc-publish.yml
kwankyu: OK. I will do it tomorrow. It is late at night here :-)

kwankyu:
Tests:

pr: https://github.com/kwankyu/sage/actions/runs/9635036620

push to `develop`: https://github.com/kwankyu/sage/actions/runs/9634977035

push tag: https://github.com/kwankyu/sage/actions/runs/9634868865

doc-pdf: https://github.com/kwankyu/sage/actions/runs/9634808127

.github/workflows/doc-publish.yml
kwankyu: Done in https://github.com/sagemath/sage/pull/38220/commits/a74cc417cc292f00b447c3415b77e371e63e11cc

kwankyu:
"Deploy Documentation" fails naturally because doc-build.yml and doc-publish.yml are not in harmony here.

.github/workflows/doc-publish.yml
mkoeppe: Thanks.

Aren't we missing ways to refer to the latest livedoc?

kwankyu: Good point. Let me think..

kwankyu: One simple solution is to deploy twice for tag push: once to "doc-tag--sagemath.netlify.app" and once again to the production site "sagemath.netlify.app".

kwankyu: It seems to be the only solution. As far as I see, netlify cli does not support multiple aliases.

kwankyu: With the simple solution, we have choices to make. For second deploymen,t we may use

(1) production site:  "sagemath.netlify.app".
(2) or draft site: "doc-latest--sagemath.netlify.app"
(3) or draft site with another name "doc-...--sagemath.netlify.app"

I like (2). Which do you prefer?

kwankyu: "doc-release" is not bad...

mkoeppe: "latest" typically means the latest _stable_ release

but where would the livedoc for the newest beta go?

kwankyu: "doc-release". See https://github.com/sagemath/sage/pull/38220/commits/501573ed9081eaa70c51267dbbd090ff2ae244da

mkoeppe:
> doc-build.yml and doc-publish.yml do not work in harmony here.

Just because of the upload/download-artifact versions or other changes too?

kwankyu:
I changed all "docs" to "doc" to be consistent with "livedoc".

"doc-build.yml" coming from this PR uploads "doc.zip" while "doc-publish.yml" coming from the develop branch downloads "docs.zip". So there is a mismatch.

mkoeppe:
a cleaner migration path would be to make doc-publish accept either version

kwankyu:
The friction will last briefly. Few PRs will be affected. I think "cleaner migration path" is unnecessary hassle.

.github/workflows/doc-publish.yml
kwankyu: It works: https://github.com/kwankyu/sage/actions/runs/9641539263

mkoeppe:
OK.

mkoeppe:
Let's merge it.

kwankyu:
Thanks.

Ideally CI should be much more robust than the target software. I expect that with this PR, our doc CI will be robust enough, at least for some time. But there are some warnings that some actions used in our CI will be deprecated soon, which would break CI again...

kwankyu:
Specifically, 

netlify/actions/cli@master:
```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
This is minor. Perhaps netlify will fix it.

potiuk/get-workflow-origin@v1_5:
```
[@octokit/rest] `const Octokit = require("@octokit/rest")` is deprecated. Use `const { Octokit } = require("@octokit/rest")` instead
```
According to the author, get-workflow-origin is not maintained. So this is more serious.

PR #37884, opened 2024-04-27
gap: support gap 4.13.1 in sagelib (does NOT touch sage-the-distro)

tornaria:
In gap 4.13 there are some improvements, e.g. converting fp groups to permutation groups, computing abelianization of fp groups, which lead to different generators.

This PR fixes doctests so they pass using gap 4.13.

I don't see a way to have this work with gap 4.12.2 at the same time, so this is not *in principle* ready to merge.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.

### :hourglass: Dependencies

#37883: don't use deprecated  `LaTeX()` gap function.

PR #38222, opened 2024-06-15
Reject system gap 4.13

mkoeppe:
### :memo: Checklist

### DIFF
--- a/build/pkgs/gap/distros/conda.txt
+++ b/build/pkgs/gap/distros/conda.txt
@@ -1,1 +1,1 @@
-gap-defaults>=4.12.2
+gap-defaults>=4.12.2,<4.13.0
--- a/build/pkgs/gap/spkg-configure.m4
+++ b/build/pkgs/gap/spkg-configure.m4
@@ -3,19 +3,21 @@ SAGE_SPKG_CONFIGURE([gap], [
   sage_spkg_install_gap=yes
 
   m4_pushdef([GAP_MINVER],["4.12.2"])
+  m4_pushdef([GAP_LTVER],["4.13.0"])
 
   SAGE_SPKG_DEPCHECK([ncurses readline zlib], [
     AC_PATH_PROG(GAP, gap)
     AS_IF([test -n "${GAP}"], [
-      AC_MSG_CHECKING([for gap version GAP_MINVER or newer])
-
-      # GAP will later add the "user" path to the list of root paths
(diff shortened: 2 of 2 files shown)


mkoeppe:
Thanks, Volker.

tornaria:
-1

Nothing here explains why this should be a blocker, not even why this should happen at all.

tornaria:
https://github.com/sagemath/sage/pull/37884#issuecomment-2183538972 shows that whatever issue you have is not about gap 4.13 but about a particular (broken?) install of gap.

mkoeppe:
> Nothing here explains why this should be a blocker, not even why this should happen at all.

Because we're trying to make a release, and your PR #37884 is not ready.

tornaria:
> > Nothing here explains why this should be a blocker, not even why this should happen at all.
> 
> Because we're trying to make a release, and your PR #37884 is not ready.

Come on, #37884 is 100% ready, blame your #38169. But note that updating 4.12.2 to 4.13.1 doesn't need anything special for us (https://github.com/void-linux/void-packages/pull/49571/files#diff-d5e8ac4c14235608656fe702205ba0e985de306ecc14756391643152370bfbd3).

mkoeppe:
Please @tornaria. As you know, #37884 is not in a shape to be merged because nobody implemented the proposed solutions to support both the old and the new version (see https://github.com/sagemath/sage/pull/37884#issuecomment-2081814558).

tornaria:
> Please @tornaria. As you know, #37884 is not in a shape to be merged because nobody implemented the proposed solutions to support both the old and the new version (see [#37884 (comment)](https://github.com/sagemath/sage/pull/37884#issuecomment-2081814558)).

Please @mkoeppe. As you know, #37884 is 100% ready to run sagemath with gap 4.13.0 or 4.13.1. The reason forhy #38169 is not ready escapes me.

tornaria:
... and there is nothing here that explains the reason to reject system gap 4.13.

mkoeppe:
@tornaria The Release Manager will simply not merge a PR such as #37884 that requires a newer version of a library than what is provided by the Sage distribution -- because he tests with the Sage distribution, the tests will fail, and he'll set it back to "needs work".

So whenever you want to prepare a PR that adds support for a newer version of a library, 
- either find a way to support both the old and the new version (including passing all tests);
- or combine the PR with the upgrade of the library.

But I'm pretty sure that you know all that. So what's with the "Come on" and "Please"?

PR #38228, opened 2024-06-16
`QuaternionAlgebra`: Unify ABC and factory

mkoeppe:
We make a change:
- from a UniqueFactory `QuaternionAlgebra` and abstract base class `QuaternionAlgebra_abstract`
- to an abstract base class `QuaternionAlgebra` with a `__classcall_private__` method

What is done here: We consider `QuaternionAlgebraFactory` internal and remove it outright. (+ docstring/doctest cosmetics)

Possible implementation variants:
- Minimally invasive change: Keep the UniqueFactory and delegate to it by `__classcall_private__`. (This was done in c2d2fca580953f544259fa08004744772c646d02)
- Change that maintains full compatibility: Add class methods `QuaternionAlgebra.create_key`, `QuaternionAlgebra.create_object` that delegate to the factory functions, so that calls to `create_key`, `create_object` continue to work without change
- This could be combined with deprecating `QuaternionAlgebraFactory`.

Also @saraedum (who introduced the use of `UniqueFactory` for this class)
- Does `UniqueFactory` offer any benefits over `CachedRepresentation`/`UniqueRepresentation` at all, or should we just replace uses of `UniqueFactory` by the more commonly used `CachedRepresentation`/`UniqueRepresentation`?

### :memo: Checklist

### DIFF
--- a/src/sage/algebras/quatalg/quaternion_algebra.py
+++ b/src/sage/algebras/quatalg/quaternion_algebra.py
@@ -20,7 +20,7 @@ TESTS:
 
 Pickling test::
 
-    sage: Q.<i,j,k> = QuaternionAlgebra(QQ,-5,-2)
+    sage: Q.<i,j,k> = QuaternionAlgebra(QQ, -5, -2)
     sage: Q == loads(dumps(Q))
     True
 """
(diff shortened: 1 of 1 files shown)


tscrim:
The factory is doing nothing special. +1 to converting it to the ABC, but it should also inherit from `UniqueRepresentation`  (or at least `CachedRepresentation`, but I don't see a need for it to have its own equality) to match the previous behavior.

mkoeppe:
@tscrim Do you have a specific opinion on the three "open questions" in the PR description?

tscrim:
I am for deprecating it. Although I can be convinced that could be removed outright (with an understanding that it is breaking our deprecation policy, but I think it can be justified in doing so here).

mkoeppe:
The failure in CI Conda is unrelated.

tscrim:
Would it be possible to have `QuaternionAlgebra_abstract` as a `lazy_import` of `QuaternionAlgebra` with a deprecation?

mkoeppe:
As discussed in https://github.com/sagemath/sage/pull/38207#discussion_r1638770236, importing such a `lazy_import` will work when this class is used in `isinstance`. But it won't work it the user tries to subclass that class.

tscrim:
However it will still emit a message if used, even if the code breaks on subclassing. That is better than it suddenly disappearing when the deprecation period is over.

PR #38235, opened 2024-06-17
Fix diagonal matrix construction from base ring elements that have `_matrix_` methods

### LINKED ISSUE #38221: Zero matrix over an algebra
### Steps To Reproduce

G = CyclicPermutationGroup(7);
R = GF(2);
A = G.algebra(R);
S = MatrixSpace(A, 3, 3);
S.zero_matrix()

### Expected Behavior

This should return the matrix [[0,0,0],[0,0,0],[0,0,0]]

### Actual Behavior

Cell In[11], line 1
----> 1 S.zero_matrix()

File /private/var/tmp/sage-10.3-current/local/var/lib/sage/venv-python3.11.8/lib/python3.11/site-packages/sage/misc/cachefunc.pyx:2329, in sage.misc.cachefunc.CachedMethodCallerNoArgs.__call__ (build/cythonized/sage/misc/cachefunc.c:16253)()
   2327 if self.cache is None:
   2328     f = self.f
-> 2329     self.cache = f(self._instance)
   2330 return self.cache
   2331 

File /private/var/tmp/sage-10.3-current/local/var/lib/sage/venv-python3.11.8/lib/python3.11/site-packages/sage/matrix/matrix_space.py:1999, in MatrixSpace.zero_matrix(self)
   1971 """
   1972 Return the zero matrix in ``self``.
   1973 
   (...)
   1996     False
   1997 """
   1998 zero = self.base_ring().zero()
-> 1999 res = self.element_class(self, zero, False, False)
   2000 res.set_immutable()
   2001 return res

File /private/var/tmp/sage-10.3-current/local/var/lib/sage/venv-python3.11.8/lib/python3.11/site-packages/sage/matrix/matrix_generic_dense.pyx:81, in sage.matrix.matrix_generic_dense.Matrix_generic_dense.__init__ (build/cythonized/sage/matrix/matrix_generic_dense.c:5528)()
     79     [x y]
     80 """
---> 81 ma = MatrixArgs_init(parent, entries)
     82 self._entries = ma.list(coerce)
     83 

File /private/var/tmp/sage-10.3-current/local/var/lib/sage/venv-python3.11.8/lib/python3.11/site-packages/sage/matrix/args.pyx:1390, in sage.matrix.args.MatrixArgs_init (build/cythonized/sage/matrix/args.c:21785)()
   1388     ret.entries = entries
   1389 ret.set_space(space)
-> 1390 ret.finalize()
   1391 return ret

File /private/var/tmp/sage-10.3-current/local/var/lib/sage/venv-python3.11.8/lib/python3.11/site-packages/sage/matrix/args.pyx:930, in sage.matrix.args.MatrixArgs.finalize (build/cythonized/sage/matrix/args.c:15152)()
    928 if self.typ == MA_ENTRIES_MATRIX:
    929     m = <Matrix>self.entries
--> 930     self.set_nrows(m._nrows)
    931     self.set_ncols(m._ncols)
    932     self.setdefault_base(m._parent._base)

File /private/var/tmp/sage-10.3-current/local/var/lib/sage/venv-python3.11.8/lib/python3.11/site-packages/sage/matrix/args.pxd:104, in sage.matrix.args.MatrixArgs.set_nrows (build/cythonized/sage/matrix/args.c:6653)()
    102 cdef long p = self.nrows
    103 if p != -1 and p != n:
--> 104     raise ValueError(f"inconsistent number of rows: should be {p} but got {n}")
    105 self.nrows = n
    106 

ValueError: inconsistent number of rows: should be 3 but got 7

### Additional Information

_No response_

### Environment

```markdown
- **OS**: macOS 14.5
- **Sage Version**: 10.3
```

### Checklist

mkoeppe:
The matrix constructors are designed to accept a scalar as the `entries` and then construct a diagonal matrix; special case: constructing the zero matrix from 0.

However, when the scalar is a base ring element that happens to have a `_matrix_` method, that is tried first, leading to the errors reported in #38221.
Here we change it so that if the base ring (or matrix space) has been specified already and a scalar from that base ring is passed, that takes precedence over using a `_matrix_` method. (Coercion is not attempted.)

For the special case of `zero_matrix` and `MatrixSpace.zero_matrix`, we now use `entries=None`, which is the fastest and most robust path in `MatrixArgs.entries_type` to designate a zero matrix.

Fixes #38221 

### :memo: Checklist

### DIFF
--- a/src/sage/matrix/special.py
+++ b/src/sage/matrix/special.py
@@ -1000,11 +1000,23 @@ def zero_matrix(ring, nrows=None, ncols=None, sparse=False):
         [0 0 0 0 0]
         [0 0 0 0 0]
 
+    TESTS:
+
+    Check that :issue:`38221` is fixed::
+
+        sage: # needs sage.groups
+        sage: G = CyclicPermutationGroup(7)
+        sage: R = GF(2)
+        sage: A = G.algebra(R)
+        sage: zero_matrix(A, 3, 3)
+        [0 0 0]
+        [0 0 0]
+        [0 0 0]
     """
     if isinstance(ring, (Integer, int)):
         nrows, ncols = (ring, nrows)
         ring = ZZ
-    return matrix_space.MatrixSpace(ring, nrows, ncols, sparse)(0)
+    return matrix_space.MatrixSpace(ring, nrows, ncols, sparse).matrix(None)
 
 
 @matrix_method
--- a/src/sage/matrix/args.pyx
+++ b/src/sage/matrix/args.pyx
@@ -1500,6 +1500,21 @@ cdef class MatrixArgs:
             Traceback (most recent call last):
             ...
             NameError: name 'a' is not defined
+
+        Check that :issue:`38221` is fixed::
+
+            sage: # needs sage.groups
+            sage: G = CyclicPermutationGroup(7)
+            sage: R = GF(2)
+            sage: A = G.algebra(R)
+            sage: matrix(A, 3, 3, A.zero())
+            [0 0 0]
+            [0 0 0]
+            [0 0 0]
+            sage: matrix(A, 3, 3, A.one())
+            [()  0  0]
+            [ 0 ()  0]
+            [ 0  0 ()]
         """
         # Check basic Python types. This is very fast, so it doesn't
         # hurt to do these first.
@@ -1524,6 +1539,8 @@ cdef class MatrixArgs:
         cdef bint is_elt = isinstance(self.entries, Element)
         if is_elt and isinstance(self.entries, Matrix):
             return MA_ENTRIES_MATRIX
+        if is_elt and self.base is not None and self.entries.parent() == self.base:
+            return MA_ENTRIES_SCALAR
         t = type(self.entries)
         try:
             f = t._matrix_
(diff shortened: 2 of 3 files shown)


src/sage/matrix/special.py
```diff
@@ -1000,11 +1000,23 @@ def zero_matrix(ring, nrows=None, ncols=None, sparse=False):
...
 
+    TESTS:
+
+    Check that :issue:`38221` is fixed::
+
+        sage: # needs sage.groups
+        sage: G = CyclicPermutationGroup(7)
+        sage: R = GF(2)
+        sage: A = G.algebra(R)
+        sage: zero_matrix(A, 3, 3)
```
tscrim, line 1011: This is the same test as in `matrix_space.py`. It would be better to have separate tests that better isolate the need for the changes. If they cannot be separated (which is quite possible), then there isn't a point in copying the test.

mkoeppe: It's not the same test. Here it is tests the global `zero_matrix` function, there it tests the `zero_matrix` method of `MatrixSpace`. Two completely separate code paths.

tscrim: Ah, right, sorry.

tscrim (approved):
LGTM.

mkoeppe:
Thank you!

PR #36641, opened 2023-11-03
`build/pkgs/sympy`: Upgrade to 1.13.2

oscarbenjamin:
See https://groups.google.com/g/sage-devel/c/fOTkNoU02Oo/m/j116LUouBwAJ

This upgrade includes doctest changes needed for compatibility with SymPy 1.13.

This PR is also used by SymPy CI to keep track of compatibility between SymPy and Sage:
https://github.com/sympy/sympy/blob/master/.github/workflows/ci-sage.yml
https://github.com/sympy/sympy/actions/runs/6724078875/job/18275557828

PR #38250, opened 2024-06-20
provide compatibility with numpy 2.0

### LINKED ISSUE #38242: using numpy 2.0
### Problem Description

it's released, and used by scipy

### Proposed Solution

various things to be done. The 1st is due to flint headers, which use `I` in function declarations, and if you include `complex.h`, you get problems. Fix is easy, and proposed in https://github.com/flintlib/flint/pull/2027

There are also deprecations to be dealt with, e.g.
```
File "src/sage/rings/polynomial/polynomial_element.pyx", line 4639, in sage.rings.polynomial.polynomial_element.Polynomial.factor
Failed example:
    (-2*x^2 - 1).factor()
Expected:
    (-2.0) * (x^2 + 0.5000000000000001)
Got:
    doctest:warning
      File "<doctest sage.rings.polynomial.polynomial_element.Polynomial.factor[44]>", line 1, in <module>
        (-Integer(2)*x**Integer(2) - Integer(1)).factor()
      File "/usr/lib/python3.12/site-packages/numpy/linalg/linalg.py", line 8, in __getattr__
        warnings.warn(
      File "/usr/lib/python3.12/warnings.py", line 112, in _showwarnmsg
        sw(msg.message, msg.category, msg.filename, msg.lineno,
    :
    DeprecationWarning: The numpy.linalg.linalg has been made private and renamed to numpy.linalg._linalg. All public functions exported by it are available from numpy.linalg. Please use numpy.linalg.LinAlgError instead.
    (-2.0) * (x^2 + 0.5000000000000001)
```

### Alternatives Considered

n/a

### Additional Information

_No response_

dimpase:
This will fix #38242

Basically, most of the problems are with numpy switching to a different way to display its data,
showing type. We either use a compatibility version, or add ellipses in doctests.

### :memo: Checklist

- [x ] I have created tests covering the changes.


dimpase:
Codecov complains about  the lines `numpy.set_printoptions(legacy="1.25")` not covered by tests.
Not sure how to suppress these, it looks bogus to me.

antonio-rojas:
I'm still getting a test failure:
```
File "/usr/lib/python3.12/site-packages/sage/matrix/matrix1.pyx", line 709, in sage.matrix.matrix1.Matrix.numpy
Failed example:
    sorted(numpy.typecodes.items())                                       # needs numpy
Expected:
    [('All', '?bhilqpBHILQPefdgFDGSUVOMm'), ('AllFloat', 'efdgFDG'),
     ('AllInteger', 'bBhHiIlLqQpP'), ('Character', 'c'), ('Complex', 'FDG'),
     ('Datetime', 'Mm'), ('Float', 'efdg'), ('Integer', 'bhilqp'),
     ('UnsignedInteger', 'BHILQP')]
Got:
    [('All', '?bhilqnpBHILQNPefdgFDGSUVOMm'),
     ('AllFloat', 'efdgFDG'),
     ('AllInteger', 'bBhHiIlLqQnNpP'),
     ('Character', 'c'),
     ('Complex', 'FDG'),
     ('Datetime', 'Mm'),
     ('Float', 'efdg'),
     ('Integer', 'bhilqnp'),
     ('UnsignedInteger', 'BHILQNP')]
```

dimpase:
that's one I don't quite know how to fix.
I guess we don't actually need to check for values of these things, it is too verbose

kiwifb:
Why is this even tested? I would understand if it was a numpy test, but a sage test?

kiwifb:
Nitpicking, the string "# to ensure numpy 2.0 compatibility" is applied inconsistently. I suggest for it to be attached to the "if" statement line. I guess it will be clear if an `import numpy` statement had to be explicitly added.

dimpase:
> Why is this even tested? I would understand if it was a numpy test, but a sage test?

it's not really a test, it's more like part of a guide. Useful, but testing all its output isn't so important. I added `#random` tag, and added one more example on how to show a part of it

antonio-rojas:
WFM now

dimpase:
> Nitpicking, the string "# to ensure numpy 2.0 compatibility" is applied inconsistently. I suggest for it to be attached to the "if" statement line. I guess it will be clear if an `import numpy` statement had to be explicitly added.

I hope you're fine with undoing all that once numpy 2 becomes standard, see #38275

dimpase:
> WFM now

thanks!

kiwifb:
> > Nitpicking, the string "# to ensure numpy 2.0 compatibility" is applied inconsistently. I suggest for it to be attached to the "if" statement line. I guess it will be clear if an `import numpy` statement had to be explicitly added.
> 
> I hope you're fine with undoing all that once numpy 2 becomes standard, see #38275

Well, I did not say to remove it altogether. But I absolutely know what you are saying. In time those compatibility hacks will go.

dimpase:
it was just a trivial rebase

tornaria:
This PR works fine for me, except there is a change in `src/sage/symbolic/expression.pyx` that conflicts with #36641. Is there a chance you can remove that hunk here? Whoever is using sympy 1.13 will need to apply #36641 regardless.

tornaria:
Note: I've tested this PR using numpy 1.26 and using numpy 2.0 and it works fine on both cases. See https://github.com/void-linux/void-packages/pull/49571#issuecomment-2237696959.

user202729:
I come across this while working on https://github.com/sagemath/sage/pull/39152.

I think while `np.set_printoptions(legacy="1.25")` in doctest is fine, setting that in the actual code is a bad idea — the new printing option is an improvement (in my opinion, because it makes a np.float64 prints differently from a Python `float`; and also evidently in numpy developers' opinion because they change it in newer versions)

Other options I can think of:

* use `.item()`, which given a numpy object returns the corresponding Python object. In both numpy 1.26 and 2.0 the printed out value are identical.
* set the compatibility option globally in doctesting — this way it wouldn't be necessary to repeat it in many docstrings.

tornaria:
> I come across this while working on #39152.
> 
> I think while `np.set_printoptions(legacy="1.25")` in doctest is fine, setting that in the actual code is a bad idea — the new printing option is an improvement (in my opinion, because it makes a np.float64 prints differently from a Python `float`; and also evidently in numpy developers' opinion because they change it in newer versions)
> 
> Other options I can think of:
> 
>     * use `.item()`, which given a numpy object returns the corresponding Python object. In both numpy 1.26 and 2.0 the printed out value are identical.
> 
>     * set the compatibility option globally in doctesting — this way it wouldn't be necessary to repeat it in many docstrings.

You are probably right.

With a caveat: sage-the-distro is still on numpy 1.26.3 afaict.

A better way would be to make sagelib depend on numpy at least 2.0, switch everything to the new format, and let distros (sage-the-distro and others) update-or-perish. See https://github.com/sagemath/sage/issues/39192#issuecomment-2564134030.

user202729:
not really (?) it's possible to make sage compatible with both 1.26 and 2.0 by writing doctests to be compatible, and a few other things.

of course the (only) disadvantage of that approach is maintenance overhead.

PR #38155, opened 2024-06-05
docstrings: large scale formatting

gmou3:
Large scale formatting of the docstrings in accordance with the [Developer Guide](https://doc.sagemath.org/html/en/developer/coding_basics.html).
The changes are done by applying `regex` rules followed by manual checks and edits.
The commits will be kept in manageable sizes (of ~100 changed files).

Similar preceding PR: #38145.

### Changes
- `True`, `False`, `None`, `self` to ` ``True`` `, ` ``False`` `, ` ``None`` `, ` ``self`` `
- `- arg (type, default: val) -- description` to `- ``arg`` -- type (default: val); description`
- First docstring sentences: enforce prescriptive form of initial verb and add missing full stops
- Remove full stops from single-sentence `INPUT` argument descriptions
- One-line `OUTPUT`s for short output descriptions
- More math (`<math>` to `` `<math>` `` and ` ``<math>`` `  to `` `<math>` `` )
- Remove redundant whitespace

### Extra changes (not reflected in the Guide)
- `non-negative`, `non-zero`, `non-positive` to `nonnegative`, `nonzero`, `nonpositive`
- `"<string>"` to `'<string>'`

### Partial reviews by commit
- [x] 01. 5219d62 (reviewer: @kwankyu)
- [x] 02. 37fe9ec (reviewer: @davidlowryduda)
- [x] 03. 1ed6ea5 (reviewer: @davidlowryduda)
- [x] 04. a6b0382 (reviewer: @davidlowryduda)
- [x] 05. c284b26 (reviewer: @davidlowryduda)
- [x] 06. 021184a (reviewer: @davidlowryduda)
- [x] 07. 713e80d (reviewer: @davidlowryduda)
- [x] 08. 53ea68f (reviewer: @davidlowryduda)
- [x] 09. 198e7ea (stash)
- [x] 10. f8b7351 (merge)
- [x] 11. 14a3682 (reviewer: @kwankyu)
- [x] 12. 9eab12f (reviewer: @kwankyu)
- [x] 13. 6c70e9e (reviewer: @kwankyu)
- [x] 14. c9f4492 (reviewer: @kwankyu)
- [x] 15. ad120b6 (merge)
- [x] 16. 75b4801 (reviewer: @kwankyu)
- [x] 17. 6922a4c (reviewer: @kwankyu)
- [x] 18. 0716b41 (reviewer: @kwankyu)
- [x] 19. 5f652ff (reviewer: @kwankyu)
[...]

PR #38254, opened 2024-06-21
Extend the method get_knotinfo of the Link class to cover non-prime knots

soehms:
This PR implements a task from the following code-comment in `link.py`:

>        # ToDo: extension to non prime links in which case an element of the monoid
>        # over :class:`KnotInfo` should be returned

In connection with this I switch the output of `get_knotinfo` to elements of the new class `FreeKnotInfoMonoid` for all knots (prime or not) For multi-component links I keep the output as it is, since the `FreeKnotInfoMonoid` does not make sence here.

Another thing changes: Currently, if the `unique=False` option is used symmetry mutants may be listed separately, even though they are known to be isotopic because their symmetry type is reversible or amphicheiral. Now, if symmetry mutants are known to be isotopic because of their symmetry type, just the minimal of them is listed (in the order of the enum).

### :memo: Checklist


soehms:
All bots green!

soehms:
## Test `get_knotinfo` against the entire database

I haven't done this since the additional 13-crossing knots appeared, so I've done this now. On the first pass, I noticed an inconsistency between the results of the new `from_knot` method of the new monoid class and the existing `get_knotinfo` method of the `Link` class. This was fixed in the previous commit (see the test with `K12a_165`).

After that five knots are left, that cannot be recovered (non uniquely) by `get_knotinfo`:

```
sage: %time ll = [L for L in KnotInfo if not L.is_recoverable(unique=False)]
CPU times: user 2h 45min 38s, sys: 1.65 s, total: 2h 45min 39s
Wall time: 2h 45min 40s
sage: ll
[<KnotInfo.K13a_260: '13a_260'>,
 <KnotInfo.K13a_1805: '13a_1805'>,
 <KnotInfo.K13a_3534: '13a_3534'>,
 <KnotInfo.K13n_615: '13n_615'>,
 <KnotInfo.K13n_618: '13n_618'>]
```

But there must be something wrong with these in the database. All five are recorded as chiral knots:

```
sage: for K in ll:
....:     print(K, K.symmetry_type())
....:
KnotInfo.K13a_260 chiral
KnotInfo.K13a_1805 chiral
KnotInfo.K13a_3534 chiral
KnotInfo.K13n_615 chiral
KnotInfo.K13n_618 chiral
```

If this is correct, the `braid_notation` must be wrong, because it is the reverse of the `pd_notation`:

```
sage: for K in ll:
....:     print(K, K.link(use_item=K.items.braid_notation).get_knotinfo(unique=False))
....:
KnotInfo.K13a_260 [KnotInfo['K13a_260r']]
KnotInfo.K13a_1805 [KnotInfo['K13a_1805r']]
KnotInfo.K13a_3534 [KnotInfo['K13a_1805c'], KnotInfo['K13a_1805m'], KnotInfo['K13a_3534'], KnotInfo['K13a_3534r']]
KnotInfo.K13n_615 [KnotInfo['K13n_615r']]
KnotInfo.K13n_618 [KnotInfo['K13n_618r']]
```

This can be traced back to the conjugation of braids:

```
sage: a = KnotInfo.K13a_260.link().braid()
sage: b = KnotInfo.K13a_260.braid()
sage: a.is_conjugated(b)
False
sage: a.is_conjugated(b.reverse())
True
```

```
sage: a = KnotInfo.K13a_3534.link().braid()
sage: b = KnotInfo.K13a_3534.braid()
sage: a.is_conjugated(b)
False
sage: a.is_conjugated(b.reverse())
True
sage: am = a.mirror_image()
sage: bm = b.mirror_image()
sage: am.is_conjugated(bm)
False
sage: am.is_conjugated(bm.reverse())
True
```


I will clarify this with Chuck Livingston.

soehms:
> I will clarify this with Chuck Livingston.

There are much more cases with the above problem:

```
sage: %time mm = [K for K in KnotInfo if K.is_knot() and K.symmetry_type() == 'chiral' and K.link().braid().is_conjugated(K.braid().reverse())]
CPU times: user 22.6 s, sys: 30 µs, total: 22.6 s
Wall time: 22.6 s
sage: len(mm)
692
```

I didn't notice them before because they were better hidden by the original `get_knotinfo` implementation, which is still used for irreducible HOMFLY polynomials. More precisely, that implementation tries to identify the knot directly by `pd_code`, which has not been the case with the `from_knot` method of `FreeKnotInfoMonoid` until the previous commit. The smallest example in the above list `mm` is this:

```
sage: K = KnotInfo.K9_33
sage: b1 = K.braid(); b1
s0^-1*s1*s0^-1*s2^-1*s1^2*s0^-1*s1*s2^-1
sage: b2 = K.link().braid(); b2
(s0^-1*s1)^2*s1*s0^-1*s2^-1*s1*s2^-1
sage: b1.is_conjugated(b2)
False
sage: b1.is_conjugated(b2.reverse())
True
```

At the moment it is unclear to me whether there is incorrect data in the database or whether the Sage method `braid` returns reversed braids in these cases. However, this problem is independent of the changes made here.

src/sage/knots/free_knotinfo_monoid.py
```diff
@@ -0,0 +1,479 @@
...
+
+            sage: from sage.knots.free_knotinfo_monoid import FreeKnotInfoMonoid
+            sage: FKIM =  FreeKnotInfoMonoid()
+            sage: K = KnotInfo.K5_1.link().mirror_image()
+            sage: FKIM._from_knot(K)
+            (KnotInfo['K5_1m'],)
+        """
+        from sage.knots.knotinfo import KnotInfo
+
+        def search_composition(max_cr, hpoly):
```
tscrim: I think this would be better as a (hidden) method or function.

src/sage/knots/free_knotinfo_monoid.py
```diff
@@ -0,0 +1,479 @@
...
+                else:
+                    return[k]  # to be consistent with get_knotinfo
+
+        if res and not unique:
+            return sorted(list(set(res)))
+        if unique and len(res) > 1:
+            non_unique_hint = '\nuse keyword argument `unique` to obtain more details'
+            raise NotImplementedError('this (possibly non prime) knot cannot be identified uniquely by KnotInfo%s' % non_unique_hint)
+        else:
+            raise NotImplementedError('this (possibly non prime) knot cannot be identified by KnotInfo')
```
tscrim:
```suggestion
        raise NotImplementedError('this (possibly non prime) knot cannot be identified by KnotInfo')
```
Personal preference, but I don't like such `else:` statements to end as it could be that one of the `if` statements (in the future) accidentally falls through resulting in the method returning `None`.

src/sage/knots/free_knotinfo_monoid.py
```diff
@@ -0,0 +1,479 @@
...
+            [(<KnotInfo.K3_1: '3_1'>, <SymmetryMutant.itself: 's'>),
+            (<KnotInfo.K3_1: '3_1'>, <SymmetryMutant.itself: 's'>),
+            (<KnotInfo.K3_1: '3_1'>, <SymmetryMutant.mirror_image: 'm'>)]
+        """
+        wl = self.to_word_list()
+        P = self.parent()
+        return [P._index_dict[w] for w in wl]
+
+
+class FreeKnotInfoMonoid(IndexedFreeAbelianMonoid):
```
tscrim, line 92: Perhaps this should also inherit from `Singleton`?

src/sage/knots/free_knotinfo_monoid.py
```diff
@@ -0,0 +1,479 @@
...
+            sage: from sage.knots.free_knotinfo_monoid import FreeKnotInfoMonoid
+            sage: FKIM =  FreeKnotInfoMonoid()
+            sage: FKIM4 =  FreeKnotInfoMonoid(4)
+            sage: TestSuite(FKIM).run()
+            sage: TestSuite(FKIM4).run()
+        """
+        self._max_crossing_number = None
+        self._set_index_dictionary(max_crossing_number=max_crossing_number)
+        from sage.sets.set import Set
+        indices = Set(self._index_dict.keys())
```
tscrim: Not really important but does it need to be a Sage `Set`? I have an instinctual desire to avoid using them in general.

tscrim (changes requested):
Just a few minor things; LGTM otherwise.

soehms:
@tscrim Thanks for reviewing this. I will have a look at your other points later on (since I caught an influenza with fever).

tscrim:
No problem; no rush. I hope you feel better soon!

soehms:
> No problem; no rush. I hope you feel better soon!

Thank you! It was Corona, but now I think it's over.

src/sage/knots/free_knotinfo_monoid.py
soehms: Yes, because `IndexedMonoid` uses `cardinality` in `__init__`.

src/sage/knots/free_knotinfo_monoid.py
soehms: This is fixed by the current commit.

src/sage/knots/free_knotinfo_monoid.py
soehms: This is fixed by the current commit.

src/sage/knots/free_knotinfo_monoid.py
tscrim: Ah, I see. Then it is okay (an alternative might be `FiniteEnumeratedSet`). Do you also need the `.keys()` here too?

soehms: > an alternative might be FiniteEnumeratedSet

Yes, that works. Done in the current commit.

> Do you also need the .keys() here too?

Yes for easier translation, for example in the `to_knotinfo` method.

tscrim (approved):
I am glad you are feeling better. Greetings from Frankfurt.

src/sage/knots/free_knotinfo_monoid.py
tscrim: I don't quite follow with the `keys()`. A `dict` iterates over its keys and it is common Python practice to not have such calls. Well, if you prefer it, then we can leave it in, but don't be surprised if someone else comes and removes it in the future. Once you decide what you want to do (remove it or keep it), you can set a positive review.

soehms: Sorry, I completely misunderstood that. Of course I can remove the `.keys()` here.

soehms:
>  Greetings from Frankfurt.

Nice! How long will you stay?

tscrim:
> > Greetings from Frankfurt.
> 
> Nice! How long will you stay?

I was just there for 5 hours while waiting for my flight. (I was in Bochum last week for FPASC.)

src/sage/knots/free_knotinfo_monoid.py
tscrim: Thank you.

soehms:
> > > Greetings from Frankfurt.
> > 
> > 
> > Nice! How long will you stay?
> 
> I was just there for 5 hours while waiting for my flight. (I was in Bochum last week for FPASC.)

I see! I hope you enjoyed it!

src/sage/knots/free_knotinfo_monoid.py
soehms: Thank you, too.

soehms:
I just solved a trivial merge conflict. Thus setting back to positive review!

soehms:
CI shows a problem in [this run](https://github.com/sagemath/sage/actions/runs/10849990582/job/30110379688?pr=38254#step:16:48), probably an incompatibility after merging:

```
sage -t --long --random-seed=325380365030325599771092817287827401232 src/sage/knots/free_knotinfo_monoid.py
Error: Failed example:: Exception raised:
  File "/sage/src/sage/doctest/forker.py", line 714, in _run
  File "/sage/src/sage/doctest/forker.py", line 1144, in compile_and_execute
    exec(compiled, globs)
  File "<doctest sage.knots.free_knotinfo_monoid.FreeKnotInfoMonoidElement.as_knot[1]>", line 1, in <module>
    FKIM =  FreeKnotInfoMonoid()
  File "sage/misc/classcall_metaclass.pyx", line [32](https://github.com/sagemath/sage/actions/runs/10849990582/job/30110379688?pr=38254#step:16:33)1, in sage.misc.classcall_metaclass.ClasscallMetaclass.__call__
    return cls.classcall(cls, *args, **kwds)
  File "/sage/src/sage/knots/free_knotinfo_monoid.py", line 109, in __classcall_private__
    return super().__classcall__(cls, max_crossing_number, prefix=prefix, **kwds)
  File "/sage/src/sage/monoids/indexed_free_monoid.py", line 722, in __classcall__
    return super().__classcall__(cls, indices, prefix,
  File "sage/misc/fast_methods.pyx", line 276, in sage.misc.fast_methods.Singleton.__classcall__
    @staticmethod
TypeError: __classcall__() takes exactly 1 positional argument (3 given)
```

soehms:
> CI shows a problem in [this run](https://github.com/sagemath/sage/actions/runs/10849990582/job/30110379688?pr=38254#step:16:48), probably an incompatibility after merging:
> 
@tscrim, I need your help here. According to [your suggestion on July 15th](https://github.com/sagemath/sage/pull/38254#discussion_r1677247098) I added an inheritance to `Singleton` in [this commit](https://github.com/sagemath/sage/pull/38254/commits/631a112c182e1dcac318c0616537d6a10c9c722c) on July 25th. This commit has been relative to 10.5.beta0. At that time all tests passed. But, after I merged with 10.5.beta3 this is failing, now.

It took me some time to find [the commit which causes this](https://github.com/sagemath/sage/pull/38254/commits/631a112c182e1dcac318c0616537d6a10c9c722c). It was authored by @enriqueartal in connection with #38155 and changes the inheritance order in `UniqueRepresentation`. If you revert that change test-wise, the failure disappears:

```
diff --git a/src/sage/structure/unique_representation.py b/src/sage/structure/unique_representation.py
index c7bb3783658..cc843e9ac9e 100644
--- a/src/sage/structure/unique_representation.py
+++ b/src/sage/structure/unique_representation.py
@@ -1201,7 +1201,7 @@ class CachedRepresentation(WithPicklingByInitArgs):
             del cache[k]


-class UniqueRepresentation(WithEqualityById, CachedRepresentation):
+class UniqueRepresentation(CachedRepresentation, WithEqualityById):
     r"""
     Classes derived from ``UniqueRepresentation`` inherit a unique
     representation behavior for their instances
```

At the moment I have no idea to fix that properly. Of course one option would be to remove the inheritance from `Singleton` again. As far as I have seen, it did not really improve performance. What do you think?

tscrim:
Hmm…interesting. I am a bit surprised by this, but it might be a good thing that this is showing up. I might need to look at it much more closely later.

One thing to do would be to put the `Singleton` class first in the MRO, but that probably won’t fix the problem. The issue seems to be that it is calling the `__classcall__` of a base class, which takes limited arguments. As such, it likely needs to pass along `*args, **kwds`, or you deliberately skip it by replacing the `super()` with a more explicit base class.

enriqueartal:
This change was caused by a change of how to uniquely associate a GAP group to a Sage group. The MRO compiicated everything and the commit was done to solve it, from a suggestion from @tscrim . I completely trust @tscrim for possible solutions.

soehms:
> One thing to do would be to put the `Singleton` class first in the MRO, but that probably won’t fix the problem. The issue seems to be that it is calling the `__classcall__` of a base class, which takes limited arguments. As such, it likely needs to pass along `*args, **kwds`, or you deliberately skip it by replacing the `super()` with a more explicit base class.


Many thanks to have a look at it! I have experimented with your suggestions. If I combine your first point with the second, I get:

```
Failed example:
    FKIM =  FreeKnotInfoMonoid()
Exception raised:
      File "/home/sebastian/devel/sage/src/sage/doctest/forker.py", line 716, in _run
      File "/home/sebastian/devel/sage/src/sage/doctest/forker.py", line 1146, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.knots.free_knotinfo_monoid.FreeKnotInfoMonoidElement.as_knot[1]>", line 1, in <module>
        FKIM =  FreeKnotInfoMonoid()
      File "sage/misc/classcall_metaclass.pyx", line 321, in sage.misc.classcall_metaclass.ClasscallMetaclass.__call__
        return cls.classcall(cls, *args, **kwds)
      File "/home/sebastian/devel/sage/src/sage/knots/free_knotinfo_monoid.py", line 109, in __classcall_private__
        return super().__classcall__(cls, max_crossing_number, prefix=prefix, **kwds)
      File "sage/misc/fast_methods.pyx", line 298, in sage.misc.fast_methods.Singleton.__classcall__
        res = typecall(cls)
      File "sage/misc/classcall_metaclass.pyx", line 472, in sage.misc.classcall_metaclass.typecall
        return (<PyTypeObject*>type).tp_call(cls, args, kwds)
    TypeError: FreeKnotInfoMonoid.__init__() missing 1 required positional argument: 'max_crossing_number'
```

If I combine your first point with the third I get:

```
Failed example:
    TestSuite(FKIM).run()
Expected nothing
Got:
      Failure in _test_pickling:
        File "/home/sebastian/devel/sage/src/sage/misc/sage_unittest.py", line 298, in run
          test_method(tester=tester)
        File "sage/structure/sage_object.pyx", line 679, in sage.structure.sage_object.SageObject._test_pickling
          tester.assertEqual(loads(dumps(self)), self)
        File "sage/misc/persist.pyx", line 990, in sage.misc.persist.loads
          ans = unpickler.load()
        File "sage/misc/classcall_metaclass.pyx", line 321, in sage.misc.classcall_metaclass.ClasscallMetaclass.__call__
          return cls.classcall(cls, *args, **kwds)
        File "sage/misc/fast_methods.pyx", line 297, in sage.misc.fast_methods.Singleton.__classcall__
          assert cls.mro()[1] == Singleton, "{} is not a direct subclass of {}".format(cls, Singleton)
      AssertionError: <class 'sage.knots.free_knotinfo_monoid.FreeKnotInfoMonoid_with_category'> is not a direct subclass of <class 'sage.misc.fast_methods.Singleton'>
```

soehms:
> This change was caused by a change of how to uniquely associate a GAP group to a Sage group. The MRO compiicated everything and the commit was done to solve it, from a suggestion from @tscrim . 

Thanks for the info!

> I completely trust @tscrim for possible solutions.

I do so, too!

tscrim:
Sorry for being slow to respond (1 week old baby and public holidays). I need to experiment with this to figure things out. I hope to do so this week
 Keep pestering me if I don't.

tscrim:
> > 1 week old baby ...
> 
> That's really great news, Travis! Congratulations!
> 
> My kids are already adults and live more than 200 km away (both happen to be in Munich). But I still remember very well how radically their birth changed life (especially in the first case).

Thank you. Indeed, although this is the second, so we are a bit more seasoned parents.

> How about this plan:
> 
>     * We remove `Singleton` from the class again and merge this PR.
>     * Then I open a follow-up PR that adds `Singleton` back in. 
> 
> This would remove the time pressure and reduce the likelihood of further merge conflicts. What do you think?

-1 Let's get this right the first time. I won't need much time, but I just need to play around a bit see what is happening.

soehms:
> Thank you. Indeed, although this is the second, so we are a bit more seasoned parents.
> 

Yes, the second should be more relaxed (at least in the beginning, before they have their own personality).
 
> -1 Let's get this right the first time. I won't need much time, but I just need to play around a bit see what is happening.

No problem!

tscrim:
This is actually independent of the change in #38155, which somehow just made the bug appear. There are two issues:

1. This should not inherit from `Singleton`, which is for classes that have precisely 1 instance. So by definition, it cannot take parameters. Contrast this with `UniqueRepresentation`, which is 1 instance for each (normalized) input.
2. The generic `IndexedMonoid.__classcall__` was designed with a particular API with the goal to avoid code duplication. We need to work around this.

Here is a diff with the changes that I was able to get the code work:
```diff
diff --git a/src/sage/knots/free_knotinfo_monoid.py b/src/sage/knots/free_knotinfo_monoid.py
index acf286fba0b..c8301a4d138 100644
--- a/src/sage/knots/free_knotinfo_monoid.py
+++ b/src/sage/knots/free_knotinfo_monoid.py
@@ -27,7 +27,7 @@ AUTHORS:
 from sage.knots.knotinfo import SymmetryMutant
 from sage.monoids.indexed_free_monoid import IndexedFreeAbelianMonoid, IndexedFreeAbelianMonoidElement
 from sage.misc.cachefunc import cached_method
-from sage.misc.fast_methods import Singleton
+from sage.structure.unique_representation import UniqueRepresentation
 
 
 class FreeKnotInfoMonoidElement(IndexedFreeAbelianMonoidElement):
@@ -89,7 +89,7 @@ class FreeKnotInfoMonoidElement(IndexedFreeAbelianMonoidElement):
         return [P._index_dict[w] for w in wl]
 
 
-class FreeKnotInfoMonoid(IndexedFreeAbelianMonoid, Singleton):
+class FreeKnotInfoMonoid(IndexedFreeAbelianMonoid):
 
     Element = FreeKnotInfoMonoidElement
 
@@ -106,12 +106,15 @@ class FreeKnotInfoMonoid(IndexedFreeAbelianMonoid, Singleton):
             sage: FreeKnotInfoMonoid(5)
             Free abelian monoid of knots with at most 5 crossings
         """
-        return super().__classcall__(cls, max_crossing_number, prefix=prefix, **kwds)
+        if not prefix:
+            prefix = 'KnotInfo'
+        # We skip the IndexedMonoid__classcall__
+        return UniqueRepresentation.__classcall__(cls, max_crossing_number, prefix=prefix, **kwds)
 
     def __init__(self, max_crossing_number, category=None, prefix=None, **kwds):
         r"""
-        Init this monoid with generators belonging to prime knots with at most
-        ``max_crossing_number`` crossings.
+        Initialize ``self`` with generators belonging to prime knots with
+        at most ``max_crossing_number`` crossings.
 
         TESTS:
 
@@ -125,8 +128,6 @@ class FreeKnotInfoMonoid(IndexedFreeAbelianMonoid, Singleton):
         self._set_index_dictionary(max_crossing_number=max_crossing_number)
         from sage.sets.finite_enumerated_set import FiniteEnumeratedSet
         indices = FiniteEnumeratedSet(self._index_dict)
-        if not prefix or prefix == 'F':
-            prefix = 'KnotInfo'
         super().__init__(indices, prefix)
 
     def _from_knotinfo(self, knotinfo, symmetry_mutant):
```
Note the normalization of the `prefix` should happen in the `__classcall__`.

tscrim:
Green bot => setting positive review.

soehms:
> Green bot => setting positive review.

Thanks!

PR #38259, opened 2024-06-22
Fixed and improvements in `is_LLL_reduced` and `approximate_closest_vector`

### LINKED ISSUE #38115: Reduce usage of `sage.modules.misc.gram_schmidt` in lattice-related methods
### Problem Description

When dealing with lattice bases I often find myself using ones which have a lower rank than the number of rows. This is not an issue in most circumstances, but calculating Gram-Schmidt orthogonalization fails since the rows are linearly dependent.

`sage.matrix.matrix_integer_dense.Matrix_integer_dense.is_LLL_reduced` uses `sage.modules.misc.gram_schmidt` internally which is as mentioned prone to errors and is also very slow for large lattices.

`sage.modules.misc.gram_schmidt` is also used in `sage.modules.free_module_integer.IntegerLattice.approximate_closest_vector` which thus suffers from the same problems.

### Proposed Solution

For `is_LLL_reduced` I suggest we instead use fpylll's method `LLL.is_reduced`. It also accepts delta and eta arguments and does not complain when there are some linearly dependent vectors. It is also way faster in my experience.

For `approximate_closest_vector` I think implementing something akin to https://github.com/fplll/fpylll/blob/master/src/fpylll/algorithms/babai.py would be better (we don't want to use that one directly, I think it has issues with precision in the norm). I have not researched what the bounds of this algorithm is.

### Alternatives Considered

We could try to speed up `gram_schmidt` and make it more resilient, but I do not think it is worth the effort. #8143 also mentioned speeding up this function.

### Additional Information

_No response_

TheBlupper:
This fixes #38115. Along with the goal stated there I have fixed some stuff I wasn't happy with from the first iteration of `approximate_closest_vector`. These changes could be breaking but seeing as there has been no release since its introduction this shouldn't be a problem.

Changes:
 - Added the `algorithm` parameter to `is_LLL_reduced` along with the new algorithm `fpLLL` which uses fpylll's `LLL.is_reduced` method, **and made it the default**, as I see no advantages with the old implementation.
 - Added the `algorithm` parameter to `approximate_closest_vector` along with the new algorithms `embedding` and `rounding_off`, **and made `embedding` the default.**
 - ~Changed the alias of `approximate_closest_vector` from `babai` to `cvp` to be more accurate but still short.~

I'm getting weird errors when building documentation locally but I don't think it's related to my changes, so I haven't been able to check if the doc build looks alright yet. (and it won't build on my own repo b/c my GitHub username has uppercase letters in it :)

### :memo: Checklist


mkoeppe:
Lint failures

src/sage/modules/free_module_integer.py
```diff
@@ -867,12 +877,39 @@ def approximate_closest_vector(self, t, delta=None, *args, **kwargs):
...
+            # if it does not exist, same effect as using the psuedo-inverse
+            return vector(ZZ,
+                [QQ(x).round('even') for x in (B*B.T).solve_left(t*B.T)])*B
+
+        else:
+            raise ValueError("algorithm must be one of 'embedding', 'nearest_plane' or 'rounding_off'")
 
-    babai = approximate_closest_vector
\ No newline at end of file
+    cvp = approximate_closest_vector
```
mkoeppe: This will need deprecation

TheBlupper: Is this because there has been a release candidate since this functions introduction or do we generally require deprecations for changes within the beta stage?

mkoeppe: Yes, it's unlikely that this PR will make it in the 10.4 release. The release manager typically only merges "blocker" PRs in this stage.

TheBlupper: Gotcha, how does one deprecate an alias like this one? I also see now that the alias gets its own entry in the reference which feels a little superfluous.

TheBlupper: Okay I opted for not changing the alias since it was a minor nitpick anyway, I think "babai" is reasonable (all the algorithms arguably in some way stem from him). I also added a small function body around the alias which redirects to the main one so as not to duplicate the entire docstring in the reference manual.

TheBlupper:
The code coverage that's missing feels a bit silly, should I still create tests for them? (one of them is unreachable so idk)

src/sage/matrix/matrix_integer_dense.pyx
```diff
@@ -3330,20 +3341,27 @@ cdef class Matrix_integer_dense(Matrix_dense):
...
-        return True
+        if algorithm == 'fpLLL':
+            from fpylll import LLL, IntegerMatrix
+            A = IntegerMatrix.from_matrix(self)
+            return LLL.is_reduced(A, delta=delta, eta=eta)
+        elif algorithm == 'sage':
+            # this is pretty slow
+            import sage.modules.misc
+            G, mu = sage.modules.misc.gram_schmidt(self.rows())
+            #For any $i>j$, we have $|mu_{i, j}| <= \eta$
```
GiacomoPope: Would be nice to have some formatting on comments

```
# Something like this
```

```
#instead of this
```

GiacomoPope (approved):
Small comment about comment formatting, otherwise good

GiacomoPope (approved):

PR #38269, opened 2024-06-24
Fix lex_BFS (and co.) for directed graphs

### LINKED ISSUE #38234: lex_BFS method of DiGraph is not well-defined (the method could compute ordering that fails doctests)
### Steps To Reproduce

Load the content of the function `_is_valid_lex_BFS_order` from `graphs/traversals.pyx`, then run
```
sage: _is_valid_lex_BFS_order(DiGraph([(0,1), (0,2), (1, 2)]), [1,0,2])
sage: _is_valid_lex_BFS_order(Graph([(0,1), (0,2), (1, 2)]), [1,0,2])
```

### Expected Behavior

```
True
True
```

### Actual Behavior

```
False
True
```

### Additional Information

As the code is written, it is not clear what is the semantic of a lexBFS ordering for directed graphs (DiGraph).

In the `lex_BFS` method in `graphs/traversals.pyx`, the graph is first converted to a simple **undirected graph**, via the hidden default parameter `to_undirected=True` of the call to the `to_simple` method of those lines
```py
if G.has_loops() or G.has_multiple_edges():
  G = G.to_simple(immutable=False)
```
It seems to indicate that directed graphs are considered as undirected for lexBFS.
The doctest of the function strengthens this impression
```py
r"""
    Lex BFS ordering of a symmetric digraph should be the same as the Lex BFS
    ordering of the corresponding undirected graph::
"""
```

But, in the function `_is_valid_lex_BFS_order` that is used in doctests, directed graphs are considered directed, as shown by the use of the methods `G.neighbor_in_iterator` and `G.has_edge` that are direction-dependent.

So it is possible to build graphs where the lexBFS ordering computed by `lex_BFS` is not considered valid.

What should be the fix ?
My first impulse would be to do the two following changes in `_is_valid_lex_BFS_order`:
- use `G.neighbor_iterator` for Graph and DiGraph
- use `G.has_edge(u,v) or G.has_edge(v,u)`

Is it the correct approch ?

I am very interested by the response as I am currently writing an extended lexBFS algorithm, and I would like to be able to rely on this function to test the output of the new implementation.

### Environment

```markdown
- **OS**: Debian 11
- **Sage Version**: 10.3
```

### Checklist

cyrilbouvier:
For directed graphs, the semantic of `lex_BFS`, `lex_DFS`, `lex_DOWN` and `lex_UP` methods are not consistent. In some parts of the code, directed graphs were converted into undirected graphs before the actual computation and in some other parts (in particular in the check function `_is_valid_lex_BFS_order`) directed graphs were explicitly considered as directed.

In this PR, the following changes are implemented:
- always consider (and convert) directed graphs into undirected graphs for `lex_*` methods
- do the same conversion in `_is_valid_lex_BFS_order` for consistency
- add the following line in the documentation of the `lex_*` methods:
```py
r"""
   Loops and multiple edges are ignored during the computation of <name of the method> and
   directed graphs are converted to undirected graphs.
"""
```
- change some doctests: with the previous changes, methods can return a different (but still valid) order
- set the copy of the graph to immutable

Fixes #38234 

### :memo: Checklist


src/sage/graphs/traversals.pyx
```diff
@@ -258,6 +267,9 @@ def lex_BFS(G, reverse=False, tree=False, initial_vertex=None, algorithm="fast")
         algorithm is in `O(n + m)`, and our implementation follows that
         complexity. See [HMPV2000]_ and next section for more details.
 
+    Loops and multiple edges are ignored during the computation of lex_BFS and
```
dcoudert: either write `lex BFS` or ``` ``lex_BFS`` ```. Same for all lex_* methods.

dcoudert (changes requested):
otherwise LGTM.

src/sage/graphs/traversals.pyx
cyrilbouvier: done

dcoudert (approved):
LGTM.

dcoudert:
I set this PR back to positive review.

PR #38274, opened 2024-06-25
Fix broken anchors in CHANGES.html of doc preview

kwankyu:
broken by careless code in #38213:
```
          # Restore the new doc from changes by "wipe out"
          (cd docs && git checkout -f)
```
which also wipes out anchors planted by `.ci/create-changes-html.sh`!

To reviewer: changes over #38220 is in https://github.com/sagemath/sage/pull/38274/commits/d2d1514148f46a4ceeea650eab1fd0a0800ebbc9

Tested in : https://github.com/kwankyu/sage/actions/runs/9659812609/job/26643993813. 
Try to open https://doc-pr-43--sagemath-test.netlify.app/html/en/reference/references/#hunk2  to check  

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

### :hourglass: Dependencies

#38220

PR #38285, opened 2024-06-26
Add version selector to doc

kwankyu:
as #38220 allows us to implement the version selector to sage doc.

Preview : https://doc-pr-53--sagemath-test.netlify.app/html/en/reference/

The script `sage-update-version`(used by the release manager) updates src/doc/website/versions.txt. The file will be accessible from https://doc-release--sagemath.netlify.app/html/en/versions.txt, after 10.4 release. 

The Preview is using https://doc-release--sagemath-test.netlify.app/html/en/versions.txt for test purpose.

### :memo: Checklist

### :hourglass: Dependencies

#38274


mkoeppe:
Nice! Would the selection box better be placed in the left sidebar?

kwankyu:
I'd rather leave it separate from the main doc elements. I like this style: https://docs.djangoproject.com/en/5.0/

kwankyu:
I think "versions" is a meta element. That it has a fixed place in the screen reflects its meta character.

kwankyu:
I could publish doc-10-3 and doc-10.2 from the 10.3 and 10.2 sage installs on my system. They include also pdf docs. 

But 10.1 and 10.0 on my system do not work. I tried to fix them but failed! 

Where can I get docs (pdfs included) for 10.1 and 10.0?

kwankyu:
artifacts in github are gone.

kwankyu:
I am building them on ubuntu vm.

kwankyu:
Docs for 10.0 and 10.1 are now uploaded.

kwankyu:
I thought about how many old docs should be supported. For now, it is 10. This number should be enough for stable releases for two years.

mkoeppe:
Very nice.

The version drop down list shows the word "versions". Would it be better to have it show the current version?

kwankyu:
> The version drop down list shows the word "versions". Would it be better to have it show the current version?

Good idea. Done. See it in Test 1.

> Would the selection box better be placed in the left sidebar?

I tried with it, but could not make it look good. So I gave it up.

Instead, in Test 2, I made the selector hide to the right so that it does not interfere with the flow of the main text, which could be an annoyance when the doc is viewed in a small screen such as mobile device.

mkoeppe:
> > Would the selection box better be placed in the left sidebar?
> 
> I tried with it, but could not make it look good.

Would it be possible to put it near the top of the left sidebar, just below the Sage logo, replacing the text 
```
  Sage 10.4.rc1 Documentation
```
by
```
  Sage Documentation
  <10.4.rc1>
```

src/doc/versions.txt
```diff
@@ -0,0 +1,20 @@
+# This file is used by the version selector of the sage doc
```
mkoeppe: How about putting this file in one of the `_static` directories instead?
Then `fetchVersions` could get it from one of the deployed documentation sites, instead of having to refer to raw.githubusercontent.com

kwankyu:
> > > Would the selection box better be placed in the left sidebar?
> > 
> > 
> > I tried with it, but could not make it look good.
> 
> Would it be possible to put it near the top of the left sidebar, just below the Sage logo, replacing the text
> 
> ```
>   Sage 10.4.rc1 Documentation

No. This text plays a role, as you can see navigating the doc. The role is related with how the doc is organized with its many subdocs (and subdocs of subdocs).

Places in the left sidebar that least intruding into the doc itself are (1) just above the logo and (2) below of the navigation list (of links).  For (2), the user would often need to scroll down to the bottom of the left sidebar to find the selector. With (1), it seems  somewhat affecting how logo looks... Among (1) and (2), I prefer (1).

src/doc/versions.txt
kwankyu: That is how I did it first, but discarded it thinking that it is a mistake.

The selector is to list currently supported docs (of the stable sage versions that released around in the last two years). The list of supported (available) docs should be independent of the deployed doc site. If it were, the site deploying an old doc would list only older docs in the selector, some of which may not be available anymore.

Using raw.githubusercontent.com is how to list only supported docs in the selector even in sites deploying an old doc. A realistic scenario is when a user is viewing a locally built doc for an old sage version.

mkoeppe: I mean, it could refer to wherever we deploy the latest doc, no?

kwankyu: Ah, I misunderstood you. You mean to fetch the file from a prominent deployed doc site like (1) "doc-release--sagemath.netlify.app" or (2) "doc.sagemath.org"? 

That would work. Then I would put it to the root of the doc, rather than "_static". I prefer (1).

kwankyu: > I mean, it could refer to wherever we deploy the latest doc, no?

Yes.

kwankyu: (1) will reflect the change of the list faster than (2).

mkoeppe: Sounds good to me.

kwankyu: There is a problem in fetching "versions.txt" from "doc-release--sagemath.netlify.app" when the browser is showing the doc from file system, because of CORS policy. The netlify server does not allow that. Hence the version selector does not work in this case.

There is no such problem with "raw.githubusercontent.com", which seems permissive.

Why is it that we should avoid "raw.githubusercontent.com/sagemath/sage/develop/..."?  I agree that "doc-release--sagemath.netlify.app" is explicitly related with a "release".

kwankyu: It is not only from the file system...

Screen Shot 2024-07-01 at 4 00 12 PM

kwankyu: Adding `_headers` file with content
```
/html/en/versions.txt
  Access-Control-Allow-Origin: *
  Access-Control-Allow-Methods: GET
  Access-Control-Allow-Headers: Content-Type
```
seems to make the doc site more permissive related with CORS. We may add the `_headers` file when we deploy "doc-release--sagemath.netlify.app" . 

But this is getting complicated... I need an excuse why I should do this :-)

kwankyu: I did it anyway. See the PR description.

kwankyu:
I prepared version selector in three styles S1, S2, and S3. In S3, I tried to follow as close as your suggestion.

mkoeppe:
in S3, I would suggest to take out the version number from the displayed title: "Sage 10.5 Documentation" -> "Sage Documentation" and instead put the word "version" before the version drop down.

I'd also suggest to style the "pr 1234" as "PR #1234" or "#1234" or "sagemath/sage#1234"

Also, would it be possible to patch the drop down menu into the old documentation?

kwankyu:
> in S3, I would suggest to take out the version number from the displayed title: "Sage 10.5 Documentation" -> "Sage Documentation" and instead put the word "version" before the version drop down.

Technically, that is part of the brand logo. I have other ideas about that element. I want to restrict the scope of this PR to just choosing among S1, S2, S3. Later in a sequel PR, we may do necessary changes to that element for the chosen style.

> I'd also suggest to style the "pr 1234" as "PR #1234" or "#1234" or "sagemath/sage#1234"

Note that there is already a link to the PR in the announcement banner. I think "pr 1234" is in rhyme with "develop" or "latest". I am not sure if your suggestions are betters. Let's take some time.

> Also, would it be possible to patch the drop down menu into the old documentation?

Maybe, for some. I am not sure if it is worth while as old docs will be replaced with version selector as time goes by. Anyway, definitely not a work item for this PR.

mkoeppe:
> I want to restrict the scope of this PR to just choosing among S1, S2, S3.

I prefer S3 even without further refinements.

kwankyu:
> in S3, I would suggest to take out the version number from the displayed title: "Sage 10.5 Documentation" -> "Sage Documentation" and instead put the word "version" before the version drop down.

The version in the displayed title is the genuine version of the doc while the "current version" in the selector is based on the url. The "current version" can be "pr 1234" or "local" (for doc from the file system). We should keep the former.

I improved S3. Check  https://doc-pr-49--sagemath-test.netlify.app/html/en/reference/. I like this S3 more than S1 or S2.

kwankyu:
> The version drop down list shows the word "versions". Would it be better to have it show the current version?

In the present S3, It is meaningless to show "current version" in the version selector. I will remove this and use "other version".

mkoeppe:
Looks good to me

kwankyu:
I like it too!

mkoeppe:
Feel free to set to positive review when ready

kwankyu:
Thanks. Many thanks for the style suggestion!

mkoeppe:
PR description still says "This PR is best to be merged after 10.4 release."

kwankyu:
That was the plan, but now along with 38332, the versions selector and many fixes are better to be right in 10.4 release. 

Thanks.

kwankyu:
I now learned the reality, how netlify works. 

Screenshot 2024-07-03 at 9 40 38 PM

As the picture shows, old deployments, about 3 months old, are removed from netlify. We need extra resources to sustain versioned docs...

PR #38301, opened 2024-06-28
graph: modular decomposition of a single vertex should be a single tree node

cyrilbouvier:
The modular decomposition tree computed by Sage for a graph with a single vertex is wrong (tested in Sage from 9.8-10.4.rc0)

```py
sage: Graph(1).modular_decomposition()
(PRIME, [0])
```

Sage returns a tree with a `PRIME` root with a single child representing the vertex `0`, but according to the recursive definition of a modular decomposition tree, it should be a single node tree ([see Wikipedia for example](https://en.wikipedia.org/wiki/Modular_decomposition#:~:text=As%20a%20base%20case%2C%20if,is%20a%20single%20tree%20node.))

This PR fixes this issue. It is an easy fix, as the case of single vertex graphs was treated separately in the `modular_decomposition` method of `Graph`.
Two doctests were modified to reflect this changes.

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/graph.py
+++ b/src/sage/graphs/graph.py
@@ -8209,7 +8209,7 @@ class Graph(GenericGraph):
         if not self.order():
             D = None
         elif self.order() == 1:
-            D = create_normal_node(self.vertices(sort=False)[0])
+            D = create_normal_node(next(self.vertex_iterator()))
         else:
             D = habib_maurer_algorithm(self)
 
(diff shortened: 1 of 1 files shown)


src/sage/graphs/graph.py
```diff
@@ -8209,8 +8209,7 @@ def modular_decomposition(self, algorithm=None, style='tuple'):
         if not self.order():
             D = None
         elif self.order() == 1:
-            D = create_prime_node()
-            D.children.append(create_normal_node(self.vertices(sort=False)[0]))
+            D = create_normal_node(self.vertices(sort=False)[0])
```
dcoudert: you could use `D = create_normal_node(next(self.vertex_iterator()))`

cyrilbouvier: done in 420692b

dcoudert (approved):
LGTM.

PR #38312, opened 2024-06-30
`build/pkgs/cbc` 2.10.11, `sage_numerical_backends_coin`

mkoeppe:
- Our `cbc` package is ancient and fails to build on various platforms
- the current `sage_numerical_backends_coin` package has not been updated for Cython 3
- current `cbc` tarballs no longer seem to contain dependencies such as Clp. 

Switching `cbc` to install via `coinbrew`.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #38319


mkoeppe:
@culler Here is a tentative fix for `sage_numerical_backends_coin`, but on my macOS machine I'm getting segfaults in the Sage test suite when it is installed. I'll check what it looks like on the other platforms here in the CI.

culler:
Wow.  That does much more than add a few noexcept qualifiers.  It adds yet another build system, with its own spkg!  I see that lots of jobs were killed due to abort, but I don't see a stack trace.

mkoeppe:
Yes... Unfortunately our ancient version of cbc fails to build on many platforms, including the "manylinux" wheel building environment that I'm trying to set up. Hence the upgrade. But it looks like this won't be ready for 10.4.

Issue #38300, opened 2024-06-28
Use gap iterators to iterate over the conjugacy classes of subgroups


mantepse:
### Problem Description

Currently we can only compute a list of all subgroups up to conjugacy, using `G.conjugacy_classes_subgroups()`.  If we only want to look at the first few, we are out of luck.

### Proposed Solution

Use Gap's iterators:
```
def conjugacy_classes_subgroups_iterator(self):
    it = libgap.Iterator(self._libgap_().ConjugacyClassesSubgroups())
    while not it.IsDoneIterator().sage():
        sub = it.NextIterator()
        yield self.subgroup(gap_group=sub.Representative())
```

### Alternatives Considered

bäh.

### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

fchapoton:
Note that Gap does not provide true iterators for that, currently.

PR #38317, opened 2024-06-30
make "conjugacy_classes_subgroups" an iterator (twice)

fchapoton:
fixing #38300 

by turning the methods `conjugacy_classes_subgroups` into an iterator (twice)

### :memo: Checklist


tscrim:
What is the reasoning behind doing the iterator in the two different implementations (in particular, for the permutation groups)?

fchapoton:
well, it was for consistency. In my opinion, the question here is : can we switch to iterators and still keep the same name for these methods ?

tscrim:
> well, it was for consistency.

By consistency, you mean in the output order, correct?

> In my opinion, the question here is : can we switch to iterators and still keep the same name for these methods ?

We kinda, sorta, shouldn't. We've done it before, but perhaps we should call this `*_iterator` and "deprecate" the list output methods saying they will change to iterators. It's a bit of a pain, but we really should do it that way.

fchapoton:
(1) I was thinking to the consistency of being both iterators
(2) painful indeed to have to do a double deprecation, first for the existing method, then for the temporary iterator.  I am tempted to have an underscored iterator method to avoid the second one. But this means that we provide no official API for an iterator until more than one year from now.

mantepse:
I would rather have it be an iterator without deprecation right away (option 1), but I can live with this solution (option 2), too.  I could also live with two methods, one producing a list and the other an iterator (option 3), as for example `subgraph_search` and `subgraph_search_iterator`.  In fact, I think having both methods is what I'd like best.

I don't know how to reach a decision.  @tscrim and @fchapoton, could you do a ranking?

tscrim:
Sorry for my slow response.

In many ways, this would be the correct way to go, but the downside is you get that annoying deprecation warning appearing, even when using it as an iterator (as indicated by the "failing" doctest). To get around this, we need to do a 2-stage deprecation, which is bothersome, which leads back to the original conclusion: take the pain with the backwards incompatible change.

Some other thoughts. Perhaps the thing to do is make all such methods uniformly be iterators across all groups, so it is one big break and consistency change. Making it a `view`-like object feels a bit overkill.

We do have some precedent for simply changing the output to an iterator (beyond the Python3 switch). Well, perhaps we should just try it (the switch to an iterator without deprecation) here to see what happens...

mantepse:
So, apparently having both methods is not a desirable option.  Thus, unless @fchapoton objects, I'd suggest to go back to cf8cdbe78befde70b15b2b1ac6df170f22a201c9

I had a brief look which other methods might better return an iterator and found 

* `conjugacy_classes_representatives`
* `conjugacy_classes` (which already has a partner `conjugacy_classes_iterator` for the symmetric group)
* `subgroups`
* `cosets`

Are these the methods you had in mind, @tscrim?

tscrim:
Yes, those are some that I was thinking of. I didn't do a detailed check (sorry!); it was just based on my recollection that there were a lot of `list` output methods that could be iterators.

mantepse:
@fchapoton, would you like me to provide iterators for the other methods, to share the workload?

fchapoton:
So here is a branch with several direct replacements with iterators.

I avoided to do that for "cosets".

I could not refrain myself from pep8 cleanup in one of the modified file, sorry.

mantepse:
Unfortunately, trying your branch, I figured that I made a very silly mistake.  What `libgap.Iterator(self._libgap_().ConjugacyClassesSubgroups())` does is the same what python would do: it computes the list of all conjugacy classes of subgroups, and then turns it into an iterator.

I cannot find the misleading bit of information anymore, that made me think it would do something magical.

I'm sorry.

There is a true gap iterator `AllSubgroupsIterator` in a package `utils`, which is not a standard gap package though.

I'll have a look whether it is easy to save the situtation.

fchapoton:
closing as gap does not really provide iterators for the moment

PR #38324, opened 2024-07-01
Implement formal Dirichlet series

kedlaya:
We implement formal Dirichlet series (truncated to fixed precision) over an arbitrary base commutative ring. By analogy with polynomials and power series, the underlying representations can be either dense or sparse.

One motivation for this is to eventually develop L-functions as an object class.


tscrim:
It would probably be good to have the lazy Dirichlet series' `truncate()`method return an element here. Definitely adding an `O()` and `add_bigoh()` method to return here, along with having the corresponding coercion.

kedlaya:
So this raises some questions due to the fact that these are fixed-precision rings, and the elements don't currently carry their own precision. (In that way they are more like an `IntegerModRing` than a fixed-precision p-adic ring.)

For example, should `truncate` return in the same parent? That would look like:
```sage
sage: R = DirichletSeriesRing(ZZ, 10)
sage: f = R([1,2,3,4,5,6,7])
sage: f
1 + 2*2^-s + 3*3^-s + 4*4^-s + 5*5^-s + 6*6^-s + 7*7^-s + O(10^-s)
sage: f.truncate(5)
1 + 2*2^-s + 3*3^-s + 4*4^-s + O(10^-s)
```
although one could specify an optional argument to decide between this and creating a new parent. (In that case what should the default be?)

tscrim:
I was confused about what `truncate` should do; that would return an element in the (lazy) series (with infinite precision); yes, it should return something in the same parent it started. I was thinking of `polynomial()` for the lazy power series ring (of course, that name doesn't make sense here), and I was thinking the series ring would be a decent initial target while there is no direct sum analog implemented (if it even makes sense to have it).

I see. That is quite a bit different than how the power series rings are implemented. So it is more like how `RR` is implemented, right? Is that better for the computations you would expect/want/etc. or for ease of an initial implementation?

kedlaya:
It does make mathematical sense to consider the ring of "Dirichlet polynomials" (Dirichlet series with finitely many coefficients), which would then be an exact ring whenever the coefficient ring is exact.

However, in practice this concept doesn't play much of a role, so I didn't bother to maintain the parallelism with ordinary polynomials/power series in this respect.

tscrim:
The question I have is should the implementation 

1. use the same parent allow arbitrary (but potentially only finite) precision (like how the power series is implemented); or
2. everything is always considered with a fixed precision (like how `RR` is implemented).

Option 1 means the series knows exactly how much precision it has, so the computation is guaranteed to be correct, and coercion can end up in the higher default precision ring with no changes to the internal structure of the elements. I think it requires a bit more code for the elements, but less for the parents. It is basically trivial to extend this to exact elements I believe.

Option 2 can lead to incorrect results if there is any loss of precision for any of the computations. I don't think this happens, but I need to check. If you want to start mixing precisions, then you need to coerce elements and truncate accordingly. There is no exact elements.

If you're happy with Option 2 (the current implementation), then we can continue with it. I just want to know if you considered the first option and if there was any particular reason for not choosing it.

src/sage/rings/dirichlet_series_ring.py
```diff
@@ -0,0 +1,116 @@
+from sage.misc.misc_c import prod
```
mkoeppe, line 38: This needs a module docstring and a copyright header

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,393 @@
...
+        EXAMPLES::
+
+            sage: R = DirichletSeriesRing(ZZ, 10)
+            sage: f = R([1, 2, 3])
+            sage: 1/f
+            1 - 2*2^-s - 3*3^-s + 4*4^-s + 12*6^-s - 8*8^-s + 9*9^-s + O(10^-s)
+        """
+        c = other[1]
+        if not c.is_unit():
+            return ValueError("Leading coefficient must be a unit")
```
mkoeppe: `raise`

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,410 @@
...
+            sage: R = DirichletSeriesRing(ZZ, 10)
+            sage: f = R([1, 2, 3])
+            sage: f
+            1 + 2*2^-s + 3*3^-s + O(10^-s)
+        """
+        coeffs = self.coefficients()
+        ans = ""
+        prec = self.parent().precision()
+        atomic_repr = self.parent().base_ring()._repr_option('element_is_atomic')
+        for i in range(1, prec):
```
mkoeppe: Check if this can be refactored using `repr_lincomb`

src/sage/rings/dirichlet_series_ring.py
```diff
@@ -0,0 +1,153 @@
+r"""
```
mkoeppe, line 1: The new modules should be added to the documentation index

kedlaya:
> The question I have is should the implementation
> 
>     1. use the same parent allow arbitrary (but potentially only finite) precision (like how the power series is implemented); or
> 
>     2. everything is always considered with a fixed precision (like how `RR` is implemented).
> 
> 
> Option 1 means the series knows exactly how much precision it has, so the computation is guaranteed to be correct, and coercion can end up in the higher default precision ring with no changes to the internal structure of the elements. I think it requires a bit more code for the elements, but less for the parents. It is basically trivial to extend this to exact elements I believe.
> 
> Option 2 can lead to incorrect results if there is any loss of precision for any of the computations. I don't think this happens, but I need to check. If you want to start mixing precisions, then you need to coerce elements and truncate accordingly. There is no exact elements.
> 
> If you're happy with Option 2 (the current implementation), then we can continue with it. I just want to know if you considered the first option and if there was any particular reason for not choosing it.

I guess I was thinking of these as being like an `IntegerModRing`, in which the precision of an element is specified by the ambient ring. If there is enough interest in these Dirichlet series rings to warrant a refactor later, then so be it.

tscrim:
> I guess I was thinking of these as being like an `IntegerModRing`, in which the precision of an element is specified by the ambient ring.

I am guessing you mean the p-adics since `IntegerModRing` is exact as `ZZ` is exact? Otherwise I don't understand this statement.

What I am worried about is if there is some operation you can do that loses precision. Well, perhaps it isn't a major thing to worry about at this point...

Hence, it seems like the major design points are set, correct? Let me know if you are ready for me to do a full review.

src/sage/rings/dirichlet_series_ring.py
kedlaya: Which file(s) do I need to touch to do this? I couldn't find this in the developer's guide.

kedlaya:
> > I guess I was thinking of these as being like an `IntegerModRing`, in which the precision of an element is specified by the ambient ring.
> 
> I am guessing you mean the p-adics since `IntegerModRing` is exact as `ZZ` is exact? Otherwise I don't understand this statement.
> 
> What I am worried about is if there is some operation you can do that loses precision. Well, perhaps it isn't a major thing to worry about at this point...
> 
> Hence, it seems like the major design points are set, correct? Let me know if you are ready for me to do a full review.

There is no natural analogue of the p-adic field (or the Laurent series ring over a base) in this context; all we have is the Dirichlet series ring. I mention `IntegerModRing` because a p-adic ring with fixed precision (resp. a power series ring with fixed precision) is mathematically equivalent to working with integers modulo a power of p (resp. with polynomials modulo a power of the variable).

I can't think of a natural operation that would cause a loss of precision in this setting.

src/sage/rings/dirichlet_series_ring.py
tscrim: @kedlaya I would put it in `src/doc/en/reference/power_series/index.rst`.

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,404 @@
...
+                if i > 0 and i < prec:
+                    self.data[i] = base_ring(j)
+        elif callable(data):
+            z = parent.zero()
+            for i in range(1, prec):
+                j = data(i)
+                if j != z:
+                    self.data[i] = j
+
+    def __eq__(self, other):
```
tscrim: Implementing
```suggestion
    def _richmp_(self, other, op):
```
would be better I think.

You will also need to implement a `__hash__`.

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,404 @@
...
+            False
+            sage: R(1) == 1
+            True
+            sage: 1 == R(1)
+            True
+        """
+        other = other - self
+        return all(not other[i] for i in range(1, other.parent().precision()))
+
+    def __bool__(self):
```
tscrim: This method needs doctests.

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,404 @@
...
+            ans = repr(coeffs[1]) + " + "
+        else:
+            ans = ""
+        # Use repr_lincomb for the other coefficients.
+        ans += repr_lincomb([(str(i) + '^-s', coeffs[i]) for i in sorted(coeffs.keys()) if i > 1])
+        # Append the precision marker.
+        ans += " + O({}^-s)".format(prec)
+        return ans
+
+    def _add_(self, other):
```
tscrim, line 189: Why not use `@abstract_method`here and below?

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,404 @@
...
+            sage: g.parent()
+            Dirichlet Series Ring over Integer Ring with fixed precision 5
+        """
+        parent = self.parent()
+        if new_parent:
+            from sage.rings.dirichlet_series_ring import DirichletSeriesRing
+            parent = DirichletSeriesRing(parent.base_ring(), prec, parent.is_sparse())
+        return parent({i: j for i,j in self.coefficients().items() if i < prec})
+
+class DirichletSeries_dense(DirichletSeries_generic):
```
tscrim, line 309: These classes' methods  will all need docstrings and doctests.

src/sage/rings/dirichlet_series_ring_element.py
```diff
@@ -0,0 +1,404 @@
...
+        out = {}
+        prec = parent.precision()
+        for i, x in self.data.items():
+            for j, y in other.data.items():
+                k = i*j
+                if k < prec:
+                    out[k] = out[k] + x*y if k in out else x*y
+        return parent(out)
+
+class DirichletSeriesIterator:
```
tscrim, line 382: I don't see the need for this class. Why not just use the default iterator of the defining list?

tscrim (changes requested):
Not sure if you are fully ready, but here is the start of a full review as well.

src/sage/rings/dirichlet_series_ring.py
```diff
@@ -0,0 +1,153 @@
...
+            sage: S = DirichletSeriesRing(ZZ, 10, sparse=True)
+            sage: R.has_coerce_map_from(ZZ)
+            True
+            sage: R.has_coerce_map_from(S)
+            False
+            sage: S.has_coerce_map_from(R)
+            True
+        """
+        self.Element = DirichletSeries_sparse if sparse else DirichletSeries_dense
+        CommutativeRing.__init__(self, base_ring, names=None, category=base_ring.category())
```
fchapoton: please do not use `CommutativeRing.__init__` at all but instead set the category in `Parent.__init__` to `CommutativeRings()` or rather `CommutativeAlgebras(base_ring)`

src/sage/rings/dirichlet_series_ring.py
```diff
@@ -0,0 +1,153 @@
...
+            sage: R = DirichletSeriesRing(ZZ, 10)
+            sage: R
+            Dirichlet Series Ring over Integer Ring with fixed precision 10
+        """
+        return "Dirichlet Series Ring over {} with fixed precision {}".format(self.base_ring(), self.__precision)
+
+    def is_sparse(self):
+        """
+        Return `True` if this ring uses sparse internal representation.
+
```
fchapoton, line 93: double backtick around True

Issue #38329, opened 2024-07-02
Finding an Elliptic curve from a Lattes map


Nathabolin:
### Problem Description

Sagemath has pre-existing functions to take any elliptic curve and produce its lattes map. We want to make a function which would invert this process. 

Pull request for code can be found at Fixes #38330

### Proposed Solution

Our solution is to construct a symbolic lattes map from a short form elliptic curve and a symbolic conjugating matrix. 

We then conjugate the map with the matrix producing a general form a conjugated lattes map. 

We then create an ideal from the set of equations generated by matching term by term the input map and symbolic map. 

Finding the variety of that ideal gives back the entries of the matrix and desired coefficients for a short form elliptic curve. 

### Alternatives Considered

There is no other function that produces these results already extant in sagemath

### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

PR #38330, opened 2024-07-02
Lattes to curve function code commits for enhancement #38329 

Nathabolin:
- **Fixing issue #38329**
- **Adding Function with documentation to be reviewed for the above ticket**
- **Lattes to curve function**

### :memo: Checklist


bhutz:
A number of simple things to fix here:

- 6843: Finds a short Weierstrass model associated to this Lattes map
- 6844 - extra white space
- 6845 ``self`` is assumed to be
- 6848, 6852, 6853 ``self``
- 6866, 6882, 6895 white space between arguments
- 6875 extra blank line
- 6911, 6920, 6928, 6934, 6942 fix indent
- 6914,6915,6916 extra white space
- 6918 extra \
- 6912-25, 6929 spacing needed

Less simple: I think you should add at least one doc tests that checks the output is correct. Something like

E, m  = F.Lattes_to_curve()
P.Lattes_map(E,2).conjugate(m) == F

Finally, you cannot in general assume that u=1. The problem you are running into in the more general case is that the solution is not defined over QQ, not that there isn't a solution.

bhutz:
I did a full functionality test and found a couple things. There are a couple more minor coding things too.

- need to wrap the very long output lines in the doctests
- `and the defining polynomial of any required extension` is not actually part of the output. That is just part of the elliptic curve definition, so remove that phrase.
- move imports to top of file
- 6939-40 These lines are not used
```
P = ProjectiveSpace(self.base_ring(), 1, "x, y")
 x, y = P.gens()
```
- I would suggest a parameter that defaults to true `check_Lattes=True`. If that parameter is true, then call `is_Lattes` after the degree check (6920). Add a test against a simple map
```
P.<x,y>=ProjectiveSpace(QQ,1)
f=DynamicalSystem([x^4,y^4])
f.Lattes_to_curve()
```
- There are some issues over number fields. The following fails due to a naming conflict on the variables
```
K.<a>=QuadraticField(2)
P.<x,y>=ProjectiveSpace(K,1)
E=EllipticCurve([1,a])
f=P.Lattes_map(E,2)
print(f)
f.Lattes_to_curve()
```
I would suggest renaming the variables to something no-one would typically use such as
```R = PolynomialRing(self.base_ring(),6,'avar,bvar,uvar,vvar,wvar,tvar')
    a = pts[0]['avar']
    b = pts[0]['bvar']
    u = pts[0]['uvar']
    v = pts[0]['vvar']
    t = pts[0]['tvar']
    w = pts[0]['wvar']
```
Nothing else needs to change for this since you use
```
a, b, u, v, w, t = R.gens()
```
but you do need to fix another coercion problem from this
```
eq += [R(g.coefficient({x:i})) - R(g_sym.coefficient({z:i}))]
```
Add a number field example after you fix these.
- Similarly there are problems coercing to QQbar when the field doesn't have the coercion specified
```
R.<t>=QQ[]
K.<c>=NumberField(t^3-2)
P.<x,y>=ProjectiveSpace(K,1)
E=EllipticCurve([1,c])
f=P.Lattes_map(E,2)
f.Lattes_to_curve()
```
This can be fixed by changing the code to
```
This can be fixed by changing the code to do:
phi = R.base_ring().embeddings(QQbar)[0]
eq = [poly.change_ring(phi) for poly in eq]
I = R.ideal(eq)
pts = I.variety()
```
- Finally: This is returning an error, but it should work:
```
P.<x,y>=ProjectiveSpace(QQbar,1)
E=EllipticCurve([1,2])
f=P.Lattes_map(E,2)
f.Lattes_to_curve()
```
so you just need to allow QQbar through in your check against numberfields

Nathabolin:
"move imports to top of file"
I'm going to do this of course but oddly enough 
the file already does try to import them

lazy_import('sage.rings.qqbar', 'number_field_elements_from_algebraics') 

is on line 118

bhutz:
I having all the functionality working in this version.  I still have a couple comments though.

- The example
"""
sage: P.<x,y> = ProjectiveSpace(QQbar,1)
sage: F = DynamicalSystem_projective([x^4,y^4])
sage: F.Lattes_to_curve(check_lattes=True)
...
NotImplementedError: Base ring must be a number field
"""
is failing for `is_Lattes` not for `Lattes_to_curve`. I also don't see a reason why it doesn't allow QQbar. Why don't you make the same check in that function too.

- Since the default is not to check if its Lattes, maybe add a check
"""
if len(pts) == 0:
	raise ValueError("No Solutions found. Check if map is Lattes.")
"""

- There are some white spacing issues in the later examples.

vbraun:
Merge conflict

Nathabolin:
@vbraun Could you point me to where to see the merge conflicts?

vbraun:
Looks like you fixed the conflict in the last merge commit

PR #38338, opened 2024-07-06
Simplify get_units_of_order in btquotient

Eloitor:
Simplified the implementation of get_units_of_order in btquotient.

### :memo: Checklist


Eloitor:
With this changes, can I remove the `# random` in this test? https://github.com/sagemath/sage/blob/fc3a044fdfaf520aeb6fc2657d2aed9cba41cf10/src/sage/modular/btquotients/btquotient.py#L2542-L2550

Eloitor:
ready for review

Eloitor:
`vectors_by_length` does not use LLL. If we want to have LLL, we could replace the pari call with something like
```python
n, norm_max, vecs = pari(OM).qfminim(flag=2)
```

Eloitor:
We should add that the returned units are modulo ±1, but I don't know the best wording for this.

PR #38341, opened 2024-07-09
Implement EllipticCurve_with_prime_order() constructor

### LINKED ISSUE #38342: EllipticCurve_with_order() crashes with overflow error on values of 64 bits or more
### Steps To Reproduce

Building sagemath from the source code on Ubuntu 22.04 and on version `10.4.beta8` (or any version after https://github.com/sagemath/sage/pull/37119 I suppose), run

```py
sage: from sage.schemes.elliptic_curves.ell_finite_field import EllipticCurve_with_order
sage: N = randint(2^63, 2^64-1)
....: for E in EllipticCurve_with_order(N):
....:     print(E)
...
OverflowError: Python int too large to convert to C long
```

### Expected Behavior

`EllipticCurve_with_order()` should handle big integers or precise that it doesn't in its docstring.

### Actual Behavior

Feeding an integer of 64 bits or more to the function overflows with the error `OverflowError: Python int too large to convert to C long`.

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Ubuntu 22.04
- **Sage Version**: 10.4.beta8, Release Date: 2024-06-01, built from source code
```

### Checklist

JosePisco:
Me and @grhkm21 suggest this diff against develop that implements the `EllipticCurve_with_prime_order(N)` constructor. Using the prime order `N` in input, this method finds another prime `p` and constructs an elliptic curve `E/Fp` with `#E(Fp) = N`.

It follows Algorithm 2.2 of the paper [Constructing Elliptic Curves of prime order](https://arxiv.org/abs/0712.2022) by Bröker and Stevenhagen. The running time is quite random depending on the input parameter but can turn out to be fast for some larger values (≃ 256 bits primes). It's also worth noticing that some values will make this function run for a **very** long time.

There had been a [PR](https://github.com/sagemath/sage/pull/37119) by @grhkm21 and @GiacomoPope that implements the `EllipticCurve_with_order()` method. This PR would intuitively fit nice into their work but I felt uncomfortable with it returning an iterator. I felt like returning a single curve was more handy so I implemented this method in a separate function that does so but I'm open to suggestions if this is of any interest to the community.

Fixes https://github.com/sagemath/sage/issues/38342

### :memo: Checklist

### DIFF
--- a/src/sage/schemes/elliptic_curves/all.py
+++ b/src/sage/schemes/elliptic_curves/all.py
@@ -27,10 +27,10 @@ from sage.misc.lazy_import import lazy_import
 lazy_import('sage.schemes.elliptic_curves.jacobian', 'Jacobian')
 
 lazy_import('sage.schemes.elliptic_curves.ell_finite_field', 'special_supersingular_curve')
-lazy_import('sage.schemes.elliptic_curves.ell_finite_field', 'EllipticCurve_with_prime_order')
 lazy_import('sage.schemes.elliptic_curves.ell_rational_field',
             ['cremona_curves', 'cremona_optimal_curves'])
 
+from sage.schemes.elliptic_curves.ell_finite_field import EllipticCurve_with_prime_order
 from sage.schemes.elliptic_curves.cm import (cm_orders,
                                              cm_j_invariants,
                                              cm_j_invariants_and_orders,
--- a/src/sage/schemes/elliptic_curves/all.py
+++ b/src/sage/schemes/elliptic_curves/all.py
@@ -27,7 +27,7 @@ from sage.misc.lazy_import import lazy_import
 lazy_import('sage.schemes.elliptic_curves.jacobian', 'Jacobian')
 
 lazy_import('sage.schemes.elliptic_curves.ell_finite_field', 'special_supersingular_curve')
-
+lazy_import('sage.schemes.elliptic_curves.ell_finite_field', 'EllipticCurve_with_prime_order')
 lazy_import('sage.schemes.elliptic_curves.ell_rational_field',
             ['cremona_curves', 'cremona_optimal_curves'])
 
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -2821,14 +2821,19 @@ def EllipticCurve_with_prime_order(N):
 
     TESTS::
 
-        sage: N = next_prime(123456789)
+        sage: for N in prime_range(3, 100):
+        ....:     E = EllipticCurve_with_prime_order(N)
+        ....:     assert E.order() == N
+
+        sage: N = 15175980689839334471
         sage: E = EllipticCurve_with_prime_order(N)
         sage: E.order() == N
         True
 
-        sage: for N in prime_range(3, 100):
-        ....:     E = EllipticCurve_with_prime_order(N)
-        ....:     assert E.order() == N
+        sage: N = next_prime(123456789)
+        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E.order() == N
+        True
 
         sage: N = 123456789
         sage: E = EllipticCurve_with_prime_order(N)
@@ -2841,17 +2846,23 @@ def EllipticCurve_with_prime_order(N):
         ...
         ValueError: input order is not prime
 
+        sage: E = EllipticCurve_with_prime_order(-7)
+        Traceback (most recent call last):
+        ...
+        ValueError: input order is not prime
+
     .. NOTE::
 
         Depending on the input, this function may run forever.
 
     ALGORITHM: [BS2007]_, Algorithm 2.2
     """
-    from sage.arith.misc import is_prime, next_prime
+    from sage.arith.misc import is_prime
     from sage.combinat.subset import powerset
-    from sage.functions.other import floor
+    from sage.functions.other import ceil
     from sage.misc.functional import symbolic_prod as product
     from sage.quadratic_forms.binary_qf import BinaryQF
+    from sage.rings.fast_arith import prime_range
     from sage.schemes.elliptic_curves.cm import hilbert_class_polynomial
 
     if not is_prime(N):
@@ -2862,15 +2873,10 @@ def EllipticCurve_with_prime_order(N):
     # algorithm to terminate after a number of rounds that is polynomial in
     # loglog N.
     r = 0
-    S = []
-    # First prime. Iterating over the primes by chunks of size log(`N`).
-    p = next_prime(0)
 
     while True:
-        while p < (r + 1) * floor(N.log()):
-            S.append(p)
-            p = next_prime(p)
-        r += 1
+        # Iterating over the odd primes by chunks of size log(`N`).
+        S = prime_range(3, ceil((r + 1) * N.log()))
 
         # Every possible products of distinct elements of `S`.
         # There probably is a more optimal way to compute all possible products
@@ -2900,3 +2906,7 @@ def EllipticCurve_with_prime_order(N):
                             for Et in E.twists():
                                 if Et.order() == N:
                                     return Et
+
+        # At this point, no discriminant has been found, moving to next round
+        # and extending the prime list.
+        r += 1
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -1416,9 +1416,17 @@ class EllipticCurve_finite_field(EllipticCurve_field):
         if q <= 100:
             return self.order() == value
 
+        # This might be slow
+        # if value.is_prime():
+        #     num_checks = 1
+
         # Is value * random == identity?
         for _ in range(num_checks):
-            G = self.random_point()
+            while True:
+                G = self.random_point()
+                if not G.is_zero():
+                    break
+
             if not (value * G).is_zero():
                 return False
 
@@ -2887,16 +2895,49 @@ def EllipticCurve_with_prime_order(N):
     It works for large primes::
 
         sage: N = 0x6cbc824032974516623e732462f4b74b56c4ffbd984380d9
-        sage: E = next(EllipticCurve_with_prime_order(N))
+        sage: E = next(EllipticCurve_with_prime_order(N)); E
+        Elliptic Curve defined by y^2 = x^3 + 2666207849820848272386538889427721639173508298483739490459*x
+         + 77986137112576 over Finite Field of size 2666207849820848272386538889427721639173508298487130585243
         sage: E.order() == N
         True
 
-    But the execution time largely depends on the input::
+    The execution time largely depends on the input, specifically the smallest
+    discriminant ``D`` for which we can apply CM method on. Here it takes
+    slightly longer, though still within `1` second::
 
         sage: N = 200396817641911230625970463749415493753
-        sage: E = next(EllipticCurve_with_prime_order(N))
+        sage: E = next(EllipticCurve_with_prime_order(N)); E
         sage: E.order() == N
         True
+
+    Note that the iterator does *not* return all curves with the given order::
+
+        sage: any(E.base_ring() is GF(7) for E in EllipticCurve_with_prime_order(7))
+        False
+        sage: EllipticCurve(GF(7), [0, 5]).order()
+        7
+
+    However, experimentally it returns many of them. Here it returns all of them::
+
+        sage: N = 23
+        sage: curves = list(EllipticCurve_with_prime_order(N)); curves
+        [Elliptic Curve defined by y^2 = x^3 + 20*x + 20 over Finite Field of size 23,
+         Elliptic Curve defined by y^2 = x^3 + 10*x + 16 over Finite Field of size 23,
+         Elliptic Curve defined by y^2 = x^3 + 14*x + 14 over Finite Field of size 17,
+         Elliptic Curve defined by y^2 = x^3 + 16*x + 25 over Finite Field of size 31,
+         Elliptic Curve defined by y^2 = x^3 + 13*x + 6 over Finite Field of size 19,
+         Elliptic Curve defined by y^2 = x^3 + 24*x + 3 over Finite Field of size 29]
+        sage: import itertools
+        sage: # These are the only primes, by the Weil-Hasse bound
+        sage: for q in prime_range(17, 35):
+        ....:     K = GF(q)
+        ....:     for u in itertools.product(range(q), repeat=2):
+        ....:         try: E = EllipticCurve(GF(q), u)
+        ....:         except ArithmeticError: continue
+        ....:         if E.order() == N:
+        ....:             assert any(E.is_isomorphic(E_) for E_ in curves)
+
+
 
     TESTS::
 
@@ -2944,41 +2985,62 @@ def EllipticCurve_with_prime_order(N):
 
     ALGORITHM: [BS2007]_, Algorithm 2.2
     """
+    import itertools
+    from sage.misc.verbose import verbose
+    from sage.combinat import gray_codes
+    from sage.sets.primes import Primes
     from sage.arith.misc import is_prime, legendre_symbol
-    from sage.combinat.subset import powerset
-    from sage.functions.other import ceil
-    from sage.misc.functional import symbolic_prod as product, log
     from sage.quadratic_forms.binary_qf import BinaryQF
-    from sage.rings.fast_arith import prime_range
     from sage.schemes.elliptic_curves.cm import hilbert_class_polynomial
 
     if not is_prime(N):
         raise ValueError("input order is not prime")
 
-    # The algorithm consists of multiple "search rounds" for a suitable
-    # discriminant `D`, `r` defines the number of rounds. We expect this
-    # algorithm to terminate after a number of rounds that is polynomial in
-    # loglog N.
-    r = 0
-    prime_start = 3
-    prime_end = ceil((r + 1) * log(N))
+    if N < 1000:
+        # Log how long this algorithm will take
+        from sage.rings.fast_arith import prime_range
+        num = sum(1 for p in prime_range(3, 4 * N) if legendre_symbol(N, p) == 1)
+        verbose(f"Total work to enumerate curves with order {N}: 2^{num}", level=2)
+
+    # The algorithm considers smooth discriminants `D`, sorted by their largest
+    # prime factor. We expect this algorithm to terminate after a number of
+    # rounds that is polynomial in loglog N.
     S = []
 
-    while True:
-        # Iterating over the odd primes by chunks of size log(`N`).
-        for p in prime_range(prime_start, prime_end):
-            if legendre_symbol(N, p) == 1:
-                # Equivalent to p* = (-1)^((p - 1) / 2) * p in [BS2007]_ page 5.
-                S.append(-p if p >> 1 & 1 else p)
+    for p in Primes():
+        if p == 2:
+            continue
 
-        # Every possible products of distinct elements of `S`.
-        # There probably is a more optimal way to compute all possible products
-        # of elements of S than using a powerset. Here many multiplications are
-        # done multiple times.
-        for e in powerset(S):
-            D = product(e)
-            if D % 8 != 5 or D >= 0:
+        if legendre_symbol(N, p) != 1:
+            continue
+
+        verbose(f"Considering {len(S) + 1}th valid prime {p}", level=3)
+
+        # Equivalent to p* = (-1)^((p - 1) / 2) * p in [BS2007]_ page 5.
+        p_star = -p if p >> 1 & 1 else p
+
+        # later we need x^2 + (-D)y^2 = 4N, and since y = 0 has no solution, we need
+        # p = |p_star| <= |-D| <= 4N
+        if p > 4 * N:
+            break
+
+        # Although this is O(2^n) instead of O(n2^n), it misses out on the optimisation where
+        # when D becomes too large (see -D > 4 * N below), one can cut the branch entirely
+        # so perhaps this is slower
+        D = p_star
+        for idx, dx in itertools.chain([(None, 1)], gray_codes.product([2] * len(S))):
+            assert abs(dx) == 1, f"hmm, gray codes shouldn't do that. file a bug report"
+
+            if idx is not None:
+                if dx == 1:
+                    D *= S[idx]
+                else:
+                    D //= S[idx]
+
+            if D % 8 != 5 or D >= 0 or -D > 4 * N:
                 continue
+
+            verbose(f"Testing {D=}", level=3)
 
             Q = BinaryQF([1, 0, -D])
             sol = Q.solve_integer(4 * N, algorithm='cornacchia')
@@ -2989,19 +3051,13 @@ def EllipticCurve_with_prime_order(N):
             for p_i in [N + 1 - x, N + 1 + x]:
                 if is_prime(p_i):
                     H = hilbert_class_polynomial(D)
-                    for j0 in H.roots(ring=GF(p_i), multiplicities=False):
-                        E = EllipticCurve(j=j0)
+                    K = GF(p_i)
+                    for j0 in H.roots(ring=K, multiplicities=False):
+                        E = EllipticCurve(K, j=j0)
                         # `E.twists()` also contains E.
                         for Et in E.twists():
-                            if Et.order() == N:
+                            # num_checks=1 is sufficient for prime order
+                            if Et.has_order(N, num_checks=1):
                                 yield Et
 
-        # At this point, no discriminant has been found, moving to next round
-        # and extending the prime list.
-        r += 1
-
-        # For small `N`, the value `(r+1)log(N)` can be less than 3. When that's
-        # the case, we don't need to worry about prime duplicates in `S` since
-        # it will be empty.
-        prime_start = max(3, prime_end)
-        prime_end = ceil((r + 1) * log(N))
+        S.append(p_star)
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -2863,6 +2863,8 @@ def EllipticCurve_with_order(m, *, D=None):
                         continue
                     # This tests whether the curve has given order
                     if Et.has_order(m_val):
+                        # TODO: remove after 38617
+                        Et.set_order(m_val, check=False)
                         seen.add(Et)
                         yield Et
 
@@ -3167,8 +3169,10 @@ def EllipticCurve_with_prime_order(N):
                         E = EllipticCurve(K, j=j0)
                         # `E.twists()` also contains E.
                         for Et in E.twists():
-                            # `num_checks=1` is sufficient for prime order.
+                            # `num_checks=1` is sufficient for prime order
                             if Et.has_order(N, num_checks=1):
+                                # TODO: remove after 38617
+                                Et.set_order(N, check=False)
                                 yield Et
 
         if p != 1:
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -1353,7 +1353,9 @@ class EllipticCurve_finite_field(EllipticCurve_field):
 
         - ``num_checks``-- integer (default: `8`); the number of times to check
           whether ``value`` times a random point on this curve equals the
-          identity
+          identity. If ``value`` is a prime and the curve is over a field of
+          order at least `5`, it is sufficient to pass in ``num_checks=1`` -
+          see the examples below.
 
         .. NOTE::
 
@@ -1403,14 +1405,20 @@ class EllipticCurve_finite_field(EllipticCurve_field):
             sage: E.has_order(N^2 + N)
             True
 
+        Due to the nature of the algorithm (testing multiple of points) and the Hasse-Weil bound, we see that for testing prime orders, ``num_checks=1`` is sufficient::
+
+            sage: p = random_prime(1000)
+            sage: E = EllipticCurve(GF(p), j=randrange(p))
+            sage: q = random_prime(p + 20, lbound=p - 20)
+            sage: E.has_order(q, num_checks=20) == E.has_order(q, num_checks=1)
+            True
+
         AUTHORS:
 
          - Mariah Lenox (2011-02-16): Initial implementation
 
          - Gareth Ma (2024-01-21): Fix bug for small curves
         """
-        # This method does *not* use the cached value of `_order` even when
-        # available.
         q = self.base_field().order()
         a, b = Hasse_bounds(q, 1)
         if not a <= value <= b:
@@ -2980,7 +2988,7 @@ def EllipticCurve_with_prime_order(N):
          Elliptic Curve defined by y^2 = x^3 + 20*x + 20 over Finite Field of size 23,
          Elliptic Curve defined by y^2 = x^3 + 10*x + 16 over Finite Field of size 23]
         sage: import itertools
-        sage: # These are the only primes, by the Weil-Hasse bound
+        sage: # These are the only primes, by the Hasse-Weil bound
         sage: for q in prime_range(17, 35):
         ....:     K = GF(q)
         ....:     for u in itertools.product(range(q), repeat=2):
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -2753,7 +2753,7 @@ def EllipticCurve_with_order(m, *, D=None):
     Return an iterator for elliptic curves over finite fields with the given order. The curves are
     computed using the Complex Multiplication (CM) method.
 
-    A :sage:`~sage.structure.factorization.Factorization` can be passed for ``m``, in which case
+    A :class:`~sage.structure.factorization.Factorization` can be passed for ``m``, in which case
     the algorithm is more efficient.
 
     If ``D`` is specified, it is used as the discriminant.
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -2800,60 +2800,71 @@ def EllipticCurve_with_prime_order(N):
     - ``N`` -- integer; the order for which we seek an elliptic curve. Must be a
       prime number.
 
-    OUTPUT: an elliptic curve `E/\mathbb F_p` of order ``N``
+    OUTPUT: an iterator of elliptic curves `E/\mathbb F_p` of order ``N``
 
     EXAMPLES::
 
         sage: N = next_prime(int(b'sagemath'.hex(), 16))
-        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E = next(EllipticCurve_with_prime_order(N))
         sage: E
         Elliptic Curve defined by y^2 = x^3 + 4757897140353078952*x +
         1841350074072114366 over Finite Field of size 8314040074357871443
         sage: E.order() == N
         True
 
+    You can directly iterate over the function call; here only on the first 10
+    curves::
+
+        sage: N = 54675917
+        sage: for _, E in zip(range(10), EllipticCurve_with_prime_order(N)):
+        ....:     assert E.order() == N
+
     It works for large primes::
 
         sage: N = 0x6cbc824032974516623e732462f4b74b56c4ffbd984380d9
-        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E = next(EllipticCurve_with_prime_order(N))
         sage: E.order() == N
         True
 
     But the execution time largely depends on the input::
 
         sage: N = 200396817641911230625970463749415493753
-        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E = next(EllipticCurve_with_prime_order(N))
         sage: E.order() == N
         True
 
     TESTS::
 
         sage: for N in prime_range(3, 100):
-        ....:     E = EllipticCurve_with_prime_order(N)
+        ....:     E = next(EllipticCurve_with_prime_order(N))
+        ....:     assert E.order() == N
+
+        sage: N = 113
+        sage: for _, E in zip(range(30), EllipticCurve_with_prime_order(N)):
         ....:     assert E.order() == N
 
         sage: N = 15175980689839334471
-        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E = next(EllipticCurve_with_prime_order(N))
         sage: E.order() == N
         True
 
         sage: N = next_prime(123456789)
-        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E = next(EllipticCurve_with_prime_order(N))
         sage: E.order() == N
         True
 
         sage: N = 123456789
-        sage: E = EllipticCurve_with_prime_order(N)
+        sage: E = next(EllipticCurve_with_prime_order(N))
         Traceback (most recent call last):
         ...
         ValueError: input order is not prime
 
-        sage: E = EllipticCurve_with_prime_order(0)
+        sage: E = next(EllipticCurve_with_prime_order(0))
         Traceback (most recent call last):
         ...
         ValueError: input order is not prime
 
-        sage: E = EllipticCurve_with_prime_order(-7)
+        sage: E = next(EllipticCurve_with_prime_order(-7))
         Traceback (most recent call last):
         ...
         ValueError: input order is not prime
@@ -2901,7 +2912,7 @@ def EllipticCurve_with_prime_order(N):
         # done multiple times.
         for e in powerset(S):
             D = product(e)
-            if D % 8 != 5 or D >= 0 or D >= prime_start^2:
+            if D % 8 != 5 or D >= 0:
                 continue
 
             Q = BinaryQF([1, 0, -D])
@@ -2918,7 +2929,7 @@ def EllipticCurve_with_prime_order(N):
                         # `E.twists()` also contains E.
                         for Et in E.twists():
                             if Et.order() == N:
-                                return Et
+                                yield Et
 
         # At this point, no discriminant has been found, moving to next round
         # and extending the prime list.
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -2862,7 +2862,7 @@ def EllipticCurve_with_order(m, *, D=None):
                     except ValueError:
                         pass
 
-def EllipticCurve_with_prime_order(N):
+def EllipticCurve_with_prime_order(N, max_D=10**8):
     r"""
     Given a prime number ``N``, find another prime number `p` and construct an
     elliptic curve `E` defined over `\mathbb F_p` such that
@@ -2877,15 +2877,24 @@ def EllipticCurve_with_prime_order(N):
 
     EXAMPLES::
 
-        sage: N = next_prime(int(b'sagemath'.hex(), 16))
+        sage: N = next_prime(int.from_bytes(b'sagemath', 'big'))
         sage: E = next(EllipticCurve_with_prime_order(N))
         sage: E
-        Elliptic Curve defined by y^2 = x^3 + 4757897140353078952*x +
-        1841350074072114366 over Finite Field of size 8314040074357871443
+        Elliptic Curve defined by y^2 = x^3 + 4757897140353078952*x + 1841350074072114366
+         over Finite Field of size 8314040074357871443
         sage: E.order() == N
         True
 
-    You can directly iterate over the function call; here only on the first 10
+    The returned curves are sometimes random because
+    :meth:`~sage.schemes.elliptic_curves.ell_finite_field.EllipticCurve_finite_field.twists` is
+    not deterministic. However, it's always isomorphic::
+
+        sage: E = next(EllipticCurve_with_prime_order(23)); E                           # random
+        Elliptic Curve defined by y^2 = x^3 + 12*x + 11 over Finite Field of size 17
+        sage: E.is_isomorphic(EllipticCurve(GF(17), [3, 5]))
+        True
+
+    You can directly iterate over the iterator; here only on the first 10
     curves::
 
         sage: N = 54675917
@@ -2920,13 +2929,12 @@ def EllipticCurve_with_prime_order(N):
     However, experimentally it returns many of them. Here it returns all of them::
 
         sage: N = 23
+        sage: set_random_seed(1337)  # as the function returns random twists of curves
         sage: curves = list(EllipticCurve_with_prime_order(N)); curves
-        [Elliptic Curve defined by y^2 = x^3 + 20*x + 20 over Finite Field of size 23,
-         Elliptic Curve defined by y^2 = x^3 + 10*x + 16 over Finite Field of size 23,
-         Elliptic Curve defined by y^2 = x^3 + 14*x + 14 over Finite Field of size 17,
-         Elliptic Curve defined by y^2 = x^3 + 16*x + 25 over Finite Field of size 31,
-         Elliptic Curve defined by y^2 = x^3 + 13*x + 6 over Finite Field of size 19,
-         Elliptic Curve defined by y^2 = x^3 + 24*x + 3 over Finite Field of size 29]
+        [Elliptic Curve defined by y^2 = x^3 + 3*x + 5 over Finite Field of size 17,
+         Elliptic Curve defined by y^2 = x^3 + 19*x + 14 over Finite Field of size 31,
+         Elliptic Curve defined by y^2 = x^3 + 2*x + 9 over Finite Field of size 19,
+         Elliptic Curve defined by y^2 = x^3 + 7*x + 18 over Finite Field of size 29]
         sage: import itertools
         sage: # These are the only primes, by the Weil-Hasse bound
         sage: for q in prime_range(17, 35):
@@ -2937,7 +2945,43 @@ def EllipticCurve_with_prime_order(N):
         ....:         if E.order() == N:
         ....:             assert any(E.is_isomorphic(E_) for E_ in curves)
 
+    The algorithm is efficient for small ``N`` due to the low number of
+    suitable discriminants (see the ``recur`` internal function of the code for
+    details). The following runs in less than a second::
 
+        sage: len(list(EllipticCurve_with_prime_order(next_prime(5000))))
+        402
+
+    There is different verbose data for level `2` and `3`. Note that if you try
+    to compute the 5th curve in the iterator, the algorithm takes a long time
+    to run, as it is trying to compute a large Hilbert class polynomial, as
+    noted in the verbose log::
+
+        sage: set_random_seed(1337)  # as the function returns random twists of curves
+        sage: for _, E in zip(range(3), EllipticCurve_with_prime_order(10^9 + 7)):
+        ....:     print(E)
+        Elliptic Curve defined by y^2 = x^3 + 703734957*x + 232615553 over Finite Field of size 999969307
+        Elliptic Curve defined by y^2 = x^3 + 278352808*x + 442354703 over Finite Field of size 999969307
+        Elliptic Curve defined by y^2 = x^3 + 22138181*x + 343211325 over Finite Field of size 999969307
+        sage: from sage.misc.verbose import set_verbose
+        sage: set_verbose(2)
+        sage: for _, E in zip(range(3), EllipticCurve_with_prime_order(10^9 + 7)):
+        ....:     print(E)
+        verbose 2 (2865: ell_finite_field.py, EllipticCurve_with_prime_order) Computing the Hilbert class polynomial H_-667
+        Elliptic Curve defined by y^2 = x^3 + 276281332*x + 798205466 over Finite Field of size 999969307
+        Elliptic Curve defined by y^2 = x^3 + 515293837*x + 588191363 over Finite Field of size 999969307
+        Elliptic Curve defined by y^2 = x^3 + 684893162*x + 679980762 over Finite Field of size 999969307
+        sage: set_verbose(3)
+        sage: for _, E in zip(range(3), EllipticCurve_with_prime_order(10^9 + 7)):
+        ....:     print(E)
+        verbose 3 (2865: ell_finite_field.py, EllipticCurve_with_prime_order) Considering 1th valid prime 19
+        verbose 3 (2865: ell_finite_field.py, EllipticCurve_with_prime_order) Considering 2th valid prime 23
+        verbose 3 (2865: ell_finite_field.py, EllipticCurve_with_prime_order) Considering 3th valid prime 29
+        verbose 3 (2865: ell_finite_field.py, EllipticCurve_with_prime_order) Testing D=-667
+        verbose 2 (2865: ell_finite_field.py, EllipticCurve_with_prime_order) Computing the Hilbert class polynomial H_-667
+        Elliptic Curve defined by y^2 = x^3 + 620755650*x + 816246309 over Finite Field of size 999969307
+        Elliptic Curve defined by y^2 = x^3 + 651416562*x + 875284469 over Finite Field of size 999969307
+        Elliptic Curve defined by y^2 = x^3 + 149283707*x + 72132991 over Finite Field of size 999969307
 
     TESTS::
 
@@ -2986,6 +3030,7 @@ def EllipticCurve_with_prime_order(N):
     ALGORITHM: [BS2007]_, Algorithm 2.2
     """
     import itertools
+    from sage.misc.misc_c import prod
     from sage.misc.verbose import verbose
     from sage.combinat import gray_codes
     from sage.sets.primes import Primes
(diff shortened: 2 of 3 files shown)


GiacomoPope:
From my perspective the work was done during a Sage days and most probably can be improved but I would be interested in tracing that over flow error... I thought we tested it for larger values and it was fine and fast enough. Don't know...

JosePisco:
> From my perspective the work was done during a Sage days and most probably can be improved but I would be interested in tracing that over flow error... I thought we tested it for larger values and it was fine and fast enough. Don't know...

Now that I read my comment again, I might have wrongfully expressed myself. I wanted to argue why this code that would intuitively fit inside your work is written separately.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2787,3 +2789,115 @@ def find_q(m, m4_fac, D):
...
+        Traceback (most recent call last):
+        ...
+        ValueError: input order is not prime
+
+        sage: E = EllipticCurve_with_prime_order(0)
+        Traceback (most recent call last):
+        ...
+        ValueError: input order is not prime
+
+    .. NOTE::
```
vincentmacri:
```suggestion
    .. WARNING::
```

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2787,3 +2789,115 @@ def find_q(m, m4_fac, D):
...
+        raise ValueError("input order is not prime")
+
+    # The algorithm consists of multiple ‘search rounds’ for a suitable
+    # discriminant `D`, `r` defines the number of rounds. We expect this
+    # algorithm to terminate after a number of rounds that is polynomial in
+    # loglog N.
+    r = 0
+    S = []
+    # First prime. Iterating over the primes by chunks of size log(`N`).
+    p = next_prime(0)
```
vincentmacri:
```suggestion
    p = 2
```

vincentmacri: Section 2.2 of the paper says to only consider odd primes. Should this be `p = 3`?

vincentmacri (commented):
I don't think I'm familiar enough with Sage or what you're doing to approve this, but hopefully this feedback helps make the review easier for someone who is more familiar.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2787,3 +2789,115 @@ def find_q(m, m4_fac, D):
...
+
+    while True:
+        while p < (r + 1) * floor(N.log()):
```
vincentmacri: Might be a bit cleaner if you use the `primes` or `prime_range` function here. `prime_range` is faster but uses more memory. `primes` is an iterator so closer to what you're doing here.

vincentmacri: I think `S = prime_range(3, ceil((r + 1) * floor(N.log())) + 1)` should work (assuming it's supposed to start at 3). Then you can remove the whole while loop. I might have an off-by-one error or something here but I think it's correct.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: Nice catch thanks! I don't think it changes the results but it did change the computation time since no discriminant can be even.
cef063eae9e4a06a31144907dacdcd865c081764

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: Indeed, this is quite faster to use `prime_range()`.

Original 
```py
sage: %%timeit S = []
....: p = 3
....: r = 0
....: while p < (r + 1) * floor(log(N)):
....:      S.append(p)
....:      p = next_prime(p)
....: 
729 µs ± 32.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
compared to
```py
sage: %timeit S = prime_range(3, ceil((11 + 1) * log(N)))
35.1 µs ± 239 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

And it's interesting to notice that iterating over it and appending to a list is slower (neglecting the list initialization of approx ≃9 ns)
```py
sage: %%timeit
....: t = []
....: for p in prime_range(ceil(11 * log(N)), ceil((11 + 1) * log(N))):
....:     t.append(p)
....: 
42.6 µs ± 550 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

I suggest changes in cef063eae9e4a06a31144907dacdcd865c081764

Thanks for the review!

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: I'm not sure what's the fundamental difference here? It doesn't feel like a warning, same as `factor()` can take a very long time?

vincentmacri: The way it's worded ("function may run forever") makes it sound like the algorithm isn't proven to always terminate.

If it does terminate but possibly only after a very long time, then I think `NOTE` is okay. In that case word the note to say "long time" or something like that. If there are certain inputs that make it more likely to take a long time (for example, very large numbers) then it would be nice if the note mentioned which kinds of cases cause it to take a long time.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2787,3 +2789,125 @@ def find_q(m, m4_fac, D):
...
+
+    # The algorithm consists of multiple ‘search rounds’ for a suitable
+    # discriminant `D`, `r` defines the number of rounds. We expect this
+    # algorithm to terminate after a number of rounds that is polynomial in
+    # loglog N.
+    r = 0
+
+    while True:
+        # Iterating over the odd primes by chunks of size log(`N`).
+        S = prime_range(3, ceil((r + 1) * N.log()))
```
vincentmacri: If this is algorithm 2.2 in the paper, shouldn't the lower bound be `r log N`?

Also, `prime_range` rounds the inputs to the nearest integer and doesn't include the right endpoint (the rounding behaviour is a bit unintuitive in my opinion because you get things like `prime_range(2.1, 5.3)` is `[2, 3]`, but oh well).

So I think this should be:

~~`S = prime_range(max(ceil(r * N.log()), 3), floor((r + 1) * N.log()) + 1)`~~

If you have any test cases where the correct answer requires the low/high point of the range to be included, those would be great to add to the TESTS in the docstring.

vincentmacri: Oh, I misread the paper. You need to add to `S` on each iteration.

Maybe do `S.extend(prime_range(max(ceil(r * N.log()), 3), floor((r + 1) * N.log()) + 1))` instead of `S = [...]`.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2787,3 +2789,125 @@ def find_q(m, m4_fac, D):
...
+
+        # Every possible products of distinct elements of `S`.
+        # There probably is a more optimal way to compute all possible products
+        # of elements of S than using a powerset. Here many multiplications are
+        # done multiple times.
+        for e in powerset(S):
+            D = product(e)
```
vincentmacri: In the paper, the algorithm takes a product of the `p^*` terms, not the `p` terms. Should there be a computation of the `p^*` terms somewhere? I also don't see where you do the square root mod `N` part of the algorithm.

grhkm21:
> This PR would intuitively fit nice into their work but I felt uncomfortable with it returning an iterator. I felt like returning a single curve was more handy so I implemented this method in a separate function that does so but I'm open to suggestions if this is of any interest to the community.

I disagree, I mean CM method inherently generates multiple curves, so it makes sense to return an iterator. If you want a single curve, just use the (builtin) `next` function.

grhkm21:
Also I will have to look closer at the paper and the implementation, but why don't use you just use `EllipticCurve_with_order`? It seems to implement the same CM stuff based on `hilbert_polynomial`.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: I was not so fine either with it and the fact that you point it out confirms it. What about fddd496e474a73745a9267e9f745f67dd6b24612 ?

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: > shouldn't the lower bound be r log N?

Yes, and as you mentioned, you are increasing the basis of primes at each round. Though I think recreating the entire table is fine because `log N` will never be big to slow the first primes computation and since we expect `loglog N` rounds, I don't think it will make such a difference. However, to stick to the algorithm, I suggest changes in 45b053a248b5e98c7afb11a88400102595a7a00d.

> Also, prime_range rounds the inputs to the nearest integer and doesn't include the right endpoint

`ceil((r + 1) * N.log())` does just fine.

> If you have any test cases where the correct answer requires the low/high point of the range to be included, those would be great to add to the TESTS in the docstring.

This means to have D to have a prime factor of `log N`. The test
```py
sage: for N in prime_range(3, 100):
....:     E = EllipticCurve_with_prime_order(N)
....:     assert E.order() == N
```
should take care of this. It also tests the twists.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: The square roots are used to solve the diophantine equation for `4 * N` and `BinaryQF` does it for us.

The `p*` are defined as `p* = (-1)^(p-1 / 2) * p`. Basically, it checks the parity of `p-1 / 2` to negate or not `p`. This is a shady part for me because this wouldn't work and it kept yielding non-discriminants for the Hilbert polynomial. I found a hack around that seems to work but maybe I'm wrong here.

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: > > Also, prime_range rounds the inputs to the nearest integer and doesn't include the right endpoint
> 
> `ceil((r + 1) * N.log())` does just fine.

When I wrote that I was thinking about the case where `(r + 1) * N.log()` is an integer, but that can't happen since `N` is an integer, so yeah the ceiling here is fine.

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: This is great, just adding a suggestion for formatting.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2787,3 +2789,135 @@ def find_q(m, m4_fac, D):
...
+        Traceback (most recent call last):
+        ...
+        ValueError: input order is not prime
+
+    .. NOTE::
+
+        Depending on the input, this function may run for a *very* long time.
+        This algorithm consists of multiple "search rounds" for a suitable
+        discriminant `D`. We expect this algorithm to terminate after a number
+        of rounds that is polynomial in `loglog N`. In practice (cf. Section 5),
```
vincentmacri:
```suggestion
        of rounds that is polynomial in `\log(\log N)`. In practice (cf. Section 5),
```

LaTeX formatting is encouraged in the style guide. https://doc.sagemath.org/html/en/developer/coding_basics.html#documentation-strings

vincentmacri (commented):
Looks good to me, but someone who better understands the algorithm and paper you're working off of should take a look too.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: Thanks for this, I'm still quite new to this kind of formatting (see 11270fb3999d20f594bcfd76741716d974d9781b).

JosePisco:
I have tried to comply 100% with the algorithm from the paper [Constructing Elliptic Curves of prime order](https://arxiv.org/abs/0712.2022) by Bröker and Stevenhagen i.e.

- Taking $p$ or $-p$ depending on the parity of $(p-1) / 2$
- Checking if $D \equiv 5 \mod 8$ and not $-D$
- Solving the diophantine equation $x^2 - Dy^2 = 4N$ and not $x^2 + Dy^2 = 4N$

I can't explain why the actual implementation works while the attempts at strictly following the algorithm don't.
I provide a patch that applies the changes described above if anybody is curious (but it fails the tests).
[changes.diff](https://github.com/user-attachments/files/16636326/changes.txt)

grhkm21:
Hi, I'm ready to review this. As @vincentmacri mentioned, you implemented the algorithm incorrectly. I am able to get an algorithm working using $p^*$ instead of $p$ terms. I also added a few performance tricks here and there, mostly the `curve_has_order` auxiliary function (it's a bit slow because of the try-except but maybe copy the code from `.set_order` that I modified a few months back), and also you can specify to use `algorithm="cornacchia"`.

And also as I said above, please return an iterator instead of a single curve, just like many other algorithms do.

```sage
sage: def curve_has_order(E, N):
....:     if any(E.random_point() * N != 0 for _ in range(20)):
....:         return False
....:     try:
....:         E.set_order(N)
....:         return True
....:     except ValueError:
....:         return False
....: 
....: def f(N):
....:     assert is_prime(N)
....: 
....:     r = 0
....:     while r < 1:
....:         primes = []
....:         for p in prime_range(max(3, floor(r * log(N))), floor((r + 1) * log(N))):
....:             if legendre_symbol(N, p) == 1:
....:                 p_ = p if p % 4 == 1 else -p
....:                 primes.append((p_, GF(N)(p_).sqrt()))
....:         for subset in powerset(primes):
....:             D, sqrtD = prod(u[0] for u in subset), prod(u[1] for u in subset)
....:             if D % 8 != 5 or D >= 0:
....:                 continue
....:             Q = BinaryQF([1, 0, -D])
....:             sol = Q.solve_integer(4 * N, algorithm="cornacchia")
....:             if sol is None:
....:                 continue
....:             x, y = sol
....:             assert x**2 - D * y**2 == 4 * N
....:             for p in [N + 1 + x, N + 1 - x]:
....:                 if not is_prime(p):
....:                     continue
....:                 h = hilbert_class_polynomial(D)
....:                 K = GF(p)
....:                 for j in h.roots(ring=K, multiplicities=False):
....:                     E = EllipticCurve(K, j=j)
....:                     for E_ in E.twists():
....:                         if curve_has_order(E_, N):
....:                             yield E_
....:         r += 1
....: 
....: N = 123456789012345678901234567890123456789012345678901234568197
....: it = f(N)
....: [next(it) for _ in range(5)]
[Elliptic Curve defined by y^2 = x^3 + 70969083882016979692051040054897244304487459580456198765042*x + 23963839073844928337866754804687093774080822288916599802630 over Finite Field of size 123456789012345678901234567890654833374525085966737125236501,
 Elliptic Curve defined by y^2 = x^3 + 32301896591105181383197983392011966935739308240089212446888*x + 20440933360506659987728474271841544365948756313127780399770 over Finite Field of size 123456789012345678901234567890654833374525085966737125236501,
 Elliptic Curve defined by y^2 = x^3 + 49640879391385567455601737809134376710338058924528599296877*x + 57903173466042191227456380331494397476264372226969014648600 over Finite Field of size 123456789012345678901234567890654833374525085966737125236501,
 Elliptic Curve defined by y^2 = x^3 + 113468659391060172590889752354736732425162987733160523980142*x + 72243181080242569121408088391312469724962221744363796902853 over Finite Field of size 123456789012345678901234567890654833374525085966737125236501,
 Elliptic Curve defined by y^2 = x^3 + 75763109110881858499138647331042804275382035844950124504205*x + 24123673357232490706020024337399069487889990527522975347683 over Finite Field of size 123456789012345678901234567890654833374525085966737125236501]```

**Comment by JosePisco** [comment]
Thanks for the review!

From your comment and code and my notes I was able to propose commit db678986c3931d35ad3e9f3f1d07a9c748d93a17 complying entirely with Algorithm 2.2 from the paper.

Few things I'm not convinced about:
- Your `curve_has_order()` function is probabilistic? I get the method and believe it's a smart way to do it but is sampling 20 points enough? And do we not fall even with `.order()` while trying with `E.set_order(N)` after?
- I'm against returning an iterator; simply from a pragmatic point of view we want a constructor for elliptic curves to return one instance of a curve, not a range of them.

**Comment by vincentmacri** [comment]
>     * I'm against returning an iterator; simply from a pragmatic point of view we want a constructor for elliptic curves to return one instance of a curve, not a range of them.

I don't have a strong opinion either way, but unless there's some kind of canonical curve with the given prime order, what's wrong with an iterator? If a user is fine with an arbitrary one, they could use `next(elliptic_curves_with_prime_order(N))` to easily get the first one.

**Comment by JosePisco** [comment]
> unless there's some kind of canonical curve with the given prime order
There isn't. As @grhkm21 mentionned: "CM method inherently generates multiple curves" and that makes it more logical to return an iterator on a theoretical point of view.

My point is on a programming and software point of view: we are defining a special constructor of `EllipticCurve` and that would make sense to align with other constructors that just return an `EllipticCurve` and not an iterator.

Maybe we could satisfy everybody by introducing a parameter `iterator` in `EllipticCurve_with_prime_order(N, iterator=False)`. So that the user can chose if he wants an iterator or not. I get that people can just use `next(EllipticCurve_with_prime_order(N))` at this point but it feels like we stray further from other elliptic curves constructors in sage.

Open debate, please take part :)

**Comment by JosePisco** [comment]
We should also discuss how to efficiently check the order of the curve candidates.
@grhkm21 proposed to probabilistically check curves' orders instead of computing them and comparing naively. This method is smarter **and** faster only if `set_order()` is deterministic.
```py
def curve_has_order(E, N):
    if any(E.random_point() * N != 0 for _ in range(20)):
        return False
    try:
        E.set_order(N)
        return True
    except ValueError:
        return False
```
However, grhkm found out while discussing this with me that it was deterministically incorrect in some *rare* cases (see #38617). 

So I think it would be nice to discuss which method to adopt for this case (and given how the issue is addressed, maybe implement `E.has_order(N)` for elliptic curves).

vincentmacri:
> My point is on a programming and software point of view: we are defining a special constructor of `EllipticCurve` and that would make sense to align with other constructors that just return an `EllipticCurve` and not an iterator.

I agree a constructor shouldn't return an iterator. But why does this need to be a constructor? We already have `elliptic_curves.rank` which returns a list of curves with the specified rank and torsion order. Instead of adding the constructor `EllipticCurve_with_prime_order`, why not add the function `elliptic_curves.prime_order` which returns an iterator?

vincentmacri:
> why not add the function `elliptic_curves.prime_order` which returns an iterator?

Maybe not that specific name since it seems `elliptic_curves` is for stuff involving an elliptic curves database, but my point is why not view this as a utility function rather than a constructor.

grhkm21:
This should definitely be viewed as an utility function rather than a constructor. I still don't see any valid argument for returning a curve rather than an iterator.

grhkm21:
> We should also discuss how to efficiently check the order of the curve candidates. @grhkm21 proposed to probabilistically check curves' orders instead of computing them and comparing naively. This method is smarter **and** faster only if `set_order()` is deterministic.
> 
> ```python
> def curve_has_order(E, N):
>     if any(E.random_point() * N != 0 for _ in range(20)):
>         return False
>     try:
>         E.set_order(N)
>         return True
>     except ValueError:
>         return False
> ```
> 
> However, grhkm found out while discussing this with me that it was deterministically incorrect in some _rare_ cases (see #38617).
> 
> So I think it would be nice to discuss which method to adopt for this case (and given how the issue is addressed, maybe implement `E.has_order(N)` for elliptic curves).

@JosePisco I just realised that, no, #38617 doesn't affect our case. This is because prime order ~~curves~~ abelian groups are always cyclic *and* `N` is prime for this method, so `set_order` is correct. In fact, the following code is correct

```sage
from sage.groups.generic import has_order
def curve_has_order_prime(E, p):
    assert is_prime(p)
    while (P := E.random_point()).is_zero(): pass
    return has_order(P, p)
```

I think you have to treat the case where `E` is defined over $\mathbb{F}_2$ or $\mathbb{F}_3$ separately, where it's possible for the curve to only contain the point at infinity, but the easier option is to skip $p = 3$ in your function as well :)

grhkm21:
Also it seems how you/I implement the loop is buggy, because on the second iteration of the `while True` when you loop over the powerset, you will also loop over the powerset of the first iteration. More simply, if you run the following code, it outputs duplicate curves

```sage
it = EllipticCurve_with_prime_order(5)
while True: print(next(it))
```

JosePisco:
> Also it seems how you/I implement the loop is buggy, because on the second iteration of the `while True` when you loop over the powerset, you will also loop over the powerset of the first iteration. More simply, if you run the following code, it outputs duplicate curves
> 
> ```python
> it = EllipticCurve_with_prime_order(5)
> while True: print(next(it))
> ```

That's true, I just found out. The algorithm wants to extend the prime list but we treat the powerset the same way. I need to think about a way around to avoid computing same multiplications or if you have ideas please share.

grhkm21:
Please don't do whatever you did that caused the force push.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2788,155 @@ def find_q(m, m4_fac, D):
...
+    Given a prime number ``N``, find another prime number `p` and construct an
+    elliptic curve `E` defined over `\mathbb F_p` such that
+    `\#E(\mathbb F_p) = N`.
+
+    INPUT:
+
+    - ``N`` -- integer; the order for which we seek an elliptic curve. Must be a
+      prime number.
+
+    OUTPUT: an iterator of elliptic curves `E/\mathbb F_p` of order ``N``
```
grhkm21: Can you comment on whether the iterator eventually returns all such curves? Recall that there are finitely many of a given order.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2788,155 @@ def find_q(m, m4_fac, D):
...
+
+            x, _ = sol
+            for p_i in [N + 1 - x, N + 1 + x]:
+                if is_prime(p_i):
+                    H = hilbert_class_polynomial(D)
+                    for j0 in H.roots(ring=GF(p_i), multiplicities=False):
+                        E = EllipticCurve(j=j0)
+                        # `E.twists()` also contains E.
+                        for Et in E.twists():
+                            if Et.order() == N:
```
grhkm21: As I commented, this is very inefficient and should be replaced with my fix after changing `prime_start` to 5.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2788,155 @@ def find_q(m, m4_fac, D):
...
+            sol = Q.solve_integer(4 * N, algorithm='cornacchia')
+            if sol is None:
+                continue
+
+            x, _ = sol
+            for p_i in [N + 1 - x, N + 1 + x]:
+                if is_prime(p_i):
+                    H = hilbert_class_polynomial(D)
+                    for j0 in H.roots(ring=GF(p_i), multiplicities=False):
+                        E = EllipticCurve(j=j0)
```
grhkm21: Can you make this `EllipticCurve(K, j=j0)` instead just for clarity?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2788,155 @@ def find_q(m, m4_fac, D):
...
+        # There probably is a more optimal way to compute all possible products
+        # of elements of S than using a powerset. Here many multiplications are
+        # done multiple times.
+        for e in powerset(S):
+            D = product(e)
+            if D % 8 != 5 or D >= 0:
+                continue
+
+            Q = BinaryQF([1, 0, -D])
+            sol = Q.solve_integer(4 * N, algorithm='cornacchia')
```
grhkm21: Note that for $x^2 + (-D)y^2 = 4N$ to have a solution, you need $-D \leq 4N$, so that is an easy check as well. In fact this gives a bound for your loop on line 2902, to make the algorithm eventually terminate instead of running infinitely (while not yielding any curve)

grhkm21 (changes requested):
To resolve the issue of duplicated curves, or just the whole prime choice thing in general, maybe take a step back and think about what `e` is taking value over. It's just the power set of the odd primes in the natural ascending (?) order. So in fact I don't see the point of segmenting into `[prime_start, prime_end]` intervals, unless you're making it probabilistic and sampling random subsets, which you aren't. So I suggest deleting `prime_start` and `prime_end` entirely, append primes to `S` one by one, and in the loop for `e` (and `D`), just mandate the last prime added to be included. Then you can halt the algorithm when $D$ is too large as mentioned.

If you write your own power-set generator algorithm then you can make it break faster for parity reasons but I don't think it matters too much here.

src/sage/schemes/elliptic_curves/all.py
```diff
@@ -27,7 +27,7 @@
 lazy_import('sage.schemes.elliptic_curves.jacobian', 'Jacobian')
 
 lazy_import('sage.schemes.elliptic_curves.ell_finite_field', 'special_supersingular_curve')
-
+lazy_import('sage.schemes.elliptic_curves.ell_finite_field', 'EllipticCurve_with_prime_order')
```
grhkm21: Replace this with `lazy_import('...', ['a', 'b'])` just like below.

grhkm21 (changes requested):

grhkm21:
@JosePisco  has notified me that he isn't free the next few weeks, so I applied the following patch file and created this patch file (waiting for him to apply or give me push access). 

https://gist.github.com/grhkm21/12b3b96b309d29ce3f9f213514316253

As a result, I won't be reviewing this PR either. @vincentmacri can you review it when it's ready? Or you can apply the patch and review that

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,96 @@ def is_ordinary(self, proof=True):
         """
         return not is_j_supersingular(self.j_invariant(), proof=proof)
 
+    def has_order(self, value, num_checks=8):
```
vincentmacri, line 1346: Do you want this to be a publicly-documented function or just an internal one? If internal, rename to `_has_order`. If you wanted it to be public then it's fine.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,96 @@ def is_ordinary(self, proof=True):
...
+
+            sage: # long time (10s)
+            sage: p = random_prime(2**128, lbound=2**127)
+            sage: K = GF((p, 2), name="a")
+            sage: E = EllipticCurve(K, [K.random_element() for _ in range(2)])
+            sage: N = E.order()
+            sage: E.has_order(N, num_checks=20)
+            True
+            sage: E.has_order(N + 1)
+            False
```
vincentmacri, line 1393: Can you add an example of a curve that is big enough that this function is useful? The example in the docstring for `set_order` should suffice.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,96 @@ def is_ordinary(self, proof=True):
...
+        # This method does *not* use the cached value of `_order` even when
+        # available.
+        q = self.base_field().order()
+        a, b = Hasse_bounds(q, 1)
+        if not a <= value <= b:
+            return False
+
+        # For really small values, the random tests are too weak to detect wrong
+        # orders So we go with computing directly instead.
+        if q <= 100:
```
vincentmacri:
```suggestion
        if q <= 2^128:
```

`EllipticCurve([F.random_element(), F.random_element()]).order()` takes under a second (141 ms) for me for `F = GF(previous_prime(2^128))`, I think you can make the bound here much larger than 100. I'm not tied to `2^128` specifically, but 100 seems way too small.

For `F = GF(previous_prime(2^64))` it's pretty much instant (6.62 ms).

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+    Given a prime number ``N``, find another prime number `p` and construct an
+    elliptic curve `E` defined over `\mathbb F_p` such that
+    `\#E(\mathbb F_p) = N`.
+
+    INPUT:
+
+    - ``N`` -- integer; the order for which we seek an elliptic curve. Must be a
+      prime number.
+
+    OUTPUT: an iterator of elliptic curves `E/\mathbb F_p` of order ``N``
```
vincentmacri: Echoing what @grhkm21 said in a previous comment.

```suggestion
    OUTPUT: an iterator of (some) elliptic curves `E/\mathbb F_p` of order ``N``
```

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+
+    .. NOTE::
+
+        Depending on the input, this function may run for a *very* long time.
+        This algorithm consists of multiple "search rounds" for a suitable
+        discriminant `D`. We expect this algorithm to terminate after a number
+        of rounds that is polynomial in `\log\log N`. In practice (cf. Section
+        5), this number is usually 1.
```
vincentmacri:
```suggestion
        of rounds that is polynomial in `\log\log N`. In practice, this number
        is usually 1.
```

I don't like referring to a section of a paper without specifying which paper.

Alternatively:

```suggestion
        of rounds that is polynomial in `\log\log N`. In practice, this number
        is usually 1 (see Section 5 of [BS2007]_).
```

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+
+    ALGORITHM: [BS2007]_, Algorithm 2.2
+
+    EXAMPLES::
+
+        sage: N = next_prime(int.from_bytes(b'sagemath', 'big'))
```
vincentmacri:
```suggestion
        sage: N = 8314040072427107567
```

We aren't doing cryptography here, there's no need for "nothing up my sleeve" numbers. It just makes the docs harder to read.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
                         yield Et
                     except ValueError:
                         pass
+
+def EllipticCurve_with_prime_order(N):
```
vincentmacri, line 2879:
```suggestion
def elliptic_curves_with_prime_order(N):
```

and make the corresponding change wherever the function is used.

Unless there is some (well) established naming convention, I don't like the current naming because it sounds like it constructs a single elliptic curve, when in reality it is an iterator of elliptic curves. If some naming convention I'm unaware of necessitates the capital letters for the name, can we at least have `EllipticCurves_with_prime_order`?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+    You can directly iterate over the iterator; here only on the first 10
+    curves::
+
+        sage: N = 54675917
+        sage: for _, E in zip(range(10), EllipticCurve_with_prime_order(N)):
+        ....:     assert E.has_order(N)
+
+    It works for large primes::
+
+        sage: N = 0x6cbc824032974516623e732462f4b74b56c4ffbd984380d9
```
vincentmacri:
```suggestion
        sage: N = 2666207849820848272386538889527600954292544013630953455833
```

I wouldn't expect all Sage users to be familiar with Python hex syntax (or even hex), and this fits within the 80 character limit anyway.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+    S = [(-p if p % 4 == 3 else p) for p in prime_range(3, min(1000, 4 * N))
+         if legendre_symbol(N, p) == 1]
+
+    def abs_products_under(bound):
+        """
+        This function returns an iterator of all numbers with absolute value not
+        exceeding ``bound`` expressable as product of distinct elements in ``S``
+        in ascending order.
+        """
+        import heapq
```
vincentmacri, line 3123: I'm guessing the reason for this import is that using `hq.pop(0)` would be too slow?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+    """
+    import itertools
+    from sage.arith.misc import is_prime, legendre_symbol
+    from sage.misc.verbose import verbose
+    from sage.quadratic_forms.binary_qf import BinaryQF
+    from sage.rings.fast_arith import prime_range
+    from sage.schemes.elliptic_curves.cm import hilbert_class_polynomial
+    from sage.sets.primes import Primes
+
+    if not is_prime(N):
```
vincentmacri, line 3099: Since this code relies on a few probabilistic algorithms, including `has_order` which sometimes returns the wrong answer, I think it would be worth adding an optional `proof` argument to this function and using it for this prime check and any calls to `has_order`.

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri, line 1346: Add an optional `proof` argument which should default to `True`.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+            EllipticCurve(GF(3), [0, 2, 0, 0, 2]),
+            EllipticCurve(GF(5), [2, 0])
+        ]
+        return
+
+    # We start with small primes directly to accelerate the search. Note that
+    # 1000 is a magic constant, it's just fast enough to compute without
+    # sacrificing much speed.
+    # The if-then-else term is (-1)^((p - 1) / 2) * p in [BS2007]_ page 5.
+    S = [(-p if p % 4 == 3 else p) for p in prime_range(3, min(1000, 4 * N))
```
vincentmacri, line 3114: Why do you construct `S` differently than in the paper? I.e. why aren't you iterating over `p` in `[r \log N, (r + 1) \log N]`? Could you add an explanation in the comments here?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+            abs_n, n, idx = heapq.heappop(hq)
+            yield n
+            for nxt in range(idx + 1, len(S)):
+                if abs_n * abs(S[nxt]) <= bound:
+                    heapq.heappush(hq, (abs_n * abs(S[nxt]), n * S[nxt], nxt))
+                else:
+                    break
+
+    # We add p = 1 to process the small primes.
+    for p in itertools.chain([1], Primes()):
```
vincentmacri, line 3135: This algorithm won't terminate (fine because it's an iterator). Can you explain in the comments this deviation from the paper?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2861,290 @@ def find_q(m, m4_fac, D):
...
+
+    .. NOTE::
+
+        Depending on the input, this function may run for a *very* long time.
+        This algorithm consists of multiple "search rounds" for a suitable
+        discriminant `D`. We expect this algorithm to terminate after a number
+        of rounds that is polynomial in `\log\log N`. In practice (cf. Section
+        5), this number is usually 1.
+
+    ALGORITHM: [BS2007]_, Algorithm 2.2
```
vincentmacri:
```suggestion
    ALGORITHM: Based on [BS2007]_, Algorithm 2.2
```

There are enough deviations that I wouldn't call this quite the same algorithm.

vincentmacri (changes requested):
I haven't gotten to the main logic yet. After you incorporate these changes I'll take another (closer) look.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: Because basically the paper used primes in ranges of order logarithmic in N to give an estimation of the number of rounds until we find a curve (loglog N). You can see it as a way to turn the algorithm to give an argument to when it finishes.

Here since we opted to find them all, we can omit this and normally iterate over the primes one at a time.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: Yes it will, because when $p$ is large, it breaks (see below, when $p > 4N$). That's why `list(EllipticCurve_with_prime_order(N))` is finite.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: Basically we introduced `has_order` as a way to check a candidate curve has the right prime order `N`.
The method we used before was naive and much more costly: compute order of the curve and compare to `N`.
Now we want to check that the curve has the right order so we sample random points and check that they are of order `N` which is much easier to do (only a few double and add). The number of samples can be reduced to 1 since N is prime and we use `set_order` after this and it's supposed to yield a deterministic result.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: > Why do you construct `S` differently than in the paper? I.e. why aren't you iterating over `p` in `[r \log N, (r + 1) \log N]`? Could you add an explanation in the comments here?

In short, for most (moderate size, say $< 2^{512}$) value of $N$ (experimentally), I find that the minimal suitable discriminant `D` has largest prime factor < 1000. Hence if we enumerate $1000$-smooth discriminants $D$ we will have a high probability of finding a curve.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: That's out of scope for this PR, the method is probabilistic as noted in `.. NOTE:` already.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: That is incorrect, please read my notes above. The `.has_order` call right now is *deterministically (100%) correct*.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: No, it's for `heapq.heappush`, so that we enumerate $D$ in increasing absolute value order.

src/sage/schemes/elliptic_curves/ell_finite_field.py
JosePisco: Mmmh, since we don't do the "rounds" anymore, this comment is probably outdated or we could update the expectation using results from the paper.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: I will consult opinions from other Sage contributors, but seeing that `EllipticCurve_with_order` got merged in another PR, I don't see any problem with the naming.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: 6.62ms is quite far away from instant. Also, I am unable to reproduce your timings on my laptop, though admittedly it's a slow one.

```sage
sage: p = random_prime(2**64, lbound=2**63)
....: E = EllipticCurve(GF(p), j=randrange(p))
....: %time N = E.order(); N
....: %time E.has_order(N)
....: %time _ = E.random_point()
CPU times: user 44.7 ms, sys: 3.31 ms, total: 48 ms
Wall time: 55.8 ms
12084593079744292894
CPU times: user 5.8 ms, sys: 0 ns, total: 5.8 ms
Wall time: 5.87 ms
True
CPU times: user 214 μs, sys: 0 ns, total: 214 μs
Wall time: 217 μs```

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 13 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:1393]: How is $p \sim 2^{128}$ not "big enough"?

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 12 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:1346]: Yes, the method should be public, and the naming is correct.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 13 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:1393]: You're right, not sure what I was thinking. Maybe I misread this part. I still have a question about this example.

~~Since this algorithm sometimes fails and this test is random, are we potentially introducing a test that might randomly fail? That could be annoying for the CI. I'd prefer something like `p = nextprime(2^127)` or something than a test that might randomly fail.~~

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 14 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py]: How about `q < 2^32` then? Or even `q < 1000`. I'd like if this was provably correct at least for the sizes of examples someone might do by hand.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 18 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:2879]: If there is an established convention here then you don't need to change it. I'd still prefer `EllipticCurves_with_prime_order` (plural curves) though but not a huge deal.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 21 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3099]: I'm confused. Doesn't it give the wrong answer for this example?

```
sage: E = EllipticCurve(GF(127), [0, 1])
sage: E.order()
108
sage: E.has_order(126)
True
```

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 20 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3123]: Why not use `hq.append`?

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 21 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3099]: image

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 20 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3123]: @vincentmacri do you know about priority queues...?

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 14 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py]: I will change it for now, but expect it to be reverted back once # 38617 is resolved.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 13 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:1393]: Nevermind, I think I misunderstood the failure example. This `has_order` can give false positives but never false negatives, is my understanding correct?

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 16 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py]: I will delete the note entirely, it's covered in the tests already

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 13 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:1393]: Your understanding is correct, but it's not why this test doesn't fail. For that you have to understand the bug analysis in # 38617. To summarise it, for `E.has_order(N)` to pass, you need $\mathrm{LCM}(\mathrm{ord}(P)) \mid N$ over all $P$ sampled in the random test. Here since $E$ has ("real") prime order $N$, the LCM would also be $N$ (because $P \neq O_E$ in `has_order`, patched in this PR as well). But $N \mid N + 1$ never holds for prime $N$ so the test never fails.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 21 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3099]: So #38617 is unrelated? Your docstring says "This demonstrates the bug mentioned in the NOTE above. The return value should be ``False`` after :issue:`38617` is fixed" for one of your examples.

Can you either update the docstring for that to explain the error better or if able, fix the bug causing the false positive?

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 22 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3114]: Sounds good, please add an explanation of this in the comments in the code.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 20 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3123]: > @vincentmacri do you know about priority queues...?

Not super familiar, my background is mainly mathematical as well as some low-level cryptography programming so not too much in terms of data structures.

I'll trust that this part is fine, just wanted to make sure the extra import had a reason.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 23 above)
Comment (vincentmacri) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3135]: > Yes it will, because when p is large, it breaks (see below, when p > 4 N ). That's why `list(EllipticCurve_with_prime_order(N))` is finite.

Missed that part, seems good. Please still add an explanation in the comments about how this gives multiple curves whereas the algorithm in the paper only gives one.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 21 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3099]: I think you're confusing two pieces of code here.

One part is the `has_order` function, which is implemented for all elliptic curves. That returns false positives (`True` when real answer is `False`) sometimes, which is bug #38617. That is documented correctly in the docstrings.

However, the bug only affects curves with a specific structure. In particular, when the curve has prime order or the tested value is prime (and at least $5$ or something), the `has_order` function always return the correct value. In fact it does with `num_checks=1`.

This leads to the other part when I use `has_order`, which is in `EllipticCurve_with_prime_order`. There the tested value is prime, so as mentioned above, there is no bug, so `EllipticCurve_with_prime_order` has no bug. This is why the `proof` argument you suggested is unnecessary.

File: src/sage/schemes/elliptic_curves/ell_finite_field.py (hunk 22 above)
Comment (grhkm21) [reply src/sage/schemes/elliptic_curves/ell_finite_field.py:3114]: I have added it now, it's a bit lengthy but should be okay

File: src/sage/schemes/elliptic_curves/ell_finite_field.py [hunk 25]
```diff
@@ -2786,3 +2863,313 @@ def find_q(m, m4_fac, D):
...
+    rather `p^*`). We find this difficult to implement without testing
+    duplicate `D`s, so we instead enlarge the table one prime at a time
+    (effectively replacing `[r\log(N), (r + 1)\log(N)]` in the paper by `[r,
+    r]`). To compensate for the speed loss, we begin the algorithm by
+    prefilling `S` with the primes below `1000` (satisfying quadratic
+    reciprocity properties). The constant `1000` is determined experimentally
+    to be fast for many purposes, and for most `N` we tested we are able to
+    find a suitable small `D` without increasing the size of `S`.
+
+<<<<<<< HEAD
```
vincentmacri:
```suggestion
```

Guessing this made it in when doing some git operations?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2786,3 +2863,313 @@ def find_q(m, m4_fac, D):
...
+    the algorithm because smaller `D` means the Hilbert class polynomial is
+    computed quicker.
+
+    Finally, to avoid repeatedly testing the same `D`s, we require the latest
+    prime to be added to the table to be included as a factor of `D` (see code
+    for more explanation). As we need to find integers `x, y` such that `x^2 +
+    (-D)y^2 = 4N` with `D < 0` and `N` prime, we actually need `|D| \leq 4N`,
+    so we terminate the algorithm when the primes in the table are larger than
+    that bound. This makes the iterator return all curves it can find in finite
+    time :)
```
vincentmacri:
```suggestion
    time.
```

grhkm21:
Hey @vincentmacri , I added a few changes to `EllipticCurve_with_order` (not `with_prime_order`) as well, mainly refactoring + reversing the order of `D` to start searching from smaller absolute value `D`. I am a coauthor of that method so I hope you can review that too ^_^ (it's very similar in nature)

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: As much as I wish software documentation had more emojis, I don't think that's how Sage docs are written.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2772,17 +2852,324 @@ def find_q(m, m4_fac, D):
...
+                            level=2)
+                    H = hilbert_class_polynomial(D)
+                    K = GF(p_i)
+                    for j0 in H.roots(ring=K, multiplicities=False):
+                        E = EllipticCurve(K, j=j0)
+                        # `E.twists()` also contains E.
+                        for Et in E.twists():
+                            # `num_checks=1` is sufficient for prime order.
+                            if Et.has_order(N, num_checks=1):
+                                Et.set_order(N, check=False)
```
vincentmacri, line 3183: Do you think it's worth calling `set_order` in `has_order`, at least when you're in the situation where you know `has_order` will return the correct result?

grhkm21:
This PR fixes #38342 as well after reversing the order of searching $D$.

vincentmacri:
> This PR fixes #38342 as well after reversing the order of searching D .

Add "Fixes #38342" to the PR description to make sure the issue gets closed after this is merged.

vincentmacri (changes requested):
I think it looks fine otherwise, but I'll take another look tomorrow or Friday before approving.

grhkm21:
> > This PR fixes #38342 as well after reversing the order of searching D .
> 
> Add "Fixes #38342" to the PR description to make sure the issue gets closed after this is merged.



> > This PR fixes #38342 as well after reversing the order of searching D .
> 
> Add "Fixes #38342" to the PR description to make sure the issue gets closed after this is merged.

@JosePisco

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: Yeah we can do that, I was worried because there's a bug report open right now, but I think it'll be fine...

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,102 @@ def is_ordinary(self, proof=True):
...
+            sage: E.has_order(N^2 + N)
+            True
+
+        AUTHORS:
+
+         - Mariah Lenox (2011-02-16): Initial implementation
+
+         - Gareth Ma (2024-01-21): Fix bug for small curves
+        """
+        # This method does *not* use the cached value of `_order` even when
```
vincentmacri: Why not?

grhkm21: I think that's the point of the method, to check the order "independently" in case the order is wrong (?) I don't know if it makes sense. If you think it makes more sense to use the cached order (when available) I can change it

vincentmacri: My understanding of the point of this method is it's a faster version of `E.order() == N` meant to be used when `E.order()` would take too long to compute. So if we know the order then we should use it. The value of `_order` shouldn't be allowed to be wrong (unless the user has used `proof=False` or something in which case they should know to be aware of wrong answers).

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1436,6 +1436,7 @@ def has_order(self, value, num_checks=8):
             if not (value * G).is_zero():
                 return False
 
+        self._order = value
```
vincentmacri: Just to make sure, there is no path to this line that is impacted by the false positive bug? I.e. if this line is reached, the order is 100% equal to `value`?

grhkm21: No, it's the other way around I think. It's possible for the code to reach this line with a wrong value.

vincentmacri: I think this can be reached in the bugged case. I think you were right before, let's not set it here for now until the issues are sorted out (sorry for changing my mind).

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: If you change it to use `_order` you might need to tweak the test showcasing the bug.

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: > No, it's the other way around I think. It's possible for the code to reach this line with a wrong value.

Let's not set `order` here then, but maybe add a `TODO` comment to do so in the future when/if the bugs are sorted out.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: done

vincentmacri:
You can put the `set_order` calls back in the two generator functions as well (and mention to remove those in the comment in `has_order` about after #38617 is fixed).

JosePisco:
Not a fan at all to address multiple problems (https://github.com/sagemath/sage/issues/38342) in a single PR.
It's starting to contain lots of changes and I think we should resolve the problems atomically.

grhkm21:
I don't see problems if they fit in the PR naturally, as it was required to create the example of failing `has_order`. If you want to, feel free to isolate it into a PR.

vincentmacri:
> Not a fan at all to address multiple problems (#38342) in a single PR. It's starting to contain lots of changes and I think we should resolve the problems atomically.

It's an incredibly minor change and I've already reviewed it. Might not be the best example of best practices but splitting it up at this point seems like it would be a waste of time.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,105 @@ def is_ordinary(self, proof=True):
...
+            True
+
+        AUTHORS:
+
+         - Mariah Lenox (2011-02-16): Initial implementation
+
+         - Gareth Ma (2024-01-21): Fix bug for small curves
+        """
+        # This method does *not* use the cached value of `_order` even when
+        # available.
```
vincentmacri:
```suggestion
```

It does use it now.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,105 @@ def is_ordinary(self, proof=True):
...
+        r"""
+        Return ``True`` if the curve has order ``value``.
+
+        INPUT:
+
+        - ``value`` -- integer in the Hasse-Weil range for this curve
+
+        - ``num_checks``-- integer (default: `8`); the number of times to check
+          whether ``value`` times a random point on this curve equals the
+          identity
```
vincentmacri:
```suggestion
          identity. If the order of the base field is prime, it is sufficient
          to pass in ``num_checks = 1`` because of the Hass bound.
```

vincentmacri: Correct me if I'm wrong about why `num_checks = 1` is sufficient here. Feel free to reword.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1341,6 +1343,105 @@ def is_ordinary(self, proof=True):
...
+        # For really small values, the random tests are too weak to detect wrong
+        # orders So we go with computing directly instead.
+        # In #38341, the bound has been increased to a large value (2^64), but
+        # it should be decreased (to ~100) after bug #38617 is fixed.
+        if q <= 2**64 or hasattr(self, "_order"):
+            return self.order() == value
+
+        # This might be slow
+        # if value.is_prime():
+        #     num_checks = 1
```
vincentmacri, line 1436:
```suggestion
```

vincentmacri (changes requested):

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: Hmm, not quite, it's if the order of the *curve* is prime. I will edit it.

src/sage/schemes/elliptic_curves/ell_finite_field.py
grhkm21: I always keep these comments because they're valuable for contributors later.

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: Sure, the commented out code I asked to remove was checking if `value` is prime.

src/sage/schemes/elliptic_curves/ell_finite_field.py
vincentmacri: I think it's redundant with the added explanation but if you still think it's valuable that's fine.

vincentmacri (approved):
LGTM. I don't have permission to add the positive review label so feel free to do that yourself.

You don't happen to know what the process is to get that permission, do you?

grhkm21:
I think I can get someone to add you to the group. Thanks for the review.

vincentmacri:
> I think I can get someone to add you to the group. Thanks for the review.

That would be great. Thanks for the contribution!

PR #38347, opened 2024-07-10
support passing two base points to .log() for elliptic-curve points

yyyyx4:
The points on an elliptic curve over a finite field form a group of rank up to $2$. In this patch we add support for passing two base points instead of just one to the `.log()` method, which will decompose the given point as a linear combination of the given points. This functionality is already available via the `.abelian_group()` method of the elliptic curve, but the latter is much slower since it relies only on generic-group algorithms and does not exploit the Weil pairing:
```
sage: F = GF((5, 60), 'a')
sage: E = EllipticCurve(F, [1, 1])
sage: A = E.abelian_group()
sage: P, Q = E.gens()[::-1]
sage: T = randrange(P.order()) * P + randrange(Q.order()) * Q
sage: %time A.discrete_log(T, [P,Q])
CPU times: user 47.4 s, sys: 74 ms, total: 47.5 s
Wall time: 47.6 s
(2474, 185989333112663415489036252299763200191)
sage: %time T.log([P, Q])
CPU times: user 1.43 s, sys: 3.34 ms, total: 1.44 s
Wall time: 1.44 s
(2474, 185989333112663415489036252299763200191)
```

### ⌛ Dependencies
- #38350


src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -3982,6 +3998,17 @@ def log(self, base):
...
             400
 
+        ::
+
+            sage: # needs sage.rings.finite_rings
+            sage: F = GF((5, 60), 'a')
+            sage: E = EllipticCurve(F, [1, 1])
+            sage: E.abelian_group().invariants()
+            (4464, 194301464603136995341424045476456938000)
+            sage: P, Q = E.gens()[::-1]
```
GiacomoPope: Is this reversal because of point order? Might be worth making that clear?

```py
P, Q = E.gens()
if P.order() != 4464:
    P, Q = Q, P
```

or something?

GiacomoPope (approved):
This is great! Thank you, means I can stop hand-writing this function all the time myself haha.

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -3997,14 +4024,52 @@ def log(self, base):
...
+                self.set_order(multiple=n, check=False)
+            if not self.order().divides(n):
+                raise ValueError('ECDLog problem has no solution (order does not divide order of base)')
+
+            # find the solution modulo the part where P1,P2 are independent
+            z = P1.weil_pairing(P2, n)
+            o = generic.order_from_multiple(z, n1.gcd(n2), operation='*')
+            x0 = self.weil_pairing(P2, n).log(z, o)
+            y0 = P1.weil_pairing(self, n).log(z, o)
+            T = self - x0*P1 - y0*P2
```
GiacomoPope, line 4096: Is it worth checking if T is zero here and exiting early?

```py
if not T:
    return x0, y0
```
?

yyyyx4:
Thanks! I applied your suggestions, added a new random test, and seem to have fixed the bugs that were caught by the random test.

GiacomoPope:
Cool! Thanks for making the changes.

yyyyx4:
@GiacomoPope I've now simplified this following #38359 and I think it should be ready to be reviewed/merged.

src/sage/schemes/elliptic_curves/ell_point.py
```diff
@@ -3950,20 +3955,31 @@ def log(self, base):
         For anomalous curves with `\#E = p`, the
         :meth:`padic_elliptic_logarithm` function is called.
 
+        For two-dimensional logarithms, we first compute the Weil pairing of
+        `Q` with `P_1` and `P_2` and their logarithms to base the pairing of
+        `P_1` and `P_2`; this allows to read off `x` and `y` modulo the part
+        of the order where `P_1` and `P_2`. Modulo the rest of the order the
```
GiacomoPope: not sure i understand this comment --

> modulo the part of the order where `P_1` and `P_2`

GiacomoPope: should this be "modulo the part of the order where P_1 and P_2 are independent"?

GiacomoPope (approved):
comment about comment, but this looks great.

src/sage/schemes/elliptic_curves/ell_point.py
yyyyx4: Indeed, that was a typo, thanks. Fixed.

yyyyx4:
The CI seems to be morally green.

PR #38356, opened 2024-07-13
some tweaks in singular pexpect interface

fchapoton:
Mainly trying to simplify the conversion of Singular polynomials to sage in the pexpect interface, using appropriate methods in singular.

I have not checked if speed is enhanced or deteriorated.

Also adding a conversion for Singular's "vector" type with a doctest.

### :memo: Checklist


fchapoton:
This turns out to be more subtle than expected, with several unexpected triggered problems. Some still to be handled.

fchapoton:
The failing doctests pass when run alone in the console. How to debug that ?

src/sage/interfaces/singular.py
```diff
@@ -1466,10 +1466,34 @@ def __len__(self):
...
+
+            sage: A = singular.matrix(2, 4)
             sage: len(A)
             4
+            sage: A = singular.matrix(2, 2)
+            sage: len(A)
+            2
         """
+        if self.type() == 'matrix':
+            return int(self.ncols())
```
user202729, line 1488: This is technically a breaking change, but it seems to make sense since it's consistent with `list`. Do leave an explanation though. (on the other hand there's the question why not the other option i.e. make `list` return the flattened list)

src/sage/interfaces/singular.py
```diff
@@ -2042,7 +2050,7 @@ def _sage_(self, R=None):
             return self.sage_matrix(R, sparse=True)
         elif typ == 'matrix':
             return self.sage_matrix(R, sparse=False)
-        elif typ == 'list':
+        elif typ == 'list' or typ == 'vector':
```
user202729: Looks like singular consider vectors as vectors over an implicit infinite (unbounded) dimensional module, which I don't think sage has an analogue for. `vector` is close but it doesn't automatically extend.

```
> print(gen(3));
[0,0,1]
> [0, 1];
gen(2)
```

converting to `list` is somewhat reasonable, but then you break round-trip. But then round-trip is already broken with `singular(singular.list([x, x]).sage())` return a vector instead of a list, so maybe it's fine.

src/sage/interfaces/singular.py
```diff
@@ -1797,111 +1820,87 @@ def sage_poly(self, R=None, kcache=None):
...
-        # as we know what to expect.
-
-        is_short = self.parent().eval('short')
+        P = self.parent()
+        variable_prod = "*".join(R.variable_names())
+
+        # The output of self.coef is a Singular matrix. When iterating over
+        # this matrix, each term is a pair (monomial, coefficient)
+
+        # warning: self.coef has a bug for the zero polynomial !
```
user202729, line 1837: what's the bug? I get `1,0` which looks sane enough?

PR #38359, opened 2024-07-13
Homogenise `.log()` api across implementations of finite field elements

### LINKED ISSUE #38350: inconsistent API for `.log()` method in finite fields
After #37329, the `.log()` method takes an optional `order=` argument for *some* implementations of finite-field elements, and throws exceptions for some others when a third argument (in addition to `self` and the base element) is passed.

This inconsistency is annoying: All implementations of this method should accept the `order=` argument and either do something useful with it or discard it silently but gracefully.

AZ-0:
Fixes #38350 by adding `order=` and `check=` arguments to the `.log()` method of finite field elements, under all implementations.
- `element_pari_ffelt.pyx`: done by @yyyyx4 in #37329.
- `element_ntl_gf2e.pyx`: if provided, does not compute `base_order`.
- `element_givaro.pyx`: if provided, passes `order` to the underlying `discrete_log` call.
- `integer_mod.pyx`: the argument is discarded (unless `check=True`, in which case the order is checked).

### :memo: Checklist

### DIFF
--- a/src/sage/rings/finite_rings/element_givaro.pyx
+++ b/src/sage/rings/finite_rings/element_givaro.pyx
@@ -1411,10 +1411,17 @@ cdef class FiniteField_givaroElement(FinitePolyExtElement):
         """
         return Integer(self._cache.log_to_int(self.element))
 
-    def log(FiniteField_givaroElement self, base):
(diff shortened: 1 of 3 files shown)


yyyyx4 (approved):
Thanks a lot for taking care of this!

PR #38361, opened 2024-07-14
EllipticCurve: Raise error on unexpected keyword argument

user202729:
As described in the title. Currently the constructor just silently ignores the arguments it doesn't recognize, which can be confusing if the user pass in `ring=R` and see it has no effect.

### :memo: Checklist

- [ ] I have linked a relevant issue or discussion. (not aware of one)

### :hourglass: Dependencies

None.


src/sage/schemes/elliptic_curves/constructor.py
```diff
@@ -472,13 +472,25 @@ def create_object(self, version, key, **kwds):
...
         """
         R, x = key
 
         if R is QQ:
             from .ell_rational_field import EllipticCurve_rational_field
             return EllipticCurve_rational_field(x, **kwds)
-        elif isinstance(R, NumberField):
+        elif kwds:
+            raise TypeError(f"Unexpected keyword arguments: {kwds}")
+
```
kwankyu, line 511: Error message starts in small letters, as in
```sage
ImportError: cannot import name SteenrodAlgebra
```

src/sage/schemes/elliptic_curves/constructor.py
```diff
@@ -472,13 +472,25 @@ def create_object(self, version, key, **kwds):
...
+        EXAMPLES::
+            sage: E = EllipticCurve.create_object(0, (QQ, (1, 2, 0, 1, 2)), rank=2)
+            sage: type(E)
+            Elliptic Curve defined by y^2 + x*y = x^3 + 2*x^2 + x + 2 over Rational Field
+            sage: EllipticCurve.create_object(0, (GF(3), (1, 2, 0, 1, 2)), rank=2)
+            Traceback (most recent call last):
+            ...
+            TypeError: Unexpected keyword arguments: {'rank': 2}
         """
```
kwankyu, line 503: Add a blank line after `EXAMPLES::`.

src/sage/schemes/elliptic_curves/ell_rational_field.py
```diff
@@ -187,21 +194,23 @@ def __init__(self, ainvs, **kwds):
...
-            self._set_rank(kwds['rank'])
+            self._set_rank(kwds.pop('rank'))
         if 'regulator' in kwds:
-            self.__regulator = (kwds['regulator'], True)
+            self.__regulator = (kwds.pop('regulator'), True)
         if 'torsion_order' in kwds:
-            self._set_torsion_order(kwds['torsion_order'])
+            self._set_torsion_order(kwds.pop('torsion_order'))
+        if kwds:
+            raise TypeError(f"Unexpected keyword arguments: {kwds}")
```
kwankyu, line 208: Here too.

src/sage/schemes/elliptic_curves/constructor.py
```diff
@@ -472,13 +472,25 @@ def create_object(self, version, key, **kwds):
             Keyword arguments are currently only passed to the
             constructor for elliptic curves over `\QQ`; elliptic
             curves over other fields do not support them.
+
+        EXAMPLES::
+            sage: E = EllipticCurve.create_object(0, (QQ, (1, 2, 0, 1, 2)), rank=2)
+            sage: type(E)
+            Elliptic Curve defined by y^2 + x*y = x^3 + 2*x^2 + x + 2 over Rational Field
+            sage: EllipticCurve.create_object(0, (GF(3), (1, 2, 0, 1, 2)), rank=2)
```
kwankyu: Why repeat the same test `type(E)` as in line 467? This test would fail by the way...

The new test may better be in `TESTS::`.

user202729:
Thanks for the review, I will fix these soon. (By the way, can I get permission to run tests on GitHub Actions CI?)

src/sage/schemes/elliptic_curves/ell_rational_field.py
user202729: Alright, but actually there's quite a few messages that doesn't follow that convention.

```bash
[schemes]$ grep -r "Error.*'[A-Z]"
affine/affine_homset.py:            raise TypeError('F must be a numerical field')
affine/affine_space.py:            raise ValueError('Must have the same base ring')
elliptic_curves/ell_curve_isogeny.py:        raise ValueError('E1 must be a short Weierstrass curve')
elliptic_curves/ell_curve_isogeny.py:        raise ValueError('E2 must be a short Weierstrass curve')
elliptic_curves/ell_field.py:            raise ValueError('Hasse invariant only defined in positive characteristic')
elliptic_curves/ell_generic.py:            raise ValueError('Frobenius isogeny only exists in positive characteristic')
elliptic_curves/ell_number_field.py:            raise ValueError('There is insufficient data to determine the rank - 2-descent gave lower bound %s and upper bound %s' % (lower, upper))
elliptic_curves/ell_point.py:                raise ValueError('Value %s illegal for point order' % value)
elliptic_curves/ell_point.py:                    raise ValueError('Value %s illegal: outside max Hasse bound' % value)
elliptic_curves/ell_point.py:                raise ValueError('Value %s illegal: %s * %s is not the identity' % (value, value, self))
elliptic_curves/ell_point.py:            raise ValueError('This point %s is not of order n=%s' % (P, n))
elliptic_curves/ell_point.py:            raise ValueError('This point %s is not in Ker(pi - 1)' % P)elliptic_curves/ell_point.py:            raise ValueError('Point %s not in Ker(pi - q)' % Q)
elliptic_curves/ell_point.py:            raise ValueError('ECDLog problem has no solution (order does not divide order of base)')
```

src/sage/schemes/elliptic_curves/constructor.py
user202729: Indeed, probably a copy paste error somewhere.

The example is meant to address the note, but yes I suppose it doesn't need to be user-facing. (Actually thinking about it, the note is probably unnecessary now that the user will be informed of the error when appropriate anyway.)

src/sage/schemes/elliptic_curves/ell_rational_field.py
kwankyu: Yes.

src/sage/schemes/elliptic_curves/constructor.py
kwankyu: I now saw the note. From my point of view, by the note, the error message for unexpected keyword arguments is not really necessary. But the note is not so visible to users. Hence I think the error message helps the users.

user202729:
The data are invalid for elliptic curves not over ℚ anyway, so I check at construction time and delete them. The `names` looks legal enough.

```
sage: EllipticCurve("15a").rank()
0
sage: K.<a> = NumberField(x^2 + 7)
sage: EllipticCurve("15a").change_ring(K).rank()
1
sage: EllipticCurve(GF(7), "15a")
Elliptic Curve defined by y^2 + x*y + y = x^3 + x^2 + 4*x + 4 over Finite Field of size 7
```

kwankyu:
@yyyyx4

yyyyx4 (approved):
Looks good to me, thanks!

vbraun:
Followup at https://github.com/sagemath/sage/pull/38917

PR #36942, opened 2023-12-21
Fix coercion problem with Algebraic Real Field

RuchitJagodara:
- Simplify the expression before converting it to algebraic number
- Added a doctest

This patch fixes #12745. With this commit, the coerce
issue of Algebraic Real Field is fixed. Now it is giving correct answer for below cases

```python
sage: x = exp(2*I*pi/7) + exp(-2*I*pi/7)
sage: QQbar(x) in AA
True
sage: AA(x)
1.246979603717467?
```

### :memo: Checklist

- [x] The title is concise, informative, and self-explanatory.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.

### :hourglass: Dependencies

None

Issue #12745, opened 2012-03-25
Coercion problem with Algebraic Real Field


11d1fc49-71a1-44e1-869f-76be013245a0:
<div id="comment:0"></div>

The following algebraic number is clearly real, but the AA class seems to have trouble digesting it:

```
----------------------------------------------------------------------
| Sage Version 5.0.beta10, Release Date: 2012-03-23                  |
| Type notebook() for the GUI, and license() for information.        |
----------------------------------------------------------------------
**********************************************************************
*                                                                    *
* Warning: this is a prerelease version, and it may be unstable.     *
*                                                                    *
**********************************************************************
sage: x = exp(2*I*pi/7) + exp(-2*I*pi/7)
sage: QQbar(x) in AA
True
sage: AA(x)
ERROR: An unexpected error occurred while tokenizing input
The following traceback may be corrupted or invalid
The error message is: ('EOF in multi-line statement', (1327, 0))

ERROR: An unexpected error occurred while tokenizing input
The following traceback may be corrupted or invalid
The error message is: ('EOF in multi-line statement', (4541, 0))

ERROR: An unexpected error occurred while tokenizing input
The following traceback may be corrupted or invalid
The error message is: ('EOF in multi-line statement', (1327, 0))

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)

/storage/masiao/sage-5.0.beta10/devel/sage-main/<ipython console> in <module>()

/storage/masiao/sage-5.0.beta10/local/lib/python2.7/site-packages/sage/structure/parent.so in sage.structure.parent.Parent.__call__ (sage/structure/parent.c:7940)()

/storage/masiao/sage-5.0.beta10/local/lib/python2.7/site-packages/sage/structure/coerce_maps.so in sage.structure.coerce_maps.DefaultConvertMap_unique._call_ (sage/structure/coerce_maps.c:3344)()

/storage/masiao/sage-5.0.beta10/local/lib/python2.7/site-packages/sage/structure/coerce_maps.so in sage.structure.coerce_maps.DefaultConvertMap_unique._call_ (sage/structure/coerce_maps.c:3247)()

/storage/masiao/sage-5.0.beta10/local/lib/python2.7/site-packages/sage/rings/qqbar.pyc in _element_constructor_(self, x)
    689                 raise ValueError("Cannot coerce algebraic number with non-zero imaginary part to algebraic real")
    690         elif hasattr(x, '_algebraic_'):
[...]

PR #38362, opened 2024-07-14
Give deprecation warning when raising negative AlgebraicReal to fractional odd power

### LINKED ISSUE #36735: number_field_elements_from_algebraics incorrectly simplify root of unity?
### Steps To Reproduce

Run
```
L, [ww], phi = number_field_elements_from_algebraics([(-1)^(2/3)])
assert ww != 1
```

### Expected Behavior

That ww become a primitive 3rd root of unity.

### Actual Behavior

ww = 1.

### Additional Information

_No response_

### Environment

```markdown
- **OS**: Linux
- **Sage Version**: SageMath version 10.1, Release Date: 2023-08-20
```

### Checklist

user202729:
Previously, the behavior of `AA` is inconsistent with most other cases:

```sage
sage: AA(-8)^(1/3)
-2
sage: QQbar(-8)^(1/3)
1.000000000000000? + 1.732050807568878?*I
```

This will give a deprecation warning when the user attempt to rely on that behavior.

------

### Previous description

Intended to fix​ https://github.com/sagemath/sage/issues/36735 , but now it doesn't.

This fix tries to be more conservative in how to converts symbolic expression of the form `a^b`: it still first convert `a` to `self.field` then raise the result to power of `b`, but it checks if `a<0` and `b` is a fraction then raise an error (I think the previously computed result in this case is never expected).

It does not fix​ https://github.com/sagemath/sage/issues/12745 though (an alternative, as mentioned in https://github.com/sagemath/sage/pull/36942 , is to convert to QQbar first then convert it to AA, then this code wouldn't be necessary. What do you think of that alternative approach?)

### :memo: Checklist

### DIFF
--- a/src/sage/rings/qqbar.py
+++ b/src/sage/rings/qqbar.py
@@ -118,6 +118,9 @@ We can convert from symbolic expressions::
     sage: QQbar((-8)^(1/3))
     1.000000000000000? + 1.732050807568878?*I
     sage: AA((-8)^(1/3))
+    doctest:warning...
+    DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
+    See https://github.com/sagemath/sage/issues/38362 for details.
     -2
     sage: QQbar((-4)^(1/4))
     1 + 1*I
@@ -134,10 +137,13 @@ symbolic expressions are algebraic numbers::
     sage: QQbar(sqrt(2) + QQbar(sqrt(3)))                                               # needs sage.symbolic
     3.146264369941973?
 
-Note the different behavior in taking roots: for ``AA`` we prefer real
-roots if they exist, but for ``QQbar`` we take the principal root::
+Currently for ``AA`` we prefer real roots if they exist, but for ``QQbar`` we take the principal root.
+However, this behavior will change in the future::
 
     sage: AA(-1)^(1/3)
+    doctest:warning...
+    DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
+    See https://github.com/sagemath/sage/issues/38362 for details.
     -1
     sage: QQbar(-1)^(1/3)
     0.500000000000000? + 0.866025403784439?*I
@@ -4360,10 +4366,13 @@ class AlgebraicNumber_base(sage.structure.element.FieldElement):
             sage: AA(-2).nth_root(5, all=True) == QQbar(-2).nth_root(5, all=True)   # long time
             True
         """
+        n = ZZ(n)
         if not all:
-            return self ** ~ZZ(n)
+            if self.parent() is AA and self.sign() < 0 and n % 2 != 0 and abs(n) > 1:
+                return -((-self) ** ~n)
+            return self ** ~n
         else:
-            root = QQbar(self) ** ~ZZ(n)
+            root = QQbar(self) ** ~n
             zlist = [root]
             zeta = QQbar.zeta(n)
             for k in range(1, n):
@@ -6373,6 +6382,9 @@ class AlgebraicNumberPowQQAction(Action):
     TESTS::
 
         sage: AA(-8)^(1/3)
+        doctest:warning...
+        DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
+        See https://github.com/sagemath/sage/issues/38362 for details.
         -2
         sage: AA(-8)^(2/3)
         4
@@ -6398,6 +6410,9 @@ class AlgebraicNumberPowQQAction(Action):
             sage: act = AlgebraicNumberPowQQAction(QQ, AA); act
             Right Rational Powering by Rational Field on Algebraic Real Field
             sage: act(AA(-2), 1/3)
+            doctest:warning...
+            DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
+            See https://github.com/sagemath/sage/issues/38362 for details.
             -1.259921049894873?
 
         ::
@@ -6429,8 +6444,13 @@ class AlgebraicNumberPowQQAction(Action):
 
         # Parent of the result
         S = self.codomain()
-        if S is AA and d % 2 == 0 and x.sign() < 0:
-            S = QQbar
+        if S is AA and x.sign() < 0:
+            if d % 2 == 0:
+                S = QQbar
+            elif d != 1:
+                from sage.misc.superseded import deprecation
+                deprecation(38362, "Taking the root of an algebraic real number "
+                            "will yield the principal root in the future.")
 
         # First, check for exact roots.
         if isinstance(x._descr, ANRational):
--- a/src/sage/symbolic/expression_conversion_algebraic.py
+++ b/src/sage/symbolic/expression_conversion_algebraic.py
@@ -26,6 +26,7 @@ from sage.functions.all import exp
 from sage.symbolic.expression_conversions import Converter
 from sage.symbolic.operators import add_vararg, mul_vararg
 from sage.symbolic.ring import SR
+from sage.rings.qqbar import AlgebraicRealField
 
 
 #############
@@ -73,6 +74,15 @@ class AlgebraicConverter(Converter):
             sage: a = AlgebraicConverter(QQbar)
             sage: a.arithmetic(f, f.operator())
             1.414213562373095?
+
+        Note that converting an expression where an odd root is taken will take the real root
+        if the target is ``AA``, but this behavior will change in the future::
+
+            sage: AA((-1)^(2/3))
+            doctest:warning...
+            DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
+            See https://github.com/sagemath/sage/issues/38362 for details.
+            1
 
         TESTS::
 
--- a/src/sage/features/sagemath.py
+++ b/src/sage/features/sagemath.py
@@ -879,8 +879,8 @@ class sage__rings__number_field(JoinFeature):
     should be marked likewise::
 
         sage: # needs sage.rings.number_field
-        sage: AA(-1)^(1/3)
-        -1
+        sage: AA(2)^(1/2)
+        1.414213562373095?
         sage: QQbar(-1)^(1/3)
         0.500000000000000? + 0.866025403784439?*I
 


kwankyu:
Would you consider #38564?

user202729:
> Would you consider #38564?

@kwankyu I'm happy with that option also, but some other people might have trouble with previous code give different result silently.

 https://github.com/sagemath/sage/pull/36942#issuecomment-1870480695

@mezzarobba @nbruin  ?

mezzarobba:
> I'm happy with that option also, but some other people might have trouble with previous code give different result silently.
> 
> [#36942 (comment)](https://github.com/sagemath/sage/pull/36942#issuecomment-1870480695)

I believe we should change the semantics of fractional powers in `AA` (and everywhere else!) to always use complex principal roots, and use a different function name (#12074) for real nth roots. IMO the current behavior is a bug (even though it is clearly intentional and documented), so I personally would be ok with just fixing it. I would ask on sage-devel before doing it, and draw attention to the change in the release notes. The alternative, I suppose, is to first give a introduce an alternative real rational power function and give a deprecation warning in all cases where the result would differ for a while. I'm not sure which option is the least inconvenient for everyone involved...

kwankyu:
This PR and my PR #38564 both propose changes in the semantics of AA(symbolic expression). We cannot view the current behavior as a bug. That is just a different interpretation of fractional powers of negative reals, which number theorists find inconvenient

Hence I think we should follow the deprecation rules. For a year, we advertise the upcoming semantics change. Then after a year, we switch to the semantics we want.

To implement that, I suggest to use this PR for the deprecation purpose, and after the deprecation period use my PR to change the semantics. What do you think?

kwankyu:
By the way, I think one-year deprecation period is too long, in general. A half year would be enough.

mezzarobba:
On Tuesday, 27 August 2024 05:12:49 CEST Kwankyu Lee wrote:
> We cannot view the current behavior as a bug.

Not in isolation, but I'd argue that the current semantics is wrong 
because it breaks the assumptions of the coercion system, and, more 
generally, the overall consistency of the Sage library (one expects 
methods with the same name applied to mathematically equivalent objects 
to behave in a consistent way).

(I'm not trying to argue for bypassing the deprecation rules here, just 
trying to explain why I believe it would be acceptable.)

> To implement that, I suggest to use this PR for the deprecation
> purpose, and after the deprecation period use my PR to change the
> semantics. What do you think?

Sounds good to me.

-- 
Marc

kwankyu:
> ... but I'd argue that the current semantics is wrong because it breaks the assumptions of the coercion system, and, more generally, the overall consistency of the Sage library (one expects methods with the same name applied to mathematically equivalent objects to behave in a consistent way). 

I agree with that, why we are working here.

kwankyu:
Hmm. It seems that the cause of the "bug" in #36735 is elsewhere... See this:

```sage
sage: RR(-1)^(1/3)
0.500000000000000 + 0.866025403784439*I
sage: CC(-1)^(1/3)
0.500000000000000 + 0.866025403784439*I
sage: QQbar(-1)^(1/3)
0.500000000000000? + 0.866025403784439?*I
sage: AA(-1)^(1/3)
-1
```

kwankyu:
@videlec Any other idea on these inconsistencies?

kwankyu:
In accordance with the opinion in https://groups.google.com/g/sage-devel/c/s81ieq2vpVo, I also fixed the first inconsistency in https://github.com/sagemath/sage/pull/38564.

@user202729 Are you willing to repurpose this PR for appropriate deprecation warnings?

user202729:
Alright, I try to change the code accordingly.

If you also want `AA(-1)^(1/3)` to raise the warning however, we need more changes. One is here…

```python
        if isinstance(x._descr, ANRational):
            rt = rational_exact_root(abs(x._descr._value), d)
            if rt is not None:
                if x._descr._value < 0:
                    if S is AA:
# TODO add the warning here
                        return AlgebraicReal(ANRational((-rt)**n))
```

I haven't found the other.

nbruin:
> Surely the element should belong to its base ring?

No, it does not:
```
sage: R.<x>=QQ[]
sage: x.base_ring()
Rational Field
```
It looks like it's a shortcut for `.parent().base_ring()`. In a way, Qbar is a quadratic extension of AA and AA is apparently considered as a primitive construct, since it's not obviously a finite degree extension of something conveniently represented:
```
sage: QQbar.base_ring()
Algebraic Real Field
sage: AA.base_ring()
Algebraic Real Field
```
I'm not sure I agree (I think of QQbar as not particularly related to any complex embedding), but given that the implementation really does realize QQbar as a subfield of CC I guess it makes sense ...

kwankyu:
> Alright, I try to change the code accordingly.

Thanks.

```sage
sage: AA((-1)^(1/3))
/Users/kwankyu/GitHub/sage-dev/src/sage/symbolic/expression_conversions.py:212: DeprecationWarning: Conversion of negative number to fractional power to AA is deprecated
See https://github.com/sagemath/sage/issues/36735 for details.
  return self.arithmetic(ex, operator)
-1
```

Your deprecation warnings should point to this PR #38362.

> 
> If you also want `AA(-1)^(1/3)` to raise the warning however, ...

Yes. That should also be done in this PR:

```diff
--- a/src/sage/rings/qqbar.py
+++ b/src/sage/rings/qqbar.py
@@ -6441,6 +6441,10 @@ class AlgebraicNumberPowQQAction(Action):
             if rt is not None:
                 if x._descr._value < 0:
                     if S is AA:
+                        if d % 2 != 0 and d != 1 and x.sign() < 0:
+                            from sage.misc.superseded import deprecation
+                            deprecation(38362, "Raising negative algebraic real number "
+                                        "will give primitive root in future.")
                         return AlgebraicReal(ANRational((-rt)**n))
                     else:
                         z = QQbar.zeta(2 * d)._pow_int(n)
```

kwankyu:
Please rebase to the develop branch.

src/sage/symbolic/expression_conversion_algebraic.py
```diff
@@ -97,6 +98,28 @@ def arithmetic(self, ex, operator):
...
+            sage: AA((-sqrt(27)+1)^5)
+            -1300.937032397459?
+            sage: AA((-sqrt(27)+1)^(5/2))
+            Traceback (most recent call last):
+            ...
+            ValueError: Cannot coerce algebraic number with nonzero imaginary part to algebraic real
+
+        Another test (need to be splitted from the above for the warning to appear)::
+
+            sage: AA((-sqrt(27)+1)^(5/3))
```
kwankyu: Instead of "Test that deprecation warnings are issued correctly", how about explaining what is deprecated? As this is for users, perhaps we may place these examples in "EXAMPLES" block.

kwankyu:
Please update the PR title and the description, as the deprecation warnings will point to this PR. To avoid confusion, you may keep the original description below the new one.

user202729:
Actually there's one more
```sage
sage: ZZ(-1)^(1/3)
(-1)^(1/3)
sage: ZZ(-1).nth_root(3)
-1
```
Should we also give warning?

src/sage/rings/qqbar.py
```diff
@@ -118,6 +118,9 @@
     sage: QQbar((-8)^(1/3))
     1.000000000000000? + 1.732050807568878?*I
     sage: AA((-8)^(1/3))
+    doctest:warning...
+    DeprecationWarning: Raising negative algebraic real number will give principal root in future
+    See https://github.com/sagemath/sage/issues/38362 for details.
     -2
```
kwankyu, line 124: You omitted a period at the end here.

But I suggested to you a broken sentence: "Raising negative algebraic real number will give principal root in future." Sorry!

A good sentence is "Taking the root of an algebraic real number will yield the principal root in the future."

src/sage/symbolic/expression_conversion_algebraic.py
```diff
@@ -97,6 +107,20 @@ def arithmetic(self, ex, operator):
             sage: L = QuadraticField(3, embedding=-AA(3).sqrt())
             sage: bool(L.gen() == -sqrt(3))
             True
+
+        More test for deprecation warning::
+
```
kwankyu: "More tests"?

kwankyu:
> Actually there's one more
> 
> ```python
> sage: ZZ(-1)^(1/3)
> (-1)^(1/3)
> sage: ZZ(-1).nth_root(3)
> -1
> ```
> 
> Should we also give warning?

I don't think so. The doc of `nth_root` clearly says that the operation is within reals. `nth_root` for elements in `ZZ` and `QQ` seems to be treated separately in sage, similarly like
```sage
sage: ZZ(-1)^(1/3)
(-1)^(1/3)
sage: QQ(-1)^(1/3)
(-1)^(1/3)
```
Anyway we are not "fix"-ing this.

kwankyu:
> Any idea why does the `AA((-sqrt(27)+1)^(5/3))` test fail? In my local testing it does give the warning.

Because a deprecation warning is issued only once in the same sage session.

kwankyu:
> Any idea why does the `AA((-sqrt(27)+1)^(5/3))` test fail? In my local testing it does give the warning.

It seems that deprecation warnings are skipped in "TESTS" block...

kwankyu:
> > Any idea why does the `AA((-sqrt(27)+1)^(5/3))` test fail? In my local testing it does give the warning.
> 
> It seems that deprecation warnings are skipped in "TESTS" block...

@mkoeppe ?

nbruin:
> Actually there's one more
> 
> ```python
> sage: ZZ(-1)^(1/3)
> (-1)^(1/3)
> sage: ZZ(-1).nth_root(3)
> -1
> ```
> 
> Should we also give warning?
> 
> (though there's a difference, `nth_root` for `ZZ` and `QQ` currently give error if the result is not in the original field.)
> 
> Because of how the current code is implemented, `AA(-1).nth_root(3)` will give a warning (and have its behavior change once we change the behavior of `**`) anyway. What's the desired behavior here?
> 
> Meanwhile:
> 
> ```
> sage: AA(-1).nth_root(2)
> I
> ```
> 
> unlike what happens for `ZZ(-1).nth_root(2)` (gives error).
> 
> Any idea why does the `AA((-sqrt(27)+1)^(5/3))` test fail? In my local testing it does give the warning.

It looks like you haven't formatted the expected output exactly as given in the example for `deprecation`:
https://github.com/sagemath/sage/blob/e042294b5ed8e1cb564965079e20e4270ccde340/src/sage/misc/superseded.py#L91
Perhaps you get better results if you follow that formatting exactly?

kwankyu:
Strange thing is happening here. First remove a typo: a period at the end
```diff
--- a/src/sage/symbolic/expression_conversion_algebraic.py
+++ b/src/sage/symbolic/expression_conversion_algebraic.py
@@ -112,7 +112,7 @@ class AlgebraicConverter(Converter):
 
             sage: AA((-sqrt(27)+1)^(5/3))
             doctest:warning...
-            DeprecationWarning: Raising negative algebraic real number will give principal root in future
+            DeprecationWarning: Raising negative algebraic real number will give principal root in future.
             See https://github.com/sagemath/sage/issues/38362 for details.
             -10.916550432858478?
             sage: AA((-sqrt(27)+1)^5)
```

Then as we are observing,
```
$ sage -t src/sage/symbolic/expression_conversion_algebraic.py
Running doctests with ID 2024-08-28-17-21-12-30034980.
Running with SAGE_LOCAL='/Users/kwankyu/GitHub/sage-dev/local' and SAGE_VENV='/Users/kwankyu/GitHub/sage-dev/local/var/lib/sage/venv-python3.9'
Using --optional=homebrew,pip,sage,sage_spkg
Features to be detected: 4ti2,SAGE_SRC,benzene,bliss,buckygen,conway_polynomials,coxeter3,csdp,cvxopt,cvxopt,database_cremona_ellcurve,database_cremona_mini_ellcurve,database_cubic_hecke,database_ellcurves,database_graphs,database_jones_numfield,database_knotinfo,dvipng,ecm,fpylll,fricas,gap_package_atlasrep,gap_package_design,gap_package_grape,gap_package_guava,gap_package_hap,gap_package_polycyclic,gap_package_qpa,gap_package_quagroup,gfan,glucose,graphviz,imagemagick,ipython,jmol,jupymake,kenzo,kissat,latte_int,lrcalc_python,lrslib,mathics,matroid_database,mcqd,meataxe,mpmath,msolve,nauty,networkx,numpy,palp,pandoc,pdf2svg,pdftocairo,pexpect,phitigra,pillow,plantri,polytopes_db,polytopes_db_4d,pplpy,primecountpy,ptyprocess,pycosat,pycryptosat,pynormaliz,pyparsing,python_igraph,requests,rpy2,rubiks,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.homfly,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.sat,sage.schemes,sage.symbolic,sage_numerical_backends_coin,sagemath_doc_html,scipy,singular,sirocco,sphinx,symengine_py,sympy,tdlib,threejs,topcom
Doctesting 1 file.
sage -t --warn-long 48.5 --random-seed=150437855707950653970655890706877376821 src/sage/symbolic/expression_conversion_algebraic.py
File "src/sage/symbolic/expression_conversion_algebraic.py", line 81, in sage.symbolic.expression_conversion_algebraic.AlgebraicConverter.arithmetic
Failed example:
    AA((-1)^(2/3))
Expected:
    doctest:warning...
    DeprecationWarning: Raising negative algebraic real number will give principal root in future
    See https://github.com/sagemath/sage/issues/38362 for details.
    1
Got:
    doctest:warning
      File "<doctest sage.symbolic.expression_conversion_algebraic.AlgebraicConverter.arithmetic[4]>", line 1, in <module>
        AA((-Integer(1))**(Integer(2)/Integer(3)))
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/rings/qqbar.py", line 1168, in _element_constructor_
        return x._algebraic_(AA)
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/symbolic/expression_conversion_algebraic.py", line 330, in algebraic
        return AlgebraicConverter(field)(ex)
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/symbolic/expression_conversions.py", line 212, in __call__
        return self.arithmetic(ex, operator)
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/symbolic/expression_conversion_algebraic.py", line 134, in arithmetic
        return self.field(base**expt)
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/rings/qqbar.py", line 6443, in _act_
        deprecation(38362, "Raising negative algebraic real number "
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/misc/superseded.py", line 99, in deprecation
        warning(issue_number, message, DeprecationWarning, stacklevel)
      File "/Users/kwankyu/GitHub/sage-dev/src/sage/misc/superseded.py", line 180, in warning
        warn(message, warning_class, stacklevel)
      File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/warnings.py", line 109, in _showwarnmsg
        sw(msg.message, msg.category, msg.filename, msg.lineno,
    :
    DeprecationWarning: Raising negative algebraic real number will give principal root in future.
    See https://github.com/sagemath/sage/issues/38362 for details.
    1
File "src/sage/symbolic/expression_conversion_algebraic.py", line 113, in sage.symbolic.expression_conversion_algebraic.AlgebraicConverter.arithmetic
Failed example:
    AA((-sqrt(27)+1)^(5/3))
Expected:
    doctest:warning...
    DeprecationWarning: Raising negative algebraic real number will give principal root in future.
    See https://github.com/sagemath/sage/issues/38362 for details.
    -10.916550432858478?
Got:
    -10.916550432858478?
1 item had failures:
   2 of  19 in sage.symbolic.expression_conversion_algebraic.AlgebraicConverter.arithmetic
    [61 tests, 2 failures, 0.46 s]
sage -t --warn-long 48.5 --random-seed=150437855707950653970655890706877376821 src/sage/symbolic/expression_conversion_algebraic.py  # 2 doctests failed
Total time for all tests: 0.5 seconds
    cpu time: 0.5 seconds
    cumulative wall time: 0.5 seconds
Features detected for doctesting: 
```

Then remove the first example issuing the deprecation warning,
```diff
--- a/src/sage/symbolic/expression_conversion_algebraic.py
+++ b/src/sage/symbolic/expression_conversion_algebraic.py
@@ -75,15 +75,6 @@ class AlgebraicConverter(Converter):
             sage: a.arithmetic(f, f.operator())
             1.414213562373095?
 
-        Note that converting an expression where an odd root is taken will take the real root
-        if the target is ``AA``, but this behavior will change in the future::
-
-            sage: AA((-1)^(2/3))
-            doctest:warning...
-            DeprecationWarning: Raising negative algebraic real number will give principal root in future
-            See https://github.com/sagemath/sage/issues/38362 for details.
-            1
-
         TESTS::
 ```

Then voila!
```
$ sage -t src/sage/symbolic/expression_conversion_algebraic.py
too many failed tests, not using stored timings
Running doctests with ID 2024-08-28-17-23-34-76f52432.
Running with SAGE_LOCAL='/Users/kwankyu/GitHub/sage-dev/local' and SAGE_VENV='/Users/kwankyu/GitHub/sage-dev/local/var/lib/sage/venv-python3.9'
Using --optional=homebrew,pip,sage,sage_spkg
Features to be detected: 4ti2,SAGE_SRC,benzene,bliss,buckygen,conway_polynomials,coxeter3,csdp,cvxopt,cvxopt,database_cremona_ellcurve,database_cremona_mini_ellcurve,database_cubic_hecke,database_ellcurves,database_graphs,database_jones_numfield,database_knotinfo,dvipng,ecm,fpylll,fricas,gap_package_atlasrep,gap_package_design,gap_package_grape,gap_package_guava,gap_package_hap,gap_package_polycyclic,gap_package_qpa,gap_package_quagroup,gfan,glucose,graphviz,imagemagick,ipython,jmol,jupymake,kenzo,kissat,latte_int,lrcalc_python,lrslib,mathics,matroid_database,mcqd,meataxe,mpmath,msolve,nauty,networkx,numpy,palp,pandoc,pdf2svg,pdftocairo,pexpect,phitigra,pillow,plantri,polytopes_db,polytopes_db_4d,pplpy,primecountpy,ptyprocess,pycosat,pycryptosat,pynormaliz,pyparsing,python_igraph,requests,rpy2,rubiks,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.homfly,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.sat,sage.schemes,sage.symbolic,sage_numerical_backends_coin,sagemath_doc_html,scipy,singular,sirocco,sphinx,symengine_py,sympy,tdlib,threejs,topcom
Doctesting 1 file.
sage -t --random-seed=305929579149378824671940413996301385865 src/sage/symbolic/expression_conversion_algebraic.py
    [60 tests, 0.48 s]
All tests passed!
Total time for all tests: 0.5 seconds
    cpu time: 0.5 seconds
    cumulative wall time: 0.5 seconds
Features detected for doctesting: 
```

That is, the second example (`AA((-sqrt(27)+1)^(5/3))`) now issues the deprecation warning!

**Comment by kwankyu** [comment]
Does Python issue a specific warning just once? If so, then everything is explained.

**Comment by kwankyu** [comment]
The warning filter seems working here: https://docs.python.org/3/library/warnings.html#the-warnings-filter

**Comment by kwankyu** [comment]
It seems that "always" action is in effect while "default" action is in effect during doctest.

**Comment by kwankyu** [comment]
> Actually there's one more
> 
> ```python
> sage: ZZ(-1)^(1/3)
> (-1)^(1/3)
> sage: ZZ(-1).nth_root(3)
> -1
> ```
> 
> Should we also give warning?

From https://groups.google.com/g/sage-devel/c/s81ieq2vpVo: x^(1/n) and x.nth_root(n) do not behave in the same way. All x.nth_root(n) gives an n-th  root in the same field to which x belongs while all x^(1/n) gives the principal n-th root of unity, with the exception of AA(-1)^(1/n).

```sage
sage: ZZ(-1).nth_root(3)
-1
sage: _.parent()
Integer Ring
sage: QQ(-1).nth_root(3)
-1
sage: _.parent()
Rational Field
sage: RR(-1).nth_root(3)
-1.00000000000000
sage: _.parent()
Real Field with 53 bits of precision
sage: CC(-1).nth_root(3)
0.500000000000000 + 0.866025403784439*I
sage: _.parent()
Complex Field with 53 bits of precision
sage: QQbar(-1).nth_root(3)
0.500000000000000? + 0.866025403784439?*I
sage: _.parent()
Algebraic Field
sage: AA(-1).nth_root(3)
-1
sage: _.parent()
Algebraic Real Field
```
```sage
sage: ZZ(-1)^(1/3)
(-1)^(1/3)
sage: QQ(-1)^(1/3)
(-1)^(1/3)
sage: RR(-1)^(1/3)
0.500000000000000 + 0.866025403784439*I
sage: CC(-1)^(1/3)
0.500000000000000 + 0.866025403784439*I
sage: QQbar(-1)^(1/3)
0.500000000000000? + 0.866025403784439?*I
sage: QQbar(ZZ(-1)^(1/3))
0.500000000000000? + 0.866025403784439?*I
sage: QQbar(QQ(-1)^(1/3))
0.500000000000000? + 0.866025403784439?*I
sage: AA(-1)^(1/3)
-1
```

user202729:
> Just delete the test?

Also okay by me, but the case where the base (`a` part in `a^b`) belong to ℚ and where it doesn't are handled by different code path (e.g. the suggested change in https://github.com/sagemath/sage/pull/38362#issuecomment-2313822146 raise warning when the base belong to ℚ but doesn't when it doesn't), so I figure might as well be careful.

Will apply suggested changes.

user202729:
> It looks like you haven't formatted the expected output exactly as given in the example for `deprecation`:
> 
> https://github.com/sagemath/sage/blob/e042294b5ed8e1cb564965079e20e4270ccde340/src/sage/misc/superseded.py#L91
> 


I see. I follow the example in some other existing code for example

https://github.com/sagemath/sage/blob/e042294b5ed8e1cb564965079e20e4270ccde340/src/sage/geometry/cone.py#L258

https://github.com/sagemath/sage/blob/e042294b5ed8e1cb564965079e20e4270ccde340/src/sage/geometry/toric_lattice_element.pyx#L121

kwankyu:
>  but the case where the base (`a` part in `a^b`) belong to ℚ and where it doesn't are handled by different code path (e.g. the suggested change in [#38362 (comment)](https://github.com/sagemath/sage/pull/38362#issuecomment-2313822146) raise warning when the base belong to ℚ but doesn't when it doesn't), so I figure might as well be careful.

With #38362,
```sage
sage: AA((-sqrt(27)+1)^(5/3))
...
ValueError: Cannot coerce algebraic number with nonzero imaginary part to algebraic real
```
Do you mean something else?

On the other hand,
```sage
sage: AA(-sqrt(27)+1)^(1/3)
-1.612935813642020?
sage: QQbar(-sqrt(27)+1)^(1/3)
0.806467906821010? + 1.396843389287713?*I
```
also with current #38362. I think this should be fixed.

user202729:
In my local testing, `AA(-sqrt(27)+1)^(1/3)` raises the warning but return the real root. Or do you mean it should return the principal root?

kwankyu:
Yes. There should be no difference with `AA(-1)^(1/3)`.

As the discussion on sage-devel goes, there are arguments that `AA(-1)^(1/3)` should give an element in `AA`, that is `AA(-1)^(1/3) == AA(-1).nth_root(3)`. I don't agree with this.

user202729:
I thought the point of giving DeprecationWarning in the first place is that we only change the behavior after half a year or something? Why do you want to change the behavior in this pull request?

kwankyu:
No fix here, only deprecation warnings. When I said " I think this should be fixed.", I meant #38362.

kwankyu:
> ... we only change the behavior after half a year or something.

It is a year after, as the current policy dictates.

kwankyu:
I mentioned #38362 by mistake, instead of #38564. I corrected them. Sorry for confusion.

```sage
sage: AA((-1)^(1/3))
/Users/kwankyu/GitHub/sage-dev/src/sage/symbolic/expression_conversion_algebraic.py:120: DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
See https://github.com/sagemath/sage/issues/38362 for details.
  return self.field(base**expt)
-1
sage: AA((-sqrt(27)+1)^(1/3))
/Users/kwankyu/GitHub/sage-dev/src/sage/symbolic/expression_conversion_algebraic.py:120: DeprecationWarning: Taking the root of an algebraic real number will yield the principal root in the future.
See https://github.com/sagemath/sage/issues/38362 for details.
  return self.field(base**expt)
-1.612935813642020?
```
This messages are not quite right, since these, after #38564, will raise an exception since the cube root is not coercible to AA.

kwankyu:
> Not necessarily. But in order to determine what's the correct new behavior (exception or silent different answer) we need to actually perform the computation.
> 
> ```python
> sage: AA((-1)^(1/9) / sin(1/9*pi) - (-1)^(1/3) * 2 / sqrt(3))  # current behavior
> -1.769103861783836?
> sage: AA(QQbar((-1)^(1/9) / sin(1/9*pi) - (-1)^(1/3) * 2 / sqrt(3)))  # after the change it will be equivalent to (silent different value instead of exception)
> 2.170127150264997?
> 
> sage: AA((-1)^(1/3)*2^(1/3)-(-2)^(1/3))  # current behavior
> 0.?e-18
> sage: QQbar((-1)^(1/3)*2^(1/3)-(-2)^(1/3))  # after the change it will be equivalent to (silent different behavior but value happen to be equal)
> 0.?e-18 + 0.?e-18*I
> ```

You seem to say that the result of `AA(...)` by the future behavior is not predictable, so it is futile to try to give appropriate message depending on the expected result.  Right.

> I think the current warning isn't completely _wrong_ either — it gives principal root, then _as a result of giving principal root_ the conversion ends up with the error.

Right. OK.

By the way, there are still doctest failures.

kwankyu (approved):
Thanks. LGTM.

JohnCremona:
I do not like this, for reasons I stated in the sage-devel thread.  I like AA(-1)^(1/3) returning -1 (in AA).  I would also like RR(-1)^(1/3) to return -1 (in RR).

This negative opinion also applies, of course, to #38564

dimpase:
> I do not like this, for reasons I stated in the sage-devel thread. I like AA(-1)^(1/3) returning -1 (in AA). I would also like RR(-1)^(1/3) to return -1 (in RR).
> 
> This negative opinion also applies, of course, to #38564

I second this. We don't have to ape what Pari etc do here, but rather to respect that people don't want to leave a real field they work in.

user202729:
How about `AA((-1)^(1/3))` ? 

(If that one can be changed then the motivating issue https://github.com/sagemath/sage/issues/36735 is good. Otherwise we got to change the code of `number_field_elements_from_algebraics` directly.)

JohnCremona:
How about this compromise.  There are already nth_root methods to handle exponents of the form 1/n, which (as Dima pointed out) can return a value in the same real parent (AA, RR etc) whenever n is odd;   so it would be easy to define methods to implement raising to a rational power p/q:  x.power(rat) = x.nth_root(rat.denom())**rat.num() or something like that.  If that method existed for AA, I would no longer object to a warning when trying x^(p/q) suggesting that the power() method is also available.

I also note that in Sage, as opposed to python, 1/3 is a Rational and not a float, so when evaluating user code  x^(1/3) with x in AA, Sage would know that the exponent was a Rational.

kwankyu:
> ...define methods to implement raising [an element in AA] to a rational power p/q:

That is a valid concern. Yes, there should be an easy and obvious way to do that. This is a major drawback of changing semantics of AA(x)^(p/q).  Let's ponder upon this.

dimpase:
Python does it wrong, while Fortran does it right (it does not go to a complex type if it can `**` within reals).

While I understand the desire to mimic Python here, it's really questionable, and I would like to know why they went this way

user202729:
As it happens, Python also take the principal root even with `Fraction`:

```
>>> from fractions import Fraction
>>> (-1)**Fraction(1/3)
(0.5000000000000001+0.8660254037844386j)
>>> (-1.)**Fraction(1/3)
(0.5000000000000001+0.8660254037844386j)
```

Probably just an oversight, because casting to `float` is the most natural way.

```python
    def __rpow__(b, a):
        """a ** b"""
        if b._denominator == 1 and b._numerator >= 0:
            # If a is an int, keep it that way if possible.
            return a ** b._numerator

        if isinstance(a, numbers.Rational):
            return Fraction(a.numerator, a.denominator) ** b

        if b._denominator == 1:
            return a ** b._numerator

        return a ** float(b)
```

And once the exponent is `float`, you have no way to "take the principal root".

kwankyu:
> ... so it would be easy to define methods to implement raising to a rational power p/q: x.power(rat) = x.nth_root(rat.denom())**rat.num() or something like that. 

Yes
```diff
diff --git a/src/sage/structure/element.pyx b/src/sage/structure/element.pyx
index e62726dedf7..477c6f1b359 100644
--- a/src/sage/structure/element.pyx
+++ b/src/sage/structure/element.pyx
@@ -2976,6 +2976,10 @@ cdef class RingElement(ModuleElement):
             return False
         return self._parent.ideal(self).is_prime()
 
+    def fractional_power(self, f):
+        m = f.numerator()
+        n = f.denominator()
+        return self.nth_root(n)**m
```
Then
```sage
sage: x = RR(-8)
sage: x.fractional_power(2/3)
4.00000000000000
sage: x = AA(-8)
sage: x.fractional_power(2/3)
4 
```
You name it if there is a better one.

I may implement this in a new PR if the deprecation proposal is adopted.

JohnCremona:
That is exactly what I had in mind!  Thanks for actually implementing it.  (I'll let others determine whether this method is in the right place.  I would suggest just calling the method "power", and making sure that it works with int and Integer as well as Rational arguments (but not e.g. 0.5: perhaps a helpful message could be output if the argument cannot be coerced into QQ).  For RR and AA an error could be raised if the object is negative and the denominator even.

This really is a can of worms:

sage: RR(-1).sqrt()
1.00000000000000*I
sage: RR(-1).nth_root(2)
ValueError: taking an even root of a negative number
sage: RR(-1)^(1/2) # same as RR(-1)^(0.5)
6.12323399573677e-17 + 1.00000000000000*I

Does anyone *not* agree that x.sqrt() and x.nth_root(2) should always do the same thing and give the same answer?  sqrt() has an argument extend with default True, hence the complex answer for negative x.  But nth_root does not.  I would suggest redefining sqrt() on RR to be nth_root(2), but of course that would break some code and would require a deprecation.  And/or add an extend keyword to nth root for the case x<0, n even....

dimpase:
probably `pow()` rather than `power()`, to match how it's done for the standard Python numerical tower.

kwankyu:
> Does anyone _not_ agree that x.sqrt() and x.nth_root(2) should always do the same thing and give the same answer? 

I agree.

>  sqrt() has an argument extend with default True, hence the complex answer for negative x. But nth_root does not. I would suggest redefining sqrt() on RR to be nth_root(2), but of course that would break some code and would require a deprecation. And/or add an extend keyword to nth root for the case x<0, n even....

Not in the scope of the present PR.

kwankyu:
> ... I would suggest just calling the method "power", and making sure that it works with int and Integer as well as Rational arguments ...

I added `pow` method, with the name suggested by Dima, in #38564. See https://github.com/sagemath/sage/pull/38564/commits/541c338670426b2b4b9bbf6b2e71dba1703b8836.

We may proceed as follows:

1. Test with #38564. 
2. If all of you are positive with #38564, the code for `pow` method is copied into this PR.
3. This PR is reviewed.
4. If this PR is merged, then a year later #38564 is merged.   

During the deprecation period, you may replace AA(x)^(p/q) or AA(x^(p/q)) with AA(x).pow(p/q).

hgranath:
Just a humble suggestion: Why not call it `__pow__` instead of `pow`, so one could use the notation

  `AA(-1)**(1/3)`

instead of the more awkward

  `AA(-1).pow(1/3)`

Sarcasm aside, would it not be very confusing and annoying to users if the first notation do not work in the special case of `AA`?

It seems to me that the discussion has gone full circle, and that the new function `pow`, which gives the same result as `AA(-1)**(1/3)` currently does, indeed is doing the right thing in `AA`. After all, what is the purpose of having the subclass `AA` of `QQbar` if it is not permitted to have some different behaviors?

I personally would be annoyed if the use of `AA(-1)**(1/3)` is deprecated. I would imagine `AA` is more of a specialist's than a beginner's tool (as far as I am aware the corresponding functionality only exists in Sagemath, I know `QQbar` also exist i Flint but if I am not mistaken they do not seem to have something corresponding to `AA`), and a user using `AA` for its intended use cases would, I assume,  probably expect `AA(-1)**(1/3)` to be real. And if not, just cast the element to `QQbar` before taking the power. Explicit is better than implicit.

Just 2 cents from a usually silent Sagemath user.

kwankyu:
> Just a humble suggestion: Why not call it `__pow__` instead of `pow`, so one could use the notation
> 
> `AA(-1)**(1/3)`
> 
> instead of the more awkward
> 
> `AA(-1).pow(1/3)`
>

`AA(-1)**(1/3)` would still work but give the same result as `P(-1)**(1/3)` for other parents P = ZZ,QQ,RR,CC,QQbar.

kwankyu:
> .... After all, what is the purpose of having the subclass `AA` of `QQbar` if it is not permitted to have some different behaviors?

`AA` behaves differently in some other methods. `AA(-8).nth_root(3) != QQbar(-8).nth_root(3)`.

kwankyu:
We start voting.

+1 from me.

kwankyu:
@user202729 Would you copy the change in src/sage/structure/element.pyx from https://github.com/sagemath/sage/commit/541c338670426b2b4b9bbf6b2e71dba1703b8836 ? You may improve it in any way.

dimpase:
I'm -1 here - as I think blindly following Python here is wrong.

JohnCremona:
-1 to be consistent with what I have been saying.  But I think that this pr has changed a bit since my first comments so if my misgivings have been addressed I could change my vote.

user202729:
I guess I'll wait until a consensus is reached before implementing. (

Note to self (if this gets implemented): instead of isinstance do something like

```python
        try:
            exponent = QQ(exponent)
```

adapt from `src/sage/rings/asymptotic/asymptotic_ring.py`.

nbruin:
> Note to self (if this gets implemented): instead of isinstance do something like
> 
> ```python
>         from sage.rings.rational_field import QQ
>         try:
>             exponent = QQ(exponent)
>         except (TypeError, ValueError):
> ```

Ouch! No, don't do that!! `QQ(pi.n())` yields a rational number! Conversion can really mean "is there any way in which you could interpret this as an element of ..."? Perhaps try to coerce in to QQ, but in this case you're really looking at the "action" of QQ via exponentiation. There could be other actions. Can't you just let the coercion system figure out what type your exponent should have?

nbruin:
For the vote: Personally I think `.nth_root` is the important function to have, and that one should NOT extend the parent (so, it should raise an error on negative elements with an even denominator), and I think it already does that.

I would expect on something that is defined to be a real algebraic, `x^(1/3)` to again be a real algebraic, but the use is so ambiguous in general that I'm not super-attached to that. Rather, I'd advise people to avoid fractional exponents on algebraic numbers and spell out what you mean. I know I'd be super annoyed if I got hit by a deprecation warning for something like exponentiation, so you should be really convinced that you'll make sage significantly better before subjecting users to a deprecation. It would be good to have a really convincing real world example where the changed behaviour is really essential.

hgranath:
-1 from me, if a non-developer's vote counts

Would the proposed change not introduce a bug, since the documented description of `AA` no longer would be valid?

>   Also, ``AlgebraicNumber`` and ``AlgebraicReal`` are closed
    under addition, subtraction, multiplication, division (except by
    0), and rational powers (including roots), except that for a
    negative ``AlgebraicReal``, taking a power with an even denominator returns
    an ``AlgebraicNumber`` instead of an ``AlgebraicReal``.

At least I could not find in any PR that the documentation would be corrected, but I might miss something.

kwankyu:
> ... I know I'd be super annoyed if I got hit by a deprecation warning for something like exponentiation, ...

With the deprecation warning, there are just a few doctest failures that this PR fixes. It implies that fractional powers of algebraic reals are not used much in the sage library.

Ordinary users who used ``AA(x)^(2/3)``, for example, in their code will meet the deprecation warning, visit this PR, and will know that ``AA(x).pow(2/3)`` does the same computation, and change their code accordingly.

> ... so you should be really convinced that you'll make sage significantly better before subjecting users to a deprecation. 

This change improves sage to be more consistent, and help fixing a couple of defects. 

> It would be good to have a really convincing real world example where the changed behaviour is really essential.

There is no such thing. Note that we have lived with the inconsistency more than a decade. I can live and die undisturbed with the inconsistency. Dima claims that (-1)**(1/3) giving the principal root is a design flaw of python. He is living well with it :-)

kwankyu:
> -1 from me, if a non-developer's vote counts

You are welcome. Just be responsible.

> Would the proposed change not introduce a bug, since the documented description of `AA` no longer would be valid?
> 
> > Also, `AlgebraicNumber` and `AlgebraicReal` are closed
> > under addition, subtraction, multiplication, division (except by
> > 0), and rational powers (including roots), except that for a
> > negative `AlgebraicReal`, taking a power with an even denominator returns
> > an `AlgebraicNumber` instead of an `AlgebraicReal`.
> 
> At least I could not find in any PR that the documentation would be corrected, but I might miss something.

The documentation should be updated in this PR. The current behavior
```sage
sage: AA(-1)^(1/4)
0.7071067811865475? + 0.7071067811865475?*I
sage: AA(-1)^(1/3)
-1
```
of course is consistent with the doc. This PR regards that as inconsistent behavior.

user202729:
@nbruin Alright, noted.

Some part in the existing code sort of do that though…

```python
sage: B = Berkovich_Cp_Affine(Qp(3))
sage: B(3, power=pi.n())
Type II point centered at 3 + O(3^21) of radius 3^245850922/78256779

sage: R.<x> = AsymptoticRing(growth_group='x^ZZ', coefficient_ring=ZZ)
sage: (1 + x)^(pi.n())
x^(245850922/78256779) + 245850922/78256779*x^(167594143/78256779) + 20601587289174923/6124123459454841*x^(89337364/78256779) + 1840491502630793355722972/1437762528405818857851417*x^(11080585/78256779) + 5098430634169557348880906924655/112514664439935388672910755005843*x^(-67176194/78256779) - 68498633075303442672509897293313532614/8805035229335182485655080241215399359697*x^(-145432973/78256779) + 4980979927288756272499089867645620034593240711/2067161088087893078113740853992180597265648907889*x^(-223689752/78256779) - 159170594950314274711966547531337367060627633645699096/161769368427893781189576754874137344728305890876260829531*x^(-301946531/78256779) + 6007626122806689076132415401891741471828772582757215638379497/12659549714031261070027065209722539002049849146701660082964140649*x^(-380203310/78256779) - 2284119337133569676886386334094220369253571089201542312095647795535070/8916260257894118369835704095222608474020660324910933528410817777243386139*x^(-458460089/78256779) + 104717755458887735895303392161722000487360506512316602734063646485366199482123/697757808508503026668072961689230627154962056480623119816535588003006898291386281*x^(-536716868/78256779) - 5109435066716738943030764777346834689842782240198560949438807023116814038100752595524/54604278615974240978794492118789587869297264407389641269773146055986362175064493795848899*x^(-614973647/78256779) + 785541979272120315935638662081031478279682412720434757807041439583840109007904194278527539007/12819464892314166137910792768472075576166030718100971371213749413114139115243944205185855219308963*x^(-693230426/78256779) - 41889353917822702810716721407423346015941075854991908158444781766335287806418143228222339116811863614/1003210030976088498023768431397117386035322733213639016282401349583452886516900872753560125823458050210177*x^(-771487205/78256779) + 2308364326665488334070391960384050247827590432575763751626804416553212448760293884316494965985096655029004205/78507985684678911874287982883020876516023937326775708282989284003654014597065191363982476229778549651068725039883*x^(-849743984/78256779) - 392303739892841898459699560171662605529040807859442574430029453222144458841593597253986709862424103253883933741890544/18431346256383243877519891376497042757702325666274012157011085572926562238355914187491655606559599536952636987773670350571*x^(-928000763/78256779) + 22753635621769426250560607286879206806844868021981194114109476418913929143826309649225898221511824553074692076619741868405317/1442377790658260855426177207554534869243061447451633122814529824230602364316764099913499057146645531291804846219429822643487270809*x^(-1006257542/78256779) - 1346824556019137914684670165324844676021493274771364609381686427070781920578166283680054973360038829336011070083897477716236573185342/112875839998051784287437300546432365710148156976642566481176518443723094261214494162064614831833409933640356361722905136620579141013064211*x^(-1084514321/78256779) + 730325259438610909274800496728115834122957380147366459793504942644792536767452967285184100338117799765489463260260743039518468921733993141391/79499696986502090221838879746968631936436181400005829383992146174558759123164278471137326719534183933939682299525341018630333218212422818657327321*x^(-1162771100/78256779) - 44694795014485209971024209255847244703714141482477445292484185000764333013099038472024080529087558208573575012291735287940959662023181532231982557900/6221390217639660057528506185966100149382028295424166788814779520918140235215700721010251655741381615063668317044166416994588869354008347574223540890159059*x^(-1241027879/78256779) + O(x^(-1319284658/78256779))

sage: Cusp(pi.n())
245850922/78256779

sage: Cusp(pi.n(), 2)
122925461/78256779

sage: QuadraticField(pi.n())
Number Field in a with defining polynomial x^2 - 245850922/78256779 with a = 1.772453850905516?
```

dimpase:
Sorry, the dispute here transcends the question whether this PR should go ahead. The dispute is in fact about the 3 alternatives I outlined, the alternatives discussed on sage-devel.

I don't think it's correct to vote exclusively on this PR at all, the vote should be about these 3 alternatives.

dimpase:
@kwankyu as you are a proponent of 2), I think you have jumped ahead with this PR and have created a bias towards 2) before the decision on the 3 alternatives was made. 

I don't like this. We really should first plan, and only then implement, not the other way around.

kwankyu:
I summarize your comments as vote -1.

dimpase:
I would rather like you calling this vote off, and do a vote on the 3 (or more?) alternatives.

nbruin:
> > It would be good to have a really convincing real world example where the changed behaviour is really essential.
> 
> There is no such thing. Note that we have lived with the inconsistency more than a decade. I can live and die undisturbed with the inconsistency. Dima claims that (-1)**(1/3) giving the principal root is a design flaw of python. He is living well with it :-)

In that case I think the cons of changing outweigh the pros of the change and then we're better off keeping the current behaviour for the sake of not disturbing user code and/or the annoyance of a deprecation message. Perhaps until someone finds a convincing reason to overcome the cons. @mezzarobba had some more spirited argument for changing rational (odd denominator) exponentiation behaviour on AA.

kwankyu:
The inconsistency has been repeatedly demonstrated in my comments here and in https://github.com/sagemath/sage/pull/38564. I think the inconsistency justifies changing the behavior. How one weighs solving inconsistency vs keeping the status quo depends on one's value system. Each one has his own preference. I can't change that.

JohnCremona:
On Tue, 3 Sept 2024, 17:49 Kwankyu Lee, ***@***.***> wrote:

> The inconsistency has been repeatedly demonstrated in my comments here and
> in #38564 <https://github.com/sagemath/sage/pull/38564>. I think the
> inconsistency justifies changing the behavior. How one weighs solving
> inconsistency vs keeping the status quo depends on one's value system. Each
> one has his own preference. I can't change that.
>

I don't think we all agree that there an inconsistency.

John

kwankyu:
> I don't think we all agree that there an inconsistency. John —

Yes. That is why we are voting now. I count your comment as -1 vote.

user202729:
There are two parts of the behavior here. One is `AA(-1)^(1/3)` and the other is `AA((-1)^(1/3))`.

People seems to be `-1` the first one, but what do you think about the second one?

Note that not fixing that one will make the diagram

```
SR ---> AA
   \    |
    \   |
     v  v
      QQbar
```

non-commutative.

nbruin:
Probably useful for people who try to make their minds up:
```
sage: parent(AA(-2)^RealField(1000)(1/3))
Complex Field with 1000 bits of precision
```
The coercion system in sage is so eager that it coerces `-2` into `RealField(1000)` in order to evaluate this. I was expecting this to raise an error, but it does not. So sage forces our hand by having the exponent space QQ embedded in the reals, so it gets a real topology. So unless we curtail the coercion system with respect to the exponentiation operator, I think our hand is forced to follow the branch choice on RealField.

kwankyu:
The coercion in powering was introduced by Jeroen Demeyer in 2017 https://github.com/sagemath/sage/issues/24247.

hgranath:
I think certain examples and arguments here are a bit misleading for the discussion at hand. What I, and I assume other number theorists, care so much about is the case `x**r` where `x` is in `AA` and `r` is a `Rational` (and of course this is the _only_ [[1]](https://en.wikipedia.org/wiki/Gelfond%E2%80%93Schneider_theorem) case I will ever care about in this context).

Any examples in pure python, or e.g. this example,

>   sage: parent(AA(-2)^RealField(1000)(1/3))
>   Complex Field with 1000 bits of precision

is about `x**f` where `f` is (some kind of) floating point number. That has absolutely no meaning in `AA` or `QQbar` (and Sagemath apparently does the only reasonable thing and casts/coerces `x` to a float before taking the power, I see no problem with that).

[1] [Gelfond–Schneider theorem](https://en.wikipedia.org/wiki/Gelfond%E2%80%93Schneider_theorem)

kwankyu:
> ... What I, and I assume other number theorists, care so much about is the case `x**r` where `x` is in `AA` and `r` is a `Rational` 

After deprecation, you may use `x.pow(r)` for the computation you care. What you really care is the notation `**`?

JohnCremona:
> I think certain examples and arguments here are a bit misleading for the discussion at hand. What I, and I assume other number theorists, care so much about is the case `x**r` where `x` is in `AA` and `r` is a `Rational` (and of course this is the _only_ [[1]](https://en.wikipedia.org/wiki/Gelfond%E2%80%93Schneider_theorem) case I will ever care about in this context).
> 
> Any examples in pure python, or e.g. this example,
> 
> > sage: parent(AA(-2)^RealField(1000)(1/3))
> > Complex Field with 1000 bits of precision
> 
> is about `x**f` where `f` is (some kind of) floating point number. That has absolutely no meaning in `AA` or `QQbar` (and Sagemath apparently does the only reasonable thing and casts/coerces `x` to a float before taking the power, I see no problem with that).
> 
> [1] [Gelfond–Schneider theorem](https://en.wikipedia.org/wiki/Gelfond%E2%80%93Schneider_theorem)

I agree with this.  When a is something algebraic (not a floating point in RR or CC) and I see a^(1/n) or a**(1/n) I interpret it is "a root of the polynomial X^n-a"; so if there is just one such root in a's parent (e.g. AA) then that's what I expect to see.

kwankyu:
Do you think the current behavior 
```sage
sage: AA(-1)^(1/4)
0.7071067811865475? + 0.7071067811865475?*I
```
should be fixed?

nbruin:
> I think certain examples and arguments here are a bit misleading for the discussion at hand. What I, and I assume other number theorists, care so much about is the case `x**r` where `x` is in `AA` and `r` is a `Rational` (and of course this is the _only_ [[1]](https://en.wikipedia.org/wiki/Gelfond%E2%80%93Schneider_theorem) case I will ever care about in this context).

I agree, but the problem is that sage *automatically coerces* to floats, so implicitly QQ (for exponents) is imbued with the real metric. That means people can use AA-elements seamlessly in contexts where the continuity in the exponent is expected. To illustrate:
```
sage: eps=RR(2^-50)
sage: abs(AA(-2)^(1/3)-AA(-2)^(1/3+eps))  #notice the large jump here?
2.18224727194344
sage: abs(AA(-2)^(1/3+eps)-AA(-2)^(1/3+2*eps))
3.53358019796514e-15
```
The coercion framework is a little too eager. I think it should not just by default coerce into a common parent for exponentiation. It's rare that exponentiation designates a binary operation instead of an action. But that's such an invasive change that we'd need a rather convincing reason to implement that. (Jeroen's argument was: multiplication also often designates an action and there we try common parent as well. I'm not sure I'm convinced by that reasoning. Rings are rather central objects in mathematics)

kwankyu:
I am curious if number theorists are happy with
```sage
sage: a = QQbar(1)
sage: (AA(-2) + a - a)^(1/3)
0.6299605249474365? + 1.091123635971722?*I
sage: AA(-2)^(1/3)
-1.259921049894873?
```
I guess such a weird example cannot be constructed in sage without using `AA`.

JohnCremona:
I should not presume to speak for all number theorists, but it is likely that they would be happy that for a in any exact field or ring (including AA and QQbar, number fields, fine fields), we would only allow a**e or a^e for *integer* exponents e.  That only uses the ring structure.  For rational powers, which may not exist or be unique within the same ring or field, we can have methods like nth_root() or pow() as shortcuts to getting roots of polynomials like X^n-a or X^d-a^n (for rational exponent n/D).

All the issues about choices of branch are relevant over CC (or RR as a subfield of CC perhaps) but not for exact fields.

dimpase:
Mixing QQbar and AA in one expression is asking for 😵‍💫

kwankyu:
> All the issues about choices of branch are relevant over CC (or RR as a subfield of CC perhaps) but not for exact fields.

For a happy way out, see https://groups.google.com/g/sage-devel/c/s81ieq2vpVo/m/GNw9w_jpAAAJ

hgranath:
> sage: eps=RR(2^-50)
> sage: abs(AA(-2)^(1/3)-AA(-2)^(1/3+eps))  #notice the large jump here?
> 2.18224727194344
> sage: abs(AA(-2)^(1/3+eps)-AA(-2)^(1/3+2*eps))
> 3.53358019796514e-15

I guess the problem here is again with the case of `x**f` for floats `f`. If `x` is in `AA` but is negative, maybe it should not be allowed at all, so maybe throw a `ValueError` or something?

If something like this happened in my computations I would take it as a sign that I am doing something very stupid, so it would be good to be clearly alerted :-)

nbruin:
> I guess the problem here is again with the case of `x**f` for floats `f`. If `x` is in `AA` but is negative, maybe it should not be allowed at all, so maybe throw a `ValueError` or something?

That's to some extent the RBF solution (where the rigorous ball results indeed suggest that for balls covering the branch cut you should decline giving an answer, so they have an additional reason to return a NaN here. Indeed, for CBF you see that for balls covering the branch cut  you get

> 
> If something like this happened in my computations I would take it as a sign that I am doing something very stupid, so it would be good to be clearly alerted :-)

nbruin:
apologies. I clicked the wrong button which "closed" the PR. I've reopened it. As far as I can see it's now restored to its original state, but let me know if there is something further I should do to restore.

kwankyu:
No problem. That happens :-)

kwankyu:
The tally is:

negative votes: @JohnCremona @dimpase @hgranath
positive votes: @kwankyu @user202729 

Thank you for your participation in the discussion. It seems that algebraists like the current behavior of AA(-1)^(1/3) and think how other analytical fields behave as irrelevant. That is understandable.

@user202729 Sorry for the outcome. It is wise that you try to solve your problem at hand in another PR https://github.com/sagemath/sage/pull/38606. I hope that number theorists here pay attention to the PR.

nbruin:
I'm still on the fence and I think other people might change their opinion as well, depending on the further evidence that arises. I think the discussion here shows that the majority of the participants prefer a branch choice in the algebraic reals that does not extend the base field if possible, and this is also the documented behaviour for AA. That's for AA *in isolation*.

Note that for interactive work or for intentional code designed specifically for a particular computation, one can freely choose to use `nth_root` or `^`, depending on what gives the desired result. The change is the painful part in those cases; particular in code that has been written for sage and is not part of sage, since that won't be picked up with doctests.

The real complications from design choices around `^` come from examples that implicitly use the coercion framework -- coercion usually only gets complicated from the second level onwards, at least for unintended consequences.

For instance, people may not have good control over whether an element gets created in `AA` or in`QQbar`: for a polynomial `f=x^3+x+1` over `QQ`, the command `f.roots(QQbar)` gives back one root that has parent `AA` and two others that lie in `QQbar`.

If someone came up with a reason to evaluate, say, Puiseux polynomials (an expression like `x^(1/5)+x^(2/5)`) at the result, they would be exposed to the different branch cuts between AA and QQbar without asking for it. I'm currently not convinced by this example because I don't see a mathematical purpose in doing so and if there were I would expect one would be better off doing this with floats anyway, but I can imagine that an example along these lines that does make mathematical sense, or something else, would show quite nasty complications.

hgranath:
On Thu, Sep 5, 2024 at 6:15 PM nbruin ***@***.***> wrote:

> For instance, people may not have good control over whether an element
> gets created in AA or inQQbar: for a polynomial f=x^3+x+1 over QQ, the
> command f.roots(QQbar) gives back one root that has parent AA and two
> others that lie in QQbar.
>

Sounds like a bug in the roots method to me if that is indeed the case. But
maybe it has been fixed lately? In Sagemath 10.4 I get

sage: R.<x> = QQ[]
....: f = x^3 + x + 1
....: for r in f.roots(QQbar):
....:     print(parent(r[0]))
....:
Algebraic Field
Algebraic Field
Algebraic Field

PR #38365, opened 2024-07-14
add access to printlevel in libsingular

fchapoton:
this is adding and using a simplified access to libsingular ``printlevel`` state variable

useful in turning down the comments easily

### :memo: Checklist


fchapoton:
maybe we should also create a `set_printlevel` method too ?

kwankyu:
> maybe we should also create a `set_printlevel` method too ?

Good idea.

fchapoton:
done and used

kwankyu (approved):
Thanks. LGTM.

PR #38366, opened 2024-07-14
Check input of composition

mantepse:
Fix #14862

### DIFF
--- a/src/sage/combinat/composition.py
+++ b/src/sage/combinat/composition.py
@@ -155,6 +155,20 @@ class Composition(CombinatorialElement):
             [1, 0, 0, 0, 1, 1, 0, 1, 0, 0, 1, 0, 0, 0, 0]
             sage: Composition(code=_)
             [4, 1, 2, 3, 5]
+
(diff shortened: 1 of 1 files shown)


mantepse:
@fchapoton, I think that this is an improvement, so let's get it in.

In the longer run I'd really like to make `Composition` only accept positive integers, and use `IntegerVector` for weak compositions (perhaps `WeakComposition` should be an alias).  However, this can be done incrementally, I am currently doing it for `tableau.py`.

mantepse (commented):
Looks good to me.  Perhaps we should have a `check` argument to bypass the checks, but I don't think it is necessary right now.

mantepse:
@fchapoton, I just realised that I cannot approve this formally, because I created the pull request...

fchapoton:
I think it would be good to have Travis' opinion on that. @tscrim what do you think ?

tscrim:
I think it would be good to add a `check` to the `__init__`, but I wouldn't make a requirement for setting this to a positive review.

PR #38367, opened 2024-07-14
convert two methods to libsingular in affine subscheme

fchapoton:
this is using #38365 to convert some methods to use `libsingular`

### :memo: Checklist

### :hourglass: Dependencies

#38365


fchapoton:
not clear to me how to avoid the remaining verbosity...

fchapoton:
so there is a failing doctest, damn

PR #38368, opened 2024-07-14
tableau/do not abuse composition

mantepse:
`Compositions` are lists of *positive* integers, whereas `IntegerVectors` are lists of *non-negative* integers, according to the documentation.

Nevertheless, `Compositions` are abused for lists of non-negative integers in several places.  Since `Composition` did not check its input at all before #38366, and only for non-negative after #38366, this does not raise errors.  Note, however, at least some methods of `Composition` do assume positivity.

Here we fix one such instance.


mantepse:
I will disclose one potential problem: `IntegerVectors` compare differently to `Compositions` and `Partitions`:
```
sage: a = IntegerVectors()([2,1])
sage: b = Composition([2,1])
sage: c = Partition([2,1])
sage: d = [2,1]
sage: e = tuple([2,1])
sage: matrix([[ZZ(x == y) for x in [a,b,c,d,e]] for y in [a,b,c,d,e]])
[1 0 0 0 0]
[0 1 1 1 0]
[0 1 1 1 0]
[0 1 1 1 0]
[0 0 0 0 1]
```

tscrim:
This is a constant pet-peeve of mine with combinatorial object comparisons. I think comparing with, e.g., `list` is what should be done, but it means duplicating code. We probably should have some mix-in class that says any subclass should consider itself list-like for comparisons. One counterpoint would be that tuples and lists do not compare as equal in Python, but I don't think that is a strong argument as we could consider it an implicit coercion.

fchapoton:
Why is this on top of the other pull request ? This seems to be very independent.

mantepse:
> Why is this on top of the other pull request ? This seems to be very independent.

I agree.  It happened, because I used the other pull request (modified to check for positivity instead of non-negativity) to look for abuse.

mantepse:
> This is a constant pet-peeve of mine with combinatorial object comparisons. I think comparing with, e.g., `list` is what should be done, but it means duplicating code. We probably should have some mix-in class that says any subclass should consider itself list-like for comparisons. One counterpoint would be that tuples and lists do not compare as equal in Python, but I don't think that is a strong argument as we could consider it an implicit coercion.

Would it make sense to have `ClonableList` compare with `list`?

tscrim:
`ClonableList` base comparison is worse; it takes the parent into account by design (which has lead to subtle comparisons returning false without coercions). A separate mixin class is the way to go IMO; it also gives a class-level indication of how the object should compare.

mantepse:
OK.  I think I don't know what to do precisely, though.

I'm afraid that I somehow the list of things I'd like to do gets longer and longer, and the list of things I actually finished does not :-(.

mantepse:
I think this one is ready, the difference in equality is minor, I would say.

mantepse:
I now removed the abuse in a few more easy cases.

Within combinat, the remaining abuses are:
```
src/bin/sage -t src/sage/combinat/colored_permutations.py  # 13 doctests failed
src/bin/sage -t src/sage/combinat/k_tableau.py  # 1 doctest failed
src/bin/sage -t src/sage/combinat/superpartition.py  # 2 doctests failed
src/bin/sage -t src/sage/combinat/affine_permutation.py  # 24 doctests failed
src/bin/sage -t src/sage/combinat/set_partition_ordered.py  # 4 doctests failed
```

For example, `OrderedSetPartition` is documented as:
```
   An ordered partition of a set.

   An ordered set partition p of a set s is a list of pairwise
   disjoint nonempty subsets of s such that the union of these subsets
   is s. These subsets are called the parts of the partition.
```
I think that, mathematically, this is entirely correct and allowing empty sets as part is not a good idea in the long run.

However, `OrderedSetPartition` is (ab)used in several places for, what I would call dissections of a set, i.e., where the number of parts is fixed.

In any case, this ticket is ready for review.

src/sage/combinat/superpartition.py
```diff
@@ -322,6 +327,7 @@ def to_composition(self) -> Composition:
             [2, 1, 0, 3, 3]
             sage: SuperPartition([[2,1,0],[3,3]]).to_composition().parent()
             Compositions of nonnegative integers
+
```
fchapoton:
```suggestion
```

src/sage/combinat/composition_tableau.py
```diff
@@ -187,19 +188,20 @@ def size(self):
...
-        Return a composition where entry `i` is the number of times that `i` appears in
-        ``self``.
+        Return a (weak) composition where entry `i` is the
+        number of times that `i` appears in ``self``.
 
         EXAMPLES::
 
             sage: CompositionTableau([[1],[3,2],[4,4]]).weight()
             [1, 1, 1, 2, 0]
+
```
fchapoton:
```suggestion
```

src/sage/combinat/composition_tableau.py
```diff
@@ -187,19 +188,20 @@ def size(self):
...
         """
         w = {i: 0 for i in range(1, self.size() + 1)}
         for row in self:
             for i in row:
                 w[i] += 1
-        return Composition([w[i] for i in range(1, self.size() + 1)])
+        return IntegerVectors()([w[i] for i in range(1, self.size() + 1)])
```
fchapoton, line 201: This could use a list for w instead of a dict, no ?

fchapoton (approved):
ok, good

PR #38371, opened 2024-07-15
Disjoint direct product decomposition of a permutation group

Newtech66:
This PR implements the disjoint direct decomposition of a permutation group, which is a partition of its domain such that the group is isomorphic to the direct product of its projection onto each part of the partition. Each part is a union of orbits.

The algorithm used is an adaptation of [https://arxiv.org/abs/2004.11618v3](https://arxiv.org/abs/2004.11618v3). Additionally, the output of this algorithm is guaranteed to be the _finest_ such partition, that is the groups formed from each part are themselves d.d.p indecomposable.

An implementation of the algorithm in the paper by the authors is available at [https://github.com/peal/DisjointDirectProdDecomposition](https://github.com/peal/DisjointDirectProdDecomposition).

### :memo: Checklist

### DIFF
--- a/src/sage/groups/perm_gps/permgroup.py
+++ b/src/sage/groups/perm_gps/permgroup.py
@@ -1581,6 +1581,74 @@ class PermutationGroup_generic(FiniteGroup):
         p = self._libgap_().SmallestMovedPoint()
         return self._domain_from_gap[Integer(p)]
 
+    @cached_method
+    def disjoint_direct_product_decomposition(self):
+        r"""
+        Returns the finest partition of `self.domain()` such that `self`
+        is isomorphic to the direct product of the projections of `self`
+        onto each part of the partition. Each part is a union of orbits
+        of `self`.
+
+        EXAMPLES::
+            sage: H = PermutationGroup([[(1,2,3),(7,9,8),(10,12,11)],[(4,5,6),(7,8,9),(10,11,12)],[(5,6),(8,9),(11,12)],[(7,8,9),(10,11,12)]])
+            sage: S = H.disjoint_direct_product_decomposition();S
+            {{1, 2, 3}, {4, 5, 6, 7, 8, 9, 10, 11, 12}}
+            sage: A = libgap.Stabilizer(H, list(S[0]), libgap.OnTuples);A
+            Group([ (7,8,9)(10,11,12), (5,6)(8,9)(11,12), (4,5,6)(7,8,9)(10,11,12) ])
+            sage: B = libgap.Stabilizer(H, list(S[1]), libgap.OnTuples);B
+            Group([ (1,2,3) ])
+            sage: T = PermutationGroup(gap_group=libgap.DirectProduct(A,B))
+            sage T.is_isomorphic(H)
+            True
+        """
+        from sage.combinat.set_partition import SetPartition
+        from sage.sets.disjoint_set import DisjointSet
+        H = self._libgap_()
+        if self.is_trivial():
+            return SetPartition(DisjointSet(self.domain()))
+        if libgap.NrMovedPoints(H) == self.degree() and libgap.IsTransitive(H):
+            return SetPartition([self.domain()])
+        O = libgap.Orbits(H)
+        k = len(O)
+        OrbitMapping = dict()
+        for i in range(k):
+            for x in O[i]:
+                OrbitMapping[x]=i
+        C = libgap.StabChain(H,libgap.Concatenation(O))
+        X = libgap.StrongGeneratorsStabChain(C)
+        P = DisjointSet(k)
+        R = libgap.List([])
+        identity = libgap.Identity(H)
+        for i in range(k-1):
+            libgap.Append(R,O[i])
+            Xp = libgap.List([])
+            while True:
+                try:
+                    if libgap.IsSubset(O[i],C['orbit']):
+                        C = C['stabilizer']
+                    else:
+                        break
+                except ValueError:  #this should catch a GAPError but I don't know how to make it work
+                    break
+            for x in X:
+                xs = libgap.SiftedPermutation(C,x)
+                if xs != identity:
+                    cj = OrbitMapping[libgap.SmallestMovedPoint(libgap.RestrictedPerm(x,R))]
+                    libgap.Add(Xp,xs)
+                    if libgap.RestrictedPerm(xs,O[i+1]) != identity:
+                        P.union(i+1,cj)
+                else:
+                    libgap.Add(Xp,x)
+            X = Xp
+        final_partition = DisjointSet(self.domain())
+        for part in P:
+            grp = [self._domain_from_gap[Integer(x)]
+                   for i in part
+                   for x in O[i]]
+            for i in range(1,len(grp)):
+                final_partition.union(grp[0],grp[i])
+        return SetPartition(final_partition)
+
     def representative_action(self, x, y):
         r"""
         Return an element of ``self`` that maps `x` to `y` if it exists.
--- a/src/sage/groups/perm_gps/permgroup.py
+++ b/src/sage/groups/perm_gps/permgroup.py
@@ -1612,9 +1612,9 @@ class PermutationGroup_generic(FiniteGroup):
             sage: PermutationGroup(PermutationGroup(gap_group=B).gens(),domain=list(S[0])).disjoint_direct_product_decomposition()
             {{1, 2, 3}}
 
-        Counting the number of connected subgroups::
+        Counting the number of "connected" permutation groups of degree `n`::
 
-            sage: seq = [sum(1 for G in SymmetricGroup(n).conjugacy_classes_subgroups() if len(G.disjoint_direct_product_decomposition()) == 1) for n in range(1,8)];seq
+            sage: seq = [sum(1 for G in SymmetricGroup(n).conjugacy_classes_subgroups() if len(G.disjoint_direct_product_decomposition()) == 1) for n in range(1,8)]; seq
             [1, 1, 2, 6, 6, 27, 20]
             sage: oeis(seq) # optional -- internet
             0: A005226: Number of atomic species of degree n; also number of connected permutation groups of degree n.
--- a/src/sage/groups/perm_gps/permgroup.py
+++ b/src/sage/groups/perm_gps/permgroup.py
@@ -1590,10 +1590,12 @@ class PermutationGroup_generic(FiniteGroup):
         of ``self``.
 
         The algorithm is from [CJ2022]_, which runs in time polynomial in
-        `n \cdot |G|`, where `n` is the degree of the group and `|G|` is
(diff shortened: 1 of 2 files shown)


src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,74 @@ def smallest_moved_point(self):
...
         return self._domain_from_gap[Integer(p)]
 
+    @cached_method
+    def disjoint_direct_product_decomposition(self):
+        r"""
+        Returns the finest partition of `self.domain()` such that `self`
+        is isomorphic to the direct product of the projections of `self`
+        onto each part of the partition. Each part is a union of orbits
+        of `self`.
+
```
mantepse, line 1599:
```suggestion

The algorithm is from [CJ2022]_, which runs in time polynomial in `n\cdot g`, where `n` is the degree of the group and `g` the number of generators, see Theorem 4.5.
```

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,74 @@ def smallest_moved_point(self):
...
+        if self.is_trivial():
+            return SetPartition(DisjointSet(self.domain()))
+        if libgap.NrMovedPoints(H) == self.degree() and libgap.IsTransitive(H):
+            return SetPartition([self.domain()])
+        O = libgap.Orbits(H)
+        k = len(O)
+        OrbitMapping = dict()
+        for i in range(k):
+            for x in O[i]:
+                OrbitMapping[x]=i
```
mantepse: before and after assignment `=` there should be a space (except when used for keyword arguments, and after `,` there should also be a space.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,74 @@ def smallest_moved_point(self):
...
+        EXAMPLES::
+            sage: H = PermutationGroup([[(1,2,3),(7,9,8),(10,12,11)],[(4,5,6),(7,8,9),(10,11,12)],[(5,6),(8,9),(11,12)],[(7,8,9),(10,11,12)]])
+            sage: S = H.disjoint_direct_product_decomposition();S
+            {{1, 2, 3}, {4, 5, 6, 7, 8, 9, 10, 11, 12}}
+            sage: A = libgap.Stabilizer(H, list(S[0]), libgap.OnTuples);A
+            Group([ (7,8,9)(10,11,12), (5,6)(8,9)(11,12), (4,5,6)(7,8,9)(10,11,12) ])
+            sage: B = libgap.Stabilizer(H, list(S[1]), libgap.OnTuples);B
+            Group([ (1,2,3) ])
+            sage: T = PermutationGroup(gap_group=libgap.DirectProduct(A,B))
+            sage T.is_isomorphic(H)
```
mantepse: There is a `:` missing

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,83 @@ def smallest_moved_point(self):
...
+    @cached_method
+    def disjoint_direct_product_decomposition(self):
+        r"""
+        Returns the finest partition of the underlying set such that ``self``
+        is isomorphic to the direct product of the projections of ``self``
+        onto each part of the partition. Each part is a union of orbits
+        of ``self``.
+
+        The algorithm is from [CJ2022]_, which runs in time polynomial in
+        `n \cdot |G|`, where `n` is the degree of the group and `|G|` is
```
mantepse: `G` usually denotes a group and `|G|` usually denotes the size of the group, so I'd suggest to use another letter, for example, `X` for the generating set.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,83 @@ def smallest_moved_point(self):
...
+            sage: B = libgap.Stabilizer(H, list(S[1]), libgap.OnTuples);B
+            Group([ (1,2,3) ])
+            sage: T = PermutationGroup(gap_group=libgap.DirectProduct(A,B))
+            sage: T.is_isomorphic(H)
+            True
+            sage: PermutationGroup(PermutationGroup(gap_group=A).gens(),domain=list(S[1])).disjoint_direct_product_decomposition()
+            {{4, 5, 6, 7, 8, 9, 10, 11, 12}}
+            sage: PermutationGroup(PermutationGroup(gap_group=B).gens(),domain=list(S[0])).disjoint_direct_product_decomposition()
+            {{1, 2, 3}}
+        """
```
mantepse, line 1628:
```suggestion
            sage: [sum(1 for G in SymmetricGroup(n).conjugacy_classes_subgroups() if len(G.disjoint_direct_product_decomposition()) == 1) for n in range(1,8)]
            [1, 1, 2, 6, 6, 27, 20]
            sage: oeis(_) # optional -- internet
            0: A005226: Number of atomic species of degree n; also number of connected permutation groups of degree n.
        """
```

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,93 @@ def smallest_moved_point(self):
...
+            {{4, 5, 6, 7, 8, 9, 10, 11, 12}}
+            sage: PermutationGroup(PermutationGroup(gap_group=B).gens(),domain=list(S[0])).disjoint_direct_product_decomposition()
+            {{1, 2, 3}}
+
+        Counting the number of connected subgroups::
+
+            sage: # optional -- internet
+            sage: seq = [sum(1 for G in SymmetricGroup(n).conjugacy_classes_subgroups() if len(G.disjoint_direct_product_decomposition()) == 1) for n in range(1,8)];seq
+            [1, 1, 2, 6, 6, 27, 20]
+            sage: oeis(seq)
```
mantepse:
```suggestion
            sage: oeis(seq)  # optional -- internet
```

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,93 @@ def smallest_moved_point(self):
...
+        The algorithm is from [CJ2022]_, which runs in time polynomial in
+        `n \cdot |X|`, where `n` is the degree of the group and `|X|` is
+        the size of a generating set, see Theorem 4.5.
+
+        EXAMPLES::
+
+        The example from the original paper::
+
+            sage: H = PermutationGroup([[(1,2,3),(7,9,8),(10,12,11)],[(4,5,6),(7,8,9),(10,11,12)],[(5,6),(8,9),(11,12)],[(7,8,9),(10,11,12)]])
+            sage: S = H.disjoint_direct_product_decomposition();S
```
mantepse: Here, and below, there should be a space after `;`.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,93 @@ def smallest_moved_point(self):
...
+            sage: T.is_isomorphic(H)
+            True
+            sage: PermutationGroup(PermutationGroup(gap_group=A).gens(),domain=list(S[1])).disjoint_direct_product_decomposition()
+            {{4, 5, 6, 7, 8, 9, 10, 11, 12}}
+            sage: PermutationGroup(PermutationGroup(gap_group=B).gens(),domain=list(S[0])).disjoint_direct_product_decomposition()
+            {{1, 2, 3}}
+
+        Counting the number of connected subgroups::
+
+            sage: # optional -- internet
```
mantepse: Oh, sorry, I only saw this now.  I think it's better to mark only the `oeis` line as `optional -- internet`, because the numbers can be computed, and are instructive.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,93 @@ def smallest_moved_point(self):
...
+        Returns the finest partition of the underlying set such that ``self``
+        is isomorphic to the direct product of the projections of ``self``
+        onto each part of the partition. Each part is a union of orbits
+        of ``self``.
+
+        The algorithm is from [CJ2022]_, which runs in time polynomial in
+        `n \cdot |X|`, where `n` is the degree of the group and `|X|` is
+        the size of a generating set, see Theorem 4.5.
+
+        EXAMPLES::
```
mantepse: There should only be a single `:` here, because what follows is not code.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,93 @@ def smallest_moved_point(self):
...
+                    libgap.Add(Xp, x)
+            X = Xp
+        final_partition = DisjointSet(self.domain())
+        for part in P:
+            grp = [self._domain_from_gap[Integer(x)]
+                   for i in part
+                   for x in O[i]]
+            for i in range(1, len(grp)):
+                final_partition.union(grp[0], grp[i])
+        return SetPartition(final_partition)
```
mantepse: Possibly (but I'm not sure!)
```suggestion
        return SetPartition([[self._domain_from_gap[Integer(x)]
                              for i in part
                              for x in O[i]]
                             for part in P]
                            + [[x] for x in self.fixed_points()])
```
is easier to understand.

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,89 @@ def smallest_moved_point(self):
...
+            0: A005226: Number of atomic species of degree n; also number of connected permutation groups of degree n.
+        """
+        from sage.combinat.set_partition import SetPartition
+        from sage.sets.disjoint_set import DisjointSet
+        H = self._libgap_()
+        if self.is_trivial():
+            return SetPartition(DisjointSet(self.domain()))
+        if libgap.NrMovedPoints(H) == self.degree() and libgap.IsTransitive(H):
+            return SetPartition([self.domain()])
+        O = libgap.Orbits(H)
```
ChrisJefferson: The order of 'orbits' isn't fixed, and the order inside each orbit, isn't fixed, so if you'd like to make the result of the algorithm more consistent, you could sort the members of O, and the sort O itself (the time taken to do this will be fairly trivial compared to the rest of the algorithm).

src/sage/groups/perm_gps/permgroup.py
```diff
@@ -1581,6 +1581,89 @@ def smallest_moved_point(self):
         p = self._libgap_().SmallestMovedPoint()
         return self._domain_from_gap[Integer(p)]
 
+    @cached_method
+    def disjoint_direct_product_decomposition(self):
+        r"""
+        Returns the finest partition of the underlying set such that ``self``
```
fchapoton: Return (imperative mode) and not Returns

Newtech66:
Why doesn't the method appear in the table of methods on the Permutation Groups page in the doc preview?

mantepse:
That seems to be a problem of `gen_rest_table_index` with decorators.  `one`,  `smallest_moved_point`, `orbits`, ... don't appear either.

mantepse (approved):
Looks good to me!

dimpase (approved):

Issue #37305, opened 2024-02-12
Implementing cryptographic schemes in `sage.crypto`


grhkm21:
### Problem Description

As it is right now, `sage.crypto` does not have too many advanced features especially outside symmetric cryptanalysis. It would be great to improve on that.

### Proposed Solution

I propose to include implementation of popular and basic cryptography primitives/schemes in that directory. It can serve as a reference implementation, and also help cryptographers prototype new advanced protocols easier.

For basic primitives, there are two directions. One direction is the "structural" side, for example creating classes for encryption schemes, OT, commitment schemes and others. The other direction is the "implementation" side, for example implementing a basic scheme using those classes.

Take encryption schemes as a concrete example. A class for a public key encryption scheme would have methods like `.generate_key`, `.encrypt(msg, pk)`, `.decrypt(msg, sk)`. An implementation can simply be wrapping around AES.

What do everyone think? Any strong opinion on whether to include this or not?

vincentmacri:
I don't have an opinion on how this should be implemented in Sage, but as someone who has used Sage to prototype new cryptographic schemes before, I do think this would be nice to have.

I think it would be nice if Sage implemented "toy" implementations (i.e. minimal effort put towards things like side-channel resistance, and not intended for real-world encryption use) of well-known cryptographic schemes (RSA, ECDH, and some of the new post-quantum schemes that are being standardized) to provide something to compare against for correctness if someone is trying to implement some well-known scheme with some improved algorithm.

vincentmacri:
Closing this open-ended issue in favour of the Discussions thread #41218.

PR #38374, opened 2024-07-15
Add framework for key exchange schemes and Diffie-Hellman

### LINKED ISSUE #11568: Diffie-Hellman key exchange
<div id="comment:0"></div>

 Diffie-Hellman was the first public-key algorithm ever invented, way back in 1976 . It gets its security from the difficulty of calculating discrete logarithms in a finite field, as compared with the ease of calculating exponentiation in the same field. Diffie-Hellman can be used for key distribution—Alice and Bob can use this algorithm to generate a secret key—but it cannot be used to encrypt and decrypt messages.

CC:  nguyenminh2@gmail.com

Component: **cryptography**

Keywords: **public key exchange, Diffie-Hellman**

Author: **Tevin Joseph K.O.**

_Issue created by migration from https://trac.sagemath.org/ticket/11568_

vincentmacri:
Motivated by (but does not yet close) #37305. Closes #11568.

This PR adds a basic framework to add public key exchange schemes to Sage, and includes an implementation of the Diffie-Hellman primarily as an example of this new framework.

Open to suggestions to improve the structure of the code. This code was based on the existing code for public-key encryption in Sage.

### :memo: Checklist

### :hourglass: Dependencies

None


vincentmacri:
@grhkm21 Thoughts on this?

src/sage/crypto/key_exchange/pke.py
```diff
@@ -0,0 +1,41 @@
...
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#                  https://www.gnu.org/licenses/
+# ****************************************************************************
+
+from sage.misc.superseded import experimental
+
+class KeyExchangeScheme:
+
+    @experimental(37305)
```
vincentmacri: This is marked as experimental because if this gets merged I'd like to add more cryptographic schemes to Sage in the near(ish) future as I have spare time. I don't want to be tied down yet to any particular way of structuring the code.

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,214 @@
...
+from sage.rings.finite_rings.finite_field_constructor import GF
+from sage.rings.finite_rings.finite_field_prime_modn import FiniteField_prime_modn_with_category
+from sage.rings.finite_rings.integer_mod import IntegerMod_int
+from sage.structure.sage_object import SageObject
+
+from typing import Union
+
+class DiffieHellman(KeyExchangeScheme):
+
+    @experimental(37305)
```
vincentmacri, line 38: See comment on `KeyExchangeScheme` on why this is marked experimental.

src/doc/en/reference/references/index.rst
```diff
@@ -5343,6 +5343,11 @@ REFERENCES:
              Mathematical Physics, Analysis and Geometry 1, 171-191 (1998).
              :doi:`10.1023/A:1009724323513`
 
+.. [PP2010] \C. Paar and J. Pelzl.
+            *Understanding Cryptography*.
```
vincentmacri: I wasn't sure whether or not to include the subtitle which is "A Textbook for Students and Practitioners".

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+    def __init__(self, p: Integer, g: Union[Integer, IntegerMod_int], proof: bool = True):
+        """
+        Create an instance of the Diffie-Hellman key exchange scheme using the
+        given prime ``p`` and base ``g``.
+
+        INPUT:
+
+        - ``p`` -- prime integer defining the field `\\GF{p}` that the key exchanges
```
JosePisco: Should you specify the type of your parameters ? See the [Docstring INPUT block](https://doc.sagemath.org/html/en/developer/coding_basics.html#the-docstring-of-a-function-content) from the developer guide.

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
+r"""
+Diffie-Hellman Public Key Exchange Scheme
+
+This module contains a toy implementation of the Diffie-Hellman public key exchange
+scheme.
```
JosePisco: You should format your code to fit into 80 cols.
```suggestion
This module contains a toy implementation of the Diffie-Hellman public key
exchange scheme.
```

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+
+from sage.misc.superseded import experimental
+
+from sage.crypto.key_exchange.key_exchange import KeyExchangeScheme
+
+from sage.arith.misc import is_prime
+from sage.misc.prandom import randint
+from sage.rings.integer import Integer
+from sage.rings.finite_rings.finite_field_constructor import GF
+from sage.rings.finite_rings.finite_field_prime_modn import FiniteField_prime_modn
```
JosePisco: 80 cols format, though I don't know the style for line carriage imports.
```suggestion
from sage.rings.finite_rings.finite_field_prime_modn import \
    FiniteField_prime_modn
```

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+from sage.rings.finite_rings.integer_mod import IntegerMod_int
+from sage.structure.proof.proof import WithProof
+
+from typing import Union
+
+class DiffieHellman(KeyExchangeScheme):
+
+    @experimental(37305)
+    def __init__(self, p: Integer, g: Union[Integer, IntegerMod_int], proof: bool = True):
```
JosePisco: 80 cols format & default values for function parameters should stick to the `=`. We should check if the style says to indent with 4 spaces or to align with the first parameter.
```suggestion
    def __init__(self, p: Integer, g: Union[Integer, IntegerMod_int], proof:
        bool=True):
```

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        Create an instance of the Diffie-Hellman key exchange scheme using the
+        given prime ``p`` and base ``g``.
+
+        INPUT:
+
+        - ``p`` -- prime integer defining the field `\\GF{p}` that the key exchanges
+          will be performed over, must be at least 5
```
JosePisco: 80 cols format
```suggestion
        - ``p`` -- prime integer defining the field `\\GF{p}` that the key
          exchanges will be performed over, must be at least 5
```

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+
+        - ``g`` -- base for the key exchange, (coerceable to) an element of `\\GF{p}` from `2` to `p - 2`
```
JosePisco: 80 cols format
```suggestion
        - ``g`` -- base for the key exchange, (coerceable to) an element of
          `\\GF{p}` from `2` to `p - 2`
```

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+            self._field = GF(p, impl='modn')
+        else:
+            with WithProof('arithmetic', False):
+                self._field = GF(p, impl='modn')
+
+        self._p = p
+        self._g = self._field(g)
+
+        # While these values won't cause mathematical problems, they do completely
+        # break the security of the Diffie-Hellman scheme.
```
JosePisco: 80 cols format
```suggestion
        # While these values won't cause mathematical problems, they do
        # completely break the security of the Diffie-Hellman scheme.
```

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+        Get the string representation of the Diffie-Hellman instance.
+
+        TESTS::
+
+            sage: from sage.crypto.key_exchange.diffie_hellman import DiffieHellman
+            sage: DH = DiffieHellman(7, 3)
+            sage: DH
+            Diffie-Hellman key exchange over Finite Field of size 7 with generator 3
+        """
+        return f'Diffie-Hellman key exchange over {self._field} with generator {self._g}'
```
JosePisco: I think you can bypass the 80 cols format using something like
```suggestion
        return f'Diffie-Hellman key exchange over {self._field} with ' \
               f'generator {self._g}'
```
But again I don't know where the carriage return should be indented here.

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,314 @@
...
+        Get the LaTeX representation of the Diffie-Hellman instance.
+
+        TESTS::
+
+            sage: from sage.crypto.key_exchange.diffie_hellman import DiffieHellman
+            sage: DH = DiffieHellman(7, 3)
+            sage: latex(DH)
+            \text{Diffie-Hellman key exchange over }\Bold{F}_{7}\text{ with generator }3
+        """
+        return f'\\text{{Diffie-Hellman key exchange over }}{self._field._latex_()}\\text{{ with generator }}{self._g}'
```
JosePisco: Same here for 80 cols format
```suggestion
        return f'\\text{{Diffie-Hellman key exchange over }}' \
               f'{self._field._latex_()}\\text{{ with generator }}{self._g}'
```

src/sage/crypto/key_exchange/key_exchange.py
```diff
@@ -0,0 +1,42 @@
+r"""
+Public key exchange schemes.
+
+This module contains base classes for public key exchange schemes. The classes defined in
+this module should not be called directly. It is the responsibility of child classes to
+implement specific cryptosystems.
```
JosePisco: 80 cols format
```suggestion
This module contains base classes for public key exchange schemes. The classes
defined in this module should not be called directly. It is the responsibility
of child classes to implement specific cryptosystems.
```

JosePisco (changes requested):
Code looks good to me overall. Just being picky with style issues. We should check if docstring `EXAMPLES` and `TESTS` should be formatted and if so, how to do it properly.

Nice contribution, it's good to see old issues being solved ^^'

src/sage/crypto/key_exchange/diffie_hellman.py
vincentmacri: Lots of code in Sage doesn't follow that (at least not in the format specified there). Since the type annotations show up in the documentation I figured that specifying types in the docstring would be unnecessary duplication. That part of the developer guide doesn't say anything about type annotation despite elsewhere in the developer guide saying that we should use type annotations when possible, so I had the impression that that specific part might be outdated.

grhkm21:
I can review later. I just got back from vacation.

src/sage/crypto/key_exchange/diffie_hellman.py
grhkm21: Personally I don't care.

vincentmacri:
> We should check if docstring `EXAMPLES` and `TESTS` should be formatted and if so, how to do it properly.

I can't find anything specific. Since the docstring `EXAMPLES` go into the documentation, I think it's most important that they're readable there.

I'm guessing you raised this because some of the examples are over 80 characters in the source code. I think most instances of this are because of the indenting before an example is already 12 characters. I don't think splitting `from sage.crypto.key_exchange.diffie_hellman import DiffieHellman` into two lines in the examples will help readability. I'm not sure there's much to be done for the tests where the return values are over 80 characters. I assume this is why the linter doesn't enforce the 80 character rule.

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,338 @@
...
+    FiniteField_prime_modn
+from sage.rings.finite_rings.integer_mod import IntegerMod_abstract
+from sage.structure.proof.proof import WithProof
+
+from typing import Union
+
+class DiffieHellman(KeyExchangeScheme):
+
+    @experimental(37305)
+    def __init__(self, p: Integer, g: Union[Integer, IntegerMod_abstract],
```
vincentmacri, line 39: The conditions on `g` are actually looser than the type annotation implies.

For instance, `DiffieHellman(7, 1/2)` is allowed because 2 has an inverse in GF(7). I'm not sure what the best way to deal with this is. I expect most users using this class will either pass in an integer for g or *maybe* an element of GF(p), which is why I only included these in the annotation. I'm not sure if there is an easy way to do an annotation for "anything that is coercable to an element of GF(p)". I mention in the docstring that `g` needs to be coercable to an element of `GF(p)`.

If people have thoughts on this I'm open to changing this.

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,338 @@
...
+        - ``proof`` -- (default: ``True``) whether to require a proof that
+          ``p`` is prime. If ``False``, a probabilistic test can be used for
+          checking that ``p`` is prime. This should be set to ``False``
+          when using large (cryptographic size) primes, otherwise checking
+          primality will take too long.
+
+        .. WARNING::
+
+            This is a toy implementation for educational use only! Do not use this
+            implementation, or any cryptographic features of Sage, in any setting where
```
JosePisco: More 80 cols format here

vincentmacri:
I'm not sure why the last build failed. A test is failing in a component that isn't touched by this PR for one test, and in another test Docker is failing to build.

src/sage/crypto/key_exchange/diffie_hellman.py
JosePisco: I don't think this problem inherits from how you constructed this piece of code and since this is a toy framework, I don't think this would be a problem.

If you really want to check this maybe you can try to play with

```py
if isinstance(g, Integer) or isinstance(g, IntegerMod_abstract):
    blah
```
but to me, this is ok.

JosePisco (approved):
> I can't find anything specific. Since the docstring EXAMPLES go into the documentation, I think it's most important that they're readable there.
>
> I'm guessing you raised this because some of the examples are over 80 characters in the source code. I think most instances of this are because of the indenting before an example is already 12 characters. I don't think splitting from sage.crypto.key_exchange.diffie_hellman import DiffieHellman into two lines in the examples will help readability. I'm not sure there's much to be done for the tests where the return values are over 80 characters. I assume this is why the linter doesn't enforce the 80 character rule.

Yes this would make sense and formatting examples would probably make them less readable for the next person.

The rest of the code looks really neat :+1:

vincentmacri:
> > I can't find anything specific. Since the docstring EXAMPLES go into the documentation, I think it's most important that they're readable there.
> > I'm guessing you raised this because some of the examples are over 80 characters in the source code. I think most instances of this are because of the indenting before an example is already 12 characters. I don't think splitting from sage.crypto.key_exchange.diffie_hellman import DiffieHellman into two lines in the examples will help readability. I'm not sure there's much to be done for the tests where the return values are over 80 characters. I assume this is why the linter doesn't enforce the 80 character rule.
> 
> Yes this would make sense and formatting examples would probably make them less readable for the next person.
> 
> The rest of the code looks really neat 👍

Thanks for the review!

The last commit was just changing "public key exchange" to "key exchange" which I realized is the more standard terminology.

src/sage/crypto/key_exchange/diffie_hellman.py
vincentmacri: I don't think I want to enforce the typing annotations, I'm happy with `DiffieHellman(7, 1/2)` being allowed. I guess one way to look at it is the type annotations tell you the types that the API *guarantees* to work, i.e. we could decide not to allow rationals later without a deprecation cycle (not that the deprecation rules apply now anyway since it's marked experimental).

Especially since this class is primarily for pedagogical purposes, I lean a lot more to "allow anything (or at least don't enforce any restrictions) that won't throw an error".

That's why I didn't do anything to prevent something like `DiffieHellman(7, GF(9)(14))`. That will work (and respects the type annotations). 14 will be reduced mod 3 to 2, and the `__init__` in DiffieHellman will convert 2 mod 3 to 2 mod 7. I don't know *why* a user would do this, but since the class is for primarily pedagogical use I don't want to stop a user from trying anyway to answer the question "what if I do this?".

vincentmacri:
@grhkm21 Sorry for the ping, but could I get a review on this?

src/sage/crypto/key_exchange/all.py
```diff
@@ -0,0 +1,4 @@
+from sage.misc.lazy_import import lazy_import
+
+lazy_import('sage.crypto.key_exchange.diffie_hellman', 'DiffieHellman')
+del lazy_import
```
grhkm21, line 6: Didn't think of this before, but seems like a good idea!

src/sage/crypto/key_exchange/diffie_hellman.py
grhkm21: Sounds *rational* to me (please laugh). I lean more towards not enforcing everything too, as long as it's reasonable it's okay.

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,341 @@
...
+        Create an instance of the Diffie-Hellman key exchange scheme using the
+        given prime ``p`` and base ``g``.
+
+        INPUT:
+
+        - ``p`` -- prime integer defining the field `\\GF{p}` that the key
+          exchanges will be performed over, must be at least 5
+
+        - ``g`` -- base for the key exchange, (coerceable to) an element of
+          `\\GF{p}` from `2` to `p - 2`
```
grhkm21:
```suggestion
        r"""
        Create an instance of the Diffie-Hellman key exchange scheme using the
        given prime ``p`` and base ``g``.

        INPUT:

        - ``p`` -- prime integer defining the field `\GF{p}` that the key
          exchanges will be performed over, must be at least 5

        - ``g`` -- base for the key exchange, (coerceable to) an element of
          `\GF{p}` from `2` to `p - 2`
```
I didn't know `\GF{p}` is a thing lol

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,341 @@
...
+
+        .. WARNING::
+
+            This is a toy implementation for educational use only! Do not use
+            this implementation, or any cryptographic features of Sage, in any
+            setting where security is needed!
+
+        REFERENCES:
+
+        For more information, see [PP2010]_, section 8.1.
```
grhkm21:
```suggestion
        For more information, see Section 8.1 of [PP2010]_.
```
Following other examples :D

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,341 @@
...
+        EXAMPLES::
+
+            sage: from sage.crypto.key_exchange.diffie_hellman import DiffieHellman
+            sage: DH = DiffieHellman(13, 2)
+            doctest:...: FutureWarning: This class/method/function is marked as experimental. It, its functionality or its interface might change without a formal deprecation.
+            See https://github.com/sagemath/sage/issues/37305 for details.
+
+        This is an example of a full key exchange using a cryptographically
+        large prime. This is the prime from the 8192-bit MODP group in
+        RFC3526::
```
grhkm21: Can you put a link for this? https://tools.ietf.org/html/rfc3526

src/sage/crypto/key_exchange/diffie_hellman.py
```diff
@@ -0,0 +1,341 @@
...
+    def __init__(self, p: Integer, g: Union[Integer, IntegerMod_abstract],
+                 proof: bool = True) -> None:
+        """
+        Create an instance of the Diffie-Hellman key exchange scheme using the
+        given prime ``p`` and base ``g``.
+
+        INPUT:
+
+        - ``p`` -- prime integer defining the field `\\GF{p}` that the key
+          exchanges will be performed over, must be at least 5
```
grhkm21: Why do we not allow DH over extension fields?

grhkm21 (changes requested):
comments. I have a few more comments (I think) but I have to test your code first, and my Sage is building now.

src/sage/crypto/key_exchange/diffie_hellman.py
vincentmacri: In the literature, Diffie-Hellman is always over a prime field. If we allowed non-prime fields, we'd be implementing a different scheme.

src/sage/crypto/key_exchange/diffie_hellman.py
vincentmacri: I think it's a thing Sage defines, like `\Bold`.

src/sage/crypto/key_exchange/all.py
vincentmacri: I just copied from the `crypto/public_key` folder for this.

I'm now thinking I may have done this wrong (and maybe `public_key` did too but I think that's out of scope for this PR, `public_key` needs some refactoring I think). I'm going to do some more testing of this.

vincentmacri: I was doing it wrong. After the commit I just pushed, `DiffieHellman` can now be accessed without the user needing to explicitly import it.

This does raise the question of how we want the schemes to be accessed by the user. Most people probably won't use this feature (although I guess that applies to most things in Sage) so maybe it makes sense to access it as something like `crypto.DiffieHellman`, to avoid polluting the auto-suggest feature in some interfaces, but several of the existing symmetric ciphers are just accessed directly so I've done the same for consistency.

Happy to change it to something like `crypto.DiffieHellman` or `key_exchange.DiffieHellman` if that's preferred though.

vincentmacri: Right now I have it so `DiffieHellman` is accessible without import, but `KeyExchangeScheme` requires an import. I don't think an abstract base class needs to be exposed to users without them explicitly importing it.

grhkm21: For my PR of a similar nature, I created a `catalog` file, see [here](https://github.com/sagemath/sage/pull/36581/files#diff-81cd27b8fea2497cd1c834dd59f615575a64134d79461fb383ae5337ff095970). (Suggested by Travis, another Sage contributor)

vincentmacri: Good idea, I've implemented this.

I still didn't include `KeyExchangeScheme` in the catalogue because it's an abstract base class, but the catalogue should be useful as more key exchange schemes are added in the future.

grhkm21: I think having `key_exchange` in global scope (I think?) is *a bit* questionable. Do you think it's a good idea?

vincentmacri: I think it's less questionable than `SubstitutionCryptosystem` being global scope (which it currently is).

I don't think it's too different than how `distributions` is global scope in the PR you linked. The alternative (which maybe is better) would be to access the key exchange schemes under `crypto.key_exchange`, so you'd do `crypto.key_exchange.DiffieHellman`. We could then in the future move the other cryptography code to `crypto.[something]` as well and take things like `SubstitutionCryptosystem` out of global scope.

After writing that I've convinced myself that it should be changed to `crypto.key_exchange`.

And the reason I didn't consider putting it under `crypto` directly (so `crypto.DiffieHellman`) is some schemes (like RSA) have multiple uses. RSA can be used for public key encryption, key exchange, and cryptographic signing. I think having something like `crypto.pke.RSA`, `crypto.key_exchange.RSA`, and `crypto.signing.RSA` would be better than `crypto.RSAEncryption`, `crypto.RSAKeyExchange`, and `crypto.RSASign`. Especially because for discoverability, a user is more likely to want to know "what key exchange schemes are in Sage" than "what RSA variants are in Sage". Thinking about the future and how many cryptography schemes (public key encryption, key exchange, signing, symmetric ciphers, etc.) could be implemented in Sage, putting it all under `crypto` would be too much.

grhkm21: I think `crypto.key_exchange.DiffieHellman` looks good (definitely not `crypto.DiffieHellman`), but again I don't have a strong opinion. 

> After writing that I've convinced myself that it should be changed to crypto.key_exchange.

Lol, it happens :)

grhkm21:
I'll look through it again once but I think it's okay. I can approve when you reply to the {crypto.}key_exchange.DiffieHellman scope thing

src/sage/crypto/key_exchange/all.py
vincentmacri: Okay, I realized that making it accessible under `crypto.key_exchange` would involve some major restructuring to the `crypto` module and I think that would be out of scope for this PR. Unless I'm forgetting some syntax, I don't think Python import statements let you do `from a.b.c import x as y.z`. Basically above I was advocating to change `from sage.crypto.all import *` to `import sage.crypto.all as crypto` in the `all.py` file in `src/sage` which currently only `stats` does (but `stats` still imports everything to global scope anyway).

For now I think it's fine to leave it as `key_exchange` in global scope. It's at least better than what we have for other stuff in the `crypto` module. But if you have other suggestions I'm open.

grhkm21: Okay if it's too much irrelevant stuff for this PR then it's okay to leave it (and optionally make a new issue that hopefully someone gets to.)

grhkm21 (approved):
LGTM. Looking forward to implementing some crypto protocols myself too! ~~Get people away from Magma~~ x)

grhkm21:
conda 3.11 CI is failing but clearly the errors are irrelevant.

vincentmacri:
Thanks for the review! By the way, do you know what needs to happen for the PR status to change from "needs review" to "positive review"?

> Looking forward to implementing some crypto protocols myself too!

To avoid duplicated work, I'll say that I'll probably be doing ECDH next, then CSIDH.

grhkm21:
@vincentmacri As long as there is positive review (and not disputed e.g. one positive one negative) then you or the reviewers can change it to positive review. I forgot to, thanks for the reminder.

Regarding the crypto, I have some CSIDH code lying around too if you want them.

vincentmacri:
> @vincentmacri As long as there is positive review (and not disputed e.g. one positive one negative) then you or the reviewers can change it to positive review. I forgot to, thanks for the reminder.

Thanks! I don't think I have the permission to set labels so I can't do it myself.
 
> Regarding the crypto, I have some CSIDH code lying around too if you want them.

Sure!

PR #38375, opened 2024-07-16
moving ideal_monoid to the category of rings

fchapoton:
Let us first try a simple thing, on the way to move all `ideal` stuff to the category framework.

This requires to enhance all doctests in `finite_dimensional_algebras` that use ideals to have associative ``self``. 

### :memo: Checklist


fchapoton:
maybe we should rather always declare that algebras made from matrices are associative ?

fchapoton:
I may have been over-optimistic in declaring these algebras in doctests as commutative..

fchapoton:
@tscrim , would you please  have time to look at that ? My plan for the sequel is to move most ideal stuff to categories and then deprecate the "Algebra" class, only used in one place right now.

src/sage/algebras/finite_dimensional_algebras/finite_dimensional_algebra_ideal.py
```diff
@@ -39,17 +41,21 @@ class FiniteDimensionalAlgebraIdeal(Ideal_generic):
 
     EXAMPLES::
 
+        sage: cat = CommutativeAlgebras(GF(3)).FiniteDimensional().WithBasis()
         sage: A = FiniteDimensionalAlgebra(GF(3), [Matrix([[1, 0], [0, 1]]),
-        ....:                                      Matrix([[0, 1], [0, 0]])])
+        ....:                                      Matrix([[0, 1], [0, 0]])],
+        ....:     category=cat)
```
tscrim:
```suggestion
        ....:                                      Matrix([[0, 1], [0, 0]])],
        ....:                               category=cat)
```
I think this alignment is more in the spirit of PEP8?

tscrim (approved):
LGTM modulo my one suggestion (repeated a few times). If you change or disagree, you can set a positive review.

I would like to remove the `IdealMonoid_nc` with better testing/usage of the element(i.e., ideal) class, but that would be for later if I have the energy and will for it.

fchapoton:
thanks ! alignment done as you wished ; though I am not an adept of pep8 for the documentation

tscrim:
Thank you. LGTM.

PR #38379, opened 2024-07-17
disconnect one bad file in pbori

fchapoton:
The file gbrefs in pbori seems bad : not documented, not used, not tested.

So I remove the only place where something is imported from this file, and add a note in the file.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/polynomial/pbori/__init__.py
+++ b/src/sage/rings/polynomial/pbori/__init__.py
@@ -39,5 +39,4 @@ from .nf import normal_form
 # Import some high-level modelling functionality
 from .blocks import declare_ring
 from .blocks import HigherOrderBlock, AlternatingBlock, Block
-from .gbrefs import load_file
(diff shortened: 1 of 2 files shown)


src/sage/rings/polynomial/pbori/__init__.py
```diff
@@ -39,5 +39,4 @@
 # Import some high-level modelling functionality
 from .blocks import declare_ring
 from .blocks import HigherOrderBlock, AlternatingBlock, Block
-from .gbrefs import load_file
```
mkoeppe, line 42: Can `load_file` be removed without deprecation?

fchapoton: It is not in the global namespace. One can certainly not exclude that it is used by some people importing from pbori maybe. Being not tested, it is probably very much broken anyway.

fchapoton:
closing as not so interesting

Issue #38384, opened 2024-07-17
Implement the recently computed dimension formula for a newspace with character


erick-ross:
### Problem Description

An explicit dimension formula was recently computed for the newspaces of modular forms with character. See https://arxiv.org/abs/2407.08881.
Implementation of this formula in Sage would greatly speed up `dimension_new_cusp_forms()`. Currently, Sage computes the newspace convolution term by term [here](https://github.com/sagemath/sage/blob/master/src/sage/modular/arithgroup/congroup_gamma1.py#L644)

### Proposed Solution

We can add in this formula as another possible `algorithm`  for `dimension_new_cusp_forms()` now. Perhaps later when the paper is formally published, this could be changed to be the default algorithm. 

### Alternatives Considered

NA

### Additional Information

I've already implemented this formula in Sage and tested it against Sage's current  `dimension_new_cusp_forms()`. I plan to submit a pull request shortly.

### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

PR #38385, opened 2024-07-17
Implemented an Explicit Dimension Formula for the Newspace with character

erick-ross:
This PR implements a [recently computed](https://arxiv.org/abs/2407.08881) explicit dimension formula for newspaces of modular forms with character.
This completes Issue #38384  
The tests verify this formula against the current dimension computation in Sage.

I added some additional documentation underneath `dimension_new_cusp_forms()` for this change. But I can't view the documentation preview, because for some reason, I can't get the docs to build on my machine. 

### :memo: Checklist

### DIFF
--- a/src/sage/modular/arithgroup/congroup_gamma1.py
+++ b/src/sage/modular/arithgroup/congroup_gamma1.py
@@ -645,8 +645,10 @@ class Gamma1_class(GammaH_class):
             return Gamma0(N).dimension_new_cusp_forms(k, p)
 
         if algorithm == "Ross":
-            if p != 0: raise ValueError("Algorithm 'Ross' not defined for p-new subspace")
-            if k < 2:  raise ValueError("Algorithm 'Ross' only defined for k >= 2")
+            if p != 0:
+                raise ValueError("Algorithm 'Ross' not defined for p-new subspace")
+            if k < 2:
+                raise ValueError("Algorithm 'Ross' only defined for k >= 2")
             return _dimension_new_cusp_forms_Ross(N, k, eps)
 
         from .congroup_gammaH import mumu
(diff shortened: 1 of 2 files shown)


src/sage/modular/arithgroup/congroup_gamma1.py
```diff
@@ -650,3 +657,246 @@ def dimension_new_cusp_forms(self, k=2, eps=None, p=0, algorithm="CohenOesterle"
...
         old = Gamma1_constructor(N//p).dimension_cusp_forms(k, eps_p, algorithm)
         return self.dimension_cusp_forms(k, eps, algorithm) - 2*old
+
+
+
+def _dimension_new_cusp_forms_Ross(N,k,chi):
+    r"""An internal function to implement the explicit dimension formula
+        given in Theorem 1.4 of https://arxiv.org/abs/2407.08881. For tests, 
+        see test_dim_formula_Ross() in sage/modular/arithgroup/tests.py  
+    """
```
tscrim:
```suggestion
    r"""
    Compute the dimension formula given in Theorem 1.4 of :arxiv:`2407.08881`.
    
    For more tests, see :meth:`sage.modular.arithgroup.tests.Test.test_dim_formula_Ross()`.
    """
```
However this function does need at least one doctest of its own.

src/sage/modular/arithgroup/congroup_gamma1.py
```diff
@@ -650,3 +657,246 @@ def dimension_new_cusp_forms(self, k=2, eps=None, p=0, algorithm="CohenOesterle"
         eps_p = eps.restrict(N//p)
         old = Gamma1_constructor(N//p).dimension_cusp_forms(k, eps_p, algorithm)
         return self.dimension_cusp_forms(k, eps, algorithm) - 2*old
+
+
+
```
tscrim:
```suggestion
        return self.dimension_cusp_forms(k, eps, algorithm) - 2*old


```
One too many blank lines.

src/sage/modular/arithgroup/tests.py
```diff
@@ -18,6 +18,8 @@
 from .arithgroup_perm import ArithmeticSubgroup_Permutation, EvenArithmeticSubgroup_Permutation, OddArithmeticSubgroup_Permutation
 from sage.modular.arithgroup.all import Gamma, Gamma0, Gamma1, GammaH
 from sage.rings.finite_rings.integer_mod_ring import Zmod
+from sage.modular.dirichlet import DirichletGroup
+from sage.rings.integer_ring import ZZ
```
tscrim:
```suggestion
```
The `ZZ` import is not needed and we can make the other import local.

src/sage/modular/arithgroup/tests.py
```diff
@@ -415,3 +420,23 @@ def test_todd_coxeter(self):
...
+        r"""
+        Test the explicit dimension formula implemented in 
+        Gamma1._dimension_new_cusp_forms_Ross()
+
+        EXAMPLES::
+
+            sage: from sage.modular.arithgroup.tests import Test
+            sage: Test().test_dim_formula_Ross(100,16)
+        """
+        for N in range(3,N_ub+1):
```
tscrim:
```suggestion
        """
        from sage.modular.dirichlet import DirichletGroup
        for N in range(3,N_ub+1):
```

src/sage/modular/arithgroup/tests.py
```diff
@@ -415,3 +420,23 @@ def test_todd_coxeter(self):
             for j in range(i + 1, len(reps)):
                 assert reps[i] * ~reps[j] not in G
                 assert reps[j] * ~reps[i] not in G
+
+
+    def test_dim_formula_Ross(self, N_ub=50, k_ub=10):
+        r"""
+        Test the explicit dimension formula implemented in 
+        Gamma1._dimension_new_cusp_forms_Ross()
```
tscrim:
```suggestion
        Test the dimension formula given in Theorem 1.4
        of :arxiv:`2407.08881`
```

src/sage/modular/arithgroup/tests.py
```diff
@@ -415,3 +420,23 @@ def test_todd_coxeter(self):
             for j in range(i + 1, len(reps)):
                 assert reps[i] * ~reps[j] not in G
                 assert reps[j] * ~reps[i] not in G
+
+
```
tscrim:
```suggestion

```

src/sage/modular/arithgroup/congroup_gamma1.py
```diff
@@ -650,3 +657,246 @@ def dimension_new_cusp_forms(self, k=2, eps=None, p=0, algorithm="CohenOesterle"
...
+    ret += -1 * c3(k) * rho(f, chi, f) * beta_rho_f(N//f, f)
+    ret += -1 * c4(k) * rhopm(f, chi, f) * beta_rhopm_f(N//f, f)
+    ret += ZZ(-1)/2 * 2**omega(f) * beta_sigma_f(N//f, f) 
+    ret += c0(k,f) * moebius(N//f)
+    assert ret.is_integral()
+    return ZZ(ret)
+
+
+
+
```
tscrim:
```suggestion
    return ZZ(ret)
```

tscrim (changes requested):
Thank you for your contribution. @JohnCremona, perhaps you might have some thoughts on this PR?

For me, I would prefer not to have all of those internal functions as they make the code harder to follow (i..e, take a more iterative style approach instead of functional) as most of those are called in exactly one place. You are also going to get a lot of Python `int`s being mixed in instead of Sage `Integer`s too.

JohnCremona:
I will take a look -- saw the preprint but have not yet read it as I'm at ANTS this week and then on holdiday for a few days.

eross156:
Thank you for all your feedback and updates. Especially thank you for all the style updates to match this repo; I'm still trying to figure out all the code standards here. 
I'm happy to make all the update suggestions.
Is that something that I should do the "commit suggestions" thing in GitHub. Or do I need to allow a maintainer to make those changes?

tscrim:
Sorry for the delay in responding (traveling).

Yes, usually you use `ZZ(123)`, which can run into problems if you end up dividing two `int`s.

I don't think it is so bad to get rid of all internal functions, but I can see some benefits to keeping some of them. However, the more trivial ones would be better to absorb as inline: `psi_local`, `psi`, `beta_psi_f`, `beta_sigma_f`, `rho`, `beta_rho_f`, `rhopm`, `beta_rhopm_f`, `c3`, `c4`, `c0`, `omega`. Many of these can be written as a `prod` and you could avoid refactoring `chi.conductor()` in a few places. (Also, it is better to use `x -= y` instead of `x = -1 * y` as a micro-optimization.) You should also always return a Sage `Integer` (in particular, line 896 to `return ZZ.zero()`).

eross156:
Ok, thanks for the feedback. I will plan to add in these changes on Monday or Tuesday (I'm currently travelling as well).

erick-ross:
This PR is two years old, and now hopelessly behind sagemath. 
So I think it is better that I just start a new PR from scratch.

PR #38396, opened 2024-07-20
Newton to poly function for enhancement #38393

### LINKED ISSUE #38393: Recovering Newton map from Dynamical system
### Problem Description

While the existing function is_newton has the ability to the return the necessary conjugation to have the dynamical system in the right form. We think that It would be better to have a dedicated function that can do more than would be appropriate for a simple boolean check function such as is_newton.

### Proposed Solution

the function is_newton will be deprecated so that calling return conjugation true, will return an error message.

Newton_to_poly(self, return_conjugation = false, check_newton = false)

Input: Self, a newton map, Return conjugation, will give a matrix that conjugates the original dynamic system, check_newton errors if map isnt newton, this will call is_newton

Output: Polynomial, and matrix if desired

### Alternatives Considered

No function returns the a polynomial corresponding to a newton map so a function is necessary

### Additional Information

_No response_

Nathabolin:
Fixes #38393 

### :memo: Checklist


bhutz:
Some comments here
=
 - deprecate the return_conjugation in is_newton
 - looks like QQbar should work for is_newton
 - *Newton* -> a Newton map
 - look at some of the example spacing: ` = ` and around input `, `
 - you should use `keyword=True` in examples
 - don't need to extra variable V. Can just say; if self.is_newton():

- This is wrong. Probably need is_newton to be true by default and note that if you set it to be false, you can get wrong answers.
```
P.<x,y>=ProjectiveSpace(QQbar,1)
f=DynamicalSystem([x^4,y^4])
M,h=f.Newton_to_poly(return_conjugation=True)
M,h
```
- This should probably work too. Take a look at the field of definition of the returned conjugation. Note that replacing QQbar by a numberfield fails with the same error.
```
A.<z>=AffineSpace(QQbar,1)
F = z^3 + z+ 1
f = DynamicalSystem_affine(z - F/(F.derivative(z)))
print(f)
g=f.homogenize(1)
m=matrix(QQ,2,2,[3,4,1,2])
g=g.conjugate(m)
M,h=g.Newton_to_poly(return_conjugation=True)
M,h
g.conjugate(M)
```

bhutz:
A couple comments on this

- still need to return the conjugation until we're done deprecating
- note the failing doc tests that needs to be fixed due to the deprecation
- looking at the code I see two options for the field of definitions
  - keep M, self, and poly in their minimal fields of definition and return both embeddings to their compositum
  - move M, poly to the compositum and return the embedding that moves self to the compositum
I think the second option is probably a simpler experience for the users, so is my preference.

PR #38405, opened 2024-07-22
Call more general algorithm when lattice basis isn't trivial

### LINKED ISSUE #38400: Discrete Gaussian sampling over lattices is skewed
### Steps To Reproduce

Consider this code

```python
from sage.stats.distributions.discrete_gaussian_lattice import DiscreteGaussianDistributionLatticeSampler
D = DiscreteGaussianDistributionLatticeSampler(B = matrix([[1, 1],[0, 2]]), sigma=20)
S = [D() for _ in range(2^12)]
l = [vector(v.list() + [S.count(v)]) for v in set(S)]
plt = list_plot(l, color='black', marker='o', size =100)
plt.show()
```

### Expected Behavior

It should show a nice spherical distribution.

### Actual Behavior

But it shows a skewed distribution.

### Additional Information

The issue is this:

```python
    def __call__(self):
        if not self.is_spherical: 
            v = self._call_non_spherical()
        elif self._c_in_lattice: # <- here
            v = self._call_in_lattice()
        else:
            v = self._call()
        v.set_immutable()
        return v

    def _call_in_lattice(self):
        w = self.VS([d() for d in self.D], check=False) # <- here
        return w * self.B + self._c
```

It is not right to output `w*B` whenever `c in L` because that ignores the skewedness of the basis. Instead, `_call_in_lattice` should be called `_call_trivial` or so and the check in call should check if the basis happens to be the identity.

### Environment

```markdown
- **OS**:
- **Sage Version**:
```

### Checklist

malb:
fixes #38400


src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -314,13 +316,13 @@ def f_or_hat(x):
...
         if self.B.nrows() != self.B.ncols():
-            raise NotImplementedError("basis must be a square matrix for now")
-
-        if self.is_spherical and not self._c_in_lattice:
-            raise NotImplementedError("lattice must contain 0 for now")
+            raise NotImplementedError("Basis must be a square matrix.")
 
         if self.B.base_ring() != ZZ:
-            raise NotImplementedError("lattice must be integral for now")
+            raise NotImplementedError("Lattice must be integral.")
```
mkoeppe: I think lowercase and no period is actually the correct style

malb: Fixed.

malb:
Linter complaint seems unrelated to the PR:

```
sage/rings/complex_interval.pyx:2392:1: E302 expected 2 blank lines, found 1
1       E302 expected 2 blank lines, found 1
```

mkoeppe:
Yes, that is fixed by CI Fix #38423.

src/sage/stats/distributions/discrete_gaussian_lattice.py
```diff
@@ -675,11 +678,20 @@ def __call__(self):
             sage: mean_L = sum(L) / len(L)          # long time
             sage: norm(mean_L.n() - D.c()) < 0.25   # long time
             True
+
+            sage: import numpy
+            sage: M = matrix(ZZ, [[1,2],[0,1]])
+            sage: D = distributions.DiscreteGaussianDistributionLatticeSampler(M, 20.0)
+            sage: L = [D() for _ in range(2^12)]                                               # long time
+            sage: div = (numpy.mean([abs(x) for x,y in L])/numpy.mean([abs(y) for x,y, in L])) # long time
```
mkoeppe:
```suggestion
            sage: div = numpy.mean([abs(x) for x,y in L])/numpy.mean([abs(y) for x,y in L])     # long time
```

mkoeppe (approved):
Otherwise LGTM

src/sage/stats/distributions/discrete_gaussian_lattice.py
malb: done :)

PR #38409, opened 2024-07-23
`xgcd` for multiple arguments

seewoo5:
`xgcd` for more than 2 elements. Implement as an another option for the existing `xgcd` function. If there's only one input `a` is given (and `b = None` as default), then returns `xgcd` result for the tuple/list `a`. It calls the 2-variable version of `xgcd` recursively. Also see [this](https://ask.sagemath.org/question/73642/xgcd-for-several-arguments/) post in the forum.

It would be great if anyone can suggest more complex testcases, e.g. rings other than integers or polynomials.

### :memo: Checklist


seewoo5:
Not strongly related to this PR, but there's a fully commented `XGCD_python` right below `XGCD`, which looks deprecated and could be deleted safely.

src/sage/arith/misc.py
```diff
@@ -1970,12 +1970,22 @@ def xgcd(a, b):
 
     INPUT:
 
+    One of the following:
+
     -  ``a, b`` -- integers or more generally, element of a ring for which the
        xgcd make sense (e.g. a field or univariate polynomials).
 
+    -  ``a`` -- list or tuple of at least two integers or more generally, element
```
grhkm21:
```suggestion
    -  ``a`` -- a list or tuple of at least two integers or more generally, elements
```

src/sage/arith/misc.py
```diff
@@ -1970,12 +1970,22 @@ def xgcd(a, b):
...
+       of a ring which the xgcd make sense.
+
     OUTPUT:
 
-    -  ``g, s, t`` -- such that `g = s\cdot a + t\cdot b`
+    One of the following:
+
+    -  ``g, s, t`` -- when two inputs ``a, b`` are given. They satisfy `g = s\cdot a + t\cdot b`.
+
+    -  ``r`` -- when only ``a`` is given (and ``b = None``). The tuple ``r`` has length one longer
```
grhkm21:
```suggestion
    -  ``r`` -- a tuple, when only ``a`` is given (and ``b = None``). It has length one longer
```
Might have to line wrap this.

src/sage/arith/misc.py
```diff
@@ -1970,12 +1970,22 @@ def xgcd(a, b):
...
+       than the length of ``a``, and its entries satisfy ``r[0] = \sum_{0 \le i < len(a)} r[i + 1] * a[i]``.
```
grhkm21:
```suggestion
       than the length of ``a``, and its entries satisfy `r_0 = \sum_{i = 0}^{len(a) - 1} r_{i + 1}a_i`.
```
Single backticks means latex, double backticks means code blocks. Here you probably want latex as a math equation.

src/sage/arith/misc.py
```diff
@@ -2045,18 +2065,34 @@ def xgcd(a, b):
...
+        except TypeError:
+            b = py_scalar_to_element(b)
         return a.xgcd(b)
-    except AttributeError:
-        a = py_scalar_to_element(a)
-        b = py_scalar_to_element(b)
-    except TypeError:
-        b = py_scalar_to_element(b)
-    return a.xgcd(b)
+    else:
```
grhkm21: This `else` is not necessary since the branch above returns. So remove this and unindent the rest below :D

src/sage/arith/misc.py
```diff
@@ -2045,18 +2065,34 @@ def xgcd(a, b):
...
+        # xgcd of >=2 elements
+        if not isinstance(a, (tuple, list)):
+            raise TypeError("Input `a` has to be a tuple or a list.")
```
grhkm21:
```suggestion
            raise TypeError("input `a` should be a tuple or a list")
```
Errors should start with a lowercase alphabet, according to one of the hundreds of PEP I am too lazy to find. Also no punctuation at the end (IIRC)

src/sage/arith/misc.py
```diff
@@ -2045,18 +2065,34 @@ def xgcd(a, b):
...
+        if len(a) < 2:
+            raise ValueError("At least two integers or elements should be given.")
```
grhkm21:
```suggestion
            raise ValueError("at least two elements should be given")
```

src/sage/arith/misc.py
```diff
@@ -2045,18 +2065,34 @@ def xgcd(a, b):
...
+        if len(a) == 2:
+            return xgcd(a[0], a[1])
+        else:  # Compute xgcd recursively
+            xgcd_t = xgcd(a[1:])  # xgcd_t[0] = sum(xgcd_t[i] * a[i], 1 \le i \le n-1)
```
grhkm21:
```suggestion
            # xgcd_t[0] = sum(xgcd_t[i] * a[i], 1 \le i \le n-1)
            xgcd_t = xgcd(a[1:])
```
I don't think it will be too difficult to implement this iteratively instead to avoid stack overflow. Do you mind doing that? If you follow the call trace then `xgcd([1,2,3,4,5])` calls `xgcd([2,3,4,5])` then ... then `xgcd([4,5])`, before any computation is done. So I think you can instead process `a` in reverse order in pairs and keep the same computation

grhkm21 (changes requested):
Thanks for contributing :)

grhkm21:
> Not strongly related to this PR, but there's a fully commented `XGCD_python` right below `XGCD`, which looks deprecated and could be deleted safely.

Feel free to!

src/sage/arith/misc.py
```diff
@@ -1955,7 +1955,7 @@ def xlcm(m, n):
     return (l, m, n)
 
 
-def xgcd(a, b):
+def xgcd(a, b=None):
     r"""
     Return a triple ``(g,s,t)`` such that `g = s\cdot a+t\cdot b = \gcd(a,b)`.
```
saraedum: Should this be updated?

src/sage/arith/misc.py
```diff
@@ -1970,12 +1970,22 @@ def xgcd(a, b):
```
saraedum, line 1958: What about the `NOTE` above? Is that still accurate?

src/sage/arith/misc.py
```diff
@@ -1970,12 +1970,22 @@ def xgcd(a, b):
...
+
     OUTPUT:
 
-    -  ``g, s, t`` -- such that `g = s\cdot a + t\cdot b`
+    One of the following:
+
+    -  ``g, s, t`` -- when two inputs ``a, b`` are given. They satisfy `g = s\cdot a + t\cdot b`.
+
+    -  ``r`` -- a tuple, when only ``a`` is given (and ``b = None``). It has length one longer
+       than the length of ``a``, and its entries satisfy `r_0 = \sum_{i = 0}^{len(a) - 1} r_{i + 1}a_i`.
```
saraedum: So, formally, all the `r_i` could be zero. I think you need to spell out somewhere that `r_0` is the gcd. Maybe at the top of the function.

src/sage/arith/misc.py
```diff
@@ -2045,18 +2069,34 @@ def xgcd(a, b):
...
-        b = py_scalar_to_element(b)
-    except TypeError:
-        b = py_scalar_to_element(b)
-    return a.xgcd(b)
+    else:
+        # xgcd of >=2 elements
+        if not isinstance(a, (tuple, list)):
+            raise TypeError("input `a` should be a tuple or a list")
+        if len(a) < 2:
+            raise ValueError("at least two elements should be given")
```
saraedum: Is there any reason not to allow the xgcd of a single element? Of course it would probably just be `a = 1×a` but it's maybe convenient in general algorithms.

Note that `gcd` allows that.

saraedum: Actually, gcd also allows an empty list. Not sure if that's reasonable here.

src/sage/arith/misc.py
seewoo5: Updated!

src/sage/arith/misc.py
seewoo5: I think so, I also checked locally without update and gives `TypeError: xgcd() missing 1 required positional argument: 'b'` without assigning default value.

src/sage/arith/misc.py
seewoo5: I didn't know that `gcd` allows single element, that's why I only allowed two or more elements. I'd update it too.

For empty list, it gives integer zero, so I can just add as returning `(0,)`, but I also don't know if this is the best or just allow at least one elements.

seewoo5: Of course I can add it for completeness, which looks good too.

src/sage/arith/misc.py
seewoo5: I'm not 100% sure about this, but I believe so. It would be great if someone can double check.

seewoo5:
Just curious, is it common to add random test cases? At least for `xgcd`, I think adding randomly generated tests worths to be added (e.g. 1000 integers, check if it gives the correct gcd and linear combination).

grhkm21:
IMO it's encouraged, but it's not too common right now afaik. Put it under a `TEST::` section.

src/sage/arith/misc.py
saraedum: `(0,)` sounds good to me. I mean, nobody is going to type this in the SageMath repl, but it might show up in an algorithm and it's a reasonable answer.

src/sage/arith/misc.py
saraedum: I mean, do you want to remove the reference to `a` and `b` and just say "the elements" (and rephrase the rest accordingly)? Because now b might be None and a a list.

src/sage/arith/misc.py
```diff
@@ -2045,18 +2070,33 @@ def xgcd(a, b):
...
+    # xgcd of >=2 elements
+    if not isinstance(a, (tuple, list)):
+        raise TypeError("input `a` should be a tuple or a list")
+    if len(a) < 2:
+        raise ValueError("at least two elements should be given")
+    # Compute xgcd recursively
+    res = xgcd(a[0], a[1])
+    for i in range(2, len(a)):
+        g, s, t = xgcd(res[0], a[i])
+        res = [g] + [s * c for c in res[1:]] + [t]
```
saraedum: Just nitpicking here now, but I believe that this can be simplified doing something like that.

```suggestion
    if len(a) == 0:
        return (ZZ(0),)
    a = Sequence(a, use_sage_types=True)
    res = [a.universe().zero()]
    for b in a:
        g, s, t = xgcd(res[0], b)
        res = [g] + [s * c for c in res[1:]] + [t]
```

saraedum: The `Sequence` is what `gcd` also does to make sure that everything lives in the same parent. Otherwise, I am not sure if you could end up in trouble if the elements live in completely different parents. (It's probably fine though, I haven't thought about it too hard.)

saraedum: (Feel free to do things like that or don't, it's just a suggestion.)

seewoo5: This looks better. Just for clarification, the return type would be still tuple, right?

src/sage/arith/misc.py
seewoo5: Oh yes you're complete right. I'll update.

src/sage/arith/misc.py
saraedum: I meant the documentation here, i.e., the first line.

src/sage/arith/misc.py
saraedum: Yes, tuple is the correct return type. The return type should be the same as it was before.

seewoo5: How should I type check `a`? (Or just do not, since `a = Sequence(...)` line eventually gives an error if input cannot be transformed into `Sequence`?)

saraedum: I wouldn't type check. In Python we usually go for duck-typing. Though it has some weird effects:
```
sage: gcd(2 + 2*I)
2
```

Then again, the lack of type checking means that you can put in an iterator:
```
sage: gcd(range(1000))
1
```

So, just being consistent with `gcd` is probably best.

src/sage/arith/misc.py
```diff
@@ -1955,27 +1956,41 @@ def xlcm(m, n):
     return (l, m, n)
 
 
-def xgcd(a, b):
+def xgcd(a, b=None):
     r"""
-    Return a triple ``(g,s,t)`` such that `g = s\cdot a+t\cdot b = \gcd(a,b)`.
+    When both ``a`` and ``b`` are given, then return a triple ``(g,s,t)``
```
grhkm21, line 1945:
```suggestion
    Return the greatest common divisor and the Bézout coefficients of the input arguments.

    When both ``a`` and ``b`` are given, then return a triple ``(g,s,t)``
```
The first line should be a one sentence summary of the algorithm, I believe.

src/sage/arith/misc.py
```diff
@@ -2044,19 +2077,45 @@ def xgcd(a, b):
...
-    return a.xgcd(b)
+
+    # xgcd for several elements (possibly more than one)
+    if len(a) == 0:
+        return (ZZ(0),)
+    a = Sequence(a, use_sage_types=True)
+    res = [a.universe().zero()]
+    for b in a:
+        g, s, t = xgcd(res[0], b)
+        res = [g] + [s * c for c in res[1:]] + [t]
```
grhkm21:
```suggestion
        res[0] = g
        for i in range(1, len(res)):
            res[i] *= s
        res.append(t)
```
Should be much faster :)

grhkm21 (changes requested):
changes

grhkm21 (approved):
Looks good to me. Any objections? @saraedum

grhkm21:
ah wait sorry, please fix linter error (CI)  @seewoo5 

image

seewoo5:
> ah wait sorry, please fix linter error (CI) @seewoo5
> 
> image

I guess this is for other files after merging development branch?
```
sage/rings/complex_interval.pyx:2392:1: E302 expected 2 blank lines, found 1
1       E302 expected 2 blank lines, found 1
```

grhkm21:
ah indeed ([here](https://github.com/sagemath/sage/pull/38423)), but also fix it here I guess.

seewoo5:
After #38423 is merged, I merged `develop` branch and there are new installation errors related to the versions - any guesses why?
Screenshot 2024-08-04 at 2 19 48 PM

PR #38414, opened 2024-07-23
Properly check dimension in curve constructor

kwankyu:
fix a bug reported in https://groups.google.com/g/sage-devel/c/qiUQEnRGGag

Along the way, we improve computation of dimension of ideals of multi-variate polynomial rings for principal ideals (this is needed to soothe elliptic curve code that complains of slow dimension computation).

### :memo: Checklist

### DIFF
--- a/src/doc/en/reference/references/index.rst
+++ b/src/doc/en/reference/references/index.rst
@@ -3008,7 +3008,10 @@ REFERENCES:
             Texts in Mathematics, Springer, 2001.
 
 .. [Gre2006] \R. M. Green, *Star reducible Coxeter groups*, Glasgow
-            Mathematical Journal, Volume 48, Issue 3, pp. 583-609.
+             Mathematical Journal, Volume 48, Issue 3, pp. 583-609.
+
+.. [Ger2008] Gert-Martin Greuel and Gerhard Pfister. A Singular introduction
+             to commutative algebra. Vol. 348. Berlin: Springer, 2008.
 
 .. [Gri2021] \O. Gritsenko, *On strongly regular graph with parameters (65; 32; 15; 16)*,
               :arxiv:`2102.05432`.
--- a/src/sage/rings/polynomial/multi_polynomial_ideal.py
+++ b/src/sage/rings/polynomial/multi_polynomial_ideal.py
@@ -1201,7 +1201,6 @@ class MPolynomialIdeal_singular_repr(
             sage: R.<x,y> = PolynomialRing(GF(2147483659^2), order='lex')
             sage: I = R.ideal(0)
             sage: I.dimension()
-            verbose 0 (...: multi_polynomial_ideal.py, dimension) Warning: falling back to very slow toy implementation.
             2
 
         ALGORITHM:
--- a/src/sage/rings/polynomial/multi_polynomial_ideal.py
+++ b/src/sage/rings/polynomial/multi_polynomial_ideal.py
@@ -1229,11 +1229,11 @@ class MPolynomialIdeal_singular_repr(
             if self.ngens() == 1:
                 g = self.gen(0)
                 if g.is_unit():
-                    self.__dimension = -1
+                    self.__dimension = Integer(-1)
                 elif g:
-                    self.__dimension = self.ring().ngens() - 1
+                    self.__dimension = Integer(self.ring().ngens() - 1)
                 else:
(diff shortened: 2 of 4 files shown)


nbruin:
Thanks for that correction!

A question: did the doctest you fixed already find this bug?

I'm wondering if the design is flawed: currently, we have `Curve_generic -> AffineCurve->AffineCurve_field`. The problem arises on `Curve_generic`: there the constructor just quietly does whatever the parent constructor does and yet `dimension` is overridden to a hardcoded `1`. Hence, `Curve_generic` seems flawed to me: it silently accepts non-curves and then overrides `dimension`. That sounds like a flawed design to me. I guess the problem is that `Curve_generic` doesn't have the means to check the validity of its input, but then it should perhaps be an abstract class (so that it cannot be instantiated) and/or it shouldn't override `dimension`: if we want to support curves over rings, then I'm not so sure they should be of dimension 1. Relative dimension 1 surely, but if you have a curve over `Spec(QQ[t])` then the resulting scheme has (absolute) dimension 2.

kwankyu:
> Thanks for that correction!
> 
> A question: did the doctest you fixed already find this bug?

Yes. But I will add a new test (from the report in sage-devel).

> I'm wondering if the design is flawed: currently, we have `Curve_generic -> AffineCurve->AffineCurve_field`. The problem arises on `Curve_generic`: there the constructor just quietly does whatever the parent constructor does and yet `dimension` is overridden to a hardcoded `1`. Hence, `Curve_generic` seems flawed to me: it silently accepts non-curves and then overrides `dimension`. That sounds like a flawed design to me. I guess the problem is that `Curve_generic` doesn't have the means to check the validity of its input, but then it should perhaps be an abstract class (so that it cannot be instantiated) and/or it shouldn't override `dimension`.

 `Curve_generic` is an "abstract class" in the sense that always its subclass is instantiated. As you say, the dimension check should happen in a subclass. `Curve_generic` does not override `dimension` since there is no `dimension` in the subclass. 

This prompts me a better way to fix the bug. Thanks. I will work more.

> if we want to support curves over rings, then I'm not so sure they should be of dimension 1. Relative dimension 1 surely, but if you have a curve over `Spec(QQ[t])` then the resulting scheme has (absolute) dimension 2.

For curves over rings, surely relative dimension should be used. Now we are dealing with curves over fields. So no difference.

nbruin:
> `Curve_generic` does not override `dimension` since there is no `dimension` in the subclass.

It does: https://github.com/kwankyu/sage/blob/b4ec7f65df42e5f7e1df24e468da136511c817fd/src/sage/schemes/curves/curve.py#L119

kwankyu:
> > `Curve_generic` does not override `dimension` since there is no `dimension` in the subclass.
> 
> or do you mean that on `AlgebraicScheme_subscheme` there is no method `dimension` so `Curve_generic` doesn't override a method but instead defines it for the first time? 

Yes.

> It's there, it's just raising "NotImplemented". So ... indeed not a very useful method, but it does make me think that Curve_generic should perhaps also no override it 

That is not relevant to our problem.

Anyway, removing `dimension` method from `Curve_generic` would fix the bug. But this is not to correct a design flaw. Objects of `Curve_generic` are curves, and curves have dimension 1 by definition. So there is nothing wrong with the method. It feels wrong to remove `dimension` just to fix the bug. 

> and leave it to subclasses that can actually test their input.

As a "better way", I thought of adding `dimension` method to the subclasses of `Curve_generic` so that `dimension` of `Curve_generic` is overridden. But this does not seem to be actually a better way, because if we use `dimension` in `__init__` (note that they are defined in the same class), we are making a circular dependency, at least conceptually.

Another way might be to reverse the order of superclasses in the class definition:
```sage
class AffineCurve_field(AlgebraicScheme_subscheme_affine_field, AffineCurve)
```
instead of
```sage
class AffineCurve_field(AffineCurve, AlgebraicScheme_subscheme_affine_field):  
```
But the original order of superclasses seems right: `AffineCurve` seems to be more concrete than `AlgebraicScheme_subscheme_affine_field`. I didn't try this method, but I guess there will be lots of unexpected doctest failures.

The current method to fix the bug seems to be the simplest.

kwankyu:
This seems to be the case where `super()` was introduced for. A little bit more intuitive now.

kwankyu:
anything else?

PR #38429, opened 2024-07-26
provide the fraction field of symmetric functions

mantepse:
We provide a naive implementation of the fraction field of symmetric functions.  In particular, this allows computations like
```
sage: s[1]/(1+s[1])
s[1]/(s[] + s[1])
```

### DIFF
--- a/src/sage/combinat/sf/sfa.py
+++ b/src/sage/combinat/sf/sfa.py
@@ -418,6 +418,15 @@ class SymmetricFunctionsBases(Category_realization_of_parent):
             return self.base_ring().is_integral_domain()
 
         def fraction_field(self):
+            r"""
+            Return the fraction field of ``self``.
+
+            EXAMPLES::
+
+                sage: s = SymmetricFunctions(QQ).s()
+                sage: s.fraction_field()
+                Fraction Field of Symmetric Functions over Rational Field in the Schur basis
+            """
             if not self.is_integral_domain():
                 raise TypeError("self must be an integral domain.")
             if hasattr(self, "__fraction_field") and self.__fraction_field is not None:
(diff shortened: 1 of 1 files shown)


src/sage/combinat/sf/sfa.py
```diff
@@ -417,6 +417,26 @@ def is_integral_domain(self, proof=True):
             """
             return self.base_ring().is_integral_domain()
 
+        def fraction_field(self):
```
tscrim, line 421:
```suggestion
        @cached_method
        def fraction_field(self):
```

src/sage/combinat/sf/sfa.py
```diff
@@ -417,6 +417,26 @@ def is_integral_domain(self, proof=True):
...
+            Return the fraction field of ``self``.
+
+            EXAMPLES::
+
+                sage: s = SymmetricFunctions(QQ).s()
+                sage: s.fraction_field()
+                Fraction Field of Symmetric Functions over Rational Field in the Schur basis
+            """
+            if not self.is_integral_domain():
+                raise TypeError("self must be an integral domain.")
```
tscrim:
```suggestion
                raise TypeError("self must be an integral domain")
```
to follow Python's conventions. (Should also be changed in the poly rings at some point.)

src/sage/combinat/sf/sfa.py
```diff
@@ -417,6 +417,26 @@ def is_integral_domain(self, proof=True):
...
+            if hasattr(self, "__fraction_field") and self.__fraction_field is not None:
+                return self.__fraction_field
+            else:
+                import sage.rings.fraction_field
+                K = sage.rings.fraction_field.FractionField_generic(self)
+                self.__fraction_field = K
+            return self.__fraction_field
```
tscrim:
```suggestion
            from sage.rings.fraction_field import FractionField_generic
            return FractionField_generic(self)
```
since we can use the `@cached_method` in the Python class.

tscrim (commented):
Thank you. A few minor things.

tscrim (approved):
LGTM. Once tests pass, you can set a positive review.

mantepse:
Thank you so much, Travis!

The failures are not reproducible on my computer, and should be unrelated anyway, therefore setting this positive review.

PR #36046, opened 2023-08-08
Fix sqrt(sqrt(2)) memory leak in ginac numeric.cpp

vbraun:
Fix the memory leak when computing the outer sqrt in `sqrt(sqrt(2))`, as reported on sage devel at https://groups.google.com/g/sage-devel/c/6zpxQKXtJgk

The unit test is a WIP and depends on https://github.com/tekknolagi/valgrind/pull/1

- Resolves #33074

PR #38432, opened 2024-07-27
fix memory leak in is_planar

dcoudert:
Fixes the memory leak in `is_planar` reported at https://ask.sagemath.org/question/78467/memory-leak-in-is_planar/

We now ensure that memory is released before returning the answer.

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/planarity.pyx
+++ b/src/sage/graphs/planarity.pyx
@@ -135,6 +135,7 @@ def is_planar(g, kuratowski=False, set_pos=False, set_embedding=False):
         elif status == NONEMBEDDABLE:
             # We now know that the graph is nonplanar.
             if not kuratowski:
+                gp_Free(&theGraph)
                 return False
             # With just the current edges, we have a nonplanar graph,
             # so to isolate a kuratowski subgraph, just keep going.
@@ -150,6 +151,7 @@ def is_planar(g, kuratowski=False, set_pos=False, set_embedding=False):
     if status == NONEMBEDDABLE:
         # Kuratowski subgraph isolator
         if not kuratowski:
+            gp_Free(&theGraph)
             return False
         g_dict = {}
         for i in range(1, theGraph.N + 1):


dcoudert:
@videlec I'm not sure how to show that this fixes the memory leak. At least when I check on my laptop the memory usage with htop, its not increasing anymore when running this very long example
```py
sage: count = 0
....: for g in graphs.nauty_geng('-C 10'):
....:     if g.is_planar():
....:         count = count+1
....: print(count)
```

user202729:
@dcoudert Maybe you can look into `tests/memcheck`? `#36046` is an example pull request that adds a test for memcheck.

dcoudert:
I found a way to check the memory consumption. However I'm not sure I can use that to make a doctest. At least we can see here that the memory leak is fixed with this PR.
```py
import psutil
from time import sleep

def get_process_memory():
    process = psutil.Process(os.getpid())
    mem_info = process.memory_info()
    return mem_info.rss

def my_count(n):
    leak = get_process_memory()
    count = 0
    for g in graphs.nauty_geng(f'-C {n}'):
        if g.is_planar():
            count += 1
    sleep(2)  # to ensure that subprocesses have released memory 
    leak = get_process_memory() - leak
    return count, leak
```

Before:
```py
sage: my_count(8)
(2893, 11714560)
sage: my_count(9)
(36496, 350650368)
```

With this PR:
```py
sage: my_count(8)
(2893, 0)
sage: my_count(9)
(36496, 0)
sage: my_count(10)
(545808, 0)
```

tscrim (approved):

tscrim:
There generally are not good ways to check against memleaks using doctests, even with #36046 IMO. Nor do I think we should add (doc)tests each time we encounter and fix a memleak, at least when it is straightforward. As such, I am setting this to a positive review.

dcoudert:
Thanks you.

PR #38418, opened 2024-07-24
add method is_vertex_cut to (di)graphs

dcoudert:
This PR answers a query from https://ask.sagemath.org/question/78391/does-sage-have-a-function-to-determine-vertex-cuts/

It adds to (di)graphs a method to check if a set of vertices is a vertex cut of the (di)graph, that is if its removal from the (di)graph increases the number of (strongly) connected components. This generalizes method `is_cut_vertex`.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation and checked the documentation preview.

PR #38435, opened 2024-07-27
add method `is_edge_cut` to graphs

dcoudert:
Similarly to #38418, this PR adds a method to check whether a set of edges forms an edge cut of the (di)graph, that is if the removal of these edges increases the number of (weakly) connected components.

### :memo: Checklist


src/sage/graphs/connectivity.pyx
```diff
@@ -23,7 +23,8 @@ Here is what the module can do:
     :meth:`connected_components_sizes` | Return the sizes of the connected components as a list.
     :meth:`blocks_and_cut_vertices` | Return the blocks and cut vertices of the graph.
     :meth:`blocks_and_cuts_tree` | Return the blocks-and-cuts tree of the graph.
-    :meth:`is_cut_edge` | Return ``True`` if the input edge is a cut-edge or a bridge.
+    :meth:`is_cut_edge` | Check whether the input edge is a cut-edge or a bridge.
+    :meth:`is_edge_cut` | Check whether ``edges`` is an edge cut of ``G``.
     :meth:`is_cut_vertex` | Check whether the input vertex is a cut-vertex.
```
kwankyu, line 28: For uniformity,

```
Check whether the input edges form an edge cut 
```
or
```
Check whether the set of edges is an edge cut 
```
or similar, omitting ``` `` ```

src/sage/graphs/connectivity.pyx
```diff
@@ -732,13 +734,173 @@ def blocks_and_cuts_tree(G):
...
+
+    This method is not working for (di)graphs with multiple edges. Furthermore,
+    edge labels are ignored.
+
+    INPUT:
+
+    - ``G`` -- a Sage (Di)Graph
+
+    - ``edges`` -- an iterable container of edges
+
```
kwankyu, line 753: whenever possible, just explain the inputs in plain mathematical terms.

"a (di)graph" instead of "a Sage (Di)Graph".

"a list of edges" instead of "an iterable container of edges".

src/sage/graphs/connectivity.pyx
```diff
@@ -732,13 +734,173 @@ def blocks_and_cuts_tree(G):
...
+        sage: G.allow_multiple_edges(True)
+        sage: is_edge_cut(G, [(0, 5), (2, 5)])
+        Traceback (most recent call last):
+        ...
+        ValueError: This method is not known to work on graphs with
+         multiedges. Perhaps this method can be updated to handle them, but in
+         the meantime if you want to use it please disallow multiedges using
+         allow_multiple_edges().
+
```
kwankyu, line 804: The message is too verbose. How about 
```
ValueError: This method does not work on graphs with multiedges
```
and put "if you want to use it please disallow multiedges using allow_multiple_edges()." in the explanation with some examples in "EXAMPLES:".

src/sage/graphs/connectivity.pyx
```diff
@@ -732,13 +734,173 @@ def blocks_and_cuts_tree(G):
...
+    TESTS:
+
+    If `G` is not a Sage graph, an error is raised::
+
+        sage: is_edge_cut('I am not a graph', [(0, 0)])
+        Traceback (most recent call last):
+        ...
+        TypeError: the input must be a Sage graph
+    """
```
kwankyu, line 832: this "test" should be removed. We don't need to test cases when a user gives inputs invalid according to the doc

src/sage/graphs/generic_graph.py
```diff
@@ -243,7 +243,8 @@
     :meth:`~GenericGraph.connected_components_sizes` | Return the sizes of the connected components as a list.
     :meth:`~GenericGraph.blocks_and_cut_vertices` | Compute the blocks and cut vertices of the graph.
     :meth:`~GenericGraph.blocks_and_cuts_tree` | Compute the blocks-and-cuts tree of the graph.
-    :meth:`~GenericGraph.is_cut_edge` | Return ``True`` if the input edge is a cut-edge or a bridge.
+    :meth:`~GenericGraph.is_cut_edge` | Check whether the input edge is a cut-edge or a bridge.
+    :meth:`~GenericGraph.is_edge_cut` | Check whether ``edges`` is an edge cut of ``G``.
     :meth:`~GenericGraph.is_cut_vertex` | Return ``True`` if the input vertex is a cut-vertex.
```
kwankyu, line 248: The same comment as above.

src/sage/graphs/connectivity.pyx
```diff
@@ -732,13 +734,173 @@ def blocks_and_cuts_tree(G):
...
+    from sage.graphs.generic_graph import GenericGraph
+    if not isinstance(G, GenericGraph):
+        raise TypeError("the input must be a Sage graph")
+
```
kwankyu: I think this input check is not necessary.

kwankyu: We may assume that the user is responsible with what (s)he is doing.

src/sage/graphs/connectivity.pyx
dcoudert: I will do the changes, but note that in this case we can give a list of edges or a set of edges or...

src/sage/graphs/connectivity.pyx
dcoudert: this is the error message of `_scream_if_not_simple`.  We may decide to change the error message of `_scream_if_not_simple`,  but it should be done in a separate PR.

src/sage/graphs/connectivity.pyx
dcoudert: done

src/sage/graphs/connectivity.pyx
kwankyu: Then you may say "a set of edges". A method that accepts a python set would also accept a python list, an iterator, and the like.

src/sage/graphs/connectivity.pyx
dcoudert: we do such test in many methods that can be directly imported. So I'm in favor of letting it.

kwankyu: An input check may be necessary only when it is likely that a user inputs a wrong type of value and silently get a garbage output, and hence it is necessary to guard the user.

I think it is not such a case here.

src/sage/graphs/connectivity.pyx
kwankyu: OK. Then it is not in the scope of this PR.

src/sage/graphs/connectivity.pyx
dcoudert: I have removed it in both `is_edge_cut` and `is_cut_edge`.

kwankyu: Thanks.

src/sage/graphs/connectivity.pyx
```diff
@@ -732,13 +734,160 @@ def blocks_and_cuts_tree(G):
     return g
 
 
+def is_edge_cut(G, edges):
+    """
+    Check whether the input edges form an edge cut.
+
```
kwankyu, line 740:
```
Check whether ``edges`` form an edge cut.
```

src/sage/graphs/generic_graph.py
kwankyu:
```
Check whether the input edges form an edge cut
```

src/sage/graphs/connectivity.pyx
```diff
@@ -732,13 +734,160 @@ def blocks_and_cuts_tree(G):
...
+        if len(set(DS.find(v) for v in comp)) > 1:
+            return True
+
+    return False
+
+
 def is_cut_edge(G, u, v=None, label=None):
     """
-    Return ``True`` if the input edge is a cut-edge or a bridge.
+    Check whether the input edge is a cut-edge or a bridge.
```
kwankyu, line 887: Is the input edge (u, v)? Is G the graph?

Then
```
Check whether the edge ``(u, v)`` is a cut-edge or a bridge of graph ``G``
```
?

dcoudert: I changed accordingly.

src/sage/graphs/connectivity.pyx
dcoudert: done.

kwankyu: Thanks.

src/sage/graphs/generic_graph.py
kwankyu: No plan to resolve?

dcoudert: I did not answer this specific comment, but the change has been done

kwankyu: No, not yet.

kwankyu: You did it in `src/sage/graphs/connectivity.pyx` but not in `src/sage/graphs/generic_graph.py`.

kwankyu (approved):
Thanks. LGTM.

dcoudert:
Thank you for the review.

dcoudert:
I have only rebased the PR on last beta. I set it back to positive review.

PR #38455, opened 2024-07-30
unitary DFT for symmetric group algebra

jacksonwalters:
### :memo: Checklist

### :hourglass: Dependencies

- #39200: Uses the decomposition.

### DIFF
--- a/src/sage/combinat/symmetric_group_algebra.py
+++ b/src/sage/combinat/symmetric_group_algebra.py
@@ -2035,10 +2035,23 @@ class SymmetricGroupAlgebra_n(GroupAlgebra_class):
           (:meth:`antipode`) of the seminormal basis instead of
           the seminormal basis.
 
-        EXAMPLES::
+        - ``form`` -- string (default: ``"modular"`` if `p|n!` else ``"seminormal"``);
+          one of the following:
 
-            sage: QS3 = SymmetricGroupAlgebra(QQ, 3)
-            sage: QS3.dft()
+          * ``"seminormal"`` -- use the seminormal basis
+          * ``"modular"`` -- use the modular DFT, which uses the Peirce decomposition
+            of the group algebra into blocks with central orthogonal idempotents
+          * ``"unitary"`` -- use the unitary DFT, which computes the extended
+            Cholesky decomposition of an `S_n`-invariant symmetric bilinear form as
+            a change-of-basis matrix (for positive characteristics, it must be
+            a finite field of square order)
+
+        EXAMPLES:
+
+        The default is the seminormal DFT when the characteristic does not divide the group order::
+
+            sage: QQ_S3 = SymmetricGroupAlgebra(QQ, 3)
+            sage: QQ_S3.dft()
             [   1    1    1    1    1    1]
             [   1  1/2   -1 -1/2 -1/2  1/2]
             [   0  3/4    0  3/4 -3/4 -3/4]
@@ -2046,11 +2059,36 @@ class SymmetricGroupAlgebra_n(GroupAlgebra_class):
             [   1 -1/2    1 -1/2 -1/2 -1/2]
             [   1   -1   -1    1    1   -1]
 
-        Over fields of characteristic `p > 0` such that `p \mid n!`, we use the
+        The unitary form works in characteristic zero::
+
+            sage: U = QQ_S3.dft(form='unitary'); U
+            [-1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2]
+            [       1/3*sqrt3        1/6*sqrt3       -1/3*sqrt3       -1/6*sqrt3       -1/6*sqrt3        1/6*sqrt3]
+            [               0              1/2                0              1/2             -1/2             -1/2]
+            [               0              1/2                0             -1/2              1/2             -1/2]
+            [       1/3*sqrt3       -1/6*sqrt3        1/3*sqrt3       -1/6*sqrt3       -1/6*sqrt3       -1/6*sqrt3]
+            [-1/6*sqrt3*sqrt2  1/6*sqrt3*sqrt2  1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2  1/6*sqrt3*sqrt2]
+            sage: U*U.H == 1
+            True
+
+        Over finite fields of square order with characteristic `p > n`, we can perform the unitary DFT::
+
+            sage: GF25_S3 = SymmetricGroupAlgebra(GF(5**2), 3)
+            sage: U = GF25_S3.dft(form='unitary'); U
+            [       1        1        1        1        1        1]
+            [2*z2 + 4   z2 + 2 3*z2 + 1 4*z2 + 3 4*z2 + 3   z2 + 2]
+            [       0        2        0        2        3        3]
+            [       0   z2 + 1        0 4*z2 + 4   z2 + 1 4*z2 + 4]
+            [2*z2 + 4 4*z2 + 3 2*z2 + 4 4*z2 + 3 4*z2 + 3 4*z2 + 3]
+            [       1        4        4        1        1        4]
+            sage: U*U.H == 1
+            True
+
+        Over fields of characteristic `p > 0` such that `p|n!`, we use the
         modular Fourier transform (:issue:`37751`)::
 
-            sage: GF2S3 = SymmetricGroupAlgebra(GF(2), 3)
-            sage: GF2S3.dft()
+            sage: GF2_S3 = SymmetricGroupAlgebra(GF(2), 3)
+            sage: GF2_S3.dft()
             [1 0 0 0 1 0]
             [0 1 0 0 0 1]
             [0 0 1 0 0 1]
@@ -2066,7 +2104,108 @@ class SymmetricGroupAlgebra_n(GroupAlgebra_class):
             return self._dft_seminormal(mult=mult)
         if form == "modular":
             return self._dft_modular()
+        if form == "unitary":
+            return self._dft_unitary()
         raise ValueError("invalid form (= %s)" % form)
+
+    def _dft_unitary(self):
+        """
+        Return the unitary form of the discrete Fourier transform for ``self``.
+
+        EXAMPLES::
+
+            sage: QQ_S3 = SymmetricGroupAlgebra(QQ, 3)
+            sage: QQ_S3._dft_unitary()
+            [-1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2]
+            [       1/3*sqrt3        1/6*sqrt3       -1/3*sqrt3       -1/6*sqrt3       -1/6*sqrt3        1/6*sqrt3]
+            [               0              1/2                0              1/2             -1/2             -1/2]
+            [               0              1/2                0             -1/2              1/2             -1/2]
+            [       1/3*sqrt3       -1/6*sqrt3        1/3*sqrt3       -1/6*sqrt3       -1/6*sqrt3       -1/6*sqrt3]
+            [-1/6*sqrt3*sqrt2  1/6*sqrt3*sqrt2  1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2 -1/6*sqrt3*sqrt2  1/6*sqrt3*sqrt2]
+            sage: GF49_S3 = SymmetricGroupAlgebra(GF(7**2), 3)
+            sage: GF49_S3._dft_unitary()
+            [5*z2 + 5 5*z2 + 5 5*z2 + 5 5*z2 + 5 5*z2 + 5 5*z2 + 5]
+            [2*z2 + 5   z2 + 6 5*z2 + 2 6*z2 + 1 6*z2 + 1   z2 + 6]
+            [       0 4*z2 + 5        0 4*z2 + 5 3*z2 + 2 3*z2 + 2]
+            [       0 3*z2 + 2        0 4*z2 + 5 3*z2 + 2 4*z2 + 5]
+            [2*z2 + 5 6*z2 + 1 2*z2 + 5 6*z2 + 1 6*z2 + 1 6*z2 + 1]
+            [5*z2 + 5 2*z2 + 2 2*z2 + 2 5*z2 + 5 5*z2 + 5 2*z2 + 2]
+
+        TESTS::
+
+            sage: QQ_S3 = SymmetricGroupAlgebra(QQ, 3)
+            sage: U = QQ_S3._dft_unitary()
+            sage: U*U.H == 1
+            True
+            sage: GF25_S3 = SymmetricGroupAlgebra(GF(5**2), 3)
+            sage: U = GF25_S3._dft_unitary()
+            sage: U*U.H == 1
+            True
+            sage: GF5_S3 = SymmetricGroupAlgebra(GF(5), 3)
+            sage: U = GF5_S3._dft_unitary()
+            Traceback (most recent call last):
+            ...
+            ValueError: the base ring must be a finite field of square order
+            sage: Z5_S3 = SymmetricGroupAlgebra(Integers(5), 3)
+            sage: U = Z5_S3._dft_unitary()
+            Traceback (most recent call last):
+            ...
+            ValueError: the base ring must be a finite field of square order
+            sage: F.<x> = FunctionField(GF(3))
+            sage: GF3_x_S3 = SymmetricGroupAlgebra(F, 5)
+            sage: U = GF3_x_S3._dft_unitary()
+            Traceback (most recent call last):
+            ...
+            ValueError: the base ring must be a finite field of square order
+            sage: GF9_S3 = SymmetricGroupAlgebra(GF(3**2), 3)
+            sage: U = GF9_S3._dft_unitary()
+            Traceback (most recent call last):
+            ...
+            NotImplementedError: not implemented when p|n!; dimension of invariant forms may be greater than one
+        """
+        from sage.matrix.special import diagonal_matrix
+        F = self.base_ring()
+        G = self.group()
+
+        if F.characteristic() == 0:
+            from sage.misc.functional import sqrt
+            from sage.rings.number_field.number_field import NumberField
+            dft_matrix = self.dft()
+            n = dft_matrix.nrows()
+            diag = [sum(dft_matrix[i, j] * dft_matrix[i, j].conjugate() for j in range(n))
+                    for i in range(n)]
+            primes_needed = {factor for d in diag for factor, _ in d.squarefree_part().factor()}
+            names = [f"sqrt{factor}" for factor in primes_needed]
+            x = PolynomialRing(QQ, 'x').gen()
+            K = NumberField([x**2 - d for d in primes_needed], names=names)
+            dft_matrix = dft_matrix.change_ring(K)
+            for i, d in enumerate(diag):
+                dft_matrix[i] *= ~sqrt(K(d))
+            return dft_matrix
+
+        # positive characteristic case
+        assert F.characteristic() > 0, "F must have positive characteristic"
+        if not (F.is_field() and F.is_finite() and F.order().is_square()):
+            raise ValueError("the base ring must be a finite field of square order")
+        if F.characteristic().divides(G.cardinality()):
+            raise NotImplementedError("not implemented when p|n!; dimension of invariant forms may be greater than one")
+        q = F.order().sqrt()
+
+        def conj_square_root(u):
+            if not u:
+                return F.zero()
+            z = F.multiplicative_generator()
+            k = u.log(z)
+            if k % (q+1) != 0:
+                raise ValueError(f"unable to factor as {u} is not in base field GF({q})")
+            return z ** ((k//(q+1)) % (q-1))
+
+        dft_matrix = self.dft()
+        n = dft_matrix.nrows()
+        for i in range(n):
+            d = sum(dft_matrix[i, j] * dft_matrix[i, j].conjugate() for j in range(n))
+            dft_matrix[i] *= ~conj_square_root(d)
+        return dft_matrix
 
     def _dft_seminormal(self, mult='l2r'):
         """
--- a/src/sage/combinat/symmetric_group_representations.py
+++ b/src/sage/combinat/symmetric_group_representations.py
@@ -67,6 +67,7 @@ def SymmetricGroupRepresentation(partition, implementation='specht',
       * ``'seminormal'`` -- for Young's seminormal representation
       * ``'orthogonal'`` -- for Young's orthogonal representation
       * ``'specht'`` -- for Specht's representation
+      * ``'unitary'`` -- for the unitary representation
 
     - ``ring`` -- the ring over which the representation is defined
 
@@ -132,6 +133,23 @@ def SymmetricGroupRepresentation(partition, implementation='specht',
         sage: spc.verify_representation()
         True
 
+    The unitary representation::
+
+        sage: unitary = SymmetricGroupRepresentation([3,1], "unitary"); unitary
+        Unitary representation of the symmetric group corresponding to [3, 1]
+        sage: unitary_GF49 = SymmetricGroupRepresentation([3,1], "unitary", ring=GF(7**2)); unitary_GF49
+        Unitary representation of the symmetric group corresponding to [3, 1]
+        sage: unitary_GF49([2,1,3,4])
+        [6 0 0]
+        [0 1 0]
+        [0 0 1]
+        sage: unitary_GF49([3,2,1,4])
+        [       4 2*z2 + 3        0]
+        [5*z2 + 5        3        0]
+        [       0        0        1]
+        sage: unitary_GF49.verify_representation()
+        True
+
     By default, any representation matrices that are computed are cached::
 
         sage: spc = SymmetricGroupRepresentation([3,2], "specht")
@@ -193,6 +211,7 @@ def SymmetricGroupRepresentations(n, implementation='specht', ring=None,
       * ``'seminormal'`` -- for Young's seminormal representation
       * ``'orthogonal'`` -- for Young's orthogonal representation
       * ``'specht'`` -- for Specht's representation
+      * ``'unitary'`` -- for the unitary representation
 
     - ``ring`` -- the ring over which the representation is defined
 
@@ -239,6 +258,23 @@ def SymmetricGroupRepresentations(n, implementation='specht', ring=None,
         [ 0  1 -1 -1  1]
         [ 0  1  0 -1  1]
 
+    The unitary representation.
+
+    ::
+
+        sage: unitary = SymmetricGroupRepresentations(3, "unitary"); unitary
+        Unitary representations of the symmetric group of order 3! over Symbolic Ring
+        sage: unitary_GF49 = SymmetricGroupRepresentations(4, "unitary", ring=GF(7**2)); unitary_GF49
+        Unitary representations of the symmetric group of order 4! over Finite Field in z2 of size 7^2
+        sage: unitary_GF49([3,1])([2,1,3,4])
+        [6 0 0]
+        [0 1 0]
+        [0 0 1]
+        sage: unitary_GF49([3,1])([3,2,1,4])
+        [       4 2*z2 + 3        0]
+        [5*z2 + 5        3        0]
+        [       0        0        1]
+
     .. NOTE::
 
         The implementation is based on the paper [Las]_.
@@ -253,6 +289,8 @@ def SymmetricGroupRepresentations(n, implementation='specht', ring=None,
         return YoungRepresentations_Orthogonal(n, ring=ring, cache_matrices=cache_matrices)
     elif implementation == "specht":
         return SpechtRepresentations(n, ring=ring, cache_matrices=cache_matrices)
+    elif implementation == "unitary":
+        return UnitaryRepresentations(n, ring=ring, cache_matrices=cache_matrices)
     else:
         raise NotImplementedError("only seminormal, orthogonal and specht are implemented")
 
@@ -337,7 +375,7 @@ class SymmetricGroupRepresentation_generic_class(Element):
             sage: spc == SymmetricGroupRepresentation([3])
             True
         """
-        if not isinstance(other, type(other)):
+        if not isinstance(other, type(self)):
             return False
         return (self._ring, self._partition) == (other._ring, other._partition)
 
@@ -659,7 +697,7 @@ class YoungRepresentation_generic(SymmetricGroupRepresentation_generic_class):
                 iu = index_lookup[u]
                 iv = index_lookup[v]
                 M[iu, iu], M[iu, iv], M[iv, iu], M[iv, iv] = \
-                    self._2x2_matrix_entries(beta)
+                    self._2x2_matrix_entries(self._ring(beta))
         return M
 
     def _representation_matrix_uncached(self, permutation):
@@ -914,7 +952,7 @@ class SpechtRepresentation(SymmetricGroupRepresentation_generic_class):
         """
         if permutation is None:
             permutation = Permutation(range(1, 1 + self._n))
-        Q = matrix(QQ, len(self._yang_baxter_graph))
+        Q = matrix(self._ring, len(self._yang_baxter_graph))
         for i, v in enumerate(self._dual_vertices):
             for j, u in enumerate(self._yang_baxter_graph):
                 Q[i, j] = self.scalar_product(tuple(permutation.action(v)), u)
@@ -998,6 +1036,198 @@ class SpechtRepresentations(SymmetricGroupRepresentations_class):
             Specht representations of the symmetric group of order 4! over Integer Ring
         """
         return "Specht representations of the symmetric group of order %s! over %s" % (self._n, self._ring)
+
+# #### Unitary Representation ###############################################
+
+
+class UnitaryRepresentation(SymmetricGroupRepresentation_generic_class):
+    r"""
+    A unitary representation of the symmetric group.
+
+    In characteristic zero, this is the same as the orthogonal representation since
+    they are defined over the reals. In positive characteristic, we are able to construct
+    representations when the field is finite, has square order, and the characteristic does not divide
+    the order of the group. In this case, we construct the representation by first
+    constructing the orthogonal representation `\rho` and then taking the extended
+    Cholesky decomposition of the unique solution `U` to the equation
+    `\rho(g)^T U \rho(g) = U` for all `g` in `G`.
+    """
+    def __init__(self, parent, partition):
+        r"""
+        Initialize ``self``.
+
+        EXAMPLES::
+
+            sage: U = SymmetricGroupRepresentation([2,1], "unitary")
+            sage: TestSuite(U).run()
+            sage: U = SymmetricGroupRepresentation([2,1], "unitary", GF(7))
+            Traceback (most recent call last):
+            ...
+            ValueError: the base ring must be a finite field of square order
+            sage: U = SymmetricGroupRepresentation([2,1], "unitary", GF(7**2))
+            sage: TestSuite(U).run()
+        """
+        if parent._ring.characteristic() == 0:
+            orth = SymmetricGroupRepresentation(partition, 'orthogonal')
+            self.representation_matrix = orth.representation_matrix
+            self._representation_matrix_uncached = orth._representation_matrix_uncached
+        else:
+            if not (parent._ring.is_field()
+                    and parent._ring.is_finite()
+                    and parent._ring.order().is_square()):
+                raise ValueError("the base ring must be a finite field of square order")
+            from sage.arith.misc import factorial
+            if parent._ring.characteristic().divides(factorial(parent._n)):
+                raise NotImplementedError("not implemented when p|n!; dimension of invariant forms may be greater than one")
+            self._q = parent._ring.order().sqrt()
+            self._specht = Permutations(sum(partition)).algebra(parent._ring).specht_module(partition)
+        super().__init__(parent, partition)
+
+    def _repr_(self):
+        r"""
+        String representation of ``self``.
+
+        EXAMPLES::
+
+            sage: SymmetricGroupRepresentation([2,1], "unitary")
+            Unitary representation of the symmetric group corresponding to [2, 1]
+        """
+        return f"Unitary representation of the symmetric group corresponding to {self._partition}"
+
+    @cached_method
+    def representation_matrix(self, permutation):
+        r"""
+        Return the matrix representing the ``permutation`` in this
+        irreducible representation.
+
+        .. NOTE::
+
+            This method caches the results.
+
+        EXAMPLES::
+
+            sage: unitary_specht = SymmetricGroupRepresentation([3,1], 'unitary', ring=GF(7**2))
+            sage: unitary_specht.representation_matrix(Permutation([2,1,3,4]))
+            [6 0 0]
+            [0 1 0]
+            [0 0 1]
+            sage: unitary_specht.representation_matrix(Permutation([3,2,1,4]))
+            [       4 2*z2 + 3        0]
+            [5*z2 + 5        3        0]
+            [       0        0        1]
+            sage: all(A * A.H == 1 for A in [unitary_specht.representation_matrix(g)
+            ....:                            for g in Permutations(4)])
+            True
+        """
+        ret = self._representation_matrix_uncached(permutation)
+        ret.set_immutable()
+        return ret
+
+    @lazy_attribute
+    def _unitary_change_basis_matrix(self):
+        """
+        Compute the change of basis matrix.
+
+        We first compute a `G`-invariant symmetric bilinear form. This yields a solution `U`
+        to the equation `\rho(g)^T U \rho(g) = U` for all `g` in `G`. We can
+        solve this equation by computing the null space of the matrix of coefficients of the linear
+        equations. Generically, this null space will be one dimensional (it may have higher dimension
+        in the modular case when the decomposition factors have multiplicity). We then take the
+        extended Cholesky decomposition of the unique solution to the equation.
+
+        EXAMPLES::
+
+            sage: unitary_specht = SymmetricGroupRepresentation([3,1], 'unitary', ring=GF(7**2))
+            sage: unitary_specht._unitary_change_basis_matrix
+            [       1        4        4]
+            [       0 2*z2 + 2 3*z2 + 3]
+            [       0        0 6*z2 + 1]
+            sage: unitary_specht = SymmetricGroupRepresentation([2,2], 'unitary', ring=GF(7**2))
+            sage: unitary_specht._unitary_change_basis_matrix
+            [       1        4]
+            [       0 2*z2 + 2]
+        """
+        from sage.rings.polynomial.polynomial_ring_constructor import PolynomialRing
+        G = Permutations(self._n)
+        F = self._ring
+        rho = self._specht.representation_matrix
+
+        # compute the invariant symmetric bilinear matrix
+        d_rho = self._specht.dimension()
+        R = PolynomialRing(F, 'u', d_rho**2)
+        U_vars = R.gens()
+        Utemp = matrix(R, d_rho, d_rho, U_vars)
+
+        def augmented_matrix(g):
+            rho_g = rho(g)
+            equation_matrix = rho_g.transpose() * Utemp * rho_g.conjugate() - Utemp
+            augmented_system = []
+            for i in range(d_rho):
+                for j in range(d_rho):
+                    linear_expression = equation_matrix[i, j]
+                    row = [linear_expression.coefficient(u) for u in U_vars]
+                    augmented_system.append(row)
+            return augmented_system
+
+        total_system = sum((augmented_matrix(g) for g in G), [])
+        null_space = matrix(F, total_system).right_kernel()
+        U = matrix(F, d_rho, d_rho, null_space.basis()[0])
+        return U.cholesky(extended=True).H
+
+    def _representation_matrix_uncached(self, permutation):
+        r"""
+        Return the matrix representing the ``permutation`` in this
+        irreducible representation.
+
+        We use the unitary change of basis matrix to conjugate
+        the representation matrix. These are unitary since
+
+        .. MATH::
+
+            (A rho(g) A^{-1})^* \cdot (A rho(g) A^{-1})
+              = (A^{-1})^* rho(g)^* A^* A rho(g) A^{-1}
+              = (A^{-1})^*H rho(g)^T U rho(g) A^{-1}
+              = (A^{-1})^* U A^{-1} = 1.
+
+        .. NOTE::
+
+            This method caches the results.
+
+        EXAMPLES::
+
+            sage: unitary_specht = SymmetricGroupRepresentation([2,2], 'unitary', ring=GF(7**2))
+            sage: unitary_specht._representation_matrix_uncached(Permutation([3,1,4,2]))
+            [       4 5*z2 + 4]
+            [2*z2 + 2        3]
+            sage: unitary_specht._representation_matrix_uncached(Permutation([1,2,4,3]))
+            [6 0]
+            [0 1]
+            sage: all(A * A.H == 1 for A in [unitary_specht._representation_matrix_uncached(g)
+            ....:                            for g in Permutations(4)])
+            True
+        """
+        assert self._ring.characteristic() > 0
+        rho = self._specht.representation_matrix
+        A = self._unitary_change_basis_matrix
+        return A * rho(permutation) * A.inverse()
+
+
+class UnitaryRepresentations(SymmetricGroupRepresentations_class):
+    _default_ring = SR
+
+    Element = UnitaryRepresentation
+
+    def _repr_(self):
+        r"""
+        String representation of ``self``.
+
+        EXAMPLES::
+
+            sage: spc = SymmetricGroupRepresentations(4)
+            sage: spc
+            Specht representations of the symmetric group of order 4! over Integer Ring
+        """
+        return "Unitary representations of the symmetric group of order %s! over %s" % (self._n, self._ring)
 
 # ##### Miscellaneous functions ############################################
 
--- a/src/sage/matrix/matrix2.pyx
+++ b/src/sage/matrix/matrix2.pyx
@@ -12943,7 +12943,186 @@ cdef class Matrix(Matrix1):
         else:
             return subspace
 
-    def cholesky(self):
+    def _cholesky_extended_ff(self):
+        r"""
+        Performs the extended Cholesky decomposition of a Hermitian matrix over a finite field of square order.
+
+        INPUT:
+
+        - ``self`` -- a square matrix with entries from a finite field of square order
+
+        .. SEEALSO::
+
+            :meth:`cholesky()`
+
+        OUTPUT:
+
+        For a Hermitian matrix `A` the routine returns a matrix `B` such that,
+
+        .. MATH::
+
+            A = B B^*,
+
+        where `B^*` is the conjugate-transpose.
+
+        ALGORITHM:
+
+        First, we ensure the matrix is square and defined over a finite field of square order. Then we can perform the conjugate-symmetric
+        version of Gaussian elimination, but the resulting decomposition matrix `L` might not be lower triangular.
+
+        This is a translation of ``BaseChangeToCanonical`` from the GAP ``forms`` package (for a Hermitian form).
+
+        EXAMPLES:
+
+        Here we use the extended decomposition, where the result may not be a lower triangular matrix::
+
+            sage: U = matrix(GF(17**2),[[0,1],[1,0]])
+            sage: B = U._cholesky_extended_ff(); B
+            [13*z2 + 6 3*z2 + 16]
+            [13*z2 + 6 14*z2 + 1]
+            sage: U == B * B.H
+            True
+            sage: U = matrix(GF(13**2),[[1,4,7],[4,1,4],[7,4,1]])
+            sage: B = U._cholesky_extended_ff(); B
+            [        1         0         0]
+            [        4  7*z2 + 3         0]
+            [        7 6*z2 + 10 12*z2 + 6]
+            sage: U == B * B.H
+            True
+            sage: U = matrix(GF(7**2), [[0, 1, 2], [1, 0, 1], [2, 1, 0]])
+            sage: B = U._cholesky_extended_ff(); B
+            [4*z2 + 2     6*z2        0]
+            [4*z2 + 2       z2        0]
+            [5*z2 + 6       z2       z2]
+            sage: U == B * B.H
+            True
+
+        TESTS:
+
+        If the matrix is not full rank, we compute the rank and throw an exception.
+
+            sage: U = matrix(GF(3**2),[[1,4,7],[4,1,4],[7,4,1]])
+            sage: U._cholesky_extended_ff()
+            Traceback (most recent call last):
+            ...
+            ValueError: matrix is not full rank
+            sage: U = matrix(GF(3**2),[[0,4,7],[4,1,4],[7,4,1]])
+            sage: U._cholesky_extended_ff()
+            Traceback (most recent call last):
+            ...
+            ValueError: matrix is not full rank
+        """
+        from sage.matrix.constructor import identity_matrix
+
+        if not self.is_hermitian():
+            raise ValueError("matrix is not Hermitian")
+
+        F = self._base_ring
+        n = self.nrows()
+        if self.fetch("rank") not in [n, None]:
+            raise ValueError("matrix is not full rank")
+        if not (F.is_finite() and F.order().is_square()):
+            raise ValueError("the base ring must be a finite field of square order")
+
+        q = F.order().sqrt(extend=False)
+
+        def conj_square_root(u):
+            if u == 0:
+                return 0
+            z = F.multiplicative_generator()
+            k = u.log(z)
+            if k % (q + 1) != 0:
+                raise ValueError(f"unable to factor: {u} is not in base field GF({q})")
+            return z ** (k//(q+1))
+
+        if self.nrows() == 1 and self.ncols() == 1:
+            return self.__class__(F, [conj_square_root(self[0][0])])
+
+        A = copy(self)
+        D = identity_matrix(F, n)
+        row = -1
+
+        # Diagonalize A
+        while row != n - 1:
+            row += 1
+
+            # Look for a non-zero element on the main diagonal, starting from `row`
+            i = row
+            while i < n and A[i, i].is_zero():
+                i += 1
+
+            if i == row:
+                # Do nothing since A[row, row] != 0
+                pass
+            elif i < n:
+                # Swap to ensure A[row, row] != 0
+                A.swap_rows(row, i)
+                A.swap_columns(row, i)
+                D.swap_rows(row, i)
+            else:
+                # All entries on the main diagonal are zero; look for an off-diagonal element
+                i = row
+                while i < n - 1:
+                    k = i + 1
+                    while k < n and A[i, k].is_zero():
+                        k += 1
+                    if k == n:
+                        i += 1
+                    else:
+                        break
+
+                if i == n - 1:
+                    # All elements are zero; terminate
+                    row -= 1
+                    r = row + 1
+                    break
+
+                # Fetch the non-zero element and place it at A[row, row + 1]
+                if i != row:
+                    A.swap_rows(row, i)
+                    A.swap_columns(row, i)
+                    D.swap_rows(row, i)
+
+                A.swap_rows(row + 1, k)
+                A.swap_columns(row + 1, k)
+                D.swap_rows(row + 1, k)
+
+                b = ~A[row + 1, row]
+                A.add_multiple_of_column(row, row + 1, b**q)
+                A.add_multiple_of_row(row, row + 1, b)
+                D.add_multiple_of_row(row, row + 1, b)
+
+            # Eliminate below-diagonal entries in the current column
+            a = ~(-A[row, row])
+            for i in range(row + 1, n):
+                b = A[i, row] * a
+                if not b.is_zero():
+                    A.add_multiple_of_column(i, row, b**q)
+                    A.add_multiple_of_row(i, row, b)
+                    D.add_multiple_of_row(i, row, b)
+
+        # Count how many variables have been used
+        if row == n - 1:
+            if A[n - 1, n - 1]:  # nonzero entry
+                r = n
+            else:
+                r = n - 1
+
+        if r < n:
+            self.cache('rank', r)
+            raise ValueError("matrix is not full rank")
+
+        # Normalize diagonal elements to 1
+        for i in range(r):
+            a = A[i, i]
+            if not a.is_one():
+                # Find an element `b` such that `a = b*b^q = b^(q+1)`
+                b = conj_square_root(a)
+                D.rescale_row(i, 1 / b)
+
+        return D.inverse()
+
+    def cholesky(self, extended=False):
         r"""
         Return the Cholesky decomposition of a Hermitian matrix.
 
@@ -12988,6 +13167,15 @@ cdef class Matrix(Matrix1):
         of original base ring, then we move to either its algebraic
         closure or the algebraic reals, depending on whether or not
         imaginary numbers are required.
+
+        Over finite fields, the Cholesky decomposition might
+        not exist, but when the field has square order (i.e.,
+        `\GF{q^2}`), then we can perform the conjugate-symmetric
+        version of Gaussian elimination, but the resulting
+        decomposition matrix `L` might not be lower triangular.
+
+        This is a translation of ``BaseChangeToCanonical`` from
+        the GAP ``forms`` package (for a Hermitian form).
 
         EXAMPLES:
 
@@ -13159,6 +13347,30 @@ cdef class Matrix(Matrix1):
             ...
             ValueError: matrix is not positive definite
 
+        Here we use the extended decomposition, where the result
+        may not be a lower triangular matrix::
+
+            sage: U = matrix(GF(5**2),[[0,1],[1,0]])
+            sage: B = U.cholesky(extended=True); B
+            [3*z2 4*z2]
+            [3*z2   z2]
+            sage: U == B * B.H
+            True
+            sage: U = matrix(GF(11**2),[[1,4,7],[4,1,4],[7,4,1]])
+            sage: B = U.cholesky(extended=True); B
+            [        1         0         0]
+            [        4  9*z2 + 2         0]
+            [        7 10*z2 + 1  3*z2 + 3]
+            sage: U == B * B.H
+            True
+            sage: U = matrix(GF(3**2), [[0, 1, 2], [1, 0, 1], [2, 1, 0]])
+            sage: B = U.cholesky(extended=True); B
+            [2*z2    2    0]
+            [2*z2    1    0]
+            [   0    1   z2]
+            sage: U == B * B.H
+            True
+
         TESTS:
 
         This verifies that :issue:`11274` is resolved::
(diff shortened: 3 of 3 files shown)


jacksonwalters:
https://github.com/sagemath/sage/issues/38456

jacksonwalters:
To get better results over \Q, we can work over a number field containing all necessary square roots:

#compute a field which contains all the square roots required
def containing_field(SGA):
    required_square_roots = []
    for partition in Partitions(SGA.group().degree()):
        specht_module = SGA.specht_module(partition)
        rho = specht_module.representation_matrix
        group_size = SGA.group().cardinality()
        P = (1/group_size)*sum(rho(g)*rho(g).conjugate().transpose() for g in SGA.group())
        d, L = P.eigenmatrix_left()
        required_square_roots += [specht_module.dimension(),SGA.group().cardinality()] + d.diagonal()
    required_square_roots = flatten([[numerator(QQ(q)),denominator(QQ(q))] if q in QQ else q for q in required_square_roots])
    K = SGA.base_ring()
    for n in set(required_square_roots):
        R.<x> = PolynomialRing(K)
        if (x**2 - n).is_irreducible():
            gen_name = "sqrt"+str(n).replace("/","over")
            K = K.extension(sqrt(n).minpoly(),names=gen_name)
    return K

jacksonwalters:
getting environment error "pytest is not installed in the venv, skip checking tests that rely on it
Error: Process completed with exit code 1." added ~3 lines to handle case when diagonal contains higher degree algebraic numbers (n=5 and above).

jacksonwalters:
relevant MSE questions:

- [Weyl's unitary trick over number fields](https://math.stackexchange.com/questions/4952935/weyls-unitary-trick-over-number-fields/4952958)
- [Roots of a polynomial over $\mathbb{Q}(\sqrt{2},\sqrt{3})$](https://math.stackexchange.com/questions/4953796/roots-of-a-polynomial-over-mathbbq-sqrt2-sqrt3/4953808#4953808)

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,70 @@ def dft(self, form=None, mult='l2r'):
...
+            K = self.base_ring()
+            for n in set(required_square_roots):
+                R = PolynomialRing(K, 'x')
+                x = R.gen()
+                if n.is_rational() and (x**2 - n).is_irreducible():
+                    gen_name = "sqrt"+str(n).replace("/","over")
+                    K = K.extension(sqrt(n).minpoly(),names=gen_name)
+                if not n.is_rational() and sqrt(n).minpoly().is_irreducible():
+                    gen_name = "deg" + str(n.minpoly().degree()) + "index" + str(list(set(required_square_roots)).index(n))
+                    K = K.extension(sqrt(n).minpoly(),names=gen_name)
```
tscrim:
```suggestion
        G = self.group()
        group_size = G.cardinality()
        data = {}
        for partition in Partitions(G.degree()):
            specht_module = self.specht_module(partition)
            rho = specht_module.representation_matrix
            P = sum(rho(g) * rho(g).conjugate().transpose() for g in G) / group_size
            d, L = P.eigenmatrix_left()
            data[partition] = (specht_module, P, d, L)

        def all_roots_field():
            required_square_roots = []
            for partition in Partitions(G.degree()):
                specht_module, P, d, L = data[partition]
                required_square_roots += [specht_module.dimension(), group_size] + d.diagonal()
            required_square_roots = flatten([[QQ(q).numerator(), QQ(q).denominator()] if q in QQ else q for q in required_square_roots])
            K = self.base_ring()
            for n in set(required_square_roots):
                R = PolynomialRing(K, 'x')
                x = R.gen()
                if n.is_rational() and (x**2 - n).is_irreducible():
                    gen_name = "sqrt"+str(n).replace("/","over")
                    K = K.extension(sqrt(n).minpoly(),names=gen_name)
                if not n.is_rational() and sqrt(n).minpoly().is_irreducible():
                    gen_name = "deg" + str(n.minpoly().degree()) + "index" + str(list(set(required_square_roots)).index(n))
                    K = K.extension(sqrt(n).minpoly(),names=gen_name)
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,70 @@ def dft(self, form=None, mult='l2r'):
...
+            return K
+
+        def unitary_change_of_basis(partition,K):
+            rho = self.specht_module(partition).representation_matrix
+            group_size = self.group().cardinality()
+            P = (1/group_size)*sum(rho(g)*rho(g).conjugate().transpose() for g in self.group())
+            d, L = P.eigenmatrix_left()
+            return L.inverse() * diagonal_matrix([sqrt(K(a)) for a in d.diagonal()]) * L
```
tscrim:
```suggestion
        def unitary_change_of_basis(partition, K):
            specht_module, P, d, L = data[partition]
            return L.inverse() * diagonal_matrix([sqrt(K(a)) for a in d.diagonal()]) * L
        K = all_roots_field()
        unitary_data = {partition: unitary_change_of_basis(partition, K) for partition in Partitions(G.degree())}
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,70 @@ def dft(self, form=None, mult='l2r'):
...
+
+        def hat(f,partition,K):
+            specht_module = self.specht_module(partition)
+            rho = specht_module.representation_matrix
+            Q = unitary_change_of_basis(partition,K)
+            unitary_factor = specht_module.dimension()/self.group().cardinality()
+            sqrt_unitary_factor = sqrt(K(unitary_factor))
+            return sqrt_unitary_factor*sum(f(g)*Q.inverse()*rho(g)*Q for g in self.group())
```
tscrim:
```suggestion
        def hat(f, partition):
            specht_module, P, d, L = data[partition]
            rho = self.specht_module(partition).representation_matrix
            Q = unitary_data[partition]
            unitary_factor = specht_module.dimension() / group_size
            sqrt_unitary_factor = sqrt(K(unitary_factor))
            return sqrt_unitary_factor * sum(f(g) * Q.inverse() * rho(g) * Q for g in G if f(g))
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,70 @@ def dft(self, form=None, mult='l2r'):
...
+
+        K = all_roots_field()
+        delta = lambda s: lambda t: 1 if t == s else 0
+        return matrix(K,[flatten([hat(delta(g),partition,K).list() for partition in Partitions(self.group().degree())]) for g in self.group()]).transpose()
```
tscrim:
```suggestion
        return matrix(K, [flatten([hat(delta(g), partition).list() for partition in Partitions(G.degree())]) for g in G]).transpose()
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,70 @@ def dft(self, form=None, mult='l2r'):
...
+
+        def hat(f,partition,K):
+            specht_module = self.specht_module(partition)
+            rho = specht_module.representation_matrix
+            Q = unitary_change_of_basis(partition,K)
+            unitary_factor = specht_module.dimension()/self.group().cardinality()
+            sqrt_unitary_factor = sqrt(K(unitary_factor))
+            return sqrt_unitary_factor*sum(f(g)*Q.inverse()*rho(g)*Q for g in self.group())
+
+        K = all_roots_field()
```
tscrim:
```suggestion
```
Moved earlier.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,70 @@ def dft(self, form=None, mult='l2r'):
...
+        EXAMPLES::
+
+            sage: QS3 = SymmetricGroupAlgebra(QQ, 3)
+            sage: QS3._dft_unitary()
+            [-1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3]
+            [      -1/3*sqrt3                0              1/2        1/6*sqrt3        1/6*sqrt3             -1/2]
+            [               0       -1/3*sqrt3        1/6*sqrt3              1/2             -1/2        1/6*sqrt3]
+            [               0       -1/3*sqrt3        1/6*sqrt3             -1/2              1/2        1/6*sqrt3]
+            [      -1/3*sqrt3                0             -1/2        1/6*sqrt3        1/6*sqrt3              1/2]
+            [-1/6*sqrt2*sqrt3  1/6*sqrt2*sqrt3  1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3 -1/6*sqrt2*sqrt3  1/6*sqrt2*sqrt3]
```
tscrim: Can you add a test showing the unitary property?

tscrim (changes requested):

src/sage/combinat/symmetric_group_algebra.py
jacksonwalters: Added U_dft*U_dft.conjugate().transpose() == identity_matrix(SGA.group().cardinality())
True

dimpase:
I checked that this builds if merged over the latest beta. Could you fix "This branch is out-of-date with the base branch" ?
The interactive rebase feature as provided by GitHub is reasonably easy to understand and work with.

tscrim:
That isn’t necessary until it is fully reviewed and only as a final check with the most recent changes.

dimpase:
These rebases unbreak the CI - so that's the main  reason to do it.

jacksonwalters:
It looks to me like the CI tests are passing (one skipped) so far.

jacksonwalters:
Work in issue [38456](https://github.com/sagemath/sage/issues/38456) is sufficient to construct this DFT. We compute G-invariant symmetric bilinear forms, factor the associated matrix $U=AA^\ast$, and obtain unitary representations $\tilde{\rho}(g)=A^\ast\rho(g)(A^{\ast})^{-1}$. The Fourier coefficients are $\hat{f}(\rho) = \sqrt{\frac{d_\rho}{|G|}}\sum_g f(g)\tilde{\rho}(g)$. However, $DFT.DFT^\ast = S$ where $S$ is a diagonal matrix consisting of signs, +1's and -1's. Factoring $S=RR^*$ with diagonal $R$, $uDFT = R^{-1}.DFT$ is unitary. Note $c=zz^\ast$ has a unique solution $z \in GF(q^2)$ when $c \in GF(q)$.

tscrim:
So the `Forms` package is a GAP package that has been included since GAP v4.9, but it is not included by our default installation nor in `gap_packages`.

However, I am thinking we should just reproduce their algorithm directly to take advantage of Sage's linear algebra packages. The only function really used is the `BaseChangeToCanonical`, and from looking at [the source code](https://github.com/gap-packages/forms/blob/77b084a7551309abbca529af39e10e31ae3c38cd/lib/forms.gi#L2279), it seems like they are just running [Gram-Schmidt by Gaussian elimination](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process#Via_Gaussian_elimination).

What do you think?

jacksonwalters:
> So the `Forms` package is a GAP package that has been included since GAP v4.9, but it is not included by our default installation nor in `gap_packages`.
> 
> However, I am thinking we should just reproduce their algorithm directly to take advantage of Sage's linear algebra packages. The only function really used is the `BaseChangeToCanonical`, and from looking at [the source code](https://github.com/gap-packages/forms/blob/77b084a7551309abbca529af39e10e31ae3c38cd/lib/forms.gi#L2279), it seems like they are just running [Gram-Schmidt by Gaussian elimination](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process#Via_Gaussian_elimination).
> 
> What do you think?

That's unfortunate that it's not included by default. I don't know what the barriers are to getting it incorporated, or how long that would take. It would certainly make things easy.

I agree. We'd been talking about doing this anyways, and I'm happy to go ahead and just rewrite it in Sage. I'll try to get a basic version working tomorrow. Then there will probably be ways to optimize it.

tscrim:
I believe it should just be getting the echelon form of the appropriately augmented Gram matrix of the initial sesquilinear form and then taking the appropriate submatrix (possibly transposed). So it should just be a few lines of code (at most).

jacksonwalters:
> I believe it should just be getting the echelon form of the appropriately augmented Gram matrix of the initial sesquilinear form and then taking the appropriate submatrix (possibly transposed). So it should just be a few lines of code (at most).

I agree it should a lot simpler than what is written in GAP. However, a straightforward augmentation and row reduction doesn't yield the correct matrix. For example, when `q=11` and `la=[2,1,1]` with `n=4`, we have

```
q = 11
F = GF(q**2)
U=matrix(F,[[1,4,7],[4,1,4],[7,4,1]])
unitary_change_of_basis(U,q)
[        1         0         0]
[        4  9*z2 + 2         0]
[        7 10*z2 + 1  3*z2 + 3]
```

which is correct. However,

```
def base_change_hermitian(U):
    F = U.base_ring()
    augmented_mat = U.augment(identity_matrix(F, U.nrows()))
    echelon_form = augmented_mat.echelon_form()
    return echelon_form[:, U.ncols():]

base_change_hermitian(U)
[7 2 9]
[2 7 2]
[9 2 7]
```

is incorrect.

tscrim:
Okay, a little bit more care is needed here since it does the RREF. We only want to operate on the strictly lower (or upper) triangle of the matrix as the other part is taken care of by the symmetry. So a short version is to use the inverse of the upper part of the LU decomposition:
```
sage: U = matrix(F,[[1,4,7],[4,1,4],[7,4,1]])
sage: Up = U.LU()[2]
sage: D = Up.diagonal()
sage: A = ~Up * matrix.diagonal([d.sqrt() for d in D])
sage: A.H * U * A
[ 1  0  0]
[ 0 10  0]
[ 0  0 10]
```
Of course, this is not very efficient since it is computing far more things than necessary. However, it is easy to code...

dimpase:
Mathematically it's not echelon form, and not even Gram-Schmid, as you have a twist by the Galois automorphism in the factorisation.

jacksonwalters:
> Mathematically it's not echelon form, and not even Gram-Schmidt, as you have a twist by the Galois automorphism in the factorisation.

For what it's worth, here is a direct translation of the GAP code into Sage which appears to be working now. Perhaps we can reduce it, but I agree with Dima that the twist makes it something slightly different.
```
#for u in GF(q), we can factor as u=aa^* using gen. z and modular arithmetic
def conj_square_root(u):
    if u == 0:
        return 0  # Special case for 0
    z = F.multiplicative_generator()
    k = u.log(z)  # Compute discrete log of u to the base z
    if k % (q+1) != 0:
        raise ValueError("Unable to factor: u is not in base field GF(q)")
    return z**((k//(q+1))%(q-1))

def base_change_matrix(mat):
    """
    Diagonalizes a Hermitian matrix over a finite field.
    Returns the base change matrix and the rank of the Hermitian form.
    
    Arguments:
        mat: The Gram matrix of a Hermitian form (Sage matrix object)
        gf: The finite field (GF(q))
    
    Returns:
        D: The base change matrix
        r: The number of non-zero rows in D*mat*D^T
    """

    gf = mat.base_ring()
    n = mat.nrows()
    q = gf.order()
    t = sqrt(q)

    A = copy(mat)
    D = identity_matrix(gf, n)
    row = 0

    # Diagonalize A
    while True:
        row += 1

        # Look for a non-zero element on the main diagonal, starting from `row`
        i = row - 1  # Adjust for zero-based indexing in Sage
        while i < n and A[i, i].is_zero():
            i += 1

        if i == row - 1:
            # Do nothing since A[row, row] != 0
            pass
        elif i < n:
            # Swap to ensure A[row, row] != 0
            A.swap_rows(row - 1, i)
            A.swap_columns(row - 1, i)
            D.swap_rows(row - 1, i)
        else:
            # All entries on the main diagonal are zero; look for an off-diagonal element
            i = row - 1
            while i < n - 1:
                k = i + 1
                while k < n and A[i, k].is_zero():
                    k += 1
                if k == n:
                    i += 1
                else:
                    break

            if i == n - 1:
                # All elements are zero; terminate
                row -= 1
                r = row
                break

            # Fetch the non-zero element and place it at A[row, row + 1]
            if i != row - 1:
                A.swap_rows(row - 1, i)
                A.swap_columns(row - 1, i)
                D.swap_rows(row - 1, i)

            A.swap_rows(row, k)
            A.swap_columns(row, k)
            D.swap_rows(row, k)

            b = A[row, row - 1]**(-1)
            A.add_multiple_of_row(row - 1, row, -b**t)
            A.add_multiple_of_column(row, row - 1, -b)
            D.add_multiple_of_row(row - 1, row, -b)

        # Eliminate below-diagonal entries in the current column
        a = -A[row - 1, row - 1]**(-1)
        for i in range(row, n):
            b = A[i, row - 1] * a
            if not b.is_zero():
                A.add_multiple_of_column(i,row - 1, b**t)
                A.add_multiple_of_row(i, row - 1, b)
                D.add_multiple_of_row(i, row - 1, b)

        if row == n - 1:
            break

    # Count how many variables have been used
    if row == n - 1:
        if not A[n - 1, n - 1].is_zero():
            r = n
        else:
            r = n - 1

    # Normalize diagonal elements to 1
    for i in range(r):
        a = A[i, i]
        if not a.is_one():
            # Find an element `b` such that `b*b^t = b^(t+1) = a`
            b = conj_square_root(a)
            D.rescale_row(i, 1 / b)

    return D
```

tscrim:
> Mathematically it's not echelon form, and not even Gram-Schmid, as you have a twist by the Galois automorphism in the factorisation.

I am pretty sure the first step of computing an orthogonal basis can be done by Gram-Schmidt. If it wasn’t, then it wouldn’t work over $\mathbb{C}$, where dot product is sesquilinear. (Although one has to be a little more careful about which side of the inner product one puts the vectors in.) In particular, when you compute the orthogonality with the new vector at each step, you can put it in the side that does not have the automorphism.

Now for normalizing the diagonal entries we have to be more careful since we need to solve $zz^* = d$, but that also has a known solution.

jacksonwalters:
Here is a complete method for obtaining the decomposition $U=AA^\ast$. It does work, but I don't understand why this is different than the result of the other two methods. Do we know if this decomposition is unique? Seems to be in the 1 dim'l case.

```
def base_change_hermitian(U):
    Up = U.LU()[2]
    D = Up.diagonal()
    A = ~Up * matrix.diagonal([d.sqrt() for d in D])
    diag = (A.H * U * A).diagonal()
    factor_diag = diagonal_matrix([conj_square_root(d) for d in diag])
    return (factor_diag*A.inverse()).H

#use the upper part of the LU decomposition, and factor the diagonal
U=matrix(F,[[1,4,7],[4,1,4],[7,4,1]])
B = base_change_hermitian(U)
print(B*B.H == U)
print(B)
True
[   1    0    0]
[   4 8*z2    0]
[   7 4*z2 2*z2]
```

jacksonwalters:
@tscrim Any idea what’s going on with the two failing checks (macos 3.11, meson/ conda python 3.11)? They were all passing before I squashed commits.

tscrim:
I usually just ignore those…

jacksonwalters:
I don't mind ignoring them, but don't they need to pass to merge?

tscrim:
They can be flaky. You can look at the logs and see that it isn’t anything related to your PR.

jacksonwalters:
Here's the error from the `macos-latest, 3.11` log:

```
sage -t --warn-long 5.0 --random-seed=16827736091212460951185318454288437369 src/sage/knots/link.py  # 1 doctest failed
sage -t --warn-long 5.0 --random-seed=16827736091212460951185318454288437369 src/sage/rings/number_field/galois_group.py  # 1 doctest failed [failed in baseline: unreported failure on macOS (sort order) in GaloisGroup_v2.artin_symbol, https://github.com/sagemath/sage/actions/runs/9525536510/job/26259809272?pr=37998]
sage -t --warn-long 5.0 --random-seed=16827736091212460951185318454288437369 src/sage/rings/number_field/number_field.py  # 2 doctests failed [failed in baseline: unreported failure on macOS (sort order) in NumberField_absolute.[optimized_]subfields, https://github.com/sagemath/sage/actions/runs/9525536510/job/26259809272?pr=37998]
sage -t --warn-long 5.0 --random-seed=16827736091212460951185318454288437369 src/sage/rings/qqbar.py  # 1 doctest failed [failed in baseline: unreported failure on macOS seen in https://github.com/sagemath/sage/actions/runs/9525536510/job/26259809272?pr=37998]
Total time for all tests: 1438.2 seconds
    cpu time: 3108.2 seconds
    cumulative wall time: 3381.7 seconds
Features detected for doctesting: conway_polynomials,database_cremona_mini_ellcurve,database_ellcurves,database_graphs,fpylll,gap_package_atlasrep,gap_package_polenta,gap_package_polycyclic,gfan,giac,lrcalc_python,mpmath,nauty,networkx,numpy,palp,pexpect,pillow,polytopes_db,pplpy,primecountpy,ptyprocess,pyparsing,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.giac,sage.libs.homfly,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.schemes,sage.symbolic,scipy,singular,sphinx,sympy
Running pytest on ['/Users/runner/work/sage/sage/src/sage', '/Users/runner/work/sage/sage/src/sage_setup', '/Users/runner/work/sage/sage/src/sage_docbuild', '/Users/runner/work/sage/sage/src/doc'] with options []
============================= test session starts ==============================
platform darwin -- Python 3.11.10, pytest-8.3.3, pluggy-1.5.0
rootdir: /Users/runner/work/sage/sage/src
configfile: tox.ini
plugins: xdist-3.6.1
collected 61 items / 2 skipped

src/sage/doctest/parsing_test.py .......sss.                             [ 18%]
src/sage/manifolds/differentiable/examples/symplectic_space_test.py .... [ 24%]
                                                                         [ 24%]
src/sage/manifolds/differentiable/symplectic_form_test.py .............. [ 47%]
.....................                                                    [ 81%]
src/sage/manifolds/differentiable/tensorfield_paral_test.py .            [ 83%]
src/sage/numerical/backends/cvxopt_backend_test.py ..                    [ 86%]
src/sage/numerical/backends/glpk_backend_test.py ..                      [ 90%]
src/sage/numerical/backends/glpk_exact_backend_test.py ..                [ 93%]
src/sage/numerical/backends/interactivelp_backend_test.py ..             [ 96%]
src/sage/numerical/backends/ppl_backend_test.py ..                       [100%]

======================== 58 passed, 5 skipped in 59.55s ========================
Error: Process completed with exit code 1.
```

Here's the one for Meson / Conda, Python 3.11:

```
Total time for all tests: 1812.4 seconds
    cpu time: 5374.2 seconds
    cumulative wall time: 5485.8 seconds
Features detected for doctesting: conway_polynomials,database_cremona_mini_ellcurve,database_ellcurves,database_graphs,fpylll,gap_package_atlasrep,gap_package_polenta,gap_package_polycyclic,gfan,giac,imagemagick,info,lrcalc_python,mpmath,nauty,networkx,numpy,palp,pexpect,pillow,polytopes_db,pplpy,primecountpy,ptyprocess,pyparsing,sage.combinat,sage.geometry.polyhedron,sage.graphs,sage.groups,sage.libs.braiding,sage.libs.ecl,sage.libs.flint,sage.libs.gap,sage.libs.giac,sage.libs.homfly,sage.libs.linbox,sage.libs.m4ri,sage.libs.ntl,sage.libs.pari,sage.libs.singular,sage.misc.cython,sage.modular,sage.modules,sage.numerical.mip,sage.plot,sage.rings.complex_double,sage.rings.finite_rings,sage.rings.function_field,sage.rings.number_field,sage.rings.padics,sage.rings.polynomial.pbori,sage.rings.real_double,sage.rings.real_mpfr,sage.schemes,sage.symbolic,scipy,singular,sphinx,sympy
Running pytest on ['/home/runner/work/sage/sage/src/sage', '/home/runner/work/sage/sage/src/doc'] with options []
============================= test session starts ==============================
platform linux -- Python 3.11.10, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/runner/work/sage/sage/src
configfile: tox.ini
plugins: xdist-3.6.1
collected 61 items / 2 skipped

src/sage/doctest/parsing_test.py ...........                             [ 18%]
src/sage/manifolds/differentiable/examples/symplectic_space_test.py .... [ 24%]
                                                                         [ 24%]
src/sage/manifolds/differentiable/symplectic_form_test.py .............. [ 47%]
.....................                                                    [ 81%]
src/sage/manifolds/differentiable/tensorfield_paral_test.py .            [ 83%]
src/sage/numerical/backends/cvxopt_backend_test.py ..                    [ 86%]
src/sage/numerical/backends/glpk_backend_test.py ..                      [ 90%]
src/sage/numerical/backends/glpk_exact_backend_test.py ..                [ 93%]
src/sage/numerical/backends/interactivelp_backend_test.py ..             [ 96%]
src/sage/numerical/backends/ppl_backend_test.py ..                       [100%]

=================== 61 passed, 2 skipped in 67.85s (0:01:07) ===================
Error: Process completed with exit code 64.
```

Not sure what these mean. I don't _think_ it has anything to do with the PR code.

---

Separately, I do need to add an assertion for the case $p \mid |G|$ since we are using the normalization factors $\sqrt{d_\rho/|G|}$. This avoid the situation when the space of $G$-invariant bilinear forms has dimension greater than one, e.g. `q=2` and `la=[3,1,1]` for `n=5`. 

This is not necessarily a bad thing, it's more a deferral to a time when better understand this case, which I would call the intersection of the modular and unitary DFT. Probably we'll have make use of two are more linearly independent forms, and I don't know how to do that (yet).

---

I'm also removing `conjugate_pos_char`. It's not necessary; one can just use `.H.transpose()`.

jacksonwalters:
@tscrim Is it worth throwing a warning on line `2181` if `len(U_mats) > 1`? In that we are later choosing `U_mats[0]`. Again, this will only happen in the modular case we are avoiding, but still.

tscrim:
You didn’t need to copy the log; there is no useful info really in it. (To my point, the last push reran the tests, which now pass.)

There is a `conjugate()` method for matrices IIRC. There is no point in transposing the matrix twice. :)

Rather than raise a warning, I would simply error out with a `NotImplementedError` since the current code doesn’t deal with that situation.

jacksonwalters:
Got rid of the log messages. Yes, looks like it was just flaky as you said. They are passing now. 

Good to know! Not sure how I missed that. Updated.

I'm now just raising a `NotImplementedError` for the modular case $p \mid |G|$. I'm also including the two key failure points in the error string

1. that the space of invariant forms can have dimension greater than one
2. the normalization factor is not defined.

It's a bit extra, but for someone trying to extend this it points them to the right place (look at how the modular DFT is performed, and try to adapt it).

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+            def hat(f, partition):
+                specht_module, P, d, L = data[partition]
+                rho = self.specht_module(partition).representation_matrix
+                Q = unitary_data[partition]
+                unitary_factor = specht_module.dimension() / G.cardinality()
+                sqrt_unitary_factor = sqrt(K(unitary_factor))
+                return sqrt_unitary_factor * sum(f(g) * Q.inverse() * rho(g) * Q for g in G if f(g))
+
+            delta = lambda s: lambda t: 1 if t == s else 0
+            return matrix(K, [flatten([hat(delta(g), partition).list() for partition in Partitions(G.degree())]) for g in G]).transpose()
```
tscrim:
```suggestion
            def hat(g, partition):
                specht_module, P, d, L = data[partition]
                rho = self.specht_module(partition).representation_matrix
                Q = unitary_data[partition]
                unitary_factor = specht_module.dimension() / G.cardinality()
                sqrt_unitary_factor = sqrt(K(unitary_factor))
                return sqrt_unitary_factor * Q.inverse() * rho(g) * Q

            PGdeg = Partitions(G.degree())
            return matrix(K, [flatten([hat(g, partition).list() for partition in PGdeg]) for g in G]).transpose()
```
I think this is equivalent since `delta` is just getting the summand `g` .

Although you probably can write this to get rid of that transpose and the `flatten`... The partitions parent `PGdeg` can also be created earlier (and reused here).

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+                K = F
+                for n in set(required_square_roots):
+                    R = PolynomialRing(K, 'x')
+                    x = R.gen()
+                    if n.is_rational() and (x**2 - n).is_irreducible():
+                        gen_name = "sqrt"+str(n).replace("/","over")
+                        K = K.extension(sqrt(n).minpoly(),names=gen_name)
+                    if not n.is_rational() and sqrt(n).minpoly().is_irreducible():
+                        gen_name = "deg" + str(n.minpoly().degree()) + "index" + str(list(set(required_square_roots)).index(n))
+                        K = K.extension(sqrt(n).minpoly(),names=gen_name)
```
tscrim:
```suggestion
                        K = K.extension(sqrt(n).minpoly(), names=gen_name)
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+                        for j in range(d_rho):
+                            linear_expression = equation_matrix[i, j]
+                            row = [linear_expression.coefficient(u) for u in U_vars]
+                            augmented_system.append(row)
+                    return matrix(F, augmented_system)
+                total_system = matrix(F,0,d_rho**2)
+                for g in G:
+                    total_system = total_system.stack(augmented_matrix(g))
+                null_space = total_system.right_kernel()
+                U_mats = [matrix(F,d_rho,d_rho,b) for b in null_space.basis()]
```
tscrim:
```suggestion
                U_mats = [matrix(F, d_rho, d_rho, b) for b in null_space.basis()]
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+                    rho_g = rho(g)
+                    equation_matrix = rho_g.transpose()*U*rho_g.conjugate() - U
+                    augmented_system = []
+                    for i in range(d_rho):
+                        for j in range(d_rho):
+                            linear_expression = equation_matrix[i, j]
+                            row = [linear_expression.coefficient(u) for u in U_vars]
+                            augmented_system.append(row)
+                    return matrix(F, augmented_system)
+                total_system = matrix(F,0,d_rho**2)
```
tscrim:
```suggestion
                total_system = matrix(F, 0, d_rho**2)
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+                raise NotImplementedError("Not implemented when p|n!. Dimension of invariant forms may be greater than one and sqrt{d_rho/|G|} is not defined. See modular DFT.")
+            q = sqrt(F.order())
+
+            def invariant_symmetric_bilinear_matrix(partition):
+                specht_module = self.specht_module(partition)
+                rho = specht_module.representation_matrix
+                d_rho = specht_module.dimension()
+                R = PolynomialRing(F,'u',d_rho**2)
+                U_vars = R.gens()
+                U = matrix(R,d_rho,d_rho,U_vars)
```
tscrim:
```suggestion
                R = PolynomialRing(F, 'u', d_rho**2)
                U_vars = R.gens()
                U = matrix(R, d_rho, d_rho, U_vars)
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+                U_mats = [matrix(F,d_rho,d_rho,b) for b in null_space.basis()]
+                return U_mats
+
+            def conj_square_root(u):
+                if u == 0:
+                    return 0
+                z = F.multiplicative_generator()
+                k = u.log(z)
+                if k % (q+1) != 0:
+                    raise ValueError("Unable to factor: u is not in base field GF(q)")
```
tscrim:
```suggestion
                    raise ValueError(f"unable to factor since {u} is not in base field GF({q})")
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+                return z**((k//(q+1)) % (q-1))
```
tscrim:
```suggestion
                return z ** ((k//(q+1)) % (q-1))
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+            def hat(f,partition):
+                specht_module = self.specht_module(partition)
+                rho = specht_module.representation_matrix
+                U = invariant_symmetric_bilinear_matrix(partition)[0]
+                A = base_change_hermitian(U)
+                sqrt_unitary_factor = sqrt(F(specht_module.dimension()/G.cardinality()))
+                return sqrt_unitary_factor*sum(f(g)*A*rho(g)*A.inverse() for g in G)
+
+            delta = lambda s: lambda t: 1 if t == s else 0
+            fourier_transform = [flatten([hat(delta(g),partition).list() for partition in Partitions(G.degree())]) for g in G]
```
tscrim:
```suggestion
            def hat(g, partition):
                specht_module = self.specht_module(partition)
                rho = specht_module.representation_matrix
                U = invariant_symmetric_bilinear_matrix(partition)[0]
                A = base_change_hermitian(U)
                sqrt_unitary_factor = sqrt(F(specht_module.dimension() / G.cardinality()))
                return sqrt_unitary_factor * f(g) * A * rho(g) * A.inverse()

            PGdeg = Partitions(G.degree())
            fourier_transform = [flatten([hat(g, partition).list() for partition in PGdeg]) for g in G]
```
Same comment as the char 0 case. Although it would likely be better to cache those square root computations somehow.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2068,8 +2068,152 @@ def dft(self, form=None, mult='l2r'):
...
+            dft_matrix = matrix(F,fourier_transform).transpose()
+            sign_diag = (dft_matrix*dft_matrix.H).diagonal()
+            factor_diag = diagonal_matrix([conj_square_root(d) for d in sign_diag])
+            return factor_diag.inverse()*dft_matrix
```
tscrim:
```suggestion
            dft_matrix = matrix(F, fourier_transform).transpose()
            sign_diag = (dft_matrix * dft_matrix.H).diagonal()
            factor_diag_inv = diagonal_matrix([~conj_square_root(d) for d in sign_diag])
            return factor_diag_inv * dft_matrix
```
Inverting a diagonal matrix is easy, and this should be faster.

tscrim (changes requested):

src/sage/combinat/symmetric_group_algebra.py
jacksonwalters: Yes, this was written to be general and obtain the Fourier coefficient for any function f. Since we only need the summand `g`, this is simpler.

src/sage/combinat/symmetric_group_algebra.py
jacksonwalters: Agree, inverting each diagonal element with ~ is better.

jacksonwalters:
@tscrim Changes have been implemented. I was able to remove all occurrences of `flatten`, but I don't see how to get rid of `transpose()`. 

I went back and looked at the DFT in characteristic zero. I noticed that if `U=SGA.dft()`, then `UU.H = D`, where `D` is diagonal. That was the original problem, to find a unitary DFT, which is solved by the first chunk of code. 

However, it seems we're now comfortable just factoring this resultant diagonal, and multiplying by its inverse to get a unitary DFT. In fact,

```
#an alternate method to create a unitary DFT in characteristic zero, avoiding other computations
U_dft_sqrt_diag = diagonal_matrix([sqrt(d) for d in (SGA_dft*SGA_dft.transpose()).diagonal()]).inverse()*SGA_dft
U_dft_sqrt_diag*U_dft_sqrt_diag.H == identity_matrix(G.cardinality())
True
```

I'm a bit conflicted, because I feel it's illuminating to see how these number fields are actually constructed. Then again, if the resulting square roots are just those on the diagonal of $UU^\ast$, perhaps that's saying the same thing. 

Do we want to replace the characteristic zero code with this simple computation?

jacksonwalters:
I think the bulk of this code should be moved to unitary representations (e.g. as another option like `orthogonal`). Then we can compute unitary rep'ns independently, and use them here if desired. 

I'm now in favor of just using the brief code which factors the resultant diagonal for the DFT, skipping the use of unitary rep'ns entirely.

jacksonwalters:
I have pushed these changes. We are now just using the usual DFT to compute $DFT.DFT^\ast = D$, a diagonal matrix, and then factoring that as $D=RR^\ast$ or $D=R^2$, and $uDFT = R^{-1}DFT$. This simplifies the code significantly. The DFT you get is a bit different than before, but it is unitary.

In the characteristic zero case for the DFT, we still explicitly compute the number field we are working over. This is better than using a symbolic ring.

I've created a class (probably incorrectly) called `UnitaryRepresentation` and `UnitaryRepresentations` in `symmetric_group_representations.py`. These are meant to enable one to compute individual unitary representations.

Note that since the orthogonal representations are real, they are also unitary in characteristic zero. I have removed the code that uses Weyl's unitary trick (even though it's cool), as well as the code that was explicitly computing the number field there. That code was causing Pari stack overflows even for n=5, and was very complicated, finding minimal polynomials for algebraic numbers on the diagonal.

I have moved the code for the finite field case, which is the meat of it. This definitely doing something new, as the orthogonal representations are definitely not unitary over `GF(q**2)`.

This will definitely need some tweaking, but I think it's a better organizational pattern to simplify the unitary DFT normalization and separate out the individual unitary representations.

jacksonwalters:
Note that there is an example for which the LDU decomposition method fails to yield a triangular decomposition $U = A^\ast A$. This case occurs when $q=3$ and $n=4$, so it is a case which we are avoiding by demanding $p \nmid n$. 

The GAP `forms` package works, but it is not installed by default. Therefore, we have opted to translate the GAP source code for BaseChangeToCanonical for a HermitianForm to Sage code. We are adding this as a standalone method:

`U.hermitian_decomposition()`

See: [PR #39200](https://github.com/sagemath/sage/pull/39200)

tscrim:
@jacksonwalters Please rebase this over #39200 and use the new method from that PR.

jacksonwalters:
> @jacksonwalters Please rebase this over #39200 and use the new method from that PR.

Will do.

jacksonwalters:
@tscrim Okay, I've successfully rebased, and replaced code in `symmetric_group_representations.py` beginning on line 1080, `_unitary_change_basis_matrix`, to now use `hermitian_decomposition()` from PR#39200.

jacksonwalters:
The doctests are currently failing. I'm using output from my laptop, but it doesn't match what's in Sage. For example,

```
 [       1        0        0]
 [       3 5*z2 + 4        0]
 [       4 3*z2 + 1       z2]
```

is a factorization of the matrix 

```
[1 3 4]
[3 1 3]
[4 3 1]
```

when it should be factorization of the matrix associated to the invariant symmetric bilinear form

```
[1 4 4]
[4 1 4]
[4 4 1]
```

I don't see any major differences between the code.

jacksonwalters:
@tscrim The only real difference is how the Specht modules are initiated. In the Sage code, we don't use SGA. It's a little concerning that setting that base_ring doesn't seem to change it:

```
rho = SymmetricGroupRepresentation(la,ring=GF(7**2))._representation_matrix_uncached; rho(Permutation([2,1,3,4])).base_ring()
Rational Field
```

tscrim:
That is a bit concerning, but since it is defined over Z, then conputationally it should be the same. So that shouldnt matter.

The invariance of the bilinear form needs to checked against the corresponding representation (and basis).

It should work for whichever set of representation matrices you use (which usually give different matrices). The existence of the unitary form is definitely an isomorphism invariant.

jacksonwalters:
> That is a bit concerning, but since it is defined over Z, then conputationally it should be the same. So that shouldnt matter.
> 
> The invariance of the bilinear form needs to checked against the corresponding representation (and basis).
> 
> It should work for whichever set of representation matrices you use (which usually give different matrices). The existence of the unitary form is definitely an isomorphism invariant.

It should be the same, but it's the only difference between what I have in the notebook version. The Sage version is definitely computing a different `U`, so the linear system must be different. I'll take a look at it.

I checked the two different forms, and only `[[1 4 4],[4 1 4],[4 4 1]]` is invariant while `[[1 3 4],[3 1 3],[4 3 1]]` fails. Here's the code (of course one must manually set `U` if it's not being computed above).

[](https://github.com/jacksonwalters/dft-finite-groups/blob/main/symmetric_group/unitary_repns_over_finite_fields.ipynb)

```
#ensure the resulting form is G-invariant, symmetric, bilinear by symbolic verification
def check_form_properties(q,partition):
    #define the representation matrix corresponding to q, partition
    F = GF(q**2)
    SGA = SymmetricGroupAlgebra(F,sum(partition))
    SM = SGA.specht_module(partition)
    rho = SM.representation_matrix
    d_rho = SM.dimension()
    G = SGA.group()

    #define variables as polynomial generators
    R_xy = PolynomialRing(F, d_rho, var_array='x,y')
    x = matrix([R_xy.gens()[2*i] for i in range(d_rho)]).transpose()
    y = matrix([R_xy.gens()[2*i+1] for i in range(d_rho)]).transpose()
    R_xy_lambda = PolynomialRing(R_xy,'lambda')
    lambda_ = R_xy_lambda.gens()[0]

    #compute the bilinear form matrix. coerce over polynomial ring
    U_mats = invariant_symmetric_bilinear_matrix(SGA,partition)
    if len(U_mats) > 1:
        print("Space of G-invariant symmetric bilinear forms has dimension > 1 for la=",partition)
        print("Dimension of space=",len(U_mats))
    U_mat = U_mats[0]
    U_form = matrix(R_xy_lambda,U_mat)
    
    #check symmetric property
    symmetric = bilinear_form(x,y,U_form) == bilinear_form(y,x,U_form)
    
    #check G-invariance property
    G_invariant = all(bilinear_form(rho(g)*x,rho(g)*y,U_form) == bilinear_form(x,y,U_form) for g in G)
    
    #check bilinear property. ISSUE: lambda_^q is a power of the ring generator, i.e. doesn't simplify.
    first_arg = bilinear_form(lambda_*x,y,U_form) == lambda_*bilinear_form(x,y,U_form)
    second_arg = bilinear_form(x,lambda_*y,U_form) == lambda_*bilinear_form(x,y,U_form) #need to amend for conjugation
    bilinear = first_arg and second_arg

    return symmetric and G_invariant and bilinear
```

If different matrices give a different form, that's fine, but isn't the space usually one dim'l? So whatever set of matrices is used to generate the linear system, the result should be the linear span of one matrix. In the above examples, they're not scalar multiples of each other.

tscrim:
That's not how linear algebra works. I just need to change my basis for the rerpreaentation vector space. This changes all of the matrices for the $S_n$ elements too, but I get a completely different matrix representing the same bilinear form.

The way the two implementations construct the representations might be slightly different (i.e., have different matrices). However, I think the problem comes from what you noted, the matrix's base ring is wrong, which makes the `conjugate()` wrong. That's my guess without testing things.

jacksonwalters:
The problem I was seeing was that `_unitary_change_basis_matrix` is outputting `[[1 3 4],[3 1 3],[4 3 1]]` rather than `[[1 4 4],[4 1 4],[4 4 1]]`. Yes, one can conjugate to change bases, but when we solve for the invariant symmetric bilinear form, there is a system of linear equations that needs to be solved. That solution space has a dimension, which is usually one. If, for example, the matrix `U=[[1 4 4],[4 1 4],[4 4 1]]` is in `null_space.basis()`, then any other solution should be a scalar multiple of it. `[[1 3 4],[3 1 3],[4 3 1]]` is not a scalar multiple of `[[1 4 4],[4 1 4],[4 4 1]]`, and checking shows it's not invariant. Perhaps I'm missing something.

In any case, you're probably right that it's conjugate. It's definitely to do with the line

```self._specht = SymmetricGroupRepresentation(partition, 'specht', ring=parent._ring)```

which I am going to try replacing with

```self._specht = Permutations(self._n).algebra(parent._ring).specht_module(self._partition)```

temporarily just to see if I can get tests passing, and match what's happening with the code in my notebook.

**Comment by jacksonwalters** [comment]
Okay, so all tests (other than one of the flaky Meson / Conda ones) are passing. 

I would like to do a separate PR addressing the base_ring issues for `symmetric_group_representations.py`. It'd be nice to have the representations defined over a minimal number field, and address the above.

File: src/sage/combinat/symmetric_group_representations.py [hunk 16]
```diff
@@ -999,6 +1001,170 @@ def _repr_(self):
...
+            [6 0]
+            [0 1]
+            sage: all(A * A.H == 1 for A in [unitary_specht._representation_matrix_uncached(g)
+            ....:                            for g in Permutations(4)])
+            True
+        """
+        assert self._ring.characteristic() > 0
+        rho = self._specht.representation_matrix
+        A = self._unitary_change_basis_matrix
+        return A.H * rho(permutation) * A.H.inverse()
```
tscrim:
```suggestion
        A = self._unitary_change_basis_matrix
        return A * rho(permutation) * A.inverse()
```

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -999,6 +1001,170 @@ def _repr_(self):
...
+                for j in range(d_rho):
+                    linear_expression = equation_matrix[i, j]
+                    row = [linear_expression.coefficient(u) for u in U_vars]
+                    augmented_system.append(row)
+            return augmented_system
+
+        total_system = sum((augmented_matrix(g) for g in G), [])
+        null_space = matrix(F, total_system).right_kernel()
+        U = matrix(F, d_rho, d_rho, null_space.basis()[0])
+        return U.hermitian_decomposition()
```
tscrim:
```suggestion
        return U.hermitian_decomposition().H
```

tscrim (changes requested):
Moving the conjugate-transpose part into the cached portion.

tscrim:
> I would like to do a separate PR addressing the base_ring issues for `symmetric_group_representations.py`. It'd be nice to have the representations defined over a minimal number field, and address the above concern.

Yes, that should be done on a separate PR. We might consider having some care as computing a minimal (or at least large enough) number field might dominate the computation (which it shouldn't).

jacksonwalters:
@tscrim I've rebased this branch on [#39200](https://github.com/sagemath/sage/pull/39200). It has been updated to call `U.cholesky(extended=True)`.

dimpase:
please rebase over the latest beta (or merge it in)

jacksonwalters:
> please rebase over the latest beta (or merge it in)

I attempted to rebase and got merge conflicts:

```
jacksonwalters@jaxmacbookair sage % git fetch origin
jacksonwalters@jaxmacbookair sage % git checkout develop
Already on 'develop'
Your branch is up to date with 'origin/develop'.
jacksonwalters@jaxmacbookair sage % git pull origin develop
From https://github.com/jacksonwalters/sage
 * branch                    develop    -> FETCH_HEAD
Already up to date.
jacksonwalters@jaxmacbookair sage % git checkout unitary_dft_symmetric_group
Updating files: 100% (5535/5535), done.
Switched to branch 'unitary_dft_symmetric_group'
Your branch is up to date with 'origin/unitary_dft_symmetric_group'.
jacksonwalters@jaxmacbookair sage % git rebase develop
Auto-merging src/sage/groups/artin.py
Auto-merging src/sage/groups/cubic_braid.py
Auto-merging src/sage/groups/finitely_presented.py
Auto-merging src/sage/groups/free_group.py
CONFLICT (content): Merge conflict in src/sage/groups/free_group.py
Auto-merging src/sage/groups/raag.py
Auto-merging src/sage/libs/gap/element.pyx
error: could not apply e9222c47428... Added gap groups for free and f.p groups, added __reduce__ methods
hint: Resolve all conflicts manually, mark them as resolved with
hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
hint: You can instead skip this commit: run "git rebase --skip".
hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
hint: Disable this message with "git config advice.mergeConflict false"
Could not apply e9222c47428... Added gap groups for free and f.p groups, added __reduce__ methods for these groups and also for artin, cubicbraid and raag
```

dimpase:
merge conflicts are expected during rebase.

you need to edit files marked as conflicted, to select appropriate chunks beween
```
>>><
===
<<<<
```
markers, then ` git add` each of them, and then run `git rebase --continue`

at every moment `git status` will give you tips what to do.

tscrim:
@jacksonwalters I would also rebase #39200, and then rebase this on top of that branch.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2066,8 +2066,74 @@ def dft(self, form=None, mult='l2r'):
...
+            raise NotImplementedError("not implemented when p|n!; dimension of invariant forms may be greater than one")
+        q = F.order().sqrt()
+
+        def conj_square_root(u):
+            if not u:
+                return F.zero()
+            z = F.multiplicative_generator()
+            k = u.log(z)
+            if k % (q+1) != 0:
+                raise ValueError(f"Unable to factor as {u} is not in base field GF({q})")
```
tscrim:
```suggestion
                raise ValueError(f"unable to factor as {u} is not in base field GF({q})")
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2066,8 +2066,74 @@ def dft(self, form=None, mult='l2r'):
...
+            primes_needed = {factor for d in diag for factor, _ in d.squarefree_part().factor()}
+            names = [f"sqrt{factor}" for factor in primes_needed]
+            x = PolynomialRing(QQ, 'x').gen()
+            K = NumberField([x**2 - d for d in primes_needed], names=names)
+            sqrt_diag_inv = diagonal_matrix([~sqrt(K(d)) for d in diag])
+            return sqrt_diag_inv * dft_matrix
+
+        # positive characteristic case
+        assert F.characteristic() > 0
+        if not (F.is_field() and F.is_finite() and F.order().is_square()):
```
tscrim, line 2188: Please add some tests checking these conditions on the field.

tscrim (changes requested):
Please also add to the documentation of `dft` about all of the `form` inputs (I should have caught this on the modular DFT PR...).

tscrim:
Also, don't forget to rebase this over #39200.

src/sage/combinat/symmetric_group_algebra.py
jacksonwalters: This is addressed in commit [b33cefade43e651454b2c0fbb01adee86cff94a7](https://github.com/jacksonwalters/sage/commit/b33cefade43e651454b2c0fbb01adee86cff94a7). I added four tests. 

Note the assert will always pass, since if a ring is not characteristic zero, it must be positive characteristic.

jacksonwalters:
> Please also add to the documentation of `dft` about all of the `form` inputs (I should have caught this on the modular DFT PR...).

This is addressed in commit [b33cefade43e651454b2c0fbb01adee86cff94a7](https://github.com/jacksonwalters/sage/commit/b33cefade43e651454b2c0fbb01adee86cff94a7).

I'm not sure how much detail is required re: how the methods differ, and what they're actually computing.

jacksonwalters:
@tscrim Okay, commits are squashed. I think this is in good shape now.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2035,6 +2035,15 @@ def dft(self, form=None, mult='l2r'):
...
           the seminormal basis.
 
+        - ``form`` -- string (default: `None`); if set to None, we use the
+            modular DFT if p|n!, otherwise we use the seminormal basis.
+            If set to `seminormal`, we use the seminormal basis. If set to
+            `modular`, we use the modular DFT which uses the Peirce decomposition
+            of the group algebra into blocks with central orthogonal idempotents.
+            If set to `unitary`, we use the unitary DFT which computes the extended
+            Cholesky decomposition of an S_n-invariant symmetric bilinear form as
+            a change-of-basis matrix over a finite field of square order.
```
tscrim:
```suggestion
        - ``form`` -- string (default: ``"modular"`` if `p |n!` else ``"seminormal"``);
          one of the following:

          * `"seminormal"` -- use the seminormal basis
          * ``"modular"`` -- use the modular DFT, which uses the Peirce decomposition
            of the group algebra into blocks with central orthogonal idempotents
          * ``"unitary"`` -- use the unitary DFT. which computes the extended
            Cholesky decomposition of an `S_n`-invariant symmetric bilinear form as
            a change-of-basis matrix (for positive characteristics, it must be
            a finite field of square order)
```

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -999,6 +1001,170 @@ def _repr_(self):
         """
         return "Specht representations of the symmetric group of order %s! over %s" % (self._n, self._ring)
 
+# #### Unitary Representation ###############################################
+
+
+class UnitaryRepresentation(SymmetricGroupRepresentation_generic_class):
```
tscrim, line 1043: Please add documentation here describing the mathematics.

tscrim (changes requested):
Thank you. Three things left. Two are comments below and the other is add the examples `QS3.dft(form="unitary")` and `GF4S3.dft(form="unitary")` (note I changed the field to `GF(4)` here).

tscrim:
> `GF4S3.dft(form="unitary")` (note I changed the field to `GF(4)` here).

Ah, right, it needs to be in the non-modular case. Sorry. I wanted an example showing the unitary versus the usual case for some positive char field.

jacksonwalters:
> > `GF4S3.dft(form="unitary")` (note I changed the field to `GF(4)` here).
> 
> Ah, right, it needs to be in the non-modular case. Sorry. I wanted an example showing the unitary versus the usual case for some positive char field.

Definitely, I'm surprised I had forgotten to add those. I added two, one for `q=5` and `q=7` when `n=3`.

src/sage/combinat/symmetric_group_representations.py
jacksonwalters: I added some documentation for `__init__`, `_unitary_change_basis_matrix`, and `_representation_matrix_uncached`. I try to capture the three key computations: 1) computing a G-invariant symmetric bilinear form `U` 2) computing its extended Cholesky decomposition `A` 3) check basis-change by `A` is unitary.

tscrim: Please move the `__init__` docstring to the class-level. The `__init__` docstring should just be "Initialize \`\`self\`\`" with the examples. The class-level doesn't need examples, but it should have the mathematics since it is public readable.

jacksonwalters: Makes sense. I've moved the docstring from `__init__` up to the class level.

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -999,6 +1001,186 @@ def _repr_(self):
...
+    r"""
+    This is a unitary representation of the symmetric group.
+    In characteristic zero, this is the same as the orthogonal representation since
```
tscrim:
```suggestion
    r"""
    A unitary representation of the symmetric group.

    In characteristic zero, this is the same as the orthogonal representation since
```
Thanks. Here is one formatting change.

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -999,6 +1001,186 @@ def _repr_(self):
...
+        null_space = matrix(F, total_system).right_kernel()
+        U = matrix(F, d_rho, d_rho, null_space.basis()[0])
+        return U.cholesky(extended=True).H
+
+    def _representation_matrix_uncached(self, permutation):
+        r"""
+        Return the matrix representing the ``permutation`` in this
+        irreducible representation. We use the unitary change of basis
+        matrix to conjugate the representation matrix. These are unitary
+        since `(A * rho(g) * A.inverse()).H * (A * rho(g) * A.inverse()) = A.inverse().H * rho(g).H * A.H * A * rho(g) * A.inverse() = A.inverse().H * rho(g).T * U * rho(g) * A.inverse() = A.inverse().H * U * A.inverse() = 1`.
```
tscrim:
```suggestion
        Return the matrix representing the ``permutation`` in this
        irreducible representation.

        We use the unitary change of basis matrix to conjugate
        the representation matrix. These are unitary since

        .. MATH::

            (A rho(g) A^{-1})^* \cdot (A rho(g) A^{-1})
              = (A^{-1})^* rho(g)^* A^* A rho(g) A^{-1}
              = (A^{-1})^*H rho(g)^T U rho(g) A^{-1}
              = (A^{-1})^* U A^{-1} = 1.
```

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -999,6 +1001,186 @@ def _repr_(self):
...
+        """
+        ret = self._representation_matrix_uncached(permutation)
+        ret.set_immutable()
+        return ret
+
+    @lazy_attribute
+    def _unitary_change_basis_matrix(self):
+        """
+        Compute the change of basis matrix. We first compute a G-invariant symmetric bilinear form. This
+        yields a solution to the equation `rho(g).T * U * rho(g) = U` for all `g` in `G`. We can
```
tscrim:
```suggestion
        Compute the change of basis matrix.

        We first compute a `G`-invariant symmetric bilinear form. This yields a solution `U`
        to the equation `\rho(g)^T U \rho(g) = U` for all `g` in `G`. We can
```

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -999,6 +1001,186 @@ def _repr_(self):
...
+
+class UnitaryRepresentation(SymmetricGroupRepresentation_generic_class):
+    r"""
+    This is a unitary representation of the symmetric group.
+    In characteristic zero, this is the same as the orthogonal representation since
+    they are defined over the reals. In positive characteristic, we are able to construct
+    representations when the field is finite, has square order, and the characteristic does not divide
+    the order of the group. In this case, we construct the representation by first
+    constructing the orthogonal representation and then taking the extended Cholesky
+    decomposition of the unique solution to the equation `rho(g).T * U * rho(g) = U` for all `g` in `G`.
```
tscrim:
```suggestion
    constructing the orthogonal representation `\rho` and then taking the extended
    Cholesky decomposition of the unique solution `U` to the equation
    `\rho(g)^T U \rho(g) = U` for all `g` in `G`.
```

tscrim (approved):
Thank you. Some additional formatting things for readability.

jacksonwalters:
> Thank you. Some additional formatting things for readability.

Committed, squashed.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2035,28 +2035,57 @@ def dft(self, form=None, mult='l2r'):
           (:meth:`antipode`) of the seminormal basis instead of
           the seminormal basis.
 
+        - ``form`` -- string (default: ``"modular"`` if `p |n!` else ``"seminormal"``);
```
tscrim:
```suggestion
        - ``form`` -- string (default: ``"modular"`` if `p|n!` else ``"seminormal"``);
```

tscrim (changes requested):
Okay, one last mystery to solve. Why do you get a different ordering of the columns than the bots. I don't see where there could be something that produces different orders. Tests are passing locally for you?

src/sage/combinat/symmetric_group_algebra.py
tscrim: Or the other way
```suggestion
        - ``form`` -- string (default: ``"modular"`` if `p | n!` else ``"seminormal"``);
```

jacksonwalters:
> Okay, one last mystery to solve. Why do you get a different ordering of the columns than the bots. I don't see where there could be something that produces different orders. Tests are passing locally for you?

Yes, was looking into that. I didn't notice it was the column ordering; I thought they were totally different. I checked that both versions are unitary, so they're passing in that sense. I also compared my [local code](https://github.com/jacksonwalters/dft-finite-groups/blob/main/symmetric_group/unitary_dft_over_finite_fields.ipynb) closely to the Sage code - they seem to match line for line.

jacksonwalters:
Checking to see if the multiplicative generators match, and if the seminormal DFTs match. These are really the only two things it could be. The rest of the computation is just computing conjugate square roots of the diagonal and multiplying by the inverse.

jacksonwalters:
@tscrim Mystery solved. The multiplicative generators were the same, but I was using `G.algebra(F)` in my local code. This causes the `.dft()` to be different. I switched to `SymmetricGroupAlgebra(F,n)` and they line up now.

tscrim:
> @tscrim Mystery solved. The multiplicative generators were the same, but I was using `SymmetricGroup(n).algebra(F)` in my local code. This causes the `.dft()` to be different. I switched to `SymmetricGroupAlgebra(F,n)` and they line up now.

I see. That makes sense. The groups (`Permutations(n)` and`SymmetricGroup(n)`) and how they are ordered/go through different things/etc. are slightly different, leading to the slight change in output.

jacksonwalters:
It looks like tests are passing now. Let me know if there's anything else I need to do.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2035,22 +2035,60 @@ def dft(self, form=None, mult='l2r'):
           (:meth:`antipode`) of the seminormal basis instead of
           the seminormal basis.
 
-        EXAMPLES::
+        - ``form`` -- string (default: ``"modular"`` if `p|n!` else ``"seminormal"``);
+          one of the following:
 
-            sage: QS3 = SymmetricGroupAlgebra(QQ, 3)
-            sage: QS3.dft()
+          * `"seminormal"` -- use the seminormal basis
```
tscrim:
```suggestion
          * ``"seminormal"`` -- use the seminormal basis
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2035,22 +2035,60 @@ def dft(self, form=None, mult='l2r'):
...
+            [       1        1        1        1        1        1]
+            [2*z2 + 4   z2 + 2 3*z2 + 1 4*z2 + 3 4*z2 + 3   z2 + 2]
+            [       0        2        0        2        3        3]
+            [       0   z2 + 1        0 4*z2 + 4   z2 + 1 4*z2 + 4]
+            [2*z2 + 4 4*z2 + 3 2*z2 + 4 4*z2 + 3 4*z2 + 3 4*z2 + 3]
+            [       1        4        4        1        1        4]
+            sage: U*U.H == 1
+            True
+
         Over fields of characteristic `p > 0` such that `p \mid n!`, we use the
```
tscrim:
```suggestion
        Over fields of characteristic `p > 0` such that `p|n!`, we use the
```
Just to make it uniform. Or you can change the one you added.

tscrim (changes requested):
Sorry, I know my comments are coming in small batches over time rather than just being done with it all. This should be the last round though.

Two minor formatting things, plus you should add the `"unitary"` input to the `SymmetricGroupRepresentations` docstring with an example.

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2066,8 +2104,103 @@ def dft(self, form=None, mult='l2r'):
...
+            z = F.multiplicative_generator()
+            k = u.log(z)
+            if k % (q+1) != 0:
+                raise ValueError(f"unable to factor as {u} is not in base field GF({q})")
+            return z ** ((k//(q+1)) % (q-1))
+
+        dft_matrix = self.dft()
+        sign_diag = (dft_matrix * dft_matrix.H).diagonal()
+        conj_sqrt_diag_inv = diagonal_matrix([~conj_square_root(d) for d in sign_diag])
+        return conj_sqrt_diag_inv * dft_matrix
```
tscrim:
```suggestion
        n = dft_matrix.nrows()
        for i in range(n):
            d = sum(dft_matrix[i, j] * dft_matrix[i, j].conjugate() for j in range(n))
            dft_matrix[i] *= ~conj_square_root(d)
        return dft_matrix
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2066,8 +2104,103 @@ def dft(self, form=None, mult='l2r'):
...
+        """
+        from sage.matrix.special import diagonal_matrix
+        F = self.base_ring()
+        G = self.group()
+
+        if F.characteristic() == 0:
+            from sage.misc.functional import sqrt
+            from sage.rings.number_field.number_field import NumberField
+            dft_matrix = self.dft()
+            diag = (dft_matrix * dft_matrix.H).diagonal()
```
tscrim:
```suggestion
            n = dft_matrix.nrows()
            diag = [sum(dft_matrix[i, j] * dft_matrix[i, j].conjugate() for j in range(n))
                    for i in range(n)]
```

src/sage/combinat/symmetric_group_algebra.py
```diff
@@ -2066,8 +2104,103 @@ def dft(self, form=None, mult='l2r'):
...
+            primes_needed = {factor for d in diag for factor, _ in d.squarefree_part().factor()}
+            names = [f"sqrt{factor}" for factor in primes_needed]
+            x = PolynomialRing(QQ, 'x').gen()
+            K = NumberField([x**2 - d for d in primes_needed], names=names)
+            sqrt_diag_inv = diagonal_matrix([~sqrt(K(d)) for d in diag])
+            return sqrt_diag_inv * dft_matrix
```
tscrim:
```suggestion
            dft_matrix = dft_matrix.change_ring(K)
            for i, d in enumerate(diag):
                dft_matrix[i] *= ~sqrt(K(d))
            return dft_matrix
```

tscrim (changes requested):
Sorry, now I am doing some obvious optimizations. This will be it from me. Anything else I can do on a followup PR.

jacksonwalters:
> Sorry, I know my comments are coming in small batches over time rather than just being done with it all. This should be the last round though.
> 
> Two minor formatting things, plus you should add the `"unitary"` input to the `SymmetricGroupRepresentations` docstring with an example.

No worries. Committed. These are all very helpful additions. Okay, I've added the `"unitary"` input options for both the`SymmetricGroupRepresentation` and `SymmetricGroupRepresentations` constructors with examples.

jacksonwalters:
> Sorry, now I am doing some obvious optimizations. This will be it from me. Anything else I can do on a followup PR.

Sounds good. It seems to be in a usable state now, and the documentation is much better. Agreed re: the followup PR.

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -238,6 +257,23 @@ def SymmetricGroupRepresentations(n, implementation='specht', ring=None,
         [ 0  0  0 -1  1]
         [ 0  1 -1 -1  1]
         [ 0  1  0 -1  1]
+
```
tscrim:
```suggestion

```
Trailing whitespace.

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -131,6 +132,23 @@ def SymmetricGroupRepresentation(partition, implementation='specht',
         [ 0  1  0 -1  1]
         sage: spc.verify_representation()
         True
+
```
tscrim:
```suggestion

```
Trailing whitespace.

tscrim (approved):
Thank you. Just two last details for the linter.

The bottleneck is now in computing the seminormal form, which seems like a harder problem to optimize (with a fair bit of the time spent in multiplying elements in the SGA).

tscrim:
And some failing tests

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -239,6 +258,23 @@ def SymmetricGroupRepresentations(n, implementation='specht', ring=None,
         [ 0  1 -1 -1  1]
         [ 0  1  0 -1  1]
 
+    The unitary representation.
+
+    ::
+
+        sage: unitary = SymmetricGroupRepresentations(3, "unitary"); unitary
+        Unitary representations of the symmetric group of order 3! over Symbolic Ring
+        sage: unitary_GF49 = SymmetricGroupRepresentations(3, "unitary", ring=GF(7**2)); unitary_GF49
```
tscrim:
```suggestion
        sage: unitary_GF49 = SymmetricGroupRepresentations(4, "unitary", ring=GF(7**2)); unitary_GF49
```

src/sage/combinat/symmetric_group_representations.py
```diff
@@ -239,6 +258,23 @@ def SymmetricGroupRepresentations(n, implementation='specht', ring=None,
...
+        Unitary representations of the symmetric group of order 3! over Finite Field in z2 of size 7^2
+        sage: unitary_GF49([3,1])([2,1,3,4])
+        [6 0 0]
+        [0 1 0]
+        [0 0 1]
+        sage: unitary_GF49([3,1])([3,2,1,4]))
```
tscrim:
```suggestion
        sage: unitary_GF49([3,1])([3,2,1,4])
```

tscrim (commented):
This should address the doctests.

jacksonwalters:
> This should address the doctests.

Okay, those minor fixes seem to have doctests passing now.

tscrim:
This is a positive review from me.

PR #38459, opened 2024-07-31
Fix coercion error on constant arguments to HypJac point construction

nbruin:
Fixes error reported by Guninski in https://groups.google.com/g/sage-devel/c/mLbbFh8E7gU/m/JPlhZtTLAwAJ

```
K.<u>=QQ[]
C=HyperellipticCurve(u^5+u+5^2-2);
J=C.jacobian();
print("J=",J);
P=J(1,5); #error here
```

The problem is that the point constructor expects polynomials describing a mumford representation. When constants are supplied, these need to be coerced into a polynomial ring. Originally this was hardcoded to be in `x`. We change it here to be the same ring as where the hyperelliptic polynomials defining C live (which is consistent with the arithmetic that happens with them)

### :memo: Checklist

- [ X] The title is concise and informative.
- [ X] The description explains in detail what this PR is about.
- [ X] I have linked a relevant issue or discussion.


vincentmacri:
Can you add a doctest that makes sure the bug is fixed?

From https://doc.sagemath.org/html/en/developer/review.html:
> Bugfixes: If the PR contains a bugfix, does it add a doctest illustrating that the bug has been fixed? This new doctest should contain the issue or PR number, for example See :issue:`12345`.


Other than that, looks good.

nbruin:
someone should put doctests on this

vincentmacri:
> (I was surprised a but labelled this with `needs review`. How would it know the pull request is ready for review?)

I think the bot assumes all non-draft PRs are ready for review. If you aren't ready for a review you can set the PR to a draft PR in GitHub.

vincentmacri:
Another note: [`hyperelliptic_generic`](https://github.com/sagemath/sage/blob/develop/src/sage/schemes/hyperelliptic_curves/hyperelliptic_generic.py) does some stuff with "printing names", the names of the variables used to represent the curve in output doesn't necessarily match the names of the variables in the ring of the defining polynomial.

This is why when you create a hyperelliptic curve over `u`, the curve is displayed as `y^2 = [polynomial in x]` unless the user specifies variable names. For consistency, should the output here use the same variable names as Sage outputs for the curve equation?

nbruin:
Changing the way the printing names are handled would be an incompatible change (but only mildly so), so the bar for making such a change is a little higher. One way would be to just extract the name of the hyperelliptic polynomials to use as default printing name for the first variable. But then an example like
```
R.<y> = QQ[]
C = HyperellipticCurve(y^6+y+1)
```
would print problematically: the name of the variable `y` as printed in `y^2+h*y=f` isn't really necessarily specified, but if the variable on the hyperelliptic quotient is already named `y`, you'd get a clash by setting it to `y`. I guess a silly work-around would be to just change that default to, say, `z` if the first default name ends up being `y` already. I'm not sure this kind of workaround for default names has a precedent in sage.

vincentmacri:
> Changing the way the printing names are handled would be an incompatible change (but only mildly so), so the bar for making such a change is a little higher. One way would be to just extract the name of the hyperelliptic polynomials to use as default printing name for the first variable. But then an example like
> 
> ```
> R.<y> = QQ[]
> C = HyperellipticCurve(y^6+y+1)
> ```
> 
> would print problematically: the name of the variable `y` as printed in `y^2+h*y=f` isn't really necessarily specified, but if the variable on the hyperelliptic quotient is already named `y`, you'd get a clash by setting it to `y`. I guess a silly work-around would be to just change that default to, say, `z` if the first default name ends up being `y` already. I'm not sure this kind of workaround for default names has a precedent in sage.

I agree with all that but I think it's out of scope of this PR. To clarify, I meant that should the code you're modifying here use the ring that the polynomial defining the curve is over (which is what you have right now in your PR), or should it use the variable names that Sage outputs to describe the curve? I think the way Sage is handling variable names for hyperelliptic curves is a bit weird anyway, so I'm fine with whatever you think makes sense.

In other words, my question is: should the output of a Mumford polynomial for a Jacobian element of a curve defined over `u` look like `(x - 1, y - 5)` or `(u - 1, y - 5)`, and does the answer to that change what you want to do in this PR? After writing this I think the answer to the second question is "no, that should be handled by the `_repr_` function for hyperelliptic Jacobian elements", but I'll leave this mini-rant for future reference if someone things the answer to the first question is `(u - 1, y - 5)` and wants to make a PR changing it.

src/sage/schemes/hyperelliptic_curves/jacobian_homset.py
```diff
@@ -113,28 +113,28 @@ def __call__(self, P):
             (x^2 + 2*x + 2, y + 2*x + 1)
```
vincentmacri, line 114: Add a test making sure the bug is fixed.

vincentmacri (changes requested):
Looks good other than needing a doctest that makes sure the bug is fixed. I have no preference on if this should be in the `EXAMPLES` block or if a `TESTS` block should be added.

nbruin:
> In other words, my question is: should the output of a Mumford polynomial for a Jacobian element of a curve defined over `u` look like `(x - 1, y - 5)` or `(u - 1, y - 5)`, and does the answer to that change what you want to do in this PR? After writing this I think the answer to the second question is "no, that should be handled by the `_repr_` function for hyperelliptic Jacobian elements", but I'll leave this mini-rant for future reference if someone things the answer to the first question is `(u - 1, y - 5)` and wants to make a PR changing it.

I think the design decision is already taken:
```
sage: R.<y>=QQ[]
sage: C=HyperellipticCurve(y^6+y+1)
sage: J=C.jacobian()
sage: V=J(y,y^0)
sage: V[0]
y
sage: V[1]
1
sage: (3*V)[0]
Returning ambiguous form of degree genus+1.
y^3 - 1/64*y + 1/8
sage: (3*V)[1]
Returning ambiguous form of degree genus+1.
1/8*y^2 - 1/2*y - 1
```
The defining polynomials of a mumford representation really live in the parent of the defining hyperelliptic polynomials. I suspect no mathematically important part ever refers to the ring in which the "Y" of "Y^2+h*Y=f" lives. So `J(QQ)._printing_ring` seems to be exactly that: a ring purely used to print. The names used there can be set when `HyperellipticCurve` is defined. It would make sense to borrow names from the relevant rings if nothing is given, with the caveat that we don't have a place to borrow a good name for `y` from. Note that the current naming already can lead to very confusing situations:
```
sage: K.<x>=FunctionField(QQ)
sage: P.<X>=K[]
sage: C=HyperellipticCurve(X^6+X+x^2)
sage: J=C.jacobian()
sage: D=J(X,P(x))
sage: 3*D
Returning ambiguous form of degree genus+1.
(x^3 - 1/64/x^6*x + 1/8/x^4, y - 1/8/x^3*x^2 + 1/2/x*x + x)
sage: (3*D)[0]
Returning ambiguous form of degree genus+1.
X^3 - 1/64/x^6*X + 1/8/x^4
sage: (3*D)[1]
Returning ambiguous form of degree genus+1.
1/8/x^3*X^2 - 1/2/x*X - x
```
As you can see, the default choices of print names really mess up the original relatively unambiguous naming introduced by the user.

vincentmacri:
> As you can see, the default choices of print names really mess up the original relatively unambiguous naming introduced by the user.

Agreed, but fixing it is out of scope for this PR I think. The only thing I think this PR needs before merging is a doctest covering the bug fix.

src/sage/schemes/hyperelliptic_curves/jacobian_homset.py
```diff
@@ -111,30 +111,39 @@ def __call__(self, P):
             (x + 2, y)
             sage: D1 + D2
             (x^2 + 2*x + 2, y + 2*x + 1)
+
+        TESTS:
+
+        Test :issue:`38459`::
+
+            sage: C=HyperellipticCurve(u^5-1)
```
vincentmacri:
```suggestion
            sage: C = HyperellipticCurve(u^5 - 1)
```

src/sage/schemes/hyperelliptic_curves/jacobian_homset.py
```diff
@@ -111,30 +111,39 @@ def __call__(self, P):
...
+            sage: J=C.jacobian()
```
vincentmacri:
```suggestion
            sage: J = C.jacobian()
```

src/sage/schemes/hyperelliptic_curves/jacobian_homset.py
```diff
@@ -111,30 +111,39 @@ def __call__(self, P):
...
+            sage: J(u-1,0)
```
vincentmacri:
```suggestion
            sage: J(u - 1, 0)
```

src/sage/schemes/hyperelliptic_curves/jacobian_homset.py
```diff
@@ -111,30 +111,39 @@ def __call__(self, P):
             (x + 2, y)
             sage: D1 + D2
             (x^2 + 2*x + 2, y + 2*x + 1)
+
+        TESTS:
+
+        Test :issue:`38459`::
+
```
vincentmacri, line 119:
```suggestion

            sage: K.<u> = QQ[]
```

vincentmacri (changes requested):
Formatting and fix the broken test.

src/sage/schemes/hyperelliptic_curves/jacobian_homset.py
```diff
@@ -111,30 +111,41 @@ def __call__(self, P):
...
+            sage: K.<u> = QQ[]
+            sage: K.<u> = QQ[]
```
vincentmacri:
```suggestion
```

This line got added twice.

vincentmacri (changes requested):

vincentmacri:
The test failures seem relevant to the changed code.

My guess from looking at the logs is that the `TypeError: Unable to coerce [...] to a rational` errors are the main problem, and the `NameError` failures are caused by the `TypeError`s.

nbruin:
Yes, the value ring of a point set does not need to equal the base ring of the scheme, so the polynomial rings used to describe divisors representing points in them should be base changed too. One would expect that to be checked in homset to begin with. Luckily we do have tests for it elsewhere.

vincentmacri:
The test failures now look to be unrelated. It would be nice if you could add test cases covering lines 139 and 144 but since this is a bug fix PR and those lines didn't have test cases covering them before (and the bug fix isn't related to those test cases) I'm good with this as it is now.

vincentmacri (approved):

vincentmacri (approved):
I'd set the label to "positive review" but I don't think I have the permission for it.

nbruin:
> I'd set the label to "positive review" but I don't think I have the permission for it.
set positive review on @vincentmacri 's behalf

PR #38479, opened 2024-08-06
Choose special name to avoid '@c' by singular's absFactorize

### LINKED ISSUE #33327: variable name '@c' is not alphanumeric
<div id="comment:0"></div>

Multivariate polynomial rings over `QQbar`
with a variable named `c` have problems.

Both of the following

```
sage: A.<a, c> = QQbar[]
sage: factor(a)

sage: A.<c> = PolynomialRing(QQbar, 1)
sage: factor(c)
```
cause this value error:

```
ValueError: variable name '@c' is not alphanumeric
```

No such problem with univariate polynomial rings
or multivariate ones with other variable names:

```
sage: A.<a> = QQbar[]
sage: factor(a)
a
sage: A.<c> = QQbar[]
sage: factor(c)
c
sage: A.<a, b> = QQbar[]
sage: factor(a)
a
```
Multivariate polynomial rings over `QQ`,
even using variable name `c`, are fine too:

```
sage: A.<a, c> = QQ[]
sage: factor(a)
a
sage: A.<c> = PolynomialRing(QQ, 1)
sage: factor(c)
c
```

---

Initial report by Anne Vaugon:

```
sage: A.<a, b, c, x, y, S> = QQbar[]
sage: p = (a - x)^2 + y^2 - b^2
sage: q = x^2 + y^2 - c^2
sage: r = a * y - 2 * S
sage: r1 = p.resultant(q, x)
sage: r2 = r.resultant(r1, y)
sage: factor(r2 - a^2 * 16 * S^2)
...
ValueError: variable name '@c' is not alphanumeric
```

CC:  @slel

Component: **interfaces**

Keywords: **singular**

_Issue created by migration from https://trac.sagemath.org/ticket/33327_

nbruin:
Fixes #33327

### :memo: Checklist

### DIFF
--- a/src/sage/rings/qqbar.py
+++ b/src/sage/rings/qqbar.py
@@ -981,7 +981,7 @@ class AlgebraicField_common(sage.rings.abc.AlgebraicField_common):
         else:
             norm_f = numfield_f
 
-        R = norm_f._singular_().absFactorize('"_SAGE_ALGEBRAIC_"')
+        R = norm_f._singular_().absFactorize('"SAGE_ALGEBRAIC"')
 
         singular.setring(R)
         L = singular('absolute_factors')
--- a/src/sage/rings/qqbar.py
+++ b/src/sage/rings/qqbar.py
@@ -927,6 +927,15 @@ class AlgebraicField_common(sage.rings.abc.AlgebraicField_common):
             sage: # needs sage.libs.singular
             sage: AA['x','y'](1).factor()   # indirect doctest
             1
+
+        Test :issue:`#33327`::
+
+            sage: # needs sage.libs.singular
+            sage: S.<a,c> = QQbar[]
+            sage: p = a^2 + 7*c^2
+            sage: factor(p)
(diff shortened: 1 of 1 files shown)


dimpase:
Are you sure `"SAGE_ALGEBRAIC"` won't pop up somewhere back in Sage data?
Otherwise looks good to me.

dimpase (approved):

nbruin:
No, it's obviously very doable to create examples that generate a `SAGE_ALGEBRAIC` identifier name, and by having introduced the name in the codebase now certainly increases encountering it somewhere. However, the name should not escape from the internal qqbar routine here, so I think it's a definite improvement. If we end up running into examples with this name without people feeling they had it coming, we can write an adaptive name generation routine. The ring gets flattened, so getting an overview of the names involved should be quite doable. Presently I think it would be overengineering to do that.

dimpase:
OK, fine. 
I tried this kind of workaround, but I forgot an extra pair of `' '` and thought it's trickier than that. :-)

Issue #38480, opened 2024-08-07
Docstring of __call__ does not appear in the HTML documentation?


user202729:
Example:

* the following documentation https://github.com/sagemath/sage/blob/develop/src/sage/lfunctions/dokchitser.py#L490-L507 is not included in https://doc.sagemath.org/html/en/reference/lfunctions/sage/lfunctions/dokchitser.html
* the following documentation https://github.com/sagemath/sage/blob/develop/src/sage/rings/valuation/valuation_space.py#L211-L218 is not included in https://doc.sagemath.org/html/en/reference/valuations/sage/rings/valuation/valuation_space.html

thecaligarmo:
This is by design. No private methods (methods which start with a `_`) are listed in the documentation. As far as I'm aware there is no movement to alter this behavior globally. Maybe @mkoeppe might have more details on this.

There are two alternatives:
1. If you have a local version of sage, you can have the documentation show all methods (even private ones) by setting `SAGE_DOC_UNDERSCORE` as an environmental variable. You would then need to `make` sage again to get the updated documentation (make sure `SAGE_DOC_UNDERSCORE` appears as an environmental variable).
2. You can add `.. automethod:: __call__` for each class you'd like the `__call__` method to appear in the documentation. (If you have a list and would prefer someone else handle it, just add the list as a comment and I can take care of it)

PR #38482, opened 2024-08-07
Dokchitser: Pass internal parameter over properly

user202729:
Previously the `c` parameter was completely unused. I assume this is what it should be used for.

The documentation is copied from the PARI code.

I don't think there's a way to actually test it, but there is already code in the docstring that tries to pass $c ≠ 1$.

The documentation update is invisible because of https://github.com/sagemath/sage/issues/38480 .

### :memo: Checklist

### DIFF
--- a/src/sage/lfunctions/dokchitser.py
+++ b/src/sage/lfunctions/dokchitser.py
@@ -486,6 +486,10 @@ class Dokchitser(SageObject):
 
         - ``s`` -- complex number
 
+        - ``c`` -- internal parameter, call with `c>1` to get the same value
+          with a different cutoff point (`c` close to `1`); should return the
+          same answer, good to check if everything works with right precision
+
         .. NOTE::
 
            Evaluation of the function takes a long time, so each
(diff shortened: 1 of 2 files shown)


user202729:
I don't know what the `der` is supposed to do (previously it is just passed to the `c` parameter which is then discarded), but given the name my best guess is `derivative`. So the change the code accordingly.

Let's see if the tests pass.

user202729:
Maybe @robertwb can give a review and check if the intention of `der` is correct? You created the `GrossZagierLseries` class.

fchapoton:
note that the Dokchister implementation is rather here as an legacy (maybe even obsolete) one, as there is similar code included in pari, for which we do have an interface already in place

user202729:
I guess we could add that to the documentation too, pointing out it is deprecated and what is the equivalent new interface.

That said, the program is just a script in pari/gp?

fchapoton:
does not build, typo on line 526

fchapoton (approved):
ok ; why not

PR #38483, opened 2024-08-07
allow supplying a value of q for special_supersingular_curve()

yyyyx4:
This is a follow-up to #36665: There, a CM endomorphism is chosen automatically to construct a "special" supersingular elliptic curve. In some applications, it can be useful to explicitly supply the desired degree of such an endomorphism to Bröker's algorithm, which is made possible by this patch.

Note that some of the subroutines used in the new code paths are relatively fragile, which can cause the functionality added here to fail for some inputs, but this has been documented in the appropriate places.

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_finite_field.py
+++ b/src/sage/schemes/elliptic_curves/ell_finite_field.py
@@ -2565,10 +2565,12 @@ def is_j_supersingular(j, proof=True):
     return E.trace_of_frobenius() % p == 0
 
 
-def special_supersingular_curve(F, *, endomorphism=False):
+def special_supersingular_curve(F, q=None, *, endomorphism=False):
     r"""
-    Given a finite field ``F``, construct a "special" supersingular
-    elliptic curve `E` defined over ``F``.
+    Given a finite field ``F`` of characteristic `p`, and optionally
+    a positive integer `q` such that the Hilbert conductor of `-q`
+    and `-p` equals `p`, construct a "special" supersingular elliptic
+    curve `E` defined over ``F``.
 
     Such a curve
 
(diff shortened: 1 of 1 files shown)


src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -2694,42 +2750,49 @@ def special_supersingular_curve(F, *, endomorphism=False):
...
+            q = 1
+        elif p % 3 == 2:
+            q = 3
+        elif p % 8 == 5:
+            q = 2
         else:
-            assert False  # should never happen
+            from sage.arith.misc import legendre_symbol
+            for q in map(ZZ, range(3,p,4)):
+                if not q.is_prime():
```
GiacomoPope, line 2771: is it work just iterating through `primes()` here instead of checking primality?

GiacomoPope (approved):
one tiny comment otherwise this patch seems nice.

src/sage/schemes/elliptic_curves/ell_finite_field.py
yyyyx4: I don't think `primes()` supports the "step" argument (`4`). What could work is iterating through the primes and skipping the ones that aren't congruent to 3 modulo 4, but it's the same amount of code, and this part should be negligible in terms of computation anyway.

GiacomoPope: yeah my thinking was it's maybe faster to ask for the primes and check if it's the right congruence versus ensuring the right congruence and checking primality, but i agree that the majority of CPU time is not spent here (hence the positive review either way :) )

yyyyx4: Thank you for the review!

PR #38487, opened 2024-08-08
convenience methods for field embeddings

yyyyx4:
Currently, the best(?) way to compute embeddings between fields is somewhat hidden inside the `Hom` object. In this patch we add helper methods to access such embeddings. In addition, we add an `.algebraic_closure()` method to non-prime finite fields.

These seem useful.

### DIFF
--- a/src/sage/rings/finite_rings/finite_field_base.pyx
+++ b/src/sage/rings/finite_rings/finite_field_base.pyx
@@ -1871,6 +1871,12 @@ cdef class FiniteField(Field):
             sage: F.gen(3)
             z3
 
+        For finite fields, the algebraic closure is always (isomorphic
+        to) the algebraic closure of the prime field:
+
+            sage: GF(5^2).algebraic_closure() == F
+            True
+
         The default name is 'z' but you can change it through the option
         ``name``::
 
@@ -1890,13 +1896,18 @@ cdef class FiniteField(Field):
 
         .. NOTE::
 
-            This is currently only implemented for prime fields.
+            For non-prime finite fields, this method currently simply
+            returns the algebraic closure of the prime field. This may
+            or may not change in the future when extension towers are
+            supported properly.
 
         TESTS::
 
             sage: GF(5).algebraic_closure() is GF(5).algebraic_closure()
             True
         """
+        if not self.is_prime_field():
+            return self.prime_subfield().algebraic_closure()
         from sage.rings.algebraic_closure_finite_field import AlgebraicClosureFiniteField
         return AlgebraicClosureFiniteField(self, name, **kwds)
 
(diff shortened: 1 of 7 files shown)


grhkm21:
One minor (relevant) error with tests/docs/CI.

yyyyx4:
Thanks, should be fixed.

src/sage/rings/finite_rings/finite_field_base.pyx
```diff
@@ -1871,6 +1871,12 @@ cdef class FiniteField(Field):
             sage: F.gen(3)
             z3
 
+        For finite fields, the algebraic closure is always (isomorphic
+        to) the algebraic closure of the prime field:
```
GiacomoPope: I think this needs a double colon for the block?

```suggestion
        to) the algebraic closure of the prime field::
```

GiacomoPope (approved):

GiacomoPope:
Tiny comment about documentation, otherwise looks good to me.

yyyyx4:
Indeed, thanks, done.

vbraun:
Test fail

src/sage/rings/finite_rings/finite_field_base.pyx
```diff
@@ -1920,6 +1931,107 @@ cdef class FiniteField(Field):
...
+            [
+            Ring morphism:
+              From: Cyclotomic Field of order 3 and degree 2
+              To:   Cyclotomic Field of order 6 and degree 2
+              Defn: zeta3 |--> zeta6 - 1,
+            Ring morphism:
+              From: Cyclotomic Field of order 3 and degree 2
+              To:   Cyclotomic Field of order 6 and degree 2
+              Defn: zeta3 |--> -zeta6
+            ]
```
S17A05: I think these tests should not be here, since the fields to be embedded are not finite fields

src/sage/rings/finite_rings/finite_field_base.pyx
```diff
@@ -1920,6 +1931,107 @@ cdef class FiniteField(Field):
...
+            from sage.rings.algebraic_closure_finite_field import AlgebraicClosureFiniteField_generic
+            if not isinstance(K, AlgebraicClosureFiniteField_generic):
+                raise NotImplementedError('computing embeddings into this ring not implemented')
+        g = self.gen()
+        if (emb := K.coerce_map_from(self)) is not None:
+            return self.hom([emb(g)])
+        try:
+            r = g.minpoly().change_ring(K).any_root()
+        except ValueError:
+            raise ValueError(f'no embedding from {self} to {K}')
```
S17A05, line 1967: This can also raise a `TypeError` if I tested it correctly, e.g. when asking for an embedding from `GF(2)` into `GF(7).algebraic_closure()`; not sure if also excepting `TypeError` leads to some other issue, though

src/sage/rings/finite_rings/finite_field_base.pyx
```diff
@@ -1920,6 +1931,107 @@ cdef class FiniteField(Field):
...
+              Defn: a |--> ...
+        """
+        try:
+            return super().an_embedding(K)
+        except (NotImplementedError, ValueError):
+            pass
+        if K not in FiniteFields():
+            from sage.rings.algebraic_closure_finite_field import AlgebraicClosureFiniteField_generic
+            if not isinstance(K, AlgebraicClosureFiniteField_generic):
+                raise NotImplementedError('computing embeddings into this ring not implemented')
```
S17A05, line 1960: It might be a nice addition to compare here (or somewhere before) the characteristic of both fields, so that a lot of basic cases where embeddings cannot exist (e.g. `GF(2)` into `QQbar`) are caught with a more explicit error message.

src/sage/rings/ring.pyx
```diff
@@ -1478,6 +1478,55 @@ cdef class Field(CommutativeRing):
...
+        H = self.Hom(K)
+        try:
+            return H.natural_map()
+        except TypeError:
+            pass
+        from sage.categories.sets_cat import EmptySetError
+        try:
+            return H.an_element()
+        except EmptySetError:
+            raise ValueError(f'no embedding from {self} to {K}')
```
S17A05, line 1541: Unless I made a mistake in setup, this seems to not work as intended: E.g. for `GF(2).an_embedding(GF(7))` it does not raise an `EmptySetError`, and instead just returns `None`.

src/sage/rings/finite_rings/finite_field_base.pyx
```diff
@@ -1920,6 +1931,107 @@ cdef class FiniteField(Field):
...
+        g = self.gen()
+        rs = []
+        if (emb := K.coerce_map_from(self)) is not None:
+            rs.append(emb(g))
+        rs += [r for r,_ in g.minpoly().roots(ring=K) if r not in rs]
+        return [self.hom([r]) for r in rs]
```
S17A05, line 2002: One more thing: This also seems to fail when considering different characteristics, e.g. `GF(2).embeddings(GF(7))` runs into the issue that it cannot defined a homorphism mapping `1` to `6`; possibly this and all the other issues can already be prevented by the characteristic comparison

yyyyx4:
Thanks for the review, @S17A05! I think I've addressed all your comments.

S17A05:
> Thanks for the review, @S17A05! I think I've addressed all your comments.

Thanks for the changes, looks good to me now! (For the record: The timeout in the long [p-z] tests is unrelated)

PR #38488, opened 2024-08-08
implement smooth_part() and coprime_part()

yyyyx4:
Two generic convenience functions for Euclidean domains. These are both easier to use and faster than trial division.


GiacomoPope (approved):
Nice addition! Thank you

PR #38490, opened 2024-08-08
prevent crash in order_from_multiple() due to rounding error

### LINKED ISSUE #38489: order_from_multiple() fails for some inputs
Example (Sage 10.4):
```sage
sage: from sage.groups.generic import order_from_multiple
sage: plist = [43, 257, 547, 881]
sage: m = prod(plist[:-1])
sage: elt = Zmod(m)(plist[-1])
sage: order_from_multiple(elt, m, plist=plist)
# ...
UnboundLocalError: cannot access local variable 'k' where it is not associated with a value
```

(Discovered by Jonas Meers.)

yyyyx4:
Due to what appears to be rounding errors, the value of `k` here can be zero, which results in the list `L` being split as `L1 == []` and `L2 == L`. This causes an error in the next recursive call.

Simple workaround: If either `L1` or `L2` would end up being empty, we split the lists in the middle instead.

Resolves #38489.

### DIFF
--- a/src/sage/groups/generic.py
+++ b/src/sage/groups/generic.py
@@ -1257,6 +1257,17 @@ def order_from_multiple(P, m, plist=None, factorization=None, check=True,
         sage: K.<a> = GF(3^60)
         sage: order_from_multiple(a, 3^60 - 1, operation='*', check=False)
         42391158275216203514294433200
+
+    TESTS:
+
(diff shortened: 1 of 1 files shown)


grhkm21:
god.

grhkm21:
good*

PR #38492, opened 2024-08-09
expose fplll enumeration routines in IntegralLattice

yyyyx4:
This is useful for situations where we'd like to iterate over short-ish or close-ish vectors without necessarily requiring *the* shortest/closest solution, and without knowing a priori how many vectors we need to try. Among other things, it shows up in some algorithms for solving (quaternion) norm equations.

### DIFF
--- a/src/sage/modules/free_quadratic_module_integer_symmetric.py
+++ b/src/sage/modules/free_quadratic_module_integer_symmetric.py
@@ -39,6 +39,7 @@ AUTHORS:
 
 - Simon Brandhorst (2017-09): First created
 - Paolo Menegatti (2018-03): Added IntegralLatticeDirectSum, IntegralLatticeGluing
+- Lorenz Panny (2024): enumeration routines for short and close vectors
 """
 
 # ****************************************************************************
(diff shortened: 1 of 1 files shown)


yyyyx4:
(Cc: @GiacomoPope)

src/sage/modules/free_quadratic_module_integer_symmetric.py
```diff
@@ -1527,6 +1528,105 @@ def short_vectors(self, n, **kwargs):
...
+            [(1, 0, 6, -9), (1, -1, 5, -9), (2, 0, 6, -9), (1, 0, 5, -9), (1, -1, 5, -10),
+             (2, 1, 6, -9), (1, 0, 5, -10), (2, 0, 5, -9), (1, 0, 6, -8), (1, -1, 6, -9)]
+        """
+        L = self.LLL()
+        dim = L.dimension()
+        gram = L.gram_matrix()
+        basis = L.basis_matrix()
+
+        import fpylll
+        gmat = fpylll.IntegerMatrix(dim, dim)
```
GiacomoPope, line 1554: There's the `IntegerMatrix.from_matrix()` method, could you not do

```
gmat = fpyll.IntegerMatrix.from_matrix(gram)
gso = fpylll.GSO.Mat(gmat, gram=True)
```

src/sage/modules/free_quadratic_module_integer_symmetric.py
```diff
@@ -1527,6 +1528,105 @@ def short_vectors(self, n, **kwargs):
...
+            for j in range(dim):
+                gmat[i,j] = gram[i,j]
+        gso = fpylll.GSO.Mat(gmat, gram=True)
+        ok = gso.update_gso()
+        assert ok
+
+        coord = None
+        if target is not None:
+            coord = basis.solve_left(target)
+            Mu = 1 + matrix([gso.get_mu(i,j) for j in range(dim)] for i in range(dim))
```
GiacomoPope, line 1565: There's also a `to_matrix()` method, so something like:

```py
Mu = 1 + matrix(gso.to_matrix())
```

might work

src/sage/modules/free_quadratic_module_integer_symmetric.py
```diff
@@ -1527,6 +1528,105 @@ def short_vectors(self, n, **kwargs):
...
+        while True:
+            enum = fpylll.Enumeration(gso, count, fpylll.EvaluatorStrategy.BEST_N_SOLUTIONS)
+            try:
+                combs = enum.enumerate(0, dim, bound, 0, coord)
+            except fpylll.EnumerationError:
+                combs = []
+            if len(combs) < count:
+                bound *= 2
+                continue
+            for length,comb in combs:
```
GiacomoPope, line 1580:
```suggestion
            for length, comb in combs:
```

GiacomoPope (approved):
Some very small comments which could be ignored. Thank you for adding this to Sage, I should have done it ages ago.

src/sage/modules/free_quadratic_module_integer_symmetric.py
grhkm21: This doesn't work for me https://sagecell.sagemath.org/?q=yduvls (10.3)

GiacomoPope: oh for the MatGSO type it's the property `.gso.G`

```
Mu = 1 + matrix(gso.G)
```

https://sagecell.sagemath.org/?q=qifgtc

grhkm21: Ahha yes. I think this should be added before it's merged, seems quicker :-)

yyyyx4:
Thanks for the reviews! I think this is now already in the process of getting merged. Perhaps your suggested changes would best be done as a (small) follow-up?

GiacomoPope:
Sure I'll make a PR for those two. It's essentially meaningless tho!

jhpalmieri:
This seems to be causing doctest failures for me on OS X 14.6:
```
sage -t --random-seed=64986521010388899927548704615315435332 src/sage/modules/free_quadratic_module_integer_symmetric.py
File "src/sage/modules/free_quadratic_module_integer_symmetric.py", line 1544, in sage.modules.free_quadratic_module_integer_symmetric.FreeQuadraticModule_integer_symmetric._fplll_enumerate
Failed example:
    [next(close) for _ in range(10)]
Expected:
    [(1, 0, 6, -9), (1, -1, 5, -9), (2, 0, 6, -9), (1, 0, 5, -9), (1, -1, 5, -10),
     (2, 1, 6, -9), (1, 0, 5, -10), (2, 0, 5, -9), (1, 0, 6, -8), (1, -1, 6, -9)]
Got:
    [(1, 0, 6, -9),
     (1, -1, 5, -9),
     (2, 0, 6, -9),
     (1, 0, 5, -9),
     (1, -1, 5, -10),
     (2, 1, 6, -9),
     (1, 0, 5, -10),
     (1, 0, 6, -8),
     (2, 0, 5, -9),
     (1, -1, 6, -9)]
File "src/sage/modules/free_quadratic_module_integer_symmetric.py", line 1601, in sage.modules.free_quadratic_module_integer_symmetric.FreeQuadraticModule_integer_symmetric.enumerate_short_vectors
Failed example:
    [next(short) for _ in range(20)]
Expected:
    [(1, -1, 1, 0), (2, -2, 2, 0), (3, -3, 3, 0), (0, 3, 2, 4), (1, 2, 3, 4),
     (4, 4, 1, 0), (3, 2, -2, -4), (3, 5, 0, 0), (4, 1, -1, -4), (-1, 4, 1, 4),
     (2, 1, 4, 4), (5, 3, 2, 0), (2, 3, -3, -4), (2, 6, -1, 0), (5, 0, 0, -4),
     (-2, 5, 0, 4), (4, -4, 4, 0), (6, 2, 3, 0), (1, 4, -4, -4), (3, 0, 5, 4)]
Got:
    [(1, -1, 1, 0),
     (2, -2, 2, 0),
     (3, -3, 3, 0),
     (0, 3, 2, 4),
     (1, 2, 3, 4),
     (4, 4, 1, 0),
     (3, 2, -2, -4),
     (3, 5, 0, 0),
     (4, 1, -1, -4),
     (-1, 4, 1, 4),
     (2, 1, 4, 4),
     (5, 3, 2, 0),
     (2, 3, -3, -4),
     (5, 0, 0, -4),
     (2, 6, -1, 0),
     (-2, 5, 0, 4),
     (4, -4, 4, 0),
     (6, 2, 3, 0),
     (1, 4, -4, -4),
     (3, 0, 5, 4)]
2 items had failures:
   1 of   7 in sage.modules.free_quadratic_module_integer_symmetric.FreeQuadraticModule_integer_symmetric._fplll_enumerate
   1 of   4 in sage.modules.free_quadratic_module_integer_symmetric.FreeQuadraticModule_integer_symmetric.enumerate_short_vectors
    [238 tests, 2 failures, 3.93 s]
```
Can you please open up a follow-up to fix this?

GMS103:
FWIW, here is what I get on Apple Silicon for ```make ptestlong``` (SageMath 10.5.beta2).

macOS 12.7.6:  All tests passed!
macOS 13.6.9:  This failure (among others).
macOS 14.6.1:  This failure (among others).

videlec:
Is the enumeration order relevant? The only difference in the output is the transposition of the two vectors `(5, 0, 0, -4)` and `(2, 6, -1, 0)`.

grhkm21:
Yeah I don't think the output order is/should be deterministic for any reason. I can make a PR and add a `sorted`.

grhkm21:
Actually, since the norm of the vectors aren't necessarily non-decreasing, the first n outputs of the iterator might not be the same. So I think it's best to mark it #random.

PR #38493, opened 2024-08-10
compute endomorphism orders for elliptic curves over finite fields (rank-2 case)

yyyyx4:
For an elliptic curve E over a finite field with an endomorphism algebra of rank two over ℚ, the endomorphism ring is a superorder of the "Frobenius order" ℤ[π].

This patch adds a simple method to compute the exact imaginary-quadratic order containing ℤ[π] which is isomorphic to the endomorphism ring of the curve, and generalizes the algorithm to supersingular elliptic curves with a (rational) endomorphism algebra of rank 2.


JohnCremona:
I'll look at this tomorrow. I did implement a special case of Sutherland 's algorithm already for our CM testing algorithm last year.

yyyyx4:
Thank you! I think the implementation you wrote is the function I'm calling here. I just extended it to work for the supersingular case, too.

JohnCremona:
Apologies to @yyyyx4 for promising and then not delivering -- sheer forgetfulness on my part.  I cannot promise to be able to review this in the next several days.  Feel free to send me reminders.

yyyyx4:
No worries, @JohnCremona! I do think it would be nice to get this into the next release, but it is by no means urgent.

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1652,30 +1660,71 @@ def height_above_floor(self, ell, e):
...
-            raise ValueError("{} is not ordinary".format(self))
-        if e == 0:
-            return 0
+            if ell == self.base_field().characteristic():
+                # In this (exceptional) case, the Frobenius can always be divided
+                # by the maximal possible power of the characteristic. The reason
+                # is that Frobenius must be of the form phi o [p^k] where phi is
+                # a purely inseparable isogeny of degree 1 or p, hence this [p^k]
+                # can always be divided out while retaining an endomorphism.
+                assert (ell**(2*e)).divides(self.base_field().cardinality())
```
JohnCremona: Perhaps better to test that the ell-valuation is at least 2*e rather than computing ell**(2*e)?

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1652,30 +1660,71 @@ def height_above_floor(self, ell, e):
...
+            # the level in the volcano. (The underlying reason is that there can
+            # be multiple non-F_q-isomorphic curves with a given j-invariant in
+            # the isogeny graph.)
+            # Example: y^2 = x^3 ± x over F_p with p congruent to 3 modulo 4 have
+            # distinct (rational) endomorphism rings.
+            # Thus we run the "probing the depths" algorithm with F_q-isomorphism
+            # classes of curves instead.
+            E0 = [self] * 3
+            E1 = [phi.codomain() for phi in self.isogenies_prime_degree(ell)]
+            assert E1
```
JohnCremona, line 1691: I don't think that including assert statements is good practice for production code (though of course essential while testing).  Does the developers' guide say something about this?  But the alternative would be to test the condition anyway and raise an error if it fails, so perhaps it does not make much difference.  (Laos, I am sure thta my own code in Sage does include assert statements...)

src/sage/schemes/elliptic_curves/ell_finite_field.py
```diff
@@ -1652,30 +1660,71 @@ def height_above_floor(self, ell, e):
...
+                for i in range(3):
+                    isogs = E1[i].isogenies_prime_degree(ell)
+                    try:
+                        step = next(phi for phi in isogs if not phi.codomain().is_isomorphic(E0[i]))
+                    except StopIteration:
+                        return h
+                    E0[i], E1[i] = step.domain(), step.codomain()
+                h += 1
+                assert h <= e
+            raise NotImplementedError
```
JohnCremona: A comment here would be helpful.

JohnCremona (changes requested):
Looks good, I made a few rather trivial comments but basically approve.

src/sage/schemes/elliptic_curves/ell_finite_field.py
yyyyx4: This is essentially unreachable code unless something goes *seriously* wrong (cosmic rays). I've turned it into an assertion with a comment for clarity.

src/sage/schemes/elliptic_curves/ell_finite_field.py
yyyyx4: I don't think there are any rules about it. I left these in since they are very quick sanity checks that might detect if something is really wrong with the algorithm. Exceptions are for a different purpose; they communicate errors that can occur as part of the *expected* functionality of the code (e.g., invalid inputs). In my opinion, there should be assertions all over the code base anyway, with the default in production environments being that they are turned off (`python -O` or `PYTHONOPTIMIZE=1`), but that's a different discussion...

yyyyx4:
Thanks for the review, @JohnCremona! If you are happy with my latest changes, don't forget to approve the pull request (I think I can no longer set "positive review" myself, contrary to the old Trac days).

JohnCremona (approved):

PR #38499, opened 2024-08-11
CI / `tox docker-incremental`: Fix incremental build from a different Sage version

mkoeppe:
As reported in https://groups.google.com/g/sage-devel/c/H6SHSpLF87E @kwankyu 

Reproducer: 
`DOCKER_TARGETS=bootstrapped FROM_DOCKER_TAG=10.5.beta1 tox -e docker-ubuntu-jammy-incremental-standard`; then in the built container, `cat src/VERSION.txt` should give 10.5.beta2, but it gives 10.5.beta1.

### :memo: Checklist

### DIFF
--- a/.ci/write-dockerfile.sh
+++ b/.ci/write-dockerfile.sh
@@ -283,8 +283,8 @@ $ADD .upstream.d /new/.upstream.d
 ADD .ci /.ci
 RUN if [ -d /sage ]; then                                               \
         echo "### Incremental build from \$(cat /sage/VERSION.txt)" &&  \
-        printf '/src\n!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n!/src/VERSION.txt\n' >> /sage/.gitignore && \
-        printf '/src\n!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n!/src/VERSION.txt\n' >> /new/.gitignore && \
+        printf '/src/*\n!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n!/src/VERSION.txt\n' >> /sage/.gitignore && \
+        printf '/src/*\n!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n!/src/VERSION.txt\n' >> /new/.gitignore && \
         if ! (cd /new && /.ci/retrofit-worktree.sh worktree-image /sage); then \
             echo "retrofit-worktree.sh failed, falling back to replacing /sage"; \
             for a in local logs; do                                     \


kwankyu:
I want to understand the situation 
```
...
#31 12.26   [sagelib-10.5.beta1]   Deleting build directory /sage/local/var/lib/sage/venv-python3.10/var/tmp/sage/build/sagelib-10.5.beta1
#31 12.26   [sagelib-10.5.beta1]   Finished installing sagelib-10.5.beta1
#31 12.26   [sagelib-10.5.beta1]   real 0m13.860s user 0m9.170s sys 0m5.237s
#31 12.26   [sagelib-10.5.beta1]   Error: Selecting a different version of a package is no longer supported
#31 12.26   [sagelib-10.5.beta1]   Error: Selecting a different version of a package is no longer supported
```
So the old sage install is of version beta1, and the version of the new src is of version beta2, but due to the bug, the file `src/VERSION.txt` in the docker image still contains `10.5.beta1`. Right? What does the error message "Error: Selecting a different version of a package is no longer supported"?

Your fix seems to indicate that the bug may have caused subtile errors in "CI Linux incremental". No?

mkoeppe:
> So the old sage install is of version beta1, and the version of the new src is of version beta2, but due to the bug, the file `src/VERSION.txt` in the docker image still contains `10.5.beta1`. Right?

Yes

> What does the error message "Error: Selecting a different version of a package is no longer supported" mean?

The `build/bin/Makefile` (which was generated at `configure` time) thinks that 10.5.beta1 is to be installed.
But the `sage-spkg` script knows better and complains (correctly).

> Your fix seems to indicate that the bug may have caused subtile errors in "CI Linux incremental". No?

Not really. This discrepancy (`src/VERSION.txt` wrong at the time of running `configure` but correct at build time) has been there ever since I introduced the incremental builds using "retrofit-worktree" but has been unnoticed (or ignored).
The error is new thanks to a recently tightened check in `sage-spkg`.

kwankyu:
Other files listed in `!/src/doc/bootstrap\n!/src/bin\n!/src/*.m4\n!/src/*.toml\n` are also affected?

mkoeppe:
Yes

mkoeppe:
> Your fix seems to indicate that the bug may have caused subtile errors in "CI Linux incremental".

OK, possibly, yes

kwankyu:
Anyway, the discrepancy would have existed only for a short period after a release. So perhaps not a major issue affecting "CI Linux Incremental". 

My impression is that "CI Linux Incremental" is not reliable. I wondered if this PR improves the situation. Perhaps not much...

mkoeppe:
> My impression is that "CI Linux Incremental" is not reliable.

Can you be more specific?

kwankyu:
I simply mean that I have the impression that workflow runs of "CI Linux Incremental" usually fail. So I tend to ignore the checks.

kwankyu:
Is it working well?

kwankyu (approved):
OK. Thanks for the fix.

mkoeppe:
> I simply mean that I have the impression that workflow runs of "CI Linux Incremental" usually fail. So I tend to ignore the checks.

There are some that regularly fail because of known bugs that also affect the non-incremental builds for the same platform. An example: Platforms with system giac often fail because of an upstream giac bug (which we patch in the SPKG).

Other incremental builds fail because they expose limitations of our build system in incremental builds. For example #38169 exposes that we do not have proper tracking of dependencies on system libraries (this is known and tracked in #29711).

mkoeppe:
Thanks for the review!

PR #38183, opened 2024-06-10
`build/pkgs/cypari` 2.2.0

mkoeppe:
- https://pypi.org/project/cypari2/2.2.0/
- https://github.com/sagemath/cypari2/blob/master/NEWS
- Another step toward #32431

### :memo: Checklist

- [x] The title is concise and informative.
- [ ] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #37885, opened 2024-04-27
build: show all cython warnings

tornaria:
Motivated by https://github.com/sagemath/sage/pull/37792#issuecomment-2053663025, change cython configuration so we show all warnings.

This will lead to a number of warnings, but it's IMO better to know about that and works towards fixing as much warnings as possible (and we have a fair number of warnings that are indeed relevant and *not* just false positives).

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.

PR #37667, opened 2024-03-24
Fix noexcept clauses (#37560)

tornaria:
In #36507 I added a lot of `noexcept` clauses guided by a warning which is not quite right (see https://github.com/cython/cython/pull/5999#issuecomment-1986868208).

A new warning in 3.0.9 shows the mistake (incorrectly added `noexcept` clauses). This broke some doctests (#37560) and a workaround was implemented for 10.3 (#37583) since we were too close to release.

This PR now does the proper fix, removing all the incorrect `noexcept`, and also adding a few missing `noexcept`.

Note: if one tries this PR with cython <= 3.0.8 it seems it's wrong in the sense that it will show 40k more warnings after the PR. These are *incorrect* warnings.

If one uses cython 3.0.9 each of these 40k lines give a correct warning before this PR and an incorrect warnings after the PR, and there is no way to avoid 40k warnings.

To confirm this PR is a good one, one needs to use cython 3.0.9 + https://github.com/cython/cython/pull/6087 or wait for cython 3.0.10. In this case, the warnings are correct and one will get 40k warnings before the PR and no warning after the PR.

The last commit reverts the workaround from #37583 so we don't silence these warnings on `cython()` now that we have our own code right.

@vbraun: this is "almost trivial" but touches too many files. Could we merge it in beta0 to avoid conflicts? I did this "kind of" automatically and I've been testing it for two weeks in almost all my builds. The changes made should not affect behaviour at all in the legacy mode we are using cython.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.

PR #38500, opened 2024-08-12
cython 3.0.11 compatibility

kiwifb:
I recently installed cython 3.0.11 in Gentoo and it caused issue building sage. The issues are not present when building with cython 3.0.10. The fix is just to add some `noexcept` statements in some of the place cython complains about. More details of the issue and some of the error messages encountered at https://github.com/cschwan/sage-on-gentoo/issues/794

### :memo: Checklist

### DIFF
--- a/src/sage/matroids/matroid.pxd
+++ b/src/sage/matroids/matroid.pxd
@@ -9,7 +9,7 @@ cdef class Matroid(SageObject):
 
     # virtual methods
     cpdef frozenset groundset(self)
-    cpdef int _rank(self, frozenset X) except -1
+    cpdef int _rank(self, frozenset X) except? -1
 
     # internal methods, assuming verified input
     cpdef frozenset _max_independent(self, frozenset X)
--- a/src/sage/matroids/matroid.pxd
+++ b/src/sage/matroids/matroid.pxd
@@ -9,28 +9,28 @@ cdef class Matroid(SageObject):
 
     # virtual methods
     cpdef frozenset groundset(self)
-    cpdef int _rank(self, frozenset X)
+    cpdef int _rank(self, frozenset X) noexcept
 
     # internal methods, assuming verified input
     cpdef frozenset _max_independent(self, frozenset X)
     cpdef frozenset _circuit(self, frozenset X)
     cpdef frozenset _fundamental_circuit(self, frozenset B, e)
     cpdef frozenset _closure(self, frozenset X)
-    cpdef int _corank(self, frozenset X)
+    cpdef int _corank(self, frozenset X) noexcept
     cpdef frozenset _max_coindependent(self, frozenset X)
     cpdef frozenset _cocircuit(self, frozenset X)
     cpdef frozenset _fundamental_cocircuit(self, frozenset B, e)
     cpdef frozenset _coclosure(self, frozenset X)
     cpdef frozenset _augment(self, frozenset X, frozenset Y)
 
-    cpdef bint _is_independent(self, frozenset X)
-    cpdef bint _is_basis(self, frozenset X)
-    cpdef bint _is_circuit(self, frozenset X)
-    cpdef bint _is_closed(self, frozenset X)
-    cpdef bint _is_coindependent(self, frozenset X)
-    cpdef bint _is_cobasis(self, frozenset X)
-    cpdef bint _is_cocircuit(self, frozenset X)
-    cpdef bint _is_coclosed(self, frozenset X)
+    cpdef bint _is_independent(self, frozenset X) noexcept
+    cpdef bint _is_basis(self, frozenset X) noexcept
+    cpdef bint _is_circuit(self, frozenset X) noexcept
+    cpdef bint _is_closed(self, frozenset X) noexcept
+    cpdef bint _is_coindependent(self, frozenset X) noexcept
+    cpdef bint _is_cobasis(self, frozenset X) noexcept
+    cpdef bint _is_cocircuit(self, frozenset X) noexcept
+    cpdef bint _is_coclosed(self, frozenset X) noexcept
 
     cpdef _minor(self, contractions, deletions)
     cpdef _has_minor(self, N, bint certificate=*)
@@ -105,7 +105,7 @@ cdef class Matroid(SageObject):
     cpdef is_coclosed(self, X)
 
     # verification
-    cpdef bint is_valid(self)
+    cpdef bint is_valid(self) noexcept
 
     # enumeration
     cpdef SetSystem circuits(self, k=*)
@@ -184,8 +184,8 @@ cdef class Matroid(SageObject):
     cpdef _is_3connected_CE(self, certificate=*)
     cpdef _is_3connected_BC(self, certificate=*)
     cpdef _is_3connected_BC_recursion(self, basis, fund_cocircuits)
-    cpdef bint is_paving(self)
-    cpdef bint is_sparse_paving(self)
+    cpdef bint is_paving(self) noexcept
+    cpdef bint is_sparse_paving(self) noexcept
     cpdef girth(self)
 
     # representability
@@ -195,8 +195,8 @@ cdef class Matroid(SageObject):
     cpdef _local_ternary_matroid(self, basis=*)
     cpdef ternary_matroid(self, randomized_tests=*, verify=*)
     cpdef is_ternary(self, randomized_tests=*)
-    cpdef bint is_regular(self)
-    cpdef bint is_graphic(self)
+    cpdef bint is_regular(self) noexcept
+    cpdef bint is_graphic(self) noexcept
 
     # matroid k-closed
     cpdef is_k_closed(self, int k)
--- a/src/sage/matroids/matroid.pxd
+++ b/src/sage/matroids/matroid.pxd
@@ -9,7 +9,7 @@ cdef class Matroid(SageObject):
 
     # virtual methods
     cpdef frozenset groundset(self)
-    cpdef int _rank(self, frozenset X) noexcept
+    cpdef int _rank(self, frozenset X) except -1
 
     # internal methods, assuming verified input
     cpdef frozenset _max_independent(self, frozenset X)
--- a/src/sage/matroids/matroid.pyx
+++ b/src/sage/matroids/matroid.pyx
@@ -492,7 +492,7 @@ cdef class Matroid(SageObject):
         """
         raise NotImplementedError("subclasses need to implement this")
 
-    cpdef int _rank(self, frozenset X):
+    cpdef int _rank(self, frozenset X) noexcept:
         r"""
         Return the rank of a set ``X``.
 
@@ -732,7 +732,7 @@ cdef class Matroid(SageObject):
                 XX.pop()
         return frozenset(XX)
 
-    cpdef int _corank(self, frozenset X):
+    cpdef int _corank(self, frozenset X) noexcept:
         """
         Return the corank of a set.
 
@@ -904,7 +904,7 @@ cdef class Matroid(SageObject):
 
     # override the following methods for even better efficiency
 
-    cpdef bint _is_independent(self, frozenset X):
+    cpdef bint _is_independent(self, frozenset X) noexcept:
         """
         Test if input is independent.
 
@@ -925,7 +925,7 @@ cdef class Matroid(SageObject):
         """
         return len(X) == self._rank(X)
 
-    cpdef bint _is_basis(self, frozenset X):
+    cpdef bint _is_basis(self, frozenset X) noexcept:
         """
         Test if input is a basis.
 
@@ -958,7 +958,7 @@ cdef class Matroid(SageObject):
         """
         return self._is_independent(X)
 
-    cpdef bint _is_circuit(self, frozenset X):
+    cpdef bint _is_circuit(self, frozenset X) noexcept:
         """
         Test if input is a circuit.
 
@@ -990,7 +990,7 @@ cdef class Matroid(SageObject):
             Z.add(x)
         return True
 
-    cpdef bint _is_closed(self, frozenset X):
+    cpdef bint _is_closed(self, frozenset X) noexcept:
         """
         Test if input is a closed set.
 
@@ -1019,7 +1019,7 @@ cdef class Matroid(SageObject):
             XX.discard(y)
         return True
 
-    cpdef bint _is_coindependent(self, frozenset X):
+    cpdef bint _is_coindependent(self, frozenset X) noexcept:
         """
         Test if input is coindependent.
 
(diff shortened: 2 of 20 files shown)


mkoeppe:
Confirmed the problem with the upgrade PR #38501, https://github.com/sagemath/sage/actions/runs/10345037926/job/28631469828?pr=38501#step:9:12644

mkoeppe:
Fix confirmed using #38501

kiwifb:
It is rather weird that only those cause problems out of the list of warnings I get. And there is very clear cython 3.0.10 to 3.0.11 boundary as well. Something like that happening with say cython 3.1.0 would not have been so weird.

mkoeppe:
see also https://github.com/sagemath/sage/pull/37885#issuecomment-2096518938

tornaria:
> It is rather weird that only those cause problems out of the list of warnings I get. And there is a very clear cython 3.0.10 to 3.0.11 boundary as well. Something like that happening with say cython 3.1.0 would not have been so weird.

It's a regression, see: https://github.com/cython/cython/issues/6335

If this PR gets rid of every "implicit noexcept" warning, perhaps `legacy_implicit_noexcept` can be switched off?

kiwifb:
>If this PR gets rid of every "implicit noexcept" warning, perhaps `legacy_implicit_noexcept` can be switched off?

It does not, but I would consider finishing the job if there is not too much more to do.

kiwifb:
Actually, I should properly check what's left, maybe it will be a nice surprise.

kiwifb:
The whole of the `cypari2` folder still need to be done. And there are those two bits that may need to be dealt with
```
warning: sage/categories/morphism.pyx:801:26: noexcept clause is ignored for function returning Python object
warning: sage/categories/morphism.pyx:826:22: noexcept clause is ignored for function returning Python object
```

kiwifb:
Once you remove duplicate warnings, it amounts to 15 lines, not too onerous
```
warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:25:30: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:45:34: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:60:28: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:63:22: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:65:30: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:68:32: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/gen.pxd:27:26: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/pari_instance.pxd:6:29: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/pari_instance.pxd:7:29: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/pari_instance.pxd:8:26: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/paridecl.pxd:5330:37: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/stack.pxd:11:21: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/stack.pxd:12:20: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/stack.pxd:14:32: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
warning: /usr/lib/python3.12/site-packages/cypari2/stack.pxd:19:22: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
```

kiwifb:
Actually this is not inside sage but the installed copy of cypari2. Does that prevent us from removing `legacy_implicit_noexcept`? Do we need a new release of `cypari2` to get rid of it?

mkoeppe:
For cypari2 see #38183

kiwifb:
I'd rather have proper release rather than rc, but that could be OK so long as we have final releases in time for 10.5.

mkoeppe:
I held off with cutting the releases when I ran into https://github.com/sagemath/sage/pull/38183#issuecomment-2159159277

gmou3 (approved):
I had noticed warnings about these files during building, which specifically mentioned a future lack of support of implicit `noexcept`s.

Most such removals happened in #37667. After that, #38057 added many function return types. This combination for `C` return types (`int`, `bint`, `void`) seems to yield an issue.

mkoeppe:
> I'd rather have proper release rather than rc

https://pypi.org/project/cypari2/2.2.0/

tornaria:
> Once you remove duplicate warnings, it amounts to 15 lines, not too onerous
> 
> ```
> warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:25:30: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
> ...
> ```

Afaict, this was already fixed on version 2.1.5 of cypari2 (march 2024).

kiwifb:
> > Once you remove duplicate warnings, it amounts to 15 lines, not too onerous
> > ```
> > warning: /usr/lib/python3.12/site-packages/cypari2/convert.pxd:25:30: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
> > ...
> > ```
> 
> Afaict, this was already fixed on version 2.1.5 of cypari2 (march 2024).

Right, I am still on 2.1.4. I have let things slip a bit since some packages have been moved to the main gentoo tree. Time to put in some PR for Gentoo.

tornaria:
> I had noticed warnings about these files during building, which specifically mentioned a future lack of support of implicit `noexcept`s.
> 
> Most such removals happened in #37667. After that, #38057 added many function return types. This combination for `C` return types (`int`, `bint`, `void`) seems to yield an issue.

But see, you have a problem here. As an example see https://github.com/sagemath/sage/blob/develop/src/sage/matroids/matroid.pyx#L495:
```python
...
    cpdef int _rank(self, frozenset X):
        r"""
        Return the rank of a set ``X``.

        This method does no checking on ``X``, and ``X`` may be assumed to
        have the same interface as ``frozenset``.

        INPUT:

        - ``X`` -- an object with Python's ``frozenset`` interface

        OUTPUT: integer

        .. NOTE::

            Subclasses should implement this method.

        EXAMPLES::

            sage: M = sage.matroids.matroid.Matroid()
            sage: M._rank(frozenset([0, 1, 2]))
            NotImplementedError: subclasses need to implement this
            ...
        """
        raise NotImplementedError("subclasses need to implement this")
...
```
Since this function is indeed raising an exception, I don't think it should be `noexcept` (neither implicit nor explicit).

Functions that return python types are `except NULL` or something like that (because a python type is a pointer so `NULL` is always there to flag an exception). When you switch the return type to a C type, this is no longer the case. You can tell cython what is a good value to use to flag an exception (e.g. rank is >= 0 so `-1` seems a good choice).

src/sage/matroids/matroid.pxd
```diff
@@ -9,28 +9,28 @@ cdef class Matroid(SageObject):
...
-    cpdef int _rank(self, frozenset X)
+    cpdef int _rank(self, frozenset X) noexcept
 
     # internal methods, assuming verified input
     cpdef frozenset _max_independent(self, frozenset X)
     cpdef frozenset _circuit(self, frozenset X)
     cpdef frozenset _fundamental_circuit(self, frozenset B, e)
     cpdef frozenset _closure(self, frozenset X)
-    cpdef int _corank(self, frozenset X)
+    cpdef int _corank(self, frozenset X) noexcept
```
gmou3, line 19:
```suggestion
    cpdef int _corank(self, frozenset X) except -1
```

src/sage/matroids/matroid.pyx
```diff
@@ -492,7 +492,7 @@ cdef class Matroid(SageObject):
         """
         raise NotImplementedError("subclasses need to implement this")
 
-    cpdef int _rank(self, frozenset X):
+    cpdef int _rank(self, frozenset X) noexcept:
```
gmou3:
```suggestion
    cpdef int _rank(self, frozenset X) except -1:
```

src/sage/sets/disjoint_set.pxd
```diff
@@ -19,8 +19,8 @@ cdef class DisjointSet_class(SageObject):
     cpdef number_of_subsets(self)
 
 cdef class DisjointSet_of_integers(DisjointSet_class):
-    cpdef int find(self, int i)
-    cpdef void union(self, int i, int j)
+    cpdef int find(self, int i) noexcept
```
gmou3:
```suggestion
    cpdef int find(self, int i) except -1
```

src/sage/sets/disjoint_set.pxd
```diff
@@ -19,8 +19,8 @@ cdef class DisjointSet_class(SageObject):
...
+    cpdef void union(self, int i, int j) noexcept
```
gmou3:
```suggestion
    cpdef void union(self, int i, int j) except *
```

src/sage/sets/disjoint_set.pyx
```diff
@@ -448,7 +448,7 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
         for i, parent in enumerate(l):
             self.union(parent, i)
 
-    cpdef int find(self, int i):
+    cpdef int find(self, int i) noexcept:
```
gmou3:
```suggestion
    cpdef int find(self, int i) except -1:
```

src/sage/sets/disjoint_set.pyx
```diff
@@ -492,7 +492,7 @@ cdef class DisjointSet_of_integers(DisjointSet_class):
             raise ValueError('i must be between 0 and %s (%s given)' % (card - 1, i))
         return OP_find(self._nodes, i)
 
-    cpdef void union(self, int i, int j):
+    cpdef void union(self, int i, int j) noexcept:
```
gmou3:
```suggestion
    cpdef void union(self, int i, int j) except *:
```

gmou3 (changes requested):
Got it, thanks @tornaria!

kiwifb:
@gmou3 are those all the changes we can do to follow @tornaria's suggestion? If so, I will accept all of them and we can put things back into positive review. For now I am switching back to need review.

gmou3:
Yes, these are all (for the files in question). However, in order to make things build, the signatures of the heirs of the `_rank` function must be consistent. A few doctest adaptations are also required.

@kiwifb I opened a PR to your branch that settles everything.

kiwifb:
> Yes, these are all (for the files in question). However, in order to make things build, the signatures of the heirs of the `_rank` function must be consistent. A few doctest adaptations are also required.
> 
> @kiwifb I opened a PR to your branch that settles everything.

Thank you for that.

src/sage/matroids/matroid.pxd
kiwifb: Did you forget that one in your PR?

gmou3: This function does not raise an error, so I think better leave it as is.

gmou3:
After adding `except -1` to `_rank` and playing around with `-1` rank values you get some `Exception ignored` messages:

```
sage: def r(X):
....:     return -1
sage: M = Matroid(groundset=[0,1,2], rank_function=r)
sage: M.is_valid()
Exception ignored in: 'sage.matroids.matroid.Matroid.is_valid'
False
sage: M._rank(frozenset())
-1
sage: M._corank(frozenset())
Exception ignored in: 'sage.matroids.matroid.Matroid._corank'
0
```
I can personally ignore such ignorations. I am mostly insecure about whether these changes slow down `_rank`, which is a core function that is often called ad nauseam.

mkoeppe:
Test with the update PR https://github.com/sagemath/sage/pull/38500 looks OK.

mkoeppe (approved):
LGTM

tornaria:
> After adding `except -1` to `_rank` and playing around with `-1` rank values you get some `Exception ignored` messages:
> 
> ```
> sage: def r(X):
> ....:     return -1
> sage: M = Matroid(groundset=[0,1,2], rank_function=r)
> sage: M.is_valid()
> Exception ignored in: 'sage.matroids.matroid.Matroid.is_valid'
> False
> sage: M._rank(frozenset())
> -1
> sage: M._corank(frozenset())
> Exception ignored in: 'sage.matroids.matroid.Matroid._corank'
> 0
> ```
> 
> I can personally ignore such ignorations. I am mostly insecure about whether these changes slow down `_rank`, which is a core function that is often called ad nauseam. (Some basic testing doesn't show a difference.)

When you tag a function as `except -1` you are making a promise to cython that your function will never return -1 on its own, so the value is used to indicate an exception was raised. The caller will assume an exception has been raised without checking it first (to save time).

If you want cython to use `-1` to indicate an exception but your function *might* return `-1`, you should tag the function as `except? -1`. In this case, when the caller sees a `-1` return value it will check first before assuming an exception has happened.

So, the warning you see means: the function returned `-1` and so cython assumed there an exception was raised, but in fact there was no exception to be found anywhere.

I suggest reading http://docs.cython.org/en/latest/src/userguide/language_basics.html#error-return-values (click on "cython" to see the cython syntax which is used in sagemath .pyx files)

If you want to support the possibility of having an implementation of these methods to return `-1` then I think `except? -1` seems the way to go. This should not slowdown anything except when the function indeed returns `-1` (rare) or when the function raises an exception (speed not really critical). IOW, I don't think there's any real harm in adding the `?` just in case.

gmou3:
@tornaria Thanks for the reference. Exactly what I needed.

@kiwifb After merging the question marks we're good to go.

gmou3:
did you maybe miss my PR Mr. 🥝? https://github.com/kiwifb/sage/pull/2

gmou3 (approved):

kiwifb:
> did you maybe miss my PR Mr. 🥝? [kiwifb#2](https://github.com/kiwifb/sage/pull/2)

I completely missed it. Good job poking me about it. 
The fruit "brand" was named after the nickname for new zealander which itself comes from the name of one iconic New Zealand animal https://www.doc.govt.nz/nature/native-animals/birds/birds-a-z/kiwi/

gmou3:
Cool, didn't know that. Cute animal!

PR #38508, opened 2024-08-14
Hashing affine group elements

### LINKED ISSUE #38381: AffineGroup elements should be hashable
### Problem Description

Currently `AffineGroup` elements do not implement hash and so can not be used as dictionary keys.

Currently the internal variables of `AffineGroupElement` are mutable, but the `matrix` method is cached suggesting that the values are expected to be immutable. This makes it possible to produce (far fetched) bugs where `matrix` returns something incompatible with the current state of the transformation:
```sage
sage: from sage.groups.affine_gps.affine_group import AffineGroup
sage: G = AffineGroup(2, QQ)
sage: G
Affine Group of degree 2 over Rational Field
sage: g = G([[1, 2], [3, 4]], [5,6])
sage: g.matrix()
[1 2|5]
[3 4|6]
[---+-]
[0 0|1]
sage: g.A()[0,0] = 9
sage: g.A()[0,0] = 9
sage: g
      [9 2]     [5]
x |-> [3 4] x + [6]
sage: g.matrix()
[1 2|5]
[3 4|6]
[---+-]
[0 0|1]
sage: hash(g)
TypeError
```

### Proposed Solution

1. Make affine group elements immutable by calling `set_immutable()` on the parameters `_A` and `_b` in the constructor.
2. Implement a hash.

### Alternatives Considered

If mutability is desired (for some reason I'm not aware of), then the `matrix` method should not be cached to avoid the bug above. In this case, I would imagine we'd also want to implement something explaining how to change the values.

### Additional Information

_No response_

tscrim:
Fixes #38381 and makes sure the defining parts of the group element are immutable.

### :memo: Checklist

### DIFF
--- a/src/sage/groups/affine_gps/group_element.py
+++ b/src/sage/groups/affine_gps/group_element.py
@@ -119,15 +119,17 @@ class AffineGroupElement(MultiplicativeGroupElement):
             A = A.matrix()
         except AttributeError:
             pass
-        if isinstance(A, Matrix) and A.nrows() == A.ncols() == parent.degree()+1:
+        if isinstance(A, Matrix) and A.nrows() == A.ncols() == parent.degree() + 1:
             g = A
             d = parent.degree()
             A = g.submatrix(0, 0, d, d)
-            b = [ g[i,d] for i in range(d) ]
+            b = [g[i,d] for i in range(d)]
             convert = True
         if convert:
             A = parent.matrix_space()(A)
             b = parent.vector_space()(b)
+            A.set_immutable()
+            b.set_immutable()
         if check:
             # Note: the coercion framework expects that we raise TypeError for invalid input
             if not isinstance(A, Matrix):
@@ -138,6 +140,14 @@ class AffineGroupElement(MultiplicativeGroupElement):
                 raise TypeError('b must be an element of ' + str(parent.vector_space()))
             parent._element_constructor_check(A, b)
         super().__init__(parent)
+        if not A.is_immutable():
+            from copy import copy
+            A = copy(A)
+            A.set_immutable()
+        if not b.is_immutable():
+            from copy import copy
+            b = copy(b)
+            b.set_immutable()
         self._A = A
         self._b = b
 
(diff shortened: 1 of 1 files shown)


src/sage/groups/affine_gps/group_element.py
```diff
@@ -138,6 +140,14 @@ def __init__(self, parent, A, b=0, convert=True, check=True):
                 raise TypeError('b must be an element of ' + str(parent.vector_space()))
             parent._element_constructor_check(A, b)
         super().__init__(parent)
+        if not A.is_immutable():
+            from copy import copy
```
fchapoton: maybe rather import copy globally in the file ?

tscrim: Indeed, that is better. Done.

src/sage/groups/affine_gps/group_element.py
```diff
@@ -45,6 +45,7 @@
 from sage.structure.element import MultiplicativeGroupElement
 from sage.structure.richcmp import richcmp, richcmp_not_equal
 
+from copy import copy
```
mkoeppe: imports from stdlib go to the top of the file, before `from sage...` imports

src/sage/groups/affine_gps/group_element.py
```diff
@@ -497,6 +515,27 @@ def _richcmp_(self, other, op):
...
+
+            sage: F = AffineGroup(3, QQ)
+            sage: g = F([1,2,3,4,5,6,7,8,0], [10,11,12])
+            sage: h = F([1,2,3,4,5,6,7,8,0], [10,11,0])
+            sage: hash(g) == hash(h)
+            False
+            sage: hash(g) == hash(copy(g))
+            True
+            sage: f = g * h
+            sage: hash(f) == hash((f.A(), f.b()))
```
mkoeppe: I don't think it's a good idea to pin this implementation detail in a doctest

tscrim:
I've addressed the reviewer comments (although I disagree with the detail in the doctest; someone who is changing the hash will clearly see the doctest needs to be updated, doctests are not, and should not be, fixed forever).

@fchapoton Could you do the final review here? Should be quick.

src/sage/groups/affine_gps/group_element.py
```diff
@@ -40,12 +40,13 @@
...
 #*****************************************************************************
 
+from copy import copy
+
 from sage.structure.element import Matrix
 from sage.misc.cachefunc import cached_method
 from sage.structure.element import MultiplicativeGroupElement
 from sage.structure.richcmp import richcmp, richcmp_not_equal
 
-
```
fchapoton: two blank lines here please

fchapoton (approved):
ok, good to go

tscrim:
Thank you.

PR #38509, opened 2024-08-14
`build/pkgs/playwright`: New package for screenshotting 3d graphics as png

mkoeppe:
After installing the package using `sage -i playwright`, we can now save threejs renders as png files. Upon first use, the chromium browser is installed automatically.
```
sage: g = polytopes.cube().plot()
sage: g.save('goo.png')
```

- Resolves #33513 

Relevant API:
- https://playwright.dev/python/docs/api/class-browser#browser-new-page
- https://playwright.dev/python/docs/api/class-page#page-screenshot

The ["webpdf" technology of nbconvert](https://nbconvert.readthedocs.io/en/latest/usage.html#convert-webpdf) also currently uses playwright. https://github.com/jupyter/nbconvert/blob/main/pyproject.toml#L66

Discussion: https://groups.google.com/g/sage-devel/c/pj76chTb348

### :memo: Checklist


kiwifb:
Yum python packages all the way. I was afraid of big bad nodejs. I have questions :)
1) does it pop up a window while it is working or it is completely headless?
2) can it use an installed browser or does it have to be a playwright packaged one? I can see why it could need a playwright packaged one, got a bit bitten by firefox snap in ubuntu 22.04 not working with selenium.

mkoeppe:
> * does it pop up a window while it is working or it is completely headless?

Here on macOS, I didn't see a window pop up. Haven't tested on Linux.

> * can it use an installed browser or does it have to be a playwright packaged one? I can see why it could need a playwright packaged one, got a bit bitten by firefox snap in ubuntu 22.04 not working with selenium.

I haven't seen any options to use browsers other than the ones it is providing. I'd also be quite skeptical about any solution that promised doing that.

mkoeppe:
nbconvert installs chromium at runtime. 
https://github.com/jupyter/nbconvert/blob/55ff3e9ac1d892165bfb4352379672019b66ddff/nbconvert/exporters/webpdf.py#L86

We probably should do the same instead of my current solution of using the script package `playwright_chromium`

mkoeppe:
> nbconvert installs chromium at runtime. https://github.com/jupyter/nbconvert/blob/55ff3e9ac1d892165bfb4352379672019b66ddff/nbconvert/exporters/webpdf.py#L86
> 
> We probably should do the same

Done now

mkoeppe:
Doesn't show in HTML documentation

PR #38511, opened 2024-08-14
fixed fast_callable()  missing INPUT and OUTPUT docs

SyedAyaanAli6786:
PR resolves #32277

### :memo: Checklist

- [ x] I have created tests covering the changes.
- [ x] I have updated the documentation and checked the documentation preview.

### DIFF
--- a/src/doc/en/reference/documentation/conf.py
+++ b/src/doc/en/reference/documentation/conf.py
@@ -1,1 +1,55 @@
 ../conf_sub.py
+
+
+def fast_callable(expression, vars=None, domain=None, hold=False, use_cache=False):
+    """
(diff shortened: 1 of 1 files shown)


src/doc/en/reference/documentation/conf.py
```diff
@@ -1 +1,55 @@
-../conf_sub.py
\ No newline at end of file
+../conf_sub.py
+
+
+def fast_callable(expression, vars=None, domain=None, hold=False, use_cache=False):
```
mkoeppe, line 4: That's not the right place to add it.

Look for the definition of `fast_callable`, for example, by typing `git grep 'def fast_callable' `. Change it there.

Issue #34892, opened 2023-01-04
Fixing the linter (again)


fchapoton:
<div id="comment:0"></div>

by removing all trailing whitespaces recently introduced

Depends on #34839

CC:  @MatteoCati @mkoeppe

Component: **refactoring**

Author: **Frédéric Chapoton**

Branch/Commit: **[`c41e4d7`](https://github.com/sagemath/sagetrac-mirror/commit/c41e4d78f268671ce9dcd533f811ba7ffe710109)**

Reviewer: **Matthias Koeppe**

_Issue created by migration from https://trac.sagemath.org/ticket/34892_

fchapoton:
Branch: **[u/chapoton/34892](https://github.com/sagemath/sagetrac-mirror/tree/u/chapoton/34892)**

fchapoton:
<div id="comment:1"></div>

New commits:
<table><tr><td><a href="https://github.com/sagemath/sagetrac-mirror/commit/7b4aa1c7cf2557eb1a858535cab72824f087cc10"><code>7b4aa1c</code></a></td><td><code>full pep8 fix for t-sequences</code></td></tr></table>

fchapoton:
Commit: **[`7b4aa1c`](https://github.com/sagemath/sagetrac-mirror/commit/7b4aa1c7cf2557eb1a858535cab72824f087cc10)**

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
<div id="comment:2"></div>

Branch pushed to git repo; I updated commit sha1. New commits:
[...]

PR #38524, opened 2024-08-18
Sort #34892 doctest output vectors to fix randomness

grhkm21:
Reported at https://github.com/sagemath/sage/pull/38492#issuecomment-2295021400 by @jhpalmieri and @GMS103. Added an extra test because why not.

### DIFF
--- a/src/sage/modules/free_quadratic_module_integer_symmetric.py
+++ b/src/sage/modules/free_quadratic_module_integer_symmetric.py
@@ -1537,11 +1537,11 @@ class FreeQuadraticModule_integer_symmetric(FreeQuadraticModule_submodule_with_b
             sage: L = IntegralLattice('A4')
             sage: t = vector([1.2, -3/11, 5.5, -9.1])
             sage: short = L.enumerate_short_vectors()   # implicit doctest
-            sage: [next(short) for _ in range(10)]
+            sage: [next(short) for _ in range(10)]  # random
             [(0, 0, 0, 1), (0, 0, 1, 1), (0, 1, 1, 1), (1, 1, 1, 1), (0, 0, 1, 0),
              (1, 1, 1, 0), (0, 1, 1, 0), (0, 1, 0, 0), (1, 1, 0, 0), (1, 0, 0, 0)]
             sage: close = L.enumerate_close_vectors(t)  # implicit doctest
-            sage: [next(close) for _ in range(10)]
+            sage: [next(close) for _ in range(10)]  # random
             [(1, 0, 6, -9), (1, -1, 5, -9), (2, 0, 6, -9), (1, 0, 5, -9), (1, -1, 5, -10),
              (2, 1, 6, -9), (1, 0, 5, -10), (2, 0, 5, -9), (1, 0, 6, -8), (1, -1, 6, -9)]
         """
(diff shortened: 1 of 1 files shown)


grhkm21:
By the way, is this intended?
```python
sage: L = IntegralLattice(Matrix(QQ, [[1000, 0], [0, 1]]))
....: L.0.norm()
1
```

yyyyx4:
> By the way, is this intended?
> 
> ```python
> sage: L = IntegralLattice(Matrix(QQ, [[1000, 0], [0, 1]]))
> ....: L.0.norm()
> 1
> ```

This is because the given matrix is interpreted as the Gram matrix of the lattice: The basis vectors are always `(1,0)` and `(0,1)`, and their pairwise inner products are specified by the matrix entries. To compute the norm, you'd have to do `L.0 * L.gram_matrix() * L.0`, I suppose. Perhaps there's a shortcut.

yyyyx4:
Indeed, presumably this is not optimal. The `parent` of this vector is set to the correct lattice, but the `.norm()` method behaves the same as for the standard Euclidean lattice regardless of the actual parent. This seems like something worth fixing.

grhkm21:
Yeah that's what I was thinking, `v.parent()` gives the lattice

yyyyx4 (commented):
Your patch (thanks!) will fix the doctest failures, but it also leaves us with tests which basically only check that the method doesn't crash. I'm not sure how to best resolve this. Practically, perhaps we can get away with a `sorted()` and deal with it properly if/when it breaks again on some other setup?

jhpalmieri:
If you don't want to use `sorted`, you could check membership of some entries. Do you know whether `(3, 2, -2, -4)` will alway be among the first 20 entries, for example? I don't know the code at all. Are there any meaningful tests you can run on the output?

> > By the way, is this intended?
> > ```python
> > sage: L = IntegralLattice(Matrix(QQ, [[1000, 0], [0, 1]]))
> > ....: L.0.norm()
> > 1
> > ```
> 
> This is because the given matrix is interpreted as the Gram matrix of the lattice: The basis vectors are always `(1,0)` and `(0,1)`, and their pairwise inner products are specified by the matrix entries. To compute the norm, you'd have to do `L.0 * L.gram_matrix() * L.0`, I suppose. Perhaps there's a shortcut.

Why does the second line start with `....:` instead of `sage:`?

grhkm21:
For now, I will add change it to a `sorted` call, and if it breaks in any other architecture then ~~too bad~~ we can figure out from there, perhaps testing membership of a short vector as suggested.

jhpalmieri:
> > > > By the way, is this intended?
> > > > ```python
> > > > sage: L = IntegralLattice(Matrix(QQ, [[1000, 0], [0, 1]]))
> > > > ....: L.0.norm()
> > > > 1
> > > > ```
> > > 
> > > 
> > > This is because the given matrix is interpreted as the Gram matrix of the lattice: The basis vectors are always `(1,0)` and `(0,1)`, and their pairwise inner products are specified by the matrix entries. To compute the norm, you'd have to do `L.0 * L.gram_matrix() * L.0`, I suppose. Perhaps there's a shortcut.
> > 
> > 
> > Why does the second line start with `....:` instead of `sage:`?
> 
> Because I copied this from my REPL and it's not a doctest.

Sorry, I thought you were quoting from the code.

grhkm21:
I'll sort it by the norm *and* the original vector then

yyyyx4 (approved):
Looks good, thanks!

PR #38535, opened 2024-08-20
fix issue 38527 with `spqr_tree_to_graph`

### LINKED ISSUE #38527: graphs: spqr_tree with algorithm="Hopcroft_Tarjan" (default parameter) is buggy as it relies on vertices being sorted
### Steps To Reproduce

To trigger the bug, one must first apply the following patch to sage:
```
diff --git a/src/sage/graphs/base/static_sparse_backend.pyx b/src/sage/graphs/base/static_sparse_backend.pyx
index ea65e2117d9..178aa8cf82a 100644
--- a/src/sage/graphs/base/static_sparse_backend.pyx
+++ b/src/sage/graphs/base/static_sparse_backend.pyx
@@ -511,10 +511,6 @@ cdef class StaticSparseBackend(CGraphBackend):
             True
         """
         vertices = list(G)
-        try:
-            vertices.sort()
-        except TypeError:
-            pass
         cdef StaticSparseCGraph cg = <StaticSparseCGraph> StaticSparseCGraph(G, vertices)
         self._cg = cg
 ```

This reverts some changes from #26447 that were introduced to fixed a failing doctest in `connectivity.pyx` due to `spqr_tree`. It indeed fixed the doctest but it hides the fact that the function is buggy, as it relies on the fact that the vertices of static graphs are sorted and it added a extra `n*log(n)` complexity to all algorithms that use static graphs (via `(Di)Graph(immuatable=True)`. This extra cost is not documented and invalidates most complexity claims of algorithms using static graphs.

Then, one can trigger the bug with the following code:

```py
sage: from sage.graphs.connectivity import spqr_tree, spqr_tree_to_graph
sage: G = Graph('LlCG{O@?GBoMw?')
sage: T1 = spqr_tree(G, algorithm="Hopcroft_Tarjan")
sage: T2 = spqr_tree(G, algorithm="cleave")
sage: G1 = spqr_tree_to_graph(T1)
sage: G2 = spqr_tree_to_graph(T2)
sage: G.is_isomorphic(G1)
sage: G.is_isomorphic(G2)
```

### Expected Behavior

```py
True
True
```

### Actual Behavior

```py
False
True
```

### Additional Information

Disclaimer: I did not know of SPQR tree before looking at this code so I may write some stupid things.

The call to `spqr_tree` with `algorithm="cleave"` is ok with and without the code that sort the vertices in `StaticSparseBackend`. The call with `algorithm="Hopcroft_Tarjan"` returns a tree that cannot be used to rebuild the original graph when the vertices are not sorted in `StaticSparseBackend`.
After a quick look at the returned trees, it seems to come from the fact the edges of the graphs that correspond to vertices of the spqr tree are either represented as (u,v, label) or (v, u, label). 
For example the following code shows that the edges between 0 and 2 is sometimes stored as (0, 2) and sometimes as (2, 0):
```py
for v in T1:
  for e in v[1].edge_iterator():
    if (e[0], e[1]) in [(0, 2), (2, 0)]:
      print (e)
print ("-")
for v in T2:
  for e in v[1].edge_iterator():
    if (e[0], e[1]) in [(0, 2), (2, 0)]:
      print (e)
```
output:
```py
(0, 2, 'newVEdge11')
(2, 0, 'newVEdge11')
-
(0, 2, None)
(0, 2, None)
```
It is not the case when the vertices of `StaticSparseBackend` are sorted (ie when the above patch is not apply):
```py
(0, 2, 'newVEdge11')
(0, 2, 'newVEdge11')
-
(0, 2, None)
(0, 2, None)
```

(Note that the labels are different between the two algorithms but it does not seem to matter). 

### Environment

```markdown
- **OS**: debian 11
- **Sage Version**: since sage-9 at least
```

### Checklist

dcoudert:
Fixes #38527. 

This PR makes the comparison of edges more robust in `spqr_tree_to_graph`, so that edges `(u, v, label)` and `(v, u, label)` are no longer considered as distinct edges.

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/connectivity.pyx
+++ b/src/sage/graphs/connectivity.pyx
@@ -2634,10 +2634,27 @@ def spqr_tree_to_graph(T):
         sage: H.is_isomorphic(G)
         True
 
-    TESTS::
+    TESTS:
+
+    Check that the method is working for the empty SPQR tree::
 
         sage: H = spqr_tree_to_graph(Graph())
         sage: H.is_isomorphic(Graph())
+        True
+
+    Check that :issue:`38527` is fixed::
+
+        sage: from sage.graphs.connectivity import spqr_tree, spqr_tree_to_graph
+        sage: G = Graph('LlCG{O@?GBoMw?')
+        sage: T1 = spqr_tree(G, algorithm="Hopcroft_Tarjan")
+        sage: T2 = spqr_tree(G, algorithm="cleave")
+        sage: T1.is_isomorphic(T2)
+        True
+        sage: G1 = spqr_tree_to_graph(T1)
+        sage: G2 = spqr_tree_to_graph(T2)
+        sage: G.is_isomorphic(G1)
+        True
+        sage: G.is_isomorphic(G2)
         True
     """
     from sage.graphs.graph import Graph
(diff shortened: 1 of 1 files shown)


cyrilbouvier (approved):
LGTM

dcoudert:
Can you set this PR to positive review ? Or I can certainly do it on your behalf if you agree.

cyrilbouvier:
> Can you set this PR to positive review ? Or I can certainly do it on your behalf if you agree.

I do not have the right to add labels to issue, so I cannot add the positive review label.
I agree that you can add it on my behalf.

dcoudert:
I have put it in a positive review on your behalf. Thank you for raising this issue and for the review.

PR #38539, opened 2024-08-20
some details in strassen.pyx

fchapoton:
just a few minor details in the code of the file `strassen.pyx`

including a wrong float-division by 2

### :memo: Checklist


src/sage/matrix/strassen.pyx
```diff
@@ -733,11 +731,9 @@ class int_range:
             sage: J - I
             [(6, 1), (20, 4)]
         """
-        all = self.to_list()
-        for i in right.to_list():
-            if i in all:
-                all.remove(i)
-        return int_range(all)
+        right_list = right.to_list()
```
dcoudert: unless elements of `right_list` are not hashable, it would be better to make it a set to speed up tests in next line.

src/sage/matrix/strassen.pyx
```diff
@@ -555,7 +559,7 @@ class int_range:
         if indices is None:
             self._intervals = []
             return
-        elif range is not None:
+        if range is not None:
```
dcoudert, line 562: variable `range` will have to be renamed at some point.

src/sage/matrix/strassen.pyx
```diff
@@ -757,17 +753,16 @@ class int_range:
             sage: J * I
             [(4, 2)]
         """
-        intersection = []
         all = self.to_list()
-        for i in right.to_list():
-            if i in all:
-                intersection.append(i)
+        intersection = [i for i in right.to_list() if i in all]
```
dcoudert: a faster way to test `i in all` might be better. Either make `all` a set or use binary search ?

src/sage/matrix/strassen.pyx
```diff
@@ -757,17 +753,16 @@ class int_range:
             sage: J * I
             [(4, 2)]
         """
-        intersection = []
         all = self.to_list()
```
dcoudert: variable `all` should be renamed to something different from Python functions.

src/sage/matrix/strassen.pyx
fchapoton: I have left this one as it is, because it appears as a named function argument.

dcoudert: Indeed, it would need a `@rename_keyword` warning.

dcoudert (approved):
LGTM.

fchapoton:
back to positive after trivial merge

dcoudert:
Indeed, but I think it has nothing to do with this PR.

PR #38542, opened 2024-08-21
Fixes in pyx files

fchapoton:
This is fixing a few issues about division in various cython files.

All the fixes were found using the command `cython -a -3 *.pyx`.

### :memo: Checklist


fchapoton:
@kedlaya : could you please have a look at the 2 failing doctests in hypergeometric motives ? I wonder if the old results were correct.

roed314:
I just checked in Magma, which agrees with Sage's old results (note that there's a `t <-> 1/t` difference in convention, thus the `1/2`):
```
H := HypergeometricData([[3],[4]]);
EulerFactor(H,1/2,7);
7*$.1^2 - 3*$.1 + 1
```

Indeed, 21 is too big since it needs to be a weight 1 Weil polynomial.

roed314:
I haven't looked at the code in detail, but I wonder why you even want to change the signature here.  It doesn't seem to relate to the C division convention.

fchapoton:
Thanks. So my changes are not good. I will have to investigate. Probably I will undo the changes here about hypergeometric motives.

fchapoton:
I wanted to fix the warning displayed by "cython -a -3 hypergeometric_misc.pyx".

kedlaya:
Off the top of my head I am guessing that changing `p` to `unsigned long long` is changing the behavior of `%` down the line. This is of course fixable, let me have a look.

fchapoton:
I have tried to make a smaller change in hgm,  and this seems to work. Not sure that it appeases "cython -a -3" buit this is maybe not important

kedlaya:
> I have tried to make a smaller change in hgm, and this seems to work. Not sure that it appeases "cython -a -3" buit this is maybe not important

On my end changing only `f` to be `unsigned int` seems to fix the warning without disturbing the doctest (which makes sense as the issue is with raising an integer to a possibly negative integer power).

fchapoton:
Thanks. So this is now ready for review.

fchapoton:
please review, if you have time

src/sage/combinat/designs/subhypergraph_search.pyx
```diff
@@ -180,7 +180,7 @@ cdef hypergraph h_init(int n,list H) noexcept:
     cdef hypergraph h
     h.n          = n
     h.m          = len(H)
-    h.limbs      = (n+63)/64 # =ceil(n/64)
+    h.limbs      = (n+63)//64 # =ceil(n/64)
```
dcoudert: 2 spaces before #

src/sage/schemes/elliptic_curves/mod_sym_num.pyx
```diff
@@ -456,24 +456,24 @@ cdef int best_proj_point(llong u, llong v, llong N,
     else:  # cases like (p:q) mod p*q drop here
         p = llgcd(u, N)
         q = llgcd(v, N)
-        Nnew = N / p / q
-        w = ((u/p) * llinvmod(v/q, Nnew)) % Nnew
-        y0 = N/q
+        Nnew = N // p // q
```
dcoudert: I know it's ok, but isn't it better for readability to write `(N // p) // q`?

dcoudert (changes requested):
otherwise LGTM.

fchapoton:
merci ! Voilà !

dcoudert (approved):
LGTM.

PR #38446, opened 2024-07-29
Implement basic multivariate polynomial species

Newtech66:
Related to #30727.

We implement basic functionality for multivariate polynomial species, using its representation as a pair of a permutation group and a mapping between the domain of the permutation group and some variables. We provide addition, multiplication, and (partitional) composition (for some special cases). We also allow it to be constructed as a group action (or a sequence thereof). Atomic and molecular decompositions are automatically computed thanks to #38371.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38544, opened 2024-08-21
Lazy species

mantepse:
We provide an implementation of combinatorial species based on the lazy series framework and #38446.

dependencies: #38974


src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,486 @@
...
+            [((1, 2, 3), E_3), ((1, 2, 3), E_3)]
+
+        """
+        def add_structures(labels):
+            yield from self.structures(labels)
+            yield from other.structures(labels)
+
+        result = super()._add_(other)
+        result.structures = add_structures
+        return result
```
mantepse: @tscrim, I'm not sure whether this is the best way to do it.  I tried to create a separate class `SumSpeciesElement` inheriting from `LazySpeciesElement` that only overrides the method `structures` (and will override `generating_series`, etc.), but got stuck.

I think I would like to turn the result of `super()._add_(other)` into a `SumSpeciesElement` instance.

mantepse: One other thing: I am not sure, but perhaps it would actually be nice to take some decisions of `LazyModuleElement._add_` into account.  For example, if one of the arguments is zero, I might not want to return a SumSpeciesElement, but rather the other summand.

However, this smells badly like duplicating a lot of code.

mantepse: Rethinking this: I guess it happens next to never that we are in one of the cases treated specially by `LazyModuleElement._add_ ` (or `_mul_`, etc.) So maybe my afterthought is thinking about the empty set.

mantepse: Does the following look correct to you?  At least it works.  I admit that the arguments `super` requires are not super-clear to me.
```
class LazySpeciesElement(LazyCompletionGradedAlgebraElement):
...
    def _add_(self, other):
        return SumSpeciesElement(self, other)
...
class SumSpeciesElement(LazySpeciesElement):
    def __init__(self, left, right):
        self._left = left
        self._right = right
        coeff_stream = super(LazySpeciesElement, type(left))._add_(left, right)._coeff_stream
        super().__init__(self.parent(), coeff_stream)

    def structures(self, labels):
        yield from self._left.structures(labels)
        yield from self._right.structures(labels)
```

mantepse:
This is now mostly functional.  The next decision is how to go about replacing the old species framework - @fchapoton, it would be great to have your input on this!

There is some functionality which is definitively missing.  I know of the following:

1. relabelling of structures
2. generating isotypes
3. the idea of `SpeciesStructure`, which provides a parent for structures and may make it easier to manipulate them
4. functorial composition
5. obtain a system of equations for a recursively defined species

1-3 are probably very easy, I think that 5 does not really belong to the species code, but should rather be a feature of the lazy framwork. Most likely, there is more.

More questions:

* should we try to transfer the doctests into the new framework?
* should we split the file into more components? If so, along which lines?

mantepse:
I realize that I have no idea whether a species structure should have any structure at all.  I see the following two options:
1. `structures` indeed returns just arbitrary objects, only the species knows how to relabel a structure and how to check whether two structures are isomorphic.
1. `structures` returns `SpeciesStructure`s.  Such a `SpeciesStructure`
    - knows the species which produced it
    - knows its labels
    - knows how to print itself
    - knows how to relabel itself
    - can test whether it is isomorphic to another structure
    - can possibly return a canonical labelling of itself.

Special species classes, such as `SumSpecies`, `ProductSpecies` and `CompositionSpecies`, may provide additional methods, for example it could know the summands, factors, etc.

The current approach is the second one, and I'm leaning towards it, although this doesn't quite follow the math.  I think we could additionally add a `facade` option to the `structures` method or the species constructor itself.

mantepse:
I am encountering yet another design question, @tscrim, I think that's for you :-)

I added a method `restrict(self, min_degree=None, max_degree=None)` to `LazyModuleElement`, which removes all elements of degree smaller than `min_degree` and larger than `max_degree`, similar to `truncate`.

Now consider a "special" lazy species, such as `GraphSpecies(LazySpeciesElement)` or `SetSpecies(LazySpeciesElement)`.  One reason to have them is their special implementation of `structures` and `isotypes`, which should yield graphs or sets, rather than the generic coset representatives corresponding to the molecular decomposition.  For example, in my private branch I have

```
sage: from sage.rings.lazy_species import LazySpecies
sage: L = LazySpecies(QQ, "X")
sage: P = L.SetPartitions()
sage: list(P.structures([1,2,3]))
[{{1, 2, 3}}, {{1, 2}, {3}}, {{1, 3}, {2}}, {{1}, {2, 3}}, {{1}, {2}, {3}}]
```
What I'd like to have is that restriction preserves this, but of course, it doesn't:
```
sage: list(P.restrict(3, 3).structures([1,2,3]))
[(E_3, ((1, 2, 3),), 0),
 (E_3, ((1, 2, 3),), 1),
 (X*E_2, ((2,), (1, 3))),
 (X*E_2, ((3,), (1, 2))),
 (X*E_2, ((1,), (2, 3)))]
```
I think that `restrict` *cannot* return an instance of `SetPartitionSpecies` because, the way I defined it, `SetPartitionSpecies` is the species of set partitions of arbitrary finite sets.

I guess we'll have to have a video call to sort this out, I'm afraid my explanation is as unclear as possible :-)

mantepse:
Here is the next interesting question.  I would like to make `define_implicitly` work.  This is tricky if weights and composition is involved (it is very likely that the problem exists also with `LazySymmetricFunctions`.

Consider the species of sets $\mathcal E$ and denote the species of sets of size $n$ with $\mathcal E_n$.  I know no good reference for this, but there are two different kinds of elements in a $\lambda$-ring, monomials and constants.  Suppose that $q$ is a monomial in the $\lambda$-ring.  We then have, for example, $\mathcal E((1+q)X)_3 = (1 + q^3) \mathcal E_3 + (q + q^2) X \mathcal E_2$.  To illustrate further, $\mathcal E(3 X)_3 = 3\mathcal E_3 + 6 X \mathcal E_2 + X^3$.

The implicit solver uses (essentially) the infinite polynomial ring to represent "undetermined" coefficients.  For example, in degree 2, an undetermined species would be $f_0 X^2 + f_1 \mathcal E_2$, because $X^2$ and $\mathcal E_2$ are a basis for species of degree 2.  Unfortunately, this idea fails with composition.  Right now, I don't see a way to compute, for example, $\mathcal E(f_0 X)_3$ correctly.  The coefficients of $\mathcal E_3$, $X \mathcal E_2$ and $X^3$ depend polynomially on $f_0$.

mantepse:
https://github.com/sagemath/sage/pull/38544#issuecomment-2481392240

This is what I must decide next, I think.

mantepse:
Note to myself: Labelle 2013 has an explicit expression for molecular expansion of the combinatorial logarithm, i.e., `L.Sets().restrict(1).revert()`, and also for compositions of this with an arbitrary species.

I don't know yet how to best integrate this into the framework.  Maybe `revert` could detect this special case, or we could learn from this special case how to make `revert` fast.

mantepse:
Apparently, I need to do something so `LazySpecies` is included in the documentation, but I don't know what.

fchapoton:
you need to add them in some index file in src/doc

```
git grep -c "species" src/doc/en/reference/
src/doc/en/reference/combinat/module_list.rst:19
src/doc/en/reference/references/index.rst:2
```
so maybe in the second file ?

mantepse:
> you need to add them in some index file in src/doc
> 
> ```
> git grep -c "species" src/doc/en/reference/
> src/doc/en/reference/combinat/module_list.rst:19
> src/doc/en/reference/references/index.rst:2
> ```
> 
> so maybe in the second file ?

Should I put it into the `combinat` documentation?  I admit that species are mostly interesting to combinatorialists.  Does it matter that the module itself is in `rings`?

fchapoton:
it does not matter that the module is in rings. I suggest to put a link to it in

https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/species/all.html#sage-combinat-species-all

so in

src/sage/combinat/species/all.py

(which is not where I said  before, sorry)

mantepse:
I built the pdf locally.  It took forever, but it worked.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+- Mainak Roy, Martin Rubey, Travis Scrimshaw (2024-2025)
+
+EXAMPLES:
+
+    We can reproduce the molecular expansions from Appendix B in
+    [GL2011]_ with little effort.  The molecular expansion of the
+    species of point determining graphs can be computed as the
+    species of graphs composed with the compositional inverse of the
+    species of non-empty sets.::
```
tscrim:
```suggestion
EXAMPLES:

    We can reproduce the molecular expansions from Appendix B in
    [GL2011]_ with little effort.  The molecular expansion of the
    species of point determining graphs can be computed as the
    species of graphs composed with the compositional inverse of the
    species of non-empty sets::
```
Everything before `EXAMPLES:` in this docstring is overindented by 4 spaces.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+    bi-point-determining graphs we use Corollary (4.6) in
+    [GL2011]_::
+
+        sage: B = G(2*Ep.revert() - X)
+        sage: B.truncate(6)
+        1 + X + E_2(X^2) + (P_5+5*X*E_2(X^2))
+
+"""
+from sage.functions.other import binomial, factorial
+from sage.misc.lazy_list import lazy_list
```
tscrim:
```suggestion
        1 + X + E_2(X^2) + (P_5+5*X*E_2(X^2))
"""

from sage.functions.other import binomial, factorial
from sage.misc.lazy_list import lazy_list
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+    The weight of a composition `n_1, n_2, \dots` is `\sum_i w_i n_i`.
+
+    INPUT:
+
+    - ``n`` -- a nonnegative integer, the sum of the parts
+    - ``d`` -- a nonnegative integer, the total weight
+    - ``weight_multiplicities`` -- an iterable,
+      ``weight_multiplicities[i]`` is the number of positions with
+      weight `i+1`.
```
tscrim:
```suggestion
    - ``n`` -- nonnegative integer; the sum of the parts
    - ``d`` -- nonnegative integer; the total weight
    - ``weight_multiplicities`` -- iterable;
      ``weight_multiplicities[i]`` is the number of positions with
      weight ``i+1``
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+        sage: list(weighted_compositions(3, 4, [2,2]))
+        [[2, 0, 1, 0],
+         [1, 1, 1, 0],
+         [0, 2, 1, 0],
+         [2, 0, 0, 1],
+         [1, 1, 0, 1],
+         [0, 2, 0, 1]]
+
+    """
```
tscrim:
```suggestion
         [0, 2, 0, 1]]
    """
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+            return
+    except TypeError:
+        pass
+    if _w0 > d:
+        return
+    for s in range(n + 1):
+        for c in weighted_compositions(n - s, d - s * (_w0 + 1), weight_multiplicities, _w0=_w0+1):
+            m = weight_multiplicities[_w0]
+            for v in map(list, IntegerVectors(s, length=m)):
+                yield v + c
```
tscrim:
```suggestion
    from sage.combinat.integer_lists.invlex import IntegerListsBackend_invlex
    for s in range(n + 1):
        for c in weighted_compositions(n - s, d - s * (_w0 + 1), weight_multiplicities, _w0=_w0+1):
            m = weight_multiplicities[_w0]
            for v in IntegerListsBackend_invlex(s, length=m)._iter():
                yield v + c
```
This should be faster without heavy parent/element objects.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+         ([3, 0], [0, 1]),
+         ([2, 1], [1, 0]),
+         ([2, 1], [0, 1]),
+         ([1, 2], [1, 0]),
+         ([1, 2], [0, 1]),
+         ([0, 3], [1, 0]),
+         ([0, 3], [0, 1])]
+    """
+    k = len(n_vec)
+    for d_vec in IntegerVectors(d, length=k):
```
tscrim:
```suggestion
    from sage.combinat.integer_lists.invlex import IntegerListsBackend_invlex
    for d_vec in IntegerListsBackend_invlex(d, length=k)._iter():
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+            sage: L = LazySpecies(ZZ, "X")
+            sage: E = L(lambda n: SymmetricGroup(n))
+            sage: F = L(lambda n: SymmetricGroup(n))
+            sage: list(E.structures([1,2,3]))
+            [(E_3, ((1, 2, 3),))]
+            sage: list((E+F).structures([1,2,3]))
+            [(E_3, ((1, 2, 3),)), (E_3, ((1, 2, 3),))]
+
+        """
```
tscrim:
```suggestion
            [(E_3, ((1, 2, 3),)), (E_3, ((1, 2, 3),))]
        """
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+        TESTS::
+
+            sage: from sage.rings.species import PolynomialSpecies
+            sage: L = LazySpecies(QQ, "X, Y")
+            sage: P = PolynomialSpecies(QQ, "X, Y")
+            sage: XY = L(P(PermutationGroup([], domain=[1, 2, 3]), {0: [1], 1: [2, 3]}))
+            sage: XY._test_structures()
+        """
+        n = 3
```
tscrim: Wouldn't it be better to make this `n` something that is passed in with `**options` from the tester?

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+        P = self.parent()
+        if P._arity == 1:
+            labels = list(range(n))
+            s = list(self.structures(labels))
+            assert len(s) == len(set(s)), f"structures for {labels} are {s}, which is not a set"
```
tscrim:
```suggestion
            tester.assertEqual(len(s), len(set(s)), f"structures for {labels} are {s}, which is not a set")
```
and similar such changes.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+    def polynomial(self, degree=None, names=None):
+        r"""
+        Return ``self`` as a polynomial if ``self`` is actually so.
+
+        INPUT:
+
+        - ``degree`` -- ``None`` or an integer
+        - ``names`` -- names of the variables; if it is ``None``, the name of
+          the variables of the series is used
```
tscrim:
```suggestion
        - ``degree`` -- (optional) integer
        - ``names`` -- (default: name of the variables of the series) names of the variables
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+            shapes = IntegerVectors(n, length=P._arity)
+            for shape in shapes:
+                s = list(self.isotypes(*shape))
+                assert len(s) == len(set(s)), f"isotypes for {shape} are {s}, which is not a set"
+                coeff = self.isotype_generating_series()[n].coefficient(list(shape))
+                assert len(s) == coeff, f"the number of isotypes for {shape} is {len(s)}, but the generating series gives {coeff}"
+
+    def polynomial(self, degree=None, names=None):
+        r"""
+        Return ``self`` as a polynomial if ``self`` is actually so.
```
tscrim:
```suggestion
        Return ``self`` as a polynomial if ``self`` is actually so or up to
        specified degree.
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+        If ``degree`` is not ``None``, the terms of the series of
+        degree greater than ``degree`` are first truncated.  If
+        ``degree`` is ``None`` and the series is not a polynomial
+        polynomial, a ``ValueError`` is raised.
+
+        EXAMPLES::
+
+            sage: L = LazySpecies(ZZ, "X")
+            sage: E = L(lambda n: SymmetricGroup(n))
```
tscrim:
```suggestion
            sage: E = L(SymmetricGroup)
```
A bit more concise way of constructing things. :)

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+                raise ValueError("not a polynomial species")
+        else:
+            m = degree + 1
+
+        return R.sum(self[:m])
+
+    def __call__(self, *args):
+        """
+
+        EXAMPLES::
```
tscrim:
```suggestion
    def __call__(self, *args):
        """
        Evaluate ``self`` at ``*args``.

        EXAMPLES::
```

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+            sage: L.<X> = LazySpecies(QQ)
+            sage: L.Graphs().generating_series().truncate(7)
+            1 + X + X^2 + 4/3*X^3 + 8/3*X^4 + 128/15*X^5 + 2048/45*X^6
+        """
+        P = self.parent()
+        L = LazyPowerSeriesRing(P.base_ring().fraction_field(),
+                                P._laurent_poly_ring._indices._indices.variable_names())
+        cis = self.cycle_index_series()
+        return L(lambda n: cis[n].coefficient(Partition([1]*n)))
```
tscrim:
```suggestion
        cis = self.cycle_index_series()
        return L(lambda n: cis[n].coefficient(_Partitions([1]*n)))
```
A bit faster.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+
+class SumSpeciesElement(LazySpeciesElement):
```
tscrim: Am I correct in that we cannot use the streams for this because when we work with finite species, then it would lose this information (because it would then create a new exact stream)? Or is there another reason for not extracting the data from the streams?

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+        weight_exp = [lazy_list(lambda j, g=g: len(coeff(g, j+1)))
+                      for g in args]
+
+        def flat(g):
+            # function needed to work around python's scoping rules
+            return itertools.chain.from_iterable((coeff(g, j) for j in itertools.count()))
+
+        args_flat1 = [lazy_list(flat(g)) for g in args]
+
+        def coefficient(n):
```
tscrim, line 988: I think it would be better to make this an `@staticmethod` of the class. You will have to document it, but then it is documented and tested separately. It also is better under equality checks and pickling IIRC.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+            sage: LazySpecies(QQ, "X, Y, Z")._arity
+            3
+        """
+        super().__init__(PolynomialSpecies(base_ring, names))
+        self._arity = len(names)
+        if self._arity == 1:
+            self.Graphs = lambda: GraphSpecies(self)
+            self.SetPartitions = lambda: SetPartitionSpecies(self)
+            self.Sets = lambda: SetSpecies(self)
+            self.Cycles = lambda: CycleSpecies(self)
```
tscrim: As annoying as it might be, I think this class should be split into the uni- and multi-arity cases. Then all of these examples can be documented, tested, and more easily found by a user.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1395 @@
...
+
+        EXAMPLES::
+
+            sage: P.<X> = LazySpecies(QQ)
+            sage: p = P.SetPartitions()
+            sage: list(p.isotypes(3))
+            [[3], [2, 1], [1, 1, 1]]
+        """
+        if labels in ZZ:
+            yield from Partitions(labels)
```
tscrim, line 1668:
```suggestion
            yield from _Partitions(labels)
```

tscrim (changes requested):

src/sage/rings/lazy_species.py
mantepse: Can I tell the tester then to test `for n in range(4)`?  we really should test all of these four values, as I discovered the hard way in another pull request :-)

src/sage/rings/lazy_species.py
mantepse: I'm not sure I understand your question correctly.  If I do, then #40163 probably answers your question.

src/sage/rings/lazy_species.py
mantepse: I certainly want that `L.<X> = LazySpecies(QQ)` and `L.<X,Y> = LazySpecies(QQ)` works.  How would I do this with two separate classes?

src/sage/rings/lazy_species.py
mantepse: It would have to take `left`, `args_flat`, `R`, `L`, `gv`, `weight_exp` as arguments, right?  That sounds a bit hard to test.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1400 @@
...
+
+        EXAMPLES::
+
+            sage: L = LazySpecies(ZZ, "X")
+            sage: E = L(SymmetricGroup)
+            sage: F = L(SymmetricGroup)
+            sage: list(E.structures([1,2,3]))
+            [(E_3, ((1, 2, 3),))]
+            sage: list((E+F).structures([1,2,3]))
+            [(E_3, ((1, 2, 3),)), (E_3, ((1, 2, 3),))]
```
mantepse: This is wrong, actually.

src/sage/rings/lazy_species.py
tscrim: I think there is a way to pass these sorts of options via, e.g., `TestSuite(foo).run(structures_sizes=[1,2,3])` and then extract that from the constructed tester. I will check today for how exactly to do this (if it is possible).

src/sage/rings/lazy_species.py
tscrim: It seems like these elements are duplicating what we have done with the streams (e.g., we have an addition and multiplication stream). Is there something preventing you from just using the data of the streams?

src/sage/rings/lazy_species.py
tscrim: Ah, right, it uses a lot of data constructed elsewhere in the method.

src/sage/rings/lazy_species.py
tscrim: You would do something like this:
```python
class LazySpecies(…):
    @staticmethod
    def __classcall_private__(cls, …):
        if amity == 1:
            return LazySpeciesUnivariate(…)
        return LazySpeciesMultivariate(…)

    Other common methods

class LazySpeciesUnivariate(LazySpecies):
    Univariate methods

class LazySpeciesMultivariate(LazySpecies):
    Multivariate methods
```
(We use this pattern in many different places, including (essentially) polynomials.)

src/sage/rings/lazy_species.py
mantepse: Yes, I think so.  Please excuse me if the following is obvious to you.

A lazy species is represented as a stream of homogeneous polynomial species.  While in principle, we can derive all information from this stream, this is often infeasible in practice.  For example, if we only want to get the generating series, it is enough to know the generating series of the summands.

Put differently, given an instance `F` of `LazySpeciesElement`, it might be impossible to obtain `F[20]`, but very easy to obtain `F.generating_series()[100]`.

In the current PR, this possibility is not yet implemented, which is actually a bug.  In any case, #40163 puts the setup created here to use.

src/sage/rings/lazy_species.py
mantepse: I am really a bit unsure whether this is a good thing to do at this point.  Like so?  (tests pass)
```diff
diff --git a/src/sage/rings/lazy_species.py b/src/sage/rings/lazy_species.py
index b018355c1bd..543a159936b 100644
--- a/src/sage/rings/lazy_species.py
+++ b/src/sage/rings/lazy_species.py
@@ -1083,7 +1083,9 @@ class LazySpecies(LazyCompletionGradedAlgebra):
         """
         from sage.structure.category_object import normalize_names
         names = normalize_names(-1, names)
-        return super().__classcall__(cls, base_ring, names, sparse)
+        if len(names) == 1:
+            return LazySpeciesUnivariate(base_ring, names, sparse)
+        return LazySpeciesMultivariate(base_ring, names, sparse)
 
     def _first_ngens(self, n):
         r"""
@@ -1126,11 +1128,45 @@ class LazySpecies(LazyCompletionGradedAlgebra):
         """
         super().__init__(PolynomialSpecies(base_ring, names))
         self._arity = len(names)
-        if self._arity == 1:
-            self.Graphs = lambda: GraphSpecies(self)
-            self.SetPartitions = lambda: SetPartitionSpecies(self)
-            self.Sets = lambda: SetSpecies(self)
-            self.Cycles = lambda: CycleSpecies(self)
+
+
+class LazySpeciesUnivariate(LazySpecies):
+
+    @staticmethod
+    def __classcall_private__(cls, base_ring, names, sparse=True):
+        """
+        Normalize input to ensure a unique representation.
+
+        EXAMPLES::
+
+            sage: LazySpecies(QQ, "X")
+            Lazy completion of Polynomial species in X over Rational Field
+        """
+        return super().__classcall__(cls, base_ring, names, sparse)
+
+    def Graphs(self):
+        return GraphSpecies(self)
+    def SetPartitions(self):
+        return SetPartitionSpecies(self)
+    def Sets(self):
+        return SetSpecies(self)
+    def Cycles(self):
+        return CycleSpecies(self)
+
+
+class LazySpeciesMultivariate(LazySpecies):
+
+    @staticmethod
+    def __classcall_private__(cls, base_ring, names, sparse=True):
+        """
+        Normalize input to ensure a unique representation.
+
+        EXAMPLES::
+
+            sage: LazySpecies(QQ, "X, Y")
+            Lazy completion of Polynomial species in X, Y over Rational Field
+        """
+        return super().__classcall__(cls, base_ring, names, sparse)
 
 
 class SetSpecies(LazySpeciesElement):
```

mantepse: Ideally, the documentation provided by `L.Sets?` would be identical to that of `SetSpecies`, right?

src/sage/rings/lazy_species.py
tscrim: Okay, thanks for the explanation (it's hard to see what changes between this and #40163 on GH; perhaps I just don't know the system well enough yet...).

src/sage/rings/lazy_species.py
tscrim: You don't need any `__classcall_private__` in the subclasses since no additional parsing is done. However, yes.

Ideally, you would simply be able to do `Graphs = GraphSpecies`, but then the method invocation doesn't work. However, we can make it work by
```python
class LazySpeciesUnivariate(LazySpecies):
    pass

from types import MethodType
LazySpeciesUnivariate.Sets = MethodType(SetSpecies, LazySpeciesUnivariate)
LazySpeciesUnivariate.Graphs = MethodType(GraphSpecies, LazySpeciesUnivariate)
LazySpeciesUnivariate.SetPartitions = MethodType(SetPartitionSpecies, LazySpeciesUnivariate)
LazySpeciesUnivariate.Cycles = MethodType(CycleSpecies, LazySpeciesUnivariate)
```
From a small experiment, this will work as you would expect and mean you will not have to duplicate the documentation. Although it is slightly ugly and hacky....

tscrim: On the other hand, given that we have to do something like this, it is probably better to just have one `LazySpecies` class and add these on in the `__init__` after checking the arity.

mantepse: Alternatively, I could have a one line standard documentation (i.e., the way it currently is) in the individual classes, and proceed as you proposed originally (i.e., essentially the diff I posted above), where I put proper documentation.  After all, the user won't ever see the individual classes, I think.

Indeed, I think that looks better than what is currently in the PR.  Three tiny questions:

1. Question
```
sage: L = LazySpecies(QQ, "X")
sage: L?
```
gives the following.  I would prefer the initial sentence (after `Init docstring:`) to be something like `The ring of lazy species`.  Similarly for the class `GraphSpecies`, etc.  I.e., no class docstring, and abuse (?) the `__init__` docstring to provide doc about the class.

Would this be against conventions?
```
Signature:      L(self, x=0, *args, **kwds)
Type:           LazySpeciesUnivariate_with_category
String form:    Lazy completion of Polynomial species in X over Rational Field
File:           ~/sage/src/sage/rings/lazy_species.py
Init docstring:
   Initialize the ring of lazy species.

   EXAMPLES:

   We provide univariate and multivariate (mostly known as multisort)
   species:

      sage: LazySpecies(QQ, "X")
      Lazy completion of Polynomial species in X over Rational Field

      sage: LazySpecies(QQ, "X, Y")
      Lazy completion of Polynomial species in X, Y over Rational Field

   In the univariate case, several basic species are provided as
   methods:

      sage: L.<X> = LazySpecies(QQ)
      sage: L.SetPartitions()
      sage: L.Sets()
      sage: L.Cycles()
      sage: L.Graphs()
```
2. Question

Perhaps I'd like to override the `_repr_` of the basic species provided, i.e.,
```
sage: L.<X> = LazySpecies(QQ)
sage: L.Graphs()
Graph species
sage: L.Graphs()[4]
2*E_4 + 2*E_2^2 + 2*X*E_3 + 2*E_2(E_2) + 2*X^2*E_2 + E_2(X^2)
```
The drawback is that one doesn't get the full molecular expansion (i.e., the series representation) anymore.  Not sure.  One advantage is that the output is faster, especially for 'complicated' species like graphs.

3. Question

Currently we have
```
sage: L.Graphs() is L.Graphs()
False
```
I think that's not optimal, do you agree?  (I guess I could simply make `GraphSpecies` inherit from `UniqueRepresentation`)

tscrim: My suggestion has the benefit of getting the full documentation interactively, and I believe in the html doc.

1. As you indicated, we generally have class-level docstrings that have all of the important stuff and then the `__init__` docstring is technical stuff (typically a `TestSuite(foo).run()` test and possibly `INPUT:`s if not useful to the user (e.g., constructed from a parent object)). Of course, you can just use the `__init__` docstring as the class-level docstring (as you said, by not including the latter). Typically that kind of information is at best unhelpful to the user, but it can be counterproductive by making it too technical. If that's they way you want to go, I would proceed with caution.
2. I have no objections. The repr doesn't have to follow the other lazy objects.
3. I agree. Either `UniqueRepresentation` the classes (elements with inherited from UR is not unprecedented). Alternatively, cache it in some way with the parent.

mantepse:
This is now ready for another round.

I have a very rough implementation of functorial composition, this will be yet another PR :-)

mantepse:
@tscrim, should I update the branch?

tscrim:
Up to you, but I don't think it is necessary. I will try to look at this later tonight.

mantepse:
There was a non-deterministic doctest, and one that actually belongs to a different ticket, sorry.

src/sage/rings/lazy_species.py
tscrim: Ah right, this is partitions of `n`; sorry.

However, should this raise an error if `labels not in ZZ`?

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1663 @@
...
+        EXAMPLES::
+
+            sage: L = LazySpecies(QQ, "X")
+            sage: L.Graphs().generating_series().truncate(7)
+            1 + X + X^2 + 4/3*X^3 + 8/3*X^4 + 128/15*X^5 + 2048/45*X^6
+        """
+        P = self.parent()
+        L = LazyPowerSeriesRing(P.base_ring().fraction_field(),
+                                P._laurent_poly_ring._indices._indices.variable_names())
+        return L(lambda n: 2 ** binomial(n, 2) / factorial(n))
```
tscrim:
```suggestion
        return L(lambda n: 2**binomial(n, 2) / factorial(n))
```
PEP8 allows this (and perhaps slightly encourages having higher precedent operators not having a space around them). Anyways, up to you.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1663 @@
...
+
+        EXAMPLES::
+
+            sage: L = LazySpecies(QQ, "X")
+            sage: G = L.Graphs()
+            sage: list(G.isotypes(2))
+            [Graph on 2 vertices, Graph on 2 vertices]
+        """
+        if labels in ZZ:
+            yield from (G.canonical_label().copy(immutable=True) for G in graphs(labels))
```
tscrim, line 1568: Should this raise an error if `labels not in ZZ`?

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1663 @@
...
+        r"""
+        Return the species of set partitions.
+
+        EXAMPLES::
+
+            sage: L = LazySpecies(QQ, "X")
+            sage: G = L.SetPartitions()
+            sage: set(G.isotypes(4))
+            {[1, 1, 1, 1], [2, 1, 1], [2, 2], [3, 1], [4]}
+            sage: set(G.structures(["a", 1, x]))
```
tscrim:
```suggestion
            sage: sorted(set(G.structures(["a", 1, x])), key=str)
```
?

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1663 @@
...
+        b = P(lambda n: 0 if n else coeff_stream[1].coefficient(X_mol))  # TODO: we want a lazy version of Stream_exact
+        b_inv = P(coefficient)  # TODO: we want a lazy version of Stream_exact
+        g = P.undefined(valuation=1)
+        g.define(b_inv * (X - (self - b * X)(g)))
+        return g
+
+    compositional_inverse = revert
+
+
+class LazySpeciesElement_generating_series_mixin:
```
tscrim:
```suggestion
class GeneratingSeriesMixin:
```
I think a more concise name is better here. The docstring does the link with `LazySpeciesElement`. If you want the long name, then let’s follow Python naming conventions.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1663 @@
...
+            sage: list(XY.structures([], [1, 2]))
+            []
+        """
+        yield from self[sum(map(len, labels))].structures(*labels)
+
+    def _test_structures(self, tester=None, max_size=4, **options):
+        r"""
+        Check that structures and generating series are consistent.
+
+        We check all structures on 3 labels.
```
tscrim: This is no longer accurate.

tscrim (changes requested):
A few small things.

I do really think putting the specific species for the univariate case as methods onto the parent class is much better: no repetition of documentation, it directly gives you the corresponding class (which makes it easier to use `?`), and it’s a fairly natural way to handle the simple redirection. Yet if you still remain unconvinced, we can leave it the way it currently is.

mantepse:
> A few small things.

Thank you!

> I do really think putting the specific species for the univariate case as methods onto the parent class is much better: no repetition of documentation, it directly gives you the corresponding class (which makes it easier to use `?`), and it’s a fairly natural way to handle the simple redirection. Yet if you still remain unconvinced, we can leave it the way it currently is.

I'm confused now - I think I followed your idea of separating univariate and multivariate species (https://github.com/sagemath/sage/pull/38544#discussion_r2107874764) and creating proper methods for each.  Apparently I misunderstood you.  So, currently, the pattern is
```python
class LazySpeciesUnivariate(LazySpecies):
    def Graphs(self):
        return GraphSpecies(self)
...
class GraphSpecies(LazySpeciesElement_generating_series_mixin,
                   LazySpeciesElement, UniqueRepresentation,
                   metaclass=InheritComparisonClasscallMetaclass):
    def __init__(self, parent):
...
    def _repr_(self):
...
    def isotypes(self, labels):
...
    def generating_series(self):
...
    def cycle_index_series(self):
```

What did you have in mind?

src/sage/rings/lazy_species.py
mantepse, line 476:
```suggestion
        We check all structures with less than ``max_size`` labels.
```

src/sage/rings/lazy_species.py
mantepse: I used the idea from `LazyPowerSeries_gcd_mixin`.  But yes, it seems that camel case is more conventional:
```
combinat/diagram_algebras.py:2162:class UnitDiagramMixin:
combinat/partition_kleshchev.py:863:class KleshchevCrystalMixin:
combinat/crystals/tensor_product.py:656:class QueerSuperCrystalsMixin:
groups/galois_group.py:48:class _GMixin:
groups/perm_gps/symgp_conjugacy_class.py:18:class SymmetricGroupConjugacyClassMixin:
groups/libgap_mixin.py:23:class GroupMixinLibGAP:
modules/fp_graded/steenrod/module.py:102:class SteenrodModuleMixin:
rings/lazy_species.py:799:class LazySpeciesElement_generating_series_mixin:
rings/lazy_series.py:6221:class LazyPowerSeries_gcd_mixin:
```
It seems to me that the main class the mixin is intended for is usually part of the name, so I'd go with `LazySpeciesElementGeneratingSeriesMixin`, if that's ok for you?

src/sage/rings/lazy_species.py
mantepse: unfortunately, it won't help the randomness, because also the individual sets of each set partitions are probably randomly ordered.  So, should I mark the example as `#random`?

Also, using `set` is probably not wise: I guarantee, that we iterate over a set.  It is probably clearer to use `list`.

src/sage/rings/lazy_species.py
mantepse: Oh yes, I'm not sure what it should do.  `NotImplementedError` or `ValueError`.  I don't know whether, some day in the future, it will make sense to allow as input a set of labels, and get representatives using this set of labels.

src/sage/rings/lazy_species.py
tscrim: If I named that other class, I will take the full blame. Otherwise I missed that, and I take full blame. The CamelCase would have been best for that class too.

src/sage/rings/lazy_species.py
tscrim: Then we should return sorted lists for the inner sets as well, but the inner lists are `SetPartition` objects, right? Those should have a consistent display ordering...

src/sage/rings/lazy_species.py
tscrim: Then raising a `NotImplementedError` is best.

tscrim:
> > A few small things.
> 
> Thank you!
> 
> > I do really think putting the specific species for the univariate case as methods onto the parent class is much better: no repetition of documentation, it directly gives you the corresponding class (which makes it easier to use `?`), and it’s a fairly natural way to handle the simple redirection. Yet if you still remain unconvinced, we can leave it the way it currently is.
> 
> I'm confused now - I think I followed your idea of separating univariate and multivariate species ([#38544 (comment)](https://github.com/sagemath/sage/pull/38544#discussion_r2107874764)) and creating proper methods for each. Apparently I misunderstood you.

No, you understood me correctly, I just changed my mind with what might be the best practice. (I"m not 100% convinced yet either though.) Sorry for the confusion.

So what I am thinking is your original idea is best with one small modification (from [this comment](https://github.com/sagemath/sage/pull/38544#discussion_r2110726800)):
```python
def __init__(self,...):
    if self._arity == 1:
        from types import MethodType
        self.Sets = MethodType(SetSpecies, LazySpeciesUnivariate)
        self.Graphs = MethodType(GraphSpecies, LazySpeciesUnivariate)
        self.SetPartitions = MethodType(SetPartitionSpecies, LazySpeciesUnivariate)
        self.Cycles = MethodType(CycleSpecies, LazySpeciesUnivariate)
```
This way it isn't a `lambda` function, but has the full (class-level) docstring. The downside is that they won't be included in the doc. The other variation is keeping both classes, but having the univariate as
```python
class LazySpeciesUnivariate(LazySpecies):
    """
    Univariate lazy species.
    """

from types import MethodType
LazySpeciesUnivariate.Sets = MethodType(SetSpecies, LazySpeciesUnivariate)
LazySpeciesUnivariate.Graphs = MethodType(GraphSpecies, LazySpeciesUnivariate)
LazySpeciesUnivariate.SetPartitions = MethodType(SetPartitionSpecies, LazySpeciesUnivariate)
LazySpeciesUnivariate.Cycles = MethodType(CycleSpecies, LazySpeciesUnivariate)
```
(Python doesn't seem to have a good mechanism in Python for converting classes into methods like how I am doing.)

src/sage/rings/lazy_species.py
mantepse: You are quite right, I'm puzzled.  Let's try.

mantepse:
Indeed, it seems that `AbstractSetPartition._repr_` behaves badly:
```
sage -t --warn-long 5.0 --random-seed=89871645352471344146790937796251140637 src/sage/rings/lazy_species.py
Error: Failed example:: Got: [{{'a', x}, {1}},
 {{'a'}, {1}, {x}},
 {{1, 'a', x}},
 {{1, 'a'}, {x}},
 {{x, 1}, {'a'}}]
                                                                                                                                                                                                                                                                                                                                                                                          
    sorted(G.structures(["a", 1, x]), key=str)
Expected:
    [{{'a', x}, {1}},
     {{'a'}, {1}, {x}},
     {{1, 'a', x}},
     {{1, 'a'}, {x}},
     {{1, x}, {'a'}}]
Got:
    [{{'a', x}, {1}},
     {{'a'}, {1}, {x}},
     {{1, 'a', x}},
     {{1, 'a'}, {x}},
     {{x, 1}, {'a'}}]
1 item had failures:
   1 of   5 in sage.rings.lazy_species.LazySpeciesUnivariate.SetPartitions
    [290 tests, 1 failure, 15.54s wall]
```
Oh yes.  The code is
```python
    def _repr_(self):
...
        try:
            s = [sorted(x) for x in self]
        except TypeError:
            s = [sorted(x, key=str) for x in self]
        return '{' + ', '.join('{' + repr(x)[1:-1] + '}' for x in s) + '}'
```
and it's a certain M. Rubey who is to blame. #38974.  Embarassing.

mantepse:
> ```python
>     def _repr_(self):
> ...
>         try:
>             s = [sorted(x) for x in self]
>         except TypeError:
>             s = [sorted(x, key=str) for x in self]
>         return '{' + ', '.join('{' + repr(x)[1:-1] + '}' for x in s) + '}'
> ```

I think the problem is that, for the set partition $\\{\\{1, x\\}, \\{"a"\\}\\}$, the order of `1` and `x` is random, because they are coerced into the symbolic ring.

mantepse:
> So what I am thinking is your original idea is best with one small modification [...]

I tried the second variant, but it actually doesn't work the way I did it.  Apparently, `MethodType` passes the *class* rather than the *instance of the class*.

However, it seems to me that the current approach (i.e., separate classes for univariate and multivariate species and providing proper methods defined in the class) makes sense.  I think that we really want to see the methods (`Sets`, `Graphs`, ...) in the documentation, and it seems that the pattern of not having a class docstring for `SetSpecies` and using the `SetSpecies.__init__` docstring just for running the testsuite is easy to understand and extend.

mantepse:
Ping?

tscrim:
> > So what I am thinking is your original idea is best with one small modification [...]
> 
> I tried the second variant, but it actually doesn't work the way I did it. Apparently, `MethodType` passes the _class_ rather than the _instance of the class_.

Ah, right, it would pass the class rather than the instance. I think I messed up my initial experiment. However, this does work:
```python
sage: class Foo():
....:     def __init__(self, B):
....:         print(B)
sage: class Bar():
....:     def __init__(self, v):
....:         self.v = v
....:         from types import MethodType
....:         self.F = MethodType(Foo, self)
....:     def __repr__(self):
....:         return "Bar: %s"  % self.v
sage: B = Bar(5)
sage: B.F()
Bar: 5
<__main__.Foo object at 0x723f4645e660>
```

> However, it seems to me that the current approach (i.e., separate classes for univariate and multivariate species and providing proper methods defined in the class) makes sense. I think that we really want to see the methods (`Sets`, `Graphs`, ...) in the documentation, and it seems that the pattern of not having a class docstring for `SetSpecies` and using the `SetSpecies.__init__` docstring just for running the testsuite is easy to understand and extend.

Yes, that is the downside with my `MethodTypes` approach. Anyways, let's go with the current version then.

mantepse:
Does there remain anything that I should do or do you just want to have a closer look?

The next (substantial) improvements are #40129, #40163 and #40186.  Arithmetic product is easy and in the pipeline.  After that, we have truly superseeded the old species framework.

src/doc/en/reference/combinat/module_list.rst
```diff
@@ -407,3 +407,4 @@ Comprehensive Module List
     sage/combinat/words/words
     sage/combinat/yang_baxter_graph
     sage/rings/cfinite_sequence
+    sage/rings/lazy_species
```
tscrim: We probably should put this with the L's rather than at the bottom. Also, would there be some other natural place to link this doc? It's hard to find just being in the big combinat module list.

src/sage/rings/lazy_species.py
```diff
@@ -0,0 +1,1758 @@
...
+    INPUT:
+
+    - ``n_vec`` -- a `k`-tuple of non-negative integers.
+
+    - ``d`` -- a non-negative integer, the total sum of the parts in
+      all components
+
+    - ``weight_multiplicities_vec`` -- `k`-tuple of iterables, an
+      iterable, ``weight_multiplicities_vec[j][i]`` is the number of
+      positions with weight `i+1` in the `j`-th component.
```
tscrim:
```suggestion
    - ``n_vec`` -- a `k`-tuple of non-negative integers

    - ``d`` -- a non-negative integer, the total sum of the parts in
      all components

    - ``weight_multiplicities_vec`` -- `k`-tuple of iterables, where
      ``weight_multiplicities_vec[j][i]`` is the number of
      positions with weight `i+1` in the `j`-th component
```

tscrim (approved):
One doc thing and one general question (that I probably should have asked much earlier, sorry).

Should we call these `Lazy(Combinatorial)Species` and not just `(Combinatorial)Species`? Having `Lazy` in front suggests there is or will be a finite precision/non-lazy version, but I don't think this will happen.

Once addressed, it will be a positive review.

mantepse:
> One doc thing and one general question (that I probably should have asked much earlier, sorry).

No problem!  I'm amazed and it really makes me happy that you do thorough reviews!

> Should we call these `Lazy(Combinatorial)Species` and not just `(Combinatorial)Species`? Having `Lazy` in front suggests there is or will be a finite precision/non-lazy version, but I don't think this will happen.

I put the `Lazy` in front because of `PolynomialSpecies`, which is not a very good name, and because of uniformity and discoverability.  The latter, because I would think that to many formal power series guys species won't occur as a possibly better tool, even when it is.

I agree that a finite precision version won't happen.

We could have `CombinatorialSpecies` for what's now `PolynomialSpecies` and `LazyCombinatorialSpecies` instead of `LazySpecies`.  (By coincidence, `LazyC + tab` is currently unused, whereas we have `LazySy + tab` and `LazySp + tab`.  I'd support that, but I won't advocate for it.

The reason for not advocating: `AtomicSpecies` and `MolecularSpecies` were useful to me frequently (mostly to generate all of a given grade), whereas `PolynomialSpecies` are only a unwanted necessity.  `PolynomialSpecies` occur when extracting a term of a `LazySpecies` and to construct `LazySpecies` by passing a coefficient function.

src/doc/en/reference/combinat/module_list.rst
mantepse: you mean, move it between `k_tableau` and `lr_tableau`?

tscrim: Yes, that’s correct.

tscrim:
I don’t think the lazy in from makes it very discoverable compared to the API we have for similar such objects (mainly `completion` aka `formal_series_ring`). It would only be if someone is already working with the other lazy series that they would find it, but they wouldn’t know what it is and ignore it. I think `CombinatorialSpecies` would be the ideal name with my current understanding. `PolynomialSpecies` will just remain as it is, which I feel fits well enough with the power series vs polynomials differences. Putting it another way, I don’t think it should be defined by its laziness (in contrast to FPS with finite precision where we have different implementations).

src/doc/en/reference/combinat/module_list.rst
mantepse: Should I move `cfinite_sequence`, too, or in a different PR?

mantepse:
> I think `CombinatorialSpecies` would be the ideal name with my current understanding. `PolynomialSpecies` will just remain as it is, which I feel fits well enough with the power series vs polynomials differences. Putting it another way, I don’t think it should be defined by its laziness (in contrast to FPS with finite precision where we have different implementations).

Just a brief update: I am not sure.  Essentially, that does mean that we should make `LazyPowerSeriesRing` an implementation of `PowerSeriesRing`.

Another thing I just noticed is that `CombinatorialSpecies(min=None)` is currently in the global namespace: it designates a currently undefined species, with valuation `min`.   Not sure how to do the deprecation here.

I guess we could first leave things as they are (perhaps rename `LazySpecies` to `LazyCombinatorialSpecies`)  and after the deprecation period drop `Lazy`.

src/doc/en/reference/combinat/module_list.rst
tscrim: While I would normally say it’s better to move it in a different PR, we can just get it done here

tscrim:
> > I think `CombinatorialSpecies` would be the ideal name with my current understanding. `PolynomialSpecies` will just remain as it is, which I feel fits well enough with the power series vs polynomials differences. Putting it another way, I don’t think it should be defined by its laziness (in contrast to FPS with finite precision where we have different implementations).
> 
> Just a brief update: I am not sure. Essentially, that does mean that we should make `LazyPowerSeriesRing` an implementation of `PowerSeriesRing`.

In the sense of a common abstract class and creation (with the default precision being passed as `oo`), then I would not be opposed. However, because there is both a finite precision and lazy version, it makes some sense to have them be separate. That being said, I do think it would be better to have more common API and base classes, but that should be done separately.

> Another thing I just noticed is that `CombinatorialSpecies(min=None)` is currently in the global namespace: it designates a currently undefined species, with valuation `min`. Not sure how to do the deprecation here.
> 
> I guess we could first leave things as they are (perhaps rename `LazySpecies` to `LazyCombinatorialSpecies`) and after the deprecation period drop `Lazy`.

Hmm….that’s really annoying. We can’t really just do a drop-in replacement as it has different API, right? We could do a 2 step deprecation: the `CombinatorialSpecies` and then `Lazy[Combinatorial]Species` after removing `Lazy` from it. Yet that’s a bit of work. There might not be a good option here. :/

mantepse:
I would like to move on here.  Meanwhile, I have also implemented the arithmetic product.  I agreed to give an online course on enumeration with sagemath at 128.5, but I guess it is too late to have this even in develop.

tscrim (approved):
Okay, we can merge this now. I would like to deprecate `LazyCombinatorialSpecies` once we have the full replacement for the old version ready, but we can address that at that time.

mantepse:
Thank you!

antonio-rojas:
This is causing one test failure here (Arch, all dependencies up to date)
```
sage -t --warn-long 30.0 --random-seed=98465613109387680556179046791300572857 sage-src/src/sage/rings/lazy_species.py
File "sage-src/src/sage/rings/lazy_species.py", line 647, in sage.rings.lazy_species.LazyCombinatorialSpeciesElement.__call__
Failed example:
    A[5]
Expected:
    X*E_4 + X^2*E_3 + 3*X^3*E_2 + X*E_2(X^2) + 3*X^5
Got:
    X*E_4 + X^2*E_3 + X*E_2(X^2) + 3*X^3*E_2 + 3*X^5
1 item had failures:
   1 of  37 in sage.rings.lazy_species.LazyCombinatorialSpeciesElement.__call__
    [302 tests, 1 failure, 10.73s wall]
```

mantepse:
Thank you for reporting!  The problem is that there is no good ordering on the terms.  Shall I mark it as random?

antonio-rojas:
> Thank you for reporting! The problem is that there is no good ordering on the terms. Shall I mark it as random?

I'd say mark that one as random, and then test equality against the expected output

mantepse:
see #40304

PR #38546, opened 2024-08-21
use Parent in Hecke algebras

fchapoton:
switch from the auld class `CommutativeRing` to `Parent` for Hecke algebras.

This is part of an effort to get rid of old stuff not using the Parent / Element framework.

### :memo: Checklist


kwankyu (approved):
LGTM.

PR #38468, opened 2024-08-04
Fix buggy doc diff script

kwankyu:
Still after #38425, the script generating CHANGES.html needs improvement.

(1) Fixes #38467 

test: https://github.com/kwankyu/sage/actions/runs/10257146932/job/28377629063?pr=63
test:  https://doc-pr-63--sagemath-test.netlify.app/changes

(2) As seen in https://doc-pr-38339--sagemath.netlify.app/changes (hunk 1), the anchor may be planted in a line very far from where the diff is.

test: https://doc-pr-65--sagemath-test.netlify.app/changes  (hunk 1)

(3) As seen in https://github.com/sagemath/sage/actions/runs/10235554395/job/28316513988?pr=38449, the script may fail (and hence doc build github workflow as well) by a bug (I failed to consider that git diff counts lines from 1).

test: https://github.com/kwankyu/sage/actions/runs/10257125674/job/28377566493?pr=64
test: https://doc-pr-64--sagemath-test.netlify.app/changes (warning: slow loading)

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #37614, opened 2024-03-15
Style AUTHORS, INPUT, OUTPUT, EXAMPLES, etc.

kwankyu:
To make them look better.

Sample pages: 

- https://deploy-preview-37614--sagemath.netlify.app/html/en/reference/probability/sage/probability/probability_distribution

- https://deploy-preview-37614--sagemath.netlify.app/html/en/reference/structure/sage/structure/formal_sum.html#formal-sums

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation accordingly.

PR #38568, opened 2024-08-27
Right align sage/python tabs and alter vertical spacing

### LINKED ISSUE #37760: HTML documentation: Right-align Sage / Python inline tabs, use vertical kerning
Screenshot 2024-04-06 at 1 19 37 PM
----->
Screenshot 2024-04-06 at 1 19 37 PM

thecaligarmo:
Fixes #37760 

The css will push the tabs all the way to the right and alter the vertical spacing so that they are more "in-line".  The vertical spacing causes some issues as sometimes the text from the previous line will run into the tabs. To fix this, the previous line now has a `max-width` of `80%` so that they won't overlap.  This also makes it so that the `EXAMPLES:` line is on the same line as tabs. (As far as I could tell, there was no way to distinguish the `EXAMPLES:` text from other text).

A good page to see the changes is on:
`en/reference/rings_standard/sage/rings/integer_ring.html`
as it contains various different formats for sage/python examples.

### :memo: Checklist

### :hourglass: Dependencies

- Depends on #38468 (merged here)


mantepse:
Here is an alternative suggestion (I don't know anything about web programming and only little about user interfaces, so feel free to ignore).

Many web pages have a global language switch, often indicated by a flag of the corresponding country, in the top right (https://ux.stackexchange.com/questions/15041/placement-of-a-change-language-button)

Since the code shown is also globally changed, maybe this would be a good approach.

mkoeppe:
> Many web pages have a global language switch, often indicated by a flag of the corresponding country, in the top right

Well I thought about this idea when I made the original design, but there are several reasons why I decided to use tabs. The main one:
- Because the preparser is run to create the pure Python examples, they are often not very readable. So even people who want to copy-paste pure Python examples into a terminal or notebook may actually want to read the original Sage example and then switch the tab just to copy-paste.

mkoeppe:
> This also makes it so that the `EXAMPLES:` line is on the same line as tabs. (As far as I could tell, there was no way to distinguish the `EXAMPLES:` text from other text).

I think @kwankyu's #37614 could be adapted for this

mkoeppe:
I recommend to merge #38468 so that the "Build documentation" workflow runs through

thecaligarmo:
I've merged #38468  as requested. I've also adapted #37614 in order to add a little bit of space below `EXAMPLE` texts.

mkoeppe:
Beautiful. Thanks @thecaligarmo for working on this!

Note that in https://doc-release--sagemath.netlify.app/html/en/reference/tensor_free_modules/sage/tensor/modules/finite_rank_free_module we use a third tab "Sage Live", so more space will be needed

thecaligarmo:
Ok, so I changed a little bit how it works so that we can do variable widths depending on how many tabs there are. It should make it so it looks nice regardless of number of tabs.

kwankyu:
Screen Shot 2024-08-28 at 12 20 14 PM

As seen above, I am not sure if this improves the situation. For the three tabs, the tab labels will occupy even more space to the right. i doubt that moving tab labels to the right would only improve the situation. It creates other problem. 

What do we want to solve? Perhaps we want to move the tab labels completely out of sight. If not completely, then we may just shrink the spaces that the tab labels occupy. Or there may be other imaginative solutions. I am not sure what would be the best solution at the moment...

thecaligarmo:
I think, personally, I would generally prefer the tabs on the left. By putting it to the right it obfuscates it a little and it's harder to know they're tabs to begin with. The documentation is a little bulkier with the tabs on the left, but it makes it nicer from a UX perspective.

We can shrink the gap between the tabs and the previous sentence to make it a little smaller, but I don't think it's that big of an issue? 

An alternative design would be to put the tabs completely to the left of the code boxes. The problem with that would be that we have less room for the code boxes. This *theoretically* shouldn't be an issue as code length should be less than 80 chars, but that's not always the case.

kwankyu:
> ... If not completely, then we may just shrink the spaces that the tab labels occupy (without the "kerning"). 

Screen Shot 2024-08-28 at 1 13 11 PM

kwankyu:
> I think, personally, I would generally prefer the tabs on the left. By putting it to the right it obfuscates it a little and it's harder to know they're tabs to begin with. The documentation is a little bulkier with the tabs on the left, but it makes it nicer from a UX perspective.

I add that psychologically we don't skip the tab labels on the right, we tend to look at them, scanning all the way to the right.

mkoeppe:
> I add that psychologically we don't skip the tab labels on the right, we tend to look at them, scanning all the way to the right.

I think adding a bit of vertical spacing above the tabs might eliminate this effect.

kwankyu:
> I also prefer the tabs on the left. I don't see much difference with the spacing tweaks to the labels.

How about dimming the labels. No difference?

thecaligarmo:
> > ... If not completely, then we may just shrink the spaces that the tab labels occupy (without the "kerning"). 
> 
> Screen Shot 2024-08-28 at 1 13 11 PM
> 
> or even
> 
> 
> Screen Shot 2024-08-28 at 1 15 53 PM
> 

These look good. I can try and add this tomorrow when I wake up. I'll also put it back on the left unless we're set on it being put on the right?

kwankyu:
Good for me. Thanks.

kwankyu:
Yes, in isolation. But we see too many of them. I prefer just highlighting the underline (bottom border line).

We many choose one if the author, hopefully, prepares a few design candidates :-)

thecaligarmo:
Since all of you are using the light version, Here's what it will look like in the dark version.

With Sage being highlighted as well:
image

With Sage *not* being highlighted:
image

Personally, The I like it with the highlighting. Otherwise, the visual cue that `Sage` is the current tab is way to small and very easy to miss.

In the branch I'm pushing I have the *highlighted* version. If it looks good, let me know and I'll clean up the code and upload a final version (Currently I've commented out some things just in case we decide to go backwards. So I'll need to clean-up the code a little)

kwankyu:
How about these choices?

(AL) The tabs are on the left; the label and the border bottom of the current tab are highlighted 

(BL) The tabs are on the left; only the border bottom of the current tab is highlighted

(AR) The tabs are on the right; the label and the border bottom of the current tab are highlighted 

(BR) The tabs are on the right; only the border bottom of the current tab is highlighted

(ARK) The tabs are on the right and there is a vertical kerning; the label and the border bottom of the current tab are highlighted 

(BRK) The tabs are on the right and there is a vertical kerning; only the border bottom of the current tab is highlighted

kwankyu:
+1 to (BL)

May I call others from sage-devel?

thecaligarmo:
Yeah, feel free to call others.  

+1 to (AL)

GMS103:
(AL) +1

mkoeppe:
(BRK) > (ARK) > (BL) > (AL) >> status quo

roed314:
(AL) +1

kwankyu:
> ... the visual cue that `Sage` is the current tab is way to small and very easy to miss.

To promote (BL), let me emphasize that the necessity of the visual cue is small because the displayed examples in the current tab are strong visual cue. Note that the whole point of this PR is to move the tab labels out of sight as much as possible.

kwankyu:
Sphinx Tabs has this style:
Screenshot 2024-08-29 at 7 53 48 AM

Could we have a sample in this style (with no label highllighting)? Then we may remove all highlighting while keeping enough visual cue.

thecaligarmo:
Here's some additional samples with different options. 
We can keep (AL) and (BL) as before, I'll add options for these (AndrewMathas, your AL' I've changed to DL to keep the notation a little simpler so we don't need more primes)

(CL) Boldened (900), grey color  - I had to decrease the bolding of Sage Live / Python (500) to see the bolding
image

(DL) Tab with highlighting, but no line underneath:
image

(AT) Tab with border around (all blue):
image

(BT) Tab with border around (tab blue):
image

(CT) Tab with border around (tab gray)
image

(DT) Tabs with border around (all gray)
image

kwankyu:
Thanks!

(DT) > (BL)

kwankyu:
For easy counting, only single choice or multiple choices please (no preferences order)

mkoeppe:
I think all T choices are a mismatch with the furo style, which uses no such elements anywhere

kwankyu:
@GMS103 @mkoeppe @roed314 @thecaligarmo Would you vote again as we have more choices now? Multiple choices may help, I think.

roed314:
I'm still in favor of (AL).  I like Sage highlighted in blue, prefer the cleaner lines without the tabs, and think that the underline adds a bit.  If I had to choose a second option, it would be (DL).

mkoeppe:
(BL), (AL)

GMS103:
(AL, (BL), (DL)

kwankyu:
The tally of votes up to now:

(AL): @GMS103, @mkoeppe, @thecaligarmo, @roed314
(BL): @GMS103, @mkoeppe, @kwankyu
(DL): @GMS103, @roed314
(BT): @AndrewMathas 
(CT): @AndrewMathas 
(DT): @kwankyu
(CL): @kwankyu

kwankyu:
By the current tally, the winner is (AL). 

I will set positive review tomorrow if there is no more votes.

mkoeppe (approved):

kwankyu (approved):

kwankyu:
Thanks for the work!

AndrewMathas (approved):
Looks good to me.

thecaligarmo:
Sounds good. I'll update the code to reflect the chosen design and push when it's ready (hopefully in the next day or two). I'll update the ticket when the design is updated

kwankyu:
I thought the PR branch is already for (AL). No?

thecaligarmo:
> I thought the PR branch is already for (AL). No?

No, it's currently still on the right hand side. I stopped editing it while we were voting on it. It shouldn't take to long to do, but I'm away from my computer until Thursday night/Friday, so it should be finished by the end of this week.

kwankyu:
The doc preview shows the tabs on the left...

thecaligarmo:
Yes. For the previews I made local changes that didn't affect the branch itself. This was done because there were many options and it didn't seem viable (and would be a waste of time) to make a commit for every single option. So none of those are in the branch.

kwankyu:
I meant https://github.com/sagemath/sage/pull/38568#issuecomment-2313374714. 

Anyway, take your time.

thecaligarmo:
@kwankyu  Ah ok! I had missed my comment that I had actually updated my branch to include (AL), so I was wrong. Sorry about that. The thing that I hadn't finished doing was cleaning up the code of unnecessary things. I've gone ahead and done that and it should now be ready.

src/sage_docbuild/conf.py
```diff
@@ -1038,7 +1047,11 @@ def apply(self):
                         content = Container("", is_div=True, classes=["tab-content"])
                         content += cell_node
                         container += content
-                        parent.insert(index + 1, container)
+                        parent.insert(index, container)
+                        index += 1
+                        if isinstance(prev_node, nodes.paragraph):
+                            prev_node['classes'].append('with-sage-live-tab')
+
```
kwankyu, line 1055: there is an extra blank line inserted at L1054.

kwankyu:
Otherwise lgtm.

src/sage_docbuild/conf.py
thecaligarmo: Pushed a fix. Sorry I missed that.

kwankyu: Thanks.

Don't be sorry. We all make mistakes. One reason why we have the reviewer system.

PR #38575, opened 2024-08-28
Fix `is_homogeneous` weight bug

grhkm21:
PoC:

```sage
term_order = TermOrder("wdegrevlex", [1, 3])
R.<x, y> = PolynomialRing(Qp(5), order=term_order)
assert not (x + y).is_homogeneous()
assert (x^3 + y).is_homogeneous()
```

### DIFF
--- a/src/sage/rings/polynomial/multi_polynomial_element.py
+++ b/src/sage/rings/polynomial/multi_polynomial_element.py
@@ -1285,8 +1285,17 @@ class MPolynomial_polydict(Polynomial_singular_repr, MPolynomial_element):
             False
             sage: (x^2*y + y^2*x).is_homogeneous()
             True
+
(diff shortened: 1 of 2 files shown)


grhkm21:
CI Failure reported at #38579 and is irrelevant

GiacomoPope:
This seems like a sensible fix, but I don't know what the best thing to do with the weighted polynomials is generally. This could be merged and then a whole new class could be refactored at a later date?

grhkm21:
AFAIK there's basically no support for weighted polynomials in the Sage implementation (not Singular) of multivariate polynomial rings with the `polydict` things. It's possible to fix them all at once by going through the 11k lines of code, but for me I prefer fixing these bugs as I discover and need them. Thanks for the review!

PR #38589, opened 2024-08-30
FiniteField: make GF(p,n) the same as GF(p^n)

user202729:
It is not uncommon for the user to accidentally call `GF(p,n)` while they want to call `GF(p^n)` or equivalently `GF((p,n))`.

This makes `GF(p,n)` the same as `GF((p,n))`.

### :memo: Checklist


grhkm21:
Actually I'm more in favour of treating `GF(p, n, ...)` as `GF((p, n), ...)`.

videlec:
> Actually I'm more in favour of treating `GF(p, n, ...)` as `GF((p, n), ...)`.

I tend to agree but the (serious) drawback is an extra type checking and hence more time needed to construct the object `GF(5, 'mygen')`.

user202729:
> I tend to agree but the (serious) drawback is an extra type checking and hence more time needed to construct the object `GF(5, 'mygen')`.

Timing data:

```
sage: %time GF(5^2, "a")
CPU times: user 176 ms, sys: 9.28 ms, total: 186 ms
Wall time: 234 ms
Finite Field in a of size 5^2
sage: %time isinstance(2, (int, Integer))
CPU times: user 7 μs, sys: 1 μs, total: 8 μs
Wall time: 9.54 μs
True
sage: %time GF(5^2, "a")
CPU times: user 0 ns, sys: 259 μs, total: 259 μs
Wall time: 264 μs
Finite Field in a of size 5^2
```

9.54 μs versus 264 μs. So around 3-4% increase if the field is already cached.

Is it that bad? (On the other hand, small changes pile up…?)

src/sage/rings/finite_rings/finite_field_constructor.py
```diff
@@ -207,7 +207,7 @@ class FiniteFieldFactory(UniqueFactory):
 
     INPUT:
 
-    - ``order`` -- a prime power
+    - ``order`` -- a prime power `p^n`, or a pair `(p,n)`
 
     - ``name`` -- string, optional.  Note that there can be a
```
vincentmacri:
```suggestion
    - ``name`` -- string or integer, optional. For a string,
      ``name`` is the name of the generator over the prime
      field. For convenience, in ``name`` is an integer, then
      ``GF(p, n)`` is equivalent to ``GF((p, n))`` (see examples below).
      Note that there can be a substantial speed penalty
      (in creating extension fields) when omitting the variable name,
      since doing so triggers the computation of pseudo-Conway
      polynomials in order to define a coherent lattice of extensions
       of the prime field.  The speed penalty grows with the size of
       extension degree and with the number of factors of the
       extension degree.
```

I think this is a bit more clear (and fixes the incorrect type description).

I'm making this on GitHub which doesn't show me if I'm exceeding the line lengths, you may have to fix the formatting if the linter complains.

vincentmacri (changes requested):
I think the wording in the docstring could be more clear, but good otherwise. After you make this change as long as no (relevant) tests fail and the HTML doc looks good I'll approve.

As for if this changing the behaviour of `GF(p, n)`, the behaviour of `GF(p, n)` with `n` an integer was not documented before so I consider whatever it did before (which is just return `GF(p)`) to be undocumented behaviour and changing it isn't a breaking change as far as I'm concerned.

src/sage/rings/finite_rings/finite_field_constructor.py
user202729: The linter is pretty relaxed nowadays (the line limit has been bumped to something like 120? Not sure), this is just very old code that sticks to ≈ 80 character line limit. Either way, doesn't matter.

src/sage/rings/finite_rings/finite_field_constructor.py
```diff
@@ -207,15 +207,20 @@ class FiniteFieldFactory(UniqueFactory):
...
-      omitting the variable name, since doing so triggers the
-      computation of pseudo-Conway polynomials in order to define a
-      coherent lattice of extensions of the prime field.  The speed
-      penalty grows with the size of extension degree and with
-      the number of factors of the extension degree.
+    - ``order`` -- a prime power `p^n`, or a pair `(p,n)`
+
+    - ``name`` -- string or integer, optional. For a string,
+      ``name`` is the name of the generator over the prime
+      field. For convenience, in ``name`` is an integer, then
```
vincentmacri:
```suggestion
      field. For convenience, if ``name`` is an integer, then
```
I had a typo. Feel free to set to positive review after fixing this and there are no (relevant) CI failures.

PR #38592, opened 2024-08-30
Add is_integral method to algebraic numbers

user202729:
For symmetry with existing methods e.g. `QQ(1).is_integral` or such method on elements of number fields.

Implementation pretty much copied from `NumberFieldElement`.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/qqbar.py
+++ b/src/sage/rings/qqbar.py
@@ -4487,7 +4487,7 @@ class AlgebraicNumber_base(sage.structure.element.FieldElement):
 
     def is_integral(self):
         r"""
-        Determine if a number is an algebraic integer.
+        Check if this number is an algebraic integer.
 
         EXAMPLES::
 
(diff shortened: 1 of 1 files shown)


src/sage/rings/qqbar.py
```diff
@@ -4485,6 +4485,26 @@ def as_number_field_element(self, minimal=False, embedded=False, prec=53):
         """
         return number_field_elements_from_algebraics(self, minimal=minimal, embedded=embedded, prec=prec)
 
+    def is_integral(self):
+        r"""
+        Determine if a number is an algebraic integer.
+
```
kwankyu, line 4491: This description does not read well. The method works with the given object, namely, ``self``.  It is not just "a number". I suggest
```
Check if this number is an algebraic integer.
```
or
```
Return ``True`` if this number is an algebraic integer.
```

user202729: Okay. 

(Result of mindlessly copying from the other method I guess. In `number_field_element.pyx`:

```python
    def is_integral(self):
        r"""
        Determine if a number is in the ring of integers of this number
        field.

```
)

user202729: Fixed, thanks for the review.

kwankyu (approved):
Thanks. LGTM.

Issue #32265, opened 2021-07-22
fix containment method of Subsets to recognize python sets


saliola:
<div id="comment:0"></div>

The following two commands should return the same answer:

```
sage: {1, 2} in Subsets(3)
False
sage: {1, 2} in list(Subsets(3))
True
```

Component: **combinatorics**

Keywords: **sets, Subsets**

_Issue created by migration from https://trac.sagemath.org/ticket/32265_

PR #38602, opened 2024-09-03
Allow for python sets to be counted as sets

thecaligarmo:
Fixes Issue #32265 

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/subset.py
+++ b/src/sage/combinat/subset.py
@@ -325,10 +325,7 @@ class Subsets_s(Parent):
             True
         """
         if value not in Sets():
-            if isinstance(value, set):
-                value = Set(value)
-            else:
-                return False
+            return False
         return all(v in self._s for v in value)
 
     def cardinality(self):
--- a/src/sage/combinat/subset.py
+++ b/src/sage/combinat/subset.py
@@ -324,7 +324,7 @@ class Subsets_s(Parent):
             sage: {1, 2} in S
             True
         """
-        if value not in Sets():
+        if value not in Sets() and not isinstance(value, (set,frozenset)):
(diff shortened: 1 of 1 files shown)


src/sage/combinat/subset.py
```diff
@@ -321,9 +321,14 @@ def __contains__(self, value):
...
             False
+            sage: {1, 2} in S
+            True
         """
         if value not in Sets():
-            return False
+            if isinstance(value, set):
+                value = Set(value)
+            else:
+                return False
```
tscrim:
```suggestion
            return False
```

src/sage/combinat/subset.py
```diff
@@ -321,9 +321,14 @@ def __contains__(self, value):
             True
             sage: 2 in S
             False
+            sage: {1, 2} in S
+            True
         """
         if value not in Sets():
```
tscrim:
```suggestion
        if value not in Sets() and not isinstance(value, (set,frozenset)):
```

tscrim (approved):
I think this is fine since all it needs to do is iterate over the object. So basically we add an additional check for `set` (and `frozenset`).

tscrim:
Once this change is done, then this is positive review.

thecaligarmo:
> I think this is fine since all it needs to do is iterate over the object. So basically we add an additional check for `set` (and `frozenset`).

That looks good for me. I've used that instead.

thecaligarmo:
Just ran through and tested everything on my machine and it's all running good. I fixed one style issue (space after comma) and it looks good on my end now.

tscrim (approved):
Thank you. LGTM.

PR #38607, opened 2024-09-03
Add subs method for function field elements

### LINKED ISSUE #38514: Implement subs or __call__ for FunctionFieldElement
### Problem Description

`FunctionFieldElement` inherits from `Element` and so it should have a working `subs` function if the notion of substitution makes sense (which I think it does in this case).

### Proposed Solution

This can be added by implementing `__call__` for `FunctionFieldElement` as the `subs` implemented in `Element` will fall back to `__call__`. This solution also allows for easier function calling syntax.

### Alternatives Considered

Implementing `subs` instead of `__call__` would work but wouldn't allow the easier function calling syntax.

### Additional Information

Example:
```python
K = GF(7)
Kx.<x> = FunctionField(K)
y = polygen(Kx)
f = x^6 + 3

# I think these should return 2^6 + 3 = 67 = 4 (mod 7)
f(2) # TypeError
f.subs(x=2) # x^6 + 3
```

vincentmacri:
Fixes #38514.

### :memo: Checklist

### :hourglass: Dependencies

None


vincentmacri:
I'm not really sure what the rules/conventions are around using `# needs sage.[whatever]` in doc tests. I didn't use it in any of my tests because I wasn't really sure how/why it's used. Any feedback/guidance on that would be appreciated.

vincentmacri:
@DaveWitteMorris could I get a review? (Asking since you triaged the bug report.)

src/sage/rings/function_field/element.pyx
```diff
@@ -121,6 +124,7 @@ cdef class FunctionFieldElement(FieldElement):
         sage: isinstance(t, sage.rings.function_field.element.FunctionFieldElement)
         True
     """
+
```
tscrim:
```suggestion
```
We generally try to avoid this; at least we shouldn't add it when it is not already there.

src/sage/rings/function_field/element.pyx
```diff
@@ -174,6 +178,260 @@ cdef class FunctionFieldElement(FieldElement):
         """
         return self._x._latex_()
 
+    def subs(self, in_dict=None, **kwds):
+        r"""
+        Substitutes given generators with given values while not touching
```
tscrim:
```suggestion
        Substitute the given generators with given values while not touching
```

src/sage/rings/function_field/element.pyx
```diff
@@ -174,6 +178,260 @@ cdef class FunctionFieldElement(FieldElement):
...
+
+        def sub_recurse(self, sub_dict):
+            ff = self.parent()
+            if ff.base_field() == ff:
+                return ff(self._x.subs({ff.gen(): sub_dict[ff.gen()]}))
+            else:
+                total = ff.zero()
+                for i, v in enumerate(list(self._x)):
+                    total += sub_recurse(v, sub_dict) * sub_dict[ff.gen()]**i
+                return ff(total)
```
tscrim:
```suggestion
            total = ff.zero()
            for i, v in enumerate(list(self._x)):
                total += sub_recurse(v, sub_dict) * sub_dict[ff.gen()]**i
            return ff(total)
```
I find this more clear and usually easier to debug.

src/sage/rings/function_field/element.pyx
```diff
@@ -174,6 +178,260 @@ cdef class FunctionFieldElement(FieldElement):
...
+
+        made_substitution = False
+        if in_dict is not None:
+            for k, v in in_dict.items():
+                if k in sub_dict:
+                    sub_dict[k] = v
+                    if v != k:
+                        made_substitution = True
+                else:
+                    raise TypeError('key does not match any field generators')
```
tscrim:
```suggestion
                if k not in sub_dict:
                    raise TypeError('key does not match any field generators')
                sub_dict[k] = v
                if v != k:
                    made_substitution = True
```

src/sage/rings/function_field/element.pyx
```diff
@@ -174,6 +178,260 @@ cdef class FunctionFieldElement(FieldElement):
...
+        else:
+            used_kwds = {k: False for k in kwds}
+            for g in sub_dict:
+                for k, v in kwds.items():
+                    if str(g) == k:
```
tscrim:
```suggestion
            for g in sub_dict:
                strg = str(g)
                if strg not in kwds:
                    continue
                v = kwds[strg]
```
and de-indent to the appropriate level.

tscrim (changes requested):
Here are my suggestions; please let me know if you have any questions.

src/sage/rings/function_field/element.pyx
vincentmacri: The wording I had was copied from the subs method for another object (I think `Element`). I like your wording better so I'm accepting the suggestion but just letting you know in case consistency is more important.

vincentmacri:
Thanks @tscrim! My only question is this one above:

> I'm not really sure what the rules/conventions are around using `# needs sage.[whatever]` in doc tests. I didn't use it in any of my tests because I wasn't really sure how/why it's used. Any feedback/guidance on that would be appreciated.

tscrim:
> Thanks @tscrim! My only question is this one above:
> 
> > I'm not really sure what the rules/conventions are around using `# needs sage.[whatever]` in doc tests. I didn't use it in any of my tests because I wasn't really sure how/why it's used. Any feedback/guidance on that would be appreciated.

Sorry, I forgot to answer that. Don’t worry about it; you do not need to add anything. It is for modularization that will be fixed by whomever knowns if something is necessary.

src/sage/rings/function_field/element.pyx
tscrim: Thanks, but that is (very) old documentation before we have our current standards. It would be good to change that one. Independently, consistency is not so important.

vincentmacri:
Great! In that case is this ready to merge? I think the CI failures are unrelated. The tests pass on my machine at least.

tscrim (approved):
Indeed, I think this can be merged.

PR #38613, opened 2024-09-04
`categories/morphism.pyx`: remove some `noexcept`s

gmou3:
Remove ignored `noexcept`s for functions with `Python` return types.

```
[sagelib-10.5.beta3] [spkg-install]     warning: sage/categories/morphism.pyx:801:26: noexcept clause is ignored for function returning Python object
[sagelib-10.5.beta3] [spkg-install]     warning: sage/categories/morphism.pyx:826:22: noexcept clause is ignored for function returning Python object
```


kwankyu (approved):
LGTM.

PR #38616, opened 2024-09-04
`structure/sage_object.pyx`: remove excess `%s`s

gmou3:
To avoid the following warning:
```
warning: sage/structure/sage_object.pyx:722:42: Too few arguments for format placeholders
```


mkoeppe (approved):
LGTM, thanks.

PR #38549, opened 2024-08-23
Prepare sage for sphinx 8

kiwifb:
This PR prepares sage for sphinx 8. An update for sage_autodoc.py is needed and is provided in this PR.
The minimum known version of sphinx supporting the updated sage_autodoc.py is 7.4.7. So, sphinx is updated from 7.3.7 (which is known not to work) to 7.4.7.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38619, opened 2024-09-04
Re-add py3.9 support to sage_autodoc.py

kiwifb:
This PR corrects the fact the sage_autodoc.py changes needed for sphinx 8 dropped python 3.9 support. This is a consequence of the fact that sphinx 8 has dropped python 3.9. 
Since we still want to support python 3.9, some support has to be re-introduced.
See https://github.com/sagemath/sage/pull/38549#issuecomment-2327790930 and after.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38627, opened 2024-09-06
Require weaker Sphinx dependency for sagelib

tobiasdiez:
The version constraint added in https://github.com/sagemath/sage/pull/38549 broke the conda ci workflow and added an unnecessarily strict version constraint for sphinx as a dependency of sagelib (as opposed to sage-the-docs).

This is fixed by providing the old version constraint for sphinx in `setup.cfg`, but with `v8` now allowed.

Alternatively, the stricter version constraint should only be added to the sage-the-docs package.

### :memo: Checklist

### DIFF
--- a/src/setup.cfg.m4
+++ b/src/setup.cfg.m4
@@ -33,7 +33,7 @@ dnl From Makefile.in: SAGERUNTIME
     SPKG_INSTALL_REQUIRES_ipython
     SPKG_INSTALL_REQUIRES_pexpect
 dnl From Makefile.in: DOC_DEPENDENCIES
-    SPKG_INSTALL_REQUIRES_sphinx
+    sphinx >=5.2, <9
     SPKG_INSTALL_REQUIRES_networkx
     SPKG_INSTALL_REQUIRES_scipy
     SPKG_INSTALL_REQUIRES_sympy


kiwifb:
My update to sage_autodoc.py does require sphinx 7.4.x at the minimum otherwise building the doc breaks. Breaking doc building would also break CI (somewhere else). Now I had to walk back some changes in a follow up at #38619 to continue supporting python 3.9 :( If you are telling me that enables lower version of sphinx as well, then sure.

Is the problem that you do not an appropriate version of sphinx in conda or something like that?

kiwifb:
I hear you on the sphinxify bit. But CI also builds documentation, there is not a process in sage's build system to have one sphinx to build sage and then update it to build the documentation.

tobiasdiez:
Sage-the-distro (and the build system included in it), the ci doc build as well as the modularization sage-the-doc package will still use the stronger version constraint that you set. The changes in this PR only affect people trying to use only the library sagemath-standard (say via pypi) or devs that don't want to build the docs locally.

kiwifb:
OK, I am getting that argument. Although it is completely disjoint from the CI which would mean the CI label you added is misleading. But I am guessing you are also telling me the conda CI does not build the doc and therefore can use your lower constraint?

tobiasdiez:
> OK, I am getting that argument. Although it is completely disjoint from the CI which would mean the CI label you added is misleading. But I am guessing you are also telling me the conda CI does not build the doc and therefore can use your lower constraint?

Exactly, the conda workflow is still using sphinx v7.3 (but doesn't build the docs). The ci label is so that this PR get applied to the other PRs so that the "verify dependencies" step passes there.
https://github.com/sagemath/sage/actions/runs/10732023722/job/29763038707#step:9:9
vs 
https://github.com/sagemath/sage/actions/runs/10734686493/job/29770380015#step:9:9

kiwifb:
I see, and this particular file inherit my constraint. I must say I am always wary of dealing with sage-the-distro and avoid it as much as possible. This is clearly one of those case where I'd rather not be involved. But now that I have done all that digging on the details of your rationale, it'd be silly not to approve.

kiwifb (approved):
Good ahead if it really helps CI.

tobiasdiez:
A bit late, but thanks @kiwifb for the review!

PR #38636, opened 2024-09-07
fix something in quadratic forms

### LINKED ISSUE #38633: Diagonalization of singular quadratic form over finite field
### Steps To Reproduce

Run the following code:
```python
m = matrix(GF(11), [[1,5,0,0], [5,1,9,0], [0,9,1,5], [0,0,5,1]])
qf = QuadraticForm(m)
Q, T = qf.rational_diagonal_form(return_matrix=True)
T == identity_matrix(GF(11), 4)
```

### Expected Behavior

Since the matrix is not diagonal, the transformation matrix `T` should not be the identity matrix (prints `False`).

### Actual Behavior

Prints `True`.

### Additional Information

I believe that the problem is in `QuadraticForm._rational_diagonal_form_and_transformation`: when the matrix is singular, it triggers this section:

```python
except ZeroDivisionError:
    # Singular case is not fully supported by PARI
    pass
```

and proceeds to the general case. However, `Q` has already been modified.
Indeed, we can skip the PARI special case altogether by running:

```python
setattr(qf, "__pari__", lambda: 1/0) # raises an exception
```

after declaring `qf` and before computing the rational diagonal form. This gives the correct `T`.

### Environment

```markdown
- **OS**: Arch Linux (kernel 6.10.7)
- **Sage Version**: 10.4
```

### Checklist

fchapoton:
fix #38633 

### :memo: Checklist

### DIFF
--- a/src/sage/quadratic_forms/quadratic_form__local_field_invariants.py
+++ b/src/sage/quadratic_forms/quadratic_form__local_field_invariants.py
@@ -17,7 +17,7 @@ quadratic forms over the rationals.
 # ****************************************************************************
 
 ###########################################################################
-# TO DO: Add routines for hasse invariants at all places, anisotropic
+# TO DO: Add routines for Hasse invariants at all places, anisotropic
 # places, is_semi_definite, and support for number fields.
 ###########################################################################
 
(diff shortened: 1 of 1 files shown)


tscrim (approved):
LGTM.

PR #38649, opened 2024-09-11
Fix all three CONDA ci failing tests

fchapoton:
just some details in the code and tests

including one fix for conda runs, that should not test sympy behaviour

and also fixing 2 other conda-related failures

### :memo: Checklist


fchapoton:
now also trying to fix the other 2 conda failures

fchapoton:
who can set the CI-fix label ? after or before positive review ?

kwankyu:
If it is a hot fix for ci, ok before review.

fchapoton:
ok, I will add the label. So please review

dcoudert:
The time complexity of current version of `eliminate_parameters` is in $O(|a|*|b| + |a|^2 + |b|^2)$, due to the time needed to check if a parameter is in a list (linear in the length of the list) and to remove it from a list (linear in the length of the list).
We can have $O(|a| + |b|)$ as follows:
 
```python
def eliminate_parameters(self, a, b, z):
    r"""
    Eliminate repeated parameters by pairwise cancellation of identical
    terms in ``a`` and ``b``.

    If `p` is in `a` and in `b`, then remove the first occurence of `p` in both.
    """
    # Record the positions of the parameters in a
    pos_in_a = {}
    for pos, p in enumerate(a):
        if p in pos_in_a:
            pos_in_a[p].append(pos)
        else:
            pos_in_a[p] = [pos]
    # Reverse the lists of positions for faster access to first remaining
    # occurence of each parameter
    for v in pos_in_a.values():
        v.reverse()

    # If a parameter is in both a and b, remove its first occurence
    new_b = []
    keep_in_a = [True] * len(a)
    for p in b:
        if p in pos_in_a and pos_in_a[p]:
            keep_in_a[pos_in_a[p].pop()] = False
        else:
            new_b.append(p)

    if len(b) == len(new_b):
        return self

    new_a = [p for p, k in zip(a, keep_in_a) if k]

    return hypergeometric(new_a, new_b, z)
```

fchapoton:
good idea, but rather elsewhere please, as this PR is also used as a fix for Conda CI

fchapoton:
note also that probably nobody uses hypergeometric functions with more than 10 arguments..

dcoudert:
it seems interesting even for a small number of parameters. I agree it's a small improvement
```py
sage: a, b = [1, 1, 2, 5], [5, 1, 4]
sage: H = hypergeometric(a, b, 1/2)
sage: H.eliminate_parameters()
hypergeometric((1, 2), (4,), 1/2)
sage: eliminate_parameters(H, a, b, 1/2)
hypergeometric((1, 2), (4,), 1/2)
sage: %timeit H.eliminate_parameters()
16.3 µs ± 75 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
sage: %timeit eliminate_parameters(H, a, b, 1/2)
13.8 µs ± 45.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

fchapoton:
would please someone give a positive review ?

dcoudert (approved):
LGTM.

PR #38655, opened 2024-09-13
trying to refresh symbolic units

fchapoton:
by allowing the conversion constants in the stored data to be either strings or rationals

Part of the conversion table are replaced by integers or rational but  not systematically.

This file is not correctly indented, but this is not fixed here.

### :memo: Checklist


tscrim (changes requested):
It probably would be better to more systematically change all of the rational conversions to rationals... Otherwise LGTM.

fchapoton:
yes, I am slightly too lazy to do that.

fchapoton:
maybe enough things were converted now ?

tscrim (approved):
Thank you. Yes, this is good for now.

PR #38656, opened 2024-09-13
advertise for libgap in src/doc

fchapoton:
So that those reading the doc meet libgap rather than gap pexpect interface.

### :memo: Checklist


tscrim (changes requested):
We probably should explain in a few places why it is `libgap` rather than `gap` considering the others have a more direct translation. Something along the lines of
```
(We access GAP through its library interface `libgap`.)
```
Otherwise it might be confusing for a new user.

tscrim (approved):
Thanks. LGTM. Sorry for the delay in getting back to this.

Issue #38657, opened 2024-09-13
build/pkgs/sagetex: missing dependency on $(PYTHON_TOOLCHAIN) ?


orlitzky:
It needs `python_build`, which is part of `PYTHON_TOOLCHAIN`, though maybe there's a better answer:

```
[sagetex-3.6.1] No stamp file for package 'sagetex' in /home/mjo/src/sage.git/local/var/lib/sage/venv-python3.12/var/lib/sage/installed
[sagetex-3.6.1] No spkg-legacy-uninstall script; nothing to do
[sagetex-3.6.1] [spkg-install] Installing sagetex-3.6.1
[sagetex-3.6.1] [spkg-install] /home/mjo/src/sage.git/local/var/lib/sage/venv-python3.12/bin/python3: No module named build
[sagetex-3.6.1] [spkg-install] Warning: building with "python3 -m build --wheel --outdir=dist   ." failed.
[sagetex-3.6.1] [spkg-install] Retrying with "python3 -m build --wheel --outdir=dist --no-isolation --skip-dependency-check  .".
[sagetex-3.6.1] [spkg-install] /home/mjo/src/sage.git/local/var/lib/sage/venv-python3.12/bin/python3: No module named build
[sagetex-3.6.1] [spkg-install] ********************************************************************************
[sagetex-3.6.1] [spkg-install] Error building a wheel for sagetex-3.6.1
[sagetex-3.6.1] [spkg-install] ********************************************************************************
```

orlitzky:
@mkoeppe I could certainly fix this but I see a commit by you changing `$(PYTHON_TOOLCHAIN)` to `pip`?

mkoeppe:
Wheel packages only need pip, not the whole tool chain.

mkoeppe:
... but of course sagetex is *not* a wheel package. So adding `$(PYTHON_TOOLCHAIN)` is the correct fix. PR please?

orlitzky:
It looks like https://github.com/sagemath/sage/pull/38752 is going in. Closing this now so I don't forget.

PR #38658, opened 2024-09-13
configure.ac: add --disable-sagetex option

orlitzky:
SageTeX is a standard package, but it isn't used anywhere else in the sage library or even in the sage distribution. If you know you don't need it, the new `--disable-sagetex` option lets you avoid downloading and installing it.

(inspired by https://github.com/sagemath/sage/issues/38657, but I should have done this a long time ago)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -463,6 +463,12 @@ AC_ARG_ENABLE([r],
     done
   ])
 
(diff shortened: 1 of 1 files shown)


dimpase (approved):
lgtm

orlitzky:
thanks

Issue #38428, opened 2024-07-26
Invalid instructions in the Installation Guide for Ubuntu 24.04 LTS


egourgoulhon:
### Expected Behavior

This Debian/Ubuntu command from the  [Linux system package installation](https://doc.sagemath.org/html/en/installation/source.html#linux-system-package-installation) section of Sage 10.4 Installation Guide should run without any error:
```
sudo apt-get install bc binutils bzip2 ca-certificates cliquer cmake curl \
        ecl eclib-tools fflas-ffpack g++ gap gcc gengetopt gfan gfortran \
        glpk-utils gmp-ecm lcalc libatomic-ops-dev libboost-dev \
        libbraiding-dev libbrial-dev libbrial-groebner-dev libbz2-dev \
        libcdd-dev libcdd-tools libcliquer-dev libcurl4-openssl-dev libec-dev \
        libecm-dev libffi-dev libflint-dev libfplll-dev libfreetype-dev \
        libgap-dev libgc-dev libgd-dev libgf2x-dev libgiac-dev libgivaro-dev \
        libglpk-dev libgmp-dev libgsl-dev libhomfly-dev libiml-dev \
        liblfunction-dev liblinbox-dev liblrcalc-dev liblzma-dev libm4ri-dev \
        libm4rie-dev libmpc-dev libmpfi-dev libmpfr-dev libncurses5-dev \
        libntl-dev libopenblas-dev libpari-dev libplanarity-dev libppl-dev \
        libprimecount-dev libprimesieve-dev libpython3-dev libqhull-dev \
        libreadline-dev librw-dev libsingular4-dev libsqlite3-dev libssl-dev \
        libsuitesparse-dev libsymmetrica2-dev libz-dev libzmq3-dev m4 make \
        maxima maxima-sage meson nauty ninja-build openssl palp pari-doc \
        pari-elldata pari-galdata pari-galpol pari-gp2c pari-seadata patch \
        patchelf perl pkg-config planarity ppl-dev python3 python3-setuptools \
        python3-venv qhull-bin singular singular-doc sqlite3 sympow tachyon \
        tar texinfo tox xcas xz-utils
```

### Actual Behavior

As reported in https://ask.sagemath.org/question/78449/, running the above command in Ubuntu 24.04 fails with the errors
```
E: Package 'libbrial-dev' has no installation candidate
E: Unable to locate package libbrial-groebner-dev
E: Package 'maxima-sage' has no installation candidate

```
along with
```
[...]

PR #38659, opened 2024-09-14
Move installation details from README to the installation guide

kwankyu:
Currently we have two docs for installation: README.md and the Sage Installation Guide. There are severe problems:

- README is too complicated for a beginner to follow: too many details, jargons, too many branch points (links)
- There are many cross references between the README and the Guide. One has to hop between the two docs.

This PR aims to make README.md a light-weight easy-to-follow installation guide for newcomers and beginners trying to install Sage to their personal computers.

See the new README:

https://github.com/kwankyu/sage/tree/p/streamline-installation-guide#getting-started

**The new README does not try to be comprehensive.** That is the role of the Sage Installation Guide. Materials removed from README are incorporated to the Sage Installation Guide. The Sage Installation Guide would provide all the details for users for whom the README.md is not sufficient.

The installation guide in the README was tested with macOS, ubuntu, Windows WSL (ubuntu and fedora).

Along the way, I made extensive edits on the Sage Installation Guide. Most of them are non-technical. 

See the new Sage Installation Guide:

https://doc-pr-38659--sagemath.netlify.app/html/en/installation/

In future, **there would be little need to update the installation guide in README.** Developers could focus only on updating the Sage Installation Guide as Sage evolves. This lessens maintenance burden.

### :memo: Checklist


egourgoulhon:
Thanks for this initiative to make the  README lighter! 

The Linux dependencies listed in the new README are outdated: the `sage apt-get install ...` command will generate errors in Ubuntu 24.04 LTS, as detailed in #38428.

Basically, one should suppress `libbrial-dev`, `libbrial-groebner-dev`, `maxima-sage` from the package list and replace `libncurses5-dev` by `libncurses-dev`.

A dependencies list that is known to work for Ubuntu 24.04 is posted at https://sagemanifolds.obspm.fr/install_ubuntu.html

kwankyu:
Thanks for attention!

> The Linux dependencies listed in the new README are outdated: the `sage apt-get install ...` command will generate errors in Ubuntu 24.04 LTS, as detailed in #38428.
> 
> Basically, one should suppress `libbrial-dev`, `libbrial-groebner-dev`, `maxima-sage` from the package list and replace `libncurses5-dev` by `libncurses-dev`.

Yes. I was aware of that, but forgot to update the list at some point. The old list worked for Ubuntu 22.04. 

On the other hand, the new list may also be outdated at some point. That won't be a big problem because sage will install its own packages fetched from mirror sites for missing packages in the system. I added a remark in line 115-118.

anneschilling:
For me --filter blob:none in the command 
git clone -c core.symlinks=true --filter blob:none \
        --origin upstream --branch master https://github.com/sagemath/sage.git
did not work, so I left it out. Can it be omitted? If not, perhaps explain what it does.

kwankyu:
`--filter blob:none` is supposed to "filter" unnecessary data so that cloning becomes fast. I don't know how it works.  

I think it can be omitted. I never used it myself.

For me, `-c core.symlinks=true` is also not necessary. But it seems useful in Windows. So we should keep this.

kwankyu:
> Can it be omitted?

Done. Thanks.

README.md
```diff
@@ -11,672 +11,292 @@
...
+For your platform, there is a similar list of
+[build prerequisites](https://doc.sagemath.org/html/en/reference/spkg/_bootstrap.html#equivalent-system-packages).
+Install them using the package manager of your platform.
 
-2.  Clone the sources with `git`:
+Sage depends on many system packages. Install them by executing
 
-    - To check that `git` is available, open a terminal and enter
-      the following command at the shell prompt (`$`):
+    sudo apt-get install bc binutils bzip2 ca-certificates cliquer cmake curl \
```
mkoeppe: Strong -1 on hardcoding this list here. 
It will be untested and therefore fall out of date.

README.md
```diff
@@ -11,672 +11,292 @@
...
+        maxima meson nauty ninja-build openssl palp pari-doc \
+        pari-elldata pari-galdata pari-galpol pari-gp2c pari-seadata patch \
+        patchelf perl pkg-config planarity ppl-dev python3 python3-setuptools \
+        python3-venv qhull-bin singular singular-doc sqlite3 sympow tachyon \
+        tar texinfo tox xcas xz-utils
 
-            $ git --version
-            git version 2.42.0
+If you encounter error messages like "Unable to locate package ...", don't be
+panic. Just delete the problematic package names in the list, and rerun the
```
mkoeppe:
```suggestion
If you encounter error messages like "Unable to locate package ...", don't
panic. Just delete the problematic package names in the list, and rerun the
```

+1 on adding advice like this.

kwankyu: Thanks.

README.md
kwankyu: OK. Thanks.

kwankyu:
Matthias, I fixed an error
```diff
diff --git a/src/doc/bootstrap b/src/doc/bootstrap
index e99ecd52c28..256ba085e0b 100755
--- a/src/doc/bootstrap
+++ b/src/doc/bootstrap
-        if [ -n "PKG_SYSTEM_PACKAGES" ]; then
+        if [ -n "$PKG_SYSTEM_PACKAGES" ]; then
```
Perhaps this error was not noticed before because `PKG_SYSTEM_PACKAGES` is always non-empty?

src/doc/en/installation/index.rst
```diff
@@ -4,219 +4,143 @@
...
-    Developer's Guide
-    <https://doc.sagemath.org/html/en/developer/walkthrough.html#chapter-walkthrough>`_.
+- Then build Sage from source, choosing ``develop`` git branch, as described in
+  section :ref:`sec-installation-from-sources`.
 
-    - Then build SageMath from source as described in section
-      :ref:`sec-installation-from-sources`.
+- Alternatively, follow the instructions in section
+  :ref:`sec-installation-conda-develop`;
+  these describe an experimental method that gets all required
```
tobiasdiez: Conda is not experimental since a very long time.

Given the many (recent?) problems with installing various packages in sage-the-distro, conda should actually be mentioned first, and only then the installation from source. Would safe us a lot of headache.

src/doc/en/installation/index.rst
```diff
@@ -4,219 +4,143 @@
...
-    - Alternatively, follow the instructions in section
-      :ref:`sec-installation-conda-develop`;
-      these describe an experimental method that gets all required
-      packages, including Python packages, from conda-forge.
+**I don't want to do Sage development:**
 
-  - No development: **Do you have root access (sudo)?**
+- **I have root access (can use sudo command):** Then the easiest way to install Sage is
+  through a Linux distribution that provides it as a package.  Some
+  Linux distributions have up-to-date versions of Sage,
```
tobiasdiez, line 97: I think this can be improved. Instead of sending people to repology, just name the very few distros (arch, gentoo) that actually have an up-to date version of sage. And given that this is not the case for the bigger players like ubuntu, I would put the conda install first again (as it will apply to most people).

tobiasdiez (changes requested):
Awesome, this is a nice improvement! Thanks for the initiative. 

I think one could even go a step further and extract all of the remaining install instructions in the readme to a "Install Quick Walkthrough" in the docs. So the readme just links to the docs. Similar to how its done in scipy/numpy.

src/doc/en/installation/index.rst
kwankyu: I don't agree. Traditionally building from source is the main way of installing sage for developers.

But perhaps this may be debatable in sage-devel. If there is a consensus in favor of conda, then we may change this part later. But this is for another PR.

Not in scope of this PR.

src/doc/en/installation/index.rst
kwankyu: Sending people to repology seems better because it would have up to date information. If we hard-code them, it would be easily out of date.

I added
```diff
   You may need to consider upgrading your Linux distribution because a recent
   version of Sage is only available on a newer version of the distribution.
 
+  If there is no up-to-date version for your Linux distribution, you may
+  consider `conda`, as explained below.
+
```
instead of recommending conda first.

kwankyu:
> I think one could even go a step further and extract all of the remaining install instructions in the readme to a "Install Quick Walkthrough" in the docs. So the readme just links to the docs. Similar to how its done in scipy/numpy.

That is not the aim of this PR. The aim is: README for quick walkthrough, installation guide for full details. 

I think the quick walkthrough is short enough to put in README.  

Most projects on github contain a simple installation guide in README.

src/doc/en/installation/index.rst
tobiasdiez: Making conda the default has been already discussed on the mailing list with only positive feedback (by people other than me): https://groups.google.com/g/sage-devel/c/AvH3xq2bCfo/m/OKwqzrrYCwAJ

In either way, this mode of installation is definitely not "experimental".

src/doc/en/installation/index.rst
tobiasdiez: > Sending people to repology seems better because it would have up to date information. If we hard-code them, it would be easily out of date.

Maybe do both? `If you are on Gentoo or Arch Linux (or [any other distribution that provides an up-to-date Sage package](<link to repology>), then use your system package manager to install Sage. Otherwise ...`. This is definitely more user-friendly, which should be the major aim of these instructions.

tobiasdiez:
> > I think one could even go a step further and extract all of the remaining install instructions in the readme to a "Install Quick Walkthrough" in the docs. So the readme just links to the docs. Similar to how its done in scipy/numpy.
> 
> That is not the aim of this PR. The aim is: README for quick walkthrough, installation guide for full details.
> 
> I think the quick walkthrough is short enough to put in README.

It's 6 screens long for me... what advantages do you see of having it in readme as opposed to the main documentation?

> 
> Most projects on github contain a simple installation guide in README.

If we would have a "simple" installation method, I wouldn't mind having a short 3-line cmd in the readme...

src/doc/en/installation/index.rst
kwankyu: OK, I removed "experimental".

I don't mind whether conda is the default or not. I just don't want to make the changes here myself because I don't have much experience with conda and I am not sure if the change is desirable. 

I would be happy if you make a followup PR just for the change.

src/doc/en/installation/index.rst
kwankyu: How can we be sure that Gentoo and Arch Linux will provide an up-to-date sage? Is Gentoo providing an up-to-date sage now? It is not in the current repology list.

kwankyu:
> It's 6 screens long for me... 

It would be still 6 screens long in the main documentation. 

> what advantages do you see of having it in readme as opposed to the main documentation?

The advantage is "having it in readme".  I hope that a newcomer visiting our github repo have the impression that installing sage is not difficult. Our main installation guide is very long and complicated because it has all details explained. I think just looking at the installation guide makes one discouraged. 

What is the advantage of moving it to the main documentation? Do you want readme to be one screen long? I don't.

src/doc/en/installation/index.rst
tobiasdiez: > How can we be sure that Gentoo and Arch Linux will provide an up-to-date sage? 

If it's outdated, we need to update the recommendation. Same as for https://doc.sagemath.org/html/en/installation/linux.html.

tobiasdiez:
> > It's 6 screens long for me...
> 
> It would be still 6 screens long in the main documentation.
> 
> > what advantages do you see of having it in readme as opposed to the main documentation?
> 
> The advantage is "having it in readme". I hope that a newcomer visiting our github repo have the impression that installing sage is not difficult. Our main installation guide is very long and complicated because it has all details explained. I think just looking at the installation guide makes one discouraged.

Just to make this clear, my proposal was to move the info that is now in the readme to a *new* page in the docs. So that it as short/uncomplicated as in the readme.

> What is the advantage of moving it to the main documentation?

Automatic checks of links, consistent page layout with other install-related pages, possibility to use advanced features such as tabs, possibility to use semi-dynamic info (such as the auto-generated linux install commands), ...

src/doc/en/installation/index.rst
kwankyu: As said in the PR description, one of the goals is to make the instructions in the readme as stable as possible so that we could only focus on the main installation guide as update occasions arise.

kwankyu:
As stated in the PR description, the aim of this PR is to simplify the instruction guide in readme. README.md is a file directly accessible from the source tree while the installation guide is built from the source tree. I am not sure if removing completely the installation guide from README.md is desirable.

As I don't want to do that here, please create your own PR for that later.

kwankyu:
> Just to make this clear, my proposal was to move the info that is now in the readme to a _new_ page in the docs. So that it as short/uncomplicated as in the readme.

Instead of having two versions in the main installation guide, I hope that the detailed guide is cleaned up and updated. It is messy now and still after this PR.

grhkm21:
can we bump this

kwankyu:
OK. I will revise this soon.

kwankyu:
To repeat, the main point of this PR is to make the README  light, keeping the status quo of the traditional build procedures.

Making the README even lighter or changing build procedures departing from the status quo should go into new PRs after this one.

Please review.

PR #38661, opened 2024-09-15
src/sage/misc/package.py: use gnulib instead of conway_polynomials in tests

orlitzky:
The conway_polynomials package now has an `spkg-configure.m4`, and may not be installed as an SPKG. In that case, several tests fail. For example,

```
File "src/sage/misc/package.py", line 404, in sage.misc.package.installed_packages
Failed example:
    sorted(installed_packages().keys())
Expected:
    [...'conway_polynomials', ...]
Got:
    ['combinatorial_designs',
     'elliptic_curves',
     'execnet',
     'gnulib',
     'gnumake_tokenpool',
     'graphs',
     'h11',
     'iniconfig',
     'jmol',
     'polytopes_db',
     'pyproject_hooks',
     'pytest_mock',
     'pytest_xdist',
     'python3_venv',
     'python_build',
     'sage_conf',
     'sage_setup',
     'sagelib',
     'threejs']
```

The gnulib package seems like a safe alternative.

### DIFF
--- a/src/sage/misc/package.py
+++ b/src/sage/misc/package.py
@@ -402,9 +402,9 @@ def installed_packages(exclude_pip=True):
         sage: # optional - sage_spkg
         sage: from sage.misc.package import installed_packages
         sage: sorted(installed_packages().keys())
-        [...'conway_polynomials', ...]
-        sage: installed_packages()['conway_polynomials']  # random
-        '0.5'
+        [...'gnulib', ...]
+        sage: installed_packages()['gnulib']  # random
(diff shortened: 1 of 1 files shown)


orlitzky:
This fixes a harmless but annoying test failure with `--enable-system-site-packages` when conway-polynomials from the system is used.

"gnulib" should be a safe alternative because it's not a real package, only a placeholder for the `iconv.m4` file that is already in the repo.

dimpase:
please rebase

orlitzky:
no problem, done

dimpase (approved):
lgtm

PR #38662, opened 2024-09-15
configure.ac: disable more packages with --disable-notebook

orlitzky:
The h11 package is used only by httpcore, and httpcore is used only by httpx, and httpx is used only by jupyterlab. Since jupyterlab is already disabled by `--disable-notebook`, the others can be too.


dimpase:
please rebase

orlitzky:
sure thing, done

dimpase (approved):
LGTM. 

(The error message one gets while trying to start a notebook in such a config via  `./sage -n` is rather ugly, but that's another story)

PR #38663, opened 2024-09-15
Freetype fails to find freetype on macos

vbraun:
Freetype fails to find freetype on macos (Mac mini M2 with only XCode installed):

    src/checkdep_freetype2.c:8:10: fatal error: 'ft2build.h' file not found
    #include <ft2build.h>
             ^~~~~~~~~~~~
    2 errors generated.
    error: command '/usr/bin/gcc' failed with exit code 1

Sage skips installing zlib on macos because it is installed as part of XCode. The freetype configuration uses freetype-config which ends up using pkg-config to find zlib. In their infinite wisdom, Apple does not ship the .pc file:

    vbraun@Mini-M2 local % ./bin/freetype-config
    Package zlib was not found in the pkg-config search path.
    Perhaps you should add the directory containing `zlib.pc'
    to the PKG_CONFIG_PATH environment variable
    Package 'zlib', required by 'libpng', not found
    Package zlib was not found in the pkg-config search path.
    Perhaps you should add the directory containing `zlib.pc'
    to the PKG_CONFIG_PATH environment variable
    Package 'zlib', required by 'libpng', not found

Solution is to require a zlib.pc file via PKG_CHECK_MODULES([zlib], ...

### :memo: Checklist

### DIFF
--- a/build/pkgs/zlib/spkg-configure.m4
+++ b/build/pkgs/zlib/spkg-configure.m4
@@ -1,7 +1,8 @@
 SAGE_SPKG_CONFIGURE([zlib], [
+    PKG_CHECK_MODULES([zlib], [zlib >= 1.2.11], [zlib_cv_pc=yes], [zlib_cv_pc=no])
     AC_CHECK_LIB([z], [inflateEnd], [zlib_cv_libz=yes], [zlib_cv_libz=no])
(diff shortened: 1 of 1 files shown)


kwankyu (approved):
LGTM.

PR #38427, opened 2024-07-26
graphs: make init_short_digraph always sort neighbors but without the extra log complexity

cyrilbouvier:
This PR improve the `init_short_digraph` function that is used to initialize `StaticSparseCGraph` (used for immutable `Graph` and `DiGraph`).

Before, a boolean parameter `sort_neighbors` was used to specify if we wanted to sort the neighbors or not. It implied an extra `log` in the complexity (as `qsort` was called).
With this PR, the neighbors are always sorted at no extra cost. It is done by appending to the neighbors list the vertices in the correct order so the call to `qsort` is not needed anymore.

This PR partly reverts #38198 and mostly reverts #37662

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38198, opened 2024-06-11
Improve complexity comments for graphs

cyrilbouvier:
As noted in #37642, there is a difference in the complexity of enumerating neighbors or edges between sparse graphs and dense graphs.
This PR aims at making the comments about the time complexity of algorithms clearer in regard of this difference.

Most of the changes I made in this PR are due to the fact that the function `init_short_digraph` have the following time complexity:
- `O(n+m)` when the input graph has a sparse representation and sort_neighbors is False
- `O(n+m log(m))` when the input graph has a sparse representation and sort_neighbors is True
- `O(n^2)` when the input graph has a dense representation and sort_neighbors is False
- `O(n^2 log(m))` when the input graph has a dense representation and sort_neighbors is True

with `m` the number of edges and `n` the number of vertices of the graph.

I spotted 5 additional complexity claims in the comments that I was not able to verify:
1. for strong_orientations_iterator in orientations.py
2. for yen_k_shortest_simple_paths in path_enumeration.pyx
3. for feng_k_shortest_simple_paths in path_enumeration.pyx
4. for edge_disjoint_spanning_trees in spanning_tree.pyx
    (there is loop over the sorted edges of the graphs, which should imply a
    complexity of at least `n+m log(m)` for sparse graphs and `O(n^2 log(m))`  for dense graphs
5. for is_partial_cube in partial_cube.py
    (there is at least one enumeration of neighbors of a vertex (via calling `contracted[root]`,
    line 318) so `m` should appear in the complexity)

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38664, opened 2024-09-16
graphs: make init_short_digraph always sort neighbors but without the extra log complexity (2nd try)

cyrilbouvier:
This PR is based on the closed PR #38427 (by myself) that was breaking too much doctests.
 
This PR improve the `init_short_digraph` function that is used to initialize `StaticSparseCGraph` (used for immutable `Graph` and `DiGraph`).

Before, a boolean parameter `sort_neighbors` was used to specify if we wanted to sort the neighbors or not. It implied an extra `log` in the complexity (as `qsort` was called).
With this PR, the neighbors are always sorted at no extra cost. It is done by appending to the neighbors list the vertices in the correct order so the call to `qsort` is not needed anymore.

This PR partly reverts #38198 and mostly reverts #37662

Contrary to PR #38427, I did not include the patch that remove the sorting of vertices when StaticSparseGraph are initialized because it was breaking too many doctests. Instead I added a option to disabled said sorting and use it in my doctests to check that the new code works.

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/asteroidal_triples.pyx
+++ b/src/sage/graphs/asteroidal_triples.pyx
@@ -147,8 +147,7 @@ def is_asteroidal_triple_free(G, certificate=False):
     # module sage.graphs.base.static_sparse_graph
     cdef list int_to_vertex = list(G)
     cdef short_digraph sd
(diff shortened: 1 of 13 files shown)


dcoudert (approved):
LGTM.

We will have to work on another PR for these remaining sorts...

Issue #38668, opened 2024-09-17
Support for --disable-giac


orlitzky:
Giac is a PITA to have as a dependency:

* There's no public source repository
* There's no bug tracker
* There are no release announcements, new releases just show up in [some random directory](https://www-fourier.ujf-grenoble.fr/~parisse/debian/dists/stable/main/source/)...
* If you can tell what is the latest release, because there's no coherent version scheme
* No one knows what's in the new releases, breaking changes show up in what appear to be minor versions
* The tarball is manually generated and often contains random files from the author's computer
* The test suite for every version fails on almost every machine
* Recent releases contain undefined behavior that is retained intentionally even though it crashes glibcxx
* The build emits hundreds of compiler warnings
* As a result of all of this, each release requires heavy patching (i.e. time & effort) even on "boring" systems
* But now I have a new computer that is not boring, and those compiler warnings weren't kidding, because giac segfaults all over the place
* So I can't use giac any more, and if Sage has a hard dependency on it, then I can't use Sage any more either

Fortunately it looks like it won't be too hard to make giac optional in Sage:

1. https://github.com/sagemath/sage/pull/38669
2. https://github.com/sagemath/sage/pull/38672
3. https://github.com/sagemath/sage/pull/38690
4. https://github.com/sagemath/sage/pull/38756
5. https://github.com/sagemath/sage/pull/38686 (not strictly required, but it conflicts with several "needs" tags)
6. https://github.com/sagemath/sage/pull/38770
7. https://github.com/sagemath/sage/pull/39376

We lose a few integration examples in the doctests when the giac algorithm is skipped in favor of sympy, but that's about it.

parisseb:
I wonder if it's really worth replying such an attack.
For the records, there is a repository for giac source files inside geogebra git repo, changes are commented there. I do not maintain myself a repository. I copy the auto-generated tarballs when I build a binary deb package and it sets the version number. When I build a binary deb, I of course check my own regression tests (I have to make some times adjustements). I'm using glibcxx,without problems once some recent "features" are disabled (this is of course done by the configure script from giac if you don't patch it), because I do not want to modify working source code.
If you want to report a bug, you can do it on Xcas forum.

If sagemath chooses to make giac optional, it might be a problem for some sage users, not for me. There are plenty of users of giac worldwide, either from ports I maintain myself or embedded in other applications.
Bye!

kiwifb:
Hi Parisse,
[...]

PR #38669, opened 2024-09-17
Remove giac integration from GiNaC

orlitzky:
Phase one of https://github.com/sagemath/sage/issues/38668

AFAICS it's not doing anything anyway.


antonio-rojas:
Works fine here

orlitzky:
Thanks!

PR #35762, opened 2023-06-13
ideals of non-maximal orders in number fields

yyyyx4:
Resolves #34198.

Issue #34806, opened 2022-11-29
deprecate constructing number-field fractional ideals via orders' .ideal() method


yyyyx4:
<div id="comment:0"></div>

Currently, Sage creates a fractional ideal of the containing number field when calling `.ideal()` on a maximal order. This is a mathematically questionable choice, but perhaps more importantly, it will inevitably cause inconsistencies once #34198 is done — the same code would start returning wildly different objects depending on whether the given order is maximal or not.

In this ticket, I propose to deprecate the use of `.ideal()` on a maximal order to create number-field fractional ideals. This will presumably affect lots of user code, so I think it's good to get the deprecation warning in as soon as possible.

Related, deeper issue: #3680

CC:  @slel

Component: **number fields**

Author: **Lorenz Panny**

_Issue created by migration from https://trac.sagemath.org/ticket/34806_

yyyyx4:
Author: **Lorenz Panny**

yyyyx4:
Description changed:
``````diff
--- 
+++ 
@@ -2,3 +2,4 @@
 
 In this ticket, I propose to deprecate the use of `.ideal()` on a maximal order to create number-field fractional ideals. This will presumably affect lots of user code, so I think it's good to get the deprecation warning in as soon as possible.
 
+Related, deeper issue: #3680
``````

yyyyx4:
Branch: **[public/34806](https://github.com/sagemath/sagetrac-mirror/tree/public/34806)**

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
Commit: **[`40363f1`](https://github.com/sagemath/sagetrac-mirror/commit/40363f14428699e2ecd9f51d43ee8dbc4ed7bd7d)**

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
<div id="comment:2"></div>

Branch pushed to git repo; I updated commit sha1. New commits:
<table><tr><td><a href="https://github.com/sagemath/sagetrac-mirror/commit/40363f14428699e2ecd9f51d43ee8dbc4ed7bd7d"><code>40363f1</code></a></td><td><code>add deprecation warning for #34806</code></td></tr></table>

7ed8c4ca-6d56-4ae9-953a-41e42b4ed313:
<div id="comment:3"></div>

Branch pushed to git repo; I updated commit sha1. New commits:
[...]

PR #38671, opened 2024-09-18
construct order ideals by default for number-field orders

yyyyx4:
This completes the change begun in #34806 and #35762: When constructing ideals of an order, return an ideal of that order, rather than a fractional ideal of the ambient number field.


kwankyu (approved):
LGTM.

Finally checking ci.

yyyyx4:
Thank you!

PR #38672, opened 2024-09-18
src/sage/features/giac.py: add new feature for the giac program

orlitzky:
In preparation for adding a `--disable-giac` option, we add a new feature that detects the presence of the "giac" executable. We already have a feature for `sage.libs.giac`, but that only guards the libgiac interface; we still have code that runs "giac" behind pexpect. This will allow us to skip those tests when giac is not installed.

### DIFF
--- a/src/sage/features/giac.py
+++ b/src/sage/features/giac.py
@@ -1,0 +1,30 @@
+# sage_setup: distribution = sagemath-environment
+r"""
+Feature for testing the presence of ``giac``
+"""
+
+from . import Executable, FeatureTestResult
+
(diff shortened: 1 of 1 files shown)


dimpase (approved):
LGTM

dimpase:
should be rebased, but is good to go (I checked, rebases cleanly)

mkoeppe:
> We already have a feature for `sage.libs.giac`, but that only guards the libgiac interface

@orlitzky Quick correction - this doctest tag has already appeared but the feature definition was missing. #38715 fixes this oversight, please review.

Issue #38651, opened 2024-09-12
Latte binary `count` called `latte-count` on archlinux


videlec:
In my setup (see below for details), the program `count` that sage expects to be from LattE (integral points count in polytope) is actually a binary from llvm.

### Steps To Reproduce

With latte installed in SageMath 10.2 in a conda environment, the following computation fails
```
sage: V = [[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
....:      [0, 0, 0, 1, 0, 0, 0, 0, 0, 1],
....:      [0, 0, 0, 1, 0, 0, 0, 0, 1, 0],
....:      [0, 0, 0, 1, 0, 0, 0, 1, 0, 0],
....:      [0, 0, 0, 1, 0, 0, 1, 0, 0, 0],
....:      [0, 0, 0, 1, 0, 1, 0, 0, 0, 0],
....:      [0, 0, 0, 1, 1, 0, 0, 0, 0, 0],
....:      [0, 0, 1, 0, 0, 0, 1, 0, 0, 0],
....:      [0, 1, 0, 0, 0, 1, 0, 0, 0, 0],
....:      [1, 0, 0, 0, 1, 0, 0, 0, 0, 0]]
sage: P = Polyhedron(V)
sage: P.integral_points_count()
```

### Expected Behavior

Same as
```
sage: len(P.integral_points())
10
```

### Actual Behavior

```
File ~/sage/geometry/polyhedron/base_QQ.py:214, in Polyhedron_QQ.integral_points_count(self, verbose, use_Hrepresentation, explicit_enumeration_threshold, preprocess, **kwds)
    211     use_Hrepresentation = True
    213 from sage.interfaces.latte import count
--> 214 return count(
    215         p.cdd_Hrepresentation() if use_Hrepresentation else p.cdd_Vrepresentation(),
    216         cdd=True,
    217         verbose=verbose,
    218         **kwds)

File ~/sage/interfaces/latte.py:175, in count(arg, ehrhart_polynomial, multivariate_generating_function, raw_output, verbose, **kwds)
    173     else:
    174         err = ":\n" + err
--> 175     raise RuntimeError("LattE integrale program failed (exit code {})".format(ret_code) + err.strip())
    177 ans = bytes_to_str(ans)
    179 # There's an error handler below that uses the numOfLatticePoints
    180 # file created by latte, so we can't cleanup() the temporary
    181 # directory here. Instead we have to clean it up before the
    182 # (several) return statements.

[...]

PR #38678, opened 2024-09-19
implement is_functional for LattE executable

videlec:
Implement the method `is_functional` for the two LattE features `count` and `integrate` so that at least the problem from #38651 raises a meaningful error.

### :memo: Checklist

### DIFF
--- a/src/sage/features/latte.py
+++ b/src/sage/features/latte.py
@@ -16,14 +16,41 @@ Features for testing the presence of ``latte_int``
 #                  https://www.gnu.org/licenses/
 # ****************************************************************************
 
-from . import Executable
+import subprocess
+from . import Executable, FeatureTestResult
 from .join_feature import JoinFeature
 
 
 LATTE_URL = "https://www.math.ucdavis.edu/~latte/software.php"
 
 
-class Latte_count(Executable):
+class Latte_executable(Executable):
+    def is_functional(self):
+        r"""
+        Check whether this LattE executable works on trivial input.
+
+        EXAMPLES::
+
+            sage: from sage.features.latte import Latte_count, Latte_integrate
+
+            sage: Latte_count().is_functional() # optional - latte_int
+            FeatureTestResult('count', True)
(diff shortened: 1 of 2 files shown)


antonio-rojas:
Looks like this is not enough to fix the issue, the `count` and `integrate` functions in `sage/interfaces/latte.py` need to explicitly test for `Latte_count().is_present()` and return an error if false.

videlec:
Even better with `Feature.require()`. We now got
```
sage: P.integral_points_count()
...
FeatureNotPresentError: count is not available.
Call `count` failed with exit code 2
To install count using the arch package manager, you can try to run:
!sudo pacman -Sy 
  !sudo pacman -S latte-integrale
To install count using the Sage package manager, you can try to run:
  !sage -i latte_int
No equivalent system packages for pip are known to Sage.
Further installation instructions might be available at https://www.math.ucdavis.edu/~latte/software.php.
```

antonio-rojas:
> Even better with `Feature.require()`. We now got
> 
> ```
> sage: P.integral_points_count()
> ...
> FeatureNotPresentError: count is not available.
> Call `count` failed with exit code 2
> To install count using the arch package manager, you can try to run:
> !sudo pacman -Sy 
>   !sudo pacman -S latte-integrale
> To install count using the Sage package manager, you can try to run:
>   !sage -i latte_int
> No equivalent system packages for pip are known to Sage.
> Further installation instructions might be available at https://www.math.ucdavis.edu/~latte/software.php.
> ```

Nice, but the `pacman` part will only work if you're running sage from the distro (which is patched to use the modified executable names)

antonio-rojas:
The `is_functional` test doesn't work. The `count` command returns exit code 1, so `CalledProcessError` is always raised. Even `count --help` returns 1.

videlec:
> The `is_functional` test doesn't work. The `count` command returns exit code 1, so `CalledProcessError` is always raised. Even `count --help` returns 1.

That is very unfortunate...

PR #38680, opened 2024-09-19
Improve counting of local solutions for QuadraticForm at p=2

### LINKED ISSUE #38679: Improve counting of local solutions for QuadraticForm at p=2
### Problem Description

Given a QuadraticForm Q it is currently infeasible to compute Q.siegel_product(m) when Q.dim() >= 8. 
The reason for this is that Q.local_density(p,m) for p=2 needs to compute local (good) solutions modulo 8 which is done with a naive brute-force approach by count_congruence_solutions__good_type(2, 3, m, ..., ...).

### Proposed Solution

The solutions mod 8 can be computed efficiently by:
1. put Q in local normal form (which is already done),
2. count all solutions Q_block[v] = k mod 8 per Jordan block for each k=0,...7 (this can be done naively)
3. compute the convolution of the solutions of all blocks
4. return the count at m

Of course this could be implemented in a general way and not just for p=2 or modulo 8.

### Alternatives Considered

None.

### Additional Information

I will propose a pull request with the feature soon.

WvanWoerden:
Fixes #38679.

Introduces new method `CountAllLocalGoodTypesNormalForm(Q, p, k, m, zvec, nzvec)` to compute the number of local solutions of 'good type' for Q[v] = m mod p^k when Q is in local normal form at p.
This replaces the use of the slow brute-force function `CountAllLocalTypesNaive` indirectly used in `local_good_density_congruence_even` for (p,k)=(2,3) . 

### :memo: Checklist


src/sage/quadratic_forms/count_local_2.pyx
```diff
@@ -228,27 +228,27 @@ cdef local_solution_type_cdef(Q, p, w, zvec, nzvec):
             i += 1
 
     if not nonzero_flag:
-        return <long> 0
+        return < long > 0
```
fchapoton: this is not wanted, keep `<long>`

src/sage/quadratic_forms/count_local_2.pyx
```diff
@@ -271,13 +271,93 @@ cdef local_solution_type_cdef(Q, p, w, zvec, nzvec):
...
-        return <long> 5
+        return < long > 5
 
     # Error if we get here! =o
     print("   Solution vector is " + str(w))
     print("   and Q is \n" + str(Q) + "\n")
     raise RuntimeError("Error in IsLocalSolutionType: Should not execute this line... =( \n")
+
+
+def CountAllLocalGoodTypesNormalForm(Q, p, k, m, zvec, nzvec):
```
fchapoton: do not use CamelCase name for functions

fchapoton:
one failing doctest

WvanWoerden:
Should be resolved now. The old doctest failed the (already existing) input assumption that Q is given in local diagonal form. The new code uses this assumption and therefore gave unexpected results. Now the local diagonal form is passed to the function in the doctest. I verified the results also with release version 10.4.

WvanWoerden:
The previous build resulted in an error because some .coverage file was missing (see https://github.com/sagemath/sage/actions/runs/10962048022/job/30442695828 ).
Is there something I can do to fix this? The coverage report is fine on my local machine and it is unclear to me what this error is referring to. 

I also merged the latest develop branch again, maybe this will already resolve the issue.

fchapoton:
the documentation in `count_local_2` is not looking right ; you can see it by clicking above and navigating to that file in the reference manual

src/sage/quadratic_forms/count_local_2.pyx
```diff
@@ -281,3 +281,83 @@ cdef local_solution_type_cdef(Q, p, w, zvec, nzvec):
...
+
+    - ``Q`` -- quadratic form over `\ZZ` in local normal form at p with no zero blocks mod m_range
+    - ``p`` -- prime number > 0
+    - ``k`` -- integer > 0
+    - ``m`` -- integer >= 0 (depending only on mod `p^k`)
+    - ``zvec``, ``nzvec`` -- list of integers in ``range(Q.dim())``, or ``None``
+
+    OUTPUT:
+
+    an integer '\ge 0' representing the solutions of Good type.
```
S17A05: A few things here:
- If I see it correctly, m_range is always p^k here, so it might be nicer to write "mod `p^k`" in the first line
- In the output line '\ge 0' is not displaying properly since the wrong ticks are used; it should be `\ge 0`, or alternatively you could also write "a non-negative integer" instead (I think the latter is slightly better, but either should be fine)
- Not sure if it is common to say "representing the solutions" for the count of solutions, but maybe "giving the number of solutions" would be a bit more explicit?

WvanWoerden:
Thank you both for noticing this and for the suggestions. I've updated the docs accordingly.

S17A05:
> Thank you both for noticing this and for the suggestions. I've updated the docs accordingly.

Thanks for the changes, looks good to me now! Not sure why the workflows require approval, but since the last commit only changed documentation and the tests ran correctly before, I'll set this to positive review.

PR #38689, opened 2024-09-20
Fix tests with singular 4.4.0p3

antonio-rojas:
Fuzz some tests to make then pass with singular>=4.4.0p3, which returns different Gröbner bases in some cases

### DIFF
--- a/src/sage/libs/singular/option.pyx
+++ b/src/sage/libs/singular/option.pyx
@@ -28,7 +28,7 @@ this::
 
     sage: with opt_ctx(red_tail=False, red_sb=False):
     ....:    std(I)[-1]
-    d^2*e^6 + 8*c^3 + ...
+    d^2*e^6 +...8*c^3 + ...
 
 However, this does not affect the global state::
 
--- a/src/sage/libs/singular/option.pyx
+++ b/src/sage/libs/singular/option.pyx
@@ -26,9 +26,9 @@ By default, tail reductions are performed::
 If we don't want this, we can create an option context, which disables
 this::
 
-    sage: with opt_ctx(red_tail=False, red_sb=False): # random - depends on singular version
+    sage: with opt_ctx(red_tail=False, red_sb=False):
     ....:    notred = std(I)[-1]; notred
-    d^2*e^6 + 8*c^3 + ...
+    d^2*e^6 + ...
     sage: len(list(red)) < len(list(notred))
     True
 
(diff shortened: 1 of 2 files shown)


src/sage/rings/polynomial/multi_polynomial_ideal.py
```diff
@@ -176,7 +176,7 @@
     The Groebner basis modulo any product of the prime factors is also non-trivial::
 
         sage: I.change_ring(P.change_ring(IntegerModRing(2 * 7))).groebner_basis()
-        [x + 9*y + 13*z, y^2 + 3*y, y*z + 7*y + 6, 2*y + 6, z^2 + 3, 2*z + 10]
+        [x + ..., y^2 + 3*y, y*z + 7*y + 6, 2*y + 6, z^2 + 3, 2*z + 10]
```
orlitzky, line 179: The text above this test says that it's checking that the basis is non-trivial, so we could probably fix this one forever by simply comparing it to the trivial basis and looking for `False`

src/sage/libs/singular/option.pyx
```diff
@@ -28,7 +28,7 @@ this::
 
     sage: with opt_ctx(red_tail=False, red_sb=False):
     ....:    std(I)[-1]
-    d^2*e^6 + 8*c^3 + ...
+    d^2*e^6 +...8*c^3 + ...
```
orlitzky: One of these is supposed to be tail-reduced and the other not... we could do e.g. `reduced = std(I)[-1]` in the first case and `notreduced = std(I)[-1]` in the second, with the test then being `reduced != notreduced`?

antonio-rojas: OK. In this case it was still being tested since the last element of the reduced basis doesn't contain `8*c^3` but this makes it a bit more future proof.

src/sage/rings/polynomial/multi_polynomial_ideal.py
antonio-rojas: The list of leading monomials is deterministic so I think it's good to keep testing it, even though it's not the main goal of the test.

orlitzky:
Ok, thanks, LGTM then.

tornaria:
I now think that the "p" in singular version means "pre release" (i.e. "alpha" or "beta"; instead of "patch" as I used to think) since they keep breaking API and everything in these.

IOW, I'm proposing that we (distros, including sage-the-distro) stick with 4.4.0 until either 4.4.1 or 4.5.0 is released.

PR #38690, opened 2024-09-21
Tweak a few tests to pass when giac is not installed

orlitzky:
This is one of the prerequisites to make it possible to install sage without giac (https://github.com/sagemath/sage/issues/38668). These few tests can be made to work easily without a `# needs giac`, so let's do that.

### DIFF
--- a/src/sage/calculus/calculus.py
+++ b/src/sage/calculus/calculus.py
@@ -407,7 +407,7 @@ To check that :issue:`14821` is fixed::
     0.6321205588285577
     sage: result = integral(exp(-300.0/(-0.064*x+14.0)),x,0.0,120.0)
     ...
-    sage: result
+    sage: result  # abs tol 1e-10
     4.62770039817000e-9
 
 To check that :issue:`27092` is fixed::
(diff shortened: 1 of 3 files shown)


orlitzky:
As always, the CI fails are bogus/unrelated.

dcoudert (commented):
shouldn't you use annotations like `# needs sage.libs.giac` and/or `# needs sympy`  ?

orlitzky:
There will be a PR where I add a bunch of "needs" tags, but in these few cases I was able to update the expected output to work both with/without giac, so a new "needs" tag isn't necessary. Sympy is a standard package and our "integrate" commands already use it unconditionally; so long as nobody adds a `--disable-sympy` flag, these tests should pass unconditionally because at least one of the two packages will be installed.

It's unrelated to this issue, but integrating with `algorithm='giac'` currently uses the slow giac pexpect interface and not the library interface in `sage.libs.giac`. I've changed this in https://github.com/sagemath/sage/pull/38686 to use the better interface, and that PR does add some `# needs sage.libs.giac` tags (because after the PR, it's using the library interface).

We're also missing a way to detect the giac pexpect interface, but I've added that in https://github.com/sagemath/sage/pull/38672. In theory you could have the executable installed but not the library... some binary distros have been known to split packages up like this.

dcoudert (approved):
thank you for the explanaition. LGTM.

orlitzky:
Sure, thanks for the review!

PR #38692, opened 2024-09-22
Add make_set function for DisjointSets

### LINKED ISSUE #35599: DisjointSet() misses MakeSet method

### Problem Description

According to [documentation](https://doc.sagemath.org/html/en/reference/sets/sage/sets/disjoint_set.html), `DisjointSet()` supports only two operations `.find()` and `.union()`. 

### Proposed Solution

Add support for the operation of adding a new element (forming its own singleton set), which is referred to as `MakeSet()` operation in [Wikipedia](https://en.wikipedia.org/wiki/Disjoint-set_data_structure#Making_new_sets).

### Alternatives Considered

N/A

### Additional Information

_No response_

thecaligarmo:
This fixes #35599  by adding a `make_set` function to `DisjointSet` using `OrbitPartitions`. The documentation links to wikipedia.
From wikipedia, the method should be done in place and therefore there is no return

### :memo: Checklist

### :hourglass: Dependencies

#35599


src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,37 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
     if m_root != n_root:
         OP.num_cells -= 1
 
+
+cdef inline void OP_make_set(OrbitPartition *OP) noexcept:
+    cdef int i, n = OP.degree
+    cdef int *new_parent, *new_rank, *new_mcr, *new_size
+
+    cdef int *int_array = <int *> sig_malloc( 4*(n+1) * sizeof(int) )
+    if int_array is NULL:
```
dcoudert, line 147: Do you really want to call free on NULL ?

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,37 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
...
+        sig_free(int_array)
+    else :
+        OP.degree = n + 1
+        OP.num_cells = OP.num_cells + 1
+        new_parent = int_array
+        new_rank = int_array + (n + 1)
+        new_mcr = int_array + (2*n + 2)
+        new_size = int_array + (3 * n + 3)
+        for i from 0 <= i < n:
```
dcoudert: `for i in range(n)`

src/sage/sets/disjoint_set.pyx
```diff
@@ -834,6 +854,43 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
         cdef int j = <int> self._el_to_int[f]
         OP_join(self._nodes, i, j)
 
+    def make_set(self, new_elt = None):
+        r"""
+        Return a new disjoint set with an additional item.
```
dcoudert: You are not returning a new set, you are adding one new element to the set

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,37 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
...
+            new_parent[i] = OP.parent[i]
+            new_rank[i] = OP.rank[i]
```
dcoudert: why not using low level `memcpy` method ?

dcoudert (changes requested):
This is an interesting new functionality. However, why limiting to one extra element ?

thecaligarmo:
> This is an interesting new functionality. However, why limiting to one extra element ?

From what I understand from reading wikipedia, `DisjointSet` is a set of singleton sets where you can put sets together (`union`) and find where certain elements are located (`find`). `make_set` then allows you to add a new singleton set to your `DisjointSet`. From what I understand, you wouldn't want to add a set of multiple elements (as this defeats the point of `DisjointSet`) and if you want to add multiple singleton sets, you can call `make_set` multiple times.  In essence, I was trying to use wiki's implementation idea for `make_set`.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
thecaligarmo: Ah, sorry, I'll fix that.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
thecaligarmo: Will that work for c? If you look at all the other for loops in the .pxd file they all have the same format:
`for i from 0 <= i < n:`

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
thecaligarmo: I'm not perfect with c, but if I understand correctly, `memcpy` will also copy the amount of memory allocated for the object. Since we are working with arrays, that would mean that the new copy would not have enough memory allocated to it since we need to add an additional entry into the array.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
dcoudert: yes, it will work.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
dcoudert: The first option is to make 4 different mallocs for parent, rank, etc. when first creating the data structure.
Then, in `make_set`, you can use `realloc` to extend the size of the arrays. Realloc should copy the data for you and so you only have to allocate the last value. In case of error, one of these arrays will get a NULL pointer (unless I'm mistaken). Last, when deallocating, you have to do 4 free.

The other option is to use `memcpy` instead of the for loop to put the good values in `new_...`. It should be a call like `memcpy(new_parent, OP.parent, n)`. It remains to assign the last cell of `new_parent`.

dcoudert:
> > This is an interesting new functionality. However, why limiting to one extra element ?
> 
> From what I understand from reading wikipedia, `DisjointSet` is a set of singleton sets where you can put sets together (`union`) and find where certain elements are located (`find`). `make_set` then allows you to add a new singleton set to your `DisjointSet`. From what I understand, you wouldn't want to add a set of multiple elements (as this defeats the point of `DisjointSet`) and if you want to add multiple singleton sets, you can call `make_set` multiple times. In essence, I was trying to use wiki's implementation idea for `make_set`.

I agree that `make_set` is designed to add a new singleton set. But we could also add several new singleton sets at once to avoid multiple calls to `make_set`.

thecaligarmo:
> > > This is an interesting new functionality. However, why limiting to one extra element ?
> > 
> > 
> > From what I understand from reading wikipedia, `DisjointSet` is a set of singleton sets where you can put sets together (`union`) and find where certain elements are located (`find`). `make_set` then allows you to add a new singleton set to your `DisjointSet`. From what I understand, you wouldn't want to add a set of multiple elements (as this defeats the point of `DisjointSet`) and if you want to add multiple singleton sets, you can call `make_set` multiple times. In essence, I was trying to use wiki's implementation idea for `make_set`.
> 
> I agree that `make_set` is designed to add a new singleton set. But we could also add several new singleton sets at once to avoid multiple calls to `make_set`.

Yes, but it becomes ambiguous once you do that. For example, if I were to do: `d.make_set([1, 2])` is this adding the set `[1, 2]` or is it adding 2 sets `1` and `2`? The function `make_set` makes it sound like I would be adding only 1 set, not multiple sets, so my initial assumption would be that `make_set([1, 2])` would add a exactly 1 new set with entry `[1, 2]`. 

We could still make a function that does allow making multiple sets (say `make_sets`) which could then go through and do all the set creations at once (change `OP_make_set` to allow an optional second parameter). Or if you want to keep it as `make_set`, we can add an additional parameter called `multiple` that's a boolean that people can specify to show that they are adding multiple things instead of just one? so `make_set([1, 2])` would add 1 set `[1, 2]` and `make_set([1, 2], multiple=True)` would create two sets `1` and `2`.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
thecaligarmo: K, I've gone ahead and changed it.

For future: is there a reason why all the other ones use `for i from 0 <= i < n`? Because every single other for loop in the file uses this syntax rather than the `for i in range` syntax.

dcoudert:
Let's keep this for a future PR.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
thecaligarmo: Ok, I'm not as comfortable with C, so I'll see if I can try and use these methods.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
dcoudert: These methods have been written long time ago and Cython has evolved (improved) since. We have not update all the code yet, it's too big.

thecaligarmo:
Sounds good. 

I also tried to use the `memcpy` approach you had mentioned above. I ran the tests on my end and they seemed to work, but if there's an error, let me know. I wasn't sure if the 3rd parameter needed changing for the other 3 variables, but due to the previous memory allocation, I'm assuming not.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,35 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
...
+    cdef int *int_array = <int *> sig_malloc( 4*(n+1) * sizeof(int) )
+    if not int_array is NULL:
+        OP.degree = n + 1
+        OP.num_cells = OP.num_cells + 1
+        new_parent = int_array
+        new_rank = int_array + (n + 1)
+        new_mcr = int_array + (2*n + 2)
+        new_size = int_array + (3 * n + 3)
+
+        memcpy(new_parent, OP.parent, 4 * n * sizeof(int) )
```
dcoudert: why `4 * n` ? it should be `n`, no ?

dcoudert (changes requested):

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
thecaligarmo: !!! Yes. Sorry, I just realised my error there.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,35 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
     if m_root != n_root:
         OP.num_cells -= 1
 
+
+cdef inline void OP_make_set(OrbitPartition *OP) noexcept:
+    cdef int i, n = OP.degree
+    cdef int *new_parent, *new_rank, *new_mcr, *new_size
+
+    cdef int *int_array = <int *> sig_malloc( 4*(n+1) * sizeof(int) )
+    if not int_array is NULL:
```
dcoudert: The test should be `if int_array is not NULL:`

src/sage/sets/disjoint_set.pyx
```diff
@@ -834,6 +854,43 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
...
+            sage: e
+            {{'a'}, {'b', 'f'}, {'c', 'd', 'e'}}
+            sage: e.make_set('e'); e
+            {{'a'}, {'b', 'f'}, {'c', 'd', 'e'}}
+            sage: e.make_set(); e
+            {{'a'}, {'b', 'f'}, {'c', 'd', 'e'}, {6}}
+        """
+        if new_elt is None:
+            new_elt = self._nodes.degree
+        if not new_elt in self._int_to_el:
```
dcoudert: `if new_elt not in self._int_to_el:`

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,35 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
     if m_root != n_root:
         OP.num_cells -= 1
 
+
+cdef inline void OP_make_set(OrbitPartition *OP) noexcept:
+    cdef int i, n = OP.degree
+    cdef int *new_parent, *new_rank, *new_mcr, *new_size
+
+    cdef int *int_array = <int *> sig_malloc( 4*(n+1) * sizeof(int) )
+    if int_array is not NULL:
```
dcoudert: What if `int_array` is `NULL` ? you should certainly raise an error.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,35 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
...
+        memcpy(new_mcr, OP.mcr, n * sizeof(int) )
+        memcpy(new_size, OP.size, n * sizeof(int) )
+
+        new_parent[n] = n
+        new_rank[n] = 0
+        new_mcr[n] = n
+        new_size[n] = 1
+
+        OP.parent = new_parent
+        OP.rank   = new_rank
```
dcoudert: please avoid the extra spaces before `=`. This is not compatible with Python coding style.

src/sage/sets/disjoint_set.pyx
```diff
@@ -834,6 +854,43 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
...
+        r"""
+        Add a new element into a new set containing only the new element.
+
+        According to :wikipedia:`Disjoint-set_data_structure#Making_new_sets` the
+        `make_set` operation adds a new element into a new set containing only
+        the new element. The new set is added at the end of `self`.
+
+        INPUT:
+
+        - ``new_elt`` -- (optional) element to add. If `None`, then an integer
```
dcoudert, line 867: for the second line, you must add 2 extra spaces, like:
```
        - ``new_elt`` -- (optional) element to add. If `None`, then an integer
          is added.
```
this currently breaks building the pdf documentation (reported by CI).

src/sage/sets/disjoint_set.pyx
```diff
@@ -834,6 +854,43 @@ cdef class DisjointSet_of_hashables(DisjointSet_class):
         cdef int j = <int> self._el_to_int[f]
         OP_join(self._nodes, i, j)
 
+    def make_set(self, new_elt = None):
```
dcoudert: don't put spaces around `=` here. It should be `def make_set(self, new_elt=None):`

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -141,29 +141,31 @@ cdef inline void OP_make_set(OrbitPartition *OP) noexcept:
     cdef int i, n = OP.degree
     cdef int *new_parent, *new_rank, *new_mcr, *new_size
 
-    cdef int *int_array = <int *> sig_malloc( 4*(n+1) * sizeof(int) )
-    if int_array is not NULL:
+    cdef int *int_array = <int *> sig_malloc(4*(n+1) * sizeof(int))
+    if int_array is NULL:
+        raise MemoryError("MemoryError allocating int_array in make_set method")
+    else:
```
dcoudert: actually, you can remove the `else` and reduce the indentation of the code. Indeed, the code will be executed only if `int_array` is not `NULL`.

thecaligarmo: That's a good point. I'll switch it over.

src/sage/groups/perm_gps/partn_ref/data_structures.pxd
```diff
@@ -136,6 +136,37 @@ cdef inline void OP_join(OrbitPartition *OP, int m, int n) noexcept:
...
+    memcpy(new_rank, OP.rank, n * sizeof(int))
+    memcpy(new_mcr, OP.mcr, n * sizeof(int))
+    memcpy(new_size, OP.size, n * sizeof(int))
+
+    new_parent[n] = n
+    new_rank[n] = 0
+    new_mcr[n] = n
+    new_size[n] = 1
+
+    OP.parent = new_parent
```
dcoudert, line 169: before this statement, you must release the memory used by the previous arrays, so `sig_free(OP.parent)`.

thecaligarmo: Done. I'm not so good with memory allocation stuff, but I think we only need `sig_free(OP.parent)` correct? I tried adding the other ones and it started throwing errors, so I'm thinking that `OP.parent` holds the entire memory for all 4 variables so freeing up the memory from that element does it for all 4. If I did that wrong, just let me know.

dcoudert (approved):
LGTM.

maxale:
@thecaligarmo @dcoudert 
Thanks for implementing `.make_set()`, which effectively enables expanding `DisjointSets` with new elements or even building it from the initially empty set. However, this functionality can be further improved as explained in #39649. Please take a look.

PR #38699, opened 2024-09-23
let the category of sets handle the __len__ method of rings

fchapoton:
This removes the `__len__` method in `ring.pyx` and let the category of infinite sets handle this.

The test with `RR` fails in some other bizarre way, so it is marked as a bug.

But maybe it would be better to just remove `__len__` from rings instead ? People should use `cardinality`.

### :memo: Checklist


tscrim (changes requested):
The issue with `RR` is that it is an extension class and `__len__` is a special slot method in Cython. You can also see this failure with `ZZ`. So the thing to do is to change the `__len__` implementation to be in parallel to `Parent.__pow__`.

fchapoton:
thanks a lot, Travis, for your continuous help and useful insights !

tscrim:
You still have the issue with `RR`, `ZZ`, etc. correct?

mantepse:
I agree that `__len__` should not be a method in the category of rings.  What is the reason for having it?

fchapoton:
Should I use "NotImplementedError: infinite set" or "ValueError: infinite set" ?
Currently, both do appear. It may be a good occasion to have more uniformity.

tscrim:
I would say it would depend on if we expect the infinite set version to (eventually) work/be-implemented or not. Granted, we aren't consistent about this either... I would lean to ValueError as a general rule.

fchapoton:
So should I change the existing "NotImplementedError" ?

tscrim:
If you want to, go ahead.

fchapoton:
Well, I finally would prefer not to do further changes here.

PR #38719, opened 2024-09-26
Change wrapping of FLINT Z/nZ polynomial gcd failure

kylehofmann:
Sometimes FLINT fails to calculate GCDs in polynomial rings over Z/nZ. Since FLINT has no interface to report the failure, the GCD calculation needs to be wrapped and an error thrown. The present method of wrapping does not properly reactivate signals if an error happens, because the exception causes `sig_off()` to never be reached. This patch rectifies that. It also adds a doctest which triggers this code path; the example is designed so that the GCD exists but FLINT fails.

The error message is also adjusted, since the failure condition is more complicated than just encountering non-invertible elements (e.g., in Z/6, `gcd(2*x, x)` will work).


src/sage/libs/flint/nmod_poly_linkage.pxi
```diff
@@ -577,18 +577,37 @@ cdef inline int celement_gcd(nmod_poly_t res, nmod_poly_t a, nmod_poly_t b, unsi
...
         sig_on()
-        nmod_poly_gcd(res, a, b)
-        sig_off()
+        try:
+            nmod_poly_gcd(res, a, b)
+        finally:
+            sig_off()
     except RuntimeError:
-        raise ValueError("non-invertible elements encountered during GCD")
+        raise RuntimeError("FLINT gcd calculation failed") from None
```
tscrim:
```suggestion
        raise RuntimeError("FLINT gcd calculation failed")
```
I think it would be better to have the full context.

src/sage/libs/flint/nmod_poly_linkage.pxi
```diff
@@ -577,18 +577,37 @@ cdef inline int celement_gcd(nmod_poly_t res, nmod_poly_t a, nmod_poly_t b, unsi
...
+
+    cdef unsigned long leadcoeff = nmod_poly_get_coeff_ui(res, nmod_poly_degree(res))
+    cdef unsigned long modulus = nmod_poly_modulus(res)
+    if n_gcd(modulus,leadcoeff) == 1:
```
tscrim:
```suggestion
    if n_gcd(modulus, leadcoeff) == 1:
```
PEP8 spacing.

tscrim (changes requested):
Two very minor things. Otherwise LGTM.

src/sage/libs/flint/nmod_poly_linkage.pxi
kylehofmann: Are you really, really, really sure? I did that because the additional context you'd get is the following; I think it actively confuses the user because it makes the exception appear to originate from `sig_on()` instead of FLINT:

```
File ~/Projects/sage/src/sage/libs/flint/nmod_poly_linkage.pxi:599, in sage.rings.polynomial.polynomial_zmod_flint.celement_gcd()
    598 try:
--> 599     sig_on()
    600     try:

RuntimeError: Aborted

```

If you're sure, then we can do this, but I don't think it's a good idea.

tscrim: I do (in your words, really  really, really). Nearly all users encountering this will not care about where it originates (I also don't think it changing it helps say it is from FLINT), but it carries a bit extra useful information for developers.

kylehofmann: In that case, I concede; while I find the chained exception confusing, I'm also not a frequent Sage developer.

tscrim (approved):
Thank you. Positive review.

PR #38720, opened 2024-09-26
Raise exception when factoring zero polynomial

kylehofmann:
The zero polynomial doesn't have a factorization or square-free factorization, so
```
PolynomialRing(GF(7), 'x').zero().factor()
PolynomialRing(GF(7), 'x').zero().squarefree_decomposition()
```
should raise errors instead of returning zero. This patch fixes this behavior.


kwankyu (approved):
LGTM.

PR #38721, opened 2024-09-26
Type of Z/nZ NTL polynomial evaluation should be scalar

kylehofmann:
When a polynomial over Z/n implemented in NTL is evaluated, the result should be in Z/n.  Currently it's not; it's still an element of the polynomial ring.  Example:

```
sage: R.<x> = PolynomialRing(Zmod(4), 'x', implementation='NTL')
sage: x.subs(1).parent()
Univariate Polynomial Ring in x over Ring of integers modulo 4 (using NTL)
```

This patch fixes this behavior.


kwankyu (approved):
LGTM.

PR #38722, opened 2024-09-26
Roots of polynomials mod n

kylehofmann:
Presently, Sage implements root-finding over Z/n in a way that requires time exponential in e, where e is the largest exponent of a prime divisor of n.  For example, while Sage is capable of the following, it takes 8 minutes and 40 seconds on the laptop I'm using:
```python
R.<x> = Zmod(2^30)[]
(x^2 - 1).roots(multiplicities=False)
```
This patch implements a better algorithm.  Like the present one, it uses the Chinese Remainder Theorem to reduce to the case of prime powers.  Unlike the present algorithm, it finds roots modulo prime powers using Hensel lifting.  When the root is non-singular mod p, it uses Newton iteration and converges quadratically.  When the root is singular, lifting is done one power at a time, and the running time is bounded by the product of e and the maximum number of roots modulo any p^d with d ≤ e.  With this patch, the above code snippet takes 32 milliseconds; even lifting modulo 2^1000 takes less than a second.

Depends on #38720.


vincentmacri:
Can you add an `ALGORITHM` block to your docstrings that says where to find this algorithm in the literature?

grhkm21:
Hi, I just found out about this issue as well. FYI I believe Hensel lifting is already implemented when you solve equations over Qp or Zp, so I *think* a simpler implementation would be to change ring to that, solve it, then cast it back to the Zmod(p^e).

Aeren1564:
```cpp
for _ in range(p - 1):
```

This makes each lifting exponential in the size of $p$.

I think each lifting can be implemented in $O(\log(p) + \deg{F} + \mathrm{NumOfRoots})$.

(Note that it's almost the same as Hensel's lifting)

Let's say $r_n$ is one of the root modulo $p^n$, where $n \ge 1$, for polynomial $F(X) = \sum_{i=0}^{\infty} f_i \cdot X^i$. Each root $r_{n+1}$ of $F$ modulo $p^{n+1}$ with $r_{n+1} = r_n \mod p_n$ can be expressed as $r_{n+1} = p^n \cdot c + r_n$ for some integer $0 \le c < p$.

$$F(r_{n+1}) = \sum_{i=0}^{\infty} f_i \cdot (c \cdot p^n + r_n)^i = F(r_n) + \left(\frac{d}{dX}F \right) (r_i) \cdot p^n \cdot c = 0 \mod p^{n+1}$$

We know $F(r_n) = 0 \mod p^n$ so we only care about $\mathrm{floor}(F(r_n) / p^n) \mod p$.

Let $A := \mathrm{floor}(F(r_n) / p^n) \mod p$ and $B := \left(\frac{d}{dX}F \right) (r_i) \mod p$, then $A + B \cdot c = 0 \mod p$

1. $B \ne 0$
 $c = -A \cdot \mathrm{modinv}_p(B) \mod p$
2. $B = 0$
 If $A \ne 0$, $r_n$ cannot be lifted.
 If $A = 0$, all of $0, \cdots, p-1$ are valid values of $c$.

I'm not familiar with the syntax of sage so I'm not sure if it's allowed, but you can return a generator object instead of list so that users can iterate over the roots one by one instead of reading all at once.

kylehofmann:
The method you have described is essentially what I've implemented; see `_lift_residue_field_root` in the MR. It is exponential in $p$ because the number of roots can be exponential in $p$. For example, there are $p$ roots of $x^2$ in $\mathbf{Z}/p^2\mathbf{Z}$, namely the multiples of $p$.

Using a generator would be a major change in the semantics of the `.roots()` method and is out of scope for this MR.

kylehofmann:
I may as well add that it's possible that there's a better way in the case where $f'(r) \equiv 0 \pmod{p}$. The form of Hensel's lemma that's usually applied in this situation says that we can lift uniquely once $\lvert f(r)\rvert_p < \lvert f'(r)\rvert_p^2$. I would not be surprised if, when we expand the Taylor series out far enough that we can observe that kind of behavior, something algorithmically helpful happens. But I haven't investigated.

Aeren1564:
Oh I didn't notice that $\frac{d}{dx} F(r_n) \mod p$ will be constant when I was first reading the implementation. My bad.

src/doc/en/reference/references/index.rst
```diff
@@ -2921,6 +2921,9 @@ REFERENCES:
                  "Pyjamask"
                  https://csrc.nist.gov/CSRC/media/Projects/Lightweight-Cryptography/documents/round-1/spec-doc/Pyjamask-spec.pdf
 
+.. [Gouvea] Fernando Q. Gouvêa, *$p$-adic Numbers: An Introduction*,
```
vincentmacri:
```suggestion
.. [Gou2020] Fernando Q. Gouvêa, *$p$-adic Numbers: An Introduction*,
```

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,414 @@ def random_element(self, bound=None):
...
+          multiplicity. If ``False``, just return the unique roots, with no
+          information about multiplicities. Multiplicities are only defined
+          over fields, and this method raises ``NotImplementedError`` if this
+          is ``True`` but the ring is not a field.
+
+        - ``algorithm`` - ignored
+
+        ALGORITHM:
+
+        The algorithm is adapted from [Gouvea], section 4.5, and [Coh1993],
```
vincentmacri:
```suggestion
        The algorithm is adapted from [Gou2020]_, section 4.5, and [Coh1993]_,
```

src/doc/en/reference/references/index.rst
vincentmacri: First three letters followed by the year is what we usually do for single-authored works

src/sage/rings/finite_rings/integer_mod_ring.py
vincentmacri: Underscores make the documentation link to the reference (doesn't particularly matter since this function is internal and not included in the generated documentation, but we should still follow proper docstring formatting).

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,414 @@ def random_element(self, bound=None):
...
+        section 3.5.3.  It is a combination of the Chinese Remainder Theorem
+        and Hensel's lemma.  As a base case, if $N$ is prime, then we find
+        roots by factoring $f$.  If $N$ is a prime power $p^e$, then we find
```
vincentmacri: See the [developer guide](https://doc.sagemath.org/html/en/developer/coding_basics.html#section-latex-typeset) for how to do LaTeX formatting in our docstrings and change the formatting accordingly.

vincentmacri (changes requested):
Just some formatting feedback for now. It'll be a few days until I have time to look at the actual changes in more detail.

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,418 @@ def random_element(self, bound=None):
...
+        roots by factoring `f`.  If `N` is a prime power `p^e`, then we find
+        roots modulo `p` and lift them.  Finally, for general `N`, we first
+        factor the modulus `N` into prime powers, list all roots modulo those
+        prime powers, and combine the roots using the Chinese Remainder
+        Theorem.
+
+        Suppose that we are trying to find roots modulo `p^e` and that `r` is
+        a root of `f(x)` modulo `p`.  The easy case is when `f'(r) \not\equiv
+        0 \pmod{p}`, for then Hensel's lemma implies that there is a unique
+        `r_e \in \mathbf{Z}/p^e\mathbf{Z}` with `r_e \equiv r \pmod{p}`.
```
vincentmacri:
```suggestion
        `r_e \in \Zmod{p^e}` with `r_e \equiv r \pmod{p}`.
```

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,418 @@ def random_element(self, bound=None):
...
+        roots, and finding all the latter does not guarantee that we have
+        found all the former.  For example, if `f(x) = 2x` and `p = 2`, then
+        there is only one `p`-adic root, namely zero.  But the solutions of
+        `2x \equiv 0 \pmod{2^k}` are `0` and `2^{k-1}`; the former lifts to
+        two roots modulo `2^{k+1}`, namely `0` and `2^k`, while the latter
+        does not lift at all.  We handle this case by lifting one power at a
+        time.  While we can no longer use Newton's method to solve for a lift,
+        the Taylor series it is based on still yields constraints on the roots
+        modulo `p^{k+1}`:  If `r_k` is a root of `f` modulo `p^k`, then either
+        every lift of `r_k` to `\mathbf{Z}/p^{k+1}\mathbf{Z}` is a root of `f`
```
vincentmacri:
```suggestion
        every lift of `r_k` to `\Zmod{p^{k + 1}}` is a root of `f`
```

vincentmacri (changes requested):
More formatting comments. I hope to find time to review the actual code next week. Sage has a small set of LaTeX macros. Using them is good to ensure consistent style throughout the documentation.

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,418 @@ def random_element(self, bound=None):
...
+            fp = fpe.change_ring(Zmod(p))
+
+            mod_p_roots = fp.roots(multiplicities=False)
+
+            this_prime_power = []
+            prime_power_roots.append(this_prime_power)
+            for root in mod_p_roots:
+                this_prime_power.extend(
+                    self._lift_residue_field_root(p, e, fpe, fpe_prime, root)
+                )
```
vincentmacri, line 1954: Where do the values you add to `this_prime_power` get used?

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,418 @@ def random_element(self, bound=None):
...
+        for p, e in fac:
+            Zpe = Zmod(p**e)
+            fpe = f.change_ring(Zpe)
+            fpe_prime = fpe.derivative()
+            fp = fpe.change_ring(Zmod(p))
+
+            mod_p_roots = fp.roots(multiplicities=False)
+
+            this_prime_power = []
+            prime_power_roots.append(this_prime_power)
```
vincentmacri, line 1950: This doesn't do anything because `this_prime_power` is empty.

vincentmacri (commented):
I still need to look at the Newton iteration part of `_lift_residue_field_root` and the CRT part of `_roots_univariate_polynomial`.

The non-unique lift part of `_lift_residue_field_root` looks good.

src/sage/rings/finite_rings/integer_mod_ring.py
vincentmacri:
```suggestion
            this_prime_power = []
            for root in mod_p_roots:
                this_prime_power.extend(
                    self._lift_residue_field_root(p, e, fpe, fpe_prime, root)
                )
            prime_power_roots.append(this_prime_power)
```

I think this is more readable. I got tripped up reading this because the current code looks like you're appending an empty list and then doing nothing with `this_prime_power`, then I remembered that there's pointer/reference stuff going on.

kylehofmann: I'm in the habit of doing this kind of append immediately so that I don't forget it entirely. But I agree that it might be confusing, so it seems like a worthwhile change.

src/sage/rings/finite_rings/integer_mod_ring.py
```diff
@@ -1547,6 +1551,418 @@ def random_element(self, bound=None):
...
+        Test polynomials with content:
+
+            sage: R.<x> = Zmod(4)[]
+            sage: (2*x).roots(multiplicities=False)
+            [0, 2]
+
+            sage: R.<x> = Zmod(6)[]
+            sage: (3*x).roots(multiplicities=False)
+            [0, 4, 2]
+
```
vincentmacri, line 1826:
```suggestion

        Test polynomial with many roots:

                    sage: R.<x> = Zmod(6)[]
                    sage: f = x * (x - 1) * (x - 2) * (x - 3) * (x - 4) * (x - 5)
                    sage: len(f.roots(multiplicities=False))
                    6

```

vincentmacri: I think with the way you're doing the CRT there is the possibility for the same root to be returned twice. This test should hit that edge case unless duplicate roots are being removed elsewhere when calling `f.roots`.

If you do `f.base_ring()._roots_univariate_polynomial(f, multiplicities=False)` I think it will return a list of length 108.

vincentmacri (changes requested):
I think I found an issue with the CRT part. Rest of the code looks good.

Can you add the suggested test and if the test fails as I expect, fix the code so that the test passes?

src/sage/rings/finite_rings/integer_mod_ring.py
kylehofmann: This works as-is. The roots are deduplicated in the base case, when we look for roots over finite fields. They can't reappear during the CRT because the CRT is an isomorphism.

Also, `f.base_ring()._roots_univariate_polynomial(f, multiplicities=False)` returns a list of length 6.

vincentmacri: You're right. Not sure what went wrong when I was testing and got something else.

vincentmacri:
LGTM. Thanks for the contribution!

vincentmacri (approved):
LGTM. Thanks for the contribution!

PR #38742, opened 2024-09-29
graphs/matching_covered_graph: introduce the class `MatchingCoveredGraph`

janmenjayap:
The objective of this PR is to introduce a new class  `MatchingCoveredGraph` through a new file `src/sage/graphs/matching_covered_graph.py` in order to address the decompositions, generation methods and related concepts in the theory of Matching Covered Graphs.

This PR introduces a new class pertaining to matching, namely `MatchingCoveredGraph` and aims to list out all functions fundamentally related to matching covered graph in the file `src/sage/graphs/matching_covered_graph.py`. The initialization and some basic class methods in this context, that shall be addressed through this PR, are described below:

- [x] `__init__()`: Create a matching covered graph, that is a connected nontrivial graph wherein each edge participates in some perfect matching.
- [x] `__repr__()`: Return a short string representation of the matching covered graph.
- [x] `_subgraph_by_adding()`: Return the matching covered subgraph containing the given vertices and edges.
- [x] `_upgrade_from_graph()`: Upgrade the given graph to a matching covered graph if eligible.
- [x] `add_edge()`: Add an edge from vertex ``u`` to vertex ``v``.
- [x] `add_edges()`: Add edges from an iterable container.
- [x] `add_vertex()`: Add a vertex to the (matching covered) graph.
- [x] `add_vertices()`: Add vertices to the (matching covered) graph from an iterable container of vertices.
- [x] `allow_loops()`: Change whether loops are allowed in (matching covered) graphs.
- [x] `allows_loops()`: Return whether loops are permitted in (matching covered) graph.
- [x] `delete_vertex()`: Delete a vertex, removing all incident edges.0
- [x] `delete_vertices()`: Delete specified vertices form ``self``.
- [x] `get_matching()`: Return a :class:`~EdgesView` of ``self._matching`` (a perfect matching of the (matching covered) graph computed at the initialization).
- [x] `has_perfect_matching()`: Return whether the graph has a perfect matching.
- [x] `update_matching()`: Update the perfect matching captured in ``self._matching``.
 

This PR shall establish a foundation to address the methods related to matching covered graphs.

Fixes #38216.
Note that this issue fixes a small part of the above-mentioned issue. 

### :memo: Checklist

### :hourglass: Dependencies

Nothing as of now (up to my knowledge).

cc: @dcoudert.


janmenjayap:
Hi,
I have a doubt. the class `MatchingCoveredGraph` inherits the class `Graph` and the class `Graph` inherits the class `GenericGraph`. Still the class `MatchingCoveredGraph` is not able to access some private members of the class `GenericGraph` such as `self._backend`, as a consequence of which the following test case does not pass. In fact, creating an object of the class `MatchingCoveredGraph` throws error:

```
sage: G = graphs.CompleteGraph(4)
sage: H = MatchingCoveredGraph(G)
AttributeError                            Traceback (most recent call last)
<ipython-input-2-d169ce49e91a> in ?()
----> 1 H = MatchingCoveredGraph(G)

~/open-source/sage/src/sage/graphs/matching_covered_graph.py in ?(self, data, matching, algorithm, solver, verbose, integrality_tolerance, *args, **kwds)
    396         elif isinstance(data, MatchingCoveredGraph):
    397             Graph.__init__(self, data, *args, **kwds)
    398 
    399         else:
--> 400             self._upgrade_from_graph(matching, algorithm, solver, verbose,
    401                                      integrality_tolerance)
    402 
    403         del self.allow_loops

~/open-source/sage/src/sage/graphs/matching_covered_graph.py in ?(self, matching, algorithm, solver, verbose, integrality_tolerance)
    419 
    420             if not check:
    421                 raise ValueError("input graph is not matching covered")
    422 
--> 423         except ValueError as error:
    424             raise error

~/open-source/sage/src/sage/graphs/matching.py in ?(G, matching, algorithm, coNP_certificate, solver, verbose, integrality_tolerance)
    960     AUTHORS:
    961 
    962     - Janmenjaya Panda (2024-06-23)
    963     """
--> 964     G._scream_if_not_simple(allow_multiple_edges=True)
    965 
    966     # The graph must be nontrivial
    967     if G.order() < 2:

~/open-source/sage/src/sage/graphs/generic_graph.py in ?(self, allow_loops, allow_multiple_edges)
   1481             Traceback (most recent call last):
   1482             ...
   1483             ValueError: This method is not known to work on graphs with loops. Perhaps this method can be updated to handle them, but in the meantime if you want to use it please disallow loops using allow_loops().
   1484         """
-> 1485         pb_with_loops = not allow_loops and self.allows_loops()
   1486         pb_with_multiple_edges = not allow_multiple_edges and self.allows_multiple_edges()
   1487         if pb_with_loops or pb_with_multiple_edges:
   1488             if pb_with_loops and pb_with_multiple_edges:

~/open-source/sage/src/sage/graphs/generic_graph.py in ?(self)
   3279             False
   3280             sage: D.edges(sort=True)
   3281             []
   3282         """
-> 3283         return self._backend.loops(None)

AttributeError: 'MatchingCoveredGraph' object has no attribute '_backend'
```

A comment on this would be highly appreciated.
Thank you.

dcoudert:
Check `src/sage/graphs/bipartite_graph.py`.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,464 @@
...
+            raise ValueError('the graph is trivial')
+
+        elif isinstance(data, MatchingCoveredGraph):
+            Graph.__init__(self, data, *args, **kwds)
+
+        else:
+            self._upgrade_from_graph(matching, algorithm, solver, verbose,
+                                     integrality_tolerance)
+
+        del self.allow_loops
```
dcoudert: Many methods will fail because of that. Check how it's done for BipartiteGraph.

dcoudert (changes requested):

janmenjayap:
Hi,
Could you please mention the requirement of `*` in the method `matching()` in `src/sage/graphs/matching.py`.

Actually, the errors in the test cases that are coming up, I suppose due to the use of `*` in the method `is_matching_covered()` in  `src/sage/graphs/matching.py`.

Thank you.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,522 @@
...
+        True
+        sage: G = Graph([(0, 1), (0, 3), (0, 4), (1, 2), (1, 5), (2, 3), (2, 6), (3, 7), (4, 5), (4, 7), (5, 6), (6, 7)])
+        sage: H = MatchingCoveredGraph(G)
+        sage: H == G
+        True
+        sage: import igraph                                                      # optional - python_igraph
+        sage: G = Graph(igraph.Graph([(0, 1), (0, 3), (1, 2), (2, 3)]))          # optional - python_igraph
+        sage: H = MatchingCoveredGraph(G)
+        sage: H
+        Matching covered graph on 4 vertices
```
janmenjayap, line 342: This test seems correct, isn't it?
The graph `H` is isomorphic to a cycle graph on 4 vertices.

janmenjayap:
> Hi, Could you please mention the requirement of `*` in the method `matching()` in `src/sage/graphs/matching.py`.
> 
> Actually, the errors in the test cases that are coming up, I suppose due to the use of `*` in the method `is_matching_covered()` in `src/sage/graphs/matching.py`.
> 
> Thank you.

I have rectified the errors caused in `matching_covered_graph.py` without changing anything in `matching.py`.

src/sage/graphs/matching_covered_graph.py
janmenjayap: This issue was due to the absence of the the comment `# optional - python_igraph` in line no. 165 and 166. It has been resolved.

dcoudert:
The `*` in the list of parameters means that the parameters placed after the `*` must be explicitly named.
```py
sage: def foo(a=1, b=2, *, c=3):
....:     return a + b + c
sage: foo()
6
sage: foo(7)
12
sage: foo(7, 8)
18
sage: foo(7, 8, 9)
Cell In[7], line 1
----> 1 foo(Integer(7), Integer(8), Integer(9))

TypeError: foo() takes from 0 to 2 positional arguments but 3 were given
sage: foo(7, 8, c=9)
24
sage: foo(a=10)
15
sage: foo(b=10)
14
sage: foo(c=10)
13
sage: foo(b=20, c=10)
31
```

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+    TESTS:
+
+    An empty graph is not matching covered::
+
+        sage: G = Graph()
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: the graph is trivial
+        sage: G = Graph(None)
```
dcoudert: this is redundant with the previous test, right?

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: the graph is trivial
+        sage: G = MatchingCoveredGraph()
+        Traceback (most recent call last):
+        ...
+        ValueError: the graph is trivial
+        sage: G = MatchingCoveredGraph(None)
```
dcoudert: this is redundant with the previous test and so can be removed.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        Traceback (most recent call last):
+        ...
+        ValueError: the graph is trivial
+
+    Providing with a graph that is not connected::
+
+        sage: cycle1 = graphs.CycleGraph(4)
```
dcoudert: a simpler version:
```py
sage: G = graphs.CycleGraph(4)
sage: G += graphs.CycleGraph(6)
sage: G.connected_components_number()
2
```

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        ...
+        ValueError: input graph is not matching covered
+        sage: G = Graph({0: [1, 6, 11], 1: [2, 4], 2: [3, 5], 3: [4, 5],
+        ....:            4: [5], 6: [7, 9], 7: [8, 10], 8: [9, 10], 9: [10],
+        ....:            11: [12, 14], 12: [13, 15], 13: [14, 15], 14: [15]})
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: input graph is not matching covered
+        sage: # needs networkx
```
dcoudert, line 485: add an empty line before the line `sage: # needs networkx`.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: input graph is not matching covered
+        sage: G = Graph('F~~~w')
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: input graph is not matching covered
+        sage: # needs sage.modules
```
dcoudert, line 498: add an empty line before `sage: # needs sage.modules`. Otherwise, all the tests after will also need. networkx`.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        [0 0 0 0 0 0 0 0 0 0 0 1 0 1 0 1]
+        [0 0 0 0 0 0 0 0 0 0 0 0 1 0 1 1]
+        [0 0 0 0 0 0 0 0 0 0 0 1 0 1 0 1]
+        [0 0 0 0 0 0 0 0 0 0 0 0 1 1 1 0]
+        sage: G = Graph(M)
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: input graph is not matching covered
+        sage: # needs sage.modules
```
dcoudert, line 538: this `needs` statement is not needed has already used before.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+        sage: G = Graph([(11, 12), (11, 14), (0, 1), (0, 11), (0, 6), (1, 2),
+        ....:            (1, 4), (2, 3), (2, 5), (3, 4), (3, 5), (4, 5),
+        ....:            (6, 7), (6, 9), (7, 8), (7, 10), (8, 9), (8, 10),
+        ....:            (9, 10), (12, 13), (12, 15), (13, 14), (13, 15),
+        ....:            (14, 15)])
+        sage: H = MatchingCoveredGraph(G)
+        Traceback (most recent call last):
+        ...
+        ValueError: input graph is not matching covered
+        sage: # optional - python_igraph
```
dcoudert, line 559: add an empty line before

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+
+        elif isinstance(data, MatchingCoveredGraph):
+            Graph.__init__(self, data, *args, **kwds)
+
+        elif isinstance(data, Graph):
+            Graph.__init__(self, data, *args, **kwds)
+            self._upgrade_from_graph(matching=matching, algorithm=algorithm,
+                                     solver=solver, verbose=verbose,
+                                     integrality_tolerance=integrality_tolerance)
+        else:
```
dcoudert: remove the `else` and reduce indentation of the `raise`. It's the default action of the method.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+    def _upgrade_from_graph(self, matching=None, algorithm='Edmonds',
+                            solver=None, verbose=0,
+                            integrality_tolerance=0.001):
+        """
+        Upgrade the given graph to a matching covered graph if eligible
+
+        See documentation ``MatchingCoveredGraph?`` for detailed information.
+        """
+        try:
+            coNP_certificate = False
```
dcoudert: why do you need this variable ?

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+                                     solver=solver, verbose=verbose,
+                                     integrality_tolerance=integrality_tolerance)
+        else:
+            raise ValueError('input data is of unknown type')
+
+    def _upgrade_from_graph(self, matching=None, algorithm='Edmonds',
+                            solver=None, verbose=0,
+                            integrality_tolerance=0.001):
+        """
+        Upgrade the given graph to a matching covered graph if eligible
```
dcoudert: a `.` is missing at the end of this sentence. Same for other methods.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+            sage: J  # An object of the class MatchingCoveredGraph
+            Matching covered hexahedron: graph on 8 vertices
+        """
+        s = Graph._repr_(self).lower()
+        if "matching covered" in s:
+            return s.capitalize()
+        return "".join(["Matching covered ", s])
+
+    def allow_loops(self, new, check=True):
+        """
```
dcoudert, line 900: use `r"""`. Same for all methods.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,534 @@
...
+            sage: P = graphs.PetersenGraph()
+            sage: P.is_matching_covered()
+            True
+            sage: G = MatchingCoveredGraph(P)
+            sage: G.allow_loops(True)
+            Traceback (most recent call last):
+            ...
+            ValueError: loops are not allowed in matching covered graphs
+        """
+        if new is True:
```
dcoudert: change to `if new:`

dcoudert (changes requested):

src/sage/graphs/matching_covered_graph.py
janmenjayap: If I am not mistaken, I suppose, in that case, we need to add `return` at the end of the `if`/ `elif`?

dcoudert: I mistakenly suggested that. You are right, the code is ok like that.

dcoudert:
Now that we have this class, what happens if I add/remove a vertex or an edge ? 
You should certainly adapt methods modifying the graph to ensure that a modified graph remains in the class and otherwise raise an error without modifying the graph.

dcoudert (commented):
It's generally better to use a hidden variable like `_matching` and to define a function to get the matching, like `def get_matching(self)`.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+                    raise ValueError("input graph is not matching covered")
+
+            except Exception as exception:
+                raise exception
+
+        if success:
+            if matching:
+                # The input matching must be a valid perfect matching of the graph
+                M = Graph(matching)
+                G_simple = self.to_simple()
```
dcoudert: This should be done after the test on the validity of the matching

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+
+            sage: W = graphs.WheelGraph(6)
+            sage: G = MatchingCoveredGraph(W)
+            sage: G.add_edge((1, 4), label='label')
+            sage: G.edges(sort=False)  # No alteration to the existing graph
+            [(0, 1, None), (0, 2, None), (0, 3, None), (0, 4, None), (0, 5, None), (1, 2, None), (1, 4, 'label'), (1, 5, None), (2, 3, None), (3, 4, None), (4, 5, None)]
+
+        An expression, analogous to the syntax mentioned above may be used::
+
+            sage: W = graphs.WheelGraph(6)
```
dcoudert: Well, you can certainly combine all these tests to show the different ways of adding an edge, for instance first adding edge (1, 4), then another edge connecting different vertices (possibly adding new vertices to the graph, etc.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+                try:
+                    u, v = u
+                except Exception:
+                    pass
+
+        if u in self and v in self:
+            if u == v:
+                raise ValueError('loops are not allowed in '
+                                 'matching covered graphs')
+            G = Graph(self)
```
dcoudert: do we really need to create a new graph and check if it is matching covered ? can't we find a lighter incremental method ? 
At least, you can make add a TODO block at the top of the module to list what should be done to improve the situation.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+        Adding an existing vertex::
+
+            sage: P = graphs.PetersenGraph()
+            sage: P
+            Petersen graph: Graph on 10 vertices
+            sage: G = MatchingCoveredGraph(P)
+            sage: G
+            Matching covered petersen graph: graph on 10 vertices
+            sage: # needs random
+            sage: u = random.choice(list(G))
```
dcoudert: here it's sufficient to try to add vertex 0. No need for random.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            sage: G.add_vertex(u)
+            sage: G
+            Matching covered petersen graph: graph on 10 vertices
+
+        Adding a new/ non-existing vertex::
+
+            sage: P = graphs.PetersenGraph()
+            sage: P
+            Petersen graph: Graph on 10 vertices
+            sage: G = MatchingCoveredGraph(P)
```
dcoudert: here you can reuse the graph G created in the previous block.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            sage: G
+            Matching covered petersen graph: graph on 10 vertices
+            sage: G.add_vertex()
+            Traceback (most recent call last):
+            ...
+            ValueError: isolated vertices are not allowed in matching covered graphs
+            sage: u = 0
+            sage: while u in G:
```
dcoudert: juste try to add vertex 10 or 100. It's enough.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+        Adding a list of already existing vertices::
+
+            sage: T = graphs.TruncatedBiwheelGraph(15)
+            sage: T
+            Truncated biwheel graph: Graph on 30 vertices
+            sage: G = MatchingCoveredGraph(T)
+            sage: G
+            Matching covered truncated biwheel graph: graph on 30 vertices
+            sage: # needs random
+            sage: import random
```
dcoudert: Same comments than for `add_vertex`. Make simpler tests.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            Traceback (most recent call last):
+            ...
+            ValueError: isolated vertices are not allowed in matching covered graphs
+            sage: S3 = [2, None, None, 5]
+            sage: G.add_vertices(S3)
+            Traceback (most recent call last):
+            ...
+            ValueError: isolated vertices are not allowed in matching covered graphs
+        """
+        for vertex in vertices:
```
dcoudert: use a form like:
```py
if any(v not in self for v in vertices):
    raise ValueError(...)
```

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+          covered) graph.
+
+        EXAMPLES:
+
+        Deleting a non-existent vertex::
+
+            sage: W = graphs.WheelGraph(12)
+            sage: G = MatchingCoveredGraph(W)
+            sage: u = 0
+            sage: while u in G:
```
dcoudert: no need for that. We know that vertex 12 is not in the graph.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            ...
+            ValueError: vertex (12) not in the graph
+
+        Deleting an existing vertex::
+
+            sage: W = graphs.WheelGraph(12)
+            sage: G = MatchingCoveredGraph(W)
+            sage: # need random
+            sage: import random
+            sage: u = random.choice(list(G))
```
dcoudert: just try to remove vertex 0 or 1

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+
+        if vertex not in self:
+            raise ValueError('vertex (%s) not in the graph' % str(vertex))
+
+        raise ValueError('odd order is not allowed for '
+                         'matching covered graphs')
+
+    def delete_vertices(self, vertices):
+        r"""
+        Delete vertices (along with the incident edges) from the (matching
```
dcoudert: rephrase like:
```
Delete specified vertices form ``self``.

This method deletes the vertices from the iterable container ``vertices`` from ``self`` along with incident edges. An error is raised if the resulting graph is not matching covered.
```

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            Staircase graph: Graph on 8 vertices
+            sage: G = MatchingCoveredGraph(S)
+            sage: G
+            Matching covered staircase graph: graph on 8 vertices
+            sage: T = [6, 7]
+            sage: G.delete_vertices(T)
+            sage: G  # Matching covered graph on 6 vertices
+            Matching covered staircase graph: graph on 6 vertices
+        """
+
```
dcoudert, line 1853: remove empty line

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+        for vertex in vertices:
+            if vertex not in self:
+                raise ValueError('vertex (%s) not in the graph' % str(vertex))
+
+        try:
+            G = Graph(self)
+            G.delete_vertices(vertices)
+            G_simple = G.to_simple()
```
dcoudert: why do you need a simple copy ? you could also set parameter `multiedges=False` when creating `G`.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+
+            M = Graph(self.get_matching())
+
+            if M:
+                M.delete_vertices(vertices)
+                # The resulting matching after the removal of the input vertices
+                # must be a valid perfect matching of the resulting graph obtained
+                # after the removal of the vertices
+
+                if any(d != 1 for d in G_simple.degree()) \
```
dcoudert: I don't understand this test. For instance, how could an edge of `M` not be in `G_simple` ?

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            sage: G.get_matching()
+            [(0, 5, None), (1, 6, None), (2, 7, None), (3, 8, None), (4, 9, None)]
+            sage: M == G.get_matching()
+            True
+        """
+        return self._matching
+
+    def update_matching(self, matching):
+        r"""
+        Update the pefect matching captured in ``self._matching``.
```
dcoudert: pefect -> perfect

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+            Traceback (most recent call last):
+            ...
+            ValueError: the input is not a perfect matching of the graph
+        """
+        try:
+            M = Graph(matching)
+            G = Graph(self)
+            G_simple = G.to_simple()
+
+            if any(d != 1 for d in M.degree()):
```
dcoudert, line 2078: do this before creating `G`.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1228 @@
...
+                raise ValueError("the input is not a matching")
+
+            if any(not G_simple.has_edge(edge) for edge in M.edge_iterator()):
+                raise ValueError("the input is not a matching of the graph")
+
+            if (G_simple.order() != M.order()) or (G_simple.order() != 2*M.size()):
```
dcoudert: isn't it sufficient to check the orders of both graphs ? why do we need to check `G_simple.order() != 2*M.size()` ?

dcoudert (changes requested):
Many tests must be simplified.

src/sage/graphs/matching_covered_graph.py
janmenjayap: But, we have a comment followed by a `::` just before that. Won't that result in a different code box in the documentation UI?

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1221 @@
...
+        sage: # needs networkx
+        sage: import networkx
+        sage: G = Graph(networkx.complete_bipartite_graph(12, 12))
+        sage: H = MatchingCoveredGraph(G)
+        sage: H
+        Matching covered graph on 24 vertices
+        sage: H == G
+        True
+        sage: H.get_matching()
+        [(4, 23, None), (6, 21, None), (11, 16, None), (7, 20, None), (10, 17, None), (1, 14, None), (2, 13, None), (8, 19, None), (9, 18, None), (3, 12, None), (5, 22, None), (0, 15, None)]
```
dcoudert: you should sort the edges, for instance using `sorted(H.get_mathcing()`. Indeed, the order of the edges may vary and so the doctest may fail.

Furhtermore, could you improve the output like
```
        [(4, 23, None), (6, 21, None), (11, 16, None), (7, 20, None),
	 (10, 17, None), (1, 14, None), (2, 13, None), (8, 19, None),
	 (9, 18, None), (3, 12, None), (5, 22, None), (0, 15, None)]
```
and do the same each time you output edges in doctests.

src/sage/graphs/matching_covered_graph.py
dcoudert: this has not been answered yet.

dcoudert (changes requested):

src/sage/graphs/matching_covered_graph.py
janmenjayap: While adding a new edge to the graph, we need to check whether it lies in a perfect matching or not and that is in  the order of `|V| + |E|`. We are creating a new instance of the existing graph and after addition of the new edge, checking whether this new edge lies in a perfect matching or not using the current matching (this shall take linear time as we are using M alternating tree search method). So, even if we do not create a new instance of the graph, the overall complexity, in the worst case, is in linear.

src/sage/graphs/matching_covered_graph.py
janmenjayap: Yes. This is my bad. I have corrected it.

src/sage/graphs/matching_covered_graph.py
dcoudert: yes, it creates a new code box, but this is something we do frequently to have a lighter, yet meaningful, documentation

src/sage/graphs/matching_covered_graph.py
dcoudert: ok

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1274 @@
...
+            ...
+            ValueError: the graph obtained after the addition of edge
+            ((4, 5, None)) is not matching covered
+            sage: G.edges(sort=False) # No alteration to the existing graph
+            [(0, 1, None), (0, 3, None), (1, 2, None), (2, 3, None)]
+
+        Adding a self-loop::
+
+            sage: H = graphs.HeawoodGraph()
+            sage: G = MatchingCoveredGraph(P)
```
dcoudert: it should be `H`, not `P`.

dcoudert (changes requested):

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1268 @@
...
+            Staircase graph: Graph on 8 vertices
+            sage: G = MatchingCoveredGraph(S)
+            sage: G
+            Matching covered staircase graph: graph on 8 vertices
+            sage: T = [6, 7]
+            sage: G.delete_vertices(T)
+            sage: G  # Matching covered graph on 6 vertices
+            Matching covered staircase graph: graph on 6 vertices
+        """
+        for vertex in vertices:
```
dcoudert, line 1854: Before this for loop, you must add basic test like:
start with a test like:
```py
if not vertices:   # do nothing
    return
vertices = set(vertices)  # to remove potentially duplicated vertices
if len(vertices) % 2:  # try to remove an odd number of vertices
    raise ValueError(....)
```

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1268 @@
...
+            if vertex not in self:
+                raise ValueError('vertex (%s) not in the graph' % str(vertex))
+
+        try:
+            G = Graph(self, multiedges=False)
+            G.delete_vertices(vertices)
+
+            M = Graph(self.get_matching())
+
+            if M:
```
dcoudert: if `M` is `None` or empty, it means that self is an empty graph, right ? if so, you should not try to remove vertices from an empty graph.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1268 @@
...
+            sage: T = [6, 7]
+            sage: G.delete_vertices(T)
+            sage: G  # Matching covered graph on 6 vertices
+            Matching covered staircase graph: graph on 6 vertices
+        """
+        for vertex in vertices:
+            if vertex not in self:
+                raise ValueError('vertex (%s) not in the graph' % str(vertex))
+
+        try:
```
dcoudert, line 1863: before this block, check if `G.order() == len(vertices)`. If so, you remove all vertices and so the resulting graph is empty and has no matching. Easy case.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1268 @@
...
+            ...
+            ValueError: the input is not a perfect matching of the graph
+        """
+        try:
+            M = Graph(matching)
+
+            if any(d != 1 for d in M.degree()):
+                raise ValueError("the input is not a matching")
+
+            G = Graph(self, multiedges=False)
```
dcoudert: this copy is not needed. You want to check if `self` contains some edges and if the number of vertices is consistent with the number of vertices of the matching.

dcoudert (changes requested):

src/sage/graphs/matching_covered_graph.py
janmenjayap: Actually, this `if` statement is redundant, as `M` can not be `None` or empty, since `self` must have at least two vertices.
I will be removing this `if` statement.

janmenjayap:
If I am not mistaken, the list of methods that are remaining to be addressed through this PR, are the following:
- `delete_edge()`
- `delete_edges()`
- `is_removable_edge()`: To check whether an edge is removable, that is: the graph obtained after the removal of that edge is still matching covered
- `is_removable_doubleton()`: To check whether a pair of edges are removable together (note that, each of those edges shall **not** be removable individually as per the definition of a removable doubleton), that is: the graph obtained after the removal of the pair of edges is still matching covered
- `removble_edges()`: An iterator over all removable edges
- `removable_doubletons()`: An iterator over all removable doubletons
- `is_removable_ear()`: 
- `is_removable_double_ear()`
- `removable_ears()`
- `removable_double_ears()`

Please note that each of the methods (except from the iterators) apart from `delete_edge()` and `delete_edges()`, is a special case of these two that can be dealt with lower time complexity as compared to deleting the respective edges and checking whether the resulting graph is matching covered by computing a perfect matching (which is the bottleneck step).

Please let me know whether these methods shall be addressed in this PR or in a separate PR?
Thank you.

dcoudert:
This PR is already very large. What I suggest is:
- add a TODO block in the description of the module with the list of methods to be added.
- start adding new methods in other PRs, of course depending on this PR.

janmenjayap:
Also, shall the method `_subgraph_by_adding()` return an instance of `MatchingCoveredGraph` or an instance of `Graph`?

tobiasdiez:
The meson tests fail since the new file is not installed. For this you would need to add it at the beginning of the `sage/graphs/meson.build` file.

janmenjayap:
Thank you.

janmenjayap:
Could you please let me know why after doing `{INDEX_OF_METHODS}` and using `@doc_index()`, the methods are not appearing in the description of the html file?

dcoudert:
> Shall the method `get_matching()` be named `matching()` in order to overwrite the method `matching()` as it will take less complexity since we already have a perfect matching in `self._matching`?

I'm not sure yet. Indeed, method `matching` returns a maximum weight matching, which might be a different matching than the one stored in `_matching`.

You could also override method `has_perfect_matching` to always return `True`

> Also, shall the method `_subgraph_by_adding()` return an instance of `MatchingCoveredGraph` or an instance of `Graph`?

To be consistent, it should returned a matching covered graph. I suggest to do that in a dedicated PR.  Indeed, one should try that calls to methods internally using subgraphs are not failing when applied to matching covered graphs.... not an easy task.
 
> Could you please let me know why after doing `{INDEX_OF_METHODS}` and using `@doc_index()`, the methods are not appearing in the description of the html file?

Check how it's done in `src/sage/graphs/graph.py`. We also import `gen_thematic_rest_table_index` and the last line of the file is `__doc__ = __doc__.replace("{INDEX_OF_METHODS}", gen_thematic_rest_table_index(Graph, _additional_categories))`

janmenjayap:
Thank you so much for the comments.
I have already incorporated `_subgraph_by_adding()` in this PR.

janmenjayap:
Hi,
I have identified some more methods that are to be overwritten. The list goes as follows:

- `__hash__()`: To compute a hash (using canonical partition and set of edges).
- `_subgraph_by_deleting()`: It shall return a (matching covered) subgraph.
- `add_clique()`: The resulting graph shall be matching covered.
- `add_cycle()`: The resulting graph shall be matching covered.
- `add_ear()`: The resulting graph shall be matching covered.
- `add_path()`: The resulting graph shall be matching covered.
- `bisubdivide_edge()`: The resulting graph shall be matching covered.
- `bisubdivide_edges()`: The resulting graph shall be matching covered.
- `cartesian_product()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `clear()`: Unfortunately, $G$ can be an empty graph.
- `complement()`: The resulting graph shall be matching covered.
- `contract_edge()`: The resulting graph shall be matching covered.
- `contract_edges()`: The resulting graph shall be matching covered.
- `degree_constrained_subgraph()`
- `delete_multiedge()`: The resulting graph shall be matching covered.
- `disjoint_union()`: The resulting graph shall be matching covered.
- `disjunctive_product()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `has_loops()`: `False`.
- `is_biconnected()`: `True`.
- `is_block_graph()`: `True` if $G$ is $K_2$ `False` otherwise.
- `is_cograph()`: Need to overwrite this since `subgraph()` is overwritten.
- `is_forest()`: `True` if $G$ is $K_2$ `False` otherwise.
- `is_matching_covered()`: `True`.
- `is_path()`: `True` if $G$ is $K_2$ `False` otherwise.
- `is_subgraph()`: Need to cast `self` as `Graph` before calling this method.
- `is_tree()`: `True` if $G$ is $K_2$ False` otherwise.
- `join()`: The resulting graph shall be matching covered.
- `lexicographic_product()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `load_afile()`: To be figured out.
- `loop_edges()`: Return an empty list.
- `loop_vertices()`: Return an empty list.
- `merge_vertices()`: The resulting graph shall be matching covered.
- `number_of_loops()`: 0.
- `random_subgraph()`: The resulting graph shall be matching covered.
- `remove_loops()`: Just return.
- `save_afile()`: To be figured out.
- `strong_product()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `subdivide_edge()`: The resulting graph shall be matching covered.
- `subdivide_edges()`: The resulting graph shall be matching covered.
- `subgraph()`: The resulting graph shall be matching covered.
- `subgraph_search()`: The input graph shall be matching covered.
- `subgraph_search_count()`: The input graph shall be matching covered.
- `subgraph_search_iterator()`: The input graph shall be matching covered.
- `tensor_product()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `to_undirected()`: Return Graph(self).
- `transitive_closure()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `transitive_reduction()`: The resulting graph shall be matching covered/ need to cast `self` as `Graph` before calling.
- `union()`: The resulting graph shall be matching covered.

Any comment on this will be appreciated.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,1969 @@
...
+            sage: sorted(M) == sorted(G.get_matching())
+            True
+        """
+        return self._matching
+
+    @doc_index('Overwritten methods')
+    def has_perfect_matching(G, algorithm='Edmonds', solver=None, verbose=0,
+                             *, integrality_tolerance=1e-3):
+        r"""
+        Return whether the graph has a perfect matching.
```
dcoudert: it should be better to write `Check whether the graph has a perfect matching.`

dcoudert (changes requested):
Otherwise, LGTM.

dcoudert:
> Hi, I have identified some more methods that are to be overwritten. The list goes as follows:
> 
...
> * `bisubdivide_edge()`: The resulting graph shall be matching covered.
> * `bisubdivide_edges()`: The resulting graph shall be matching covered.

I don't know these methods. Where did you find them ?

... 
> Any comment on this will be appreciated.

This is a very long list and for some methods it's certainly not easy to adapt. Are all these methods listed in the todo list ?

janmenjayap:
That's true. `bisubdivide edge()` and `bisubdivide_edges()` are new methods, those are essentially subdividing an edge/ edges even number of times. For a matching covered graph, bisubdividing an edge/ edges always results in a new matching covered graph.

dcoudert (approved):
LGTM.

janmenjayap:
Thank you.

dcoudert:
for the first issue, it's best if you can add doctests to avoid the warnings.

For the second issue, I don't know how to fix that. Have you checked how it's done in other files ?

dcoudert:
I'm turning back this PR to needs work to let you some time to fix the issues.

dcoudert:
put it in the TESTS block.

dcoudert:
Let me know when you are satisfied with the current version.

dcoudert (commented):
do you plan to add the missing tests for codecov ?
L608, L623-624, L644, L649, L652, L1126

janmenjayap:
Yes.

dcoudert (approved):
LGTM.

janmenjayap:
Thank you.

PR #38748, opened 2024-10-01
deprecate dict method of polynomials

mantepse:
We deprecate the method `dict` for polynomials in favour of `monomial_coefficients`, because

* `monomial_coefficients` is more descriptive
* `monomial_coefficients` also works for free modules
* the function `dict` applied to a polynomial does (of course) something rather different, which may be surprising to newcomers.


mantepse:
@vbraun It is not clear to me why this was merged.  I don't think it was set to positive review.

PR #38754, opened 2024-10-02
Dropping support for power with Finite Field Polynomial Element

Shay2Shay:
Fixes issue : #32287
`Reference : pow ( Int, Finite Field)`
It was noticed that raising power with ModN field numbers and Gravio/NTL was not consistent.

- Finite Field = ModN : Coerced the FF element to Integer then performed Powering
- Finite Field = Gravio/NTL: Coerced the output to Gravio/NTL rather than Integer because it could only interpret non prime order field elements to be polynomials.

Change:
Removed support for Gravio/NTL/Pari_ffelt in powering scenario.

### :memo: Checklist

### DIFF
--- a/src/sage/structure/parent.pyx
+++ b/src/sage/structure/parent.pyx
@@ -2733,6 +2733,7 @@ cdef class Parent(sage.structure.category_object.CategoryObject):
             S_is_int = parent_is_integers(S)
             if not S_is_int:
                 from sage.rings.abc import IntegerModRing
+                from sage.rings.finite_rings.finite_field_base import FiniteField
                 if isinstance(S, IntegerModRing):
                     # We allow powering by an IntegerMod by treating it
                     # as an integer.
@@ -2741,10 +2742,7 @@ cdef class Parent(sage.structure.category_object.CategoryObject):
                     # to support. But in general this should not be
                     # allowed. See Issue #15709
                     S_is_int = True
-                from sage.rings.finite_rings.finite_field_pari_ffelt import FiniteField_pari_ffelt
-                from sage.rings.finite_rings.finite_field_ntl_gf2e import FiniteField_ntl_gf2e
-                from sage.rings.finite_rings.finite_field_givaro import FiniteField_givaro
-                if isinstance(S, (FiniteField_pari_ffelt, FiniteField_ntl_gf2e, FiniteField_givaro)):
+                elif isinstance(S, FiniteField):
                     # See Issue : #32287
                     # Disallowing Finite Field Polynomial Element
                     # to be used to raise power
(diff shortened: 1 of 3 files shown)


Shay2Shay:
@tscrim please have a look.

src/sage/structure/parent.pyx
```diff
@@ -2741,6 +2741,14 @@ cdef class Parent(sage.structure.category_object.CategoryObject):
                     # to support. But in general this should not be
                     # allowed. See Issue #15709
                     S_is_int = True
+                from sage.rings.finite_rings.finite_field_pari_ffelt import FiniteField_pari_ffelt
+                from sage.rings.finite_rings.finite_field_ntl_gf2e import FiniteField_ntl_gf2e
+                from sage.rings.finite_rings.finite_field_givaro import FiniteField_givaro
+                if isinstance(S, (FiniteField_pari_ffelt, FiniteField_ntl_gf2e, FiniteField_givaro)):
```
tscrim:
```suggestion
                elif isinstance(S, FiniteField):
```
You will also need to add
```
from sage.rings.finite_rings.finite_field_base import FiniteField
```

src/sage/structure/parent.pyx
```diff
@@ -2741,6 +2741,14 @@ cdef class Parent(sage.structure.category_object.CategoryObject):
...
+                    # See Issue : #32287
+                    # Disallowing Finite Field Polynomial Element
+                    # to be used to raise power
+                    raise TypeError('Unable to raise power using finite field polynomial element')
```
tscrim, line 2749:
```suggestion
                    _record_exception()
```
I think this is the correct thing to do.

tscrim (changes requested):
I don't think this is the correct way to fix the issue. Adding very special cases in generic code for deciding coercion should not be done without a very good reason. Contrast your code with the generic base class `IntegerModRing`. Hence, my proposed change, get every other case that is not an `IntegerModRing` but a finite field.

Shay2Shay:
I understood the 1st part `elif isinstance(S, FiniteField)` and I made desired changes.

But I couldn't understand the second part `_record_exception()`. I originally intended the function call to throw error and halt. 
But removing `raise error` will not stop the program and at it will coerce the base of power to be finite field element - this is what I am thinking.

tscrim:
Sorry, I forgot what `_record_exception` does; yes, that is not the proper thing to do there. Errors definitely should *not* be raised in that code. In turn, this made me realize something: specific coercions between specific parents should not be (dis)allowed by the coercion code. Those specific parents should be the ones to handle them. So I do not think this is the correct fix. We probably need to dig deeper into what coercions are being called and figure out a better mechanism for disallowing it.

Shay2Shay:
Understood

I will close this and look for another approach.

Shay2Shay:
@tscrim 
If I add [here] (https://github.com/Shay2Shay/sage/blob/b0f09d9d7827d657cacb12bd5f18933fa8ed8fb5/src/sage/rings/finite_rings/element_givaro.pyx#L1239C1-L1240C48)  (Same with other variants):
```
if isinstance(exp.parent(), FiniteField) and not isinstance(exp.parent(), ModNField):
    raise TypeError("Cannot raise power with polynomial")
```

How would that be?? (This works locally)

PR #38756, opened 2024-10-02
src/sage/symbolic/integration: make libgiac integration optional

orlitzky:
We turn the libgiac integrator into a no-op if libgiac isn't available. That way, in the default integration routines, it is essentially skipped. If it were tried last, we could literally skip it; but currently it lives between maxima and sympy and cannot be rearanged without messing up some doctests.

### DIFF
--- a/src/sage/symbolic/integration/external.py
+++ b/src/sage/symbolic/integration/external.py
@@ -285,7 +285,15 @@ def libgiac_integrator(expression, v, a=None, b=None):
         sage: (F.derivative(x) - f).simplify_trig()
         0
     """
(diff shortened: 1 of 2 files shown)


dimpase:
Shouldn't the test for giac presence go via src/sage/features/giac ?

orlitzky:
It's not using the "giac" executable, only the libgiac interface. I guess we could use the `sage.libs.giac` feature, but since this code is already going to `import sage.libs.giac`, it was simpler to catch the `ImportError`.

dimpase (approved):
LGTM

Issue #38410, opened 2024-07-23
Request for Built-in Function: Tutte Embedding of 3-Connected Planar Graphs


lichengzhang1:
### Problem Description

The Tutte embedding is a graph-drawing technique to position vertices of a graph so that the outer face is a convex polygon and each interior vertex is at the mean of the positions of its adjacent vertices.

**SageMath does not seem to have it built-in yet.**

### Proposed Solution

An implementation in sage can be found at [tuttedraw](https://github.com/manfredscheucher/tuttedraw) due to Manfred Scheucher. Here is the core code for it.

```
from scipy.spatial import ConvexHull
from sys import argv
from ast import literal_eval
import xml.etree.ElementTree as ET

from scipy import optimize
def tutte_layout(G,outer_face,weights):
	V = G.vertices()
	pos = dict()
	l = len(outer_face)

	a0 = pi/l+pi/2
	for i in range(l):
		ai = a0+pi*2*i/l
		pos[outer_face[i]] = (cos(ai),sin(ai))
	
	n = len(V)
	M = zero_matrix(RR,n,n)
	b = zero_matrix(RR,n,2)

	for i in range(n):
		v = V[i]
		if v in pos:
			M[i,i] = 1
			b[i,0] = pos[v][0]
			b[i,1] = pos[v][1]
		else:
			nv = G.neighbors(v)
			s = 0
			for u in nv:
				j = V.index(u)
				wu = weights[u,v]
				s += wu
				M[i,j] = -wu
			M[i,i] = s

	sol = M.pseudoinverse()*b
	return {V[i]:sol[i] for i in range(n)}




s = 'JsTB@_OBGN?'
G = Graph(s, sparse=True); 
F = G.faces()
F = [tuple(e[0] for e in f) for f in F]	
outer_face=F[0]
weights = dict()
for (u,v) in G.edges(labels=None):
     weights[u,v] = weights[v,u] = 1
G.set_pos(tutte_layout(G,outer_face,weights))
G.plot(edge_thickness=2, vertex_size=150)
```
![image](https://github.com/user-attachments/assets/78443987-16bb-4ae3-bf4e-74c544d13a02)

### Alternatives Considered

1. https://github.com/daryatodoskova/CSE306_TD9
2. https://github.com/DarlingZzzz/Tutte-Embedding

### Is there an existing issue for this?

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.

nturillo:
#38762 

[...]

PR #38762, opened 2024-10-03
Added a Tutte Embedding Layout for Graphs

nturillo:
Adds enhancement requested in #38410.

For 3-connected and planar graphs (https://en.wikipedia.org/wiki/Polyhedral_graph), there exists an embedding into
the plane called the Tutte embedding. In the Tutte embedding, an outer face is set, and all other vertices are placed inside
the outer face such that they are at the mean position of their neighbors.
(https://en.wikipedia.org/wiki/Tutte_embedding)

### :memo: Checklist


src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,94 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+
+        """
+        from sage.graphs.graph import Graph
+        from sage.matrix.constructor import zero_matrix
+        from sage.rings.real_mpfr import RR
+        
+        if (len(external_face) < 3):
+            raise ValueError("External face must have at least 3 vertices")
+
+        G = Graph(self)
```
dcoudert: Why do you need a copy of the graph?

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,94 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        if (not G.is_planar()):
+            raise ValueError("Graph must be planar")
+        C = G.subgraph(vertices=external_face)
+        if (not C.is_cycle()):
+            raise ValueError("External face must be a cycle")
+        external_face_ordered = C.depth_first_search(start=external_face[0])
+        
+        from sage.graphs.connectivity import vertex_connectivity
+        if (vertex_connectivity(G, k=4)):
```
dcoudert: you check if the vertex connectivity is more than 3. What if the graph is only 2 or 1 connected ? what if the graph is not connected ?

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,94 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        for i in range(n):
+            v = V[i]
+            if v in pos:
+                M[i,i] = 1
+                b[i,0] = pos[v][0]
+                b[i,1] = pos[v][1]
+            else:
+                nv = G.neighbors(v)
+                for u in nv:
+                    j = V.index(u)
```
dcoudert: a call to `V.index(u)` takes time $O(n)$. You should better first define a mapping from vertices to index in V like `vertex_to_int = {v: I for I, v in enumerate(V)}` and then use it. Each query will then be constant time.

dcoudert (changes requested):
The coding style will have to be improved, but let's start with the most important.

src/sage/graphs/generic_graph.py
nturillo: I was copying another layout function, but you're right I don't need this.

src/sage/graphs/generic_graph.py
nturillo: Whoops, I was getting my connectivity rules reversed. It should be fixed now. If the connectivity is more than three, that's actually okay; it's the reverse situation which induces a problem.

src/sage/graphs/generic_graph.py
nturillo: Okay good catch fixed!

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
         positions = dot2tex.dot2tex(self.graphviz_string(**options), format='positions', prog=prog)
 
         return {key_to_vertex[key]: pos for key, pos in positions.items()}
+    
+    def layout_tutte(self, external_face, external_face_pos=None, **options):
+        """
```
dcoudert: use `r"""` and remove the empty line

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+
+        Compute graph layout based on a Tutte embedding.
+
+        The graph must be 3-connected and planar.
```
dcoudert: Be more precise on the connectivity requirement. Is it exactly or at least 3 ?

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+
+        INPUT:
+
+        - ``external_face`` -- list; the external face to be made a polygon
```
dcoudert: isn't it possible to find an embedding without specifying the external face ?

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+
+        - ``external_face_pos`` -- list (default: ``None``); the positions of the vertices
```
dcoudert: in the code, you use it as a dictionary, not a list.

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+          of the external face. If ``None``, will automatically generate a unit sided regular polygon.
+        
+        - ``**options`` -- other parameters not used here
+
+        OUTPUT: a dictionary mapping vertices to positions
+
+        EXAMPLES::
+
+            sage: g = graphs.WheelGraph(n=7)
+            sage: g.layout_tutte(external_face=[1, 2, 3, 4, 5, 6])
```
dcoudert: This doctest is unstable. I obtain different results on my laptop (macOS...)
```
File "src/sage/graphs/generic_graph.py", line 21041, in sage.graphs.generic_graph.GenericGraph.layout_tutte
Failed example:
    g.layout_tutte(external_face=[1, 2, 3, 4, 5, 6])
Expected:
    {1: (-0.499999999999999, 0.866025403784438),
    2: (0.500000000000000, 0.866025403784439),
    3: (1.00000000000000, 8.85026869717109e-17),
    4: (0.500000000000000, -0.866025403784439),
    5: (-0.500000000000001, -0.866025403784438),
    6: (-1.00000000000000, 4.55896726715916e-16),
    0: (-5.55111512312578e-17, 1.86944233679563e-16)}
Got:
    {0: (4.16333634234434e-16, 7.42055745992141e-16),
     1: (-0.499999999999999, 0.866025403784439),
     2: (0.500000000000000, 0.866025403784440),
     3: (1.00000000000000, 7.31549942167514e-16),
     4: (0.500000000000000, -0.866025403784438),
     5: (-0.500000000000000, -0.866025403784437),
     6: (-1.00000000000000, 1.29124025055007e-15)}
```
You should find another test that is more stable. Printing vertex positions is never a good idea.

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            {1: (-0.499999999999999, 0.866025403784438),
+            2: (0.500000000000000, 0.866025403784439),
+            3: (1.00000000000000, 8.85026869717109e-17),
+            4: (0.500000000000000, -0.866025403784439),
+            5: (-0.500000000000001, -0.866025403784438),
+            6: (-1.00000000000000, 4.55896726715916e-16),
+            0: (-5.55111512312578e-17, 1.86944233679563e-16)}
+            sage: g = graphs.CubeGraph(n=3, embedding=2)
+            sage: g.plot(layout='tutte', external_face=['101','111','001','011'], external_face_pos={'101':(1,0), '111':(0,0), '001':(2,1), '011':(-1,1)})
```
dcoudert: this test is failing on my laptop
```
File "src/sage/graphs/generic_graph.py", line 21050, in sage.graphs.generic_graph.GenericGraph.layout_tutte
Failed example:
    g.plot(layout='tutte', external_face=['101','111','001','011'], external_face_pos={'101':(1,0), '111':(0,0), '001':(2,1), '011':(-1,1)})
Expected:
    Launched png viewer for Graphics object consisting of 21 graphics primitives
Got:
    Graphics object consisting of 21 graphics primitives
```

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            Launched png viewer for Graphics object consisting of 21 graphics primitives
+
```
dcoudert: remove empty line.

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        Compute graph layout based on a Tutte embedding.
+
+        The graph must be 3-connected and planar.
+
+        INPUT:
+
+        - ``external_face`` -- list; the external face to be made a polygon
+
+        - ``external_face_pos`` -- list (default: ``None``); the positions of the vertices 
+          of the external face. If ``None``, will automatically generate a unit sided regular polygon.
```
dcoudert: this line requires 2 more spaces indentation. See other methods for examples.

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            4: (0.500000000000000, -0.866025403784439),
+            5: (-0.500000000000001, -0.866025403784438),
+            6: (-1.00000000000000, 4.55896726715916e-16),
+            0: (-5.55111512312578e-17, 1.86944233679563e-16)}
+            sage: g = graphs.CubeGraph(n=3, embedding=2)
+            sage: g.plot(layout='tutte', external_face=['101','111','001','011'], external_face_pos={'101':(1,0), '111':(0,0), '001':(2,1), '011':(-1,1)})
+            Launched png viewer for Graphics object consisting of 21 graphics primitives
+
+        """
+        from sage.graphs.graph import Graph
```
dcoudert: you don't need to import `Graph`

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        from sage.matrix.constructor import zero_matrix
+        from sage.rings.real_mpfr import RR
+        
+        if (len(external_face) < 3):
```
dcoudert: for tests, there is no need for the extra parenthesis: `if len(external_face) <3:`

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        from sage.graphs.connectivity import vertex_connectivity
+        if (not vertex_connectivity(self, k=3)):
+            raise ValueError("Graph must be 3-connected")
+
+        from math import sin, cos, pi
+        pos = dict()
+
+        if external_face_pos is None:
+            l = len(external_face)
+            a0 = pi/l+pi/2
```
dcoudert: pycodestyle: `a0 = pi/l + pi/2`

also, it might be better to avoid variable `l` that is often confused with `1`.

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            for i, vertex in enumerate(external_face_ordered):
+                ai = a0+pi*2*i/l
+                pos[vertex] = (cos(ai),sin(ai))
```
dcoudert: add a space after `,`, so `pos[vertex] = (cos(ai), sin(ai))`.

Same in other places

src/sage/graphs/generic_graph.py
```diff
@@ -21014,6 +21016,95 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        else:
+            for v,p in external_face_pos.items():
+                pos[v] = p
+
+        V = self.vertices()
+        n = len(V)
+        M = zero_matrix(RR,n,n)
+        b = zero_matrix(RR,n,2)
+
+        vertices_to_indices = {v:i for i,v in enumerate(V)}
```
dcoudert: `vertices_to_indices = {v: i for I, v in enumerate(V)}`

src/sage/graphs/generic_graph.py
```diff
@@ -21399,6 +21493,13 @@ def plot(self, **options):
           "down".  If "up" (resp., "down"), then the root of the tree will
           appear on the bottom (resp., top) and the tree will grow upwards
           (resp. downwards). Ignored unless ``layout='tree'``.
+        
+        - ``external_face`` -- list of vertices; the external face to be made a
+        in the Tutte layout. Ignored unless ``layout='tutte''``.
```
dcoudert: 2 more spaces indentation, like other items.

dcoudert (changes requested):
* Is the proposed method working on directed graphs ? if not, it should be in `graph.py` and not in `generic_graph.py`
* you should add some tests showing the behavior of the method when the connectivity requirement is not fulfilled, or when the graph is not planar
* For all comments, please use 80 columns mode (lines of length at most 80).

src/sage/graphs/generic_graph.py
nturillo: Any graph that is k-connected for k $\geq$ 3 is 3-connected. For example, a 5-connected graph is 3-connected.

src/sage/graphs/generic_graph.py
nturillo: In order to calculate the embedding, an external face must be specified. Any subgraph of the graph that is a cycle will work as an external face. It would be possible to find an arbitrary external face automatically if none is specified, but it's unclear to me how one would be chosen from the available options. Also, that functionality is outside the scope of my original intention.

src/sage/graphs/generic_graph.py
nturillo: In many of the other layout examples, they include this sort of example. However, I've removed it from this function.

src/sage/graphs/generic_graph.py
nturillo: I believe this is a quirk of WSL, it should be fixed now.

src/sage/graphs/generic_graph.py
nturillo: In the other methods, it seems like the inputs have two spaces of indentation for an input element which requires more than a line of explanation. Is that not correct?

nturillo:
The function is intended to work on digraphs as well. Since it's an embedding, it just embeds the digraph as if it were a simple graph. I've added more examples for when one of the requirements is not met. 

Can you be more specific about the 80 column mode? For what comments? These on GitHub or in the actual code? Sorry, this is my first time contributing.

src/sage/graphs/generic_graph.py
dcoudert: Consider tha following:
```py
H = Graph(self) # take a (undirected) copy H of the graph
u, v = next(H.edge_iterator(labels=False)  # take any edge (u, v) of H
H.delete_edge(u, v)  # remove edge (u, v) from H
C = H.shortest_path(v, u) # Compute a shortest path from v to u in H minus (u, v)
```
You get a short cycle `C`, and this cycle is a face.

This can of course be done in a follow up PR, but this seems quite simple.

dcoudert:
> Can you be more specific about the 80 column mode? For what comments? These on GitHub or in the actual code? Sorry, this is my first time contributing.

In the code. As you can see, in most of the method we try to avoid long lines. Comments and documentation are in 80 columns mode. We also try to avoid long lines in the code to make it easier to read.

fchapoton:
the `Lint` checker is not happy, please fix the warnings seen there

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            ...
+            ValueError: Graph must be 3-connected
+        """
+        from sage.matrix.constructor import zero_matrix
+        from sage.rings.real_mpfr import RR
+
+        if len(external_face) < 3:
+            raise ValueError("External face must have at least 3 vertices")
+
+        if (not self.is_planar()):
```
dcoudert, line 21686: there is no need for external parenthesis

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            sage: g.layout(layout='tutte', external_face=[0,1,2,3,4,5,6,7,8,9])
+            Traceback (most recent call last):
+            ...
+            ValueError: Graph must be 3-connected
+        """
+        from sage.matrix.constructor import zero_matrix
+        from sage.rings.real_mpfr import RR
+
+        if len(external_face) < 3:
+            raise ValueError("External face must have at least 3 vertices")
```
dcoudert, line 21684: the convention for error messages is to start with a lowercase letter.

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+
+        if (not self.is_planar()):
+            raise ValueError("Graph must be planar")
+
+        C = self.subgraph(vertices=external_face)
+        if (not C.is_cycle(directed_cycle=False)):
+            raise ValueError("External face must be a cycle")
+        external_face_ordered = C.depth_first_search(start=external_face[0], ignore_direction=False)
+
+        from sage.graphs.connectivity import vertex_connectivity
```
dcoudert, line 21694: you don't need to import `vertex_connectivity`. You can directly call `self.vertex_connectivity(k=3)`

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        if (not vertex_connectivity(self, k=3)):
+            raise ValueError("Graph must be 3-connected")
+
+        from math import sin, cos, pi
+        pos = dict()
+
+        if external_face_pos is None:
+            external_face_length = len(external_face)
```
dcoudert, line 21702: you could use the helper method `self._circle_embedding(...)`.

The call should be like `pos = G._circle_embedding(external_face, return_dict=True)`

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            for v, p in external_face_pos.items():
+                pos[v] = p
+
+        V = self.vertices()
+        n = len(V)
+        M = zero_matrix(RR, n, n)
+        b = zero_matrix(RR, n, 2)
+
+        vertices_to_indices = {v:I for I, v in enumerate(V)}
+        for i in range(n):
```
dcoudert, line 21717: why not using `for i, v in enumerate(self):` ?

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        else:
+            for v, p in external_face_pos.items():
+                pos[v] = p
+
+        V = self.vertices()
+        n = len(V)
+        M = zero_matrix(RR, n, n)
+        b = zero_matrix(RR, n, 2)
+
+        vertices_to_indices = {v:I for I, v in enumerate(V)}
```
dcoudert, line 21716: a minor detail, add a space after `:`, so
`vertices_to_indices = {v: I for I, v in enumerate(V)}`

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,97 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+                b[i, 1] = pos[v][1]
+            else:
+                nv = self.neighbors(v)
+                for u in nv:
+                    j = vertices_to_indices[u]
+                    M[i, j] = -1
+                M[i, i] = len(nv)
+
+        sol = M.pseudoinverse()*b
+        return {V[i]:sol[i] for i in range(n)}
```
dcoudert, line 21731: could be `return dict(zip(V, sol))` or even `return dict(zip(self, sol))`

dcoudert (changes requested):
The code has been greatly improved, but some more can be done.

src/sage/graphs/generic_graph.py
```diff
@@ -21637,6 +21639,98 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        - ``**options`` -- other parameters not used here
+
+        OUTPUT: a dictionary mapping vertices to positions
+
+        EXAMPLES::
+            sage: g = graphs.WheelGraph(n=7)
+            sage: g.plot(layout='tutte', external_face=[0,1,2])                        # needs sage.plot
+            Graphics object consisting of 20 graphics primitives
+            sage: g = graphs.CubeGraph(n=3, embedding=2)
+            sage: g.plot(layout='tutte', external_face=['101','111','001',
```
dcoudert, line 21666: use the form
```sage:
            sage: g.plot(layout='tutte', external_face=['101','111','001',                                                    # needs sage.plot
            ....: '011'], external_face_pos={'101':(1,0), '111':(0,0), '001':(2,1),
            ....: '011':(-1,1)})
```

dcoudert (changes requested):
several failing doctests

src/sage/graphs/generic_graph.py
nturillo: Fixed!

src/sage/graphs/generic_graph.py
```diff
@@ -21641,6 +21643,98 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+        - ``external_face_pos`` -- dictionary (default: ``None``); the positions
+          of the vertices of the external face. If ``None``, will automatically
+          generate a unit sided regular polygon.
+
+        - ``**options`` -- other parameters not used here
+
+        OUTPUT: a dictionary mapping vertices to positions
+
+        EXAMPLES::
+            sage: g = graphs.WheelGraph(n=7)
```
dcoudert, line 21666: please add an empty line after `EXAMPLES::`, as done in other methods.

src/sage/graphs/generic_graph.py
```diff
@@ -21641,6 +21643,98 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            raise ValueError("external face must have at least 3 vertices")
+
+        if not self.is_planar():
+            raise ValueError("graph must be planar")
+
+        if not self.vertex_connectivity(k=3):
+            raise ValueError("graph must be 3-connected")
+
+        if external_face is None:
+            from sage.graphs.graph import Graph
```
dcoudert, line 21698: Could you make a global comment for this case explaining what this is about instead of adding a comment per line.
For example:
```py
if external_face is None:
    # Search for a short cycle and make it the external face
    from sage.graphs.graph import Graph
    ...
```

src/sage/graphs/generic_graph.py
```diff
@@ -21641,6 +21643,98 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+            H = Graph(self)  # take a (undirected) copy H of the graph
+            u, v = next(H.edge_iterator(labels=False))  # take any edge (u, v) of H
+            H.delete_edge(u, v)  # remove edge (u, v) from H
+            external_face = H.shortest_path(v, u)
+            # Compute a shortest path from v to u in H minus (u, v)H = G.
+            # Cycle existence is guaranteed since G is 3-connected.
+        else:
+            C = self.subgraph(vertices=external_face)
+            if (not C.is_cycle(directed_cycle=False)):
```
dcoudert, line 21707: This case may fail for instance when the graph is a clique of order $\geq 4$ and the given external face any subset with $\geq 4$ vertices.  More generally, it will fail if the external face has a chord.

I think we must simply check that the vertices of the external face form a cycle in the graph using
```sage
if (not self.has_edge(external_face, external_face[-1])
     or not all(self.has_edge(u, v) for u, v in zip(external_face, external_face[1:]))):
    raise ValueError("external face must be a cycle")
```
this is for undirected graphs. The case of digraphs should also be considered.

src/sage/graphs/generic_graph.py
```diff
@@ -21641,6 +21643,98 @@ def layout_graphviz(self, dim=2, prog='dot', **options):
...
+                raise ValueError("external face must be a cycle")
+            external_face = C.depth_first_search(start=external_face[0], ignore_direction=False)
+
+        pos = dict()
+        if external_face_pos is None:
+            pos = self._circle_embedding(external_face, return_dict=True)
+        else:
+            for v, p in external_face_pos.items():
```
dcoudert, line 21715: why not simply `pos = external_face_pos.copy()`  ? and the line `pos = dict()` is no longer needed.

src/sage/graphs/generic_graph.py
```diff
@@ -22026,6 +22123,13 @@ def plot(self, **options):
           appear on the bottom (resp., top) and the tree will grow upwards
           (resp. downwards). Ignored unless ``layout='tree'``.
 
+        - ``external_face`` -- list of vertices; the external face to be made a
```
dcoudert, line 22126: please add `(default: ``None``)`

dcoudert (changes requested):
some more comments.

src/sage/graphs/generic_graph.py
nturillo: Done!

src/sage/graphs/generic_graph.py
nturillo: This is my mistake; the external face should actually be a face of the graph in the graph's planar embedding. Checking for a cycle is not sufficient. The planar embedding is unique for a three-connected planar graph (due to Whitney 1933), so the code should check if the proposed external face is actually a face in the embedding created by is_planar. I believe all of this should be the same for digraphs. Fixed now!

src/sage/graphs/generic_graph.py
nturillo: Done!

src/sage/graphs/generic_graph.py
nturillo: Done!

dcoudert (approved):
This seems ok now. Thanks.

PR #38703, opened 2024-09-23
Implementation of Ore modules

xcaruso:
This PR implements modules over Ore polynomial rings.

More precisely, if $A[X;\theta,\partial]$ is a Ore polynomial ring, we propose an implementation of finite free modules $M$ over $A$ equipped with a map $f : M \to M$ such that $f(ax) = \theta(a) f(x) + \partial(a) x$ for all $a \in R$ and $x \in M$.
Such a map is called *pseudolinear* and it endows `M` with a structure of module over $A[X;\theta,\partial]$ (the map $f$ corresponding to the multiplication by $X$).

This PR includes:
- an implementation of the category of Ore modules
- an implementation of Ore modules, their submodules and their quotients (with an option to give chosen names to elements in a distinguished basis)
- a constructor to create quotients of the form $A[X;\theta,\partial] / A[X;\theta,\partial]P$
- an implementation of morphisms between Ore modules, including methods for computing kernels, cokernels, images and coimages

This is the second step (after PR #38650) towards the implemetation of Anderson motives.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [ ] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

### :hourglass: Dependencies

#38650: pseudomorphisms

PR #38767, opened 2024-10-04
polynomials/fix category

mantepse:
We put polynomials (and friends) into the category of algebras with basis.

Dependencies: #38729


tscrim:
@nbruin @fchapoton @xcaruso cc-ing you in case you have some thoughts/comments about doing this.

tscrim (approved):
Since there are no comments, then positive review. (Something I've wanted to do for awhile, thank you!)

xcaruso:
Thanks @tscrim for pointing this to me and sorry for not being reactive.
I actually do not have a strong opinion on this and I'm definitely happy with your positive review.

However, I should say that I'm quite confused with the dichotomy between the classes `FreeModule` and `CombinatorialFreeModule`. I think I understand that they support different kind of features: basically the former is more oriented towards linear algebra in finite dimension, while the former allows for infinite dimensional spaces, more abstract constructions (as tensor products, *etc*.) and custom names for elements in a basis. But, IMO, it's really shame because we cannot having both at the same time is 

Concretely, recently, I implemented Ore modules (see #38703) and was embarrassed by this. I finally chose to derive from `FreeModule` because I was interested in having submodules and quotients, two constructions which are more common in the class `FreeModule`. However, I was not entirely convinced.

Any thoughts about this? (Please answer in #38703 - and review this PR at the same time :-))

Issue #35790, opened 2023-06-17
Doctest `# optional` / `# long` / `# needs` annotation maintenance tool


mkoeppe:
`sage -fixoptional` would rewrite doctest annotations `# optional` / `# long` / `# needs` (from #35779):
- use `# optional` only for optional features, `# needs` for standard features (infrastructure for standard status of features is added in #35668)
- rewrite to use persistent (codeblock-scoped) annotations (from #35779) when it reduces noise
- align at alignment points (>= column 88 for `# needs`; see https://gist.github.com/mkoeppe/d7c6224d780dbb9641a03c7ce4357e9e)

(from https://groups.google.com/g/sage-devel/c/nUi0hn_u8SM)

PR #38770, opened 2024-10-05
Add "needs" tags for giac and libgiac

orlitzky:
Part of https://github.com/sagemath/sage/issues/38668. If it's going to be possible to disable giac, we need to guard all of the tests that use it with either `# needs giac` or `# needs sage.libs.giac`.

I think I've gotten them all. A crude way to test:

1. `git rm -r src/sage/libs/giac` and rebuild to disable sage.libs.giac
2. build sage, and then delete the giac executable to disable the pexpect interface

If you do these one at a time, it should ensure that the correct tags are used. (Typically, if giac is missing, neither sage.libs.giac nor the giac executable will be present, making it very easy to mix up the tags.)

For bonus points you can undelete `src/sage/libs/giac` after building but before testing to make sure the "needs" tags in those files are accurate.

### Dependencies:

* https://github.com/sagemath/sage/pull/38756
* https://github.com/sagemath/sage/pull/38686 (not strictly required, but it adds a few "needs sage.libs.giac" tags of its own)

### DIFF
--- a/src/doc/en/prep/Calculus.rst
+++ b/src/doc/en/prep/Calculus.rst
@@ -316,8 +316,9 @@ help it look nicer in the browser?
     - 1/10*(sqrt(5) - 3)*log(2*x^2 + x*(sqrt(5) - 1) + 2)/(sqrt(5) - 1)
     + 1/5*log(x + 1)
 
-Some integrals are a little tricky, of course. Sage tries hard to integrate using Maxima, Giac and Sympy::
+Some integrals are a little tricky, of course. Sage tries hard to integrate using Maxima, Sympy, and (optionally, if installed) Giac. The following can only be integrated by Giac::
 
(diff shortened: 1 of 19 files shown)


mantepse:
shouldn't the tag be `# optional - giac`?

orlitzky:
I think either will work, but IIRC "needs" is preferred for standard features, and `build/pkgs/giac/type` is still standard.  I tried to find some documentation for this but the description of https://github.com/sagemath/sage/issues/35790 was the best I could do.

orlitzky:
The CI fail is,

> sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/rings/tests.py  # Timed out

and doesn't have anything to do with this PR (which just adds some presently-unused tags).

tobiasdiez (approved):

PR #38774, opened 2024-10-05
rpy2 (standard) cannot depend on r (optional)

### LINKED ISSUE #38773: make sdist is broken
### Steps To Reproduce

```
make download-for-sdist
[...]
[r-none] Error: r is a dummy package and 
[r-none] cannot be installed using the Sage distribution.
make[2]: *** [Makefile:3575: r-SAGE_LOCAL-no-deps] Error 1
make[1]: *** [Makefile:3575: /home/release/Sage/local/var/lib/sage/installed/r-none] Error 2
```

### Expected Behavior

Should be able to build a source tarball

### Actual Behavior

Broken

### Additional Information

_No response_

### Environment

10.5.beta6

### Checklist

vbraun:
"make download-for-sdist" tries to download all standard sources and fails if we depend on a dummy package. Presumably fine since the r dummy package should only used at configure time.

Fixes #38773

### DIFF
--- a/build/pkgs/rpy2/dependencies
+++ b/build/pkgs/rpy2/dependencies
@@ -1,4 +1,4 @@
- r cffi tzlocal pytz jinja2 | $(PYTHON_TOOLCHAIN) pycparser $(PYTHON)
+ cffi tzlocal pytz jinja2 | $(PYTHON_TOOLCHAIN) pycparser $(PYTHON)
 
 ----------
 All lines of this file are ignored except the first.


kwankyu:
Removing r here makes `./configure` not check for the presence of r for rpy2...

kwankyu:
#38777 fixes the problem of this PR.

kwankyu (approved):
LGTM.

PR #38776, opened 2024-10-06
Fix test failures due to global mpmath state

orlitzky:
mpmath has a global `mp.pretty` variable that affects how its numbers are printed. On my machine, it's not reliable. For example:

```
Failed example:
    repr(mp.mpc(2,3))
Expected:
    '(2.0 + 3.0j)'
Got:
    "mpc(real='2.0', imag='3.0')"
```

I've only fixed the tests that are failing for me, since mpmath-1.4 is on the TODO list. A few different strategies are used:

* `print()` will use the pretty format if all we're doing is displaying an answer.
* `with workdps` can be used to change the precision only locally.
* In TESTS blocks, the ugly output is acceptable.
* When actually testing `mp.pretty`, we can first clone the `mp` object with `mp2 = mp.clone()`, and then work on the new one.

### DIFF
--- a/src/sage/functions/orthogonal_polys.py
+++ b/src/sage/functions/orthogonal_polys.py
@@ -2343,10 +2343,10 @@ class Func_ultraspherical(GinacFunction):
         sage: # needs mpmath
         sage: from mpmath import gegenbauer as gegenbauer_mp
         sage: from mpmath import mp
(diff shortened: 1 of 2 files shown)


dimpase (approved):
lgtm

orlitzky:
Thanks, I may finally be able to focus on whatever is happening with GAP soon.

PR #38777, opened 2024-10-06
Check the system R in SPKG rpy2

kwankyu:
It is problematic to put a dummy package into the dependencies list. For instance, see #38774.

We move the check for the system R to `spkg-configure.m4` of SPKG rpy2. Then we can replace R in the dependencies list with rpy2.

### :memo: Checklist

### :hourglass: Dependencies

#38774 (merged here)


kwankyu:
We can ignore the failure of "test-new" workflow. It failed because it forced installing r_jupyter, which depends on the syetem R, which is of course not existing.

kwankyu:
Thanks!

Issue #38758, opened 2024-10-03
harmonise the interfaces of orientation() and acyclic_orientation()


dimpase:
### Steps To Reproduce

for a graph `g`, outputs of `g.orientation()` and `g.acyclic_orientation()` are rather different, type-wise.
As reported by Gordon Royle on [sage-devel](https://groups.google.com/d/msgid/sage-support/f2d490eb-f11d-4fb3-8c2d-80e4802fd7b0n%40googlegroups.com)

My suggestion is to redefine orientations of a graph $\Gamma=(V,E)$ to be a function $E\to V\times V$,
and return these, eventually making this the default.
 

### Expected Behavior

should be consistent

### Actual Behavior

is not

### Additional Information

see also https://ask.sagemath.org/question/79427/the-acyclic-orientations-function-behaves-unexpectedly/

### Environment

- **OS**: all
- **Sage Version**: 10.*

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

dimpase:
@dcoudert - do you think it's a good idea?

maxale:
Then there also should be an easy way to apply an orientation function to the graph and obtain the corresponding directed graph. Something like `D = G.orient(f)`.

maxale:
Please also include `.random_orientation()` into the proposed harmonization.

dcoudert:
The current proposal is to make `acyclic_orientations` return a digraph with properly directed edges #38757

What should be done in `.random_orientation()` ?
We could also consider `strong_orientations_iterator` (which could be renamed `strong_orientations`).

Do we really need a method like `D = G.orient(f)` ? Roughly, it should be `D = DiGraph([G, [f(e) for e G.edges()]])`.

maxale:
> What should be done in `.random_orientation()` ?
> We could also consider `strong_orientations_iterator` (which could be renamed `strong_orientations`).

I do not know if anything needs to be done. My point is that a harmonization/uniformization of the interfaces should be applied to all functions that produce graph orientations. The `strong_orientations` should be included into that class as well.

> Do we really need a method like `D = G.orient(f)` ? Roughly, it should be `D = DiGraph([G, [f(e) for e G.edges()]])`.

Yes, assuming that `f` understands labeled edges and preserves the labels. Here my point is that if there is a function, then there should be a way to apply this function without thinking about underlying technicalities.

dcoudert:
[...]

PR #38778, opened 2024-10-06
add method `orient` to `Graph`

dcoudert:
As suggested by @maxale in https://github.com/sagemath/sage/issues/38758#issuecomment-2391505302, we add a method to apply an orientation function to a graph and obtain the corresponding directed graph. 

### :memo: Checklist

### DIFF
--- a/src/sage/graphs/graph.py
+++ b/src/sage/graphs/graph.py
@@ -9596,7 +9596,10 @@ class Graph(GenericGraph):
     from sage.graphs.tutte_polynomial import tutte_polynomial
     from sage.graphs.lovasz_theta import lovasz_theta
     from sage.graphs.partial_cube import is_partial_cube
-    from sage.graphs.orientations import strong_orientations_iterator, random_orientation, acyclic_orientations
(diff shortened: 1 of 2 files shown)


tscrim (commented):
Lines 172 to 175 are never reached since they are an "else" after two cases, one of which *must* be true. Is line 172 supposed to be an `if`?

dcoudert:
These lines can be reached setting `data_structure=None`, `immutable=None` and `sparse=True` or `False`.

In fact, lines 154-182 are copy-paste from method `copy` in `generic_graph.py`.  May be I should make a method like `_choose_data_structure_for_copy` in `generic_graph.py` and use it for both methods ?

tscrim:
Ah, I see. It’s fine for now. Then can you add doctests covering those cases?

dcoudert:
We'll see if the new tests are sufficient.

tscrim (approved):
That should be good. Thank you.

dcoudert:
Thanks for the review.

PR #38795, opened 2024-10-09
switch to nauty for generating posets

fchapoton:
as there is now a direct method in `nauty` and we have added the interface previously

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/posets/posets.py
+++ b/src/sage/combinat/posets/posets.py
@@ -8945,8 +8945,7 @@ class FinitePosets_n(UniqueRepresentation, Parent):
             sage: list(P)
             [Finite poset containing 2 elements, Finite poset containing 2 elements]
         """
-        from sage.graphs.digraph_generators import DiGraphGenerators
-        for dig in DiGraphGenerators().nauty_posetg(f"{self._n} o"):
+        for dig in digraphs.nauty_posetg(f"{self._n} o"):
             # We need to relabel the digraph since range(self._n) must be a linear
             # extension. Too bad we need to compute this again. TODO: Fix this.
             label_dict = dict(zip(dig.topological_sort(), range(dig.order())))
--- a/src/sage/combinat/posets/posets.py
+++ b/src/sage/combinat/posets/posets.py
@@ -8884,11 +8884,11 @@ class FinitePosets_n(UniqueRepresentation, Parent):
         sage: P.cardinality()
         5
         sage: for p in P: print(p.cover_relations())
-        []
-        [[1, 2]]
-        [[0, 1], [0, 2]]
-        [[0, 1], [1, 2]]
         [[1, 2], [0, 2]]
+        [[0, 1], [0, 2]]
+        [[0, 2]]
+        [[0, 1], [1, 2]]
+        []
     """
 
     def __init__(self, n) -> None:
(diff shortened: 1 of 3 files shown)


dcoudert (approved):
LGTM.

src/sage/combinat/posets/posets.py
```diff
@@ -8945,8 +8945,8 @@ def __iter__(self):
             sage: list(P)
             [Finite poset containing 2 elements, Finite poset containing 2 elements]
         """
-        from sage.graphs.digraph_generators import DiGraphGenerators
-        for dig in DiGraphGenerators()(self._n, is_poset):
+        from sage.graphs.digraph_generators import DiGraphGenerators as DG
+        for dig in DG().nauty_posetg(f"{self._n} o"):
```
mantepse: I would prefer not to use `as DG`, because it makes code much harder to read for me personally.  Also, the import could actually go to the module level, no?

src/sage/combinat/tutorial.py
```diff
@@ -844,15 +844,15 @@
 
 Partial orders on a set of `8` elements, up to isomorphism::
 
-    sage: C = Posets(8); C
-    Posets containing 8 elements
+    sage: C = Posets(7); C
+    Posets containing 7 elements
     sage: C.cardinality()
-    16999
+    2045
```
mantepse, line 850: Why this change?  On my computer, even *generating* all the posets now takes only half a second.

mantepse:
This is a very impressive speedup:
```
sage: %time n=7; sum(1 for dig in DiGraphGenerators()(n, is_poset))
CPU times: user 5.85 s, sys: 0 ns, total: 5.85 s
Wall time: 5.85 s
2045
sage: %time n=7; sum(1 for dig in DiGraphGenerators().nauty_posetg(f"{n} o"))
CPU times: user 57.3 ms, sys: 0 ns, total: 57.3 ms
Wall time: 66.7 ms
2045
sage: %time n=8; sum(1 for dig in DiGraphGenerators().nauty_posetg(f"{n} o"))
CPU times: user 395 ms, sys: 0 ns, total: 395 ms
Wall time: 399 ms
16999
```

dcoudert:
Actually, there is another issue: `nauty_posetg` works only for $n\leq 16$. You could use a switch to use it only in this case and use the previous (slower) method otherwise.

src/sage/combinat/tutorial.py
fchapoton: well, I think smaller doctests are always better. No need to show muscles in doctests.

mantepse:
> Actually, there is another issue: `nauty_posetg` works only for n ≤ 16 . You could use a switch to use it only in this case and use the previous (slower) method otherwise.

Great catch!

To avoid such oversights in future (not in this pull request!), wouldn't it be better to have intermediate methods that call `nauty_posetg` and do the argument parsing?  I am frequently wasting time looking up the precise string for `nauty_geng` to generate connected graphs only :-)

src/sage/combinat/tutorial.py
mantepse: OK, I don't mind - although I somewhat disagree, with two (implicit) statements:

* in the case at hand, we are only looking up the number
* in general, I think it it good to use the largest example which takes a reasonable amount of time (in particular, less than a second), because it seems to me that this helps to catch bad performance regressions.

mantepse:
I think that the issue pointed out by @dcoudert still needs fixing.

fchapoton:
We lose a lot of speed in the relabeling..

fchapoton:
merci pour les suggestions

dcoudert (approved):
LGTM.

Issue #38784, opened 2024-10-07
Minor issue with documentation for congruence subgroups


seewoo5:
### Steps To Reproduce

In the [documentation](https://doc.sagemath.org/html/en/reference/arithgroup/sage/modular/arithgroup/farey_symbol.html#sage.modular.arithgroup.farey_symbol.Farey.word_problem) of `Farey.word_problem`, the documentation does not show up correctly:

<img width="676" alt="Screenshot 2024-10-07 at 9 44 17 AM" src="https://github.com/user-attachments/assets/934fd426-b5e1-44ef-bf58-eded647de1e2">

Possibly an issue with this part:

https://github.com/sagemath/sage/blob/b0f09d9d7827d657cacb12bd5f18933fa8ed8fb5/src/sage/modular/arithgroup/farey_symbol.pyx#L364

### Expected Behavior

Sentence shows up correctly.

### Actual Behavior

Setence shows up weirdly.

### Environment

- **OS**: 
- **Sage Version**: 

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

mantepse:
Yes, it should be a backtick, not a quote after the `i`.

mantepse:
I just built the doc on my computer. Here, the problem does not show up.  However, apart from the suggestion above, I just see that the docstring has another English problem:
```
        - ``gens`` returns tuple of tuples of the form `(g,n)`,
          `(g,n)` such that the product of the matrices `g^n`
          equals `M` up to sign.
```
should probably be
```
        - ``gens`` returns a tuple of pairs `(g, n)`, where `g` is a matrix and `n` an integer, such that the product of the matrices `g^n` equals `M` up to sign.
```
Below that, there is whitespace around keyword `=`, which should be removed.

Are you going to do a pull request?

seewoo5:
@mantepse Maybe not - feel free to take PR.

PR #38797, opened 2024-10-10
make sure .division_points() returns the correct result for m=-1

### LINKED ISSUE #38796: elliptic-curve points' .division_points() method returns wrong result for m=-1
```sage
sage: E = EllipticCurve(GF(127), [1,1])
sage: P = E(72, 24)
sage: [-1*Q for Q in P.division_points(-1)]
[(72 : 103 : 1)]
```
In this example, the returned value is the negative of what it should be.

yyyyx4:
Fixes #38796: The solution is to simply take the sign of $m$ into account when $m \in \{\pm1\}$.

### DIFF
--- a/src/sage/schemes/elliptic_curves/ell_point.py
+++ b/src/sage/schemes/elliptic_curves/ell_point.py
@@ -1148,12 +1148,19 @@ class EllipticCurvePoint_field(EllipticCurvePoint,
           3
           sage: [(Q,Q._order) for Q in P.division_points(4)]
           [((-2 : -7 : 1), 6), ((1 : 2 : 1), 6), ((4 : -7 : 1), 3), ((13 : 38 : 1), 6)]
+
+        Check for :issue:`38796`::
+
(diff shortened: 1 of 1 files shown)


DaveWitteMorris:
Thanks for fixing the bug!

Please add "Fixes #38796" to the PR description (or just change the word "See" to "Fixes"), so the linked issue will close when this is merged.

yyyyx4:
That's in the commit message, which I think should work as well?

DaveWitteMorris:
I think you are probably right that the commit message will work (though I didn't realize that before). However, the PR does not show up in the "Development" sidebar of the issue. If you look at #38784 (for example), the sidebar says there is a pull request that may close the issue, and I think it is helpful for this information to be displayed clearly, so, although putting the reference in the commit message may work, I don't think it works literally "as well" as putting it in the PR description. As an alternative, I think you may be able to manually link the PR to the issue by clicking on the PR's Development sidebar.

yyyyx4:
I see, thanks for your explanation. I changed it.

PR #38808, opened 2024-10-12
faster comparison of elliptic-curve morphisms

yyyyx4:
Here we add two fast checks to the generic comparison method for elliptic-curve morphisms that will quickly detect some pairs of unequal morphisms. In the context of #35949, this speeds up the following example from 5.9 seconds to 1.8 seconds:
```sage
sage: E = EllipticCurve(GF((5, 2)), [0,1])
sage: %time _ = list(E.isogenies_degree(27))
```

### DIFF
--- a/src/sage/schemes/elliptic_curves/hom.py
+++ b/src/sage/schemes/elliptic_curves/hom.py
@@ -270,6 +270,13 @@ class EllipticCurveHom(Morphism):
         if lx != rx:
             return richcmp_not_equal(lx, rx, op)
 
+        # Check the Weierstraß scaling factor, too (should be fast)
+
+        if op == op_EQ or op == op_NE:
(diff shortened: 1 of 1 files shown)


grhkm21:
Not sure why but doctests are failing with some unsorted issue.

grhkm21:
Oh, because you modified `_richcmp_`. Of course

yyyyx4:
Oops. Just pushed another version that only runs this check when testing `==` or `!=`.

grhkm21:
I think it looks good, I'll relabel it when doctest is done.

S17A05 (approved):
LGTM, and doctests seem to have passed too.

grhkm21:
Thanks, I forgot about it 🥲

PR #38821, opened 2024-10-17
let the category setup handle the ideals

fchapoton:
This is removing the `ideal` methods in the old `Ring` and `Field` classes, moving them to the category setup.

Also removing one custom `ideal` method in multiple polynomial rings.

### :memo: Checklist


fchapoton:
note : the removal of "ideal" in multi-polynomial requires a careful check, as it may have hidden side-effects

fchapoton:
At first sight, it seems to be ok thanks to the `_ideal_class_` method in `src/sage/rings/polynomial/multi_polynomial_ring_base.pyx`

src/sage/categories/fields.py
```diff
@@ -458,6 +458,26 @@ def fraction_field(self):
             """
             return self
 
+        def ideal(self, *gens, **kwds):
+            """
+            Return the ideal generated by gens.
+
```
mantepse, line 470:
```suggestion

            INPUT:

            - an element or a list/tuple/sequence of elements, the generators

            Any named arguments are ignored.
```

src/sage/categories/fields.py
```diff
@@ -458,6 +458,26 @@ def fraction_field(self):
...
+            EXAMPLES::
+
+                sage: QQ.ideal(2)
+                Principal ideal (1) of Rational Field
+                sage: QQ.ideal(0)
+                Principal ideal (0) of Rational Field
```
mantepse, line 476:
```suggestion
                Principal ideal (0) of Rational Field

	    TESTS::

                sage: QQ.ideal(2, 4)
                Principal ideal (1) of Rational Field

                sage: QQ.ideal([2, 4])
                Principal ideal (1) of Rational Field             
```

src/sage/categories/fields.py
```diff
@@ -458,6 +458,26 @@ def fraction_field(self):
...
+            """
+            if len(gens) == 1 and isinstance(gens[0], (list, tuple)):
+                gens = gens[0]
+            if not isinstance(gens, (list, tuple)):
+                gens = [gens]
```
mantepse: I think this cannot happen: on entry `gens` is always a list, and only if it has length 1 and the first entry is a list (or tuple), we modify it.

fchapoton: oui, en effet. J'ai viré ces deux lignes.

src/sage/categories/rings.py
```diff
@@ -888,27 +921,26 @@ def ideal(self, *args, **kwds):
...
-                            elif isinstance(first, Element):
-                                gens = [first]
-                            else:
-                                raise ArithmeticError("there is no coercion from %s to %s" % (first, self))
-                        except TypeError:  # first may be a ring element
-                            pass
                         break
-            if coerce:
+
+            if len(gens) == 0:
```
mantepse:
```suggestion
            if not gens:
```

src/sage/categories/rings.py
```diff
@@ -888,27 +921,26 @@ def ideal(self, *args, **kwds):
                         break
                     elif isinstance(first, (list, tuple, GeneratorType)):
                         gens = first
+                    elif isinstance(first, Parent) and self.has_coerce_map_from(first):
+                        gens = first.gens()  # we have a ring as argument
```
mantepse: I don't quite understand, the INPUT says that `gens` should be - essentially - generators.  I don't see a doctest where they are rings, either.

fchapoton: ok, I copied that when merging the duplicated code. Let's try to remove this.

src/sage/categories/rings.py
```diff
@@ -889,26 +922,23 @@ def ideal(self, *args, **kwds):
...
                 # Use GCD algorithm to obtain a principal ideal
                 g = gens[0]
                 if len(gens) == 1:
                     try:
-                        g = g.gcd(g)  # note: we set g = gcd(g, g) to "canonicalize" the generator: make polynomials monic, etc.
-                    except (AttributeError, NotImplementedError):
+                        # note: we set g = gcd(g, g) to "canonicalize" the generator:
+                        # make polynomials monic, etc.
+                        g = g.gcd(g)
+                    except (AttributeError, NotImplementedError, IndexError):
```
mantepse, line 941: It would be good to have a test for this line, in particular because `IndexError` was added.

mantepse:
I played a little with the patch and it looks good to me.  I found a bug, but that's present in develop, too:
```
sage: I.<x> = InfinitePolynomialRing(QQ)
sage: II = I.ideal( [x[0]^3,x[1]^3+x[0]^2, x[1]^2] )
sage: x[0]^2 in II
False
sage: I.<x0, x1> = PolynomialRing(QQ)
sage: II = I.ideal( [x0^3,x1^3+x0^2, x1^2] )
sage: x0^2 in II
True
```

src/sage/categories/rings.py
fchapoton: so far, I was not able to find a ring that pass through that case.

mantepse: https://en.wikipedia.org/wiki/Euclidean_domain has $\mathbb R[x,y]/(x^2+y^2+1)$, but sage does not recognise it as principal ideal domain (or does $\mathbb Q$ vs $\mathbb R$ make a difference?):
```
sage: R.<x,y> = PolynomialRing(QQ)
sage: Q = R.quotient(x^2+y^2+1)
sage: Q in PrincipalIdealDomains()
False
```
Also, if I made no mistake, $\mathbb Q(\sqrt{19})$ would constitute an example of a principal ideal domain which is not Euclidean, but sage does not know it (or do I misunderstand the code?):
```
sage: R.<x> = QQ[]
sage: K.<a> = QQ.extension(x^2 + 19); K
Number Field in a with defining polynomial x^2 + 19
sage: K
Number Field in a with defining polynomial x^2 + 19
sage: K in PrincipalIdealDomains()
True
sage: K in EuclideanDomains()
True
```

(Disclaimer: I know nothing about this kind of math)

So, I guess it's best to leave it in.

mantepse (approved):
LGTM

src/sage/categories/rings.py
fchapoton: One can declare the category a posteriori

`Q._refine_category_(PrincipalIdealDomains())`

but then creating the ideal follows a specific code-path for quotient rings.

For the second part, I guess you mean the ring of integers rather than the number field. I tried, and there is also a specific code-path for ideals there.

fchapoton:
thanks ! So positive review ?

mantepse:
sorry, forgot

PR #38822, opened 2024-10-17
Upgrade database_knotinfo to version 2024.10.1

soehms:
The current `database_knotinfo` version 2024.10.1 is no longer compatible with Sage because the syntax for 16 knots for which two braid notations are recorded has changed (see the corresponding [Release Notes](https://github.com/soehms/database_knotinfo/releases/tag/2024.10.1)).

This is fixed by this PR. In addition, two new fields (`geometric_type` and `cosmetic_crossing`) added in this version are used to implement corresponding methods (`is_hyperbolic` and `cosmetic_crossing_conjecture_verified`) of the interface.

### :memo: Checklist

### DIFF
--- a/src/sage/knots/knotinfo.py
+++ b/src/sage/knots/knotinfo.py
@@ -1356,7 +1356,7 @@ class KnotInfoBase(Enum):
         verified = not knotinfo_bool(cosmetic_crossing)
         if not knotinfo_bool(cosmetic_crossing):
             return True
-        raise AssertionError('This seems to be a negative example to the conjecture')
+        raise AssertionError(f'{self} is a counterexample to the cosmetic crossing conjecture')
 
     @cached_method
     def homfly_polynomial(self, var1='v', var2='z', original=False):
--- a/src/sage/knots/knotinfo.py
+++ b/src/sage/knots/knotinfo.py
@@ -1354,8 +1354,9 @@ class KnotInfoBase(Enum):
         if not cosmetic_crossing or cosmetic_crossing == 'Unknown':
             return False
         verified = not knotinfo_bool(cosmetic_crossing)
-        assert verified
-        return verified
+        if not knotinfo_bool(cosmetic_crossing):
+            return True
+        raise AssertionError('This seems to be a negative example to the conjecture')
 
     @cached_method
     def homfly_polynomial(self, var1='v', var2='z', original=False):
--- a/src/sage/knots/knotinfo.py
+++ b/src/sage/knots/knotinfo.py
@@ -790,7 +790,7 @@ class KnotInfoBase(Enum):
             return (1, )
 
         braid_notation = eval_knotinfo(braid_notation)
-        if type(braid_notation) is list:
+        if type(braid_notation) in (list, tuple):
             # in some cases there are a pair of braid representations
             # in the database. If this is the case we select the
             # corresponding to the braid index.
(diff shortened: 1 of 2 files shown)


soehms:
For non-hyperbolic knots, the `geometric_type` field contains additional information. This is construction information for torus and satellite knots. Maybe we can use this for another new method `construction` and combine it with information from the `pretzel_notation` field. Would that make sense?

src/sage/knots/knotinfo.py
```diff
@@ -1183,6 +1183,23 @@ def is_amphicheiral(self, positive=False):
...
+
+        EXAMPLES::
+
+            sage: KnotInfo.K3_1.is_hyperbolic()
+            False
+            sage: KnotInfo.K5_2.is_hyperbolic()
+            True
+        """
+        geometric_type = self[self.items.geometric_type]
+        if (geometric_type == 'hyperbolic'):
```
tscrim:
```suggestion
        if geometric_type == 'hyperbolic':
```
Python :P

src/sage/knots/knotinfo.py
```diff
@@ -1309,6 +1326,37 @@ def is_oriented(self):
...
+        EXAMPLES::
+
+            sage: knots = [K for K in KnotInfo if K.is_knot() and K.crossing_number() < 10]
+            sage: all(K.cosmetic_crossing_conjecture_verified() for K in knots)
+            True
+        """
+        cosmetic_crossing = self[self.items.cosmetic_crossing]
+        if self.crossing_number() == 0:
+            return True
+        if (not cosmetic_crossing or cosmetic_crossing == 'Unknown'):
```
tscrim:
```suggestion
        if not cosmetic_crossing or cosmetic_crossing == 'Unknown':
```

src/sage/knots/knotinfo.py
```diff
@@ -1309,6 +1326,37 @@ def is_oriented(self):
...
+            return False
+        verified = not knotinfo_bool(cosmetic_crossing)
+        assert verified
+        return verified
```
tscrim: You should choose one or the other. It makes no sense to return `True` if passed as the `assert` will fail otherwise. So I would either do the return or raise an error (with no `return` statement).

(Technically, yes, there is a difference as we could turn off `assert` checks, but that isn't easy to do. Plus `assert`s are meant for things that should be true anyways.)

tscrim (approved):
Two minor things. Once addressed, positive review.

src/sage/knots/knotinfo.py
soehms: Thanks for the review! 

> You should choose one or the other. It makes no sense to return True if passed as the assert will fail otherwise. So I would either do the return or raise an error (with no return statement).

I've done this in the last commit.

src/sage/knots/knotinfo.py
```diff
@@ -1309,6 +1326,38 @@ def is_oriented(self):
...
+        """
+        cosmetic_crossing = self[self.items.cosmetic_crossing]
+        if self.crossing_number() == 0:
+            return True
+        if not cosmetic_crossing or cosmetic_crossing == 'Unknown':
+            return False
+        verified = not knotinfo_bool(cosmetic_crossing)
+        if not knotinfo_bool(cosmetic_crossing):
+            return True
+        raise AssertionError('This seems to be a negative example to the conjecture')
```
tscrim:
```suggestion
        raise AssertionError(f'{self} is a counterexample to the cosmetic crossing conjecture')
```

tscrim (approved):
Thank you. One last little cosmetic (pun intended) thing to fix. Once done, you can set a positive review.

src/sage/knots/knotinfo.py
soehms: That actually sounds better!

tscrim:
Thank you. LGTM.

soehms:
> Thank you. LGTM.

Thanks too!

PR #38825, opened 2024-10-19
Fix two tests that fail on a heavily-loaded machine

orlitzky:
The commit messages explain each change in detail, but tl;dr these two tests fail for me when the machine is under heavy load because they are guessing how long it will take some computation to complete and the guess is wrong.

### DIFF
--- a/src/sage/parallel/map_reduce.py
+++ b/src/sage/parallel/map_reduce.py
@@ -1136,21 +1136,12 @@ class RESetMapReduce():
             sage: from sage.parallel.map_reduce import RESetMapReduce
             sage: def children(x):
             ....:     print(f"Starting: {x}", flush=True)
-            ....:     sleep(float(0.5))
-            ....:     print(f"Finished: {x}", flush=True)
             ....:     return []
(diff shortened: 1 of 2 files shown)


tobiasdiez (approved):

src/sage/parallel/map_reduce.py
```diff
@@ -1136,21 +1136,12 @@ def start_workers(self):
             sage: from sage.parallel.map_reduce import RESetMapReduce
             sage: def children(x):
             ....:     print(f"Starting: {x}", flush=True)
-            ....:     sleep(float(0.5))
-            ....:     print(f"Finished: {x}", flush=True)
             ....:     return []
             sage: S = RESetMapReduce(roots=[1, 2], children=children)
             sage: S.setup_workers(2)
-            sage: S.start_workers(); sleep(float(0.4))
+            sage: S.start_workers(); sleep(float(5))  # long time
```
tornaria, line 1142: The `# long time` label causes a test failure in the cleanup section, when running tests in non-long mode.

orlitzky: https://github.com/sagemath/sage/pull/38873

PR #36226, opened 2023-09-09
Use CPU time for --warn-long

orlitzky:
Finally address #32981.

Using wall time to measure which tests are "long" doesn't make much sense. Wall time is subject to every conflating factor:

1. Faster hardware runs tests in less wall time
2. Faster compiler flags makes code that runs in less wall time
3. A busy machine will run tests slower than an unburdened one
4. Operating system priority can increase or decrease the time taken by a process
5. etc.

Ideally, everyone would agree on how long is too long for a test to run. Attaining that ideal is not so easy, but the first step is to switch`--warn-long` away from using wall time and towards using CPU time. A second phase (https://github.com/sagemath/sage/issues/33022) will normalize the CPU time.

#### --warn-long output

The user-facing portion of this can be tested with a low `--warn-long` value,

```
$ sage -t --warn-long=1 src/sage/rings/qqbar.py 
...
Doctesting 1 file.
sage -t --warn-long --random-seed=309742757431591512726158538133919135994 src/sage/rings/qqbar.py
**********************************************************************
File "src/sage/rings/qqbar.py", line 511, in sage.rings.qqbar
Warning, slow doctest:
    z, = [r for r in p4.roots(QQbar, False) if r in ival]
Test ran for 1.51s cpu, 1.51s wall
Check ran for 0.00s cpu, 0.00s wall
**********************************************************************
File "src/sage/rings/qqbar.py", line 524, in sage.rings.qqbar
Warning, slow doctest:
    z2 = QQbar.polynomial_root(p4, ival)
Test ran for 1.07s cpu, 1.07s wall
Check ran for 0.00s cpu, 0.00s wall
**********************************************************************
File "src/sage/rings/qqbar.py", line 808, in sage.rings.qqbar.AlgebraicField_common._factor_multivariate_polynomial
Warning, slow doctest:
    F = QQbar._factor_multivariate_polynomial(p)
Test ran for 2.38s cpu, 2.42s wall
Check ran for 0.00s cpu, 0.00s wall

...
```

#### pexpect accounting

Two different methods are used to account for the CPU time, one for linux, and one for everything else. This GAP command uses pexpect and should chew up a good bit of CPU time:

```
sage: from sage.doctest.util import Timer
sage: t = Timer()
sage: _ = t.start(); t.stop()    # no CPU time used between start/stop
{'cputime': 0.0, 'walltime': 7.605552673339844e-05}

[...]

PR #38826, opened 2024-10-19
src/sage/env.py: canonicalize paths in a test

orlitzky:
A test in sage.env is running sage in a subprocess to compare the values of `SAGE_ROOT` and `SAGE_LOCAL`. It does the comparison as strings, however, and can fail:

```
File "src/sage/env.py", line 14, in sage.env
Failed example:
    out == repr((SAGE_ROOT, SAGE_LOCAL))   # long time
Expected:
    True
Got:
    False
```

This despite the fact that both values are equivalent:

```
sage: out
"('/home/mjo/src/sage.git/src/sage/../..', '/usr')"
sage: repr((SAGE_ROOT, SAGE_LOCAL))
"('/home/mjo/src/sage.git', '/usr')"
```

We update the test to canonicalize the paths within the subprocess, and output only "True" or "False" instead.

### DIFF
--- a/src/sage/env.py
+++ b/src/sage/env.py
@@ -8,10 +8,14 @@ environment variables, and has the same ``SAGE_ROOT`` and ``SAGE_LOCAL``
 
     sage: env = {k:v for (k,v) in os.environ.items() if not k.startswith("SAGE_")}
     sage: from subprocess import check_output
-    sage: environment = "sage.all"
-    sage: cmd = f"from {environment} import SAGE_ROOT, SAGE_LOCAL; print((SAGE_ROOT, SAGE_LOCAL))"
+    sage: module_name = "sage.all"   # hide .all import from the linter
+    sage: cmd  = f"from {module_name} import SAGE_ROOT, SAGE_LOCAL;"
+    sage: cmd +=  "from os.path import samefile;"
+    sage: cmd += f"s1 = samefile(SAGE_ROOT, '{SAGE_ROOT}');"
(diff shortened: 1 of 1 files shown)


orlitzky:
Oh, so that's why it was using f-string substitution for a command with no variables...

```
Linting files... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:09
╭─ Error: namespace_pkg_all_import: import from .all of a namespace package ───╮
│      9     sage: env = {k:v for (k,v) in os.environ.items() if not k.starts  │
│     10     sage: from subprocess import check_output                         │
│  ❱  11     sage: cmd  =  "from sage.all import SAGE_ROOT, SAGE_LOCAL;"       │
```

orlitzky:
CI is an unrelated timeout:

```
sage -t --long --random-seed=286735480429121101562228604801325644303 src/sage/coding/linear_code.py  # Timed out
```

I'm going to start shamelessly plugging,

* https://github.com/sagemath/sage/pull/36226

On every ticket where the CI "fails" due to slow tests.

tobiasdiez (approved):

antonio-rojas:
This breaks the test on distro builds where `SAGE_ROOT` is not defined
```
> /usr/bin/python -c "from sage.all import SAGE_ROOT, SAGE_LOCAL;from os.path import samefile;s1 = samefile(SAGE_ROOT, 'None');"

  File "<string>", line 1, in <module>
  File "<frozen genericpath>", line 112, in samefile
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

PR #38829, opened 2024-10-20
support defining quaternion orders from a generating set

yyyyx4:
Same as #36759, but for orders instead of ideals.

(Eventually, both should become a subclass of the same "quaternion lattice" type anyway...)


yyyyx4:
There are many doctests failing in `src/sage/schemes/elliptic_curves/heegner.py`, but I'm not familiar with any of the math there...

PR #38835, opened 2024-10-22
Remove redundant modulo operation in vector_modn_dense

user202729:
Just a tiny change. There's no user-visible behavior seen.

(The output of `int(…)` is already reduced modulo `p` anyway.)

### :memo: Checklist

### DIFF
--- a/src/sage/modules/vector_modn_dense.pyx
+++ b/src/sage/modules/vector_modn_dense.pyx
@@ -171,13 +171,15 @@ cdef class Vector_modn_dense(free_module_element.FreeModuleElement):
         """
         Create an element.
 
+        TESTS:
+
         Note that ``coerce=False`` is dangerous::
 
             sage: V = VectorSpace(GF(7), 3)
             sage: v = V([2, 9, -5], coerce=False)
             sage: v[0] == v[1]
             False
-            sage: v[0]+1 == v[1]+1
+            sage: v[0] + 1 == v[1] + 1
             True
             sage: v[0] == v[2]
             False
--- a/src/sage/modules/vector_modn_dense.pyx
+++ b/src/sage/modules/vector_modn_dense.pyx
@@ -184,7 +184,7 @@ cdef class Vector_modn_dense(free_module_element.FreeModuleElement):
         """
         cdef Py_ssize_t i
         cdef mod_int a
-        if isinstance(x, xrange):
+        if isinstance(x, range):
             x = tuple(x)
         if isinstance(x, (list, tuple)):
             if len(x) != self._degree:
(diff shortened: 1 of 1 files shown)


src/sage/modules/vector_modn_dense.pyx
```diff
@@ -168,19 +168,32 @@ cdef class Vector_modn_dense(free_module_element.FreeModuleElement):
...
+            False
+            sage: v[0]+1 == v[1]+1
+            True
+            sage: v[0] == v[2]
+            False
+        """
         cdef Py_ssize_t i
-        cdef mod_int a, p
+        cdef mod_int a
         if isinstance(x, xrange):
```
fchapoton: please replace xrange by range here

user202729: Okay, but apparently it's intentional that Cython supports `xrange` (will just be replaced by `range`)

https://github.com/cython/cython/issues/1036

src/sage/modules/vector_modn_dense.pyx
```diff
@@ -168,19 +168,32 @@ cdef class Vector_modn_dense(free_module_element.FreeModuleElement):
             self._init(parent.degree(), parent, parent.base_ring().order())
 
     def __init__(self, parent, x, coerce=True, copy=True):
+        """
+        Create an element.
+
+        Note that ``coerce=False`` is dangerous::
```
tscrim, line 176:
```suggestion

        TESTS:

        Note that ``coerce=False`` is dangerous::
```

src/sage/modules/vector_modn_dense.pyx
```diff
@@ -168,19 +168,32 @@ cdef class Vector_modn_dense(free_module_element.FreeModuleElement):
...
+
+            sage: V = VectorSpace(GF(7), 3)
+            sage: v = V([2, 9, -5], coerce=False)
+            sage: v[0] == v[1]
+            False
+            sage: v[0]+1 == v[1]+1
```
tscrim:
```suggestion
            sage: v[0] + 1 == v[1] + 1
```
PEP8 spacing.

tscrim (changes requested):
Two extra little things. Otherwise I think this is good to go.

src/sage/modules/vector_modn_dense.pyx
user202729: Not sure if this should be test-only. The fact that "coerce=False is dangerous" feel like something that the user should know about.

tscrim: This won’t be in the public documentation, but you can make it `EXAMPLES:` if you prefer.

user202729: Ah yes, the method is a magic method. So users reading e.g. https://doc.sagemath.org/html/en/reference/modules/sage/modules/vector_modn_dense.html won't even know `coerce=` exists at all…?

Sure, sounds reasonable.

tscrim (approved):
Thanks. Positive review.

src/sage/modules/vector_modn_dense.pyx
tscrim: Yes, that's right. Users aren't supposed to really be looking at that anyways as there are other entry points. That is more for developers.

PR #38836, opened 2024-10-22
Migrate bliss extension to meson

tobiasdiez:
WIP

Follow-up to https://github.com/sagemath/sage/pull/36524.

### :memo: Checklist


tobiasdiez:
After thinking more about this, I don't see the point of these modularized distributions. Meson is now able to selectively build the bliss-part of sage depending on whether bliss is installed or not. Thus a separate module where bliss is marked as required seems to be rather pointless to me. I guess the only advantage would be shipping a binary wheel that includes bliss, but that could also easily be done as part of the main project. In summary: if someone wants to continue here, please feel free to do so.

@dimpase @orlitzky if you want to create a giac package, you can take this PR here as a blueprint.

@kwankyu The "Build and Test" are failing without any error message. I had this in a couple of PRs, mostly when there are compilation errors. This is rather inconvenient. Could you please fix this (My understanding is that you are one of the very few that understand the CI infrastructure)?

dimpase:
@tobiasdiez - I agree. 

I am busy with the move now, till next week.

orlitzky:
I only see the value if they are moved to a separate repository, but there are now many more important things to do, like move all of the `./configure` checks into meson, and figure out a way for the meson equivalent of `--disable-foo` to actually disable foo (rather than have it slowly autodetected at runtime over and over again).

tobiasdiez:
> figure out a way for the meson equivalent of `--disable-foo` to actually disable foo (rather than have it slowly autodetected at runtime over and over again).

That's easy to do. If you tell me what you want to disable, then I can prepare a PR.

(Btw, meson caches its dependency checks - at least if they provide pkgconfig data.)

dimpase:
> I only see the value if they are moved to a separate repository, but there are now many more important things to do, like move all of the `./configure` checks into meson, and figure out a way for the meson equivalent of `--disable-foo` to actually disable foo (rather than have it slowly autodetected at runtime over and over again).

If this would allow us to get rid of the `sage.feature`  things (or at least some of them - runtime checks for libraries are really a weapon of last resort), it would be a great simplification.
I believe that `sage.feature`  was introduced in the 1st place to overcome the shortcomings of the build system.

tobiasdiez:
This makes a lot of sense. I would wait, however, with implementing this until we have faded out the old setuptools-based build and only use meson. Otherwise we have to implement many workarounds since we cannot rely on the build system to check for things during build time.

PR #38843, opened 2024-10-23
Improve Matrix_mod2_dense solve_right performance

maple3142:
Use M4RI builtin `mzd_solve_left` function to solve equation.

Currently, using `solve_right` on large `GF(2)` matrix is pretty slow because it use a generic solving function in matrix2.pyx, but M4RI provides a `mzd_solve_left` function that does the same thing.

For example, try running this in sage's IPython REPL:

```python
set_random_seed(12345)
n, m = 20000, 19968
A = random_matrix(GF(2), n, m)
x = random_vector(GF(2), m)
B = A*x
%time sol = A.solve_right(B)
assert A * sol == B
```

without this PR, it took

```
CPU times: user 1min 2s, sys: 190 ms, total: 1min 2s
Wall time: 15.2 s
```

with this PR, it becomes

```
CPU times: user 12.5 s, sys: 40.8 ms, total: 12.6 s
Wall time: 1.09 s
```

### :memo: Checklist

### DIFF
--- a/src/sage/matrix/matrix_mod2_dense.pyx
+++ b/src/sage/matrix/matrix_mod2_dense.pyx
@@ -1958,9 +1958,13 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
             sage: A * A.solve_right(B) == B
             True
         """
-        cdef Matrix_mod2_dense X  # the solution
-
         cdef mzd_t *B_entries = (<Matrix_mod2_dense>B)._entries
+
+        cdef Matrix_mod2_dense X  # the solution
+        X = self.new_matrix(nrows=self._entries.ncols, ncols=B_entries.ncols)
+        if self._entries.nrows == 0 or self._entries.ncols == 0:
+            # special case: empty matrix
+            return X
         cdef rci_t rows = self._entries.nrows
         if self._entries.nrows < self._entries.ncols:
             rows = self._entries.ncols  # mzd_solve_left requires ncols <= nrows
@@ -1979,7 +1983,6 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
 
             if ret == 0:
                 # solution is placed in rhs
-                X = self.new_matrix(nrows=self._entries.ncols, ncols=B_entries.ncols)
                 rhs.nrows = self._entries.ncols
                 mzd_copy(X._entries, rhs)
                 return X
--- a/src/sage/matrix/matrix_mod2_dense.pyx
+++ b/src/sage/matrix/matrix_mod2_dense.pyx
@@ -1962,7 +1962,7 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
 
         cdef Matrix_mod2_dense X  # the solution
         X = self.new_matrix(nrows=self._entries.ncols, ncols=B_entries.ncols)
-        if self._entries.nrows == 0 or self._entries.ncols == 0:
+        if self._entries.ncols == 0 or B_entries.ncols == 0:
             # special case: empty matrix
             return X
         cdef rci_t rows = self._entries.nrows
--- a/src/sage/matrix/matrix_mod2_dense.pyx
+++ b/src/sage/matrix/matrix_mod2_dense.pyx
@@ -1970,22 +1970,24 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
         cdef mzd_t *rhs = mzd_init(rows, B_entries.ncols)
         mzd_copy(rhs, B_entries)
 
-        sig_on()
-        # although it is called mzd_solve_left, it does the same thing as solve_right
-        ret = mzd_solve_left(lhs, rhs, 0, check)
-        sig_off()
-        mzd_free(lhs)
+        cdef int ret
+        try:
+            sig_on()
+            # although it is called mzd_solve_left, it does the same thing as solve_right
+            ret = mzd_solve_left(lhs, rhs, 0, check)
+            sig_off()
 
-        if ret == 0:
-            # solution is placed in rhs
-            X = self.new_matrix(nrows=self._entries.ncols, ncols=B_entries.ncols)
-            rhs.nrows = self._entries.ncols
-            mzd_copy(X._entries, rhs)
+            if ret == 0:
+                # solution is placed in rhs
+                X = self.new_matrix(nrows=self._entries.ncols, ncols=B_entries.ncols)
+                rhs.nrows = self._entries.ncols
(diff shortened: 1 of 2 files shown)


src/sage/matrix/matrix_mod2_dense.pyx
```diff
@@ -1919,6 +1919,74 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
...
+        cdef mzd_t *lhs = mzd_init(rows, self._entries.ncols)
+        mzd_copy(lhs, self._entries)
+        cdef mzd_t *rhs = mzd_init(rows, B_entries.ncols)
+        mzd_copy(rhs, B_entries)
+
+        sig_on()
+        # although it is called mzd_solve_left, it does the same thing as solve_right
+        ret = mzd_solve_left(lhs, rhs, 0, check)
+        sig_off()
+        mzd_free(lhs)
```
user202729: Wouldn't this leak the `lhs` and `rhs` if user press ctrl+C?

maple3142: You are right, should I wrap it with a try...finally block or just disable the ability to Ctrl-C?

maple3142: Similarity, while `_right_kernel_matrix` does not enable signal but it can still leak the resulting kernel matrix when Ctrl-C (triggered during `self.new_matrix`).

user202729: Certainly handling it properly is preferable. (But no memory leak is a hard thing to test for…)

cysignals is a bit complicated…

https://cysignals.readthedocs.io/en/latest/interrupt.html#using-sig-on-and-sig-off

So far I don't exactly understand what's the difference between `try: sig_on() except KeyboardInterrupt:` versus `if not sig_on_no_except():` either.

user202729: for what it's worth, the part I don't understand is https://github.com/sagemath/cysignals/issues/205 .

In any case, I think the following will work:

```python
data = malloc(...)
if not sig_on_no_except():
	mzd_free(lhs)
	mzd_free(rhs)
	cython_check_exception()

ret = mzd_solve_left(...)
sig_off()
```

maple3142: Okay, I have pushed a version where try...finally are used to ensure allocated matrices get freed.

src/sage/matrix/matrix_mod2_dense.pyx
```diff
@@ -1919,6 +1919,81 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
...
+            True
+        """
+        cdef mzd_t *B_entries = (<Matrix_mod2_dense>B)._entries
+
+        cdef Matrix_mod2_dense X  # the solution
+        X = self.new_matrix(nrows=self._entries.ncols, ncols=B_entries.ncols)
+        if self._entries.ncols == 0 or B_entries.ncols == 0:
+            # special case: empty matrix
+            if B != 0:
+                raise ValueError("matrix equation has no solutions")
```
user202729, line 1990: Surely this is redundant (?) if B has no entries then every entries are 0?

(also even if it can happen you forget to free X here)

maple3142: There are cases where A have zero cols but B have non zero cols, so it would have a different result depends on whether B is zero. And without this, [some sage tests](https://github.com/sagemath/sage/actions/runs/11494820088/job/31993046661#step:10:9425) will fail.

As for X, I think it is a Python object so it would be handled by Python's refcount?

user202729:
Maybe add a few tests?

```
sage: matrix(GF(2), 2, 0).solve_right(matrix(GF(2), 2, 2))
[]
sage: matrix(GF(2), 2, 0).solve_right(matrix(GF(2), 2, 2)+1)
Traceback…

ValueError: matrix equation has no solutions
sage: matrix(GF(2), 2, 0).solve_right(matrix(GF(2), 2, 2)+1, check=False)
[]
sage: matrix(GF(2), 2, 2).solve_right(matrix(GF(2), 2, 0))
[]
```

(note that the old code returns empty matrix for the third case, while the new code raises error. Apparently it doesn't break any existing test case, but anyway.)

maple3142:
Tests for special case added.

> Strange that m4ri doesn't automatically handle the empty matrix case correctly (mathematically) though. Any idea why?

TBH, zero size matrix is pretty weird by itself I think, so it is not unusual to not consider that I think. It would probably be better handled by the `solve_right` in `matrix2.py` so the underlying function don't need to handle that.

maple3142:
Not sure what breaks the test using Meson on Python 3.9...

user202729:
My argument: just as zero is not a weird integer, zero-dimensional vector space (which is just a point) is not a weird space. So a zero-sized matrix is just a linear map from (or to) a zero-dimensional vector space.

The math checks out.

(Of course, programming-wise it is still a corner case.

user202729:
Looks like the failing test is https://github.com/sagemath/sage/pull/38940 . Should be unrelated to this change then.

src/sage/matrix/matrix_mod2_dense.pyx
```diff
@@ -1919,6 +1919,91 @@ cdef class Matrix_mod2_dense(matrix_dense.Matrix_dense):   # dense or sparse
...
+            sage: matrix(GF(2), 2, 0).solve_right(matrix(GF(2), 2, 2)) == matrix(GF(2), 0, 2)
+            True
+            sage: matrix(GF(2), 2, 0).solve_right(matrix(GF(2), 2, 2) + 1)
+            Traceback (most recent call last):
+            ...
+            ValueError: matrix equation has no solutions
+            sage: matrix(GF(2), 2, 0).solve_right(matrix(GF(2), 2, 2) + 1, check=False) == matrix(GF(2), 0, 2)
+            True
+            sage: matrix(GF(2), 2, 2).solve_right(matrix(GF(2), 2, 0)) == matrix(GF(2), 2, 0)
+            True
```
user202729, line 1969:
```suggestion
            True

        Check that it can be interrupted::

            sage: set_random_seed(12345)
            sage: n, m = 20000, 19968
            sage: A = random_matrix(GF(2), n, m)
            sage: x = random_vector(GF(2), m)
            sage: B = A*x
            sage: alarm(0.5); sol = A.solve_right(B)
            Traceback (most recent call last):
            ...
            AlarmInterrupt
```

user202729 (approved):
I don't have write permission anyway but… sure.

Maybe @grhkm21 would be interested?

grhkm21:
Thanks
love `# although it is called mzd_solve_left, it does the same thing as solve_right`

PR #38798, opened 2024-10-10
adding method tikz to class Graph

seblabbe:
Currently, we can create a TikzPicture from a graph as follows, but is raises an experimental feature warning:
```
sage: from sage.misc.latex_standalone import TikzPicture
sage: g = graphs.PetersenGraph()
sage: t = TikzPicture.from_graph(g)
<ipython-input-20-c4b6306d5e76>:1: FutureWarning: This class/method/function is marked as experimental. It, its functionality or its interface might change without a formal deprecation.
See https://github.com/sagemath/sage/issues/20343 for details.
  t = TikzPicture.from_graph(g)
sage: t
```
It was declared experimental during the review of #20343 because it should rather be a method of the class Graph.

This is what we do in this PR: we add a tikz method to the class Graph.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation
- [ ] and checked the documentation preview.

### :hourglass: Dependencies

None

### Screenshots

As a consequence, the following now works in Jupyter. By default, it uses the dot2tex format if dot2tex is available:

image

### Works well with sagetex

For example:
```
\begin{sagesilent}
g = graphs.PetersenGraph()
tikz = g.tikz()
\end{sagesilent}

\begin{center}
        \sageplot[scale=.5][pdf]{tikz}
\end{center}
```

PR #38847, opened 2024-10-24
Graph tikz method fixup

seblabbe:
tikz method for graphs was recently added in #38798

default value for `format` should be `None` as documented.

### :hourglass: Dependencies

#38798


dcoudert (commented):
LGTM, but this branch should be rebased on last beta.

fchapoton (approved):
ok

PR #38850, opened 2024-10-24
Fix incorrect logic in `can_convert_to_singular()`

### LINKED ISSUE #38849: `can_convert_to_singular()` should check the size of the ring for Z/NZ
### Steps To Reproduce

Currently `can_convert_to_singular(self)` checks the size of the characteristic for finite fields, but it does not check this for integer mod rings, leading to issues with the conditional logic.

```py
sage: from sage.rings.polynomial.polynomial_singular_interface import can_convert_to_singular
sage: can_convert_to_singular(GF(2**127 - 1))
True
sage: can_convert_to_singular(Zmod(2**1000))
True
```

### Expected Behavior

`_has_singular()` should be false when the size of the ring is larger than or equal to `2147483647`

### Actual Behavior

`_has_singular()` returns true for all rings of type `sage.rings.abc.IntegerModRing`

### Additional Information

_No response_

### Environment

- **Sage Version**: Latest

### Checklist

GiacomoPope:
Before this PR, the function `can_convert_to_singular(R)` would return true for `GF(p)[]` and `Zmod(N)[]` for `N` or `p` larger than the allowed bound of `p <= 2**31-1` due to a bug in the conditional logic.

This PR checks the size of integer mod rings as well as finite fields which should address this issue.

Fixes #38849

### DIFF
--- a/src/sage/rings/polynomial/polynomial_singular_interface.py
+++ b/src/sage/rings/polynomial/polynomial_singular_interface.py
@@ -424,6 +424,21 @@ def can_convert_to_singular(R):
         sage: R.<x,y> = Zmod(10^20 + 1)[]
         sage: R._has_singular
         True
+
+    Check for :issue:`38849`::
+        sage: # needs sage.rings.finite_rings
+        sage: R.<x,y> = GF(163)[]
+        sage: R._has_singular
+        True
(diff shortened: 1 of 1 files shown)


GiacomoPope:
I think I was wrong about this. Closing.

PR #38854, opened 2024-10-25
Add LaTeX representation for function fields

vincentmacri:
Adds LaTeX representation for function fields to `function_field_polymod.py`.

### :memo: Checklist

### :hourglass: Dependencies

None


src/sage/rings/function_field/function_field_polymod.py
```diff
@@ -571,7 +571,22 @@ def _repr_(self):
             sage: L._repr_()
             'Function field in y defined by y^5 - 2*x*y + (-x^4 - 1)/x'
         """
-        return "Function field in %s defined by %s" % (self.variable_name(), self._polynomial)
+        return (f"Function field in {self.variable_name()} defined by "
+                f"{self._polynomial}")
+
```
kwankyu, line 575: make this just one line (one f-string).

kwankyu:
Otherwise lgtm.

src/sage/rings/function_field/function_field_polymod.py
vincentmacri: That would exceed our 80 character line length limit and not comply with the style guide. I'm not personally a fan of line length limits, but I want to make sure you're okay with violating the style guide in this instance before I do that. Personally, I prefer consistency (even if I may disagree with a particular part of the style guide) and think we should follow the style guide.

kwankyu: I was aware of that. The 80 character limit is for readability. In the present case, I think the readability will get better by joining the dangling part into the previous line. Readability is more important than the rule.

My suggestion is a mild one. Taking it or not is just up to you. Let me know which you prefer.

vincentmacri: Fair point. I've taken the suggestion.

kwankyu: Thanks.

PR #38783, opened 2024-10-07
Fix gfortran and e-antic builds with XCode 16 on x86_64

culler:
This PR modifies the patches in the gfortran spkg and adds a patch to the boost_cropped spkg. It addresses some issues caused by changes added in XCode 16, which is the XCode version compatible with macOS 15 Sequoia.

The gfortran spkg failed to build on macOS 15 Sequoia with XCode 16 because of [this bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116809) which is specific to clang 16.  The bug report contains a patch, which is what this PR adds.  However, that patch changed files which were also being patched by the existing gcc-13.3.0-arm.patch.  Rather than have patches which must be applied in a certain order, this PR replaces the gcc-13.3.0-arm patch with a new patch file named gcc-13_3_0.patch which incorporates the new changes (no longer just for arm) with the older ones.

The e-antic spkg failed to build on macOS 15 Sequoia.  The reason turned out to be that XCode 16, unlike earlier versions, now generates an error for code which used to generate a warning about assigning an out-of-range value to an enum.  The code that triggers the error is in a boost header file.  There is a patch for the header file in [this macports ticket](https://trac.macports.org/ticket/68920).  The patch modifies a line in the header file which controls how this warning / error is handled by certain compilers.  The change adds clang 16 to the list of those compilers.   

See [https://groups.google.com/g/sage-devel/c/sv3jCczVa40](https://groups.google.com/g/sage-devel/c/sv3jCczVa40l) and [https://groups.google.com/g/sage-devel/c/yYRBm-hc918](https://groups.google.com/g/sage-devel/c/yYRBm-hc918)

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [ ] I have created tests covering the changes.
- [ ] I have updated the documentation and checked the documentation preview.

PR #38855, opened 2024-10-25
Fixes for Xcode 16 and gcc/gfortran, boost_cropped

jhpalmieri:
Fix gfortran and boost_cropped builds for Xcode 16

I am reopening the PR at #38783. The patches there work well and should be merged. (The original PR could not be reopened because the repo upon which it was based has been removed.) From that PR's description:

This PR modifies the patches in the gfortran spkg and adds a patch to the boost_cropped spkg. It addresses some issues caused by changes added in XCode 16, which is the XCode version compatible with macOS 15 Sequoia.

The gfortran spkg failed to build on macOS 15 Sequoia with XCode 16 because of this bug which is specific to clang 16. The bug report contains a patch, which is what this PR adds. However, that patch changed files which were also being patched by the existing gcc-13.3.0-arm.patch. Rather than have patches which must be applied in a certain order, this PR replaces the gcc-13.3.0-arm patch with a new patch file named gcc-13_3_0.patch which incorporates the new changes (no longer just for arm) with the older ones.

The e-antic spkg failed to build on macOS 15 Sequoia. The reason turned out to be that XCode 16, unlike earlier versions, now generates an error for code which used to generate a warning about assigning an out-of-range value to an enum. The code that triggers the error is in a boost header file. There is a patch for the header file in this macports ticket. The patch modifies a line in the header file which controls how this warning / error is handled by certain compilers. The change adds clang 16 to the list of those compilers.

See https://groups.google.com/g/sage-devel/c/sv3jCczVa40 and https://groups.google.com/g/sage-devel/c/yYRBm-hc918

### :memo: Checklist

### DIFF
--- a/build/pkgs/boost_cropped/patches/integral_wrapper.patch
+++ b/build/pkgs/boost_cropped/patches/integral_wrapper.patch
@@ -1,0 +1,13 @@
+diff --git a/boost/mpl/aux_/integral_wrapper.hpp b/boost/mpl/aux_/integral_wrapper.hpp
+index 6bc05f7..4f3f696 100644
+--- a/boost/mpl/aux_/integral_wrapper.hpp
++++ b/boost/mpl/aux_/integral_wrapper.hpp
+@@ -56,7 +56,7 @@ struct AUX_WRAPPER_NAME
+ // have to #ifdef here: some compilers don't like the 'N + 1' form (MSVC),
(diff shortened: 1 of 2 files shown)


jhpalmieri:
All of the work was done at #38783. Marc Culler is the author, I am the reviewer.

PR #38859, opened 2024-10-26
`build/pkgs/matroid_database`: mark as math

gmou3:
From https://github.com/passagemath/passagemath/pull/183:

'Currently it appears in the "Other dependencies" section in https://doc.sagemath.org/html/en/reference/spkg/; it should go in the "Math" section.'


kwankyu (approved):
LGTM.

PR #38860, opened 2024-10-26
Fix test if SAGE_ROOT is not defined

antonio-rojas:
`SAGE_ROOT` is not defined in distro builds, make the test pass in that case

### DIFF
--- a/src/sage/env.py
+++ b/src/sage/env.py
@@ -11,7 +11,7 @@ environment variables, and has the same ``SAGE_ROOT`` and ``SAGE_LOCAL``
     sage: module_name = "sage.all"   # hide .all import from the linter
     sage: cmd  = f"from {module_name} import SAGE_ROOT, SAGE_LOCAL;"
     sage: cmd +=  "from os.path import samefile;"
-    sage: cmd += f"s1 = samefile(SAGE_ROOT, '{SAGE_ROOT}');"
+    sage: cmd += f"s1 = samefile(SAGE_ROOT, '{SAGE_ROOT}') if SAGE_ROOT else True;"
     sage: cmd += f"s2 = samefile(SAGE_LOCAL, '{SAGE_LOCAL}');"
     sage: cmd += "print(s1 and s2);"
     sage: out = check_output([sys.executable, "-c", cmd], env=env).decode().strip()   # long time


orlitzky (approved):
Sorry about that, the fix looks fine to me.

src/sage/env.py
```diff
@@ -11,7 +11,7 @@
     sage: module_name = "sage.all"   # hide .all import from the linter
     sage: cmd  = f"from {module_name} import SAGE_ROOT, SAGE_LOCAL;"
     sage: cmd +=  "from os.path import samefile;"
-    sage: cmd += f"s1 = samefile(SAGE_ROOT, '{SAGE_ROOT}');"
+    sage: cmd += f"s1 = samefile(SAGE_ROOT, '{SAGE_ROOT}') if SAGE_ROOT else True;"
```
tornaria, line 14: This will fail to detect the case where `SAGE_ROOT` is set (in the parent, the sage session) but becomes `None` in the subprocess (the python session which we are testing).

Maybe
```suggestion
    sage: cmd += f"s1 = SAGE_ROOT == {SAGE_ROOT!r} or samefile(SAGE_ROOT, {SAGE_ROOT!r});"
```

tornaria: Or more verbose:
```suggestion
    sage: if SAGE_ROOT is None:
    ....:     cmd += f"s1 = SAGE_ROOT is None;"
    ....: else: 
    ....:     cmd += f"s1 = samefile(SAGE_ROOT, '{SAGE_ROOT}');"
```

orlitzky: The purpose of this doctest is supposedly to...


```python
Verify that importing ``sage.all`` works in Sage's Python without any ``SAGE_``
environment variables, and has the same ``SAGE_ROOT`` and ``SAGE_LOCAL``             
(see also :issue:`29446`)::
```

Can the two `SAGE_ROOT`s be different under normal circumstances? I'm happy to go back and improve this, I'm just a little behind.

tornaria: The point is that it should not be, and this is what we are testing.

In other words, we expect that if `SAGE_ROOT="something"` then it will be the same when used from python. Now suppose there is a bug that causes `SAGE_ROOT=None` when used from python. Then your test will pass but we want it to fail.

orlitzky:
follow-up in https://github.com/sagemath/sage/pull/38935

PR #38863, opened 2024-10-27
Fix testing of installed rst files

antonio-rojas:
Fix a regression from https://github.com/sagemath/sage/pull/36524 where `sage -t` skips doctesting installed `rst` files (which get installed as `rst.txt`)


tobiasdiez (approved):
Oh, that was a silly mistake of mine. Thanks for the fix!

PR #38865, opened 2024-10-27
make gap objects selectively hashable

### LINKED ISSUE #35926: Incorrect `__hash__` method for `GapElement`
I noticed the following in `src/sage/libs/gap/element.pyx:828`:
```julia
    def __hash__(self):
        """
        Make hashable.

        EXAMPLES::

            sage: hash(libgap(123))   # random output
            163512108404620371
        """
        return hash(str(self))
```
But at the same time the [Python documentation for `__hash__` says this](https://docs.python.org/3.11/reference/datamodel.html#object.__hash__):
> The only required property is that objects which compare equal have the same hash value

But the above method violates that. E.g. there are GAP matrices which are equal but print quite differently.

A good fix is difficult. The only "easy" fix I can think of is to return a constant value. Any better fix requires a case distinction based on what kind of GAP object this is; and even then in general returning the constant is the best I can think of (e.g. if the object is a group)

mantepse:
Before this pull request, all `GapElement` objects, including rings, functions, etc. were hashable, based on their string representation.  Since this may lead to subtle bugs, we make `GapElements` hashable only if equality of the objects implies equality of their string representation.

Fixes #35926


mantepse:
Ping?

dimpase:
Could you explain in a human language how this works now? (Classes are tricky)

mantepse:
I'm not sure whether I understand the question, so please let me know if I misunderstood.

In current develop, all GAP objects inherit from the class `GapElement`, which is hashable.  In the proposal, we remove the `__hash__` implementation - which is simply the hash of `str(self)` - of the of `GapElement` and move it to a new class, `GapElement_with_hash`.

GAP objects which inherit from `GapElement_with_hash` should therefore have the property that the string representation of equal objects is the same.  This is the case for `GapElement_Integer`, `GapElement_FiniteField`, etc.

Other GAP objects, for example groups, should only inherit from `GapElement`.

I think that this change will only be noticable for developers, because `CategoryObject` still has a hash function, which returns the hash of `repr(self)`.  However, if I try to base my hash on a GAP object, or put GAP elements into a set, I will be warned:
```
sage: G = PermutationGroup([[(1,2)],[(1,3)]])
sage: H = SymmetricGroup(3)
sage: set([G.gap(), H.gap()])
Cell In[15], line 1
----> 1 set([G.gap(), H.gap()])

TypeError: unhashable type: 'sage.libs.gap.element.GapElement'
```

dimpase:
please rebase

PR #38867, opened 2024-10-27
Fix duplicate install of config.py in meson build

antonio-rojas:
The  `configure_file` command creating this `config.py` file already has `install: true` which makes it installed by default. Installing it again here creates a duplicate entry in wheel packages, which breaks installing them with some tools (eg. python-installer).


tobiasdiez (approved):
Thanks!

PR #38870, opened 2024-10-28
pkgconf to dependencies of libsemigroups

dimpase:
without it, libsemigroups will be useless for GAP's semigroups package

see [sage-release](https://groups.google.com/g/sage-release/c/tXlIIJ0HX_Y/m/ITFAkQv0BAAJ) for a discussion

### :memo: Checklist

### DIFF
--- a/build/pkgs/libsemigroups/dependencies
+++ b/build/pkgs/libsemigroups/dependencies
@@ -1,4 +1,3 @@
-# no dependencies
-
+pkgconf
 ----------
 All lines of this file are ignored except the first.


orlitzky:
I think pkgconf should be a (build only, after the pipe) dependency of gap_packages instead, right? The GAP semigroups package uses it to detect libsemigroups: https://github.com/semigroups/Semigroups/blob/main/m4/ax_check_libsemigroup.m4

libsemigroups, on the other hand, only uses pkgconf to detect things that sage disables anyway:

https://github.com/search?q=repo%3Alibsemigroups%2Flibsemigroups%20PKG_CHECK_MODULES&type=code

(I guess we could make this more explicit with `--disable-eigen --disable-hpcombi --disable-fmt`)

culler:
I don't think Eigen was disabled in the previous version of the libsemigroups package.  When I fixed the issue with TextFlow.hpp I got another error in the include statement in the following block of code from digraph.hpp:

```
#ifdef LIBSEMIGROUPS_EIGEN_ENABLED
#include <Eigen/Core>
#endif
```

I think that means that LIBSEMIGROUPS_EIGEN_ENABLED was defined in the previous version.

orlitzky:
The spkg-install for libsemigroups does,

```
cd src
sdh_configure --disable-eigen
sdh_make
sdh_make_install
```

so unless that's buggy somehow, it should be disabled (we don't have eigen in the sage distro).

culler:
I was able to successfully build libsemigroups and gap_packages after making these two changes.

1. Remove the --disable-eigen option, so that sage/build/pkgs/libsemigroups/spkg-install.in looks like:
```
cd src
sdh_configure
sdh_make
sdh_make_install
```

2. Add a post-install script sage/build/pkgs/libsemigroups/spkg-postinst.in containing:
```
sed -i '' 's|<Eigen/Core>|"Eigen/Core"|' ${SAGE_LOCAL}/include/libsemigroups/digraph.hpp
```

It should be possible to make the include statement in digraph.hpp work correctly by adding a -I option somewhere, but I don't know where to do that and this worked for me.

I also don't know what it means when @orlitzky says "we don't have eigen in the sage distro".

orlitzky:
> I also don't know what it means when @orlitzky says "we don't have eigen in the sage distro".

I meant that we don't have `build/pkgs/eigen`, which I took for even more evidence that `--disable-eigen` had to disable it. But now I see that it's bundled with libsemigroups, so I guess it's possible that `--disable-eigen` was not working somehow and that it used the bundled version.

dimpase:
Sorry, guys, this PR has little to do with #38875 - could you please review #38875 there, and not here.

This PR fixes a potential bug which might occur if one runs `make gap_packages` right after `./configure`.

dimpase:
And could you please review **this** PR?

dimpase:
> > I also don't know what it means when @orlitzky says "we don't have eigen in the sage distro".
> 
> I meant that we don't have `build/pkgs/eigen`, which I took for even more evidence that `--disable-eigen` had to disable it. But now I see that it's bundled with libsemigroups, so I guess it's possible that `--disable-eigen` was not working somehow and that it used the bundled version.

For me, everything builds fine after `make distclean` (even on macOS without any `Eigen` or `libsemigroup` installed anywhere, I think). I'll repeat this comment on #38875 and would like to continue there.

orlitzky:
> And could you please review **this** PR?

I was. I think this commit changes the wrong package:

* libsemigroups doesn't need pkgconf
* gap_packages does

dimpase:
> > And could you please review **this** PR?
> 
> I was. I think this commit changes the wrong package:
> 
>     * libsemigroups doesn't need pkgconf
> 
>     * gap_packages does

libsemigroups could not be used by its Sage consumer(s) without pkg-config. And (probably) it might fail tests.
There is also a danger that libsemigroups won't even install it .pc file if pkg-config is not available.

orlitzky:
> libsemigroups could not be used by its Sage consumer(s) without pkg-config.

Only the consumers that rely on pkg-config for detection need pkg-config. Others would be fine without it. The packages that need pkg-config should depend on pkg-config...

> And (probably) it might fail tests.

You can quickpkg up your dev-util/pkgconf, emerge -C it, and try :)


> There is also a danger that libsemigroups won't even install it .pc file if pkg-config is not available.

This is handled by automake which doesn't know anything about pkg-config:  https://github.com/libsemigroups/libsemigroups/blob/main/Makefile.am#L39

dimpase:
anyhow,  pkg-config/pkgconf will be in pre-reqs after #40204 is merged, so no dependencies need to be touched then.

PR #38873, opened 2024-10-28
src/sage/parallel/map_reduce.py: fixup start_workers() test case

orlitzky:
In commit a1818838 I made this test wait a bit longer (to be more robust under load) and added a "long time" to the corresponding start_workers() line. But, further down, there's another test that calls finish(), which needs the workers to be started.  Without a "long time" on the finish(), it can fail sans --long.

To fix it, we combine the two tests that call start_workers() and finish(), and then mark the whole thing "long time". The finish() is supposed to delete the workers created in setup_workers() as well, so there's not much point in trying to make the "long time" more fine-grained than that.

### DIFF
--- a/src/sage/parallel/map_reduce.py
+++ b/src/sage/parallel/map_reduce.py
@@ -1133,18 +1133,16 @@ class RESetMapReduce:
 
         TESTS::
 
+            sage: # long time
             sage: from sage.parallel.map_reduce import RESetMapReduce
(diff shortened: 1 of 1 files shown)


tornaria:
Lgtm, I will test later.

tobiasdiez:
Let's mark it as CI fix so that it gets applied to the other PRs (where the meson workflow fails due to this problem).

PR #38874, opened 2024-10-28
Remove imports from `gsl.all`

tobiasdiez:
I had troubles with importing `gsl` (on Windows) and tried to fix it by changing imports from `gsl.all` to more specific imports. Turns out the real problem was actually wrong `pkg-config` entry; but I thought these more specific imports might be helpful anyway and should (slightly) reduce the size of the built library.

### :memo: Checklist

### DIFF
--- a/src/sage/probability/probability_distribution.pyx
+++ b/src/sage/probability/probability_distribution.pyx
@@ -40,7 +40,8 @@ REFERENCES:
 # ****************************************************************************
 from cysignals.memory cimport sig_malloc, sig_free
 
-from sage.libs.gsl.all cimport *
+from sage.libs.gsl.rng cimport * 
+from sage.libs.gsl.random cimport *
 import sage.misc.prandom as random
 import sage.rings.real_double
 from sage.modules.free_module_element import vector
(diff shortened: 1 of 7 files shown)


src/sage/probability/probability_distribution.pyx
```diff
@@ -40,7 +40,8 @@ REFERENCES:
 # ****************************************************************************
 from cysignals.memory cimport sig_malloc, sig_free
 
-from sage.libs.gsl.all cimport *
+from sage.libs.gsl.rng cimport * 
+from sage.libs.gsl.random cimport *
 import sage.misc.prandom as random
```
kwankyu, line 45: Trailing whitespaces here?

kwankyu:
Otherwise lgtm.

kwankyu:
By the way, the B&T using Conda/Meson would be fixed soon? Please let Checks fail only for issues of the PR itself.

src/sage/probability/probability_distribution.pyx
```diff
@@ -40,7 +40,8 @@ REFERENCES:
 # ****************************************************************************
 from cysignals.memory cimport sig_malloc, sig_free
 
-from sage.libs.gsl.all cimport *
+from sage.libs.gsl.rng cimport *
```
tobiasdiez:
```suggestion
from sage.libs.gsl.rng cimport *
```

src/sage/probability/probability_distribution.pyx
tobiasdiez: Thanks fixed

kwankyu (approved):
LGTM.

PR #38882, opened 2024-10-29
src/sage: revert "touch libgap consumers"

orlitzky:
This was a junk commit to trigger a rebuild of certain parts of sagelib after a GAP upgrade.


fchapoton (approved):
damn.. Who let this in ?

orlitzky:
Me I suppose :)

We were on our third attempt to upgrade to GAP 4.13.1 with several months invested in tracking down bugs etc., and the window for the next sage release was closing... But the CI was failing because sagelib wouldn't rebuild these files after installing the new version of GAP. A better fix would have been to figure out the Makefile spaghetti, but that wastes a few hours every time you guess wrong (and I spent a few days guessing and rebuilding to no avail). This option is stupid but it got the job done quickly.

fchapoton:
ok, I understand. Thanks for cleaning.

PR #38883, opened 2024-10-29
provide the tilde species of a species

mantepse:
See https://mathoverflow.net/questions/480823/the-tilde-species

Dependencies: #38446


fchapoton (approved):
ok, fine

PR #38884, opened 2024-10-29
remove dead code in categories/rings

mantepse:
The removed line was marked as dead by codecov in #38821.  For safety, we remove it in a separate pull request.

Dependencies: #38821


mantepse:
Indeed, you are right.   Should we make this a doctest?
```
from sage.structure.element import Element

class MyRingElement(Element):
    """
    sage: R = MyRing()
    sage: R.ideal(R(1))
    """
    def __init__(self, parent, x):
        Element.__init__(self, parent)
        self._x = parent._K(x)

    def _add_(self, other):
        P = self.parent()
        return P.element_class(P, self._x + other._x)

    def _mul_(self, other):
        return P.element_class(P, self._x * other._x)

class MyRing(Parent):
    def __init__(self):
        R.<x,y> = PolynomialRing(QQ)
        Q = R.quotient(x^2+y^2+1)
        self._K = Q
        category = PrincipalIdealDomains()
        Parent.__init__(self, base=QQ, category=category)

    Element = MyRingElement
```

mantepse:
Should we (i.e., I) add this as a test to `categories.rings`?

fchapoton:
I am not so sure. Maybe we can think about that again once the big changes are merged ?

mantepse:
> I am not so sure. Maybe we can think about that again once the big changes are merged ?

Yes, of course!  Note that the tests show that the code is not so useful for principal ideal domains without gcd.

tscrim:
#38821 has now been merged I believe.

tscrim:
There are doctest failures.

PR #38892, opened 2024-10-31
graphs/matching_covered_graph: add `maximal_barrier()` and `canonical_partition()`

janmenjayap:
The objective of this issue is to implement the methods pertaining to the canonical partition of a matching covered graph. 

More specifically, this PR aims to implement the following two methods:
- [x] `maximal_barrier()`: Return the (unique) maximal barrier containing the provided vertex. 
- [x] `canonical_partition()`: Return the canonical partition of the (matching covered) graph.

This PR shall address the methods related to canonical partition of matching covered graphs.

Fixes #38216. Note that this issue fixes a small part of the mentioned issue.

### :memo: Checklist

### :hourglass: Dependencies
This PR depends on the PR #38742.

cc: @dcoudert.


src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2070 @@
...
+            True
+        """
+        return self._matching
+
+    @doc_index('Barriers and canonical partition')
+    def maximal_barrier(self, vertex):
+        r"""
+        Return the (unique) maximal barrier containing the vertex.
+
+        We use the following theorem.
```
dcoudert: Before explaining the theorem you use, add a like explaining what is a `barrier`

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2070 @@
...
+
+        G = Graph(self, multiedges=False)
+        M = Graph(self.get_matching())
+        B = set([vertex])
+
+        # u: The M neighbor of vertex
+        u = next(M.neighbor_iterator(vertex))
+        vertex_neighbors = []
+
+        for v in G.neighbor_iterator(vertex):
```
dcoudert: why don't you simply use `vertex_neighbors = G.neighbors(vertex)` ?

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2070 @@
...
+        # In other words, there exists an odd length M-alternating vertex-w
+        # path in G, starting and ending with edges in M. Alternatively, there
+        # exists an even length M-alternating u-w path in the graph G - vertex
+        # starting with an edge not in M and ending with and edge in M.
+
+        # even: The set of all such vertex w
+        from sage.graphs.matching import M_alternating_even_mark
+        even = M_alternating_even_mark(G=G, matching=M, vertex=u)
+
+        for v in G:
```
dcoudert: `B.update(v for v if G if v not in even)` ?

dcoudert (changes requested):

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2147 @@
...
+    :class:`~MatchingCoveredGraph`:
+
+    Overwritten Methods:
+
+    - ``delete_edge()`` | Delete the edge from ``u`` to ``v``.
+    - ``delete_edges()`` | Delete edges from an iterable container.
+
+    Barriers and canonical partition:
+
+    - ``canonical_partition()`` | Return the canonical partition of the
```
dcoudert: can be removed from the TODO list

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2147 @@
...
+
+        For a matching covered graph `G`, a subset `B` of the vertex set `V` is
+        a barrier if `|B| = o(G - B)`, where `|B|` denotes the cardinality of
+        the set `B` and `o(G - B)` denotes the number of odd components in the
+        graph `G - B`. And a barrier `B` is a maximal barrier if `C` is not a
+        barrier for each `C` such that `B \subset C \subseteq V`.
+
+        Note that in a matching covered graph, each vertex belongs to a unique
+        maximal barrier. The maximal barriers of a matching covered graph
+        partitons its vertex set and the partition of the vertex set of a
```
dcoudert: partitons -> partitions

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2147 @@
...
+            - :meth:`~sage.graphs.graph.Graph.is_matching_covered`
+            - :meth:`~sage.graphs.matching_covered_graph.MatchingCoveredGraph.maximal_barrier`
+        """
+        visited = set()
+
+        maximal_barriers = []
+        for v in self:
+            if v not in visited:
+                B = self.maximal_barrier(v)
+                for u in B:
```
dcoudert: use `visited.update(B)`

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2147 @@
...
+                    visited.add(u)
+                maximal_barriers.append(B)
+
+        return sorted(maximal_barriers, key=lambda s: min(s))
```
dcoudert: why are you sorting the result ? In general it is better to avoid sorting if not necessary. Indeed, vertex labels may be of different types.

dcoudert (changes requested):

dcoudert:
in the description of this PR, you say that it fixes #38216 but this is not the case. It is part of that issue, or a follow-up of that issue.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -0,0 +1,2141 @@
...
+
+            - :meth:`~sage.graphs.graph.Graph.is_bicritical`
+            - :meth:`~sage.graphs.graph.Graph.is_matching_covered`
+            - :meth:`~sage.graphs.matching_covered_graph.MatchingCoveredGraph.canonical_partition`
+        """
+        if vertex not in self:
+            raise ValueError('vertex {} not in the graph'.format(vertex))
+
+        G = Graph(self, multiedges=False)
+        M = Graph(self.get_matching())
```
dcoudert: I understand that it's easier to build the graph of the matching, but you don't need it. You could simply scan the edges to find one incident to `vertex` and then get its neighbor. Building a graph takes some time.

Note that method `M_alternating_even_mark` will also make a graph from the matching.

Similarly, I'm not sure you need to turn `self` into a `Graph`. method `M_alternating_even_mark` starts building a simple graph out of the input graph.

dcoudert (changes requested):
in the PR description, you must indicate that this PR depends on #38742

src/sage/graphs/matching_covered_graph.py
janmenjayap: I have made the necessary changes.
Thanks for pointing this out.

src/sage/graphs/matching_covered_graph.py
```diff
@@ -2033,12 +2033,9 @@ def maximal_barrier(self, vertex):
...
             raise ValueError('vertex {} not in the graph'.format(vertex))
 
-        G = Graph(self, multiedges=False)
-        M = Graph(self.get_matching())
-        B = set([vertex])
-
         # u: The M neighbor of vertex
-        u = next(M.neighbor_iterator(vertex))
+        matching = self.get_matching()
+        u = next((a if b == vertex else b) for a, b, *_ in matching if vertex in [a, b])
```
dcoudert, line 2160: why using `*_` ?  isn't `_` sufficient ?

janmenjayap: We might have an edge with just two parameters without a label. In such a case, we might get a `ValueError` if we use `for a, b, _ in matching`:

```
...
ValueError: not enough values to unpack (expected 3, got 2)
```

And the following one, if we use `for a, b in matching:` provided we have an edge with three entries:

```
...
ValueError: too many values to unpack (expected 2)
```

dcoudert: I didn't know that. Thank you for the explanation.

dcoudert (approved):
LGTM.

dcoudert (approved):
LGTM.

PR #38899, opened 2024-10-31
src/sage/interfaces/singular.py: use GNU Info to read Singular's info

### LINKED ISSUE #32242: Use GNU Info to parse singular's Info file
<div id="comment:0"></div>

Our Singular interface currently contains a hand-written parser for Singular's "info" file. In this ticket, we add an optional GNU Info package to parse this file instead.

Relative to running "info", the hand-written parser has several drawbacks:

* The extra code is a maintenance burden. We should not be wasting
  our time (poorly) reimplementing standard tools.

* The custom parser is buggy. For example, it is supposed to raise a
  `KeyError` when documentation for a non-existent function is
  requested. However, the parser does not keep track of what section
  it's in, so, for example, `get_docstring("Preface")` will happily
  return the contents of the preface even though "Preface" is not a
  Singular function.

* The first time documentation is requested, the entire info file
  (see above) is loaded into a dictionary. With the Singular info
  file currently weighing in at around 11MiB, this is wasting a
  good chunk of memory.

* The `singular_console()` itself tries to launch GNU Info to display
  its interactive help, and cannot use our hand-written parser.

The one downside to requiring "info" is that it would add a new runtime dependency to sagelib. To avoid that, we make GNU Info optional, and suggest to the user that he install it when he tries to read Singular's documentation.

GNU Info (or its superset, Texinfo) are widespread, portable, and easy to install on all of the systems we support, so in most cases this should be a "free" improvement.

Depends on #32302

Component: **packages: standard**

Author: **Michael Orlitzky**

Branch/Commit: **[u/mjo/ticket/32242](https://github.com/sagemath/sagetrac-mirror/tree/u/mjo/ticket/32242) @ [`137a918`](https://github.com/sagemath/sagetrac-mirror/commit/137a9182a02d4aa329a78c2e936b48431c1e9017)**

_Issue created by migration from https://trac.sagemath.org/ticket/32242_

orlitzky:
Our Singular interface currently contains a hand-written parser for Singular's "info" file. This commit eliminates the custom parser in favor of launching GNU Info. GNU Info (or its superset, Texinfo) are widespread, portable, and easy to install on all of the systems we support, so in most cases this should be a "free" improvement.

The hand-written parser has several drawbacks:

* The extra code is a maintenance burden. We should not be wasting our time reimplementing standard tools.
* The custom parser is buggy. For example, it is supposed to raise a KeyError when documentation for a non-existent function is requested. However, the parser does not keep track of what section it's in, so, for example, get_docstring("Preface") returns the contents of the preface even though "Preface" is not a Singular function.
* The first time documentation is requested, the entire info file is loaded into a dictionary. This wastes a few megabytes of memory for the duration of the Sage session.
* The custom parser does not handle compression (GNU Info does transparently), and the end user or people packaging Singular may not be aware of that. If the system installation of Singular has a compressed info file, Sage won't be able to read it.

For contrast, the one downside to using GNU Info is that it adds a new runtime dependency to sagelib. To mitigate that, we do not technically require it, and instead raise a warning if the user (a) tries to read the Singular documentation and (b) has managed to find a system without GNU Info. Our singular_console() itself tries to launch GNU Info to display its interactive help, so the additional optional dependency is not so additional except in corner cases, such as a pypi installation of a subset of Sage linked against libsingular but without a full Singular install.

This was originally https://github.com/sagemath/sage/issues/32242

Since then,

* GNU Info has become a standard package in the sage distribution
* I'm now using `sage.features.info` to detect the "info" program, so nothing bad happens if it's not installed

Fixes https://github.com/sagemath/sage/issues/32242

### DIFF
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -2335,12 +2335,12 @@ def get_docstring(name, prefix=False, code=False):
     INPUT:
 
     - ``name`` -- a Singular function name
-    - ``prefix`` -- boolean (default: False); whether or not to include
-      the prefix stating that what follows is from the Singular
-      documentation.
-    - ``code`` -- boolean (default: False); whether or not to format the
-      result as a reStructuredText code block. This is intended to support
-      the feature requested in :issue:`11268`.
+    - ``prefix`` -- boolean (default: ``False``); whether or not to
+      include the prefix stating that what follows is from the
+      Singular documentation.
+    - ``code`` -- boolean (default: ``False``); whether or not to
+      format the result as a reStructuredText code block. This is
+      intended to support the feature requested in :issue:`11268`.
 
     OUTPUT:
 
--- a/src/sage/interfaces/singular.py
+++ b/src/sage/interfaces/singular.py
@@ -2269,11 +2269,9 @@ class SingularFunction(ExpectFunction):
         """
         EXAMPLES::
 
-            sage: 'groebner' in singular.groebner.__doc__
+            sage: 'groebner' in singular.groebner.__doc__  # needs info
             True
         """
-        if not nodes:
-            generate_docstring_dictionary()
 
         prefix = """
 This function is an automatically generated pexpect wrapper around the Singular
@@ -2288,15 +2286,9 @@ EXAMPLES::
     x+y,
     y^2-y
 """ % (self._name,)
-        prefix2 = """
-
-The Singular documentation for '%s' is given below.
-""" % (self._name,)
-
-        try:
-            return prefix + prefix2 + nodes[node_names[self._name]]
-        except KeyError:
-            return prefix
+        return prefix + get_docstring(self._name,
+                                      prefix=True,
+                                      code=True)
 
 
 @instancedoc
@@ -2307,15 +2299,10 @@ class SingularFunctionElement(FunctionElement):
 
             sage: R = singular.ring(0, '(x,y,z)', 'dp')
             sage: A = singular.matrix(2,2)
-            sage: 'matrix_expression' in A.nrows.__doc__
+            sage: 'matrix_expression' in A.nrows.__doc__  # needs info
             True
         """
-        if not nodes:
-            generate_docstring_dictionary()
-        try:
-            return nodes[node_names[self._name]]
-        except KeyError:
-            return ""
+        return get_docstring(self._name, code=True)
 
 
 def is_SingularElement(x):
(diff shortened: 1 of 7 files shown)


src/sage/interfaces/singular.py
```diff
@@ -2408,15 +2350,65 @@ def get_docstring(name):
...
+
     """
-    if not nodes:
-        generate_docstring_dictionary()
+    from sage.features.info import Info
+
+    if not Info().is_present():
+        raise OSError("Error: GNU Info is not installed. Singular's "
+                      "documentation will not be available.")
+    import subprocess
```
kwankyu, line 2405: just
```
OSError("GNU Info is not installed. Singular's "
```
?

kwankyu:
Is this the expected result for unknown function?

image

kwankyu:
image

```
>>`>>
```
seems broken parts... No?

kwankyu:
Perhaps I am not testing your branch on a proper platform... How can I invoke the `info` interface like
image
within sage?

src/sage/interfaces/singular.py
orlitzky: You're right, I copied that error message from some FFMpeg but there's no reason to say it again. Fixed.

orlitzky:
You should just be able to run `info` if it's already installed on your system. To browse the singular help, you should be able to run `info singular.info`.

Can you please run `info --version` for me? I think your output may be different due to an older version. Sage comes with a new version but it will also accept an old version from the system, and there is an important bugfix that I am relying on to keep the parsing simple.

orlitzky:
> Is this the expected result for unknown function?

No, but I think it may be the case with info 6.x. If so, I'll need to handle that.

orlitzky:
I updated both the feature and the `spkg-configure.m4` to check for v7.0 or greater. I think that should fix all of the issues you're seeing. With v6.x, info would return "success" even if the section you want to read does not exist.

kwankyu:
> You should just be able to run `info` if it's already installed on your system. To browse the singular help, you should be able to run `info singular.info`.
> 
> Can you please run `info --version` for me? I think your output may be different due to an older version. Sage comes with a new version but it will also accept an old version from the system, and there is an important bugfix that I am relying on to keep the parsing simple.

```bash
$ sage -sh

Starting subshell with Sage environment variables set.  Don't forget
to exit when you are done.  Beware:
 * Do not do anything with other copies of Sage on your system.
 * Do not use this for installing Sage packages using "sage -i" or for
   running "make" at Sage's root directory.  These should be done
   outside the Sage shell.

Bypassing shell configuration files...

Note: SAGE_ROOT=/home/kwankyu/GitHub/sage-dev
(sage-sh) kwankyu@HOME:sage-dev$ info --version
info (GNU texinfo) 7.1

Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
(sage-sh) kwankyu@HOME:sage-dev$ which info
/usr/bin/info
```
I am testing on Ubuntu on WSL

There is nothing changed with your updated branch.

I will force sage to build the info spkg...

kwankyu:
```sage
sage: !info --version
info (GNU texinfo) 7.0.3

Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
sage: from sage.interfaces.singular import get_docstring
sage: get_docstring('Preface')
'\n1 Preface\n*********\n\n                        SINGULAR version 4.4.0\n                  University of Kaiserslautern-Landau\n      Department of Mathematics and  Centre for Computer Algebra\n     Authors: W. Decker, G.-M. Greuel, G. Pfister, H. Schoenemann\n\n                        Copyright (C) 1986-2024\n\n\n                               *NOTICE*\n\nThis program is free software; you can redistribute it and/or modify it\nunder the terms of the GNU General Public License as published by the\nFree Software Foundation (version 2 or version 3 of the License).\n\nSome single files have a copyright given within the file:\nSingular/links/ndbm.* (BSD)\n\nThe following software modules shipped with SINGULAR have their own\ncopyright: the omalloc library, the readline library, the GNU Multiple\nPrecision Library (GMP), NTL: A Library for doing Number Theory (NTL),\nFlint: Fast Library for Number Theory, the Singular-Factory library, the\nSingular-Factory library, the Singular-libfac library, surfex, and, for\nthe Windows distributions, the Cygwin DLL and the Cygwin tools\n(Cygwin), and the XEmacs editor (XEmacs).\n\nTheir copyrights and licenses can be found in the accompanying files\nCOPYING which are distributed along with these packages.  (Since\nversion 3-0-3 of SINGULAR, al
...
sage: singular.aliggggg?
Signature:      singular.aliggggg(*args, **kwds)
Type:           SingularFunction
String form:    aliggggg
File:           ~/GitHub/sage-dev/src/sage/interfaces/singular.py
Docstring:
This function is an automatically generated pexpect wrapper around the
Singular function 'aliggggg'.

EXAMPLES:

   sage: groebner = singular.groebner
   sage: P.<x, y> = PolynomialRing(QQ)
   sage: I = P.ideal(x^2-y, y+x)
   sage: groebner(singular(I))
   x+y,
   y^2-y
Init docstring:
WARNING: the enclosing module is marked 'needs sage.libs.gap sage.libs.pari sage.libs.singular sage.symbolic',
so doctests may not pass.
Call docstring:
WARNING: the enclosing module is marked 'needs sage.libs.gap sage.libs.pari sage.libs.singular sage.symbolic',
so doctests may not pass.
```
Nothing changed....

orlitzky:
Argh, OK, I see the problem. When I updated the PR to trust the exit code from `info`, that accidentally allowed it to match things like "Preface" again. I'll take another shot at it tomorrow, thank you.

orlitzky:
I've added back the check for the section header to ensure that `get_docstring("Preface")` fails, and I've added a doctest for that now.

I also tried to fix the `singular.nonexistent?` problem. One of the functions in `sage.misc.sageinspect` was ignoring all exceptions that arise inside the question-mark help. I think the motivation for that is obsolete, so I've removed it, but we'll have to wait for the documentation build to complete to be sure.

In any case, `singular.nonexistent?` now raises an error immediately instead of showing you junk. Likewise, running

```
sage: from sage.libs.singular.function import singular_function
sage: f = singular_function("nonexistent")
sage: f?
```

should raise an error.

orlitzky:
It's looking like it might be OK now:

* The section header parsing that was required with GNU info < 7.0 is back because that's what allows `get_docstring("Preface")` to be caught and fail.
* As a result, there's no point in requiring info v7. It will accept v6 again.
* I removed the exception masking in sage's `help?`. This fixes the docs for nonexistent singular functions (they now raise an error).
* I had to make a few singular function variables in `src.sage.algebras.letterplace` local so that the docbuild wouldn't try to document them, but afterwards there doesn't seem to be any problem building the docs.
* I rebased and squashed everything because there's a new develop branch and I had to force push anyway.

src/sage/interfaces/singular.py
```diff
@@ -2341,82 +2328,125 @@ def is_SingularElement(x):
...
-    nodes.clear()
-    node_names.clear()
+    INPUT:
 
-    new_node = re.compile(r"File: singular\.[a-z]*,  Node: ([^,]*),.*")
-    new_lookup = re.compile(r"\* ([^:]*):*([^.]*)\..*")
+    - ``name`` -- a Singular function name
+    - ``prefix`` -- boolean (default: False); whether or not to include
+      the prefix stating that what follows is from the Singular
+      documentation.
```
kwankyu:
```
``False``
```
here and below.

kwankyu:
Thanks. It now works nicely, both with spkg info and system info.

One anomaly I noticed: for this input
```
sage: singular.ali?
```
I get
```
Signature:      singular.ali(*args, **kwds)
Type:           SingularFunction
String form:    ali
File:           ~/GitHub/sage-dev/src/sage/interfaces/singular.py
Docstring:
This function is an automatically generated pexpect wrapper around the
Singular function 'ali'.

EXAMPLES:

   sage: groebner = singular.groebner
   sage: P.<x, y> = PolynomialRing(QQ)
   sage: I = P.ideal(x^2-y, y+x)
   sage: groebner(singular(I))
   x+y,
   y^2-y

The Singular documentation for "ali" is given below.

   5.1.1 align
   -----------

   `*Syntax:*'
        `align (' vector_expression, int_expression `)'
        `align (' module_expression, int_expression `)'

   `*Type:*'
        type of the first argument
```
Completing "ali" to "align" is nice, but I think the documentation should show "align" instead of "ali". 

But to be fair, this anomaly is not new. Let me know if you would fix it or leave it.

kwankyu:
Some strange thing is happening in documentation:

image

https://doc-pr-38899--sagemath.netlify.app/html/en/reference/interfaces/sage/interfaces/singular#hunk2

src/sage/interfaces/singular.py
orlitzky: fixed, thanks

orlitzky:
> Completing "ali" to "align" is nice, but I think the documentation should show "align" instead of "ali".
> 
> But to be fair, this anomaly is not new. Let me know if you would fix it or leave it.

I don't think this will be easy to fix. The problem is that `singular.ali` is a "valid" singular function object created on-the-fly by the pexpect interface. You can do the same thing with e.g. `gap.ali?`. It will create a `GapFunction` called `gap.ali`, and then return whatever the docstring is for it.

The thing that I _could_ fix here is that `ali?` will load the help for `align` and not insist on an exact match for `ali`. But maybe it is nicer to return the nearest match, I don't know. It would be great if `singular.ali<tab>` worked too but for some reason "align" isn't in the list of tab completions.

orlitzky:
This is probably something a singular expert should look in to, but I think the names of the built-in singular functions can be obtained from the list returned by `reservedName();` after filtering out the data types and other non-functions.

orlitzky:
> Some strange thing is happening in documentation:

That was literally the docstring for `singular_version()`. Probably a copy/paste mistake. I've fixed it.

kwankyu:
> > Completing "ali" to "align" is nice, but I think the documentation should show "align" instead of "ali".
> > But to be fair, this anomaly is not new. Let me know if you would fix it or leave it.
> 
> I don't think this will be easy to fix. 

OK.

kwankyu (approved):
LGTM now. Thanks.

kwankyu:
Would you add "Fixes #32242" in the PR description?

orlitzky:
Done, thank you for the careful review

PR #38901, opened 2024-11-01
add "# long time" to slow-ish composite isogeny doctest

yyyyx4:
...to silence the new(?) warning about tests that take a long time.


orlitzky:
Thanks, this takes about 30s on my machine. The warnings aren't new but the threshold has been lowered and the "how long did it take" computation is much more reliable now.

PR #38902, opened 2024-11-01
division of elliptic-curve morphisms

yyyyx4:
This patch adds a class for representing symbolic fractions $\phi/d$ of elliptic-curve morphisms. As a corollary, we get left- and right-division of isogenies, which are emulated by composing with the dual and dividing out the degree. Fractional isogenies often appear in computations with endomorphism rings (e.g., $(1+\pi_q)/2$ is an endomorphism of any elliptic curve over $\mathbb F_q$ with rational $2$-torsion).

The new class is designed to fully support the `EllipticCurveHom` interface, albeit with *extremely* basic algorithms in some cases. Indeed, the goal of this patch is to establish the interface and basic functionality — any optimizations are left for future work.

### DIFF
--- a/src/sage/schemes/elliptic_curves/hom.py
+++ b/src/sage/schemes/elliptic_curves/hom.py
@@ -125,15 +125,15 @@ class EllipticCurveHom(Morphism):
             raise TypeError(f'cannot compose {type(self)} with {type(other)}')
 
         ret = self._composition_impl(self, other)
-        if ret is not NotImplemented:
-            return ret
 
-        ret = other._composition_impl(self, other)
-        if ret is not NotImplemented:
-            return ret
+        if ret is NotImplemented:
+            ret = other._composition_impl(self, other)
 
-        from sage.schemes.elliptic_curves.hom_composite import EllipticCurveHom_composite
-        return EllipticCurveHom_composite.from_factors([other, self])
+        if ret is NotImplemented:
+            from sage.schemes.elliptic_curves.hom_composite import EllipticCurveHom_composite
+            ret = EllipticCurveHom_composite.from_factors([other, self])
+
+        return ret
 
     def _add_(self, other):
         r"""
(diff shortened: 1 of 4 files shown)


src/sage/schemes/elliptic_curves/hom.py
```diff
@@ -1130,6 +1143,91 @@ def matrix_on_subgroup(self, domain_gens, codomain_gens=None):
...
+                To:   Elliptic Curve defined by y^2 = x^3 + 418*x over Finite Field of size 419
+                Via:  (Frobenius endomorphism of degree 419:
+                         From: Elliptic Curve defined by y^2 = x^3 + 418*x over Finite Field of size 419
+                         To:   Elliptic Curve defined by y^2 = x^3 + 418*x over Finite Field of size 419,
+                       Scalar-multiplication endomorphism [1]
+                         of Elliptic Curve defined by y^2 = x^3 + 418*x over Finite Field of size 419)
+              Denominator: 2
+        """
+        from sage.rings.integer import Integer
+        if not isinstance(other, (int, Integer)):
```
GiacomoPope, line 1170: Is it worth trying to cast the value here to an Integer and return NotImplemented when we cant? 

I'm thinking of the case where somehow `2` is an element of the rationals or finite field and then this fails somewhat silently

src/sage/schemes/elliptic_curves/hom_fractional.py
```diff
@@ -0,0 +1,544 @@
...
+      Numerator:   Composite morphism of degree 245 = 35*7:
+      From: Elliptic Curve defined by y^2 = x^3 + x over Finite Field of size 419
+      To:   Elliptic Curve defined by y^2 = x^3 + 269*x + 82 over Finite Field of size 419
+      Denominator: 7
+    sage: phi == psi * chi
+    True
+
+AUTHORS:
+
+- Lorenz Panny (2024)
```
GiacomoPope, line 86: 2025?

src/sage/schemes/elliptic_curves/hom_fractional.py
```diff
@@ -0,0 +1,544 @@
...
+            1
+            sage: E = EllipticCurve(GF(419), [-1,0])
+            sage: pi = E.frobenius_isogeny()
+            sage: ((3*pi - pi) / 2).inseparable_degree()
+            419
+        """
+        return self._phi.inseparable_degree() / self._domain.scalar_multiplication(self._d).inseparable_degree()
+
+
+def _torsion_gens(E, EE, l, e, psi=None):
```
GiacomoPope, line 531: is there not something appropriate we can use from the curve object itself here?

GiacomoPope (commented):
Some initial comments while reading, but will read more later when i have time. General thought is that this is fantastic and a welcome addition to Sage

GiacomoPope (approved):

PR #38903, opened 2024-11-01
a few details in combinat, following ruff and pycodestyle

fchapoton:
just fixing a few ruff PERF suggestions in 4 files

### :memo: Checklist

### DIFF
--- a/src/sage/combinat/rigged_configurations/kleber_tree.py
+++ b/src/sage/combinat/rigged_configurations/kleber_tree.py
@@ -123,13 +123,10 @@ def _draw_tree(tree_node, node_label=True, style_point=None, style_node='fill=wh
     if rpos is None:
         rpos = [0., 0.]
 
-    def draw_point(point):
-        return '(%.3f, %.3f)' % (point[0], point[1])
-
     if not tree_node.children:
         r = ''
         node_name = node_prefix + str(node_id)
-        r = "\\node (%s) at %s" % (node_name, draw_point(start))
+        r = "\\node (%s) at (%.3f, %.3f)" % (node_name, *start)
         if node_label:
             r += "{$%s$};\n" % tree_node._latex_()
         else:
@@ -187,7 +184,7 @@ def _draw_tree(tree_node, node_label=True, style_point=None, style_node='fill=wh
     rpos[0] = pos[0]
     rpos[1] = pos[1]
     point_str = ''
-    node_str = "\\node%s (%s) at %s" % (style_node, node_name, draw_point(pos))
+    node_str = "\\node%s (%s) at (%.3f, %.3f)" % (style_node, node_name, *pos)
     if node_label:
         node_str += "{$%s$};\n" % tree_node._latex_()
     else:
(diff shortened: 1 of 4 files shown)


src/sage/combinat/tableau.py
```diff
@@ -8183,25 +8182,20 @@ def random_element(self):
...
         # Get the cells in the Young diagram
-        cells = []
-        for i in range(len(p)):
-            for j in range(p[i]):
-                cells.append((i, j))
+        cells = [(i, j) for i in range(len(p)) for j in range(p[i])]
 
         m = sum(p)
-        while m > 0:
+        while m:
```
vincentmacri, line 8188: Unless this should be `while m != 0`, I think `m > 0` is more readable.

src/sage/combinat/words/finite_word.py
```diff
@@ -5675,7 +5674,7 @@ def abelian_vectors(self, n):
         size = alphabet.cardinality()
         if size == float('inf'):
             raise TypeError("The alphabet of the parent is infinite; define"
```
vincentmacri, line 5676: I don't expect you to do anything about this in this PR because it would be out of scope, but it would be nice to standardize either using `'` or `"` for strings throughout Sage.

src/sage/combinat/rigged_configurations/kleber_tree.py
```diff
@@ -122,7 +122,10 @@ def _draw_tree(tree_node, node_label=True, style_point=None, style_node='fill=wh
         start = [0., 0.]
     if rpos is None:
         rpos = [0., 0.]
-    draw_point = lambda point: '(%.3f, %.3f)' % (point[0],point[1])
+
+    def draw_point(point):
+        return '(%.3f, %.3f)' % (point[0], point[1])
```
vincentmacri: Unless I'm missing something, it looks like `draw_point` is only used once? Can't this string formatting just be done in that line? I don't think this one line function that's used in a single place is useful for readability. And at worst it adds (marginal) overhead having to call a separate function for a simple string formatting in a line where we're already doing string formatting.

vincentmacri (changes requested):
A couple small things. Looks good otherwise!

src/sage/combinat/tableau.py
fchapoton: Well,  the boolean test is supposed to be faster.

src/sage/combinat/words/finite_word.py
fchapoton: maybe, but this is an herculean task

src/sage/combinat/tableau.py
vincentmacri: Fair enough, as long as `m` can never be negative (which I think is true here so that's fine).

src/sage/combinat/rigged_configurations/kleber_tree.py
fchapoton: done

src/sage/combinat/words/finite_word.py
vincentmacri: I think it could be done with auto-formatting, but yeah it would be really big.

vincentmacri (approved):
LGTM! I'll set it to positive review once the CI run is finished unless there is a failure that looks relevant.

mantepse (approved):
Bots are happy, me too!

src/sage/combinat/words/finite_word.py
tscrim: I am *very strongly* -1 on making that uniform. Let the authors do what they want. Python allows it.

tscrim (approved):
LGTM.

Issue #38737, opened 2024-09-29
Segfault in plural


mantepse:
### Steps To Reproduce

1. Ubuntu 22.04.5 LTS
2. SageMath version 10.5.beta5, Release Date: 2024-09-22, develop branch compiled from source
3. `sage -t src/sage/rings/polynomial/plural.pyx`
4. segfault

### Expected Behavior

all tests pass

### Actual Behavior

[bug.txt](https://github.com/user-attachments/files/17178705/bug.txt)

### Additional Information

Might be related to #29528, but the segfault occured every time I ran the test.

### Environment

- **OS**: Ubuntu 22.04.5 LTS
- **Sage Version**: 10.5.beta5, Release Date: 2024-09-22, develop branch compiled from source

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

fchapoton:
I have that on all machines running gcc13

dimpase:
no problem on Fedora with gcc 14, and on macOS with its usual clang toolchain

fchapoton:
I can see that segfault on Ubuntu 22.04.4 LTS running gcc 1.4.0

tobiasdiez:
Stacktrace:
```
sage -t --warn-long 5.0 --random-seed=121929766104423602556012623849278420179 src/sage/rings/polynomial/plural.pyx
    SignalError in doctesting framework
**********************************************************************

Tests run before exception:
sage: A.<x,y,z> = FreeAlgebra(QQ, 3) ## line 50 ##
sage: P.<x,y,z> = A.g_algebra(relations={y*x:-x*y}, order = 'lex') ## line 51 ##
sage: P ## line 53 ##
Noncommutative Multivariate Polynomial Ring in x, y, z over Rational Field, nc-relations: {y*x: -x*y}
sage: y*x + 1/2 ## line 56 ##
-x*y + 1/2
sage: A.<x,y,z> = FreeAlgebra(GF(17), 3) ## line 59 ##
sage: P.<x,y,z> = A.g_algebra(relations={y*x:-x*y}, order = 'lex') ## line 60 ##
sage: P ## line 61 ##
Noncommutative Multivariate Polynomial Ring in x, y, z over Finite Field of size 17, nc-relations: {y*x: -x*y}
sage: y*x + 7 ## line 64 ##
-x*y + 7
sage: from sage.matrix.constructor import Matrix ## line 71 ##
[...]

PR #38904, opened 2024-11-01
provide a class for partitions with bounded length and minimal part

### LINKED ISSUE #38897: Computing `Partitions(...).cardinality()` with set `min_length=` can be significantly improved by complementation and using `max_length=` instead
### Problem Description

Check this out:
```
sage: %time Partitions(50,min_length=10).cardinality()
CPU times: user 1min 46s, sys: 14 ms, total: 1min 46s
Wall time: 1min 46s
158414

sage: %time number_of_partitions(50) - Partitions(50,max_length=9).cardinality()
CPU times: user 6.98 s, sys: 5 ms, total: 6.98 s
Wall time: 6.98 s
158414
```

### Proposed Solution

As the code example shows, computing cardinality with `min_length=` appears to be much slower than that with `max_length=`. Hence, it'd be beneficial to convert one into the other by complementation (in the set of all partitions) as illustrated by the example.

### Alternatives Considered

I believe the cardinalities in both cases can be computed via dynamic programming, but I did not check if Sage uses it wisely and why in one case it's much slower than in the other.

### Additional Information

_No response_

mantepse:
We provide a new class, `PartitionsSmallestGE` to handle subsets of the constraints `length`, `min_length`, `max_length` and `min_part`, to make cardinality computations with such constraints reasonable.

Fixes #38897, #39107


mantepse:
@fchapoton, could you please have a look so it doesn't rot away?

fchapoton (approved):
ok. Please avoid doing a lot of cleanup work in this kind of ticket.

mantepse:
Thank you!

mantepse:
@fchapoton, I would like to put this back to "needs work", because it doesn't quite solve Max' issue, and it's not hard to do that, but I'd like to avoid deprecation of something I just introduced.

Is this OK with you or would you prefer if I open a new pull request?

Ideally, I think there should be a single class for partitions of `n` with parts in a given set or interval and length in a given interval, which computes the cardinality efficiently using a few case distinctions and a recurrence if necessary.

mantepse:
I find it a bit frustrating that we do not need the `IntegerListsLex.__init__` at all to compute the cardinality.  It would be nice to have a pattern where such an expensive initialisation is postponed until it's needed.

Failing that, it seems that most of the time is spent in `Envelope.__init__`.  Maybe this can be optimized.

mantepse:
The solution was very simple, so I implemented it.

Max' problem is still much better solved without using `Partitions` (it seems to take roughly twice as long), because `Partitions` has `UniqueRepresentation`, which is expensive if we call it with very many different arguments - we are creating an entry in the cache not only for each size, but for each restriction.

mantepse:
@tscrim, there seems to be a problem with `FockSpace`.  In `FockSpaceTruncated.F.__init__`, the index set is set to be partitions of `F._n`, but that doesn't seem to be quite true.  Could you please check?

https://github.com/sagemath/sage/blob/39ebbe45ca5cdf87d8ae23a28b3f784130be3d1a/src/sage/algebras/quantum_groups/fock_space.py#L1734-L1759

tscrim:
> @tscrim, there seems to be a problem with `FockSpace`. In `FockSpaceTruncated.F.__init__`, the index set is set to be partitions of `F._n`, but that doesn't seem to be quite true. Could you please check?

Yes, that is not correct. It should be over all partitions of length `\leq k` (as per the doc of `FockSpaceTruncated`).

mantepse:
> Yes, that is not correct. It should be over all partitions of length `\leq k` (as per the doc of `FockSpaceTruncated`).

There is currently no class for partitions of

mantepse:
@fchapoton, is the lint error in error?

Other than that, I think that the pull request should be ready to go.

fchapoton:
somebody should add PLC0206 to the ruff-minimal ignore list in another pull request

fchapoton:
Is this definitely good to go, or not yet ? I am quite annoyed to have to check once again all the cosmetic indentation changes in the code.

mantepse:
Yes, this is ready.  There shouldn't be any cosmetic changes in the new commits, but if you prefer, I could create a separate pull request containing only the cosmetic changes from d53dccdb5210fdc129cf7875e757aa42dc430e03, and postpone this one.

There are two other things you may want me to change:

* have a better name for the class `Partitions_parts_length_restricted`
* deprecate `PartitionsGreatestLE` (which is entirely superseded by `Partitions_parts_length_restricted`)

The reason for not generalising `PartitionsGreatestLE` to accept more parameters is that the order of the parameters would be completely backwards.

tscrim:
I feel like there should be an easier way to get `number_of_partitions_length_max_part(n,k,b)` than recursively. Indeed the first step is definitely to strip off the first box in each row, and so now you have to count the number of partitions of `n-k`with max part `b-1` and length at most `k`. So you're asking for the coefficient of `q^{n-k}` in the `q`-binomial coefficient `\binom{b-1}{k}_q`. I feel like just storing that information (possibly computing it by recursion) would be faster. I think in general, you are computing all of that information anyways in the recursion.

mantepse:
> I feel like there should be an easier way to get `number_of_partitions_length_max_part(n,k,b)` than recursively. Indeed the first step is definitely to strip off the first box in each row, and so now you have to count the number of partitions of `n-k`with max part `b-1` and length at most `k`. So you're asking for the coefficient of `q^{n-k}` in the `q`-binomial coefficient `\binom{b-1}{k}_q`.

Yes, that's essentially correct (I get `[q^{n-k}]\binom{k+b-1}{k}_q`).  I should absolutely add it to the documentation.  Indeed, it also seems that there is room for improvement.  The worst case for the recurrence appears to be around small `k` and `b ~ n-k`.  Notably, in this case, the `q`-binomial is extremely fast.  On the other hand, the worst case for the `q`-binomial seems to be around `k = b = n` (which is, of course, silly).  I'm afraid that, if `k+b-1` is very large, we will get into trouble because of the size of the polynomials involved.  Hm, but in total, the performance of the `q`_binomial seems to be better indeed.

Apart from that, for the original issue we spend most of the time in constructing parents.  So, maybe give `number_of_partitions_length_max_part` a nicer name, and also let it handle `min_part` and perhaps `min_length` and `max_length`, so that it can be used also directly?  The function `number_of_partitions` contains the following documentation, though:
```
    The options of :meth:`number_of_partitions()` are being deprecated
    :issue:`13072` in favour of :meth:`Partitions_n.cardinality()` so that
    :meth:`number_of_partitions()` can become a stripped down version of
    the fastest algorithm available (currently this is using FLINT).
```

The most natural thing might be to give `number_of_partitions_length` two optional arguments, `min_part` and `max_part`, what do you think?  Although, with the `q`_binomial approach, including `max_length` becomes just as natural.

tscrim:
You're right about the q-binomial coefficient; that was a typo from me. I don't expect memory usage to be the big barrier; I think it will always be CPU time. There might be some tipping point to doing it recursively versus using the polynomials, but I wouldn't worry too much about finding it.

I'm not quite sure what you mean by the time was spent in constructing the parents. There is overhead for constructing the parents, but the cache lookup should basically be the same as for `@cached_function`... It's a little annoying to have a separate cache that has the same information that the parents would have...

For such cardinality-computing functions, it would be better to not use options and have different functions for the different input types. In essence, that is what you're doing with those options anyways.

mantepse:
Here is a plot indicating the relative performance for `n=600` and `k` and `b` in steps of `50`.  Red means that the `q`-binomial is faster.  The break-even seems to be around `k = b = 350`.

performance

I admit that I do not quite understand why the `q`-binomial is faster, but I shouldn't care too much, I guess.

mantepse:
Finally, given the good performance of `q`-binomial, I guess we should also take advantage of the fact that we can use it to compute the number of partitions of bounded length, rather than just the number of partitions of fixed length.  But it is unclear to me how to expose all this in the best way to the user - which functions with which names do we want to have?

mantepse:
It's annoying: I have re-implemented `cardinality` using `q`-binomial, and now Max' example takes forever, again.

mantepse:
I don't get it.  Individually, `q`-binomial seems to be fast, but not in this loop?  Here is a stripped down version.  On my laptop, the `D(100)` takes 750 milliseconds, whereas `D2(100)` takes 22 seconds.
```
"""
    sage: def D(n):
    ....:     s1 = number_of_partitions(n)
    ....:     s2 = sum(Q(m, 0, l, k, k)
    ....:              * Q(n-m-l^2, 0, Infinity, k+2*l, Infinity)
    ....:              for l in range(1, (n+1).isqrt())
    ....:              for m in range((n-l^2-2*l)*l//(l+1)+1)
    ....:              for k in range(ceil(m/l), min(m, n-m-l^2-2*l)+1))
    ....:     return s1 + s2
    sage: D(10)
    102
    sage: %time D(100)

    sage: def D2(n):
    ....:     s1 = number_of_partitions(n)
    ....:     s2 = sum(Q2(m, 0, l, k, k)
    ....:              * Q2(n-m-l^2, 0, Infinity, k+2*l, Infinity)
    ....:              for l in range(1, (n+1).isqrt())
    ....:              for m in range((n-l^2-2*l)*l//(l+1)+1)
    ....:              for k in range(ceil(m/l), min(m, n-m-l^2-2*l)+1))
    ....:     return s1 + s2
    sage: D2(10)
    102
    sage: %time D2(100)
"""
from sage.combinat.partition import number_of_partitions_length

@cached_function
def number_of_partitions_max_length_max_part(n, k, b):
    r"""
    Return the number of partitions of `n` with at most `k`
    parts, all of which are at most `b`.
    """
    return q_binomial(k + b, b)[n]

@cached_function
def number_of_partitions_length_max_part(n, k, b):
    r"""
    Return the number of partitions of `n` with exactly `k` parts and
    the largest part at most `b`.
    """
    if not n:
        if not k:
            return ZZ.one()
        return ZZ.zero()
    if not k or k > n or n > b * k:
        return ZZ.zero()
    if b >= n:
        return number_of_partitions_length(n, k)

    return ZZ.sum(number_of_partitions_length_max_part(n - m, k - 1, m)
                  for m in range(1, b+1))


def Q(n, a, b, k, m):
    n = ZZ(n)
    if a <= 0: a = 1
    if b > n: b = n
    if k < 0: k = 0
    if m > n: m = n
    # print(n,a,b,k,m)
    if not k and m == n and a == 1:
        # unrestricted length, parts smaller b
        return ZZ.sum(number_of_partitions_length(n, i) for i in range(b + 1))

    return ZZ.sum(number_of_partitions_length_max_part(n - (a - 1)*l, l, b - a + 1)
                  for l in range(k, m+1))


def Q2(n, a, b, k, m):
    n = ZZ(n)
    if a <= 0: a = 1
    if b > n: b = n
    if k < 0: k = 0
    if m > n: m = n
    if a == 1:
        # unrestricted min_part
        if not k and m == n:
            # unrestricted length, parts smaller max_part
            return ZZ.sum(number_of_partitions_length(n, i)
                          for i in range(b + 1))

        return (number_of_partitions_max_length_max_part(n, m, b)
                - number_of_partitions_max_length_max_part(n, k - a, b))

    return ZZ.sum(number_of_partitions_max_length_max_part(n - a * ell, ell, b - a)
                  for ell in range(k, m + 1))
```

mantepse:
I'm afraid the reason is that the recursive approach can reuse previous computations in later calls to `partitions_restricted`,  whereas this does not happen with `q`-binomial.  I am not sure whether this is the only reason.

However, I'd like to finish this, so I'll just polish the current patch a bit.

mantepse:
Next try.  One possibly controversial bit is that if keyword arguments concerning length and part bounds are contradictory, such as `min_length > max_length`, `Partitions.__classcall_private__` now returns `Partitions(-1)`. The reason for this is that handling these contradictory requirements in the new class is quite cumbersome, and also costly, because of the caching

For the same reason, the new class does not handle the special case of the empty partition.  Instead, this case is detected in `Partitions.__classcall_private__`, which returns `Partitions(0)` if appropriate.

As a side effect to the cleaner implementation, the timing for `A(200)` is now down to 5 seconds (from 8 seconds with the previous version).  Of course, using the function `number_of_partitions_max_length_max_part` would still be much faster, because almost all the time is spent in the `UniqueRepresentation` framework. 

@fchapoton, I did not do any further cosmetic changes, but I repeat my offer to separate these out into a separate pull request, and make the present pull request depend on that one.  There are also some unnecessary imports, which I didn't touch, this would be another 4 lines of changes.

tscrim:
What you need to do is cache the `q`-binomials that you have already computed, not the individual coefficients of their output. There probably is some hybrid approach that does more corner/special cases to maximize the speed and/or complementing within a rectangle.

For parent construction, there is at least some overhead for constructing each parent. In your example, you are going through the generic function, which does a number of additional computations. If you directly construct the appropriate parent class, it might be faster as well.

Well, it isn't a necessity to get the absolute best performance right now. Perhaps we can just take this as-is. Let me think just a little bit more about it. Feel free to try one or two things more too.

mantepse:
Good morning!  I tried the brutal thing, on top of this pull request:
```diff
diff --git a/src/sage/combinat/partition.py b/src/sage/combinat/partition.py
index b5537db6626..4607f92c541 100644
--- a/src/sage/combinat/partition.py
+++ b/src/sage/combinat/partition.py
@@ -9735,7 +9735,7 @@ def number_of_partitions_length(n, k, algorithm='hybrid'):
     return ZZ(libgap.NrPartitions(ZZ(n), ZZ(k)))
 
 
-@cached_function
+# @cached_function
 def number_of_partitions_max_length_max_part(n, k, b):
     r"""
     Return the number of partitions of `n` with at most `k`
@@ -9765,6 +9765,8 @@ def number_of_partitions_max_length_max_part(n, k, b):
         sage: A(100)
         10934714090
     """
+    from sage.combinat.q_analogues import q_binomial
+    return q_binomial(k + b, b)[n]
     assert n >= 0 and k >= 0 and b >= 0, f"{n, k, b} must be non-negative"
     if not n:
         return ZZ.one()
diff --git a/src/sage/combinat/q_analogues.py b/src/sage/combinat/q_analogues.py
index 8bde8028ae2..035e493ed17 100644
--- a/src/sage/combinat/q_analogues.py
+++ b/src/sage/combinat/q_analogues.py
@@ -126,6 +126,7 @@ def q_factorial(n, q=None):
     raise ValueError("argument (%s) must be a nonnegative integer" % n)
 
 
+@cached_function
 def q_binomial(n, k, q=None, algorithm='auto'):
     r"""
     Return the `q`-binomial coefficient.
```
Test:
```
sage: from sage.combinat.partition import number_of_partitions_max_length_max_part as f
sage: all(f(n, k, b) == q_binomial(k + b, b)[n] for n in range(5) for k in range(n+1) for b in range(n+1))
True
sage: %cpaste
Pasting code; enter '--' alone on the line to stop or use Ctrl-D.
:def A(n):
:    s1 = number_of_partitions(n)
:    s2 = sum(Partitions(m, max_part=l, length=k).cardinality()
:             * Partitions(n-m-l^2, min_length=k+2*l).cardinality()
:             for l in range(1, (n+1).isqrt())
:             for m in range((n-l^2-2*l)*l//(l+1)+1)
:             for k in range(ceil(m/l), min(m, n-m-l^2-2*l)+1))
:    return s1 + s2
:--
```
Timing:
```
sage: %time A(100)
CPU times: user 15.3 s, sys: 44.7 ms, total: 15.4 s
Wall time: 15.4 s
10934714090
```
whereas, with the pull request only (both in a fresh session, of course):
```
sage: %time A(100)
CPU times: user 441 ms, sys: 7.78 ms, total: 448 ms
Wall time: 448 ms
10934714090
```

mantepse:
> For parent construction, there is at least some overhead for constructing each parent. In your example, you are going through the generic function, which does a number of additional computations. If you directly construct the appropriate parent class, it might be faster as well.

I think that the experiments above in https://github.com/sagemath/sage/pull/38904#issuecomment-2506348053 show that most of this overhead is due to the `UniqueRepresentation` inheritance of `Partitions`.  If I remove that, I get 24 seconds versus 36 seconds for `A(400)` in a fresh session.

I don't see a good reason why "combinatorial" parents, such as `Partitions`, inherit from `UniqueRepresentation`, especially when it is common to create *very many* different parents.  See also Nils' comment https://groups.google.com/g/sage-devel/c/BIFLmpIxLuM

It might be convenient to have a drop in replacement for `UniqueRepresentation` that provides equality based on (preprocessed) input parameters.

mantepse:
Something like
```python
from sage.misc.classcall_metaclass import typecall
class WithEqualityByArgs:
    @staticmethod
    def __classcall__(cls, *args, **options):
        instance = typecall(cls, *args, **options)
        instance._reduction = (cls, args, options)
        return instance

    def __eq__(self, other):
        try:
            return self._reduction == other._reduction
        except AttributeError:
            return False
```
(this seems to work, although I have no idea what I am doing)

tscrim:
@mantepse Here are some reasons why combinatorial objects should still inherit from `UniqueRepresentation`.

A number of these combinatorial objects do use coercion, but it is through the equality (`Partitions` is not among them however). You get a speedup by testing`is` versus `==` for the parents. This simplifies the code because they have all of the requisites for being a subclass of `UniqueRepresentation`. You can also cache things with respect to the parent object (e.g., `P` has a `@cached_method` `foo()` that you run, but then you create another instance, you don't need to recompute `foo()`). You also get overhead by creating the parents, which I expect to be (generally) comparable if not larger than the caching since more data in general needs to be processed. If there is a discrepancy, we should investigate why.

I would say the problem is in your implementation of the computation you want. In general, you should avoid creating many parents for speed.

mantepse:
@tscrim , would you mind moving your comment to #39070?

I hope the present pr is good?

mantepse:
Could we please have this?

mantepse:
ping?

src/sage/combinat/partition.py
```diff
@@ -8784,14 +8930,198 @@ def cardinality(self):
...
+        """
+        try:
+            mu = Partition(x)
+        except ValueError:
+            return False
+        return (mu.size() == self._n
+                and (not mu
+                     or (min(mu) >= self._min_part
+                         and max(mu) <= self._max_part
+                         and self._min_length <= len(mu) <= self._max_length)))
```
```suggestion
        if mu not in _Partitions:
            return False
        if not self._n:
            return not mu
        return (sum(mu) == self._n
                and mu[-1] >= self._min_part
                and mu[0] <= self._max_part
                and self._min_length <= len(mu) <= self._max_length)
```

src/sage/combinat/partition.py
```diff
@@ -8784,14 +8930,198 @@ def cardinality(self):
...
+
+            sage: list(Partitions(9, min_part=2, max_part=4, min_length=3, max_length=4))
+            [[4, 3, 2], [3, 3, 3], [3, 2, 2, 2]]
+        """
+        yield from IntegerListsLex(self._n, max_slope=0,
+                                   min_part=self._min_part,
+                                   max_part=self._max_part,
+                                   min_length=self._min_length,
+                                   max_length=self._max_length,
+                                   element_constructor=Partition)
```
```suggestion
        yield from IntegerListsLex(self._n, max_slope=0,
                                   min_part=self._min_part,
                                   max_part=self._max_part,
                                   min_length=self._min_length,
                                   max_length=self._max_length,
                                   element_constructor=lambda x: self.element_class(self, x))
```
Otherwise this is yielding elements in the wrong parent.

src/sage/combinat/partition.py
```diff
@@ -6728,6 +6806,70 @@ def from_core_and_quotient(self, core, quotient):
...
+            sage: [0] in P
+            True
+            sage: [] in P
+            True
+        """
+        try:
+            mu = Partition(x)
+        except ValueError:
+            return False
+        return mu in Partitions(mu.size(), **self._restrictions)
```
```suggestion
        try:
            return mu in Partitions(sum(mu), **self._restrictions)
        except Exception:
            return False
```

src/sage/combinat/partition.py
```diff
@@ -8784,14 +8930,198 @@ def cardinality(self):
...
+        """
+        Return a string representation of ``self``.
+
+        TESTS::
+
+            sage: from sage.combinat.partition import Partitions_length_and_parts_restricted
+            sage: Partitions_length_and_parts_restricted(9, 1, 9, 2, 9)
+            Partitions of 9 whose parts are at least 2
+            sage: Partitions_length_and_parts_restricted(9, 3, 5, 2, 9)
+            Partitions of 9 having length between 3 and 5 and whose parts are at least 2
```
tscrim, line 9103: I think each type of conditions should be printed at least once (see the Codecov).

tscrim (changes requested):
This likely would be the last set of changes from me before a positive review.

src/sage/combinat/partition.py
mantepse: good catch!

src/sage/combinat/partition.py
mantepse: I agree, done!

mantepse:
Oh no.

tscrim:
Good news is all these failures are the same failure. The bad news is that it is for annoying reasons:
```
sage: [0,0,0,0] in Partitions(0, max_length=3)
False
sage: [0,0] in Partitions(0, max_length=3)
False
sage: [] in Partitions(0, max_length=3)
True
```
this happens because it punts off to the backend. We want the containment check to be quick since it is used in the element construction. Well, perhaps in this case we just need to suffer a bit and do
```python
def __contains__(self, item):
     try:
        return self.backend._contains(_Partitions(item))
    except Exception:
        return False
```
in the fixed `n` with constraints.

Actually, rather than call the new class with "restrictions", use "constraints" instead to avoid confusion with restricted partitions (the dual of regular or has limited number of columns of any given length).

mantepse:
I checked the other containment checks, and many of them fail to treat trailing zeros correctly.  So, there is more work to do.  I also did a performance check:

```
sage: n=15; P = Partitions(n); P1 = Partitions_length_and_parts_constrained(n, 3, 6, 2, 8)
sage: l = [P.random_element() for _ in range(100000)]
sage: %time sum(1 for mu in l if P1.__contains__(mu))
CPU times: user 125 ms, sys: 7 µs, total: 125 ms
Wall time: 125 ms
15690
sage: %time sum(1 for mu in l if P1._contains2(mu))
CPU times: user 200 ms, sys: 18 µs, total: 200 ms
Wall time: 200 ms
15690
```
where
```python
    def __contains__(self, x):
...
        if x not in _Partitions:
            return False
        if not self._n:
            return not x
        return (sum(x) == self._n
                and x[-1] >= self._min_part
                and x[0] <= self._max_part
                and self._min_length <= len(x) <= self._max_length)

    def _contains2(self, x):
        try:
            mu = Partition(x)
        except ValueError:
            return False
        return (mu.size() == self._n
                and (not mu
                     or (min(mu) >= self._min_part
                         and max(mu) <= self._max_part
                         and self._min_length <= len(mu) <= self._max_length)))
```
I'm not sure whether it's worth the effort to do something super intelligent here.  Note that `__contains__` above does not deal with trailing zeros, and that we would have the complication - whatever the more intelligent check might be - in most of the `__contains__` methods.

mantepse:
An awful lot of time is spent in line 6464 of the `_element_constructor_`, i.e., the last (!) line of
```python
        try:
            lst = list(map(ZZ, lst))
        except TypeError:
            raise ValueError(f'all parts of {lst} should be nonnegative integers')
```
Removing `{lst}` would probably make a big difference.

Also, `list(map(ZZ, lst))` seems a bit slower than `[ZZ(e) for e in lst]`.

mantepse:
If tests pass, I hope that this is be good enough.

mantepse:
The two failing tests are #39183 and most likely #38737 - in any case I cannot reproduce them locally.

mantepse:
ping?

fchapoton (approved):
let's move forward and keep further changes for later

PR #38024, opened 2024-05-18
Add Oriented Matroids Package

thecaligarmo:
This pull request merges the Oriented Matroids package that I developed (see https://github.com/thecaligarmo/oriented_matroids) into Sage so that others may develop directly into it rather than relying solely on myself.

This package should resolve ticket #18703 and in addition, my package should be removed from #31164 as this pull request will put Oriented Matroids directly into Sage rather than requiring a package.

If there are any changes needing to be made, just let me know and we can go from there.

### :memo: Checklist

- [x] The title is concise and informative.
- [x] The description explains in detail what this PR is about.
- [x] I have linked a relevant issue or discussion.
- [x] I have created tests covering the changes.
- [x] I have updated the documentation and checked the documentation preview.

PR #38906, opened 2024-11-02
Move matroid constructor

gmou3:
Move the matroid constructor from `constructor.py` to within the `Matroid` class in `matroid.pyx`.

The main `Matroid` class with its new method `__classcall_private__` are to serve as the factory method from which the appropriate subclass is called to instantiate a new matroid.

This has the advantage of identifying the `Matroid` class with the matroid constructor, without requiring an external `Matroid()` constructor function. Now,  instead of the `Matroid()` constructor function, we import directly the `Matroid` class through `all.py` . This means that `isinstance(M, Matroid)` works properly.

This PR follows the same choice of architecture as in #38024.


gmou3:
@thecaligarmo @tscrim
It seems like the `metaclass=ClasscallMetaclass` doesn't work with a `cdef class`. I have also tried overriding the `__new__` method. Any ideas of how we can make the `__classcall_private__` work within a `cdef class`?
As things stand the code compiles but the `__classcall_private__` is not called and every matroid is rendered as a generic `Matroid`.

tscrim:
Sorry for not responding sooner. Hmm...that is indeed a problem. That might be a technical limitation for `Matroid` that requires the separate constructor. We could hack around this by having the public `Matroid` class be the construction using the `ClasscallMetaclass` mechanism, which is then inherited by the lowest level Python classes. But that's ugly and not worth it IMO. The only other thing I know is this:
https://doc.sagemath.org/html/en/reference/cpython/sage/cpython/cython_metaclass.html
That might work.

Note that it is probably better (in this case) for readability to have the construction function in a separate file and then have the `__classcall_private__` redirect to that function.

gmou3:
I will heed the article's warning and not enter these murky `Cython` waters.
I'll probably close this PR soon, unless some other idea emerges.

thecaligarmo:
@gmou3 I see that I'm roughly 2 hours late, but I was going to try and see if I can figure it out. Do you still have a copy of the branch by some chance? Last few days have been busy so I didn't have time to look at things.

If not, no worries, I can try and use the commits above to figure it out and implement something.

gmou3:
Just restored it. Good luck!

PR #38907, opened 2024-11-02
Format function headers around `=` and `,`

gmou3:
A PEP8 sweep mostly of function headers around equal signs and commas.

An illustration of the type of changes:
From
```
def f(x = 0,y):
```
to
```
def f(x=0, y):
```

I am in the awkward position of announcing the abundance of such PEP8-malformations, thousands of which remain in functions calls. I may do a follow-up PR addressing those.


mantepse (approved):
If Not Now, When?

PR #38908, opened 2024-11-02
Don't (mis)use `prec_words_to_bits()`

tornaria:
We need to remove this function from cypari2, because pari 2.17 has
changed precision from words to bits which would cause confusion.

Besides, the current usage is incorrect, since `log_pari.precision()`
will give the precision in *decimal digits* not in words.

See: #38749

### :memo: Checklist


tornaria:
See also: https://github.com/sagemath/cypari2/pull/166#issuecomment-2448193076

antonio-rojas:
Already tested as part of https://github.com/sagemath/sage/pull/38749

videlec (approved):

PR #38909, opened 2024-11-03
Fix ruff issues in `manifolds`

tobiasdiez:
Fix import formatting / ordering in manifolds.

### :memo: Checklist

### DIFF
--- a/src/sage/manifolds/all.py
+++ b/src/sage/manifolds/all.py
@@ -1,4 +1,5 @@
 from sage.misc.lazy_import import lazy_import
+
 lazy_import('sage.manifolds.manifold', 'Manifold')
 lazy_import('sage.manifolds.differentiable.examples.euclidean', 'EuclideanSpace')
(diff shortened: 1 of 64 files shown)


kwankyu:
Why changing ``` ' ``` to ``` " ``` is an improvement? Personally I prefer ``` ' ``` to ``` " ``` for all strings except plain (English) sentences.

tobiasdiez:
> Why changing `'` to `"` is an improvement? Personally I prefer `'` to `"` for all strings except plain (English) sentences.

I don't have an opinion on what style should be preferred, but it should be used consistently. It looked like that `"` was used most often already in the code, so I went with it. It's also happens to be the default in ruff.

tobiasdiez:
> This is too much to review. The changes about string delimiter would better go in their own dedicated PR.

Okay, have reduced it now to step 1: only reorder imports (and always specify the full import for consistency)

fchapoton (approved):
ok, but what a boring thing to do.

PR #38922, opened 2024-11-04
remove the last use of PrincipalIdealDomain

fchapoton:
This is removing the last use of the auld class `PrincipalIdealDomain`

### :memo: Checklist

### DIFF
--- a/src/sage/rings/polynomial/polynomial_ring.py
+++ b/src/sage/rings/polynomial/polynomial_ring.py
@@ -149,7 +149,7 @@ from sage.categories.morphism import IdentityMorphism
 from sage.categories.principal_ideal_domains import PrincipalIdealDomains
 from sage.categories.rings import Rings
 
-from sage.rings.ring import (Ring, IntegralDomain, PrincipalIdealDomain)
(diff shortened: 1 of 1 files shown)


vincentmacri:
If this class isn't used anymore, can the class itself be deleted entirely? It looks like it's already deprecated.

fchapoton:
I has only be deprecated formally recently. We have to wait one year after the first release after the deprecation.

src/sage/rings/polynomial/polynomial_ring.py
```diff
@@ -2217,9 +2216,14 @@ def _element_class():
             from sage.rings.polynomial.polynomial_element_generic import Polynomial_generic_dense_field
             return Polynomial_generic_dense_field
 
+        if category is None:
+            cat = PrincipalIdealDomains()
+        else:
+            cat &= PrincipalIdealDomains()
```
mantepse: There is not `cat` here yet, is there?

fchapoton: oups, désolé

mantepse (approved):
Looks correct!

PR #38923, opened 2024-11-05
remove some deprecated functionality

yyyyx4:
These should all be older than a year.


fchapoton (approved):
ok. In principle, it's one year after the first release with the changes.

yyyyx4:
Yes, that's what I meant. Thanks!

PR #38927, opened 2024-11-05
moving "is_integrally_closed" to the category setup.

fchapoton:
This is moving a method from the auld Ring class to the category setup.

### :memo: Checklist


src/sage/categories/rings.py
```diff
@@ -423,6 +423,28 @@ def is_integral_domain(self, proof=True) -> bool:
...
+                sage: R.is_integrally_closed()
+                False
+                sage: R
+                Order of conductor 2 generated by 2*a in Number Field in a with defining polynomial x^2 + 189*x + 394
+                sage: S = K.maximal_order(); S
+                Maximal Order generated by a in Number Field in a with defining polynomial x^2 + 189*x + 394
+                sage: S.is_integrally_closed()
+                True
+            """
+            return False
```
vincentmacri: I'm not too familiar with the Sage category setup, but why isn't the default implementation to raise a `NotImplementedError`? It looks like that was the previous behaviour.

fchapoton: yes, indeed, that's another possibility and that was the previous behaviour. We can also decide to say "False" if there is no specific reason known for the property in question to hold. I have no strong opinion.

vincentmacri: > yes, indeed, that's another possibility and that was the previous behaviour. We can also decide to say "False" if there is no specific reason known for the property in question to hold. I have no strong opinion.

I feel that it's better to raise an error than report a potentially incorrect result.

vincentmacri:
```suggestion
            raise NotImplementedError
```
And change the docstring accordingly.

vincentmacri (changes requested):

src/sage/categories/rings.py
fchapoton: voila, voila

vincentmacri (approved):
LGTM

mantepse:
I'm overriding the needs-review bot.

PR #38931, opened 2024-11-05
OS X: do not use -ld_classic

jhpalmieri:
Recent versions of Xcode have deprecated `-ld_classic`, so (a) we stop using it, (b) we filter warnings about it when doctesting, and (c) we filter some other warnings related to ld in Xcode.

Without these changes, when building with the latest Xcode, I get many doctest failures because of the new warning
```
ld: warning: -ld_classic is deprecated and will be removed in a future release
```

### :memo: Checklist

### DIFF
--- a/src/bin/sage-env
+++ b/src/bin/sage-env
@@ -373,32 +373,6 @@ if [ -n "$SAGE_LOCAL" ]; then
     # Construct and export LDFLAGS
     if [ "$UNAME" = "Darwin" ]; then
        LDFLAGS="-L$SAGE_LOCAL/lib $LDFLAGS"
-       # On OS X, use the old linker if it is available.
-       # if "ld-classic" is present in the selected XCode
-       # toolchain, add "-Wl,-ld_classic" to LDFLAGS (see #36599) unless
-       # LD is already set, as it will be with conda on macOS.  When the
-       # selected toolchain is in the Xcode app the output of "xcode-select -p"
-       # is "/Applications/Xcode.app/Contents/Developer", but "ld-classic" is
-       # not in the subdirectory "usr/bin/" but rather in the subdirectory
-       # "Toolchains/XcodeDefault.xctoolchain/usr/bin/".  (See #37237.)
-       if [ -z "$LD" ]; then
-	   # Running xcode-select on a system with no toolchain writes an
-	   # error message to stderr, so redirect stderr to /dev/null. 
-	   XCODE_PATH=$(/usr/bin/xcode-select -p 2> /dev/null)
-	   if [ -n $XCODE_PATH ]; then
-               if [ -x "$XCODE_PATH/usr/bin/ld-classic" -o \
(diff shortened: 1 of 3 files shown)


culler:
@jhpalmieri: The code which this PR removes was supposed to check whether the old linker is present in the
toolchain and not add the -ld_classic option if the old linker is not present.  This would use the old linker in older
versions of XCode but would not use the flag in newer versions of XCode which do not have the old linker.  It tested the existence by just checking if ld accepts that option.  I don't see why that test should be removed in favor of never requesting the old linker, even in old versions of XCode where it was buggy.

Since you have a new XCode, you should not have been using the option.  Somehow the test failed.  I am wondering if you might have some mixed up setup, such as a new XCode and an old command line tools installation, which made the test fail.

jhpalmieri:
> Since you have a new XCode, you should not have been using the option. Somehow the test failed. I am wondering if you might have some mixed up setup, such as a new XCode and an old command line tools installation, which made the test fail.

I'm confused: I think that the newer versions of Xcode are precisely the ones that contain an `ld-classic` executable. For example I have Xcode 16.1 and a file `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld-classic` which is detected by that code in `sage-env`. Therefore the `-ld_classic` flag is added to LDFLAGS, and that leads to the many deprecation warnings. 

So I think the question is, are there versions of Xcode that are recent enough to be distributed with an `ld-classic` executable and are old enough that Sage requires us to use the `-ld_classic` flag to actually build?

culler:
Sorry.  I did not understand the situation -- I thought that the ld_classic linker had already been removed in XCode 16.1.

I don't think there is any need to use the old linker with XCode 16.1.  I think the bugs have been fixed by now.  But if you are using XCode 15.0 then you probably do want to use ld_classic.  However, I don't think the flag was deprecated in XCode 15.0.  So maybe the correct thing to do is to make the test be more sopisticated.  Maybe it should run ld with the -ld_classic flag and check if a deprecation warning is issued.

jhpalmieri:
This seems to work: "ld -ld_classic" prints a deprecation warning with the new Xcode, presumably not with Xcode 15.

culler:
This looks good to me.

I extracted the block of code that adds the ld_classic flag (if needed) as a separate script which I then ran on three systems: one with the XCode 14 command line tools, one with the XCode 15 command line tools and one with XCode 16.1.  Only the XCode 15 system got the extra flag.

jhpalmieri:
> This looks good to me.
> 
> I extracted the block of code that adds the ld_classic flag (if needed) as a separate script which I then ran on three systems: one with the XCode 14 command line tools, one with the XCode 15 command line tools and one with XCode 16.1. Only the XCode 15 system got the extra flag.

Great, thank you for checking that! I don't have access to machines running anything except Xcode 16.

culler (approved):

PR #38934, opened 2024-11-07
keep meson.build file for ext/interpreters

dimpase:
`make sagelib-clean` should not remove any `meson.build` files, but now it does, the one in
`src/sage/ext/interpreters/`. 
As  reported here: https://groups.google.com/g/sage-release/c/HHbr1W7vDoM/m/qLYBuIcpCAAJ
This PR fixes this.

### :memo: Checklist


tobiasdiez (approved):
LGTM, but it would be good if someone else has a look at this change as well.

Makefile
```diff
@@ -122,7 +122,7 @@ sagelib-clean:
 	     rm -rf c_lib .cython_version cython_debug; \
 	     rm -rf build; find . -name '*.pyc' -o -name "*.so" | xargs rm -f; \
 	     rm -f $$(find . -name "*.pyx" | sed 's/\(.*\)[.]pyx$$/\1.c \1.cpp/'); \
-	     rm -rf sage/ext/interpreters) \
+	     cd sage/ext/interpreters/; rm -f *.so *.c *.h *.py* *.pxd) \
```
tornaria:
```suggestion
	     cd sage/ext/interpreters/ && rm -f *.so *.c *.h *.py* *.pxd) \
```
I'm a bit wary of `rm -f` and wildcards. This seems dangerous if the directory was removed by some reason.

A milder alternative https://github.com/sagemath/sage/pull/36524#issuecomment-2461184829

dimpase: done

PR #38935, opened 2024-11-07
src/sage/env.py: don't pass if mismatched SAGE_ROOT is None

orlitzky:
In this file we test that `SAGE_ROOT` in a subprocess is the same as `SAGE_ROOT` in the parent. There is a special case for when `SAGE_ROOT is None`, but that special case can pass if only one of the `SAGE_ROOT` variables is `None` and the other is not -- contrary to the intent of the test.

Here we extend the special case to ensure that both `SAGE_ROOT` variables are `None` if one of them is.

Thanks to Gonzalo Tornaría for the suggestion.

### DIFF
--- a/src/sage/env.py
+++ b/src/sage/env.py
@@ -2,16 +2,21 @@
 r"""
 Sage Runtime Environment
 
-Verify that importing ``sage.all`` works in Sage's Python without any ``SAGE_``
-environment variables, and has the same ``SAGE_ROOT`` and ``SAGE_LOCAL``
-(see also :issue:`29446`)::
+Verify that importing ``sage.all`` works in Sage's Python without any
+``SAGE_`` environment variables, and has the same ``SAGE_ROOT`` and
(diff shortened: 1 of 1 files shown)


orlitzky:
wait for CI, I haven't actually tried it

kwankyu:
In what situation, `SAGE_ROOT` could be `None`?

What subprocess is this test for? `sage -python`?

orlitzky:
On 2024-11-18 19:58:58, Kwankyu Lee wrote:
> In what situation, `SAGE_ROOT` could be `None`?

I don't have an example handy, but after sagelib is installed,
SAGE_ROOT should not matter -- maybe some distro packages are setting
it to None to avoid leaking details about the build system?

> What subprocess is this test for? `sage -python`?

Essentially, yes. Whatever python is used to run sage. I think most
often it will be the system python.

kwankyu (approved):
OK. LGTM.

orlitzky:
Thanks again.

PR #38938, opened 2024-11-08
Sanity check parent of Vector_numpy_integer_dense

user202729:
As in the title.

This appears to be what the author intend it to be, previously it's possible to construct with base_ring not ZZ but the elements being returned are ZZ anyway. (plus `self._sage_dtype = ZZ`.)

### :memo: Checklist


tscrim (commented):
I am not quite sure I see the need to do the check since a user would typically never call this class directly. While the test is unlikely to make an impact on any real timings, it feels a bit overboard to me. I would trust the user to know what they are doing if another base ring was used, What are you thoughts about this?

user202729:
But… the class is never called from anywhere else? (as far as I can see)

This is unlike `Vector_real_double_dense` where it is automatically selected by `def vector():` when the base ring is `RR`.

If not called by the user what is it for?

tscrim:
I can only guess, but probably for eventually being able to set the backend of `FreeModule`. That shouldn't be too hard to do, but it probably would take a bit of time and effort to do. As-is, this feels somewhat like lip service to people who want numpy stuff...

user202729:
The problem is using that class is inherently dangerous because it has no overflow check… right (?)

tscrim:
Yes, there are a number of issues as I don't think Matthias fully implemented his idea. Well, perhaps we will just include this for now as a way to make this a little safer to use, which we can later remove if the invocation becomes less user-facing (not that it is very much so right now anyways...). Thoughts?

tscrim:
PS - I should also say the only thing I can do to contribute to implementing such a backend right now is in reviewing code.

user202729:
> perhaps we will just include this for now as a way to make this a little safer to use, which we can later remove if the invocation becomes less user-facing

Sounds good to me.

Another alternative I can see is to explicitly document the limitations.
(The original motivation for this PR is that I see the code there, but was confused what the class is used for, so I try to is instantiate it with Zmod(p) and it actually works except that it produces erratic behavior.)

I think that would be sufficient, because after all, users would never encounter this issue themselves unless they actually come across the corresponding page in the documentation.

tscrim (approved):
Then let's just get this in as-is for now. We can always make further changes/improvements down the road.

Issue #37295, opened 2024-02-11
Random Segfaults in `matrix_gap`


tobiasdiez:
### Steps To Reproduce

	https://github.com/sagemath/sage/actions/runs/7730743552/job/21076818758
	https://github.com/sagemath/sage/actions/runs/7666354980/job/20893984833

### Expected Behavior

Tests pass

### Actual Behavior

```
sage -t --random-seed=141136157000169917298631972457521844114 src/sage/matrix/matrix_gap.pyx
    Killed due to segmentation fault
**********************************************************************
Tests run before process (pid=101128) failed:
sage: M = MatrixSpace(ZZ, 2, implementation='gap') ## line 25 ##
sage: m1 = M([1, 0, 2, -3]) ## line 26 ##
sage: m2 = M([2, 2, 5, -1]) ## line 27 ##
sage: type(m1) ## line 28 ##
<class 'sage.matrix.matrix_gap.Matrix_gap'>
sage: m1 * m2 ## line 31 ##
[  2   2]
[-11   7]
sage: type(m1 * m2) ## line 34 ##
<class 'sage.matrix.matrix_gap.Matrix_gap'>
sage: M = MatrixSpace(QQ, 5, 3, implementation='gap') ## line 37 ##
sage: m = M(range(15)) ## line 38 ##
sage: m.left_kernel() ## line 39 ##
Vector space of degree 5 and dimension 3 over Rational Field
Basis matrix:
[ 1  0  0 -4  3]
[ 0  1  0 -3  2]
[ 0  0  1 -2  1]
sage: M = MatrixSpace(ZZ, 10, implementation='gap') ## line 46 ##
sage: m = M(range(100)) ## line 47 ##
sage: m.transpose().parent() is M ## line 48 ##
True
sage: UCF = UniversalCyclotomicField() ## line 52 ##
sage: M = MatrixSpace(UCF, 3, implementation='gap') ## line 53 ##
sage: m = M([UCF.zeta(i) for i in range(1,10)]) ## line 54 ##
sage: m ## line 55 ##
[               1               -1             E(3)]
[            E(4)             E(5)          -E(3)^2]
[            E(7)             E(8) -E(9)^4 - E(9)^7]
sage: (m^2)[1,2] ## line 59 ##
E(180)^32 - E(180)^33 + E(180)^68 - E(180)^69 + E(180)^104 - E(180)^141 - E(180)^156 + E(180)^176 - E(180)^177
[...]

PR #38943, opened 2024-11-09
Some documentation improvement

user202729:
Just improve some documentation.

Apparently the existing documentation uses the absolute URL ``   * - `Tutorial <../tutorial/index.html>`_`` e.g.  `sage/src/doc/en/developer/sage_manuals.rst`, none of them uses `<tutorial>`. I tried using `<chapter-tutorial>` in https://github.com/sagemath/sage/commit/955a09e746c7d9675d2a4c4f10558cec8d3eb948 but for some reason it doesn't work, so I just end up deleting it.

### :memo: Checklist

------

test failures have nothing to do with the change, see https://github.com/sagemath/sage/issues/37295


src/doc/en/developer/coding_in_cython.rst
```diff
@@ -60,14 +60,19 @@ There are several ways to create and build Cython code in Sage.
...
+#. Use ``%%cython`` cell magic in the command line.
+   Refer to :meth:`sage.repl.ipython_extension.SageMagics.cython`.
 
 #. Create a ``.pyx`` file and add it to the Sage library.
    Then run ``sage -b`` to rebuild Sage.
 
+.. _section-attach-or-load-spyx-files:
 
 Attaching or loading .spyx files
 ================================
```
kwankyu, line 72: While we are here, we should consider moving everything about ".spyx" here into the head of https://doc-pr-38943--sagemath.netlify.app/html/en/reference/repl/sage/repl/load. 

Only things about ".pyx" are concerns of sage developers.

user202729: On the other hand I think it's valid to want to disable the preparser too. For example `%%cython` cell does not preparse (if I recalled correctly).

But okay, point noted. (as an user stumble upon that page while looking for information how to use Cython in Sage, I hardly notice it is meant for developers.)

user202729: Actually are you sure? Then the developer guide page would have almost nothing left except maybe

> Create a .pyx file and add it to the Sage library. Then run sage -b to rebuild Sage.


and

> **Deprecation**  
> When making a backward-incompatible modification in Sage, the old code should keep working and display a message indicating how it should be updated/written in the future. We call this a deprecation.

I'm not sure how best to organize it. Maybe leaving it as is is fine — users who want to code in Cython would stumble across the developer's guide page anyway.

kwankyu: What I hoped for is perhaps out of the scope of this PR. You don't have to do it.

kwankyu: On the other hand, the two links you added in lines starting line 63 seem to be pointless. I suggest to leave the paragraph
```
#. Create an ``.spyx`` file and attach or load it from the command
   line. This is similar to creating a ``%cython`` cell in the
   notebook but works completely from the command line (and not from
   the notebook).
```
as it is, but just update `%cython` with  `%%cython`. I think `%cython` worked for old sage notebook.

user202729: Ah well… I didn't test the notebook version and just assume the syntax is somehow different.


image

Looks like the syntax is the same on both command-line and notebook now.

user202729: `attach()` doesn't work on the notebook though. (apparently known issue from long ago, https://ask.sagemath.org/question/36364/attachattach-does-not-work-in-sage-jupyter-sage-751/ )

My issue with the command-line is the bash command-line is also called command-line (https://doc.sagemath.org/html/en/reference/repl/options.html) so…

kwankyu: > `attach()` doesn't work on the notebook though. (apparently known issue from long ago, https://ask.sagemath.org/question/36364/attachattach-does-not-work-in-sage-jupyter-sage-751/ )

The notebook refers to the old (now deprecated) classic sage notebook, not the jupyter notebook. The paragraph itself is somewhat outdated. But, as I said above, updating the paragraph is not in the scope of your pr.

> My issue with the command-line is the bash command-line is also called command-line (https://doc.sagemath.org/html/en/reference/repl/options.html) so…

Then explicitly stating "the Sage command line" or "the Sage prompt" would make it less confusing. If you state "the Sage command line"  early in the paragraphs, there would really be no confusion.

src/doc/en/developer/coding_in_cython.rst
```diff
@@ -32,13 +32,13 @@ up-to-date information or check out the
...
-Writing cython code in Sage
+Writing Cython code in Sage
 ===========================
 
 There are several ways to create and build Cython code in Sage.
 
-#. In the Sage Notebook, begin any cell with ``%cython``. When you
-   evaluate that cell,
+#. In the Sage Notebook or the Sage command line, begin any cell with
+   a line containing ``%%cython``. When you evaluate that cell,
```
kwankyu, line 42: I think `the Sage notebook` is appropriate since `Sage Notebook` refers to the old deprecated Sage's own notebook. Now there is only Jupyter notebook with Sage kernel.

user202729: You mean is not appropriate (?) 

Anyway looking through the documentation it doesn't seem to have a special name, it's just referred to by "the notebook interface".

Maybe I can just write "In the notebook interface or the command-line interface, begin any cell with a line…"

Problem (though this is maybe too nitpicky: I'm not sure what a `sage: ` part in the command-line is called. Is "cell" appropriate?

The terminology isn't consistent throughout either. https://doc.sagemath.org/html/en/tutorial/interactive_shell.html uses "interactive shell", and  https://doc.sagemath.org/html/en/reference/repl/index.html uses "command line".

kwankyu: I meant that "the Sage Notebook" (with capital `N`) is a proper noun, and hence it implies there is some Notebook software that we developed. But there is no such thing after we dropped the classic Sage Notebook. So I suggested "the Sage notebook" (with small `n`) that rhymes well with "the Sage command line".

You may just leave it as it is, if you are reluctant to touch things out of the scope of your PR.

"sage: " is called the Sage prompt, if your question means that.

user202729: Might as well fix. Found a few other instances.

(though some other instances are from other languages, I'm not sure if it's appropriate to change it because some languages may use capitalization in other ways than English e.g. if I recalled correctly German has some capitalization rules.

kwankyu: Thanks.

kwankyu (approved):
Thanks. LGTM.

PR #38946, opened 2024-11-09
Allow Cython compilation to open a web browser containing the annotated HTML file

user202729:
As in the title.

I think this is a rather common use case, to investigate whether the code does get the desired speedup.

Remark: there's `sage.misc.viewer.browser()` which could be used instead, but then a lot of existing code in SageMath uses `webbrowser.open()` anyway?

### :memo: Checklist

### Dependencies

Depends on #38945 .


kwankyu:
This looks nice. 

It would be nicer if the new argument `view_annotate` is accepted to the cell magic `%%cython`. For example,
```
%%cython view_annotate=True
def function(int n):
...
```
or
```
%%cython --view_annotate
def function(int n):
...
```

The cell magic is defined in https://github.com/sagemath/sage/blob/develop/src/sage/repl/ipython_extension.py. Find the line
```
def cython(self, line, cell):
```
`line` seems to accept the part after "%%cython". So we may use it to provide arguments to the `cython` function. What do you think?

user202729:
Already did that in https://github.com/sagemath/sage/pull/38945 . (though whichever being merged first need to be modified to incorporate the other…)

kwankyu:
You may set #38945 as a dependency and merge it here.

kwankyu:
and instead of 
```
- ``line`` -- parsed as keyword arguments. See :func:`~sage.misc.cython.cython` for details.
```
you may at least list the possible arguments for users' convenience, leaving details to `sage.misc.cython.cython` .

user202729:
I learnt it's possible to view the result in the Sage notebook (if Jupyter-based notebook GUI is used) from looking at cython-provided magic, so I implement that here.

user202729:
@kwankyu Can you review the PR? Thanks.

kwankyu:
Conflicts?

user202729:
Huh, that only come from the test today. This one is pretty harmless.

src/sage/misc/cython.py
```diff
@@ -110,6 +111,14 @@ def cython(filename, verbose=0, compile_message=False,
...
       then save a copy of the .html file in the current directory.
 
+    - ``view_annotate`` -- boolean (default: ``False``); if ``True``, open the
+      annotated html file in a web browser using :func:`webbrowser.open`
+
+    - ``view_annotate_callback`` -- function (default: ``webbrowser.open``); a function
+      that takes a string being the path to the html file. This can be overridden to
+      change what to do with the annotated html file. Have no effect unless
+      ``view_annotate`` is ``True``
+
```
kwankyu, line 129: Period at the end in this case.

src/sage/misc/cython.py
```diff
@@ -381,6 +418,11 @@ def cython(filename, verbose=0, compile_message=False,
             shutil.copy(os.path.join(target_dir, name + ".html"),
                         os.curdir)
 
+    if view_annotate:
+        if not annotate:
+            raise ValueError("Cannot view annotated file without creating it")
+        view_annotate_callback(os.path.join(target_dir, name + ".html"))
+
     # This emulates running "setup.py build" with the correct options
```
kwankyu, line 434: ```ValueError("cannot ...")```?

src/sage/repl/ipython_extension.py
```diff
@@ -357,12 +373,23 @@ def cython(self, line, cell):
...
           - ``--sage-namespace``
           - ``--create-local-so-file``
           - ``--no-compile-message``, ``--no-use-cache``, etc.
 
           See :func:`~sage.misc.cython.cython` for details.
 
+          For ``--view-annotate``, the following additional choices are allowed:
+
+          - ``--view-annotate=none`` (default)
+          - ``--view-annotate=auto`` (same as ``--view-annotate``); select one of the
```
kwankyu, line 386: `None`?

kwankyu: I think this default `none` is useless.

src/sage/repl/ipython_extension.py
```diff
@@ -357,12 +373,23 @@ def cython(self, line, cell):
...
+            choices below automatically
+          - ``--view-annotate=webbrowser``; open the annotation in a web browser
+            (preferred if the Sage command line is used)
+          - ``--view-annotate=displayhtml``; display the annotation inline in the notebook
+            (preferred if the Sage notebook is used)
+
         - ``cell`` -- string; the Cython source code to process
```
kwankyu, line 390: How about removing the lines 389 and 391 and instead use them to explain "automatically" for "auto" option?

src/sage/repl/ipython_extension.py
kwankyu: and "auto" should be the default (for the case ``--view-annotate``).

user202729: I interpret default to be "which option is chosen if nothing is specified", in that case the default is indeed it is not viewed.

Indeed user would never need to explicitly pass `--view-annotate=none` unless to override a previous `--view-annotate` option.

kwankyu: What do you mean by "nothing is specified":
```
%%cyhon --view-annotate
```
or 
```
%%cyhon 
```
? 

Perhaps you mean the latter. But in the context, I think the former is understood. Then the default is "auto".

kwankyu:
On comand line (on mac),
```
sage: %%cython --view-annotate
....: def f(int n):
....:     return n*n
....: 
``` 
does not open a web browser.

user202729:
Not sure what's going on, but probably doesn't have anything to do with the code. There are some complaints e.g. https://stackoverflow.com/questions/65981130/python-webbrowser-module-doesnt-work-on-terminal or https://github.com/python/cpython/issues/69143

You can try `oeis.browse()` which also uses `webbrowser.open()` — if it also doesn't work it probably isn't the program's fault.

If it does work, maybe you can try this patch

```diff
diff --git a/src/sage/misc/cython.py b/src/sage/misc/cython.py
index fde36372712..3689c896b4a 100644
--- a/src/sage/misc/cython.py
+++ b/src/sage/misc/cython.py
@@ -81,7 +81,7 @@ sequence_number = {}
 
 def cython(filename, verbose=0, compile_message=False,
            use_cache=False, create_local_c_file=False, annotate=True, view_annotate=False,
-           view_annotate_callback=webbrowser.open, sage_namespace=True, create_local_so_file=False):
+           view_annotate_callback=lambda x: webbrowser.open("file://" + x), sage_namespace=True, create_local_so_file=False):
     r"""
     Compile a Cython file. This converts a Cython file to a C (or C++ file),
     and then compiles that. The .c file and the .so file are
```

src/sage/repl/ipython_extension.py
user202729: I decided to simplify the documentation. The `=none` and `=auto` option is now undocumented (the user doesn't need to know that anyway)

src/sage/repl/ipython_extension.py
```diff
@@ -357,12 +373,20 @@ def cython(self, line, cell):
...
           - ``--sage-namespace``
           - ``--create-local-so-file``
           - ``--no-compile-message``, ``--no-use-cache``, etc.
 
           See :func:`~sage.misc.cython.cython` for details.
 
+          If ``--view-annotate`` is given, the annotation is either displayed
+          inline in the Sage notebook or opened in a new web browser, depends
+          on whether the Sage notebook is used.
+
```
kwankyu, line 386: `depending on`

src/sage/repl/ipython_extension.py
```diff
@@ -357,12 +373,20 @@ def cython(self, line, cell):
...
+          You can override the selection by specifying
+          ``--view-annotate=webbrowser`` or ``--view-annotate=displayhtml``.
+
```
kwankyu, line 389: OK.

kwankyu:
> If it does work, maybe you can try this patch
> 
> ```diff
> diff --git a/src/sage/misc/cython.py b/src/sage/misc/cython.py
> index fde36372712..3689c896b4a 100644
> --- a/src/sage/misc/cython.py
> +++ b/src/sage/misc/cython.py
> @@ -81,7 +81,7 @@ sequence_number = {}
>  
>  def cython(filename, verbose=0, compile_message=False,
>             use_cache=False, create_local_c_file=False, annotate=True, view_annotate=False,
> -           view_annotate_callback=webbrowser.open, sage_namespace=True, create_local_so_file=False):
> +           view_annotate_callback=lambda x: webbrowser.open("file://" + x), sage_namespace=True, create_local_so_file=False):
>      r"""
>      Compile a Cython file. This converts a Cython file to a C (or C++ file),
>      and then compiles that. The .c file and the .so file are
> ```

Yes. That does the trick. For readability, you may want to store the lambda function to a variable first, before line 82.

src/sage/repl/ipython_extension.py
kwankyu: OK. But you said that one can use  `none` to override a previous `--view-annotate` option... Undocumented feature may be silently removed.

I see no problem with
```
--view-annotate=auto (default) 
--view-annotate=webbrowser
--view-annotate=displayhtml
--view-annotate=none 
```

Anyway, I won't insist.

user202729:
I come across https://github.com/sagemath/sage/issues/33931 , while it looks like the PR is stalled it is recommended to use webbrowser over `sage.misc.viewer`. So I'll use `webbrowser.open()` anyway.

kwankyu (approved):
Thanks. LGTM.

antonio-rojas:
A test is failing here
```
File "/usr/lib/python3.13/site-packages/sage/repl/ipython_extension.py", line 433, in sage.repl.ipython_extension.SageMagics.cython
Failed example:
    shell.run_cell('''
    %%cython --view-annotate=xx
    print(1)
    ''')
Expected:
    UsageError: argument --view-annotate: invalid choice: 'xx' (choose from 'none', 'auto', 'webbrowser', 'displayhtml')
Got:
    UsageError: argument --view-annotate: invalid choice: 'xx' (choose from none, auto, webbrowser, displayhtml)
```

user202729:
Should be fixed in https://github.com/sagemath/sage/pull/39282

user202729:
This PR leads to an annoying consequence: look at https://doc-pr-39509--sagemath.netlify.app/html/en/reference/misc/sage/misc/cython#hunk1

> sage.misc.cython.cython(filename, verbose=0, compile_message=False, use_cache=False, create_local_c_file=False, annotate=True, view_annotate=False, view_annotate_callback=<function _webbrowser_open_file at 0x7f3980c4dc60>, sage_namespace=True, create_local_so_file=False)[[source]](https://github.com/sagemath/sage/blob/develop/src/sage/misc/cython.py#L90)

the address in `<function _webbrowser_open_file at 0x7f3980c4dc60>` is essentially random, which leads to it showing up in the documentation diff every time.

kwankyu:
Right..

> (the other is to let the default be `None` and in the function body handles accordingly, but that is somewhat semantically misleading in that normally `None` should represent a function that do nothing)

I prefer this approach. I think this is rather a "standard way". You may add an explanation in the docstring to avoid confusion.

PR #38952, opened 2024-11-10
adding the ring of quantum-valued polynomials

fchapoton:
This is a q-analogue of the ring of integer-valued polynomials. The elements are polynomials in x such that their evaluation at every q-integer is a polynomial in q.

### :memo: Checklist


src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
+r"""
+Quantum-valued polynomial rings
+
+AUTHORS:
+
+- Frédéric Chapoton (2024-03): Initial version
+
+REFERENCES:
+
+.. [HaHo2017] Nate Harman and Sam Hopkins, *Quantum integer-valued
```
tscrim: Probably would be good to add this to main reference file.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+   polynomials*, J. Alg. Comb. 2017, :doi:`10.1007/s10801-016-0717-3`
+
+"""
+# ***************************************************************************
+#  Copyright (C) 2024 Frédéric Chapoton <chapoton-math-unistra-fr>
+#
+#  Distributed under the terms of the GNU General Public License (GPL)
+#                  https://www.gnu.org/licenses/
```
tscrim: This I think should match the standard GPL format in our dev guide.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+
+    Quantum-valued polynomial rings are commutative and associative
+    algebras, with a basis indexed by integers.
+
+    This is endowed with two bases, named ``B`` or ``Binomial``
+    and ``S`` or ``Shifted``.
+
+    INPUT:
+
+    - ``R`` -- commutative ring
```
tscrim, line 133: I think it would be good to have `q` be included as an argument, subsequently it would be a part of the ground ring. Now you can have an option like if the input is :
1. Only `R`, then append `q` as currently implemented.
2. Both `(R, q)`, then do `q = R(q)`.
3. Just `q` and get `R = q.parent()`.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+                    base_N = poly(N)
+                    top_coeff = remain.leading_coefficient() / base_N.lc()
+                    denom = top_coeff.denominator()
+                    if denom.is_term():
+                        numer = top_coeff.numerator()
+                        top_coeff_laurent = laurent_polys(numer) / laurent_polys(denom)
+                    else:
+                        msg = 'not a polynomial with integer values :'
+                        msg += f' {top_coeff} is not a Laurent polynomial'
+                        raise ValueError(msg)
```
tscrim, line 411: It would be good to have a test for this.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+
+                sage: A = QuantumValuedPolynomialRing(QQ).S()
+                sage: A.product_on_basis(0, 1)
+                S[1]
+            """
+            i = ZZ(n1)
+            j = ZZ(n2)
+            if j < i:
+                j, i = i, j
+            q = self.base_ring().gen()
```
tscrim: Rather than getting this every time, I think it would be better to store `self._q = self.base_ring().gen()` in the `__init__`.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+                sage: A.from_h_vector(ex.h_vector())
+                S[2] + q*S[3]
+            """
+            B = self.basis()
+            ring = self.base_ring()
+            q = ring.gen()
+            d = len(hv) - 1
+            m = matrix(ring, d + 1, d + 1,
+                       lambda j, i: (-1)**(d - j) * q_binomial(d - i, d - j, q) *
+                       q**(-d * (d - i) + binomial(d - j, 2)))
```
tscrim: Any reason why you aren't simply creating the lists of lists (or a single list) to pass to the constructor here?

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+                if P is self:
+                    return x
+                if P is not self.base_ring():
+                    return self.element_class(self, x.monomial_coefficients())
+
+            # ok, not a quantum-valued polynomial ring element
+            R = self.base_ring()
+            # coercion via base ring
+            x = R(x)
+            if x == 0:
```
tscrim:
```suggestion
            if not x:
```

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1170 @@
...
+                sage: F._poly(4).factor()
+                (1/(q^6 + 3*q^5 + 5*q^4 + 6*q^3 + 5*q^2 + 3*q + 1)) *
+                (q*x + 1) * (q^2*x + q + 1) * (q^3*x + q^2 + q + 1) *
+                (q^4*x + q^3 + q^2 + q + 1)
+            """
+            return q_binomial_x(i, i)
+
+        class Element(CombinatorialFreeModule.Element):
+
+            def umbra(self):
```
tscrim:
```suggestion
        class Element(CombinatorialFreeModule.Element):
            def umbra(self):
```

tscrim (changes requested):

src/sage/rings/polynomial/q_integer_valued_polynomials.py
fchapoton: The new attribute "_q" is not inherited by the Bases. So I have made a method ".q" there. If you have a better way, please tell me.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
fchapoton: Hmm, I would rather not allow a custom q. The ring R is not there to provide q, in my mind. I have added a sentence about that in the doc.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
tscrim: Either have an ABC for the basis or just have both of them provide it.

Nevertheless, having a method `q()` is a good idea.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
tscrim: You don’t expect anyone to want to specialize `q`? It also feels slightly off mathematically to not have `q` in the base ring, but perhaps that is just from my experience with Hecke algebras and quantum groups...

fchapoton: I am not sure. For my own usage, I have a base ring of coefficients, almost always QQ, and the variable q on top of that. For this algebra of quantum-valued polynomials, I can see no interest to have q a complex number. Of course, one can evaluate to polynomials and then evaluate q if one will.

tscrim: I believe certain Askey scheme $q$-orthogonal polynomials (e.g., [cont. q-Hermite](https://en.wikipedia.org/wiki/Continuous_q-Hermite_polynomials)) would be in this ring, and problems I've worked on, we want to consider $q \in \mathbb{R}_{>0}$. There might also be important considerations for doing things over finite fields (or $q$ satisfying some relation). I think it would be better to be more general as it is easy to add and keep the most common behavior ($q$ is a generic parameter).

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1203 @@
...
+        return ring.one()
+    return ring.prod(q_int_x(m + 2 - i) / q_int(i) for i in range(1, n + 1))
+
+
+class QuantumValuedPolynomialRing(UniqueRepresentation, Parent):
+    r"""
+    The quantum-valued polynomial ring over a base ring.
+
+    Quantum-valued polynomial rings are commutative and associative
+    algebras, with a basis indexed by integers.
```
tscrim: Can you also add the definition to this documentation?

src/sage/rings/polynomial/q_integer_valued_polynomials.py
tscrim: Thank you. However, can you put the input normalization in a `__classcall_private__`? Otherwise `(LaurentPolynomialRing(QQ, 'q'), q)` and `(QQ)` would give identical but distinct instances.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1250 @@
+r"""
+Quantum-valued polynomial rings
+
+This provide a `q`-analogue of the :class:`~sage.rings.polynomials.integer_valued_polynomials.IntegerValuedPolynomialRing`.
```
tscrim:
```suggestion
This provides a `q`-analogue of the
:class:`~sage.rings.polynomials.integer_valued_polynomials.IntegerValuedPolynomialRing`.
```

src/sage/rings/polynomial/q_integer_valued_polynomials.py
fchapoton: I have now allowed to provide an optional q. Then the ring must contain this q.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
fchapoton: done, hopefully

tscrim (commented):
Thank you. This should be the last two things.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
```diff
@@ -0,0 +1,1255 @@
...
+        sage: 1 - B[2] * B[2] / 2
+        S[0] - (1/2*q^-3)*S[2] + (1/2*q^-4+q^-3+q^-2+1/2*q^-1)*S[3]
+        - (1/2*q^-4+1/2*q^-3+q^-2+1/2*q^-1+1/2)*S[4]
+    """
+    @staticmethod
+    def __classcall_private__(cls, R, q=None) -> None:
+        """
+        Normalize the input.
+        """
+        if R not in Rings().Commutative():
```
tscrim:
```suggestion
    def __classcall_private__(cls, R, q=None) -> None:
        """
        Normalize the input.

        EXAMPLES::

            sage: q = LaurentPolynomialRing(QQ, 'q').gen()
            sage: F1 = QuantumValuedPolynomialRing(QQ)
            sage: F2 = QuantumValuedPolynomialRing(q.parent(), q)
            sage: F1 is F2
            True
        """
        if R not in Rings().Commutative():
```
Thank you. Now just to add a doctest verifying the consistency. Once done, this will be a positive review.

src/sage/rings/polynomial/q_integer_valued_polynomials.py
tscrim: This will work. It feels very slightly different than the definition in the paper, but the definition here definitely allows for more general base rings. Not sure if this makes any actual difference though.

tscrim (approved):
Thank you. LGTM.

fchapoton:
the meson conda failure seems to be related. No idea what this is doing, apparently not seeing the new module. Not the first time I met this issue, see also #38965

tscrim:
Yea, I usually ignore that since it happens so often... positive review.

tobiasdiez:
You have to add the new file in `src/sage/rings/polynomial/meson.build`, either manually or by running `tools/update-meson.py`.
Sorry, it's on my todo list to add documentation around this and post a bit more details about meson on the mailing list.

tobiasdiez:
Thanks for the fix!

PR #38954, opened 2024-11-10
Fix configure script generated by pkgconf-2.3.0

vbraun:
Pkgconf 2.3.0 changes the PKG_PROG_PKG_CONFIG macro to error out instead of marking it as not found. Arguably this is the correct default for most but not for us, since we support building on macos where Apple, in their infinite wisdom, decided not to ship their dev tools with it.

See also: https://github.com/pkgconf/pkgconf/blob/master/NEWS

    dnl PKG_PROG_PKG_CONFIG([MIN-VERSION], [ACTION-IF-NOT-FOUND])
    dnl ---------------------------------------------------------
    dnl Since: 0.16
    dnl
    dnl Search for the pkg-config tool and set the PKG_CONFIG variable to
    dnl first found in the path. Checks that the version of pkg-config found
    dnl is at least MIN-VERSION. If MIN-VERSION is not specified, 0.9.0 is
    dnl used since that's the first version where most current features of
    dnl pkg-config existed.
    dnl
    dnl If pkg-config is not found or older than specified, it will result
    dnl in an empty PKG_CONFIG variable. To avoid widespread issues with
    dnl scripts not checking it, ACTION-IF-NOT-FOUND defaults to aborting.
    dnl You can specify [PKG_CONFIG=false] as an action instead, which would
    dnl result in pkg-config tests failing, but no bogus error messages.

### :memo: Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -222,7 +222,7 @@ dnl Exit autoconf with exit code 16 in this case. This will be
 dnl caught by the bootstrap script.
 m4_exit(16)])
 
-PKG_PROG_PKG_CONFIG([0.29])
+PKG_PROG_PKG_CONFIG([0.29], [PKG_CONFIG=false])
 
 AC_CHECK_PROG(found_ranlib, ranlib, yes, no)
 if test x$found_ranlib != xyes


kwankyu:
I am new to this territory. So perhaps a stupid question: Is pkgconf or pkg-config a build prerequisite on mac? I think it should because either of them must provide the macro PKG_PROG_PKG_CONFIG in the first place. But then why it does not appear here: https://doc-release--sagemath.netlify.app/html/en/reference/spkg/_prereq ?

vbraun:
The pkgconf m4 scripts are, together with autoconf/autotools, a prerequisite for generating `configure.ac` -> `configure`. 

But they are not required for building Sage from source. The `bootstrap` script will either run autoconf or, failing that, download a pre-generated `configure`. This is `build/pkgs/configure`, and I generate a new one whenever I push a new version.

Fedora 40 -> 41 updated the pkgconf m4 scripts, so this the breakage comes from.

kwankyu (approved):
Thanks for explaining how bootstrap things work. LGTM.

PR #38956, opened 2024-11-12
Hide GAP finiteness warnings during isomorphism check 

### LINKED ISSUE #38889: Random test failures in `finitely_presented_named`
### Steps To Reproduce

_No response_

### Expected Behavior

Test pass

### Actual Behavior

```
sage -t --warn-long 5.0 --random-seed=130677747974987316498399576088509580425 src/sage/groups/finitely_presented_named.py
File "src/sage/groups/finitely_presented_named.py", line 485, in sage.groups.finitely_presented_named.AlternatingPresentation
Failed example:
    A1.is_isomorphic(A2), A1.order()
Expected:
    (True, 1)
Got:
    #I  Forcing finiteness test
    (True, 1)
```

### Additional Information

_No response_

### Environment

Linux: https://github.com/sagemath/sage/actions/runs/11572245422/job/32211815924

### Checklist

orlitzky:
To fix https://github.com/sagemath/sage/issues/38889 and others like it, we hide the `"#I  Forcing finiteness test` warning that can be printed by GAP if one of the groups is not (yet) known to be finite.


dcoudert (approved):
LGTM.

orlitzky:
Thank you!

user202729:
Has been a while, but I wonder if hiding the warning is the best idea. After all, warnings are displayed for a reason.

The original motivation is just to make the doctests pass. For that Sage has a custom `SageOutputChecker`, where you can implement `do_fixup()`.

orlitzky:
> Has been a while, but I wonder if hiding the warning is the best idea. After all, warnings are displayed for a reason.

The isomorphism check only works on finite groups, so my thought process was, if it has to check that the groups are finite... okay? Is there any action I should take? If not, it's just noise. Printing unstructured GAP messages to the console also isn't great from the sage UI perspective since there's no easy way to hide them in your own program like there would be with e.g. a python warning.

For power users and developers, the warnings are still available if you really want to see them: just use `G.gap().IsomorphismGroups(H.gap())` instead of the sage method.

PR #38958, opened 2024-11-12
Allow powers of SignGroup elements to live in other parents

behackl:
A `SignGroup` is used as an auxiliary construct in the context of [asymptotic expansions](https://doc.sagemath.org/html/en/reference/asymptotic/sage/rings/asymptotic/asymptotic_ring.html) to split the growth of terms like `(-2)^n * n^42` into, essentially, `2^n * n^42 * (-1)^n` that lives in a cartesian product `QQ^n * n^QQ  * Signs^n`.

In the current implementation of `Sign.__pow__`, the power is computed using the raw data (essentially, `int ** whatever`) and then an attempt is made to push the result back into `SignGroup`. This changes this behavior to still attempt the conversion -- but if it fails while the raw operation itself was successful, the result living in some other parent not coercing back to `SignGroup` is returned.

Simple example:

```py
sage: from sage.groups.misc_gps.argument_groups import SignGroup
sage: S = SignGroup()
sage: x = SR.var('x')
sage: S(-1)^x  # this fails on develop
(-1)^x
sage: _.parent()
Symbolic Ring
```

For asymptotic expansions, this fixes an issue preventing substitutions in expansions with terms that involve, e.g., `(-1)^n`.

### :memo: Checklist

### DIFF
--- a/src/sage/groups/misc_gps/argument_groups.py
+++ b/src/sage/groups/misc_gps/argument_groups.py
@@ -1474,10 +1474,9 @@ class Sign(AbstractArgument):
         result = self._element_ ** exponent
         P = self.parent()
         try:
-            result = P.element_class(P, self._element_ ** exponent)
+            return P.element_class(P, result)
         except (ValueError, TypeError):
-            pass
-        return result
+            return result
 
     def __invert__(self):
         r"""
--- a/src/sage/groups/misc_gps/argument_groups.py
+++ b/src/sage/groups/misc_gps/argument_groups.py
@@ -1461,6 +1461,15 @@ class Sign(AbstractArgument):
             1
             sage: S(-1)^3  # indirect doctest
             -1
+
+        Check that the results may live in other parents too::
+
+            sage: x = SR.var('x')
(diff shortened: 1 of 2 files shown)


src/sage/groups/misc_gps/argument_groups.py
```diff
@@ -1461,9 +1461,23 @@ def __pow__(self, exponent):
...
+            (-1)^x
+            sage: elem.parent()
+            Symbolic Ring
+
         """
+        result = self._element_ ** exponent
         P = self.parent()
-        return P.element_class(P, self._element_ ** exponent)
+        try:
+            result = P.element_class(P, self._element_ ** exponent)
```
fchapoton: why recompute the power ? and maybe use "return" in the try ?

behackl: You are absolutely right, I changed the code in haste and missed replacing the argument. Fixed via c431e2cd15f.

fchapoton (approved):
ok, looks good enough, although maybe not very kosher in term of the coercion model

PR #38961, opened 2024-11-13
Iteration over infinite abelian groups

### LINKED ISSUE #30751: Iteration over infinite abelian groups
<div id="comment:0"></div>

Given an AbelianGroup with infinite order generators the `__iter__` method silently returns an empty iterator.

```
sage: for a in AbelianGroup(1,[0]):
....:     print(a)
```
gives no output.

Either an error should be raised (like for the list method), or ideally an iterator over all elements (by lexicographic order of the powers of generators should be returned, as for

```
sage: for a in ZZ:
....:     print(a)
....:     if a^2 + a == 30:
....:         break
```

Component: **group theory**

_Issue created by migration from https://trac.sagemath.org/ticket/30751_

DaveWitteMorris:
Fixes #30751.

This PR adds code to the `__iter__` method of `AbelianGroup` to handle infinite groups. For backward compatibility (and because the finite case is simpler and probably more efficient), the code for the finite case has been retained.

### :memo: Checklist


src/sage/groups/abelian_gps/abelian_group.py
```diff
@@ -1301,10 +1301,27 @@ def __iter__(self):
...
+        if 0 not in invs:
+            # The group is finite
+            for t in mrange(invs):
+                yield self(t)
+        else:
+            # A similar approach works for infinite groups.
+            # (This would also work for finite groups, but is more complicated.)
+            from sage.misc.mrange import cantor_product
+            for t in cantor_product(*[range(n) if n > 0 else ZZ for n in invs]):
+                yield self(t)
```
mantepse: I think that this can be simplified slightly:
```suggestion
        if 0 not in invs:
            # The group is finite
            yield from map(self, mrange(invs))
        else:
            # A similar approach works for infinite groups.
            # (This would also work for finite groups, but is more complicated.)
            from sage.misc.mrange import cantor_product
            yield from map(self, cantor_product(*[range(n) if n else ZZ for n in invs]))
```

I suppose that negative orders should not be allowed, although:
```
sage: G = AbelianGroup([-4])
sage: G.cardinality()
-4
```

DaveWitteMorris: I agree that your code is an improvement, so I made the changes. (At least, I think I did, but I don't really understand github.)

DaveWitteMorris: I opened issue #38967 to correct the problem with negative orders.

mantepse (approved):
Perfekt!

DaveWitteMorris:
Thanks for the review and your comments!

PR #38966, opened 2024-11-13
Clean up Cygwin remnants

orlitzky:
We dropped support for Cygwin a while ago, but there is some stuff left over in the tree. Let's finally remove it.


kwankyu (approved):
Thanks. I agree that there is a zero chance that we would look back.

LGTM.

PR #38972, opened 2024-11-15
Fix build of meataxe extension with meson

antonio-rojas:
meataxe is not a library, the library is called mtx and is already searched for


antonio-rojas:
tdlib and mcqd are also not found because they are header-only, so `find_library` doesn't work.

tobiasdiez (approved):
Thanks!

tobiasdiez:
> tdlib and mcqd are also not found because they are header-only, so `find_library` doesn't work.

They also don't seem to provide pkg-config info. So just using `cc.has_header` to check for the required headers, and if then set the dependency object to `disabler()` if not found?

antonio-rojas:
> > tdlib and mcqd are also not found because they are header-only, so `find_library` doesn't work.
> 
> They also don't seem to provide pkg-config info. So just using `cc.has_header` to check for the required headers, and if then set the dependency object to `disabler()` if not found?

Done in https://github.com/sagemath/sage/pull/38984

antonio-rojas:
Setting back to positive review, it was a simple rebase

PR #38973, opened 2024-11-16
is_t_design needs to verify that every t-tuple is in at least one block

### LINKED ISSUE #38454: `is_t_design` method of the class `IncidenceStructure` sometimes returns wrong answer
### Steps To Reproduce

All you have to do is to activate SageMath.

### Expected Behavior

```sage
sage: IS = IncidenceStructure(points=[0,1,2,3,4,5], blocks=[[0,1], [1,2], [0,2]])
sage: print(IS.is_t_design(return_parameters=True))
(False, (0,0,0,0))
```
or
```sage
sage: IS = IncidenceStructure(points=[0,1,2,3,4,5], blocks=[[0,1], [1,2], [0,2]])
sage: print(IS.is_t_design(return_parameters=True))
(True, (2, 3, 2, 1))
```

### Actual Behavior

```sage
sage: IS = IncidenceStructure(points=[0,1,2,3,4,5], blocks=[[0,1], [1,2], [0,2]])
sage: print(IS.is_t_design(return_parameters=True))
(True, (1, 6, 2, 1))
```
### Additional Information

_No response_

### Environment

```markdown
- **OS**: Ubuntu 22.04.4 LTS
- **Sage Version**:9.5
```

### Checklist

- [x] I have read the documentation and troubleshoot guide

DaveWitteMorris:
Fixes #38454.

In a t-design, every t-tuple needs to be contained in the same number of blocks. The code for `is_t_design` did not verify this correctly, because it only checked the t-tuples that are contained in at least one block. We fix this by verifying that the number of t-tuples that appear in at least one block is equal to the total number of t-tuples.

### :memo: Checklist


dcoudert (approved):
It took me some time to understand why each $(t=0)$-subset of points is indeed covered by $\lambda = 3$ blocks, and so to check the correctness of the test.

LGTM.

DaveWitteMorris:
Thanks for the review!

I agree that 0-designs are surprising (definitions of "t-design" in the literature often assume t is at least 2), so I'll explain this for anyone who reads this thread in the future.

The empty set is the only 0-subset. This means the 0-subset is unique, so it is is trivially true that all 0-subsets are contained in the same number of blocks.  Hence, this condition in the definition of t-design is automatically satisfied when t = 0. This means that in order for a collection of blocks to be a 0-design, the only requirement is that all of the blocks have the same cardinality. (Also note that the empty set is a subset of every set, so the number λ of subsets containing the 0-subset is equal to the total number of blocks.)

PR #38974, opened 2024-11-16
allow incomparable labels in SetPartition

### LINKED ISSUE #29996: SetPartition prefers comparable labels
<div id="comment:0"></div>

```
sage: SetPartition([["1", 2]])
...
/home/martin/sage-develop/local/lib/python3.7/site-packages/sage/combinat/set_partition.py in __init__(self, parent, s, check)
    567         """
    568         self._latex_options = {}
--> 569         ClonableArray.__init__(self, parent, sorted(map(frozenset, s), key=min), check=check)
    570 
    571     def check(self):
...
TypeError: unsupported operand parent(s) for >: 'Integer Ring' and '<class 'str'>'
```

Component: **combinatorics**

_Issue created by migration from https://trac.sagemath.org/ticket/29996_

mantepse:
Fixes #29996


src/sage/combinat/set_partition.py
```diff
@@ -620,7 +624,11 @@ def __init__(self, parent, s, check=True):
             {}
         """
         self._latex_options = {}
-        ClonableArray.__init__(self, parent, sorted(map(frozenset, s), key=min), check=check)
+        try:
+            s = sorted(map(frozenset, s), key=min)
+        except TypeError:
```
fchapoton, line 629: comparing using min of string looks strange.. 
```
sage: min(str((3,4)))
' '
```

mantepse: We shouldn't be applying `min` to a single string here, but rather to an iterable of strings.
```
sage: min(";-)")
)
sage: min([";-)"])
';-)'
```

fchapoton (approved):
ok, good enough

PR #38975, opened 2024-11-16
support left and right actions for species

mantepse:
As in other modules, we allow the user to specify whether the group action is a left or a right action. In line with permutation groups, and also gap, the default side for an action is from the right.


tscrim (approved):
LGTM.

PR #38982, opened 2024-11-17
Fix meson build by adding missing files

tobiasdiez:
Fixes the meson build.

### :memo: Checklist


kwankyu:
If this does not fix "Build & Test using Meson", please consider turn it off to be information only.

kwankyu (approved):
LGTM.

tobiasdiez:
Thanks for the review!

> If this would not fix "Build & Test using Meson", please consider turn it off to be information only.

I understand the frustration. Disabling the workflow would however be counter productive since meson is supposed to be the replacement of the old setuptools-based build. Would love to hear your thoughts on how to archive that without making the transition period to annoying. (I will also open a discussion on sage-dev about this in the coming days.)

PR #38984, opened 2024-11-17
Fix detection of tdlib and mcqd with meson

antonio-rojas:
These are header-only libraries, so find_library doesn't work with them


tscrim (approved):
LGTM.

PR #38986, opened 2024-11-17
Allow CRT_list() to be called with one argument

user202729:
As in the title.

Opinion: should we allow `crt()` / `CRT()` to be called with one list also? Note that currently it is callable with two lists and does the exact same thing as `CRT_list()`.

Opinion: is returning `Mod(0, 1)` for empty list a sane idea? Problem is later we may want to implement (pseudocode)

```python
CRT_list([mod(x-1, x^2+1), mod(x+1, x^2-1)])
```

and in that case the correct thing to return is not clear when the list is empty.

More generally, if it's given elements `x₁ ∈ R/I₁, …, xₙ ∈ R/Iₙ` where `Iᵢ = xᵢ.parent().defining_ideal()` and all `R = xᵢ.parent()` are the same for all `i`, it should returns `x ∈ R/(I₁.intersection(I₂).intersection(I₃).….intersection(Iₙ))`. But this is not well-defined when `values` is empty.

### :memo: Checklist


user202729:
Any idea why the `TESTS:` show up in the documentation?

src/sage/arith/misc.py
```diff
@@ -3575,14 +3600,26 @@ def CRT_list(values, moduli):
         sage: ms
         [5, 7, 9]
     """
-    if not isinstance(values, list) or not isinstance(moduli, list):
+    if not isinstance(values, list) or not isinstance(moduli, (list, type(None))):
```
tscrim:
```suggestion
    if not isinstance(values, list) or (moduli is not None and not isinstance(moduli, list)):
```
This is the standard way in Python as I understand it..

src/sage/arith/misc.py
```diff
@@ -3498,11 +3497,15 @@ def crt(a, b, m=None, n=None):
 CRT = crt
 
 
-def CRT_list(values, moduli):
+def CRT_list(values, moduli=None):
     r""" Given a list ``values`` of elements and a list of corresponding
     ``moduli``, find a single element that reduces to each element of
     ``values`` modulo the corresponding moduli.
```
tscrim:
```suggestion
def CRT_list(values, moduli=None):
    r"""
    Given a list ``values`` of elements and a list of corresponding
    ``moduli``, find a single element that reduces to each element of
    ``values`` modulo the corresponding moduli.
```
This is a while-we-are-at-it thing to fix.

src/sage/arith/misc.py
```diff
@@ -3498,11 +3497,15 @@ def crt(a, b, m=None, n=None):
...
 
+    This function can also be called with one argument, each element
+    of the list is a :class:`IntegerMod_abstract` object. In this case, it returns
+    another :class:`IntegerMod_abstract` object.
```
tscrim:
```suggestion
    of the list is a :class:`modular integer <IntegerMod_abstract>`. In this case, it returns
    another :class:`modular integer <IntegerMod_abstract>`.
```
Trying to be a little less technical. Not sure if it is an improvement though.

tscrim (changes requested):
A few small things.

src/sage/arith/misc.py
user202729: The link should go to https://doc.sagemath.org/html/en/reference/finite_rings/sage/rings/finite_rings/integer_mod.html (I need to fix the link later). The elements aren't called modular integers there either, they're just called elements of ℤ/nℤ.

Edit: actually "modular integer" is mentioned exactly twice (once in `lucas` method and another in libs/giac/giac.pyx).

src/sage/arith/misc.py
user202729: Existing code does seem to do that. There's also `from types import NoneType`, but none of the existing code does that.

tscrim (approved):
Sorry for loosing track of this one. LGTM. Thanks.

user202729:
Actually I realize this is already supported by a (hard to find) function `sage.rings.finite_rings.integer_mod_ring.crt`.

It seems like bad practice to have this function also named `crt` does a completely different thing however. (besides, this function seems slightly slower because it doesn't do the binary divide and conquer thing. But being implemented in cython instead of python might lead to it being slightly faster…?)

Maybe we can just deprecate this function pointing to `CRT_list` after this pull request. Performance optimization can be considered later.

tscrim:
Indeed, let's look at removing the duplication on a followup PR/issue.

PR #38989, opened 2024-11-17
tentative of free-Algebra functor

fchapoton:
This is trying to create a functor for free Associative unital algebras.

Using as a model the case of free Pre-Lie algebras.

### :memo: Checklist


fchapoton:
So far, the merge is not working as expected for the coercion

fchapoton:
probably one needs to fix that also :
```
sage: A = algebras.Free(QQ,['x','y'])
sage: B = algebras.Free(QQ,['y'])
sage: A.has_coerce_map_from(B)
False
```

tscrim:
The problem is the `FreeAlgebra` does not like the `Alphabet` input:
```
sage: FreeAlgebra(QQ, ['x','y'], None)
Free Algebra on 2 generators (x, y) over Rational Field
sage: FreeAlgebra(QQ, Alphabet(['x','y']), None)
...
ValueError: variable name "{'x'" is not alphanumeric
```
So changing
```
-            return AssociativeFunctor(Alphabet(ret))
+            return AssociativeFunctor(tuple(ret))
```
fixes the problem for me.

fchapoton:
oh, damn.. Great ! Thanks a lot. I still have to care a little more for the degrees.

tscrim:
We probably should let the free algebra constructor take an alphabet (or anything list like) with special processing for strings. Well, that is older code IIRC, and might be in need of an overhual...:/

tscrim (approved):
LGTM.

PR #38994, opened 2024-11-18
Allow copying algebraic closure of finite field

### LINKED ISSUE #38988: algebraic_closure of finite field cannot be copied
### Steps To Reproduce

```
sage: deepcopy(GF(5).algebraic_closure())
...
TypeError: sage.misc.fast_methods.WithEqualityById.__new__(AlgebraicClosureFiniteField_pseudo_conway_with_category) is not safe, use AlgebraicClosureFiniteField_pseudo_conway_with_category.__new__()
```

### Expected Behavior

should just work

### Actual Behavior

 error

### Additional Information

_No response_

### Environment

- **OS**: Linux
- **Sage Version**: 10.5.β9

### Checklist

user202729:
Fixes https://github.com/sagemath/sage/issues/38988

I choose to make the copy return itself, because that's what `GF(p)` does at the moment anyway, and the field is immutable.

Not sure how much of a hack for me to implement `__copy__` and `__deepcopy__` instead of `__reduce__`, but pickling and unpickling at the moment works well anyway (although it returns a different instance)

### :memo: Checklist


JohnCremona:
@nbruin ?  I don't know about copying

nbruin:
I'm not sure what the copying behaviour should be for this structure. Reading the intro of the file, there is a very particular reason why the algebraic closure is *not* UniqueRepresentation. It is EqualityById, which looks like it's not a problem.
This explains why unpickling produces a new instance ... although
```
k=GF(5)
kbar=k.algebraic_closure()
GF(5).algebraic_closure() == kbar ##True
```
suggests that indirectly there IS a preferred copy: the one cached on the corresponding prime field.

The fact that multiple copies can exist would suggest that at least deep_copy should be able to produce a separate copy. Perhaps check with Peter Bruin @pjbruin ?

At the moment, only prime finite fields have algebraic closures. As far as I can see, prime finite fields are not formally UniqueRepresentation, but producing a finite field from `GF(p)` twice does produce identical results. So similarly, perhaps its algebraic closure is effectively also UniqueRepresentation, in which case we might want to fix pickle and then it wouldn't be controversial for copy to just return self.

user202729:
If I understood correctly, the problem is we cannot really fix pickle, otherwise if we dump an element to the disk, exit Sage, reopen it (and thus the computed pseudo-conway_polynomial is different), then there's no consistent way to convert the unpickled element to an element of the (new) `GF(p).algebraic_closure()`.

nbruin:
Choosing pseudo-conway polynomials in a deterministic way would probably allow solving the pickling problems. I don't immediately see why that wouldn't work for existing pickles (do those pickles store the computed Conway polynomials?), but pickle has a "version" system that allows you to recognize old pickles, that you can then reconstruct using a legacy method that you need to register somewhere.

Changing to another implementation can be done with deprecation: just write and install the new code under other names, have a deprecation period where people get a warning if they use the old routines, pushing them to the new (so that would be `algebraic_closure` and `algebraic_closure_new` or something like that. After the deprecation period, the old code can be removed (or renamed to `algebraic_closure_conway` or something like that). Once the names are freed up, you could alias the new code to the original names. If you so desire, you can deprecate the temporary "new" names. That process takes something like 6 years.

Or if you feel that the benefits of the change are large enough to inconvenience some hypothetical users that depend on conwayness then just make the change. But perhaps check beforehand if any of them are vocal.

user202729:
Actually… if pseudo-Conway polynomials are really randomly chosen, then finite fields with degree greater than 1 cannot be UniqueRepresentation either. Yet:

```python
sage: deepcopy(GF(5^2))
Finite Field in z2 of size 5^2
sage: deepcopy(GF(5^2)) is GF(5^2)
True
sage: loads(dumps(GF(5^2))) is GF(5^2)
True
```

I think modifying pickle return the algebraic closure itself would be just fine.

pjbruin:
I can think of three "obvious" ways to copy `GF(p).algebraic_closure()`:
1. return the object itself;
2. return another `AlgebraicClosureFiniteField_pseudo_conway` object with the parameter `lattice` equal to the pseudo-Conway lattice of the orginal;
3. return another `AlgebraicClosureFiniteField_pseudo_conway` object with a newly constructed pseudo-Conway lattice.

I guess it would be fine for `__copy__()` to return either of the first two; `__deepcopy__()` could perhaps return the third one. That option behaves differently in the sense that the resulting pseudo-Conway lattices could give rise to different defining polynomials for the finite subfields. If we would make `__copy__()` and/or `__deepcopy__()` use the second option, it might be wise to modify `AlgebraicClosureFiniteField_pseudo_conway` to no longer be `WithEqualityById`, but to compare the corresponding pseudo-Conway lattices.

Issue #38373, opened 2024-07-15
No canonical conversion from ℚ to ‾𝔽ₚ?


user202729:
### Steps To Reproduce

```python
p = 2**127 - 1
F = GF(p)

F.algebraic_closure()(F(5/2))  # okay
F.algebraic_closure()(5/2)  # error?
```

### Expected Behavior

The last line also works.

### Actual Behavior

The last line errors out.

Side note, the motivation is the following.

```python
p = 2**5-1
F = GF(p)
E = EllipticCurve(F, [0, 6, 0, 1, 0])
E.change_ring(F.algebraic_closure()).isogenies_prime_degree(5)  # errors?
```

### Checklist

- [X] I have searched the existing issues for a bug report that matches the one I want to file, without success.
- [X] I have read the documentation and troubleshoot guide

JohnCremona:
I agree that there should be a conversion from QQ to any field of characteristic p which works for rationals with non-negative valuation, i.e. with denominator prime to p.

fchapoton:
The error happens when it tries to **coerce** the rational in the ground field.

PR #38995, opened 2024-11-18
Isogeny small degree improvement

user202729:
Allow the following to be called without error.

```sage
sage: EllipticCurve(GF(31).algebraic_closure(), [0, 6, 0, 1, 0]).isogenies_prime_degree(5)
```

Inspired from, but independent from, https://github.com/sagemath/sage/issues/38373 . Whatever consensus on that issue might be.

------

Would be good if someone can sanity check that the output is actually correct. (although it looks correct to me.)

Strangely enough there is

```python
        if isinstance(F, sage.rings.abc.AlgebraicField):
            raise NotImplementedError("This code could be implemented for QQbar, but has not been yet.")
```

even though when I comment that part out, the code seems to work well.

### :memo: Checklist


JohnCremona:
About correctness: looks reasonable to me.  Over GF(31) there are only 2 5-isogenies, the others are defined over GF(31^2) and the output from that looks just like this output.  Perhaps it would be safer to have the doctest just report that there are six 5-isogenies, to guard against future versions making different choices?  My familiarity with the supersingular case is less than perfect!  I note that of the 6 isogenous curves there are only 3 isomorphism classes, E itself (@3) another @2 and another unique.

I think it is great that this works as it should making only a trivial change to the code in isogenies_small_degree which I wrote years ago.

Can you explain the need for the new copy methods?

user202729:
The copy is from https://github.com/sagemath/sage/pull/38994 . Basically it's needed to make it work at all because of some implementation detail.

Specifically, some part of the code requires making a new isogeny with a different pre-composed isomorphism, so it calls the internal method `_set_pre_isomorphism` (which mutates the isogeny (!)), but it wants a modified copy so it does `deepcopy` on the isogeny. Without special handling, `deepcopy` on the isogeny copies all the members, including the base field.

I think it's desirable for arbitrary objects to be copyable anyway, so I can't see any harm. (Or at least if it's desirable to make algebraic closure of finite field **not** copyable then at least raise a meaningful error, but at the moment `loads(dumps(GF(p).algebraic_closure()))` makes a copy anyway)

It might separately be a good idea to refactor the code to avoid mutable state in isogeny objects (easy to introduce accidental bugs), but currently it's just internal implementation detail anyway.

JohnCremona (approved):

S17A05:
Since this PR contains the code from #38994, which does not have its own positive review yet, it probably makes sense to close that PR and integrate the "Fixes #38988" into this PR to avoid clutter/confusion

user202729:
Since figuring out what to do with Conway polynomial and non-unique representation of algebraic closure of finite field might take a long time, I decide to modify the thing to just use shallow copy instead.

As far as I can see this doesn't cause any trouble (the whole reason we need the copy in the first place is because `_set_post_isomorphism()` mutates the object, but it only mutates direct members, so it should be safe. If for some reason some future implementer `__copy__` method can always be implemented for `EllipticCurveIsogeny` that properly copies that part)

JohnCremona:
About the pre- and post-isomorphisms:  the original implementation of isogenies was done by someone who was only interested in curves over finite fields, who wanted all isogenies to be normalised (i.e. inverse image of the standard invariantdifferential on the codomain is exactly the standard differential on the domain).  In reviewing that original code I pointed out that it all worked fine over arbitrary fields, for example over QQ and over number fields, but that in such situtations we don't want to restrict to normalised isogenies as instead we want to normalise the curve models, e.g. making them integral and globally inimal where possible.  That led to the pre and post-isomorphism stuff being tacked on.  If starting from scratch now, I am sure it would look different.

JohnCremona:
As I gave this a positive review several months ago it would be helpful for someone to tell me what now needs to be checked, as otherwise I'll have to start again.

JohnCremona:
@user202729 Why is your username anonymous?

user202729:
> As I gave this a positive review several months ago it would be helpful for someone to tell me what now needs to be checked, as otherwise I'll have to start again.

Several months ago you positive-review the implementation that just makes algebraic closure of finite field an identity.

In https://github.com/sagemath/sage/pull/38994 no consensus was reached (as far as I can tell), so I decide to change the implementation here that avoids the `deepcopy` entirely. Doing that turns out to be surprisingly difficult and requires a rather large modification.

The current situation is (as far as I remember… it's been a few months ago)

* `_set_pre_isomorphism` and `_set_post_isomorphism` is gone, instead we have `_compose_with_isomorphism` which returns a modified copy. So isomorphism is "more" immutable now.
   * consequently, `_compose_with_isomorphism` needs to do some ugly monkey patching thing to copy the attributes over. (`deepcopy` doesn't work.)
   * at least the positive side is `__clear_cached_values` is gone now, since isomorphism is immutable except at object construction.
* Of course it still need to be set at object construction, so `__set_pre_isomorphism` and `__set_post_isomorphism` remains.

I really can't see a better way to implement it without  https://github.com/sagemath/sage/pull/38994 being available.

JohnCremona:
Thanks for the explanation.

I think that the changes around line 330 (where the first error occurs in the current Sage version) are fine, but an alternative is to change line 325 to jt=f/t.   (Explanation:  Fricke_module(ell) = Fricke_polynomial(ell)/t and Fricke_polynomial(ell) is a monic polynomial in ZZ[t], of degree ell+1, for ell in [2,3,5,7,13].)

So in the definition of the polynomial f (defined in line 330) you could replace jt by f and kt by (f-1728*t) and leave out the coercion into R.

I think this is easier than what the current code does (my bad) which is to factor two rational functions in QQ(t) whose denominator is t and whose numerators are monic integer polynomials, pick out the factors of multiplicity 3 and 2 (which are certainly monic integer polynomials), multiply them together, getting a result whose parent is QQ[t] but which lies in ZZ[t], the coerce into ZZ[t] and then into F[t]!

Of course that is the easy part.  Your work on getting the pre and post isomorphisms to work over this field was harder.

PR #38997, opened 2024-11-18
    Document RealIntervalField default printing configuration

user202729:
As in the title.

Side note, is this actually a good idea? Maybe it was intended to be not modifiable… 

But modifying it has a practical advantage, sometimes you cannot control these two option otherwise e.g. if you have a list of elements you may need to extract each element and call `.str(...)` on it.

### :memo: Checklist

### DIFF
--- a/src/sage/rings/real_mpfi.pyx
+++ b/src/sage/rings/real_mpfi.pyx
@@ -120,6 +120,26 @@ satisfying, but we have chosen the latter.
     sage: a == 2
     False
 
+Some default printing options can be set by modifying module globals::
+
+    sage: from sage.rings import real_mpfi
+    sage: x = RIF(sqrt(2), sqrt(2)+1e-10); x
+    1.4142135624?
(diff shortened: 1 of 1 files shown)


tscrim (approved):
LGTM.

user202729:
The problem with this idea is… if later we want to move it to be parent-specific (like `sci_not` at the moment), then we will need a long deprecation period.

And it isn't entirely clear how to deprecate this feature either. We can't detect when the user accesses or modifies that specific field — because there's no `__getattr__`/`__setattr__` module-level available.

tscrim:
IMO the `sci_not` should be a global option. It doesn't affect the fundamental behavior of the objects, just their printing. This extends to all options that control the printing of objects. (Really, these should be moved to the current `GlobalOptions` class.)

It's easy enough to deprecate the class input. For the module-level stuff, we could default it to `None` and in the `_repr_`, then we print the deprecation. It doesn't print it when it is accessed, but at least when it is used.

PR #39007, opened 2024-11-20
Fix cpu time on Windows

tobiasdiez:
Currently, it uses `resources` which is not available on Windows. For cross-platform purposes, we change it to use `time.process_time()` instead.

Also add type hints and minor reformating.

### :memo: Checklist


orlitzky:
Does _psutil_ work on Windows?

`Timer._quick_cputime()` in `sage.doctest.util` is a much better approach than hoping that every subprocess will be able to reliably report its CPU usage. It's accurate on Linux even if your subprocesses are badly behaved, and should work anywhere else if _psutil_ happens to be installed and supports the OS.

psutil isn't a dependency of the sage library, only of sage-the-distro, but pyproject.toml supports per-platform dependencies now.

If we wanted to replace `cputime(..., subprocesses=True)` without any risk of breakage, we could require psutil for sagelib on macOS and Windows, and then switch `cputime()` to use the `_quick_cputime()` implementation. That would allow us to delete the handwritten cputime calculations for gap, gp, macailay2, magma, maple, maxima, mupad, and singular.

tobiasdiez:
> Does _psutil_ work on Windows?

Yes, at least it claims so.

> If we wanted to replace `cputime(..., subprocesses=True)` without any risk of breakage, we could require psutil for sagelib on macOS and Windows, and then switch `cputime()` to use the `_quick_cputime()` implementation. That would allow us to delete the handwritten cputime calculations for gap, gp, macailay2, magma, maple, maxima, mupad, and singular.

That sounds like a good idea but I'm not familiar enough with all the implications (especially on unix) to make these changes myself.

orlitzky:
As an intermediate step I think it makes more sense to,

 1. factor out `sage.doctest.util.Timer._quick_cputime()`
 2. Add a psutil dependency to sagelib for Windows
 3. Switch the `cputime(subprocesses=True)` implementation to `_quick_cputime()` on Windows

The background to all of this is that the existing `cputime(subprocesses=True)` implementation is not reliable. It basically asks each pexpect process to report its own cputime. Some of them get it right, but others just report zero and will never be fixed, or hang if the process is in some error state,

When I was switching the doctest framework to report cputime, I couldn't rely on that mess, so I created `_quick_cputime()`. On linux and freebsd at least, you can just read the relevant information from `/proc`, so that's what `_quick_cputime()` tries first. On other systems, it falls back to psutil, which also gives reliable answers. This works on sage-the-distro because psutil is a standard package, and it should "accidentally" work elsewhere (even for e.g. meson installs) because ipython pulls psutil in as a dependency. I have not lost any sleep over this minor discrepancy because the default `cputime(subprocesses=True)` method was already flaky.

The summary is that `_quick_cputime` is simply a better `cputime(subprocesses=True)`, but that it's only used in the doctest framework right now, and the one outstanding issue it has is that sagelib does not actually depend on psutil on the platforms that need it. So in this case I am suggesting that (on Windows only) we switch the implementations and add the platform dep. Eventually the other platforms can be migrated in the same way. Then somewhere down the road we could deprecate and remove the hacky `subprocesses=True` attempt. But for now, you can get quick and reliable cputime measurements by just asking psutil, and `_quick_cputime` will already try that for you.

tobiasdiez:
> Switch the cputime(subprocesses=True) implementation to _quick_cputime() on Windows

Note that the only change in this PR is in the `subprocesses = False` branch. To be honest, at the current state I actually don't care at all about correct numbers for windows. My current problem is that `import sage.all` fails on windows with a bunch of errors, one of them being `import resources`. So there is a very long path to go until the correct cputime readings become the most pressing concern ;-)

(I'm also not planning to implement `sage -t` support on windows but will directly try to make pytest work.)

orlitzky:
OK, we can deal with the rest later.

PR #39010, opened 2024-11-20
Don't import cysignals.alarm on Windows

tobiasdiez:
`cysignals.alarm` is not available on Windows, so we don't try to import it. Probably more work needs to be done to actually fix the methods that make use of these methods on Windows; but that's left for further PRs.

Also remove a `cysignals` import from an `all_xyz.py` file that was duplicated in the main `all.py` file.

### :memo: Checklist


dimpase (approved):
LGTM. Funny that `C:MICROS~1` still calls its platform `w32`

PR #39011, opened 2024-11-20
Correctly print import errors in `sage.all`

tobiasdiez:
When there is an error during importing `sage.all`, then it is just hidden but sage's are no longer available. This is very hard to debug and misleading during development. For this reason the fallback to `repl.all` is removed.

Furthermore, a ruff exception for "unused imports" is added for all files, since these files always contains many unused imports.

### :memo: Checklist


user202729:
It was first introduced in 787ddd9bc5cff4be18ea082a453cce4c32d6d471 though no rationale was given. Maybe it is part of the modularization?

tobiasdiez:
> It was first introduced in [787ddd9](https://github.com/sagemath/sage/commit/787ddd9bc5cff4be18ea082a453cce4c32d6d471) though no rationale was given. Maybe it is part of the modularization?

Yes, `sage.all__sagemath_repl` is part of the modularization. However, it only gives an empty python environment (supporting the sage language constructs). So it's pretty useless. If in the future there are more modularized distributions that integrate with the sage shell, then I am happy to readd a construction like this - but for the moment it only hides important errors without giving any benefit.

dimpase (approved):
lgtm

tobiasdiez:
Thanks!

PR #39012, opened 2024-11-20
Remove SAGE_DB from public interface

tobiasdiez:
The `SAGE_DB` variable was never used (as far as I can see). So we remove it (as part of the process of migrating the databases to separate python packages).

### :memo: Checklist


user202729:
Maybe some user (existing code) are using it? The function `db` and `db_save` are exported to the user afte rall.

orlitzky:
I don't think `db()` and `db_save()` have anything to do with the database packages, it looks like they're just shortcuts to pickle/unpickle without providing a path (because it will use `SAGE_DB` as the path behind the scenes). Seems silly to me to reimplement `import pickle` in the first place, but whatever.

The comment in the graphs file looks obsolete. And we can certainly remove `SAGE_DB` from the public interface.

And maybe some day deprecate / separate all of this random stuff that has nothing to do with the rest of sage or mathematics.

tobiasdiez:
> Maybe some user (existing code) are using it? The function `db` and `db_save` are exported to the user afte rall.

I couldn't find any usage (using github's search). I'm tempted to read `sage.misc` as "internal".

@orlitzky: Not sure if I should read your comment as a suggestion to change something or as a positive review ;-). Could you clarify what you mean with "The comment in the graphs file looks obsolete."

orlitzky:
> Could you clarify what you mean with "The comment in the graphs file looks obsolete."

The graphs database is now saved to `sage.env.GRAPHS_DATA_DIR`, so the comment in `isgci.py` about `SAGE_DB` is misleading.

I don't think we should remove `db()` and `db_save()` without deprecation. It looks to me like they're thin wrappers around pickling and unpicking, and, even though I think that's pretty pointless, it looks like they were intended for end users at some point.

orlitzky (approved):

tobiasdiez:
Thanks @orlitzky. I hope you are okay with me setting it officially to positive review.

